PR #10559, opened 2019-09-20
FocusTrapZone: Add aria-modal attribute

### LINKED ISSUE #6535: FocusTrapZone does not trap focus when narrator is on with scan mode and arrow keys are navigating
 
Use the demo page in edge:
https://developer.microsoft.com/en-us/fabric#/components/focustrapzone
Turn on a focus trap zone
Turn on narrator with scan mode on (caps lock + spacebar to toggle scan mode)
From inside a FocusTrapZone, use the up or down key to navigate

### Bug Report 
 
- __Package version(s)__: current
- __Browser and OS versions__: edge- latest, windows - latest
 
Requested priority: Normal
 
Products/sites affected: PQ Online
 
#### Describe the issue:
 
Repro shown above
 
#### Actual behavior:
 
Focus can leave trap zone
 
#### Expected behavior:
 
Focus cannot leave trap zone

JasonGore:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #6535

#### Description of changes

Add `aria-modal` attribute to `FocusTrapZone`. Remove `aria-modal` attribute from higher order components (Panel, Modal) since they are now redundant. 

Narrator as yet does not seem to make use of `aria-modal` to affect navigation (per #6535), but this PR more correctly aligns application of `aria-modal` to Fabric's component hierarchy in general. 

#### Focus areas to test

Update snapshots.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10559)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
+++ b/packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
@@ -93,6 +93,7 @@ export class FocusTrapZone extends React.Component<IFocusTrapZoneProps, {}> impl
         className={className}
         ref={this._root}
         aria-labelledby={ariaLabelledBy}
+        aria-modal="true"
         onFocusCapture={this._onFocusCapture}
         onFocus={this._onRootFocus}
         onBlur={this._onRootBlur}
--- a/packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
@@ -221,7 +221,6 @@ export class ModalBase extends BaseComponent<IModalProps, IDialogState> implemen
         <Layer {...mergedLayerProps}>
           <Popup
             role={isModeless || !isBlocking ? 'dialog' : 'alertdialog'}
-            aria-modal={!isModeless}
             ariaLabelledBy={titleAriaId}
             ariaDescribedBy={subtitleAriaId}
             onDismiss={onDismiss}
--- a/packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
@@ -183,7 +183,6 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
       <Layer {...layerProps}>
         <Popup
           role="dialog"
-          aria-modal="true"
           ariaLabelledBy={header ? headerTextId : undefined}
           onDismiss={this.dismiss}
           className={_classNames.hiddenPanel}
--- a/change/office-ui-fabric-react-2019-09-20-12-59-47-jg-6535-ftz-aria-modal.json
+++ b/change/office-ui-fabric-react-2019-09-20-12-59-47-jg-6535-ftz-aria-modal.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "FocusTrapZone: Add aria-modal attribute.",
+  "packageName": "office-ui-fabric-react",
+  "email": "jagore@microsoft.com",
+  "commit": "bd8f150e61169ed814584042a59eeef5017154cf",
+  "date": "2019-09-20T19:59:47.600Z"
+}
--- a/packages/office-ui-fabric-react/src/components/HoverCard/__snapshots__/HoverCard.test.tsx.snap
+++ b/packages/office-ui-fabric-react/src/components/HoverCard/__snapshots__/HoverCard.test.tsx.snap
@@ -113,6 +113,7 @@ exports[`HoverCard renders ExpandingCard correctly 1`] = `
           }
         >
           <div
+            aria-modal="true"
             onBlur={[Function]}
             onFocus={[Function]}
             onFocusCapture={[Function]}
@@ -276,6 +277,7 @@ exports[`HoverCard renders PlainCard correctly 1`] = `
           }
         >
           <div
+            aria-modal="true"
             onBlur={[Function]}
             onFocus={[Function]}
             onFocusCapture={[Function]}
--- a/packages/office-ui-fabric-react/src/components/Modal/__snapshots__/Modal.test.tsx.snap
+++ b/packages/office-ui-fabric-react/src/components/Modal/__snapshots__/Modal.test.tsx.snap
@@ -54,7 +54,6 @@ exports[`Modal renders Draggable Modal correctly 1`] = `
     onSubmit={[Function]}
   >
     <div
-      aria-modal={false}
       onKeyDown={[Function]}
       role="dialog"
       style={
@@ -88,6 +87,7 @@ exports[`Modal renders Draggable Modal correctly 1`] = `
             }
       >
         <div
+          aria-modal="true"
           className=
               ms-Dialog-main
               test-containerClassName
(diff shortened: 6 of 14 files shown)


packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -93,6 +93,7 @@ export class FocusTrapZone extends React.Component<IFocusTrapZoneProps, {}> impl
         className={className}
         ref={this._root}
         aria-labelledby={ariaLabelledBy}
+        aria-modal="true"
```
jspurlin: >aria-modal="true" [](start = 8, length = 17)

aria-modal only applies to window roles (alertdialog and dialog), see: [https://www.w3.org/TR/wai-aria-1.1/#aria-modal](https://www.w3.org/TR/wai-aria-1.1/#aria-modal) (the "used in roles" section). that being said, there's divProps on the FocusTrapZone so a user could already pass aria-modal in

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
```diff
@@ -221,7 +221,6 @@ export class ModalBase extends BaseComponent<IModalProps, IDialogState> implemen
         <Layer {...mergedLayerProps}>
           <Popup
             role={isModeless || !isBlocking ? 'dialog' : 'alertdialog'}
```
jspurlin: >role={isModeless || !isBlocking ? 'dialog' : 'alertdialog'} [](start = 12, length = 59)

I'm glad we're removing aria-modal, but we should also remove the role.... In reality the Modal should be wrapper that hosts things and isn't anything (like) by itself... it's a vehicle for hosting a dialog, not the dialog itself

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -183,7 +183,6 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
       <Layer {...layerProps}>
         <Popup
           role="dialog"
-          aria-modal="true"
```
jspurlin: >ria-modal="true" [](start = 11, length = 16)

Is the panel the object or is it just a vehicle (I believe in this case it is the object)? If it is the object, I feel like this (or whatever element that's the actual Panel, not including the layer/overlay... just the container of the contents) is the one that should have the role and aria-modal

packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
JasonGore: The applicable text from that link says:
> The aria-modal attribute is used to indicate that the presence of a "modal" element precludes usage of other content on the page. For example, when a modal dialog is displayed, it is expected that the user's interaction is limited to the contents of the dialog, until the modal dialog loses focus or is no longer displayed.

They use modal in quotes and refer to modal-like elements rather than exactly and specifically the `modal` role. In this case, FocusTrapZone (and other higher order elements such as Panel) are in fact "modal" elements that are precluding usage of other content on the page. The linked issue is also saying this, as well, and I agree with the reasoning.

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
JasonGore: Makes sense, but I'd rather limit this PR to dealing with `aria-modal` as I believe it's separate from the application of role here, and would have to consider all components using and used by Modal. Do you know of any issues that capture this? We could make one, if not.

packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
jspurlin: Hmm, I don't disagree with the general reasoning there, but just because TAB is bound to the FTZ, it doesn't mean that it is modal. The ribbon in office, for example, loops focus, but it isn't modal and the user can click/touch or use CTRL + F6 to navigate out of that content. Modal usually mean that the rest of the application can't be interacted with, with isn't always the case with FTZ

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
jspurlin: dialog still carries aria-modal inherently (with the value false) so this change is just potentially making the markup of the role worse

packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
JasonGore: Good points as well, I had thought of tying the `aria-modal` attribute to some FTZ props as well, but I'm not sure if that would be reliable.

What should we expect scan mode to do in the ribbon case? Should a user be able to navigate outside of the ribbon in scan mode?

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
JasonGore: Popup contains a FocusTrapZone, which would have the `aria-modal` property. I'm also concerned with having `aria-modal` on multiple layers unnecessarily, and `FocusTrapZone` can't know it's being used as a dialog.

What if I just close this PR and resolve the issue by making it the responsibility of users to pass `aria-modal` to FocusTrapZone?

packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
jspurlin: I would expect it to go outside of the ribbon (you could imagine a user who uses scan mode to navigate around the app and doesn't use CTRL +f6 (in the office online case) to navigate around the app, then they are stuck in the ribbon if an AT doesn't move outside of the modal area with virtual navigation turned on... (which would be really bad)

JasonGore: Related to my other comment (and yours), I'm thinking FocusTrapZone can't possibly know when to apply `aria-modal` accurately. I think I will close this PR and resolve the issue as being the responsibility of the FocusTrapZone consumer.

JasonGore:
Per our discussion on this PR, I'm going to close it without merging. I've verified users and consumers of `FocusTrapZone` can pass `aria-modal` as needed.

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
jspurlin: I think it should be the responsibility of the user to add aria-modal, but case like Dialog and Panel feel like they should have it by default

...and until modal stops having the ability to be a leaf component I guess it should have it it as well... but we're back to square one... If the modal couldn't be a leaf component I'm pretty sure it would simplify things a lot. Alternatively, if we could detect that the modal is being used as a leaf component they we could make this work

We need to make we aren't introducing accessibility bugs for users that are relying on the components to be marked up already (as a side effect of us trying to clean things up for accessibility... This might all have to be done together so no regressions are introduced)

JasonGore: Yeah, my intention was that they would still have it, since these components are using `FocusTrapZone`, and thus still would have `aria-modal`. In any case, there is no effect on navigation currently, but I do not want to misapply it and have closed this PR for now.

PR #10561, opened 2019-09-20
postprocess commonjs and amd outputs to make them easier to consume

### LINKED ISSUE #10461: AMD bundling: strip out process.env.NODE_ENV which isn't automatically defined by bundling
https://github.com/OfficeDev/office-ui-fabric-react/commit/a9245feee547762f0c8c9935cbf39d737b86c519 clearly introduces a breaking change without bumping the major version.

If fabric folks intend to use the minor versions as breaking, then please add guidelines that consumers should be fixing their fabric versions precisely up to the patch version.

kenotron:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10461

#### Description of changes

This introduces some postprocessing code for the AMD and COMMONJS module types from Fabric:

1. AMD users do not have a `process.env` defined - this is a node concept and is only declared by webpack
2. COMMONJS users who use jest to test their code have had to transform OUFR code because even if they required the 'lib-commonjs' version of our code, we still internally reference /lib everywhere (commonjs requiring esm) which is incompatible. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10561)


dzearing:
Thanks @kenotron !!!

kenotron:
You’re welcome! I’m not quite ready to merge just yet... need to think through implications and need someone who use AMD to help verify this...

scripts/tasks/postprocess-commonjs.js
```diff
@@ -0,0 +1,29 @@
...
+  const getAllPackageInfo = require('../monorepo/getAllPackageInfo');
+  const allPackages = Object.keys(getAllPackageInfo());
+
+  mod('lib-commonjs/**/*.js').asTypescript((node, modder) => {
+    if (ts.isCallExpression(node)) {
+      if (node.expression.getText() === 'require' && node.arguments.length === 1) {
+        const arg = node.arguments[0];
+        if (ts.isStringLiteral(arg)) {
+          if (belongsToFabric(arg.text) && arg.text.includes('lib/')) {
+            modder.replace(arg, `"${arg.text.replace('lib', 'lib-commonjs')}"`);
```
dzearing: This should really include the trailing slash so there are no accidents :)

scripts/tasks/postprocess-commonjs.js
```diff
@@ -0,0 +1,29 @@
...
+          }
+        }
+      }
+    }
+  });
+
+  function belongsToFabric(requireArg) {
```
dzearing: I don't understand what this is doing or how it relates to fabric. Prefer no "fabricisms" in this script if possible :)

dzearing: I think this is filtering out imports that are in the local repo

kenotron: Well, this should only convert lib-commonjs for fabric related packages so inherently fabric related. This function picks up all the local packages as something we would postprocess for a file.

kenotron:
I disabled the commonjs postprocess for now just so we can progress with the fix for the AMD post processing. That commonjs postprocess would be turned on in another isolated change.

kenotron:
Got an offline approval from @dzearing, so I'm merging this one.

Issue #10485, opened 2019-09-18
Callout does not return focus to original element when using 'hidden' prop


micahgodbolt:
The popup inside callout depends on "componentWillUnmount" to return focus to the originalFocusedElement

https://github.com/OfficeDev/office-ui-fabric-react/blob/3432f59416afac5a6f8535bb35bdba8bf438439f/packages/office-ui-fabric-react/src/components/Popup/Popup.tsx#L50-L65

Since the callout `hidden` prop is designed to not unmount the control, focus is never returned.

I'd prefer to keep the focus returning logic in the Popup, but how do we access it? Do we add `hidden` to the popup as well? Do we expose 'returnFocus' as a public prop? Other ideas?

from #8616

kenotron:
From a discussion with @micahgodbolt, breaking this down (a little hacky though):

1. expose "setOriginalFocusElement()" and "focusOriginalElement()" (or something similar)
2. in CalloutContent.base.tsx's componentDidUpdate() - if hidden prop changes, then call these functions in (1)

kenotron:
@joschect - can you comment here if you think this is an okay approach?

joschect:
ITT this kind of focus shouldn't be a thing that only the popup controls. Ideally there would be another focus function/object separate from the popup that would handle focus. In my opinion at least. So it could look something like:

``` 
export function getFocusReturner(hostElement) {
  const originalFocusedElement = getDocument(hostElement)!.activeElement as HTMLElement
  return () => {
    if (doesElementContainFocus(hostElement) && originalFocusedElement) {
        this._originalFocusedElement.focus();
      }
   }
} 
````

micahgodbolt:
Seeing as this is going to require some refactoring in the positioning/focus logic, I'm going to leave this one squarely in @joschect's capable hands.

espeo196:
Came across the same issue in something I'm working on. We are marking the callout as hidden and finding that focus seems to go back to the beginning of the page rather than the element we clicked to open the callout. Was wondering if there was an update or workaround discussed. I found it curious that FocusTrapCallout seemed to work fine as far as focus, but we don't desire forcing the focus to stay within the callout until manually dismissed.

PR #10581, opened 2019-09-24
Callout: Remove the use of 'hidden' from our example

micahgodbolt:
Due to #10485 i'm removing the 'hidden' example from our docs until we can fix the accessibility issue of the control. 

To use 'hidden' you will need to maintain your own focus on hiding the callout. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10581)


kenotron:
Great!

PR #10601, opened 2019-09-24
[Date-time] WeeklyDayPicker showFullMonth prop to expand to month picker inline

lorejoh12:
#### Pull request checklist

#### Description of changes

Adding property for @date-time WeeklyDayPicker component to allow it to expand to a full month picker, and adding an example of how it might be used.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10601)


packages/date-time/src/components/WeeklyDayPicker/WeeklyDayPicker.base.tsx
```diff
@@ -242,12 +258,20 @@ export class WeeklyDayPickerBase extends BaseComponent<IWeeklyDayPickerProps, IW
     );
   };
 
-  private _onSelectPrevWeek = () => {
-    this._navigateDate(addDays(this.state.navigatedDate, -7));
+  private _onSelectPrev = () => {
```
andrewhwng: nit: could we maintain the specificity of the function name here, something like `_onSelectPrevDateRange`? same applies to `_onSelectNext'

packages/date-time/src/components/WeeklyDayPicker/examples/WeeklyDayPicker.Inline.Expandable.Example.tsx
```diff
@@ -0,0 +1,125 @@
...
+  selectedDate?: Date;
+  expanded?: boolean;
+}
+
+export interface IWeeklyDayPickerInlineExpandableExampleProps extends Omit<IWeeklyDayPickerProps, 'strings'> {
+  showExpandButton?: boolean;
+  showNavigateButtons?: boolean;
+}
+
+export class WeeklyDayPickerInlineExpandableExample extends React.Component<
```
aneeshack4: Thanks for adding an example! 

Has there been more design consideration given to this? I see that there's just an expand/collapse button - can we consider a chevron button to bring up & down the full month view?
image

packages/date-time/src/components/WeeklyDayPicker/examples/WeeklyDayPicker.Inline.Expandable.Example.tsx
```diff
@@ -0,0 +1,125 @@
...
+              {' '}
+              {this.props.restrictedDates.length > 0
+                ? this.props.restrictedDates.map((d: Date) => d.toLocaleDateString()).join(', ')
+                : 'Not set'}
+            </span>
+          </div>
+        )}
+        {this.props.showExpandButton && (
+          <div>
+            <DefaultButton className={styles.button} onClick={this._expand} text="Expand/collapse" />
```
aneeshack4: This button needs aria-label. And something else to convey the state of the picker - whether it's expanded/collapsed. I ran Narrator over this and nothing's announced when this action is performed.

packages/date-time/src/components/WeeklyDayPicker/WeeklyDayPicker.base.tsx
```diff
@@ -55,6 +55,12 @@ export interface IWeeklyDayPickerState {
 
   /** The currently selected date in the calendar */
   selectedDate: Date;
+
+  /** Tracking whether we just toggled showFullMonth */
+  previousShowFullMonth: boolean;
```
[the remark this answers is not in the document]
lorejoh12: so we need to keep track of both the current showFullMonth as well as the previousShowFullMonth- this is specifically used for animations. In order to know whether we should be animating horizontally or vertically, we need to know if the property just changed- previously that was done using componentWillReceiveProps, where we could just compare the old props to new, but now we need to track it ourselves in the state.

packages/date-time/src/components/WeeklyDayPicker/examples/WeeklyDayPicker.Inline.Expandable.Example.tsx
lorejoh12: We plan on using this as a small control in a larger OWA component- there will be a button in a menu there used to expand/collapse. I see this more as a smaller piece, much like CalendarDayGrid is a smaller component. If someone wanted to use a more complete wrapper component around this that just had the header with a chevron arrow, that'd be fine, but I think we'd create a separate component for that.

packages/date-time/src/components/WeeklyDayPicker/examples/WeeklyDayPicker.Inline.Expandable.Example.tsx
lorejoh12: Why would we need an aria-label on a button that has text in it? Shouldn't narrator read out the text if there isn't an aria-label specified? It's true though that narrator doesn't announce anything when performing the action, I'm not really sure what the right aria behavior is there.

https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions says that we can put the aria-expanded attribute on the button, and doing so seems to announce expand/collapse correctly.

PR #10604, opened 2019-09-24
Nav 10594

### LINKED ISSUE #10594: [Accessibility][Programmatic access]: Name of the 'Basic components expand button' is not descriptive to the user.
Tracking all of the issues brought up with the 1/25 accessibility pass on Nav control 
Details on individual bugs can be found on [SharePoint](https://microsoft-my.sharepoint.com/:x:/r/personal/v-amaver_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B8b59713f-87bb-4a5e-ad92-2b7b3a0ef589%7D&action=default&uid=%7B8B59713F-87BB-4A5E-AD92-2B7B3A0EF589%7D&ListItemId=1721&ListId=%7B395E1ECB-5002-440E-81C5-6A090E98077C%7D&odsp=1&env=prodbubble) 

This issue is tracking a single entry from #7803

jdhuntington:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #10594

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10604)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
@@ -143,7 +143,15 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
       groups
     });
 
-    const finalExpandBtnAriaLabel = expandButtonAriaLabel ? `${link.name} ${expandButtonAriaLabel}` : link.name;
+    let finalExpandBtnAriaLabel = '';
+    if (link.links && link.links.length > 0) {
+      if (link.collapseAriaLabel || link.expandAriaLabel) {
+        finalExpandBtnAriaLabel = link.isExpanded ? link.collapseAriaLabel! : link.expandAriaLabel!;
+      } else {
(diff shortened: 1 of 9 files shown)


packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
```diff
@@ -218,11 +218,13 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
       groups
     });
 
+    const finalExpandBtnAriaLabel = expandButtonAriaLabel ? `${group.name} ${expandButtonAriaLabel}` : group.name;
```
dzearing: This could be a localization concern. Is the subject before the verb in all languages?

jdhuntington: Good call. This is the way other groups were being handled, but I think you're right that this is not ideal. Introduced some new attributes on links/groups to handle this correctly and deprecated the old method.

PR #10605, opened 2019-09-24
Dropdown: Error messages in Dropdown are now linked to the Dropdown (from a programmatic accessibility POV).

### LINKED ISSUE #10600: Dropdown:  'Error message' is not associated with the 'Error message example drop down'
Standalone tracking issue for item 3 in #6326.

cliffkoh:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10600

#### Description of changes

This links the Dropdown to the error message that might potentially be displayed.

I've done considerable reading of specs, looked at various examples on the internet, and tested on Narrator, ChromeVox and NVDA and believe this is the best, most spec-compliant change that is the most semantically correct.

One thing to note is that the various screen reading software differ widely in how they handle aria-labelledby/aria-describedby and when they intersect with roles or if they're both present. David MacDonald (a veteran WCAG member) has an article discussing some of this here: https://www.davidmacd.com/blog/does-aria-label-override-static-text.html

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10605)

### DIFF
--- a/packages/utilities/src/aria.ts
+++ b/packages/utilities/src/aria.ts
@@ -4,9 +4,9 @@
  *
  * @param ariaAttributes - ARIA attributes to merge
  */
-export function mergeAriaAttributeValues(...ariaAttributes: (string | undefined)[]): string | undefined {
+export function mergeAriaAttributeValues(...ariaAttributes: (string | undefined | false)[]): string | undefined {
   const mergedAttribute = ariaAttributes
-    .filter((arg: string | undefined) => arg !== undefined && arg !== null)
+    .filter((arg: string | undefined | false) => arg !== undefined && arg !== null && arg !== false)
     .join(' ')
     .trim();
   return mergedAttribute === '' ? undefined : mergedAttribute;
--- a/packages/utilities/src/aria.ts
+++ b/packages/utilities/src/aria.ts
@@ -6,7 +6,7 @@
  */
 export function mergeAriaAttributeValues(...ariaAttributes: (string | undefined | false)[]): string | undefined {
   const mergedAttribute = ariaAttributes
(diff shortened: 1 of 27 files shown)


packages/utilities/src/aria.ts
```diff
@@ -4,9 +4,9 @@
  *
  * @param ariaAttributes - ARIA attributes to merge
  */
-export function mergeAriaAttributeValues(...ariaAttributes: (string | undefined)[]): string | undefined {
+export function mergeAriaAttributeValues(...ariaAttributes: (string | undefined | false)[]): string | undefined {
   const mergedAttribute = ariaAttributes
-    .filter((arg: string | undefined) => arg !== undefined && arg !== null)
+    .filter((arg: string | undefined | false) => arg !== undefined && arg !== null && arg !== false)
```
jdhuntington: can this be simplified to the following?

    .filter((arg: string | undefined | false) => arg)

I think the only change is that it ends up filtering out `''`, which seems fine.

dzearing: if i'm reading this right, it's filtering out undefined, null, and false. Not 0.

cliffkoh: strictly speaking, I'm not sure why we check for null when the types don't allow null. Likewise, 0 is technically an invalid input and even if someone does pass it in, we probably don't want to preserve "0". I will make the change JD suggested (what I had done here was just add on to the existing pattern but I think JD's suggestion is a net improvement)

PR #10610, opened 2019-09-25
[Pickers] Edge IME Compatibility & IE demo page bug fix

andrewhwng:
#### Pull request checklist

#### Description of changes

1. There was a bug where typing inside the picker with a language requiring a composition state for typing characters (ie. Japanese) would not behave as expected in Edge. Specifically...
    - Typing a character that matches a suggestion will fill the text input to match the first suggestion, but the highlight does not occur
2. Demo page selector did not work in IE in most recent build when I was testing. Turns out it was caused due to the RegExp constructor not being supported in IE _iff_ there is a regexp passed in. It only takes a string...

#### Focus areas to test

- Any demo page in IE
- Pickers while testing composition based languages. To test, I will push another commit to add a Japanese string to the example page and provide more details on how to test it in a comment :) I'll be sure to remove the commit before merging.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10610)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Autofill/Autofill.tsx
+++ b/packages/office-ui-fabric-react/src/components/Autofill/Autofill.tsx
@@ -217,8 +217,9 @@ export class Autofill extends BaseComponent<IAutofillProps, IAutofillState> impl
   private _onInputChanged = (ev: React.FormEvent<HTMLElement>) => {
     const value: string = this._getCurrentInputValue(ev);
 
-    // Right now typing does not have isComposing, once that has been fixed any should be removed.
-    this._tryEnableAutofill(value, this._value, (ev.nativeEvent as any).isComposing);
+    if (!this._isComposing) {
+      this._tryEnableAutofill(value, this._value, (ev.nativeEvent as any).isComposing);
+    }
 
     // If it is not IE11 and currently composing, update the value
     if (!(isIE11() && this._isComposing)) {
--- a/change/office-ui-fabric-react-2019-09-24-18-35-44-u-anhw-picker-asian-langs.json
+++ b/change/office-ui-fabric-react-2019-09-24-18-35-44-u-anhw-picker-asian-langs.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Prevent suggestion autocomplete without selection",
+  "packageName": "office-ui-fabric-react",
+  "email": "anhw@microsoft.com",
+  "commit": "87c6dca185d61a8ed15782090e5b2f7a30fe7fc4",
+  "date": "2019-09-25T01:35:44.321Z"
+}
--- a/packages/office-ui-fabric-react/src/components/Autofill/Autofill.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/Autofill/Autofill.test.tsx
@@ -32,6 +32,20 @@ describe('Autofill', () => {
     expect(autofill.inputElement!.value).toBe('hello');
   });
 
+  it('correctly autofills with composable languages', () => {
+    let updatedText: string | undefined;
+    const onInputValueChange = (text: string | undefined): void => {
+      updatedText = text;
+    };
+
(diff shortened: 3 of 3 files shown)


packages/tsx-editor/src/transpiler/exampleParser.ts
```diff
@@ -93,7 +93,7 @@ export function tryParseExample(example: string, supportedPackages: IBasicPackag
 
 /** @internal */
 export function _tryParseExample(example: string, supportedPackages: string[]): IExampleInfo | string {
-  const possibleComponents = example.match(new RegExp(COMPONENT_NAME_REGEX, 'gm'));
+  const possibleComponents = example.match(new RegExp(COMPONENT_NAME_REGEX.source, 'gm'));
```
ecraig12345: Thanks for noticing this! Sorry about the extra trouble but would you mind putting this in a separate PR (or I can)? I'm not familiar enough with Autofill to be comfortable approving that part of the change, but this part is fine and should really go in ASAP.

andrewhwng: Sure! just created [PR 10611](https://github.com/OfficeDev/office-ui-fabric-react/pull/10611), I'll get rid of these changes here

andrewhwng:
**Testing**
In the most recent commit (in which I will delete before merging), I pushed an option in the demo page to include 'あいうえお' as an auto-fill option for the picker.

To get the typing to work, install Japanese on your OS, switch your typing language, and the first character can be matched by pressing the 'a' key (you'll need to then press enter to finish the composition). Auto-fill should work as expected.

Exporting to codepen will load the broken version and you can run the same test there to see the bug :)

ecraig12345:
> In the most recent commit (in which I will delete before merging)...

Added do not merge label to make that clear. :) Feel free to remove it when you're ready.

packages/office-ui-fabric-react/src/components/pickers/examples/TagPicker.Basic.Example.tsx
```diff
@@ -23,7 +23,8 @@ const _testTags: ITag[] = [
   'rose',
   'violet',
   'white',
-  'yellow'
+  'yellow',
+  'あいうえお'
```
joschect: DId you mean to add this here?

packages/office-ui-fabric-react/src/components/Autofill/Autofill.tsx
```diff
@@ -217,8 +217,9 @@ export class Autofill extends BaseComponent<IAutofillProps, IAutofillState> impl
   private _onInputChanged = (ev: React.FormEvent<HTMLElement>) => {
     const value: string = this._getCurrentInputValue(ev);
 
-    // Right now typing does not have isComposing, once that has been fixed any should be removed.
-    this._tryEnableAutofill(value, this._value, (ev.nativeEvent as any).isComposing);
+    if (!this._isComposing) {
+      this._tryEnableAutofill(value, this._value, (ev.nativeEvent as any).isComposing);
```
joschect: Can we add a test? To catch some of these IME bugs?

packages/office-ui-fabric-react/src/components/pickers/examples/TagPicker.Basic.Example.tsx
joschect: Should have read your comment. Ignore this.

andrewhwng: This was to have a deploy that's testable, I will delete it before merging! Hence the 'Do not merge' label on the PR now

packages/office-ui-fabric-react/src/components/Autofill/Autofill.tsx
andrewhwng: I've just added a simple autofill test for composable languages :)

andrewhwng:
@joschect tagging one last time for sanity checks :)

PR #10613, opened 2019-09-25
Allow custom Overflow icon in Breadcrumb component

### LINKED ISSUE #10612: Unable to set custom Overflow Icon in Breadcrumb component
### Environment Information

- **Package version(s)**: 6.204.2

#### Actual behavior:

If the default icon library is not loaded, the result is button that contains no icon:

Screen Shot 2019-09-24 at 7 09 52 PM

We generally avoid this issue by using an icon custom render function, but in the case of the Overflow Button there is none.

#### Expected behavior:

Being able to set a render function for the icon.

Requested priority: High

Products/sites affected: Teams for Edu

matidomingues:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10612

#### Description of changes

The current implementation does not allow to include custom items on the Breadcrumb component, this limits applications where the Icon library is not loaded or the ability to render a custom Icon.

In this PR I just added an IRenderFunction to the Breadcrumb container where you can add a custom function to render the Icon.

Currently, the current icon defined in the IconButton component is being defined by the `{ iconName: 'More' }` prop so we need to conditionally remove this prop and provide a custom  `onRenderMenuIcon`

Another approach would be to still use the `onRenderOverflowIcon` prop and in case it's not defined, define a custom Icon render function for the default (for example: `return <Icon iconName={'More'} />; `) the problem with this is that IconButton currently adds a few extra CSS properties (which seems to not do anything), but I did not want to get rid of them.

Also, this change should also be applied to the 7.x version, but we are currently using 6.x.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10613)

### DIFF
--- a/change/office-ui-fabric-react-2019-09-24-19-25-38-madomi-overflow-icon.json
+++ b/change/office-ui-fabric-react-2019-09-24-19-25-38-madomi-overflow-icon.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Allow setting a custom overflow icon on the Breadcrumb component",
+  "packageName": "office-ui-fabric-react",
+  "email": "madomi@microsoft.com",
+  "commit": "f5164fd724812ef0b1877445b0ac825debf96504",
+  "date": "2019-09-25T02:25:38.368Z"
+}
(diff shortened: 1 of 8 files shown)


dzearing:
Please do cherrypick into master and we'll get that merged. Thank you!

change/office-ui-fabric-react-2019-09-24-19-25-38-madomi-overflow-icon.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
dzearing: oops missed this, but we've minored since then.

PR #10625, opened 2019-09-25
Rating: Align read-only behavior with ARIA standards

### LINKED ISSUE #10126: Rating Star component losing tabindex once tabbed out of it in readonly mode.
### Environment Information

- **Package version(s)**: (latest @7)
- **Browser and OS versions**: (Chrome 76.0.3809.100, Win10)

### Describe the issue:

The Rating Star component in readonly mode, sets the tabindex = 0 for the rendered component. But when you are tabbing through it, after the first time it receives focus, the tabindex attribute is removed.

Can get a repro here:
https://fabricweb.z5.web.core.windows.net/oufr/7.22.1/#/examples/rating

and also here:
https://developer.microsoft.com/en-us/fabric/#/controls/web/rating

Steps to repro:
Using keyboard and tabs get the focus to the "Usage" section examples, tab through the elements, observe that focus actually goes to the Stars under "Half star in readOnly mode:" the first time, but when you try to tab to it again, (shift + Tab / Tab), the focus doesn't go into the element. On inspecting the element, it is seen that the tabindex property gets removed for the component's FocusZone class element.

Check in gif:
ratingstarfocus

#### Expected behavior:

The focus should actually be going to the Rating Star component in the first place, as it is a readonly component.

https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html

In short "All interactive elements must be focusable and non-interactive elements must not be focusable"

1) In this case, the readonly rating star component is not interactive and should not be focusable to begin with.
2) The tabindex is removed once we focus out of the component.

Upon investigatign further, found that the code here: 
https://github.com/OfficeDev/office-ui-fabric-react/pull/7818/files#r252543706
is causing the tabindex to be removed.

There's two potential issues here:
1) The Rating Star component shouldn't be setting the tabindex as 0 for non interactive elements. (https://github.com/OfficeDev/office-ui-fabric-react/blob/a5bbc76367eb01e993f20796955569f422a538b2/packages/office-ui-fabric-react/src/components/Rating/Rating.base.tsx#L140)
2) The _setParkedFocus could be incorrectly removing the tabindex attribute for elements.
(https://github.com/OfficeDev/office-ui-fabric-react/blob/0dde4ac50fcaee5bdfaa9a44c827e8396a16d02d/packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx#L398)

JasonGore:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10126

#### Description of changes

#10126 highlights concerns with `Rating`'s accessibility, but proposed making read-only `Ratings` unfocusable. However, [ARIA's read-only attribute](https://www.w3.org/TR/wai-aria-1.1/#aria-readonly) says that read-only elements that are otherwise operable or indirectly affected by users should be focusable. This PR aligns `Rating`'s `readOnly` prop with the ARIA standards by making read-only `Ratings` focusable. Thge PR also now allows `aria-label` contents to be dictated when users focus on read-only `Ratings`.

In addition, this PR fixes the tabindex focus behavior by applying focusability to the child `FocusZone` in readonly mode.

This PR has a side effect that `Ratings` that were previously unfocusable (such as the "Button Controlled Rating" example) are now focusable. This is consistent with the ARIA standard that read-only components that are otherwise operable should be focusable.

I've also fixed a couple of instances where ARIA attributes were being set to `''` rather than `undefined`. (ARIA attributes should be `undefined` when they have no valid value.)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10625)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Rating/Rating.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Rating/Rating.base.tsx
@@ -11,7 +11,7 @@ import {
 } from '../../Utilities';
 import { IProcessedStyleSet } from '../../Styling';
 import { Icon } from '../../Icon';
-import { FocusZone, FocusZoneDirection } from '../../FocusZone';
+import { FocusZone, FocusZoneDirection, IFocusZoneProps } from '../../FocusZone';
 import { IRatingProps, RatingSize, IRatingStyleProps, IRatingStyles } from './Rating.types';
 
 const getClassNames = classNamesFunction<IRatingStyleProps, IRatingStyles>();
@@ -128,7 +128,20 @@ export class RatingBase extends React.Component<IRatingProps, IRatingState> {
       }
     }
 
-    const ariaLabel = getAriaLabel ? getAriaLabel(rating ? rating : 0, max as number) : '';
+    const ariaLabel = getAriaLabel ? getAriaLabel(rating ? rating : 0, max as number) : undefined;
+
+    // When in read-only mode, we allow focus (per ARIA standards) and set up ARIA attributes to indicate element is read-only.
+    // https://www.w3.org/TR/wai-aria-1.1/#aria-readonly
+    const readOnlyProps: IFocusZoneProps | undefined = readOnly
+      ? ({
+          allowFocusRoot: true,
+          disabled: true,
+          'aria-label': ariaLabel,
+          'aria-readonly': true,
+          'data-is-focusable': true,
+          tabIndex: 0
+        } as IFocusZoneProps)
+      : undefined;
 
     return (
       <div
@@ -142,14 +155,12 @@ export class RatingBase extends React.Component<IRatingProps, IRatingState> {
       >
         <FocusZone
           direction={FocusZoneDirection.horizontal}
-          tabIndex={readOnly ? 0 : -1}
           className={css(this._classNames.ratingFocusZone, {
             [this._classNames.rootIsLarge]: size === RatingSize.Large,
             [this._classNames.rootIsSmall]: size !== RatingSize.Large
           })}
-          data-is-focusable={readOnly ? true : false}
           defaultActiveElement={rating ? starIds[Math.ceil(rating) - 1] && '#' + starIds[Math.ceil(rating) - 1] : undefined}
-          aria-label={readOnly ? ariaLabel : ''}
+          {...readOnlyProps}
         >
           {stars}
         </FocusZone>
--- a/change/office-ui-fabric-react-2019-09-25-10-36-25-jg-10126-rating-accessibility.json
+++ b/change/office-ui-fabric-react-2019-09-25-10-36-25-jg-10126-rating-accessibility.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Rating: Align read-only functionality with ARIA standards.",
+  "packageName": "office-ui-fabric-react",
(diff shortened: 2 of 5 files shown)


jspurlin:
Should this have a `allowDisabledFocus` prop like buttons? I know this was added due to a concern David had that some creators may want the disabled component to not be focusable (and the default setting would align with the current/legacy behavior)

JasonGore:
This PR doesn't change anything regarding `disabled` behavior. Disabled Ratings were not focusable and remain not focusable. This is also in alignment with [the ARIA standard regarding `aria-disabled`](https://www.w3.org/TR/wai-aria-1.1/#aria-disabled) in that disabled elements should not be focusable.

JasonGore:
[Here](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/10625/merge/fabric-website-resources/dist/index.html#/examples/rating) is a link to this PR's version of Rating.

jspurlin:
> This PR doesn't change anything regarding `disabled` behavior. Disabled Ratings were not focusable and remain not focusable. This is also in alignment with [the ARIA standard regarding `aria-disabled`](https://www.w3.org/TR/wai-aria-1.1/#aria-disabled) in that disabled elements should not be focusable.

// side tangent, from what I see on that site, it states:
> Disabled elements might not receive focus from the tab order. For some disabled elements, applications might choose not to support navigation to descendants.

That a different conversations for a different time...

Anyways, thanks, I was looking at this incorrectly (thinking it was about `disabled` instead of `read-only`

packages/office-ui-fabric-react/src/components/Rating/Rating.base.tsx
```diff
@@ -142,14 +155,12 @@ export class RatingBase extends React.Component<IRatingProps, IRatingState> {
       >
```
jspurlin: Is this component a slider at the end of the day? If so, it should be marked up as such: [https://www.w3.org/TR/wai-aria-1.1/#slider](https://www.w3.org/TR/wai-aria-1.1/#slider)

JasonGore: I don't think it's close enough to ARIA's definition of a slider:

> A slider represents the current value and range of possible values via the size of the slider and position of the thumb.

The [few](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Role_Img#Examples) [examples](https://www.w3.org/WAI/tutorials/forms/custom-controls/) I find don't use `slider` anywhere. If anything, it seems like we should be using `role` of `img` instead of `presentation`.

jspurlin: role img is almost always indicative of a static image. I found the right role, radioGroup: [https://www.w3.org/TR/wai-aria-1.1/#radiogroup](https://www.w3.org/TR/wai-aria-1.1/#radiogroup) and each of the items should have role radio: [https://www.w3.org/TR/wai-aria-1.1/#radio](https://www.w3.org/TR/wai-aria-1.1/#radio)

---
In reply to: [328340606](https://github.com/OfficeDev/office-ui-fabric-react/pull/10625#discussion_r328340606) [](ancestors = 328340606)

PR #10626, opened 2019-09-25
DetailsList: Narrator should announce SelectAll Checkbox as a toggle selection for all item

### LINKED ISSUE #10519: Screen reader is not reading the action performed when we check/uncheck the select all checkbox
### Environment Information

- **Package version(s)**: (6.182.1)
- **Browser and OS versions**: (Edge/Chrome and Windows 10)

### Describe the issue:

Screen reader(Narrator/NVDA) is not reading the action performed when we check/uncheck the select all checkbox

https://codepen.io/sandeepnmenon/pen/NWKOgYo?&editable=true

#### Actual behavior:

Narrator/NVDA is not reading the action performed when we check/uncheck the "select all" checkbox

#### Expected behavior:

Narrator/NVDA should read the action performed when we check/uncheck the "select all" checkbox

aneeshack4:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10519 

#### Description of changes

Right now, screen reader's interaction with the DetailsList SelectAll checkbox is ambiguous because it simply announces it as "Toggle selection" and when toggled, the states are announced as "checked" and "unchecked" so the user's not aware that all of the items are being checked and unchecked. This is because the `aria-label` on the DetailsRowCheck was set to the ariaLabelForSelectionColumn, not ariaLabelForSelectAllCheckbox. 
Before:
image
After:
image

#### Focus areas to test
Narrator should read out "Toggle selection for all items" when focus is on the select all checkbox.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10626)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.base.tsx
@@ -204,7 +204,7 @@ export class DetailsHeaderBase extends React.Component<IDetailsHeaderBaseProps, 
                     children: (
                       <DetailsRowCheck
                         id={`${this._id}-check`}
-                        aria-label={ariaLabelForSelectionColumn}
+                        aria-label={ariaLabelForSelectAllCheckbox}
                         aria-describedby={
                           !isCheckboxHidden
                             ? ariaLabelForSelectAllCheckbox && !this.props.onRenderColumnHeaderTooltip
--- a/change/office-ui-fabric-react-2019-09-25-13-40-00-detailslist.json
+++ b/change/office-ui-fabric-react-2019-09-25-13-40-00-detailslist.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "DetailsList: Narrator should announce SelectAll Checkbox should as a toggle selection for all items",
+  "packageName": "office-ui-fabric-react",
+  "email": "email not defined",
+  "commit": "24028884d7f00767de402aa6cb9e2dc9874c8f21",
+  "date": "2019-09-25T20:40:00.190Z"
+}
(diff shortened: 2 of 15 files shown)


packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.base.tsx
```diff
@@ -204,7 +204,7 @@ export class DetailsHeaderBase extends React.Component<IDetailsHeaderBaseProps,
                     children: (
                       <DetailsRowCheck
                         id={`${this._id}-check`}
-                        aria-label={ariaLabelForSelectionColumn}
+                        aria-label={ariaLabelForSelectAllCheckbox}
```
natalieethell: Since this header is only a select all checkbox in multiple selection mode (and only a header for single selection mode), should we do this instead?

```tsx
aria-label={selectionMode === SelectionMode.multiple ? ariaLabelForSelectAllCheckbox : ariaLabelForSelectionColumn
```

KevinTCoughlin: Good catch! If the checkbox is rendered with the corresponding `aria-label` for both scenarios then yes we should make it conditional.

@aneeshack4 can you verify?

aneeshack4: that works yes verified:
image

aneeshack4: Oops I accidentally added the automerge label before pushing the latest commit - I'll make another PR to fix this.

aneeshack4: PR #10641

PR #10627, opened 2019-09-25
Textfield: aria-labelledby should be set on mulitine textfields

### LINKED ISSUE #10551: Narrator is not reading TextField in case of multiline.
### Environment Information
It was repro at Edge
### Describe the issue:
In case of TextField is multiline, Narrator does not read textfield label, because Fabric control does not render "aria-labelledby" attribute. 
-->

#### Actual behavior:

Fabric control does not render "aria-labelledby" attribute. 

#### Expected behavior:

Fabric control (TextField -Multiline) must render "aria-labelledby" attribute. so that narrator can read as did in single line TextField.

aneeshack4:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10551 

#### Description of changes
Right now, aria-labelledby isn't set on multiline textfields. I believe this wasn't originally set because the [ARIA spec for `aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) says to set it on `input` (single line textfield) but not `<textarea>` (multiline textfield). Makes sense to set it though and found some [other](https://developer.paciellogroup.com/blog/2017/07/short-note-on-aria-label-aria-labelledby-and-aria-describedby/) [documentation](https://thepaciellogroup.github.io/AT-browser-tests/test-files/textarea.html) agreeing with that. 
image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10627)

### DIFF
--- a/change/office-ui-fabric-react-2019-09-25-14-32-36-textfield.json
+++ b/change/office-ui-fabric-react-2019-09-25-14-32-36-textfield.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Textfield: aria-labelledby should be set on multiline textfields too",
+  "packageName": "office-ui-fabric-react",
+  "email": "email not defined",
+  "commit": "44513977f9292c8eb860c6ae35e9420a005cc67b",
+  "date": "2019-09-25T21:32:36.811Z"
+}
--- a/change/office-ui-fabric-react-2019-09-25-14-32-36-textfield.json
+++ b/change/office-ui-fabric-react-2019-09-25-14-32-36-textfield.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Textfield: aria-labelledby should be set on multiline textfields too",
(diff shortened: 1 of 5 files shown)


change/office-ui-fabric-react-2019-09-25-14-32-36-textfield.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Textfield: aria-labelledby should be set on multiline textfields too",
```
cliffkoh:
```suggestion
  "comment": "TextField: Multiline textfields are now correctly associated with their labels with `aria-labelledby`.",
```

PR #10628, opened 2019-09-25
Website accessibility docs

micahgodbolt:
Adding accessibility docs from the design team. Probably a bit of massaging to the text from the design and engineering team to clarify/improve consistency. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10628)


ecraig12345:
I think this is a good change in principle, but practically, 90% of the info presented is irrelevant to **consumers** of our controls (a lot of the things listed are handled by the controls internally). So putting it on the website will just confuse people and make them think they need to manually add handling for a bunch of things when in fact we've already done it for them.

Instead, the public content should focus on the parts that our controls don't handle--the only one that comes to mind offhand is ensuring everything has an appropriate label/labelledby (especially worth calling out in controls where the consumer needs to provide labels for sub-elements like an X button), but there are probably others too.

packages/office-ui-fabric-react/src/components/Button/docs/ButtonA11y.md
```diff
@@ -0,0 +1,12 @@
+Button is accessible if these considerations are followed:
+
+- Use a touch target area of `44px` by `44px`
```
dzearing: hmm. 44x44 is far larger than pretty much any button currently. A default button is 32px tall. This guidance is super confusing.

packages/office-ui-fabric-react/src/components/Button/docs/ButtonA11y.md
```diff
@@ -0,0 +1,12 @@
...
+- For multiple buttons provided on the page that perform the same function, one of the targets should be `44px` by `44px`.
+- Buttons should get activated using `Enter` key as well as `SPACE` key.
+- For custom buttons made using `<div>`, `<span>`, `<svg>`or `<img>`,
+  - `role="button"` must be applied.
+  - Tabindex property should be set to "0", so that it could get tab focus.
```
dzearing: This is wrong guidance.  We should not be asking users to worry about tabIndex, when the component already manages this for you.

We need to separate 2 concerns:

1. Things the Button will manage for you, so that you (the user) don't need to. 

You don't need to worry about tabbing or spacing. You don't need to worry about tab index. You don't need to worry about "custom buttons" unless you're dealing with an `as` prop, and even then maybe WE should be providing role.button and not even asking the user to deal with it.

2. Things that you (the user) need to consider and HOW. 

For example, perhaps tooltips are a good idea for icon buttons. How? What's the recommended accessible approach? Just adding `title` would do it, but you wouldn't get a consistent look/feel. Perhaps we should add a custom tooltip here. We need a code example.

dzearing: Just saw @ecraig12345 comment, which is effectively the same thing i just said :)

joschect:
@micahgodbolt per Elizabeth's comment it sounds like most of this doesn't apply to using our controls. If it's a good distillation of general aria-specs it might make sense to add this to the wiki somewhere so that in the future we have a summary to go to help us make our components accessible.

micahgodbolt:
closing PR down for now. Link is still visible to reviewers if it's closed. Don't need to clog up our PR queue.

PR #10640, opened 2019-09-26
Dropdown: There is now a visible focus border on Dropdowns in the error state.

### LINKED ISSUE #10636: [Accessibility][Keyboard navigation]: Keyboard Focus indicator is not visible on 'Error message example drop down'.
This was originally issue 4 in #6326.

cliffkoh:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10636 

#### Description of changes

@betrue-final-final is out but discussed with @phkuo and agreed with his POV that even in an error state, the Dropdown should go to normal focus since the point of an error is to draw initial attention to it. Once a Dropdown is in the error state, for it to lose the red sharing while it has focus is fine since the user would have already known there is an error (and there is also an error text beneath).

(Thank you @phkuo :))

dropdown focus


micahgodbolt:
So just for clarity:

Before: Error state border took precedence over focus

After: Focus border takes precedence over error state border

PR #10644, opened 2019-09-27
Checkbox: improved check state handling, tests, and examples

ecraig12345:
#### Pull request checklist

#### Description of changes

I was trying to write a test for another component which consumed a Checkbox, and found that I was having trouble getting it to behave how I expected in a test (even though it seems to work okay at runtime). I think this was because of some inconsistencies in how it determines/stores the checked state. After this update:

- `state.isChecked` and `state.isIndeterminate` are the one source of truth for the current checked/indeterminate state
- In `getDerivedStateFromProps`:
  - `nextProps.checked` should always be copied into `state.isChecked` if provided (regardless of current indeterminate state)
  - `nextProps.indeterminate` should always be copied into `state.isIndeterminate` if provided (it's possible a consumer would like to programmatically transition the checkbox from checked/unchecked to indeterminate, and we should respect that)
- Internally, always set the `<input>`'s `checked` prop rather than using `defaultChecked` sometimes (it simplifies things and makes testing easier)

In the process of figuring out why my test didn't work, I added more tests for Checkbox (and simplified some of the existing tests) to make sure it was *possible* to test against and that the issue wasn't just my test.

I also modified the way the examples are grouped together and moved the indeterminate cases to a separate example.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10644)


micahgodbolt:
Adding @maxwellred for review as he did the original indeterminate work.

micahgodbolt:
@msft-github-bot make sure @maxwellred signs off on this PR

cliffkoh:
Requesting review from @dzearing as well as he was heavily involved in reviewing and iterating on @maxwellred 's original PR.

cliffkoh:
@ecraig12345 I will review when I get back from OOF but please feel free to merge if @maxwellred and @dzearing signs off.

aneeshack4:
Pinging @maxwellred and @dzearing - gentle reminder to get back to this if y'all have any more feedback

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -179,24 +170,25 @@ export class CheckboxBase extends React.Component<ICheckboxProps, ICheckboxState
   };
 
   private _onChange = (ev: React.FormEvent<HTMLElement>): void => {
-    const { disabled, onChange } = this.props;
+    const { onChange } = this.props;
```
ecraig12345: This event shouldn't be getting called to begin with if the component is disabled (other form components assume this), so the function doesn't need to include a disabled check.

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -1,175 +1,233 @@
...
-    );
-  }
-
-  private _onControlledChange = (ev: React.FormEvent<HTMLElement>, checked: boolean): void => {
-    if (this.state.indeterminateState) {
-      this.setState({ indeterminateState: false });
-    } else {
-      this.setState({ checkedState: checked! });
-    }
+const IndeterminateControlledCheckbox: React.FunctionComponent = () => {
```
ecraig12345: Switch to simplified test component: functional and with only one checkbox

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -1,175 +1,233 @@
...
   });
-  it('renders Checkbox with Uncontrolled indeterminate correctly', () => {
-    const component = mount(<Checkbox label="Uncontrolled Indeterminate checkbox" defaultIndeterminate />);
-    expect(component.find(CheckboxBase).state('isIndeterminate')).toEqual(true);
+
+  it('defaults to unchecked non-indeterminate', () => {
+    component = mount(<Checkbox componentRef={checkboxRef} />);
+
+    const input = component.find('input');
+    expect(input.prop('checked')).toBe(false);
```
ecraig12345: - Add tests for `checked`, `defaultChecked`, and `onChange` in addition to indeterminate stuff
- Verify both the item props and the component state (to ensure they're in sync)

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -117,8 +109,7 @@ export class CheckboxBase extends React.Component<ICheckboxProps, ICheckboxState
               type="checkbox"
               {...inputProps}
               data-ktp-execute-target={keytipAttributes['data-ktp-execute-target']}
-              {...checked !== undefined && { checked }}
-              {...defaultChecked !== undefined && { defaultChecked }}
+              {...isChecked !== undefined && { checked: isChecked }}
```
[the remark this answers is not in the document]
maxwellred: @ecraig12345 Is it possible to change this to `checked={ isChecked }` ?

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -28,18 +28,15 @@ export class CheckboxBase extends React.Component<ICheckboxProps, ICheckboxState
...
     }
-    if (props.checked !== undefined) {
-      return {
-        isChecked: !!props.checked
-      };
+    if (nextProps.checked !== undefined) {
+      prevState.isChecked = !!nextProps.checked;
     }
-    return null;
+    return Object.keys(state).length ? state : null;
```
cliffkoh: >state [](start = 23, length = 5)

i might be missing something but where in between line 32 and 39 is state actually modified? wouldn't this always return null?

cliffkoh: Also, if there is indeed a bug here, should the tests have caught something?

---
In reply to: [335074857](https://github.com/OfficeDev/office-ui-fabric-react/pull/10644#discussion_r335074857) [](ancestors = 335074857)

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -28,18 +28,15 @@ export class CheckboxBase extends React.Component<ICheckboxProps, ICheckboxState
...
 
-  public static getDerivedStateFromProps(props: ICheckboxProps, state: ICheckboxState): ICheckboxState | null {
-    if (!props.defaultIndeterminate && state.isIndeterminate) {
-      return {
-        isIndeterminate: !!props.indeterminate
-      };
+  public static getDerivedStateFromProps(nextProps: ICheckboxProps, prevState: ICheckboxState): ICheckboxState | null {
+    const state: Partial<ICheckboxState> = {};
+    if (nextProps.indeterminate !== undefined) {
+      prevState.isIndeterminate = !!nextProps.indeterminate;
```
cliffkoh: >prevState [](start = 6, length = 9)

is this meant to be state?

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -149,11 +140,11 @@ export class CheckboxBase extends React.Component<ICheckboxProps, ICheckboxState
   }
 
   public get indeterminate(): boolean {
-    return this.state.isIndeterminate!;
+    return !!this.state.isIndeterminate;
   }
```
cliffkoh: Nice

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -28,18 +28,15 @@ export class CheckboxBase extends React.Component<ICheckboxProps, ICheckboxState
...
+    if (nextProps.indeterminate !== undefined) {
+      prevState.isIndeterminate = !!nextProps.indeterminate;
     }
-    if (props.checked !== undefined) {
-      return {
-        isChecked: !!props.checked
-      };
+    if (nextProps.checked !== undefined) {
+      prevState.isChecked = !!nextProps.checked;
     }
```
xugao: you should not be mutating `prevState` here

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
ecraig12345: Yes, that's meant to be `state`. Not sure how I missed that...

ecraig12345: I'm wondering if it appeared to work because the mutated old state object was being preserved...

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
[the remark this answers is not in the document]
ecraig12345: fixed!

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
cliffkoh: [https://codepen.io/cliffkoh/pen/MWWeamj?editors=1010](https://codepen.io/cliffkoh/pen/MWWeamj?editors=1010) confirmed that you can mutate state and it would actually stick

---
In reply to: [335121554](https://github.com/OfficeDev/office-ui-fabric-react/pull/10644#discussion_r335121554) [](ancestors = 335121554)

PR #10364, opened 2019-09-04
SplitButton: high contrast fix

maxwellred:
#### Pull request checklist

- [] Addresses an existing issue: Fixes #10299
- [x] Include a change request file using `$ yarn change`

#### Description of changes
- fixed much of the high contrast issues shown in the Before picture. Makoto fixed the border in an earlier PR #10328 

#### BEFORE
image

#### AFTER
Split button high contrast (1)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10364)

Issue #10299, opened 2019-08-28
SplitButton borders and divider & PrimaryButton Focus not visible in high contrast/Edge


micahgodbolt:
Standard Split button divider is not visible in high contrast mode.
Primary Split button 'split' button border is not visible in high contrast mode

Edge browser.
https://fabricweb.z5.web.core.windows.net/oufr/7.28.1/#/examples/button

![image](https://user-images.githubusercontent.com/1434956/63889171-5121f200-c995-11e9-889d-a8d5f2df19b6.png)

Should be like this:

![image](https://user-images.githubusercontent.com/1434956/63889321-9e9e5f00-c995-11e9-897d-3404d1b8adb5.png)

maxwellred:
@micahgodbolt Thanks, I'll look into the high contrast selectors now.

micahgodbolt:
While you're looking at high contrast, check out the focused state of the primary button. It looks like the internal focus ring is getting covered up by the background.

PR #10656, opened 2019-09-27
SplitButton: High contrast border fix

maxwellred:
#### Pull request checklist

- [x] Addresses an existing issue: #10299 

#### Description of changes

Copies over solution from #10364 

#### BEFORE
image

#### AFTER
Split button high contrast (1)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10656)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Button/ButtonThemes.ts
+++ b/packages/office-ui-fabric-react/src/components/Button/ButtonThemes.ts
@@ -1,5 +1,6 @@
 import { IButtonStyles } from './Button.types';
 import { ITheme, HighContrastSelector } from '../../Styling';
+import { IsFocusVisibleClassName } from '../../Utilities';
 
 export function standardStyles(theme: ITheme): IButtonStyles {
   const s = theme.semanticColors;
@@ -70,6 +71,10 @@ export function standardStyles(theme: ITheme): IButtonStyles {
               color: 'Highlight'
             }
           }
+        },
+        [HighContrastSelector]: {
+          border: `1px solid white`,
+          borderLeft: 'none'
         }
       }
     },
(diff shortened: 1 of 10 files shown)


packages/office-ui-fabric-react/src/components/Button/ButtonThemes.ts
```diff
@@ -70,6 +70,10 @@ export function standardStyles(theme: ITheme): IButtonStyles {
               color: 'Highlight'
             }
           }
+        },
+        [HighContrastSelector]: {
+          border: `1px solid white`,
+          borderLeft: 'none'
```
khmakoto: Why is this here? This is different from the styling in #10364.

khmakoto:
@maxwellred Can you run `yarn update-snapshot` on the `apps/a11y-tests` folder? It seems the build is complaining about snapshots being changed.

aneeshack4:
@maxwellred Pinging since there hasn't been activity for a while now

packages/office-ui-fabric-react/src/components/Button/ButtonThemes.ts
aneeshack4: Hmm yeah I don't see this in the original fabric 7 change https://github.com/OfficeDev/office-ui-fabric-react/pull/10364/files#diff-83f447f02b4959069f9feb281f1c83b2R98

maxwellred: In Fabric 7 a border is introduced to the button in non-HC mode. Since it's not there in Fab 6 i have to add this white border only in HC for there to be one.

khmakoto:
@maxwellred once the CI passes I'll check the examples locally and, if everything looks good, I'll approve. Thanks!

packages/office-ui-fabric-react/src/components/Button/ButtonThemes.ts
aneeshack4: Slightly confused - why does this need to be added in Fabric 6 if the border is only expected for Fabric 7? There are screener changes down below so we should be careful not to break anyone still using Fabric 6.

khmakoto: I think this is because in Fabric 6 we don't use a border on the `Button` due to the background color indicating where it ends. However, this background color is no longer visible in High Contrast Mode, which is why we need a border there.

PR #10657, opened 2019-09-27
Editor: update component rendering method and add a11y props

ecraig12345:
#### Pull request checklist

#### Description of changes

Fix the React invariant error that prevented edited examples from rendering: I think this was happening because we were mixing bundled React and global React. To fix...
-  Instead of including a call to the global `ReactDOM.render()` in the transformed/eval'd example, wrap the example in a function which returns the component. That component is passed up to parents using a callback and rendered normally (using bundled or global React as the app prefers).
- In the website, always bundle React rather than loading from CDN (even in serve mode, for more realistic testing) 

Also add an aria label and accessibility help URL for the editor, per https://github.com/Microsoft/monaco-editor/wiki/Accessibility-Guide-for-Integrators .

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10657)


packages/tsx-editor/README.md
```diff
@@ -27,7 +27,7 @@ If the user's browser can't support the editor (mainly IE 11 and some mobile bro
 Any project consuming `@uifabric/tsx-editor` should follow the Webpack and runtime configuration instructions from the [`@uifabric/monaco-editor` readme](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/monaco-editor/README.md). Note that the helpers used are re-exported from this package for convenience:
 
 - `addMonacoWebpackConfig`: import from `@uifabric/tsx-editor/scripts/addMonacoWebpackConfig`
-- `configureEnvironment` and `IMonacoConfig`: import from `@uifabric/tsx-editor`
+- `configureEnvironment` and `IMonacoConfig`: import from `@uifabric/tsx-editor/lib/index-min`
```
dzearing: this seems like not an improvement. Can we avoid suggesting to import path based things?

packages/example-app-base/src/components/ExampleCard/ExampleCard.tsx
```diff
@@ -195,21 +188,33 @@ export class ExampleCardBase extends React.Component<IExampleCardProps, IExample
...
-        <CustomizerContext.Provider value={{ customizations: { settings: {}, scopedSettings: {} } }}>
-          <Customizer {...this._activeCustomizations}>
-            <ThemeProvider scheme={_schemes[this.state.schemeIndex]}>
-              <Stack styles={regionStyles}>{content}</Stack>
-            </ThemeProvider>
-          </Customizer>
-        </CustomizerContext.Provider>
-      );
+  // tslint:disable-next-line:member-ordering
+  private _getPreviewComponent = memoizeFunction(
```
kenotron: For memoizing React Functional Components, I think we should be switching to React.memo these days

ecraig12345: I'm not sure React.memo would work in this case?

packages/tsx-editor/README.md
ecraig12345: tsx-editor is kind of a special case because the normal `index` file exports all the public APIs, including lower-level ones which directly import Monaco. So other packages shouldn't directly import from that file because:
1. It greatly increases bundle size (Monaco is huge)
2. It will break Jest tests, even if the editor isn't instantiated (Monaco uses ES modules, which Jest *still* hasn't added support for, so even trying to parse it makes Jest blow up)

My solution to this was to make a separate `index-min` file which doesn't reference Monaco aside from types, but I'm open to other suggestions--maybe reversing it somehow, but I'm not sure what the new file should be called (`index-max` just sounds silly).

PR #10660, opened 2019-09-27
Teaching bubble: updated to match design toolkit 

marygans:
#### Pull request checklist

#### Description of changes

Updated teaching bubble to match latest design toolkit. Also added a separate multi-step example to differentiate between a basic teaching bubble and a multi-step teaching bubble.

#### Before 
Screen Shot 2019-09-27 at 2 39 21 PM

###### Basic Example
Screen Shot 2019-09-27 at 2 40 51 PM

###### Multi-step Example
Screen Shot 2019-09-27 at 2 41 05 PM

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10660)


packages/office-ui-fabric-react/src/components/TeachingBubble/TeachingBubbleContent.base.tsx
```diff
@@ -113,11 +114,15 @@ export class TeachingBubbleContentBase extends BaseComponent<ITeachingBubbleProp
...
-        <div className={classNames.footer}>
-          {primaryButtonProps && <PrimaryButton {...primaryButtonProps} className={classNames.primaryButton} />}
-          {secondaryButtonProps && <DefaultButton {...secondaryButtonProps} className={classNames.secondaryButton} />}
-          {customFooterContent && <span>{customFooterContent}</span>}
-        </div>
+        <Stack horizontal horizontalAlign={customFooterContent ? 'space-between' : 'end'}>
+          <Stack.Item className={classNames.footer} align="center">
+            {<span>{customFooterContent}</span>}
+          </Stack.Item>
+          <Stack.Item>
```
micahgodbolt: This is a pretty big DOM change, and will probably break people that are doing customizations on the footer thinking it covered the buttons as well. Can the footer be on the above stack?

Either way, we need to communicate this change out via Teams and the fabricreact alias so that people know this DOM change is coming.

marygans:
@micahgodbolt Yes you're right about the DOM changes potentially breaking current consumers. 
I moved the footer to the above `div` and composed a message to share to the Fabric channels cautioning them about the new changes. Anything else I should do?

micahgodbolt:
that's good with me

PR #10663, opened 2019-09-30
Changes required to support child window rendering

dzearing:
#### Description of changes

In cases where we reference the `window` object, calling `getWindow` is often not good enough; we need an element in reference to where we're rendering to get the correct window.

Hosts can create child windows and render React content there. It's essential that all DOM and async operations use the correct window to render content on. This changes ensures that critical places in the code are correctly providing the child window.

In making this all work I needed to tweak `Layer` to not render child layers during the render pass, but to first render the `<span/>` placeholder, so that we would have an element to find the correct window. Previously it would append elements to `document.body` which would have been the parent window in the child case.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10663)


packages/office-ui-fabric-react/src/components/Fabric/Fabric.styles.ts
```diff
@@ -19,7 +19,7 @@ export const getStyles = (props: IFabricStyleProps): IFabricStyles => {
   return {
     root: [
       classNames.root,
-      isFocusVisible && 'is-focusVisible',
+      isFocusVisible && 'is-focusVisible ms-Fabric--isFocusVisible',
```
dzearing: Note that without this, child windows were not receiving this class even when wrapped in a `Fabric` element, as it was only being added to `document.body`

scripts/monorepo/getAllPackageInfo.js
```diff
@@ -15,10 +15,13 @@ module.exports = function getAllPackageInfo() {
     .filter(line => line.endsWith('package.json'))
     .forEach(packageJsonFile => {
       const packageJson = readConfig(path.join(gitRoot, packageJsonFile));
-      packageInfo[packageJson.name] = {
-        packagePath: path.dirname(packageJsonFile),
-        packageJson
-      };
+
+      if (packageJson) {
```
dzearing: In one of my repos which had old directories in it, this would fail without this change.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -35,7 +35,7 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
       this._updateComposedComponentRef = this._updateComposedComponentRef.bind(this);
 
       this.state = {
-        responsiveMode: this._getResponsiveMode()
+        responsiveMode: ResponsiveMode.large
```
JasonGore: Is this an intentional change? I still see `getResponseMode` present so I wanted to make sure this wasn't a temporary change made for testing.

packages/utilities/src/dom/getWindow.ts
```diff
@@ -19,12 +21,15 @@ try {
...
-export function getWindow(rootElement?: Element | null): Window | undefined {
+export function getWindow(rootElement?: Element | Component | null): Window | undefined {
   if (_isSSR || typeof _window === 'undefined') {
     return undefined;
   } else {
-    return rootElement && rootElement.ownerDocument && rootElement.ownerDocument.defaultView
-      ? rootElement.ownerDocument.defaultView
-      : _window;
+    if (rootElement && !(rootElement as Element).ownerDocument) {
+      rootElement = findDOMNode(rootElement) as Element;
```
joschect: Don't change root element, create a new variable instead.

packages/utilities/src/dom/getDocument.ts
```diff
@@ -7,10 +10,16 @@ import { _isSSR } from './setSSR';
...
  */
-export function getDocument(rootElement?: HTMLElement | null): Document | undefined {
+export function getDocument(rootElement?: HTMLElement | Component | null): Document | undefined {
   if (_isSSR || typeof document === 'undefined') {
     return undefined;
   } else {
-    return rootElement && rootElement.ownerDocument ? rootElement.ownerDocument : document;
+    if (rootElement && !(rootElement as Element).ownerDocument) {
+      // tslint:disable-next-line:no-any
+      rootElement = findDOMNode(rootElement) as any;
```
joschect: Same as below, also why is this typed as any but the one in getWindow is typed as Element?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
dzearing: Good catch; Talked with Jon on this one too. We should change this to not render children until we know responsive mode. I had made this change to avoid touching window before being mounted, but we shouldn't assume large until we know.

packages/utilities/src/Async.ts
```diff
@@ -141,6 +143,7 @@ export class Async {
    */
   public setImmediate(callback: () => void): number {
     let immediateId = 0;
+    const win = getWindow()!;
```
joschect: Do you want to allow a passthrough for an htmlelement or react component to this?

packages/office-ui-fabric-react/src/components/Layer/Layer.base.tsx
```diff
@@ -19,213 +20,133 @@ export class LayerBase extends React.Component<ILayerProps, ILayerBaseState> {
...
 
-    const { onLayerWillUnmount, hostId } = this.props;
-    if (onLayerWillUnmount) {
-      onLayerWillUnmount();
-    }
+    this._createLayerElement();
 
     if (hostId) {
-      unregisterLayer(hostId, this);
+      registerLayer(hostId, this._createLayerElement);
```
joschect: Did you mean to call registerLayer in componentWillUnmount?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
dzearing: I've added an `unknown` size, which by default avoids rendering the inner component until the window is measured. Once that's know we render. This prevents us from accessing the window until mounted.

PR #10666, opened 2019-09-30
MessageBar: A11y bug fixes and Doc updates

### LINKED ISSUE #7762: MessageBar A11y Pass
Tracking all of the issues brought up with the 1/22 accessibility pass on MessageBar control 
Details on individual bugs can be found on [SharePoint](https://microsoft-my.sharepoint.com/:x:/r/personal/v-amaver_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B8f6b9510-2d32-45ba-bc39-37aa879c7eba%7D&action=default&uid=%7B8F6B9510-2D32-45BA-BC39-37AA879C7EBA%7D&ListItemId=1676&ListId=%7B395E1ECB-5002-440E-81C5-6A090E98077C%7D&odsp=1&env=prodbubble) 
  
 - [x] Bug1: Bug1: [Accessibility][Visual requirements]: Luminosity ratio of link 'Visit our website' under 'Error, single line, with dismiss button - custom styles' section is less than 4.5:1 ratio.  
 - [x] Bug2: [Accessibility][Keyboard navigation]: Focus indicator is not properly visible on  'visit our website' link. 
 - [x] Bug3: [Accessibility][Screen Reader]: Narrator in scan  mode doesn't announce the role of  'See More' , 'Close ('X')' buttons and 'visit our website' link  under 'Blocked, single line, with dismiss button and truncated text - custom styles' section. " 
 - [x] Bug4: [Accessibility][Screen Reader]: Narrator doesn't announce the state of 'See More' button while navigating with tab key. 
 - [x] Bug5: [Accessibility][Keyboard Navigation]: Text is disappearing from the screen when keyboard focus moves to the 'See more' button. " 
- [x] Bug6: [Accessibility][ Supporting the platform]: Controls present in the ‘MessageBar’ are not adopting high contrast settings. " 
- [x] Bug7: [Accessibility][Suggestion]: Appeared message bar should have ‘Aria-live’ property and region landmark. " 
  
 Component: MessageBar  
@micahgodbolt

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #7762

#### Description of changes

Addressed a11y issues brought up from recent pass. 

Updated examples to be easier to demo with screen readers. 

image

#### Focus areas to test

(optional)


packages/office-ui-fabric-react/src/components/MessageBar/examples/MessageBar.Styled.Example.tsx
```diff
@@ -1,222 +0,0 @@
-import * as React from 'react';
-import { MessageBarButton } from 'office-ui-fabric-react/lib/Button';
-import { Link } from 'office-ui-fabric-react/lib/Link';
-import { Label } from 'office-ui-fabric-react/lib/Label';
-import { Stack, StackItem } from 'office-ui-fabric-react/lib/Stack';
-import { MessageBar, MessageBarType } from 'office-ui-fabric-react/lib/MessageBar';
-
-const log = (text: string): (() => void) => (): void => console.log(text);
-
-export const MessageBarStyledExample: React.StatelessComponent = () => (
```
ecraig12345: Might be worth preserving one of these as an example of how to do custom styling (just pick an accessible one)

packages/office-ui-fabric-react/src/components/MessageBar/examples/MessageBar.Basic.Example.tsx
```diff
@@ -1,160 +1,184 @@
...
+      messageBarType={MessageBarType.warning}
+      ariaLabel="Aria help text here"
+      actions={
+        <div>
+          <MessageBarButton>Yes</MessageBarButton>
+          <MessageBarButton>No</MessageBarButton>
+        </div>
+      }
+    >
+      Warning lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi luctus, purus a lobortis tristique, odio augue pharetra metus,
```
ecraig12345: Maybe add back something about this example allowing overflowing to multiple lines, since it omits the `isMultiline={false}`? (we will most definitely want to rethink some of these option names/behavior when rewriting...)

packages/office-ui-fabric-react/src/components/MessageBar/examples/MessageBar.Basic.Example.tsx
```diff
@@ -1,160 +1,184 @@
...
-          Visit our website.
-        </Link>
-      </MessageBar>
-    </StackItem>
+  const WarningExample2 = () => (
+    <MessageBar
+      onDismiss={resetChoice}
+      dismissButtonAriaLabel="Close"
+      messageBarType={MessageBarType.warning}
+      ariaLabel="Aria help text here"
```
ecraig12345: Not sure what an aria-label will do on the MessageBar? Probably best to remove it or at least make it more descriptive.

packages/office-ui-fabric-react/src/components/MessageBar/examples/MessageBar.Basic.Example.tsx
```diff
@@ -1,160 +1,184 @@
...
-    <StackItem>
-      <Label>Info/Default MessageBar</Label>
-      <MessageBar>
-        Info lorem ipsum dolor sit amet, a elit sem interdum consectetur adipiscing elit.{' '}
-        <Link href="www.bing.com" target="_blank">
-          Visit our website.
-        </Link>
-      </MessageBar>
-    </StackItem>
+  const DefaultExample = () => (
```
ecraig12345: These should probably go outside the example component function.

packages/office-ui-fabric-react/src/components/MessageBar/examples/MessageBar.Basic.Example.tsx
```diff
@@ -1,160 +1,184 @@
...
-        Warning lorem ipsum dolor sit amet, a elit sem interdum consectetur adipiscing elit.{' '}
-        <Link href="www.bing.com" target="_blank">
-          Visit our website.
-        </Link>
-      </MessageBar>
-    </StackItem>
-  </Stack>
-);
+  return (
+    <Stack horizontal tokens={{ childrenGap: 16 }}>
```
ecraig12345: Recently we've been trying to move towards avoiding inline objects or functions since they mutate every render and break memoization if it's being used. (Currently most of our components aren't memoized, but I think the idea was to demonstrate best practices to people who may be copy-pasting our examples.) 

Usually for any constant objects or functions (don't depend any variables), I've been moving them somewhere outside the component. For the ones that depend on state or something you can use `useCallback` or `useMemo`.

packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.base.tsx
```diff
@@ -132,7 +132,7 @@ export class MessageBarBase extends BaseComponent<IMessageBarProps, IMessageBarS
     const { theme } = this.props;
     return (
       <div style={{ background: theme!.semanticColors.bodyBackground }}>
-        <div className={this._classNames.root}>
+        <div className={this._classNames.root} role="region">
```
dzearing: I'm not sure this is correct.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role

> Every element with a region role should include a label that describes the purpose of the content in the region, preferably with an aria-labelledby referencing a visible header. If no visible appropriate header is present, aria-label should be used..

There doesn't seem to be any way to get a label here.

micahgodbolt:
I'm planning on removing the 'aria-controls' label on the show/hide truncate button. I agree that it visually changes the message bar, but it doesn't change the content, and the description for aria-controls seems to imply this would be used for something else on the page that controls the "role=status" element:

> If another part of the page controls what appears in the status, authors SHOULD make the relationship explicit with the aria-controls attribute.

packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.base.tsx
micahgodbolt: updated

packages/office-ui-fabric-react/src/components/MessageBar/examples/MessageBar.Basic.Example.tsx
micahgodbolt: addressed

micahgodbolt:
@dzearing @ecraig12345 are we feeling ok about this one?

packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.base.tsx
```diff
@@ -149,8 +149,14 @@ export class MessageBarBase extends BaseComponent<IMessageBarProps, IMessageBarS
     const nativeProps = getNativeProps<React.HTMLAttributes<HTMLSpanElement>>(this.props, htmlElementProperties, ['className']);
 
     return (
-      <div className={this._classNames.text} id={this.state.labelId}>
-        <span className={this._classNames.innerText} role="status" aria-live={this._getAnnouncementPriority()} {...nativeProps}>
+      <div className={this._classNames.text}>
+        <span
+          className={this._classNames.innerText}
+          id={this.state.labelId}
```
ecraig12345: I'm curious, what's the reason for moving the id prop? That has the potential to break people's tests or other assumptions.

packages/office-ui-fabric-react/src/components/MessageBar/examples/MessageBar.Basic.Example.tsx
ecraig12345: Are you planning to address this? In general I don't think we should be demonstrating creating components within components.

packages/office-ui-fabric-react/src/components/MessageBar/examples/MessageBar.Basic.Example.tsx
```diff
@@ -1,160 +1,195 @@
...
+
+  return (
+    <Stack {...horizontalStackProps}>
+      <StackItem disableShrink>
+        <ChoiceGroup
+          label="Select a MessageBar Example Below"
+          selectedKey={choice}
+          // tslint:disable-next-line: jsx-no-lambda
+          onChange={(e, v) => setChoice(v!.key)}
+          options={[
```
ecraig12345: These options should probably also go in a const outside the body of the component, so the array doesn't mutate every re-render.

ecraig12345:
@micahgodbolt Left a couple more comments. Sorry to be picky about so many small things, but it's pretty important that our examples demonstrate good practices (because they'll be copy-pasted and/or used as patterns so much).

packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.base.tsx
micahgodbolt: yeah, good point. I think i can move it back

packages/office-ui-fabric-react/src/components/MessageBar/examples/MessageBar.Basic.Example.tsx
micahgodbolt: Most of the MessageBar examples need access to state in order to reset it. It'd be odd to have most of them inside and then a few outside. 

I also don't want to create addition abstractions to pass this resetChoice functionality into the control. 

I don't see this as component inside a component, but rather components inside of a stateful view.

micahgodbolt: ehh....i guess i already have basic abstraction. i can move the callback into props without too much worry.

Still, i'm always concerned with easy to read demos vs best practice code demos

packages/office-ui-fabric-react/src/components/MessageBar/docs/MessageBarDonts.md
```diff
@@ -1,2 +1,3 @@
 - Don’t use paragraphs, long sentences, or special formatting in a MessageBar. The control tries to grow to accommodate all the text and will just result in pushing the user’s main content too low on the view.
 - Don’t use buttons when a subtler link will suffice. Reserve the usage of button for when the MessageBar has a single ”hero” action that has vital usefulness to the user at that particular moment. Using more than one button is discouraged.
+- Don't overuse message bars, or load a page with multiple message bars because each MessageBar is read out loud by screen readers.
```
ecraig12345: Good addition! Might be good to be even more explicit that they're read on page load (assuming that's correct).
```suggestion
- Don't overuse MessageBars or load a page with multiple MessageBars, because screen readers will read each MessageBar on page load.
```

PR #10668, opened 2019-09-30
Add an optional prop to base button to render a persisted menu hidden on mount

whykushal93:
#### Pull request checklist

#### Description of changes
BaseButton, until a few months ago, mounted a persisted menu hidden as soon as the BaseButton was mounted. This resulted in an [Edge bug](https://github.com/OfficeDev/office-ui-fabric-react/issues/9034), where the menu's scrollbars are off.

To fix this Edge bug, some code was added to only mount the menu when it is visible for the first time in [this](https://github.com/OfficeDev/office-ui-fabric-react/pull/9124) This, however, breaks a perf. improvement I have been working on which depends on the menu mount to be triggered when hidden. 

Since this bug only reproes in Edge, I tried to get in touch with Edge teams to try and see if we can find a workaround. Unfortunately, they are pretty busy with other feature work to take a look at this. I have also unsuccessfully tried various approaches to fix this (hit and trial) myself. 

As a final solution to unblock me, I am adding this optional prop which is also deprecated from the beginning. This is so that we can remove it and make the behavior default when the bug is fixed on Edge's side. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10668)


packages/office-ui-fabric-react/src/components/Button/Button.types.ts
```diff
@@ -289,6 +289,25 @@ export interface IButtonProps
    */
   persistMenu?: boolean;
 
+  /**
+   * If true, the persisted menu is rendered hidden when the button
```
micahgodbolt: would like a note about why this is already deprecated. sounds like this flag turns on experimental support for faster 'persistMenu' that isn't supported by edge yet.

micahgodbolt: almost makes me want an 'experimental' type prop

packages/office-ui-fabric-react/src/components/Button/Button.types.ts
```diff
@@ -289,6 +289,24 @@ export interface IButtonProps
...
+   * initially mounts. Non-persisted menus will
+   * not be in the component tree unless they are being shown
+   *
+   * Note: This increases the time the button will take to mount, but
+   * can improve perceived menu open perf. when the user opens the menu.
+   *
+   * @default undefined, equivalent to false
```
khmakoto: Should be `@defaultvalue`.

joschect:
@whykushal93 this feels like a super specific fix that's only useful in a few scenarios. Would it be possible to pass in a rendering function to `menuAs` that would do this conditional rendering without the need for the button to have another prop?

whykushal93:
@joschect I tried doing that. The problem though is the way the BaseButton is written right now, it does not even try to render the menu until it is shown the first time. 

I do understand this is specific. But, this is the way I see it. We changed a certain behavior of persisted menus (that they are always present in DOM) because of the Edge specific bug. I do understand that this was necessary. All I am doing by introducing this prop is to get back to the original behavior.

joschect:
@whykushal93 Here is a [codepen](https://codepen.io/slimemold/pen/pozMyvV?editors=1010) It opens and then dismisses the menu immediately. It's definitely a hack but it would unblock your scenario for right now. The menu never flickers or opens because of the way that setState gets called.

whykushal93:
Thanks for prototyping it for me, @joschect. That sounds like it could work. However, it would complicate the code on our side though. We have handlers that do critical things when a menu is opened and when a menu is closed. I would need to disable these handlers when I phantom open and close the menu, and re-enable them on the menu when that is complete. 

I am not 100% against doing it if it is absolutely necessary. But, I feel the behavior enabled by this prop should actually be default in BaseButton, if not for the Edge bug.

PR #10674, opened 2019-10-01
Dropdown: Making multiselect options clickable outside of the options label

### LINKED ISSUE #10669: Multi select <Dropdown/> doesn't select item when clicking outside of label
### Environment Information

https://codepen.io/mshindal/pen/RwbzZXG

#### Actual behavior:

When using a `<Dropdown/>` that has multi select enabled, clicking outside of the option's label (the text) does not select the option. 

This .gif demonstrates the issue. First, I click outside of the label in a single select dropdown, which selects the option. Then, I click outside of the label in a multi select dropdown, and the option is not selected. 

recording

#### Expected behavior:

Clicking anywhere on the option in multi-select mode should select it, to be consistent with the single select behavior. 

Requested priority: High

Products/sites affected: N/A

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10669 

#### Description of changes

The behavior between single select and multi select `Dropdowns` was inconsistent because in single select `Dropdowns` one can select/unselect an option by clicking outside of the option's label if you are still within the boundaries of the `Dropdown's` callout, while in multi select `Dropdowns` one cannot select/unselect the option this way. This was happening because multi select `Dropdowns` use `Checkboxes` instead of `Buttons` for the options, which are only selected by clicking on either the checkbox itself or the label associated with it. This PR fixes this scenario by expanding the label inside of the `Checkbox` in multi select `Dropdowns` to cover the entire height width of its container.

__Before:__

Dropdown

__After:__

Dropdown

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10674)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
@@ -630,6 +630,10 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
         role="option"
         aria-selected={isItemSelected ? 'true' : 'false'}
         checked={isItemSelected}
+        // Since the Checkbox is triggered via a click on either the checkbox itself or its label, we need to make the label span the
+        // entire width of the Dropdown so users can select and unselect the checkbox as long as they are clicking in the area adjacent
+        // to it.
+        styles={{ label: { width: '100%' } }}
       />
     );
   };
--- a/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
@@ -21,7 +21,7 @@ import {
   warnMutuallyExclusive
 } from '../../Utilities';
 import { Callout } from '../../Callout';
-import { Checkbox } from '../../Checkbox';
+import { Checkbox, ICheckboxStyleProps, ICheckboxStyles } from '../../Checkbox';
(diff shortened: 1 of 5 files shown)


packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -630,6 +630,10 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
         role="option"
         aria-selected={isItemSelected ? 'true' : 'false'}
         checked={isItemSelected}
+        // Since the Checkbox is triggered via a click on either the checkbox itself or its label, we need to make the label span the
+        // entire width of the Dropdown so users can select and unselect the checkbox as long as they are clicking in the area adjacent
+        // to it.
+        styles={{ label: { width: '100%' } }}
```
dzearing: any way here to provide this in a styles file and not inline?

joschect: I don't think that this is the right fix, it still won't expand to fill the entire container height wise. So the clickbox still isn't the whole highlightable space.

khmakoto: @dzearing I'll move the styles to a styles file.

@joschect You're right, I missed that part, great catch.

PR #10675, opened 2019-10-01
Fixing boolean rendering in list

### LINKED ISSUE #10671: DetailsList component not showing boolean values in rows.
### Environment Information

- **Package version(s)**: 6.175.0
- **Browser and OS versions**: Chrome/Windows 10

https://codepen.io/linkup/pen/xxKoXeR

#### Actual behavior:

When one of the value types for DetailList is boolean the data is not shown in the table.

#### Expected behavior:

It should show, true/false accordingly

Products/sites affected: (if applicable) Dynamics 365 for Customer Insights.

linkup89:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10671 

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10675)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/DetailsList/DetailsRowFields.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/DetailsRowFields.tsx
@@ -11,6 +11,10 @@ const getCellText = (item: any, column: IColumn): string => {
     value = '';
   }
 
+  if (typeof value === 'boolean') {
+    return String(value);
+  }
+
   return value;
 };
 
--- a/packages/office-ui-fabric-react/src/components/DetailsList/DetailsRowFields.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/DetailsRowFields.tsx
@@ -12,7 +12,7 @@ const getCellText = (item: any, column: IColumn): string => {
   }
 
   if (typeof value === 'boolean') {
-    return String(value);
+    return value.toString();
   }
 
   return value;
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/DetailsList/DetailsRowFields.tsx
```diff
@@ -11,6 +11,10 @@ const getCellText = (item: any, column: IColumn): string => {
     value = '';
   }
 
+  if (typeof value === 'boolean') {
+    return String(value);
```
KevinTCoughlin: I'm admittedly micro-bench-marking here, but this is a hot code path. It looks like calling `toString` is faster compared with concatenation and `String(this)` approaches: https://jsperf.com/boolean-tostring/1.

image

Given the above, I think we should consider `value.toString()` here.

KevinTCoughlin: I'll write the unit test in a follow-up PR if you just want to weigh-in with the above

linkup89: Was using String(x) cause of the null check capabilities but there is already a null check before this, so .toString should work. And if it's faster the better :)

KevinTCoughlin: > Was using String(x) cause of the null check capabilities but there is already a null check before this, so .toString should work. And if it's faster the better :)

You should be well-guarded against `null` and `undefined` because they won't equate to `boolean`:

```js
typeof null
"object"
typeof undefined
"undefined"
```

PR #10697, opened 2019-10-03
Callout & Tooltip: Making background color themable.

### LINKED ISSUE #10679: TooltipHost background color isn't set by theme
### Environment Information

- **Package version(s)**: 7.29.0
- **Browser and OS versions**: any

#### Actual behavior:
TooltipHost uses `theme.palette.neutral` for the root background color, despite accepting a theme as a prop, so the root's background color never changes.

#### Expected behavior:
TooltipHost should use a `theme.palette.neutral*` or `theme.semanticColors.*` color instead of `theme.palette.white` for the root background color. Otherwise, the content's colors are updated from the theme (`color: palette.neutralPrimary`), but the background color isn't, making the component inaccessible and styled incorrectly. 

Requested priority: High for those using TooltipHost

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10679

#### Description of changes

`Tooltip` was using `palette.white` for its background, which was not very friendly for theming so I changed it, after conversations with design, to use `semanticColors.menuBackground` instead. However, I realized that this was not fixing the issue because `Callout` was also using `palette.white` for its background, so I also made the change to use `semanticColors.menuBackground` as part of this PR.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10697)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-02-17-24-07-tooltipBackgroundHover.json
+++ b/change/office-ui-fabric-react-2019-10-02-17-24-07-tooltipBackgroundHover.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Callout & Tooltip: Making background color themable.",
+  "packageName": "office-ui-fabric-react",
+  "email": "Humberto.Morimoto@microsoft.com",
+  "commit": "6aa84713cd454fdab87851df68bc9af845eec00d",
+  "date": "2019-10-03T00:24:07.290Z"
+}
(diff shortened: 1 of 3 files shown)


dzearing:
@khmakoto Thank you!

PR #10711, opened 2019-10-03
Data attribute conformance

micahgodbolt:
Giving this a first pass. Worked with @joschect to sort out some of the challenges when moving from try inside of test vs tests inside of try.

Not 100% sure on Jest.Mock call, and if that's being done right. Would like to have fewer exceptions on the initial set of mounted components, but finding issues i'm guessing is related to that mock call. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10711)


ecraig12345:
Is this going anywhere or should it be closed?

micahgodbolt:
good question...tossed it up to get some eyes, but haven't gotten any so far. it's not the best solution, so I'm happy to table it for now.

ecraig12345:
I guess we have too many other new things to think about right now? 😆 I could potentially take a look at this though if you could add more detail in the description about what it's doing and why.

micahgodbolt:
nah, i'll take a another stab at it later. the idea is extending our conformance testing to test other things besides classnames (like data/aria attributes). There's some re-org needed in the code to support different sets of exceptions. Probably a better all up solution we could write to support N number of tests.

PR #10714, opened 2019-10-04
Fix editor bundling issues and add example error boundary

ecraig12345:
#### Pull request checklist

#### Description of changes

Change the `tsx-editor` package to use a more conventional import structure:
- delete `index-min` files
- `index` files only include things which **don't** import Monaco
- add root-level files for public APIs `EditorWrapper`, `TsxEditor`, and `Editor` (the latter two are not exported from the main `index` due to importing Monaco)

Add an error boundary so that runtime errors in the example won't crash the whole page.

Fix a bug which was preventing full type checking from being enabled.

Disable transpiling/rendering example until types are fully loaded. (This reduces the chance of runtime errors and helps with various other edge cases.)

Remove code visibility toggle from tsx-editor and move it to ExampleCard (it's not a generally useful behavior and made the error handling logic more complicated).

Fix monaco-bundle imports and helpers so that consuming packages don't try to load non-transformed CSS. (for some reason this worked okay previously in OUFR, but when I tried to make a package consuming tsx-editor, it didn't work)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10714)

### DIFF
--- a/packages/tsx-editor/src/components/EditorErrorHandler.tsx
+++ b/packages/tsx-editor/src/components/EditorErrorHandler.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { ITransformedExample } from '../interfaces';
+import { ITransformedExample } from '../interfaces/index';
 import { EditorError } from './EditorError';
 
 export interface IEditorErrorBoundaryProps {
--- a/packages/tsx-editor/src/components/EditorErrorHandler.tsx
+++ b/packages/tsx-editor/src/components/EditorErrorHandler.tsx
@@ -1,0 +1,70 @@
+import * as React from 'react';
+import { ITransformedExample } from '../interfaces';
+import { EditorError } from './EditorError';
+
+export interface IEditorErrorBoundaryProps {
+  /**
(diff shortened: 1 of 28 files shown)


packages/tsx-editor/src/components/EditorErrorHandler.tsx
```diff
@@ -0,0 +1,70 @@
...
+      // keep track of the last known-good children, to reuse if there's an error
+      this._lastGoodChildren = props.children;
+    }
+  }
+
+  public componentDidCatch(error: Error, errorInfo: React.ErrorInfo): void {
+    // Log the error to the console so people can see the full stack/etc if they want
+    // (only in production, because React logs these errors itself in dev mode)
+    if (process.env.NODE_ENV === 'production') {
+      console.error(error.stack || error);
```
kenotron: Should we be displaying this more than console? I thought we had an error message box for them to see?

PR #10720, opened 2019-10-04
ComboBox: Making error messages accessible to screen readers

### LINKED ISSUE #10693: ComboBox errorMessage is not read by screen reader
### Environment Information

- **Package version(s)**: version 7 (latest at time of writing = 7.42.0)
- **Browser and OS versions**: Chrome 77.0.3865.90 and Edge 44.18362.329.0 (on Windows 10)

### Describe the issue:
Neither Narrator with Edge nor NVDA with Chrome reads/announces the error message of a ComboBox component when the error is set/made visible.

https://codepen.io/mirajp/pen/jONgJag

#### Actual behavior:
Error message shows up but is not read aloud.

#### Expected behavior:
Error message to show up and and be read aloud as an alert.

https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html ?

(I noticed that the rendered element doesn't have any special roles/aria-live attributes).

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10693

#### Description of changes

Error messages in `ComboBox` were not being read by screen readers, which creates an accessibility problem because non-sighted users would not be able to know that there has been an error in their input. To fix that this PR introduces 2 changes:

1. Set `aria-describedby` to the `ComboBox` so that when users focus on the `ComboBox` the screen readers correctly read that an error exists on that input.
2. Set `aria-live` on the `errorMessage` so that if the `ComboBox's` value changes to an invalid one, the error message is announced by screen readers at that moment.

I've added a new example with a dynamic error message to test for the second scenario.

One thing to consider is that we need the `errorMessage` region to be rendered and changed so that `aria-live` can work correctly, but we don't want to always render this if there is no value passed to the `errorMessage` prop. Given this, I've added a `Do` specifying to:

>Set empty strings values and not `undefined` to the `errorMessage` prop when trying to spout dynamic error messages so that aria-live can work correctly for screen readers.

For more info on why we need this:

> Assistive technologies will announce dynamic changes in the content of a live region. The live region must first be present (and usually empty), so that the browser and assistive technologies are aware of it. Any subsequent changes are then announced.
>
>Simply including an aria-live attribute or a specialized live region role (such as role="alert") in the initial markup as it's loaded will have no effect.
>
>Dynamically adding an element with an aria-live attribute or specialized role to the document also won't result in any announcement by assistive technologies (as, at that point, the browser/assistive technologies are not aware of the live region yet, so cannot monitor it for changes).
>
>Always make sure that the live region is present in the document first, and only then dynamically add/change any content.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10720)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
@@ -311,6 +311,7 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
   // Primary Render
   public render(): JSX.Element {
     const id = this._id;
+    const errorMessageId = id + '-error';
     const {
       className,
       label,
@@ -393,7 +394,7 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
                 aria-label={ariaLabel && !label ? ariaLabel : undefined}
                 aria-describedby={
                   errorMessage !== undefined
-                    ? mergeAriaAttributeValues(ariaDescribedBy, keytipAttributes['aria-describedby'], this._id + '-error')
+                    ? mergeAriaAttributeValues(ariaDescribedBy, keytipAttributes['aria-describedby'], errorMessageId)
                     : mergeAriaAttributeValues(ariaDescribedBy, keytipAttributes['aria-describedby'])
                 }
                 aria-activedescendant={this._getAriaActiveDescentValue()}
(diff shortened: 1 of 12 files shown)


packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -435,7 +439,11 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
             },
             this._onRenderContainer
           )}
-        {errorMessage && <div className={this._classNames.errorMessage}>{errorMessage}</div>}
+        {errorMessage !== undefined && (
+          <div role="region" aria-live="polite" aria-atomic="true" id={this._id + '-error'} className={this._classNames.errorMessage}>
```
joschect: `this._id + '-error'` should be either in a const or a separate function so it is consistent.

PR #10723, opened 2019-10-04
Enable editor by default

ecraig12345:
#### Pull request checklist

#### Description of changes

Turn the live example editor on by default! (leaving an option to turn it off)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10723)

### DIFF
--- a/packages/tsx-editor/src/utilities/isEditorSupported.ts
+++ b/packages/tsx-editor/src/utilities/isEditorSupported.ts
@@ -11,8 +11,8 @@ export function isEditorSupported(code: string, supportedPackages: IBasicPackage
     !!win &&
     // Required environment config available
     !!getMonacoConfig() &&
-    // Opt-in query param or session storage is set
-    getSetting('useEditor') === '1' &&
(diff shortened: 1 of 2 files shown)


packages/tsx-editor/src/utilities/isEditorSupported.ts
```diff
@@ -11,8 +10,6 @@ export function isEditorSupported(code: string, supportedPackages: IBasicPackage
     !!win &&
     // Required environment config available
     !!getMonacoConfig() &&
-    // Opt-in query param or session storage is set
-    getSetting('useEditor') === '1' &&
```
kenotron: Maybe we should keep a flag here allow to do:

```
useEditor = 0
```

to turn off the feature?

ecraig12345: Yeah I think that's a good idea.

PR #10725, opened 2019-10-05
ColorPicker: adding keyboard accessibility.

dzearing:
ColorPicker swatch rectangle and sliders were not keyboard accessible; now they are!

* Rectangle: up/down/left/right + shift modifier to go +10 instead of +1.
* Sliders: left/right/home/end + shift modifier
* Focus rect shows up but only when using the keyboard to tab around
* RTL safe

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10725)


ecraig12345:
Closing in favor of a version I'm about to submit.

PR #10728, opened 2019-10-05
cherrypick: Remove ReactDOM dependency in utilities.

dzearing:
@KevinTCoughlin 

Here's the cherrypick for the removal of `findDOMNode` from the utilities package.

This should help with bundle concerns for 6.0 users.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10728)

### DIFF
--- a/packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
+++ b/packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { findDOMNode } from 'react-dom';
 import { BaseDecorator } from './BaseDecorator';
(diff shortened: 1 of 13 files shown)


packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { findDOMNode } from 'react-dom';
```
dzearing: this looks like a bad merge!

ipip2005:
Thanks a lot!

PR #10730, opened 2019-10-07
Callout and HoverCard: Fix anchor positions

### LINKED ISSUE #10721: Component positioning going off screen
### Environment Information

- **Package version(s)**: 7.43.1

https://codepen.io/slimemold/pen/bGGGdzY

Examples of repro's in OWA:
image
image

#### Actual behavior:

Components are rendering off the page when near the edges.

#### Expected behavior:

Component positioning should render the whole component on the page

Requested priority: High

Products/sites affected: OWA

joschect:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10721

#### Description of changes

Callout and HoverCard: Fix anchor positions

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10730)

### DIFF
--- a/packages/office-ui-fabric-react/src/utilities/positioning/positioning.ts
+++ b/packages/office-ui-fabric-react/src/utilities/positioning/positioning.ts
@@ -442,22 +442,13 @@ function _getFlankingEdges(edge: RectangleEdge): { positiveEdge: RectangleEdge; 
 /**
  * Retrieve the final value for the return edge of elementRectangle.
  * If the elementRectangle is closer to one side of the bounds versus the other, the return edge is flipped to grow inward.
- * The finalized return edge should not flip if the element has already been positiuoned successfully to prevent the callout from
- * appearing to jump
  *
  * @param elementRectangle
  * @param targetEdge
  * @param bounds
- * @param previousPositions
  */
-function _finalizeReturnEdge(
-  elementRectangle: Rectangle,
-  returnEdge: RectangleEdge,
-  bounds?: Rectangle,
-  previousPositions?: boolean
-): RectangleEdge {
+function _finalizeReturnEdge(elementRectangle: Rectangle, returnEdge: RectangleEdge, bounds?: Rectangle): RectangleEdge {
   if (
-    !previousPositions &&
     bounds &&
     Math.abs(_getRelativeEdgeDifference(elementRectangle, bounds, returnEdge)) >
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/utilities/positioning/positioning.ts
```diff
@@ -481,7 +472,6 @@ function _finalizeReturnEdge(
  * @param {RectangleEdge} bounds
  * @param {RectangleEdge} [alignmentEdge]
  * @param {boolean} coverTarget
- * @param {previousPositions} previousPositions
```
jdhuntington: add comment for new param

packages/office-ui-fabric-react/src/utilities/positioning/positioning.ts
```diff
@@ -491,19 +481,17 @@ function _finalizeElementPosition(
   bounds?: Rectangle,
   alignmentEdge?: RectangleEdge,
   coverTarget?: boolean,
-  previousPositions?: boolean
+  doNotFinalizeReturnEdge?: boolean
```
jdhuntington: It's my own bias, but I think functions are easier to understand if they don't take boolean flags as args. They tend to accumulate over time and are not easy to grok when looking at invocations.

Since this method is small, can there be 2 methods instead of an `if`? I think this lets you walk back several changes and have `_positionCard` make a more straightforward method call all the way through.

packages/office-ui-fabric-react/src/utilities/positioning/positioning.ts
```diff
@@ -877,6 +867,15 @@ export function positionCallout(
   return _positionCallout(props, hostElement, elementToPosition, previousPositions);
 }
 
+export function positionCard(
```
khmakoto: Why are creating a public `positionCard` and a private `positionCard` method here?

joschect: Because then we can do whatever we want to change the private position card and keep entirely out of the api.

packages/office-ui-fabric-react/src/utilities/positioning/positioning.ts
joschect: I think that we could do that in the future. For now it's hidden enough that it should be okay.

PR #10113, opened 2019-08-09
Changing breadcrumb items to links

caiomontei:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes breadcrumb items being rendered as buttons instead of links.
- [x] Include a change request file using `$ yarn change`

#### Description of changes

Set prop to render breadcrumb item links as anchor tags instead of buttons.

#### Focus areas to test

Breadcrumb navigation

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10113)

PR #10736, opened 2019-10-08
Breadcrumb: fix accessibility and styling issues

### LINKED ISSUE #10138: Inconsistent link hover behavior in Breadcrumb
### Environment Information

https://codepen.io/atneik/pen/JjPGZzY

Items with no link (no `onClick`), behave differently when collapsed as compared to not collapsed.

#### Actual behavior:

When `onClick` is not defined in an item of Breadcrumb, the hover behavior of item is different when collapsed vs when not collapsed. 

#### Expected behavior:

 The hover behavior should be same. It should follow the not collapsed behavior, i.e. no cursor change or background color change in Dropdown upon hover when `onClick` is not defined.

Requested priority: Normal

Products/sites affected: SharePoint

### LINKED ISSUE #10750: Breadcrumbs as headings example: illogical headings
### Environment Information

### Describe the issue:

In the [Breadcrumbs as headings](https://developer.microsoft.com/en-us/fabric/#/controls/web/breadcrumb) example the order of the headings goes from `h4` to `h3` to `h2`: 

- `h4`: Files
- `h3`: This is folder 1
- `h2`: This is folder 2

This is not a logical order. If these were each headings (and they shouldn't be) Files would be the most important with "This is folder 1" being the first item under it and "This is folder 2" the first item under that. Therefore the order would be:

- `h2`: Files
- `h3`: This is folder 1
- `h4`: This is folder 2

But as I said, it doesn't make sense for each of these to be headings. This is a navigation. It might make sense for the _last_ item to be a heading but not the others.

I think this was coded this way because someone wanted to not have to have a heading on their page and decided to jam it in the breadcrumb. It's key to recognize what content is falling under each heading. 

#### Actual behavior:

Headings within the breadcrumb nav do not logically act as headings. They are broken in a way that _might_ pass automated checking.

Observed heading levels
- `h4`: Files (What is this a heading of?)
- `h3`: This is folder 1 (What is this a heading of?)
- `h2`: This is folder 2 

#### Expected behavior:

- Files (Should not be a heading)
- This is folder 1 (Should not be a heading)
- `h2`: This is folder 2 (Probably shouldn't be a heading but that will depend on context so let's leave it to the author)

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10138, fixes an issue introduced by PR #10113, and fixes #10750

#### Description of changes

#10113 switched breadcrumb items to always use links instead of buttons, which broke "clicking" `href`-less breadcrumb items using the keyboard. I'm reverting that change since it's such a major accessibility issue, and updating the `IBreadcrumbItem.as` prop to allow forcing individual breadcrumb items to be links (to facilitate whatever the PR author was trying to do).

This change also updates the hover styling for non-clickable breadcrumb items in the overflow menu to match the design spec. The current approach--marking the item as disabled and overriding the disabled text color--isn't ideal, but see https://github.com/OfficeDev/office-ui-fabric-react/issues/10138#issuecomment-539304332 for more about the reasoning and alternatives (and I'm open to suggestions).

Also, per #10750, I'm updating the breadcrumb-as-heading example to have only the last breadcrumb be a heading.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10736)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Breadcrumb/Breadcrumb.types.ts
+++ b/packages/office-ui-fabric-react/src/components/Breadcrumb/Breadcrumb.types.ts
@@ -130,8 +130,10 @@ export interface IBreadcrumbItem {
 
   /**
    * Optional prop to render the item as a heading of your choice.
+   *
    * You can also use this to force items to render as links instead of buttons (by default,
-   * any item with a `href` renders as a link, and any item without a href renders as a button).
+   * any item with a `href` renders as a link, and any item without a `href` renders as a button).
+   * This is not generally recommended because it may prevent activating the link using the keyboard.
    */
   as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'a';
 }
--- a/packages/office-ui-fabric-react/src/components/Breadcrumb/Breadcrumb.types.ts
+++ b/packages/office-ui-fabric-react/src/components/Breadcrumb/Breadcrumb.types.ts
@@ -41,7 +41,7 @@ export interface IBreadcrumbProps extends React.HTMLAttributes<HTMLElement> {
   items: IBreadcrumbItem[];
 
   /**
-   * Optional root classname for the root breadcrumb element.
+   * Optional class for the root breadcrumb element.
    */
   className?: string;
 
@@ -60,23 +60,22 @@ export interface IBreadcrumbProps extends React.HTMLAttributes<HTMLElement> {
    */
   maxDisplayedItems?: number;
 
-  /** Method to call when trying to render an item. */
-
+  /** Custom render function for each breadcrumb item. */
   onRenderItem?: IRenderFunction<IBreadcrumbItem>;
 
   /**
-   * Method to call when reducing the length of the breadcrumb.
-   * Return undefined to never reduce breadcrumb length
+   * Method that determines how to reduce the length of the breadcrumb.
+   * Return undefined to never reduce breadcrumb length.
    */
   onReduceData?: (data: IBreadcrumbData) => IBreadcrumbData | undefined;
 
   /**
-   * Aria label to place on the navigation landmark for breadcrumb
+   * Aria label for the root element of the breadcrumb (which is a navigation landmark).
    */
   ariaLabel?: string;
 
   /**
-   * Optional name to use for aria label on overflow button.
+   * Aria label for the overflow button.
    */
   overflowAriaLabel?: string;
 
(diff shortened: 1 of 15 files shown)


packages/office-ui-fabric-react/src/components/Breadcrumb/Breadcrumb.types.ts
```diff
@@ -40,7 +40,7 @@ export interface IBreadcrumbProps extends React.HTMLAttributes<HTMLElement> {
   items: IBreadcrumbItem[];
 
   /**
-   * Optional root classname for the root breadcrumb element.
+   * Optional class for the root breadcrumb element.
```
ecraig12345: Most of the changes here are just comment updates. See the next PR comment for the one actual change.

packages/office-ui-fabric-react/src/components/Breadcrumb/Breadcrumb.types.ts
```diff
@@ -99,44 +98,46 @@ export interface IBreadcrumbProps extends React.HTMLAttributes<HTMLElement> {
...
   isCurrentItem?: boolean;
 
   /**
-   * Optional prop to render item as a heading of your choice.
+   * Optional prop to render the item as a heading of your choice.
+   * You can also use this to force items to render as links instead of buttons (by default,
+   * any item with a `href` renders as a link, and any item without a href renders as a button).
    */
-  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'a';
```
ecraig12345: The only actual change in this file

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.types.ts
```diff
@@ -45,32 +44,31 @@ export interface IContextualMenuProps extends IBaseProps<IContextualMenu>, IWith
   styles?: IStyleFunctionOrObject<IContextualMenuStyleProps, IContextualMenuStyles>;
 
   /**
-   * Theme provided by High-Order Component.
+   * Theme provided by higher-order component.
```
ecraig12345: No actual changes to the interfaces in this file, just comments (I did a good bit of reading the comments while trying to figure out how to fix the styling issue, and decided to edit them)

ecraig12345:
Marking as do not merge while I work on adding tests.

PR #10741, opened 2019-10-08
Update missing sass slots and theme designer semantic slot mappings

marygans:
#### Pull request checklist

#### Description of changes

Updates to missing sass slots and theme designer semantic slot mappings. There were semantic slots added as well as some modifications previously (updates to `theme.ts` / `variants.ts` ). Whenever there are modifications there, `SemanticSlots.tsx` + Theme Designer should be updated. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10741)


apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -44,6 +44,7 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
     'bodySubtext', // neutralSecondary
```
aneeshack4: Looks like some other slots are missing that have been added in theme.ts, perhaps in previous PRs and we missed adding them in this mapping too:
`bodyBackgroundHovered` `bodyBackgroundChecked`
Might help to go through the theme.ts mapping & this one side by side to make sure they're matching fully.

aneeshack4:
`office-ui-fabric-react: [7:10:19 PM] x The public API file is out of date. Please run the API snapshot and commit the updated API file.` @marygans Try running `yarn update-snapshots` to fix this.

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -82,12 +85,16 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
     'accentButtonBackground', // accent
     'accentButtonText', // white
     // INPUTS - 14
```
aneeshack4: Probably want to change the #s in these comments to reflect the new counts of each section & the size of each group starting on [line 33 in SemanticSlotsDetailsList.tsx](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/apps/theming-designer/src/components/SemanticSlotsDetailsList.tsx#L33) will have to be changed too so that the DetailsList can resize

marygans: Thanks @aneeshack4, good call! I just updated it

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -96,6 +103,7 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
     'disabledBackground', // neutralLighter
     'disabledText', // neutralTertiary
     'disabledSubText', // neutralQuaternary
+    'disabledBodySubtext', // neutralTertiaryAlt
```
aneeshack4: disabledBodySubtext belongs under Defaults group I believe - make sure to bump # to 17 too

aneeshack4:
Thank you @marygans for doing all this work! Looks good save for that one slot that needs to be moved from inputs to defaults - I'll get Sirena the designer to take a pass over this too.

aneeshack4:
Also just noticed that some of the newly added slots are missing string names: 
image
this is kinda strange...

marygans:
> Also just noticed that some of the newly added slots are missing string names:
> image
> this is kinda strange...

Yes Sirena and I just saw that a second ago too! I think I need to add bodyBackgorundHovered/Checked to `variants`

aneeshack4:
@marygans Any update on the slots with missing string names? Also #10888 is the PR I sent out last week that will remove the need to update slots in two places like this in the future

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -34,8 +34,10 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
   let strongSlots: JSX.Element[] = [];
 
   slotNames = [
-    // DEFAULTS - 13
+    // DEFAULTS - 16
     'bodyBackground', // white
+    'bodyBackgroundHovered', // neutralLighter
```
aneeshack4: Yeah it looks like these might need to be added to variants.ts

marygans:
> @marygans Any update on the slots with missing string names? Also #10888 is the PR I sent out last week that will remove the need to update slots in two places like this in the future

Hey sorry got caught up in some other work but just pushed an update!

khmakoto:
Pinging @marygans and @aneeshack4, is there anything else that needs to be done here?

aneeshack4:
@marygans & I talked offline about closing this & her doing another update after #10888 is merged.

PR #10752, opened 2019-10-08
Announced: default aria-live to polite

### LINKED ISSUE #10742: Announced component's aria-live attribute should default to "polite"
**Reference:** https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

When an aria-live region is set to "assertive", any message added to it will interrupt the user's context and read whatever is set. This is typically reserved for urgent messages, not something like "email deleted" as described in the Announced documentation.

Instead of defaulting to a very heavy-handed value, we should instead default to "polite" and allow consumers to override the value using props if they require their message to interrupt user flow (as  described in #843).

natalieethell:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #10742

#### Description of changes

Defaults Announced to aria-live="polite". @ecraig12345 @betrue-final-final @cschlechty and I discussed this and believe that it makes more sense.

Out of our current examples, only quick actions, like "item deleted" and "item moved", should be assertive messages, which is already specified in the example.

#### Focus areas to test

Ensure that all of the Announced examples still work as expected in Narrator + Edge.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10752)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-08-16-21-44-defaultAnnouncedToPolite.json
+++ b/change/office-ui-fabric-react-2019-10-08-16-21-44-defaultAnnouncedToPolite.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "Announced: default aria live to polite",
+  "packageName": "office-ui-fabric-react",
+  "email": "naethell@microsoft.com",
+  "commit": "10d61e3f27ff20c0da970f53dddae52f7dfa9b72",
+  "date": "2019-10-08T23:21:44.223Z",
+  "file": "C:\\Users\\naethell\\office-ui-fabric-react\\change\\office-ui-fabric-react-2019-10-08-16-21-44-defaultAnnouncedToPolite.json"
+}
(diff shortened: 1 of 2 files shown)


ecraig12345:
Thanks for doing this! It might be worth explicitly adding the `aria-live` prop to `IAnnouncedProps`  (for documentation purposes--I know it's already inherited from div props) 
 and noting the default value.

Also, if you could update the examples' aria-live setting per Ben's suggestions, that would be good. (assertive for quick actions, and when bulk operation is finished)

dzearing:
All of @ecraig12345 's comments are valid; it would be great to add in a separate PR.

natalieethell:
@ecraig12345 no problem - thanks for the help and feedback! I can open another PR to add the aria-live prop explicitly. As for the examples, aria-live is actually already set on the quick actions example, but I can update the bulk operations example.

PR #10755, opened 2019-10-09
Dropdown: do not select first option automatically unless it's selected using keyboard navigation

### LINKED ISSUE #5473: Dropdown with placeholder - first option automatically selected on focus.
### Bug Report 
 
- __Package version(s)__:  > 6.25.1
- __Browser and OS versions__: Chrome 67.0.3396.99, OSX 10.12.6
 
Requested priority: Normal
 
Products/sites affected: Yammer
 
#### Describe the issue:
 
Starting in 6.25.1, the behavior of a Dropdown with a placeholder has changed. Previously clicking to open it would not change the value, now it does.
 
#### Actual behavior:
 
The first value is selected.
 
#### Expected behavior:
 
The placeholder continues to show.
 
https://codepen.io/chrismohr/pen/BVgXYv

### LINKED ISSUE #10713: Dropdown calls onChange before callout is dismissed
### Environment Information

- **Package version(s)**: 7.43.0
- **Browser and OS versions**: Chrome 77.0.3865.90,  Windows 10

https://codepen.io/aboktor/pen/LYYPdwO
#### Actual behavior:
When a dropdown with no defaultSelectedKey is clicked, onChange is called with the first item in options BEFORE the callout is dismissed
#### Expected behavior:
onChange is not called until the user actually selects an option or focus blur. Only at that time should onChange be called with the option the user selected or the first option (in case of blur).

Yes
High
Can't disclose this right now, not released publicly

xugao:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #5473 , Fixes #10713

#### Description of changes
In case of keyboard navigation to dropdown, per aria guidance, dropdown should select first option by default on focus. But we do not want to do this if user clicked and opened the dropdown.
Before:
image 
After:
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10755)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
@@ -71,9 +71,10 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
 
   /** Flag for when we get the first mouseMove */
   private _gotMouseMove: boolean;
-
   /** Flag for identifiying dropdown is opened by getting focus using keyboard */
   private _isOpenedByKeyboardFocus: boolean;
+  /** Flag for tracking whether focus is triggered by click event */
+  private _isFocusedByClick: boolean;
 
   constructor(props: IDropdownProps) {
     super(props);
@@ -279,6 +280,7 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
               onKeyDown={this._onDropdownKeyDown}
               onKeyUp={this._onDropdownKeyUp}
               onClick={this._onDropdownClick}
+              onMouseDown={this._onDropdownMouseDown}
               onFocus={this._onFocus}
             >
               <span
@@ -1045,6 +1047,10 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
     this._isOpenedByKeyboardFocus = false;
   };
 
+  private _onDropdownMouseDown = (_ev: React.MouseEvent<HTMLDivElement>): void => {
+    this._isFocusedByClick = true;
+  };
+
   private _onFocus = (ev: React.FocusEvent<HTMLDivElement>): void => {
     const { isOpen, selectedIndices, hasFocus } = this.state;
     const { multiSelect, openOnKeyboardFocus } = this.props;
@@ -1052,7 +1058,7 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
     const disabled = this._isDisabled();
 
     if (!disabled) {
-      if (!isOpen && selectedIndices.length === 0 && !multiSelect) {
+      if (!this._isFocusedByClick && !isOpen && selectedIndices.length === 0 && !multiSelect) {
         // Per aria
         this._moveIndex(ev, 1, 0, -1);
       }
@@ -1067,6 +1073,8 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
 
       this.setState(state);
     }
+
+    this._isFocusedByClick = false;
   };
 
   /**
--- a/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
@@ -70,10 +70,8 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
   private _requestAnimationFrame = safeRequestAnimationFrame(this);
   /** Flag for when we get the first mouseMove */
   private _gotMouseMove: boolean;
-  /** Flag for identifiying dropdown is opened by getting focus using keyboard */
-  private _isOpenedByKeyboardFocus: boolean;
-  /** Flag for tracking whether focus is triggered by click event */
-  private _isMouseDownBeforeFocus: boolean;
+  /** Flag for tracking whether focus is triggered by click (alternatively triggered by keyboard nav) */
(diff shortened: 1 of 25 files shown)


dzearing:
Looks like screener caught a regression; take a look at the failed screener result and check the problem state. :)

xugao:
> Looks like screener caught a regression; take a look at the failed screener result and check the problem state. :)

The changes look expected. according to the wiki, i need admin to approve these screener test failures. Could you help double check the changes and approve if look good? :)

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -1046,15 +1048,19 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
     }
   };
 
+  private _onDropdownMouseDown = (_ev: React.MouseEvent<HTMLDivElement>): void => {
```
khmakoto: nit:

```suggestion
  private _onDropdownMouseDown = (ev: React.MouseEvent<HTMLDivElement>): void => {
```

xugao: simply removed unused param

khmakoto: For context, @xugao and I talked about this offline and decided that, because this is a private method, we have the liberty of removing the parameter now and adding it later if needed without affecting public facing API.

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -1036,33 +1035,36 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
     const { isOpen } = this.state;
     const disabled = this._isDisabled();
 
-    if (!disabled && !(this._isOpenedByKeyboardFocus && isOpen)) {
+    if (!disabled && !this._shouldOpenOnFocus()) {
```
joschect: I would add a comment here about why `_shouldOpenOnFocus` is called. It might be confusing as to why that's called in `_onDropdownClick`

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -1106,4 +1108,14 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
...
   };
+
+  /**
+   * Returns true if dropdown should set to open on focus.
+   * Otherwise, isOpen state should be toggled on click
+   */
+  private _shouldOpenOnFocus(): boolean {
+    const { hasFocus } = this.state;
+    const { openOnKeyboardFocus } = this.props;
+    return !this._isFocusedByClick && openOnKeyboardFocus === true && !hasFocus;
```
khmakoto: Is there a reason we cannot do?

```suggestion
    return !this._isFocusedByClick && openOnKeyboardFocus && !hasFocus;
```

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -1106,4 +1108,14 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
       </Label>
     ) : null;
   };
+
+  /**
+   * Returns true if dropdown should set to open on focus.
+   * Otherwise, isOpen state should be toggled on click
```
khmakoto: Suggestion:

```ts
  /**
   * Determines if the Dropdown should be open on focus or not.
   */
```

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
khmakoto:
```suggestion
    // `shouldOpenOnFocus` is used here because we don't want to toggle the `isOpen` state if we're supposed to open the Dropdown on focus.
    if (!disabled && !this._shouldOpenOnFocus()) {
```

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
xugao: i missed your comment and the PR was auto-merged. sorry about that
To answer you. reason is `openOnKeyboardFocus` is type `boolean | undefined`
```
true && undefined // undefined
true && undefined === true // false
```

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
xugao: I specifically want to callout "Otherwise, isOpen state should be toggled on click" because the method is used in onClick handler, which can be quite confusing

xugao: Ah. i see your suggestion on moving the other comment inlined where the method is used. i like you suggestion better than my version. i will update these in my future PR :)

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
xugao: i will update these in my future PR :)

mbest:
What is the "aria guidance" referenced above? It seems quite strange that tabbing through an item should change it.

PR #10797, opened 2019-10-11
Perf Test: Consume new version of flamegrill with metrics

JasonGore:
#### Pull request checklist

#### Description of changes

Consume new version of flamegrill adding metrics to flamegraphs:

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10797)

### DIFF
--- a/apps/perf-test/tasks/perf-test.js
+++ b/apps/perf-test/tasks/perf-test.js
@@ -9,30 +9,87 @@ const { argv } = require('@uifabric/build').just;
 // Flamegrill Types
 /**
  * @typedef {{
+ *   scenario: string;
+ *   baseline?: string;
+ * }} Scenario
+ *
+ * @typedef {{
+ *   [scenarioName: string]: Scenario;
+ * }} Scenarios
+ *
+ * @typedef {{
  *   outDir?: string;
  *   tempDir?: string;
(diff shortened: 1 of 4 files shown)


apps/perf-test/tasks/perf-test.js
```diff
@@ -160,43 +217,40 @@ module.exports = async function getPerfRegressions() {
...
+    };
   });
+  // });
 
   console.log(`\nRunning ${iterations} iterations for each of these scenarios: ${scenarioList}\n`);
 
-  if (!fs.existsSync(tempDir)) {
-    console.log(`Making temp directory ${tempDir}...`);
-    fs.mkdirSync(tempDir);
-  }
```
khmakoto: Why is it that we don't need to create a `tempDir` anymore if it does not exist?

JasonGore: https://github.com/microsoft/flamegrill/pull/14 😄

khmakoto: Ahh yes, I remember now!

PR #10798, opened 2019-10-11
Fabric Component: add applyTheme and applyThemeToBody props.

aneeshack4:
#### Pull request checklist

#### Description of changes

Right now, there's no way to automatically apply the loaded theme to the div it's used in or the body of the whole page. Adding `applyTheme` and `applyThemeToBody` props to allow users to do this easily. This need arose when we saw that upon clicking the Export theme button on our [theme designer site](aka.ms/themedesigner), the Codepen doesn't have the theme fully applied to its background.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10798)


packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -29,7 +29,26 @@ export class FabricBase extends React.Component<
...
     const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
 
-    return <div {...divProps} className={classNames.root} ref={this._rootElement} />;
+    let fabricComponent;
+    if (this.props.applyTheme) {
+      // apply theme only to the div
+      fabricComponent = (
+        <div
+          {...divProps}
+          style={{ backgroundColor: this.props.theme!.semanticColors.bodyBackground }}
```
dzearing: please can we do this inside of the .styles file? Also, you're not applying the text color.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -29,7 +29,26 @@ export class FabricBase extends React.Component<
...
+          className={classNames.root}
+          ref={this._rootElement}
+        />
+      );
+    } else if (this.props.applyThemeToBody) {
+      // apply theme to the body of the whole page
+      document.body.style.backgroundColor = this.props.theme!.semanticColors.bodyBackground;
```
dzearing: no! Please don't. This is a permanent update. What you could do is create a "bodyTheme" class, and then conditionally add/remove it to body.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -29,7 +29,26 @@ export class FabricBase extends React.Component<
...
+      fabricComponent = <div {...divProps} className={classNames.root} ref={this._rootElement} />;
+    } else {
+      // do not apply theme
+      fabricComponent = <div {...divProps} className={classNames.root} ref={this._rootElement} />;
```
dzearing: this is repeat code. Please avoid.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -29,7 +29,26 @@ export class FabricBase extends React.Component<
...
+    }
+    return fabricComponent;
   }
 
   public componentDidMount(): void {
```
dzearing: Needs unit tests.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: done - did just the conditional add because I think if I also did 
`    else {
      document.body.classList.remove('ms-Fabric-body-themed');
    } ` it might throw errors at not finding that class right (since there's only one case where it would be added)?

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: done.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: removed.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -29,7 +29,16 @@ export class FabricBase extends React.Component<
     });
     const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
 
-    return <div {...divProps} className={classNames.root} ref={this._rootElement} />;
+    let fabricComponent;
+    // apply theme to just the div only if applyTheme is true, otherwise do not apply theme at all
+    fabricComponent = (
+      <div {...divProps} className={this.props.applyTheme ? classNames.rootThemed : classNames.root} ref={this._rootElement} />
```
dzearing: Shouldn't both be applied, rather than either/or?

aneeshack4: `classNames.root` doesn't have the backgroundColor for the theme so if applyTheme isn't true only that's appropriate but if it is, then we're using `classNames.rootThemed` which has  `color: root.textColor,
        backgroundColor: root.backgroundColor,` so that text & background color get the theme colors.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -29,7 +29,16 @@ export class FabricBase extends React.Component<
...
+    );
+    if (this.props.applyThemeToBody) {
+      // apply theme to the body of the whole page
+      document.body.classList.add('ms-Fabric-body-themed');
```
dzearing: Multiple problems here.

1. You should never access `document` or `window` until the component is mounted. Otherwise you break SSR.

2. `document.body` is the host frame, every time. It seems like on mount, you should call `getDocument` with the _`rootElement` prop to get the actual document  the content is hosted inside of.

3. The classname used should be generated from the styling file, not a global classname.

4. Once you've added the classname, you're never cleaning it up. This is tricky. You probably need a refcount solution to have the last add remove the classname.

On unmount this is still applied. Hmmm. Also, shouldn't there be

dzearing: I don't think this will work, because you are adding the global name and not the generated one. Please test that applying to the body actually applies the generated classname to the body.

packages/office-ui-fabric-react/src/components/Fabric/__snapshots__/Fabric.test.tsx.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Fabric renders correctly 1`] = `
+exports[`Fabric renders a Fabric component correctly 1`] = `
```
dzearing: I don't see any  tests that actually updated any snapshots...

packages/office-ui-fabric-react/src/components/Fabric/Fabric.test.tsx
```diff
@@ -1,11 +1,22 @@
...
-  it('renders correctly', () => {
+  it('renders a Fabric component correctly', () => {
     const component = renderer.create(<Fabric>test</Fabric>);
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  it('renders a Fabric component with applyTheme correctly', () => {
+    const renderedDOM: HTMLElement = renderIntoDocument(<Fabric applyTheme>test</Fabric>);
+    expect(renderedDOM.getElementsByClassName('ms-Fabric root-themed')).toBeTruthy();
```
dzearing: use a snapshot rather than asserts.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.test.tsx
```diff
@@ -1,11 +1,22 @@
...
+  });
+
+  it('renders a Fabric component with applyThemeToBody correctly', () => {
+    const renderedDOM: HTMLElement = renderIntoDocument(<Fabric applyThemeToBody>test</Fabric>);
+    expect(renderedDOM.getElementsByClassName('ms-Fabric-body-themed')).toBeTruthy();
```
dzearing: snapshot

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
dzearing: This isn't really expected. In current patterns where style props control the styles, you'd pass in rootThemed as a style prop and conditionalize those sections of the styling.

In the world where we don't key styling off props, you'd apply both `root` and `rootThemed` classnames to the root. A developer might be applying styling to the `root` which you just dropped.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: Changed it so that now it's adding the generated classname `bodyThemed` - trying to add a snapshot test to confirm this (I see that it does it successfully when I wrap one of the component examples with this and inspect the dom but want to of course add a test too) however I get a jest error saying `The token provided contains HTML space characters, which are not valid in tokens.` which I think might be due to the removing on unmount - looking into this.

scripts/postinstall.js
```diff
@@ -4,9 +4,7 @@ const chalk = require('chalk');
 // git v2.9.0 supports a custom hooks directory. This means we just need to checkin the hooks scripts
 spawnSync('git', ['config', 'core.hooksPath', '.githooks']);
 
-console.log(`${chalk.green(
-  'All dependencies are installed! This repo no longer builds automatically after installing dependencies.'
-)}
+console.log(`${chalk.green('All dependencies are installed! This repo no longer builds automatically after installing dependencies.')}
```
aneeshack4: No idea why this change happened... doesn't make any sense

ecraig12345: Probably because prettier was run while merging the commit from master that contained this change. (But it also should have run when this was committed in the first place, so not sure why it would be formatted differently this time.)

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: Moved the conditionalizing into the styles file.

aneeshack4:
Pinging @dzearing on this - any more feedback? Would like to close this soon since it's been open for a while

packages/office-ui-fabric-react/src/components/Fabric/Fabric.test.tsx
```diff
@@ -3,9 +3,21 @@ import * as renderer from 'react-test-renderer';
...
     expect(tree).toMatchSnapshot();
   });
+
+  it('renders a Fabric component with applyTheme correctly', () => {
+    const component = renderer.create(<Fabric applyTheme>test</Fabric>);
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders a Fabric component with applyThemeToBody correctly', () => {
```
jdhuntington: I don't believe this snapshot is useful. Since the snapshot does not encapsulate the body, it's not actually validating any behavior.

packages/office-ui-fabric-react/src/components/Fabric/__snapshots__/Fabric.test.tsx.snap
```diff
@@ -1,6 +1,58 @@
...
+<div
+  className=
+      ms-Fabric
+      {
+        -moz-osx-font-smoothing: grayscale;
+        -webkit-font-smoothing: antialiased;
+        color: #323130;
+        font-family: 'Segoe UI', 'Segoe UI Web (West European)', 'Segoe UI', -apple-system, BlinkMacSystemFont, 'Roboto', 'Helvetica Neue', sans-serif;
+        font-size: 14px;
+        font-weight: 400;
```
jdhuntington: I would expect `background-color` to appear here. Is that an incorrect assumption?

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -20,16 +21,13 @@ export class FabricBase extends React.Component<
...
-      theme: this.props.theme!,
-      className,
-      isFocusVisible: this.state.isFocusVisible
-    });
     const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
 
-    return <div {...divProps} className={classNames.root} ref={this._rootElement} />;
+    let fabricComponent;
+    fabricComponent = <div {...divProps} className={classNames.root} ref={this._rootElement} />;
+    return fabricComponent;
```
jdhuntington: nit: can be collapsed into `return <div...`

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -38,10 +36,32 @@ export class FabricBase extends React.Component<
     if (win) {
       this._disposables.push(on(win, 'mousedown', this._onMouseDown, true), on(win, 'keydown', this._onKeyDown, true));
     }
+
+    const classNames = this._getClassNamesHelper();
+    if (this.props.applyThemeToBody) {
+      const currentDoc: Document = getDocument(this._rootElement.current)!;
```
jdhuntington: Should avoid `!` and instead check the return value.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -38,10 +36,32 @@ export class FabricBase extends React.Component<
...
+      currentDoc.body.classList.add(classNames.bodyThemed);
+    }
   }
 
   public componentWillUnmount(): void {
     this._disposables.forEach((dispose: () => void) => dispose());
+
```
jdhuntington: There could be some weird edge cases here that are unlikely, but should guard against. If a developer renders this control with `applyThemeToBody=true`, then sets `applyThemeToBody=false`, then unmounts, this code will not have the intended effect.

Instead, what about having `componentDidMount()` define a callback to clean up (much like `_disposables`)? Has the added benefit of evaluating `_getClassNamesHelper()` fewer times.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -38,10 +36,32 @@ export class FabricBase extends React.Component<
...
+    if (this.props.applyThemeToBody) {
+      const currentDoc: Document = getDocument(this._rootElement.current)!;
+      currentDoc.body.classList.remove(classNames.bodyThemed);
+    }
+  }
+
+  private _getClassNamesHelper(): IProcessedStyleSet<IFabricStyles> {
+    const { className } = this.props;
+    const classNames = getClassNames(getStyles, {
+      theme: this.props.theme!,
```
jdhuntington: Should check the return value of `this.props.theme`, not use `!`.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
KevinTCoughlin: You'll probably want to clean-up regardless of props. Imagine the scenario where you render this on a page using the new props to add a class to the `document.body`. Then you remove that component while performing an in-place navigation (not a hard refresh of the document). That class will remain on `document.body` and impact other components until the document is hard-refreshed. I'm thinking of SPFx WebParts in particular.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: done.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: done

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: done.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: I tried to avoid using this ! but the only way to to do an if check for `this.props.theme` is outside getClassNames call (unless I'm forgetting something?) and when I do that, the method needs an alternate return value besides `IProcessedStyleSet<IFabricStyles>` like undefined but adding that to this method signature is introducing unnecessary ambiguity especially because I looked at our usage of ! inside getClassNames calls across the Fabric codebase and didn't find any other cases where it was avoided.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.test.tsx
aneeshack4: Spoke offline a bit - seems like it might be hard to test this sort of change to body. Right now, across all the snapshot tests in the Fabric codebase, there aren't any examples of testing the changes to body like the ones happening here but I'm trying to find a way.

packages/office-ui-fabric-react/src/components/Fabric/__snapshots__/Fabric.test.tsx.snap
aneeshack4: Hmm yes now that I look at these again, they kind of all look identical even though the DOM I can see, when I test locally by doing `applyTheme` and `applyThemeBody` to exising examples that each is changing the DOM differently.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: I'm slightly confused here - so is it recommended to do a `this._disposables.forEach((dispose: () => void) => dispose());` in `componentDidMount()` too when it's already happening in `componentDidUnmount`? And is it better to do it even if `props.applyThemeToBody` is passed in? To be more clear, what kind of an action would cause there to be a component to be removed during an in-place navigation? I'm trying to think of an example - like while scrolling or navigating to the last page, somehow the component this prop is on is removed?

KevinTCoughlin: I was advocating for removing the classes previously applied by this component when the component is unmounted via `componentDidUnmount`. This way the component cleans-up after itself and doesn't leave artifacts like classNames on `document.body`. That can have side-effects when navigating in an app where we may want the behavior on some pages but not others.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.test.tsx
aneeshack4: So the main way to test this sort of change to body seems to be with a hasClass check like this https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/Facepile/Facepile.test.tsx#L175, https://airbnb.io/enzyme/docs/api/ReactWrapper/hasClass.html but that's the same as the assert style that was not recommended in a previous comment by @dzearing

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: Done.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.test.tsx
aneeshack4: I researched a lot about using enzyme, even other solutions like puppeteer & tried a bunch of different things like:
>     const dom = mount(<Fabric applyThemeToBody>test</Fabric>);
>     const outermost = dom.childAt(0); 
>     expect(outermost.hasClass('bodyThemed')).toBe(true);
but ended up landing on just getDOMNode() - still not directly rendering into document & using enzyme so that's better than before.

packages/office-ui-fabric-react/src/components/Fabric/__snapshots__/Fabric.test.tsx.snap
aneeshack4: I believe this is because the applyTheme conditional in Fabric.styles.ts isn't working properly

aneeshack4: Fixed this! It was because the props `applyTheme` and `applyThemeToBody` weren't properly being pulled from getClassNames in the base.

aneeshack4:
Tested this on Label by wrapping those examples with `<Fabric applyTheme>`:
image
and <Fabric applyThemeToBody`:
image

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -13,37 +12,59 @@ export class FabricBase extends React.Component<
...
   }
 
   public componentWillUnmount(): void {
     this._disposables.forEach((dispose: () => void) => dispose());
+    if (this._removeClassNameFromBody) {
+      this._removeClassNameFromBody();
+    }
   }
 
+  private _getClassNamesHelper(): IProcessedStyleSet<IFabricStyles> {
```
ecraig12345: nit, you don't really need "helper" in the method name

packages/office-ui-fabric-react/src/components/Fabric/Fabric.styles.ts
```diff
@@ -4,18 +4,18 @@ import { IFabricStyleProps, IFabricStyles } from './Fabric.types';
 const inheritFont = { fontFamily: 'inherit' };
 
 const GlobalClassNames = {
-  root: 'ms-Fabric'
+  root: 'ms-Fabric',
+  bodyThemed: 'ms-Fabric-body-themed'
```
ecraig12345: Not super familiar with class name conventions but I think this would typically be `ms-Fabric-bodyThemed`?

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -13,37 +12,59 @@ export class FabricBase extends React.Component<
...
+    const { className, theme, applyTheme, applyThemeToBody } = this.props;
+    const classNames = getClassNames(getStyles, {
+      theme: theme!,
+      applyTheme: applyTheme,
+      applyThemeToBody: applyThemeToBody,
```
ecraig12345: I'm not seeing `applyThemeToBody` used in the style function.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.styles.ts
aneeshack4: Yeah I see that's how isFocusVisible is too - camelcasing

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: Oh yes `bodyThemed` which is the corresponding class in the styles file is added conditionally based on `this.props.applyThemeToBody` so we don't need it here in the getClassNames function

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
aneeshack4: changed.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -13,37 +12,58 @@ export class FabricBase extends React.Component<
...
+    const classNames = getClassNames(getStyles, {
+      theme: theme!,
+      applyTheme: applyTheme,
+      className,
+      isFocusVisible: this.state.isFocusVisible
+    });
+    return classNames;
+  }
+
+  private _addClassNameToBody = () => {
```
dzearing: any reason why this is a lambda?

aneeshack4: Since it's using `this` inside a couple times, to give it the right context of `this` and since no arguments are being used.

ecraig12345: Using a lambda to ensure the correct `this` is only necessary if the method will be passed as a callback. If it's only ever called directly (`this._addClassNameToBody()`), it's not necessary to use a lambda, since `this` inside `_addClassNameToBody` will automatically be set to the object the method was called on. (The "as an object method" section of [this article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) explains it better.) And the number of arguments doesn't really make a difference for whether something should be a lambda or not.

Overall, while it's not technically wrong to use a lambda here (everything will still work as expected as far as I'm aware), it does have the slight downside of implying that the method is passed as a callback, when in fact that's not the case.

aneeshack4: Thanks for this clarification elizabeth! I got rid of the lambda.

PR #10806, opened 2019-10-14
Pivot, Stack, Card: relax child item type checks

### LINKED ISSUE #10785: Spurious error: The children of a Pivot component must be of type PivotItem to be rendered.
This issue was reported previously here: https://github.com/OfficeDev/office-ui-fabric-react/issues/8867
It was closed due to inactivity. I have an isolated repro of the problem.

### Environment Information

- **Package version(s)**: "office-ui-fabric-react": "^7.48.0"
- **Browser and OS versions**: Windows 10, Chrome 77.0.3865.90 64-bit

I could not repro in a code-pen but I did put together a small repo that has the bug:
https://github.com/akoltz/ouifr-pivot-bug

In this case, I am able to repro the issue when I am also using a third party library, 'react-timeseries-charts' side-by-side with the Pivot. But this just happens to be the lib that triggered the issue for me... there may be others and it may not be due to bad behavior in that lib. I'm using dozens of other react office-ui-fabric-react components that never had a problem.

The repro does not require rendering any of the components in react-timeseries-charts. Simply importing it causes the Pivot to break.

#### Actual behavior:
Pivot does not render and throws a "The children of a Pivot component must be of type PivotItem to be rendered." error in the console.

#### Expected behavior:
Pivot item renders

Requested priority: Low

Products/sites affected: (if applicable) None

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10785

#### Description of changes

Under certain circumstances which still are not clear, the checks used to determine whether children of `Pivot` (and `Stack` and `Card`) are of a certain type would fail due to component function identity differences. So switch the checks to string comparison on the function name instead.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10806)


ecraig12345:
> I think `Card` also has a similar issue, could you change that as well as part of this PR?

Fixed!

packages/react-cards/src/components/Card/Card.view.tsx
```diff
@@ -95,3 +97,18 @@ export const CardView: ICardComponent['view'] = props => {
...
 };
+
+function _isReactElement(item: React.ReactNode): item is React.ReactElement {
+  return !!item && typeof item === 'object' && !!(item as React.ReactElement).type;
+}
+
+function _isCardItem(item: React.ReactNode): item is typeof CardItem {
+  // In theory, we should be able to just check item.type === CardItemType.
+  // However, under certain unclear circumstances (see https://github.com/OfficeDev/office-ui-fabric-react/issues/10785),
+  // the object identity is different despite the function body being the same.
```
khmakoto: Just a question, should we also have this comment in the `_isCardSection` method? I know it would be repeating the same info but I've seen this before in other files. I don't have a strong opinion on this either, just wanted to bring it up.

ecraig12345: I decided to omit it since it's only a few lines down from the same comment in _isCardItem

MichielDeMey:
Any particular reason we're checking `displayName` in `Stack` and `Card` but just `name` in `Pivot`? 🙂

ecraig12345:
At the time this change was made, StackItem and CardItem were function components (which may not have a `name`) whereas PivotItem was a class component (which automatically has `name`). Is this causing a problem or were you just curious?

MichielDeMey:
@ecraig12345 Sorry for the late reply. 😅

I was trying to create a custom `SuspensefulPivotItem` which would just be a `<PivotItem>` with a `<Suspense>` inside it to prevent content flashes outside a pivot item while loading suspenseful content.

```tsx
<PivotItem>
  <Suspense fallback={<Spinner />}>
    {children}
  </Suspense>
</PivotItem>
```

So I tried to create a custom Class component that would `extend PivotItem` but I could not get around the Pivot sanity check.
```
The children of a Pivot component must be of type PivotItem to be rendered.
```

I checked and I believe it was caused by the minification from Webpack (using create-react-app) that would cause the `name` property to be random, failing the check.

I then tried to create a functional component and I though I'd simply set the `displayName` manually since that won't get minified. But then I saw that the code was not checking on the `displayName` property. Hence my question. 🙂 

```tsx
const SuspensefulPivotItem = (props) => { ... }

SuspensefulPivotItem.displayName = 'PivotItem'
```

I didn't get around to finding a solution for creating a custom `PivotItem` due to the sanity check though, so any pointers are welcome!

PR #10807, opened 2019-10-14
CommandBar: update examples and docs

### LINKED ISSUE #10737: CommandBar button doesn't work in overflow mode
office-ui-fabric-react@7.45.0 on Firefox 69.0.2

CommandBar button works when CommandBar is not shrunk:

image

But doesn't work when CommandBar in overflow mode:

image

I have a custom renderer for the button provided through `commandBarButtonAs`:
```ts
    {
      key: 'editEntity',
      disabled: !entities || !entities.size,
      text: 'Edit',
      iconProps: {
        iconName: 'Edit'
      },
      commandBarButtonAs: entities && entities.size ? linkCommandBarButton(`entityDlg/${entities.getString('name')}`) : undefined
    },
```
Where it is rendered as `react-router's` `Link` component:
```ts
export interface ILinkCommandBarButtonProps extends IComponentAsProps<ICommandBarItemProps> {
  link: string;
  supText?: string;
  onClick?: () => void;
};

export class LinkCommandBarButton extends BaseComponent<ILinkCommandBarButtonProps> {
  public render(): JSX.Element {
    const { defaultRender, link, supText, onClick, ...buttonProps } = this.props;

    const onRenderText = supText ? (props: IButtonProps) => <>{props.text}<sup>{supText}</sup></> : undefined;
    const DefaultRender = defaultRender ? defaultRender as any : CommandBarButton;

    const render = (
      <DefaultRender {...buttonProps} onRenderText={onRenderText}/>
    );
    if (buttonProps.disabled) {
      return render;
    } else {
      return (
        <Link to={link} onClick={onClick}>
          {render}
        </Link>
      );
    }
  }
};

export const linkCommandBarButton = (link: string, supText?: string, onClick?: () => void) => (props: IComponentAsProps<ICommandBarItemProps>) => {
  return <LinkCommandBarButton {...props} link={link} supText={supText} onClick={onClick} />;
};
```
So, it is seem as custom renderer is somehow lost when `CommandBar` is in overflow mode.

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10737

#### Description of changes

Update CommandBar.ButtonAs.Example.tsx to handle the scenario highlighted by #10737, where the component user also wants to customize overflow/menu item rendering.

Update CommandBar examples to use function components and other modern patterns (and simplify some of the examples).

Edits to CommandBar prop docs.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10807)


packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.types.ts
```diff
@@ -32,44 +32,45 @@ export interface ICommandBarProps extends React.HTMLAttributes<HTMLDivElement> {
   componentRef?: IRefObject<ICommandBar>;
 
   /**
-   * Items to render. ICommandBarItemProps extend IContextualMenuItem
+   * Items to render. ICommandBarItemProps extends IContextualMenuItem.
```
ecraig12345: Comment changes only, aside from the one prop that was `Boolean` for some reason instead of `boolean`

PR #10811, opened 2019-10-14
CommandBar: Fixing application of user provided styles to items

### LINKED ISSUE #10774: CommandBar item with buttonStyles applied loses some of the styling
### Environment Information

- **Package version(s)**: office-ui-fabric-react 7.47.0
- **Browser and OS versions**: Chrome 77.0.3865.90, Windows 10 Pro

https://codepen.io/creimann/pen/VwwLgRR

#### Actual behavior:

When the background color of a CommandBar item with iconOnly set to true is changed then the icon is aligned top and not vertically centered.

#### Expected behavior:

The icon should be vertically centered when iconOnly is set to true.

Requested priority: Normal

Products/sites affected: -

### LINKED ISSUE #10776: Callout: Incorrect position when target is off-screen (doNotLayer)
### Environment Information

- **Package version(s)**: 7.39.1

https://codepen.io/Zhephyr54/pen/ZEEbGWO
(Scroll down to the "Show callout" button)

#### Actual behavior:

Using a Callout with doNotLayer, preventDismissOnScroll and preventDismissOnLostFocus (for a fully controlled component), when the Callout is actually displayed while the target element is off-screen (not visible at this moment), the Callout positioning is incorrect (it appears to be the closest to the target element while being still visible on screen).

#### Expected behavior:

The Callout positions itself properly in relation to the target element even when the target element is not visible on screen.

Requested priority: Normal

### Remarks

I am not sure if this behavior is actually supposed to be supported or if it would be hard to implement.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10774

#### Description of changes

Inline styles are being provided in `CommandBar.base.tsx` to `CommandBarItems`:

```ts
styles: { root: { height: '100%' }, label: { whiteSpace: 'nowrap' }, ...item.buttonStyles }
```

However, this causes a problem if the user specified `buttonStyles` since it overrides these inline styles. For example, if you used `buttonStyles` you could end up with something like:

image

This PR fixes this by applying the `root` and `label` user defined styles separately on top of the ones already present:

```ts
styles: {
   ...item.buttonStyles,
   root: { height: '100%', ...(item.buttonStyles ? (item.buttonStyles.root as object) : {}) },
   label: { whiteSpace: 'nowrap', ...(item.buttonStyles ? (item.buttonStyles.label as object) : {}) }
}
```

Which solves the issue above while respecting user defined styles:

image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10811)


packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.base.tsx
```diff
@@ -138,8 +138,8 @@ export class CommandBarBase extends BaseComponent<ICommandBarProps, {}> implemen
       allowDisabledFocus: true,
       role: 'menuitem',
       ...item,
-      styles: { root: { height: '100%' }, label: { whiteSpace: 'nowrap' }, ...item.buttonStyles },
-      className: css('ms-CommandBarItem-link', item.className),
+      styles: { label: { whiteSpace: 'nowrap' }, ...item.buttonStyles },
+      className: css(this._classNames.commandBarItems, item.className),
```
micahgodbolt: what about specificity? these will both be the same specificity. if user provides a height, can we ommit the className?

khmakoto: @micahgodbolt, you're right, thanks for that, I didn't really catch it! I'm wondering if there is a cleaner way to do this. I'll investigate.

joschect:
@khmakoto is this a case where we should use merge styles? Commandbar items should have their own getStyles function?

khmakoto:
@joschect that is a tricky thing, I don't know if we necessarily want to add a `getStyles` function here, though I agree that we might want to use `mergeStyles` to do this instead of manually. I'll delegate that decision to @micahgodbolt however, but for what it's worth the current approach works very similar to `mergeStyles` so using one or the other shouldn't affect this functionality-wise.

PR #10812, opened 2019-10-15
Announced: add `as` prop and root styles

### LINKED ISSUE #10743: Announced component should be a <span> not a <div>
#### Actual behavior:

`Announced` component returns `<div>` but has nothing inside that requires it to be a block element, this causing a lot of nesting issues when teams use `Announced` inside `<span>` and similar inline elements.

#### Expected behavior:

Announced component returns `<span>`

Are you willing to submit a PR to fix? No (I could and would but my machine isn't currently setup)

Requested priority: Normal

Products/sites affected: Outlook Web

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10743

#### Description of changes

Currently Announced is always rendered as a `div`, and there's not an idiomatic way to make it render as an `inline-block` element. This PR adds two ways to do that:
1. `styles.root` for styling the root element
2. an `as` prop so the root element can be rendered as anything (such as a `span`)

I definitely think adding `styles.root` makes sense (matches most of our other components), but I'm less certain about adding `as`. Either approach could solve the issue which prompted this PR.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10812)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-14-17-36-33-announced.json
+++ b/change/office-ui-fabric-react-2019-10-14-17-36-33-announced.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "Announced: add `as` prop and root styles",
+  "packageName": "office-ui-fabric-react",
+  "email": "elcraig@microsoft.com",
+  "commit": "2e4ea58e204193827ff747b0c661e1ac0701577f",
+  "date": "2019-10-15T00:36:33.041Z",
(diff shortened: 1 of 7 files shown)


ecraig12345:
In a shocking turn of events, adding a `root` style and not applying it anywhere does nothing. 😆 So...that's fixed now.

PR #10813, opened 2019-10-15
SpinButton: allow passing data attributes

### LINKED ISSUE #10768: SpinButton doesn't pass through other HTML attributes such as `data-testid`.
### Environment Information

- **Package version(s)**: 7.15.0
- **Browser and OS versions**: Chrome Version 77.0.3865.90 (Official Build) (64-bit)

https://codesandbox.io/s/spin-button-no-data-testid-attribute-31tzw

#### Actual behavior:

Other components forward the attribute on to their surrounding DOM element to make it reliably accessible by e2e tests

#### Expected behavior:

The attribute is not forwarded.

image

image

Requested priority: (Blocking, High, Normal, **Low**)

Can likely just wrap it another `div` to work around, but it is inconsisten with other components.

Seem to be other issues with `SpinButton` related to label spacing, guessing this component is not used as much as others and doesn't get as much attention. 

On larger note, I making components testable would be helpful. For example, it took some digging but I noticed some controls like the `ChoiceGroup` have built in ways to pass testid's to the choices which was helpful. This ensures our tests will not be coupled to the DOM structure of a particular OFUI version only the test ids.

xugao:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #10768

#### Description of changes
Allow to pass data attributes to `<SpinButton />`
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10813)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
+++ b/packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
@@ -144,7 +144,6 @@ export class SpinButton extends React.Component<ISpinButtonProps, ISpinButtonSta
       incrementButtonAriaLabel,
       decrementButtonIcon,
       decrementButtonAriaLabel,
-      title,
       ariaLabel,
       ariaDescribedBy,
       styles: customStyles,
@@ -167,7 +166,7 @@ export class SpinButton extends React.Component<ISpinButtonProps, ISpinButtonSta
       ? this.props.getClassNames(theme!, disabled, isFocused, keyboardSpinDirection, labelPosition, className)
       : getClassNames(getStyles(theme!, customStyles), disabled, isFocused, keyboardSpinDirection, labelPosition, className);
 
-    const nativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties, ['onBlur', 'onFocus']);
+    const nativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties, ['onBlur', 'onFocus', 'className']);
 
     return (
       <div className={classNames.root}>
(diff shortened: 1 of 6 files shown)


packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
```diff
@@ -165,6 +167,8 @@ export class SpinButton extends React.Component<ISpinButtonProps, ISpinButtonSta
       ? this.props.getClassNames(theme!, disabled, isFocused, keyboardSpinDirection, labelPosition, className)
       : getClassNames(getStyles(theme!, customStyles), disabled, isFocused, keyboardSpinDirection, labelPosition, className);
 
+    const nativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties, ['onBlur', 'onFocus']);
```
ecraig12345: You should probably also exclude `className`, since that's used elsewhere.

packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
```diff
@@ -180,6 +184,7 @@ export class SpinButton extends React.Component<ISpinButtonProps, ISpinButtonSta
         <KeytipData keytipProps={keytipProps} disabled={disabled}>
           {(keytipAttributes: any): JSX.Element => (
             <div
+              {...nativeProps}
               className={classNames.spinButtonWrapper}
               title={title && title}
```
ecraig12345: Since `title` is a native prop, this line can be removed.

xugao: Good catch!

PR #10819, opened 2019-10-15
Fix cases where List virtualization causes stale data to render

ThomasMichon:
The `DetailsList` and `List` have some communication issues which can result in stale data being rendered when the `items` passed to the `List` are updated.
Specifically, if:
1. `DetailsList` is initially rendered with less than 10 items (the default page size)
2. `DetailsList` is subsequently updated to have 30+ items (more than 3 pages)
3. The `renderedWindowsAhead` value is `0`, preventing idle callbacks, or `getItemCountPerPage` has been altered to make the page size the same as the window size.
Then the resulting change can cause only the first page of the content to be rendered, and even worse is that it will have stale item data.

This change addresses these circumstances by eliminating two troublesome behaviors:
1. `DetailsList` no longer calls `forceListUpdates` during `componentWillReceiveProps`, because this caused `List` to render with its previous props, queuing a state update callback with bad data. Now, `DetailsList` simply uses a `version` object which can be revved when `shouldForceListUpdates` is `true`, ensuring that the `List` will always re-render subsequently.
2. `List` no longer captures stale data in the post-update callback within `_updatePages`. What was happening was that the callback contained the original `props` and `state` from the first call, but multiple `setState` calls would be queued, so by the time the callback was fired it would be out-of-date. Now, the callback retrieves the current `state` and `props` and takes its action based on current data.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10819)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-15-12-39-20-list-virtualization.json
+++ b/change/office-ui-fabric-react-2019-10-15-12-39-20-list-virtualization.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "Fix virtualization state tracking in List and DetailsList",
+  "packageName": "office-ui-fabric-react",
+  "email": "tmichon@microsoft.com",
+  "commit": "09a835928992fa37838f347cec2d16f97d4fe037",
+  "date": "2019-10-15T19:39:20.931Z",
+  "file": "D:\\Src\\Git\\OfficeDev\\office-ui-fabric-react-1\\change\\office-ui-fabric-react-2019-10-15-12-39-20-list-virtualization.json"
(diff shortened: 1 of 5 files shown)


KevinTCoughlin:
@ThomasMichon it looks like adjusting column widths via the DetailsHeader no longer works:

http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/10819/merge/fabric-website-resources/dist/index.html#/examples/detailslist

PR #10848, opened 2019-10-15
Fix getDerivedStateFromProps signatures

ecraig12345:
#### Pull request checklist

#### Description of changes

Per React typings, the signature of `getDerivedStateFromProps` should be:

```
static getDerivedStateFromProps(
  nextProps: Readonly<IProps>,
  prevState: Readonly<IState>
): Partial<IState> | null
```

Not all of the components followed this signature, and the missing readonly part helped allow me to almost introduce a bug in checkbox in #10644. So I made this PR to fix the signature everywhere.

I also changed places which were returning the entire state to return `null` (no update) instead, and got rid of places which were copying the old state into the returned object (which is unnecessary because React merges in the update automatically, similar to a setState call).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10848)

### DIFF
--- a/change/@uifabric-charting-2019-10-15-15-20-25-derivedstate.json
+++ b/change/@uifabric-charting-2019-10-15-15-20-25-derivedstate.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Fix getDerivedStateFromProps signatures",
+  "packageName": "@uifabric/charting",
+  "email": "elcraig@microsoft.com",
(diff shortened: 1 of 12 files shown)


ecraig12345:
All the modified controls work fine as I can tell. Reviewers please take a look and verify.
http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/10848/merge/

PR #10867, opened 2019-10-16
Add role to FocusZone in BasePickerListBelow

mattlind:
#### Pull request checklist

#### Description of changes
Added role attribute to the FocusZone that contains the selected items for BasePickerListBelow.  Previously there was no role set so adding a role like "list-item" to an item produced by the onRenderItem callback didn't read out in screen readers as item number n of m.

#### Focus areas to test
- Tested that screen reader now read item "n" of "m" in the sample site: https://developer.microsoft.com/en-us/fabric#/controls/web/peoplepicker

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10867)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-16-08-58-09-BasePickerListBelowRole.json
+++ b/change/office-ui-fabric-react-2019-10-16-08-58-09-BasePickerListBelowRole.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": " Added role to the selection list for BasePickerListBelow control so screen readers will read off \"n\" of \"m\" when arrowing through the list.",
+  "packageName": "office-ui-fabric-react",
+  "email": "malind@microsoft.com",
+  "commit": "adfbe74c57a96e8b75b16c0ee11397ab06a3fd5f",
+  "date": "2019-10-16T15:58:09.775Z"
+}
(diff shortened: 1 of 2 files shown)


joschect:
@mattlind Thanks for submitting this fix! It looks like it would break aria rules for users who don't have role="listitem" already on their selected items. I think that either adding `selectedItemsContainerProps: IFocusZoneProps' or `renderSelectedItemsContainer` to props might be another way of doing this.

mattlind:
> @mattlind Thanks for submitting this fix! It looks like it would break aria rules for users who don't have role="listitem" already on their selected items. I think that either adding `selectedItemsContainerProps: IFocusZoneProps' or `renderSelectedItemsContainer` to props might be another way of doing this.

I did it this way because that was how the other instance was doing the same thing above.

```
<SelectionZone selection={this.selection} selectionMode={SelectionMode.multiple}>
            <div className={classNames.text}>
              {items.length > 0 && (
                <span id={this._ariaMap.selectedItems} className={classNames.itemsWrapper} **role={'list'}**>
                  {this.renderItems()}
                </span>
              )}
```

joschect:
@mattlind That makes sense then. Approved to bring everything in line.

shivasai09:
is it possible to  cherry pick this change to version 6?

PR #10868, opened 2019-10-16
Fixed onDismiss passed into Button, made more consistent

micahgodbolt:
#### Pull request checklist

Currently a custom onDismiss in Button's menuProps will totally replace the internal one, meaning there's no way to even add a console log without breaking the typical dismiss behavior.

This change makes that behavior more consistent with other experiences where it'll always call onDismiss and internal dismiss as long as you don't use `ev.preventDefault()` on the onDismiss function.

https://codepen.io/micahgodbolt/pen/zYYBaRx

I updated the demo to show how shift key could be used to block dismissal. 

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10868)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
+++ b/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
@@ -448,7 +448,6 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
   };
 
   private _onRenderMenu = (menuProps: IContextualMenuProps): JSX.Element => {
-    const { onDismiss = this._dismissMenu } = menuProps;
     const { persistMenu } = this.props;
     const { menuHidden } = this.state;
     const MenuType = this.props.menuAs || (ContextualMenu as React.ReactType<IContextualMenuProps>);
@@ -470,9 +469,20 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
         hidden={persistMenu ? menuHidden : undefined}
         className={css('ms-BaseButton-menuhost', menuProps.className)}
         target={this._isSplitButton ? this._splitButtonContainer.current : this._buttonElement.current}
-        onDismiss={onDismiss}
+        onDismiss={this._onDismissMenu}
       />
     );
+  };
+
+  private _onDismissMenu = (ev: any): void => {
+    const { menuProps } = this.props;
+
+    if (menuProps && menuProps.onDismiss) {
+      menuProps.onDismiss(ev);
+    }
+    if (!ev.defaultPrevented) {
+      this._dismissMenu();
+    }
   };
 
   private _dismissMenu = (): void => {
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
```diff
@@ -470,11 +469,22 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
         hidden={persistMenu ? menuHidden : undefined}
         className={css('ms-BaseButton-menuhost', menuProps.className)}
         target={this._isSplitButton ? this._splitButtonContainer.current : this._buttonElement.current}
-        onDismiss={onDismiss}
+        onDismiss={this._onDismissMenu}
       />
     );
   };
 
+  private _onDismissMenu = (ev: any): void => {
```
micahgodbolt: yes, this ev is an any in the typings.

cliffkoh: Sounds like we should fix upstream since we own those typings, don't we?

khmakoto: I agree on a general level, we could definitely have the private `_onDismissMenu` take the correct typings, but we wouldn't be able to change the public `onDismissMenu` since it could potentially break customers.

khmakoto: @cliffkoh I think we should let this PR go in if Micah doesn't respond before the but merges it. In that case I'll change the typings in a separate PR because it would be a non-public facing api change.

cliffkoh: I'm ok with that :) henc ei didn't remove automerge label or block. Thanks for volunteering to do the follow up!

cliffkoh: @khmakoto Oh i just saw your first response. We don't consider types tightening to be a breaking change. Improving our types and catching actual bugs is considered a good thing! We have communicated and solicited feedback on this policy last year and none of our partner teams expressed opposition or concerns about this policy.

khmakoto: Sounds good, I'll send a followup PR tomorrow then. Thanks for the context @cliffkoh!

KevinTCoughlin: Agree with the above. Types tightening has not been met with friction in my experience.

micahgodbolt: thanks for letting this go through and doing the typings separately. This PR is unblocking a customer.

PR #10869, opened 2019-10-16
Facepile: presence example

maxwellred:
#### Pull request checklist

- [x] Addresses an existing issue discussed with Angela and Matt Cox about whether or not Facepile should advertise this functionality more.

#### Description of changes

Added presence icons to the Facepile Basic Example. I simply added it to ALL the coins as an illustration that you CAN use presence icons. I didn't know if it would be useful to add fake individual presences to each icon but I can do that if the current implementation seems potentially confusing to consumer.

#### Screenshot
Screen Shot 2019-10-17 at 11 14 43 AM

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10869)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Facepile/examples/Facepile.Basic.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/Facepile/examples/Facepile.Basic.Example.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { Checkbox } from 'office-ui-fabric-react/lib/Checkbox';
 import { Dropdown, IDropdownOption } from 'office-ui-fabric-react/lib/Dropdown';
 import { Facepile, IFacepilePersona, IFacepileProps } from 'office-ui-fabric-react/lib/Facepile';
-import { PersonaSize } from 'office-ui-fabric-react/lib/Persona';
+import { PersonaSize, PersonaPresence } from 'office-ui-fabric-react/lib/Persona';
 import { Slider } from 'office-ui-fabric-react/lib/Slider';
 import { facepilePersonas } from '@uifabric/example-data';
 import './Facepile.Examples.scss';
@@ -11,6 +11,7 @@ export interface IFacepileBasicExampleState {
   numberOfFaces: any;
   imagesFadeIn: boolean;
   personaSize: PersonaSize;
+  personaPresence: PersonaPresence;
 }
 
 export class FacepileBasicExample extends React.Component<{}, IFacepileBasicExampleState> {
@@ -20,7 +21,8 @@ export class FacepileBasicExample extends React.Component<{}, IFacepileBasicExam
     this.state = {
       numberOfFaces: 3,
       imagesFadeIn: true,
-      personaSize: PersonaSize.size32
+      personaSize: PersonaSize.size32,
+      personaPresence: PersonaPresence.none
     };
   }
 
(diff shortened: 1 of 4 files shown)


packages/office-ui-fabric-react/src/components/Facepile/examples/Facepile.Basic.Example.tsx
```diff
@@ -63,6 +66,20 @@ export class FacepileBasicExample extends React.Component<{}, IFacepileBasicExam
             ]}
             onChange={this._onChangePersonaSize}
           />
+          <Dropdown
+            label="Persona Presence:"
+            selectedKey={this.state.personaPresence}
+            options={[
+              { key: PersonaPresence.none, text: PersonaPresence[PersonaPresence.none] },
```
ecraig12345: I'm not sure it's really necessary to have the dropdown to switch presence--seems more to the point to either choose one presence and use it for all the faces, or set the presence for each face individually (which would be more realistic).

mtennoe: Probably a bit overkill, but doesnt hurt. I am happy with either solution

maxwellred: Alright I just dropped a little randomizer function into the example file in order to change the presences around but keep them consistent to something (the name in this case). Im assuming the user would be passing/controlling what presence the coins are via the data field of the personas array but i didn't want to change that pice of code for this.

packages/office-ui-fabric-react/src/components/Facepile/examples/Facepile.Basic.Example.tsx
```diff
@@ -100,4 +102,18 @@ export class FacepileBasicExample extends React.Component<{}, IFacepileBasicExam
       }
     );
   };
+
+  private _personaPresence = (personaName: any): any => {
```
ecraig12345: Could be done more simply like this: 
```ts
  private _personaPresence = (personaName: string): PersonaPresence => {
    const presences = [
      PersonaPresence.away,
      PersonaPresence.busy, 
      PersonaPresence.online, 
      PersonaPresence.offline,
      PersonaPresence.offline
    ];

    return presences[personaName.charCodeAt(1) % 5];
  };
```

packages/office-ui-fabric-react/src/components/Facepile/examples/Facepile.Basic.Example.tsx
```diff
@@ -26,13 +26,15 @@ export class FacepileBasicExample extends React.Component<{}, IFacepileBasicExam
...
+
     const facepileProps: IFacepileProps = {
       personaSize: personaSize,
       personas: facepilePersonas.slice(0, numberOfFaces),
       overflowPersonas: facepilePersonas.slice(numberOfFaces),
       getPersonaProps: (persona: IFacepilePersona) => {
         return {
-          imageShouldFadeIn: this.state.imagesFadeIn
+          imageShouldFadeIn: this.state.imagesFadeIn,
+          presence: this._personaPresence(persona.personaName)
```
ecraig12345: You might add a note that typically the persona presence would be included when generating each facepile persona (probably when processing data from a server), but here we're assigning an arbitrary presence to each persona based on its name for demonstration purposes.

packages/office-ui-fabric-react/src/components/Facepile/examples/Facepile.Basic.Example.tsx
```diff
@@ -100,4 +106,16 @@ export class FacepileBasicExample extends React.Component<{}, IFacepileBasicExam
       }
     );
   };
+
+  private _personaPresence = (personaName: any): any => {
```
ecraig12345: Does this have to use any? As far as I can tell you should be able to do `personaName: string` and `PersonaPresence` for the return type and array type (with `!` operators in places if needed).

maxwellred: You got it!

PR #10876, opened 2019-10-17
Pickers: add missing fonts and fix color

### LINKED ISSUE #10746: Tag Picker doesn't respect theme
When switching to a dark theme, you can see that some of the colors in the Tag Picker don't match the generated theme. The input field's text color is black, even though the background color is also black.

Aside: The incorrect font (Times New Roman) in the picker items may also be part of a similar issue, although that only repros in Storybooks or in Codepen.

### Environment Information

- **Package version(s)**: office-ui-fabric-react 7.45.1
- **Browser and OS versions**:
Version 77.0.3865.90 (Official Build) (64-bit)

Codepen to repro the issue: https://codepen.io/hegi/pen/vYYONzy

#### Actual behavior:

What it looks like:
image

#### Expected behavior:

Input should look similar to `TextField`.

Requested priority: Low

Products/sites affected: Dynamics 365 Field Service

This bug was created from #10745.

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10746 

#### Description of changes
Added missing font and theme color on input text and TagItem text

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10876)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-16-21-36-43-bug-fix-tagpicker.json
+++ b/change/office-ui-fabric-react-2019-10-16-21-36-43-bug-fix-tagpicker.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "TagPicker: make tag item and input text to respect theme color and have correct font",
+  "packageName": "office-ui-fabric-react",
+  "email": "xgao@microsoft.com",
+  "commit": "5612231e960d701aecf968601e1d252dc09338d2",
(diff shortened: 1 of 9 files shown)


xugao:
the screener test failures are not expected. will check and fix

ecraig12345:
I'm having trouble seeing what changed with the screener failures...

PR #10882, opened 2019-10-17
ColorPicker: add keyboard support and aria labels

### LINKED ISSUE #8012: Make ColorPicker sliders keyboard accessible
Currently, the sliders within the ColorPicker (s/v rectangle, hue, alpha) can't be selected and moved around using the keyboard. At least to me, it seems like a user ought to be able to tab to the slider, hear its current value announced, and move it around using the arrow keys.

ColorSlider and ColorRectangle

**Have you discussed this feature with our team, and if so, who**
@betrue-final-final brought this up a long time ago under #1598 but no action was ever taken there. (haven't discussed with him, just found the issue)

Even without this, the ColorPicker is still somewhat accessible (the user can type in the text fields) but that's probably not ideal. The main functional difference is that the sliders allow specific adjustment in terms of of hue, saturation, and value, whereas the text fields only allow RGBA (and hex) adjustment. 

It's possible that accessibility of the ColorSlider could be fixed by using a regular Slider instead and re-styling it (not sure how hard that would be). ColorRectangle keyboard handling and value announcement could be implemented based on Slider.

If we do this, we should ensure the sliders have proper ARIA labels as referenced in #8463.

### LINKED ISSUE #8463: ColorPicker A11y Pass
Tracking all of the issues brought up with the 3/25 accessibility pass on ColorPicker control
Details on individual bugs can be found on [Sharepoint](https://microsoft-my.sharepoint.com/:x:/r/personal/v-amaver_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B21986988-6108-426e-b79f-214f796cc784%7D&action=default&uid=%7B21986988-6108-426E-B79F-214F796CC784%7D&ListItemId=1898&ListId=%7B395E1ECB-5002-440E-81C5-6A090E98077C%7D&odsp=1&env=prodbubble)

- ~~[ ] Bug1: [Accessibility][Programmatic access ]: Label should be defined for the sliders available for changing the colors.~~ *Doesn't block essential functionality (see below); will be fixed if we do feature #8012.*
- ~~[ ] Bug2: [Accessibility][Usable]: In scan mode backward navigation focus is going on selected color panel but screen reader is providing details about table. And forward navigation focus does not go to the selected colors quare panel, sliders, and the color picker.~~ *can't reproduce/not a bug (see below)*

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #8463, fixes #8012

[Link to built ColorPicker](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/10882/merge/fabric-website-resources/dist/index.html#/examples/colorpicker)

#### Description of changes

Add keyboard support to the color picker sliders and rectangle (expanding upon @dzearing's PR #10725).

Add more aria labels and attributes for various parts of the ColorPicker. Feedback on whether this is the "right" way to handle things is VERY welcome!

Make RTL handling (or lack of) consistent.

Misc other minor fixes called out at bottom.

##### ColorPicker aria
- Add an aria label, default "Color picker, (color) selected"
  - For the selected value, add spaces between the elements of the color for more helpful reading (not trying to run color codes together as words)
- Add `role="group"` -- not sure if this is the appropriate role (some role is required for the label to be read)
- Set the text field table to `role="group"` to prevent it from being read as a table (semantically it shouldn't be a table, so reading it as one is probably unhelpful)
- Room for improvement: I suspect it would be helpful to provide a description for the region (such as number of sliders and fields), but that means there's a LOT of words to wait through before you get to hear the value of the first focused element (the rectangle)

I wish the control could be one labeled tab stop, with some way besides tab to navigate between the sub-controls, but FocusZone won't work for that because the arrow keys are already needed to move around within the controls.

A related option would be to put the controls inside a FocusTrapZone which is initially disabled. When entering the control, focus would land on a hidden button with the overall label and interaction instructions (all other sub-controls would be non-tabbable). Activating the button would activate the trap zone and make the controls inside tabbable. Then you'd press some specific key (or another hidden button) to get out. However, that might confuse/frustrate people because it's not the most standard interaction pattern. It might also be hard to integrate well with mouse interaction.

##### ColorRectangle aria
- Add an aria label, default "Saturation and brightness"
  - I chose "brightness" because it's more commonly understood than "value"
- Add an aria description with usage info, default "Use left and right arrow keys to set saturation. Use up and down arrow keys to set brightness."
  - Room for improvement: Edge/Narrator doesn't read the description...
- Add `aria-valuetext` with the current value, default "Saturation {0} brightness {1}"
- Add `role="slider"` and related attributes (`aria-valuemin`, `aria-valuemax`, `aria-valuenow`, `aria-valuetext`). Various complications arise from the fact that aria has no notion of a 2D slider:
  - `aria-valuenow` can only be one number. In theory screen readers should use `aria-valuetext` instead if it's provided, but that doesn't always happen. Not providing `aria-valuenow` isn't an option because per the spec it defaults to halfway between min and max. So I set it to the value of the most recently adjusted dimension.
  - Strictly there should be separate `aria-valuemin` and `aria-valuemax` for each dimension, but conveniently min/max are the same for saturation and value/brightness
  - Sliders have an implicit `aria-orientation` of `horizontal`, which isn't quite accurate in this case, but `both` is not an option...

Alternatively, to avoid all this complication, we could just make the rectangle `aria-hidden`.

##### ColorSlider aria
- Add an aria label, default "Hue" or "Alpha"
- Add `role="slider"` and related attributes

##### RTL
I noticed the color picker controls were previously in a weird state of only about *halfway* flipping in RTL. If we want to fully support RTL, these are all the things that need to flip:
- Gradient styles handled by merge-styles
- Gradient styles *not* handled by merge-styles (due to frequent dynamic updates)
- Mouse movement in slider
- Mouse movement in rectangle
- Keyboard in slider
- Keyboard in rectangle
- Thumb positioning in slider
- Thumb positioning in rectangle

That's a lot of places to get and keep correct, so I suggest that we **don't flip anything** for RTL, which is what I've implemented in this PR.

##### Other
- Get rid of `UNSAFE_componentWillReceiveProps`
- Add tests for ColorSlider (most of which I'd written previously but not submitted)
- Add tests for new functionality
- Call `onChange` *before* setting state, then don't set state if event default is prevented (thanks David)
- More idiomatic gradient and overlay style handling for slider (pass in just a color for the alpha overlay, and move the static hue and alpha styles into styles file so they can be overridden if desired)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10882)


micahgodbolt:
uuugggh details list snapshots

image

ecraig12345:
@micahgodbolt I think I figured out how to get rid of that finally! (@KevinTCoughlin the small DetailsList example change should be the only part of this you need to look at, unless of course you want to look at the rest.)

ecraig12345:
Deploy link is ready for anyone interested in testing: http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/10882/merge/fabric-website-resources/dist/index.html#/examples/colorpicker

As far as I can tell, the bundle size increase reported by size auditor is due to the amount of functionality added, or possibly the KeyCodes import (so hard to avoid).

Screener "failure" is due to an incorrect accepted result.

ecraig12345:
Made a separate PR #10898 for the DetailsList snapshot thing (I'll remove that change from this PR once the other merges).

packages/office-ui-fabric-react/src/components/ColorPicker/ColorRectangle/ColorRectangle.base.tsx
```diff
@@ -81,16 +102,68 @@ export class ColorRectangleBase extends React.Component<IColorRectangleProps, IC
...
-  private _onMouseDown = (ev: React.MouseEvent<HTMLElement>): void => {
+  private _onKeyDown = (ev: React.KeyboardEvent<HTMLElement>): void => {
+    const { color } = this.state;
+    let { s, v } = color;
+
+    const increment = ev.shiftKey ? 10 : 1;
+
+    // Intentionally DO NOT flip the color picker in RTL: its orientation is not very meaningful,
+    // and getting all the math and styles flipped correctly is tricky
+    switch (ev.which) {
```
dzearing: hmmm.

ecraig12345: I'm open to trying to do RTL, it's just A LOT of things to get right (some of which were totally wrong before this PR). Copying from the description:

- Static gradient styles handled by merge-styles (easy)
- Gradient styles not handled by merge-styles (due to frequent updates)
- Mouse movement in slider
- Mouse movement in rectangle
- Keyboard in slider
- Keyboard in rectangle
- Thumb positioning in slider (not handled by merge-styles due to frequent updates)
- Thumb positioning in rectangle (same)

dzearing:
Overall, this is awesome. Great job!

packages/office-ui-fabric-react/src/components/ColorPicker/ColorRectangle/ColorRectangle.base.tsx
dzearing: Maybe ok to save for another day.

dzearing:
I did not validate that screen readers read this correctly. Would be good to review with Jon or someone

packages/office-ui-fabric-react/src/components/ColorPicker/ColorRectangle/ColorRectangle.base.tsx
ecraig12345: Opened #10900 to track

betrue-final-final:
Is there a reason the color bar below got rearranged? It's not a huge deal, but just wondering the reason.

micahgodbolt:
from narrator standpoint, it's really good. The main color picking surface works great in edge. Reads out coordinates with good labels etc, and doesn't spam you if you hold an arrow key down (unlike Chrome..yuck! sat.sat.sat.sat.saturation 8, brightness 29. All max headroom over there.

I'd suggest improving the string to be read on each hue/saturation slider change. Right now when you land on it, it says "hue 35", but then each step left or right it only says "35...36....37". This might be personal preference, but I'd prefer hearing "hue 36, hue 37 etc". Also, coming into hue and saturation, I don't know how many steps there are. Landing on huge should probably read "hue, slider at 0 of 359". 

Lastly, you could might consider adding a 'press right arrow to increase, hold down shift to increase by 10'. This one feels a bit verbose, but we should probably at least make mention of that capability in the docs. 

Oh one last thing, if you happen to pick a color via mouse, it can have decimal coordinates.  Narrator reads that out: 27.264890124345

Either need to round off what narrator reads, or simply clean that up in state.

ecraig12345:
Finally noticed @micahgodbolt's comment (thanks for looking!) -- 

> I'd suggest improving the string to be read on each hue/saturation slider change. Right now when you land on it, it says "hue 35", but then each step left or right it only says "35...36....37". This might be personal preference, but I'd prefer hearing "hue 36, hue 37 etc". Also, coming into hue and saturation, I don't know how many steps there are. Landing on huge should probably read "hue, slider at 0 of 359".
> 
> Lastly, you could might consider adding a 'press right arrow to increase, hold down shift to increase by 10'. This one feels a bit verbose, but we should probably at least make mention of that capability in the docs.

I think I'd prefer to stick with just announcing the value for the 1D sliders, since a slider is a common widget that users should be familiar with and a screen reader can fully represent. (If a consumer wanted more verbose instructions, they could pass in string overrides.) And at least in Narrator+Edge, it does actually announce the min/max when you land on the slider, a second or so after the main description finishes.

The only reason I'm using more verbose strings for the rectangle is that there's no way to properly represent (essentially) a 2D slider with ARIA. 

> Oh one last thing, if you happen to pick a color via mouse, it can have decimal coordinates. Narrator reads that out: 27.264890124345

Good catch. I'll add something to round off the coordinates before putting them in the state.

ecraig12345:
> Is there a reason the color bar below got rearranged? It's not a huge deal, but just wondering the reason.

I think I answered this on chat, but for the public record--assuming you mean the RTL version in the screener states, that changed because previously the color picker *halfway* adjusted to RTL, in a way that was really broken. So I updated it to *not adjust at all* for now and made #10900 to track the potential work of adding full proper adjustment.

PR #10888, opened 2019-10-17
Theme Designer: pull semantic slots from primary source instead of have a secondary mapping.

aneeshack4:
#### Pull request checklist

#### Description of changes

Right now, the semantic slots on [https://aka.ms/themedesigner](https://aka.ms/themedesigner) are pulled from a secondary, duplicate mapping in `SemanticSlots.tsx` that has to be updated every time the primary source in theme.ts is changed. This is causing bugs because they are not always updated at the same time with the same changes so removing this need by having `SemanticSlots.tsx` employ a bunch of string manipulation algorithms and dictionaries to properly pull the list of slots that are pulled and match the right palette values with each semantic slot and construct an accurate mapping for the theme designer.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10888)


apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -432,33 +58,107 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
     );
   };
 
-  const fillVariantSlotsList = (variantType: VariantThemeType, mappingType: {}): JSX.Element[] => {
+  const trimPaletteSlots = (paletteSlots: IPalette): IPaletteSlots => {
```
phkuo: add a comment describing this method's purpose

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -432,33 +58,107 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
...
+    for (let palette in paletteSlots) {
+      if (
+        palette.startsWith('theme') ||
+        palette.startsWith('neutral') ||
+        palette === 'black' ||
+        palette === 'white' ||
+        palette === 'redDark' ||
+        palette === 'accent'
+      ) {
+        trimmedPaletteSlots[palette] = (paletteSlots as any)[palette];
```
phkuo: shouldn't need this any here?

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -432,33 +58,107 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
...
+    return (
+      semanticString.startsWith('body') ||
+      semanticString.startsWith('primary') ||
+      semanticString.startsWith('disabled') ||
+      semanticString.startsWith('focus') ||
+      semanticString.startsWith('variant') ||
+      semanticString.startsWith('default') ||
+      semanticString.includes('action') ||
+      semanticString.includes('link') ||
+      semanticString.startsWith('accent') ||
```
phkuo: is accent both a semantic and a palette slot?

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -432,33 +58,107 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
...
+      semanticString.startsWith('list') ||
+      semanticString.startsWith('button') ||
+      semanticString.startsWith('menu') ||
+      semanticString.startsWith('input') ||
+      semanticString.startsWith('small')
+    );
+  };
+
+  const trimSemanticSlotsOrNames = <T extends unknown>(semanticSlots: T) => {
```
phkuo: comment

phkuo: do you need the `extends unknown`?
also you can say this returns a T as well

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -432,33 +58,107 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
...
+        }
+      }
+      return trimmedSemanticSlots;
+    }
+  };
+
+  const fillVariantSlotsList = (variantType: VariantThemeType): JSX.Element[] => {
     let currThemeVariant: ITheme;
+    let noneVariant = getVariant(props.theme, VariantThemeType.None);
     if (props.theme) {
```
phkuo: I think this would still work even with an undefined theme?

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -432,33 +58,107 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
...
+      const trimmedSemanticSlots = trimSemanticSlotsOrNames(currVariantSemanticSlots);
+      const currVariantPaletteSlots = noneVariant.palette; // palette slot values should be based off the default variant
+      const trimmedPaletteSlots = trimPaletteSlots(currVariantPaletteSlots);
+      const mapping: IMapping = {};
+      // Iterate through the list of semantic colors
+      // for each semantic color, check if it's hex color string is in the list of palette colors
+      // if it is, add it to the mapping
+      for (let semanticColor in trimmedSemanticSlots) {
+        if (semanticColor) {
+          const paletteColorHexStr = (trimmedSemanticSlots as any)[semanticColor];
```
phkuo: you shouldn't need this any

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -432,33 +58,107 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
...
+          for (let palette in trimmedPaletteSlots) {
+            if (trimmedPaletteSlots[palette] === paletteColorHexStr) {
+              mapping[semanticColor] = palette;
+            } else if (paletteColorHexStr === 'transparent') {
+              mapping[semanticColor] = 'transparent';
+            }
```
phkuo: else if it's a custom color?

apps/theming-designer/src/components/SemanticSlots.tsx
KevinTCoughlin: @aneeshack4, I'm also curious about the usage of `unknown` here because I am unfamiliar with the new TypeScript keyword.

What does `T extends unknown` express?

apps/theming-designer/src/components/SemanticSlots.tsx
aneeshack4: It says required because IPalette doesn't have an index signature

apps/theming-designer/src/components/SemanticSlots.tsx
aneeshack4: Yes there are semantic slots that have the string accent in them and accent is a palette slot as well.

apps/theming-designer/src/components/SemanticSlots.tsx
aneeshack4: `unknown` is a type-safe alternative to `any` https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type. T must extend something, it can technically extend `any` here too but according to the docs, `unknown` is more type safe... 

When I try to set the return type to `T` errors get thrown saying Type 'any[]' is not assignable to type 'T'.ts for `trimmedSemanticSlotNames` and Type 'ISemanticSlots' is not assignable to type for `trimmedSemanticSlots`

I would like to find a way to specify a return type for this method that would accommodate both `trimmedSemanticSlotNames` and `trimmedSemanticSlots` - that's one of the reasons

apps/theming-designer/src/components/SemanticSlots.tsx
aneeshack4: When I remove it, it says "type 'any[] | ISemanticSlots' has no index signature" so I think I do need it... I've been trying to find ways to remove all usage of any and it's been a struggle.

apps/theming-designer/src/components/SemanticSlots.tsx
aneeshack4: yes moving these 2 lines above inside the if - thanks for catching this!

apps/theming-designer/src/components/SemanticSlots.tsx
aneeshack4: What do you mean by custom color?

apps/theming-designer/src/components/SemanticSlots.tsx
cliffkoh: Consider using type overloads instead of doing what you're currently doing, which would be simpler. The current thing is.. equivalent to just saying `(semanticSlots: any) => { }` 

Hard for me to prescribe the exact syntax to use as it isn't super obvious what the possible inputs are and what the output in each case will be (maybe this is a sign that this function can use JSDoc or some inline comments)

cliffkoh: Oh I just realized you actually did add comments.. (which did not show in the old iteration) but it doesn't really explain input/outputs.

apps/theming-designer/src/components/SemanticSlots.tsx
```diff
@@ -58,6 +58,9 @@ export const SemanticSlots: React.StatelessComponent<ISemanticSlotsProps> = (pro
     );
   };
 
+  // Filters the list of palette slots pulled from getVariant to only include the ones starting with 'theme'
```
cliffkoh: please use jsdoc syntax

apps/theming-designer/src/components/SemanticSlots.tsx
aneeshack4: Nvm done - removed all any usage!

apps/theming-designer/src/components/SemanticSlots.tsx
aneeshack4: @cliffkoh showed me how to get rid of all the any usage so no longer using unknown!

apps/theming-designer/src/components/SemanticSlots.tsx
aneeshack4: Nvm done - removed all any usage!

PR #10897, opened 2019-10-17
adding string index sig to IButtonProps

### LINKED ISSUE #10896: IButtonProps should support string index sig

Be able to pass arbitrary props such as data- and aria- to a Button through an object spread onto it. Currently you can only do this in JSX.

IButtonProps

Not yet

brandonthomas:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10896

#### Description of changes

- expose string index signature for IButtonProps.

#### Focus areas to test
That Buttons can take arbitrary props from objects being spread into them. Such examples include aria- and data-

const myObject = {
“data-test-id”: “anId”
}

<Button {...myObject}\>

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10897)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-17-16-53-21-buttonPropsStringIndex.json
+++ b/change/office-ui-fabric-react-2019-10-17-16-53-21-buttonPropsStringIndex.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Adding string index signature to IButtonProps",
+  "packageName": "office-ui-fabric-react",
+  "email": "brandth@microsoft.com",
+  "commit": "2223d173132671f4b73887f07eed8106c1434f85",
+  "date": "2019-10-17T23:53:21.868Z",
+  "file": "C:\\repos\\office-ui-fabric-react\\change\\office-ui-fabric-react-2019-10-17-16-53-21-buttonPropsStringIndex.json"
+}
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -1880,6 +1880,7 @@ export interface IButton {
 
 // @public (undocumented)
 export interface IButtonProps extends React.AllHTMLAttributes<HTMLAnchorElement | HTMLButtonElement | HTMLDivElement | BaseButton | Button | HTMLSpanElement> {
+    [propertyName: string]: any;
     allowDisabledFocus?: boolean;
     ariaDescription?: string;
     ariaHidden?: boolean;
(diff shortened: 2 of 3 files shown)


brandonthomas:
I'm also hoping to backport to 6.x

dzearing:
@khmakoto If I understand correctly, here's exactly the problem:

```tsx
// Sweet
<PrimaryButton data-foo="bar"/>

// Not sweet
const things = { ['data-foo']: 'bar' };
<PrimaryButton {...things } />
```

brandonthomas:
@dzearing yup that’s exactly the issue. Not sure how else to solve. I talked to Jason about this issue and he forwarded me a typescript issue tracking how we can better type things like this. Currently it’s not possible.

brandonthomas:
Updated the description to better reflect testing focus

khmakoto:
@dzearing then, is this something we want to allow? Since like we can lose some of the benefits from typing if we do this but I'm not completely opposed to it.

brandonthomas:
@khmakoto how else can we support things like data- props being spread in? If there's another way to spread in a data object I'm more than happy to do that instead. It is weird that you can do this within JSX but not on the object of the same type.

dzearing:
We should ask the TypeScript team.

brandonthomas:
The issue is we need it for a library we make on top of Fabric so our consumers may or may not know this. For now they can do what you mentioned or ts-ignore the next line which is also less than ideal. Currently this isn't possible to support any other way but there is a thread around this. For the data- and aria- issue specifically Ryan Cavanaugh said they would open an issue on it directly.

Sort of linked to the following: https://github.com/microsoft/TypeScript/issues/6579

brandonthomas:
@dzearing I'm curious why this is not ok here but is ok in other parts of Fabric. Is there a specific reason for this or did the other examples just slip through?

micahgodbolt:
Seeing that we aren't wanting to drop a wildcard into button, should we close this PR down and open an issue tracking this problem?

brandonthomas:
> 
> 
> Seeing that we aren't wanting to drop a wildcard into button, should we close this PR down and open an issue tracking this problem?

As long as we address the underlying issue that we want to be able to pass valid props to components through an object. At the moment the only way to do this is the tsignore which is gross

ecraig12345:
> As long as we address the underlying issue that we want to be able to pass valid props to components through an object. At the moment the only way to do this is the tsignore which is gross

In Fabric we typically use casts to `any` instead of outright tsignore. Still not great but slightly less nuclear.

micahgodbolt:
yeah, casting to any on a per need basis is better than forcing the interface to basically be 'any' for everyone at all times.

PR #10901, opened 2019-10-18
Tooltip: clarify examples and docs around overflow modes

### LINKED ISSUE #10732: TooltipOverflowMode doesn't seem to have any effect
### Environment Information

- **Package version(s)**: ^7.29.0
- **Browser and OS versions**: any

#### Actual behavior:
`TooltipOverflowMode.Parent` and `TooltipOverflowMode.Self` don't seem to do anything.
[This is CodePen example](https://codepen.io/isabelacmor/pen/QWWwOqx?editors=0010), the the `p` with ellipsis should show the tooltip on hover and the other `p` should not show a tooltip. As you can see, regardless of what `TooltipOverflowMode` is used, the `p` with ellipsis never gets a tooltip on hover.

#### Expected behavior:
`TooltipOverflowMode.Parent` and `TooltipOverflowMode.Self` should do what the documentation says.

```
export declare enum TooltipOverflowMode {
    /** Only show tooltip if parent DOM element is overflowing */
    Parent = 0,
    /** Only show tooltip if tooltip host's content is overflowing */
    Self = 1
}
```

Requested priority: High

Products/sites affected: Any using Tooltip

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10732

#### Description of changes

Currently the TooltipHost overflow modes are not very well-documented, causing some confusion about usage (such as the linked issue). This PR improves the documentation for those props, as well as adding a more detailed example of the different overflow modes.

Other changes:
- TooltipHost was re-generating the ID for the Tooltip every render, which seemed like it could cause problems. I changed it to preserve the ID until the tooltip is removed (could also generate the ID in the constructor and keep it for the component lifetime).
- Clean up other Tooltip/TooltipHost prop documentation
- Update the examples to use function components
- Remove the NoScroll example since it seems to have very limited relevance (the new comment in Callout explains the previously-undocumented reason why the example was added, based on digging through old PRs)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10901)


packages/office-ui-fabric-react/src/components/Tooltip/examples/Tooltip.Absolute.Position.Example.tsx
```diff
@@ -1,42 +0,0 @@
-import * as React from 'react';
```
ecraig12345: This file was renamed and modified, not deleted.

packages/office-ui-fabric-react/src/components/Tooltip/examples/Tooltip.Overflow.Example.tsx
```diff
@@ -1,59 +1,83 @@
...
-        >
-          <div
-            style={{
-              overflow: 'hidden',
-              textOverflow: 'ellipsis',
-              whiteSpace: this.state.overflow ? 'nowrap' : 'inherit',
-              width: this.state.overflow ? '200px' : 'auto',
-              border: '1px solid black'
-            }}
+import { Toggle } from 'office-ui-fabric-react/lib/Toggle';
```
ecraig12345: The overflow example now demonstrates both Parent and Self

packages/office-ui-fabric-react/src/components/Callout/Callout.types.ts
```diff
@@ -139,6 +139,9 @@ export interface ICalloutProps extends React.HTMLAttributes<HTMLDivElement> {
 
   /**
    * CSS style to apply to the callout.
+   *
+   * If you set `overflowY` in this object, it provides a performance optimization by preventing
```
ecraig12345: This is the (previously undocumented) reason why the NoScroll example existed. I guess if this perf optimization seems useful to highlight, I could bring the example back, it just seemed like a major edge case.

packages/office-ui-fabric-react/src/components/Tooltip/TooltipHost.base.tsx
```diff
@@ -236,9 +237,10 @@ export class TooltipHostBase extends React.Component<ITooltipHostProps, ITooltip
 
   private _toggleTooltip = (isTooltipVisible: boolean): void => {
     if (this.state.isTooltipVisible !== isTooltipVisible) {
+      this._defaultTooltipId = getId('tooltip');
```
micahgodbolt: is this redundant with line 40?

packages/office-ui-fabric-react/src/components/Tooltip/TooltipHost.types.ts
```diff
@@ -28,12 +28,15 @@ export enum TooltipOverflowMode {
...
+  /**
+   * Only show tooltip if tooltip host's content is overflowing.
+   * Note that this does not check the children for overflow, only the TooltipHost root.
+   */
   Self
 }
 
 /**
- * Tooltip component props.
+ * TooltipHost props. All native props are passed through to the Tooltip itself.
```
micahgodbolt: 'all native props' isn't 100% clear. Is this due to the fact that it extends HTMLAttributes? Maybe there's a better, more consistent way we can make that known for all controls?

packages/office-ui-fabric-react/src/components/Tooltip/TooltipHost.types.ts
```diff
@@ -44,69 +47,87 @@ export interface ITooltipHostProps extends React.HTMLAttributes<HTMLDivElement |
...
   /**
-   * Only show if there is overflow. If set, the tooltip hosts observes  and only shows the tooltip if this element has overflow.
-   * It also uses the parent as target element for the tooltip.
+   * Class name to apply to the *tooltip itself*, not the host.
+   * To apply a class to the host, use `hostClassName` or `styles.root`.
+   */
+  className?: string;
+
+  /**
+   * If this is unset (the default), the tooltip is always shown even if there's no overflow.
```
micahgodbolt: These strings get placed into the 'description' of the prop. I'm worried as we try to add more 'details', the description is lost, and our props list gets harder to parse. 

image

micahgodbolt: i agree, this information is useful, but curious if there's a better way to do this in the future.

packages/office-ui-fabric-react/src/components/Tooltip/TooltipHost.base.tsx
ecraig12345: No, it's meant to re-initialize the ID since it's a different tooltip, meant to preserve the previous behavior of having a different ID each time the tooltip opens......but that may not actually be necessary/desirable, so I can remove it.

packages/office-ui-fabric-react/src/components/Tooltip/TooltipHost.types.ts
ecraig12345: This is a really weird case because instead of the native props going on this component (the TooltipHost), they go on its child Tooltip. I should clarify that in the comment.

packages/office-ui-fabric-react/src/components/Tooltip/TooltipHost.types.ts
ecraig12345: I think the "right" solution here is to introduce line break support in the descriptions...

ecraig12345: Opened #11388 to track adding line breaks

PR #10913, opened 2019-10-18
Update DatePicker disableAutoFocus functionality

phtucker:
#### Pull request checklist

#### Description of changes

This change expands the functionality of the disableAutoFocus property of DatePicker to prevent expansion of the picker when clicking on the TextField when property is set to true and allowTextInput is set to true. 

This is to address a bad experience on iOS where tapping on the TextField where allowTextInput is true shows the picker, but also expands the iOS onscreen keyboard, which in turn steals focus causing the picker to collapse but not refocus the TextField. This behavior is not consistent based on layout, but is consistently a bad experience.

The result is that a user intending to enter a date via the onscreen keyboard in iOS faces difficulties.

After this change, setting disableAutoFocus to true (and allowTextInput) will allow the DatePicker to act more like a ComboBox wherein only clicking/tapping on the calendar icon will expand the picker, click on the TextField will not (but it will collapse it if it's already expanded) allowing users to easily focus the TextField for input.

When disabledAutoFocus is true and allowTextInput is true the only way to expand the date picker is by clicking on the calendar icon, because of this the hit area of the icon is expanded by adding some padding (no change to visual).

#### Focus areas to test

DatePicker

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10913)


change/office-ui-fabric-react-2019-10-18-14-55-55-phtucker-updateDatePickerDisableAutoFocusFunctionality.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "Do not expand data picker on TextField click when disableAutoFocus is true and allowTextInput is true.",
```
evlevy: Typo: ""Do not expand _data_ picker..."

packages/date-time/src/components/DatePicker/DatePicker.base.tsx
```diff
@@ -370,20 +370,28 @@ export class DatePickerBase extends BaseComponent<IDatePickerProps, IDatePickerS
...
-      this._showDatePickerPopup();
-    } else {
-      if (this.props.allowTextInput) {
-        this.setState({
-          isDatePickerShown: false
-        });
+    if (!this.props.disableAutoFocus) {
+      if (!this.state.isDatePickerShown && !this.props.disabled) {
+        this._showDatePickerPopup();
+      } else {
```
evlevy: Use `else if ...` instead of `else { if ...`

packages/date-time/src/components/DatePicker/DatePicker.base.tsx
```diff
@@ -370,20 +370,28 @@ export class DatePickerBase extends BaseComponent<IDatePickerProps, IDatePickerS
...
+      this._dismissDatePickerPopup();
     }
   };
 
   private _onIconClick = (ev: React.MouseEvent<HTMLElement>): void => {
     ev.stopPropagation();
-    this._onTextFieldClick(ev);
+    if (!this.state.isDatePickerShown && !this.props.disabled) {
+      this._showDatePickerPopup();
+    } else {
```
evlevy: Use `else if ...` instead of else `{ if ...`

packages/date-time/src/components/DatePicker/DatePicker.styles.ts
```diff
@@ -52,8 +52,12 @@ export const styles = (props: IDatePickerStyleProps): IDatePickerStyles => {
     callout: [classNames.callout],
     icon: [
       DatePickerEvent,
-      !label && [classNames.withoutLabel, { top: '7px' }],
-      label && [classNames.withLabel, { bottom: '5px' }],
+      {
```
evlevy: Why not add these directly to DatePickerEvent?

phtucker: I had thought, from the name, that the const was used elsewhere but it doesn't seem to be. I should rename it and add these then. DatePickerEvent is a misleading name for its usage.

PR #10927, opened 2019-10-22
Panel: Use semantic colors

### LINKED ISSUE #10740: Panel: Use semantic colors
Panel is not using any semantic colors. There are 7 palette colors that need to be converted to semantic.

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10740

#### Description of changes

Use semantic colors in Panel instead of palette colors.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10927)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
@@ -188,7 +188,7 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
     theme,
     type = PanelType.smallFixedFar
   } = props;
-  const { palette, effects, fonts } = theme;
+  const { effects, fonts, semanticColors } = theme;
   const classNames = getGlobalClassNames(GlobalClassNames, theme);
   const isCustomPanel = type === PanelType.custom || type === PanelType.customNear;
 
@@ -228,7 +228,7 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
     main: [
       classNames.main,
       {
-        backgroundColor: palette.white,
+        backgroundColor: semanticColors.bodyBackground,
         boxShadow: effects.elevation64,
         pointerEvents: 'auto',
         position: 'absolute',
@@ -245,8 +245,8 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
         width: panelWidth.full,
         selectors: {
           [HighContrastSelector]: {
-            borderLeft: `3px solid ${palette.neutralLight}`,
-            borderRight: `3px solid ${palette.neutralLight}`
+            borderLeft: `3px solid ${semanticColors.variantBorder}`,
+            borderRight: `3px solid ${semanticColors.variantBorder}`
           },
           ...getPanelBreakpoints(type)
         }
@@ -310,7 +310,7 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
       classNames.headerText,
       fonts.xLarge,
       {
-        color: palette.neutralPrimary,
+        color: semanticColors.bodyText,
         lineHeight: '27px',
         margin: 0,
         overflowWrap: 'break-word',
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
```diff
@@ -245,8 +245,8 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
         width: panelWidth.full,
         selectors: {
           [HighContrastSelector]: {
-            borderLeft: `3px solid ${palette.neutralLight}`,
-            borderRight: `3px solid ${palette.neutralLight}`
+            borderLeft: `3px solid ${semanticColors.variantBorder}`,
```
ecraig12345: This is the first time I've done a color conversion, and most of the colors seemed pretty obvious except for this one and the footer border later. Not sure who to ask to confirm I used the right color here.

marygans: `neutralLight` doesn't seem right for high contrast, Im expecting something like `borderRightColor: 'Highlight'` 
I'll ask @betrue-final-final for his thoughts!

ecraig12345: Yeah I thought that was weird too--I chose `variantBorder` for the new color since it's the same as `neutralLight` usually (and might make sense semantically?) but for sure open to other suggestions.

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
```diff
@@ -346,8 +346,8 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
         transition: `opacity ${AnimationVariables.durationValue3} ${AnimationVariables.easeFunction2}`
       },
       isFooterSticky && {
-        background: palette.white,
-        borderTopColor: palette.neutralLight
+        background: semanticColors.bodyBackground,
+        borderTopColor: semanticColors.variantBorder
```
ecraig12345: @marygans Do you think `variantBorder` is correct here?

marygans: Yes this looks good! (Also checked with design for confirmation)

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
marygans: Yeah, I see why you chose it! Also, Ben says the system color reference should be `'WindowText'`

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
```diff
@@ -346,8 +346,8 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
         transition: `opacity ${AnimationVariables.durationValue3} ${AnimationVariables.easeFunction2}`
       },
       isFooterSticky && {
-        background: palette.white,
-        borderTopColor: palette.neutralLight
+        background: semanticColors.bodyBackground,
```
aneeshack4: does it make sense to user a body slot for the footer here? (just asking out of curiosity - it really might make sense to do this but wondering if that's the pattern we're meant to follow in other places throughout the codebase too)

marygans: Hi @aneeshack4 do you have more context on how the pattern different? And what is being used everywhere else?

aneeshack4: I don't have the full context but based on the recent PRs I've seen using semantic slots, it seems like the usage varies a lot. I think body makes the most sense here actually even though words wise it's a bit weird to use body on footer but there are no better options from the semantic slots list.

marygans: Yeah I see what you mean, wording-wise `body` is weird to use on a `footer`..seems to be a special case since it's a sticky footer we need that white BG layer to overlay on top of panel content when scrolling

marygans:
@micahgodbolt does this look good to go?

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
micahgodbolt: bodyBackground works for me. the sticky footer needs a bg that matches the rest of the panel so that it can sit on top of panel content scrolling under

PR #10928, opened 2019-10-22
Panel example updates

### LINKED ISSUE #10781: Panel: "Controlled visibility" example close button doesn't work
[Functional : Panel Control] : No action is getting preformed after activating the 'Close' button either by using keyboard or by mouse.

**User Experience:**
All user will face difficulty if controls will not work as per expected. 

**Test Environment:** 
OS: Version: RS5_Prerelease(1903 OS Build 18999.1)
URL: https://developer.microsoft.com/en-us/fabric#/controls/web/panel
Edge Browser Version:
Microsoft Edge: 44.18999.1.0
Microsoft EdgeHTML: 18.18999

**Usage:**
1.Panel - Controlled Visibility.

**Repro Steps:**
1. Open https://developer.microsoft.com/en-us/fabric#/controls/web/panel in Edge.
2. Navigate to 'Panel' control by using Tab key and activate it.
3. Dialogue pane will be appeared.
4. Navigate to 'Controlled Visibility' variant by using Tab key and activate it.
5. Press Tab key and focus will be on 'Close' button.
6. Now verify the issue after activating it by 'Enter' key and by using mouse also.

**Actual Result:**
After activating 'Close' button, no action performed. We are not able to close that window after activating the 'Close' button. We are able to close that pane after activating the 'Close Panel' button.

**Expected Result:**
After activating the 'Close'  button, 'Controlled Panel' pane should be closed.

**Attachment:**
No action is getting preformed after activating the 'Close' button
[Functional_No action is getting preformed after activating the 'Close' button either by using keyboard or by mouse..zip](https://github.com/OfficeDev/office-ui-fabric-react/files/3711976/Functional_No.action.is.getting.preformed.after.activating.the.Close.button.either.by.using.keyboard.or.by.mouse.zip)

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10781 

#### Description of changes

This change was initially meant to address the issue brought up in #10781 about an example with a close button which does nothing, but when I started thinking about that example and the others, there were a lot more issues which seemed worth addressing.
- Add a basic example
- Combine redundant sizing examples into one example with a size selector
- Add explanatory text for all the examples where the purpose isn't obvious
- Improve panel content in some examples to better demonstrate concepts
- Convert to function components
- Hide the close button in the controlled example

Some related "infrastructure" updates:
- Make the API reference table render enum descriptions as multiline, since some enums (such as PanelType) have long descriptions which are important to let people read.
- Exclude panel examples from snapshot testing (and delete the snapshot files)--the panel isn't open at the time the snapshot is taken, so there's no point in having a bunch of "panel" snapshots which actually just contain buttons and text.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10928)


packages/office-ui-fabric-react/src/components/Panel/examples/Panel.Sizes.Example.tsx
```diff
@@ -0,0 +1,70 @@
...
+        type={panelType}
+        customWidth={panelType === PanelType.custom || panelType === PanelType.customNear ? '888px' : undefined}
+        closeButtonAriaLabel="Close"
+        headerText="Sample panel"
+      >
+        <p>
+          This is {a} <strong>{description}</strong> panel{panelType === PanelType.smallFixedFar ? ' (the default size)' : ''}.
+        </p>
+        <p>
+          Select this size using <code>{`type={PanelType.${PanelType[panelType]}}`}</code>.
```
aneeshack4: Do we usually show code visibly like this in examples?

packages/office-ui-fabric-react/src/components/Panel/examples/Panel.LightDismissCustom.Example.tsx
```diff
@@ -1,74 +1,54 @@
...
-  };
-
-  private _closeDialogAndHidePanel = () => {
-    this._hidePanel();
-    this._closeDialog();
-  };
-}
+import { useConstCallback } from '@uifabric/react-hooks';
+
+const explanation = 'When the user attempts to close this panel, a confirmation dialog will appear.';
```
aneeshack4: I tried out this example on the PR deploy link - I don't see a confirmation dialog... maybe I'm missing something what should it look like?

packages/office-ui-fabric-react/src/components/Panel/examples/Panel.Sizes.Example.tsx
ecraig12345: Not typically, but I thought it would be helpful in this case.

packages/office-ui-fabric-react/src/components/Panel/examples/Panel.LightDismissCustom.Example.tsx
ecraig12345: Oh I might have broken that. I'll take a look.

ecraig12345: It only shows the confirmation dialog if you light dismiss. I'll update the description.

ecraig12345: Fixed!

packages/office-ui-fabric-react/src/components/Panel/examples/Panel.Sizes.Example.tsx
micahgodbolt: i wonder if it makes sense to simply hide the code of some examples, especially when the code isn't the thing you'd want to show

ecraig12345: You may be talking about different things. I think Aneesha was referring to showing literal `PanelType.whatever` in the UI. @micahgodbolt I'm not totally sure what you're referring to (I have a guess but if you could elaborate that's better).

micahgodbolt: oh i see what you mean :)

packages/office-ui-fabric-react/src/components/ComponentExamples.test.tsx
```diff
@@ -79,7 +79,9 @@ const excludedExampleFiles: string[] = [
   'Picker.CustomResult.Example.tsx',
   'ScrollablePane.Default.Example.tsx',
   'ScrollablePane.DetailsList.Example.tsx',
-  'SelectedPeopleList.Basic.Example.tsx'
+  'SelectedPeopleList.Basic.Example.tsx',
+  // Snapshots of these examples are worthless since the component isn't open by default
+  'Panel.'
```
aneeshack4: Confused by this line - why isn't it ".Examples.tsx" like the rest?

packages/example-app-base/src/components/ApiReferencesTable/ApiReferencesTable.tsx
```diff
@@ -249,6 +249,7 @@ export class ApiReferencesTable extends React.Component<IApiReferencesTableProps
         maxWidth: 400,
         isCollapsible: false,
         isResizable: true,
+        isMultiline: true,
```
aneeshack4: I don't see this being used anywhere else in this PR - am I missing something?

packages/office-ui-fabric-react/src/components/ComponentExamples.test.tsx
ecraig12345: It's so this will match all panel examples, not just one specific file

packages/example-app-base/src/components/ApiReferencesTable/ApiReferencesTable.tsx
ecraig12345: It's allowing enum value descriptions to be displayed as multi-line fields, motivated by the really long (and important) PanelType descriptions.

Today:


After this PR:

aneeshack4: Oh I see - should we perhaps have a multiline example then using this?

ecraig12345: Not sure what you mean? This is just a documentation component, not anything we have examples of.

aneeshack4: Ahh sorry my bad! Not sure why but yesterday when I was reviewing this, I think it looked like a type declaration to me instead of just the setting.

PR #10934, opened 2019-10-22
10931 dropdown alert

### LINKED ISSUE #10931: Dropdown : Screen reader does not announce the error message is displaying after activating the 'Show error message' control
**User Experience:**
Screen reader users will be confused if screen reader will not read the error message which is appeared after activating the 'Show error message' control. 

**Test Environment:**
OS: Version: RS5_Prerelease(1903 OS Build 19005.1)
URL: https://developer.microsoft.com/en-us/fabric#/controls/web/dropdown
Edge Browser Version:
Microsoft Edge: 44.19005.1.0
Microsoft EdgeHTML: 18.19005
Chrome Browser Version:
Chrome: 77.0.3865.120

**Usages:**
Dropdown with error message

**Repro Steps:**
1.https://developer.microsoft.com/en-us/fabric#/controls/web/dropdown in Edge and turn on Narrator.
2.Navigate to 'Basic input' button and activate it.
3.Navigate to 'Dropdown' link and activate it.
4.'Dropdown' pane will be appeared.
5.Navigate to 'Show error message' control and activate it.
6.Error message will be appeared and than verify the issue.

**Actual Result:**
Screen reader does not announce the error message after activating the 'show error message' control. Screen reader does not announce "This dropdown has an error".

**Expected Result:**
Screen reader should read the error message "This dropdown has an error" after activating the control.

**MAS Reference:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/Doc.aspx?sourcedoc=%7B684573A7-B089-4131-9B39-0009054125D3%7D&file=MAS%204.1.3%20%E2%80%93%20Status%20Messages.docx&action=default&mobileredirect=true&cid=3ff80ac8-a17c-49c5-893b-34a0390e615b

**Attachment:**
image
[OneDrive_2019-10-22.zip](https://github.com/OfficeDev/office-ui-fabric-react/files/3754201/OneDrive_2019-10-22.zip)

jdhuntington:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #10931

#### Description of changes

Set role=alert on error messages in `Dropdown`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10934)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-22-11-35-07-10931-dropdown-alert.json
+++ b/change/office-ui-fabric-react-2019-10-22-11-35-07-10931-dropdown-alert.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Dropdown: error message should have role=alert",
+  "packageName": "office-ui-fabric-react",
+  "email": "jdh@microsoft.com",
+  "commit": "698a3db6767fdaab67d68152af093b3e5f7d21a2",
+  "date": "2019-10-22T18:35:07.685Z",
+  "file": "/Users/jdh/projects/office-ui-fabric-react/change/office-ui-fabric-react-2019-10-22-11-35-07-10931-dropdown-alert.json"
+}
--- a/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
@@ -307,7 +307,7 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
         </KeytipData>
         {isOpen && onRenderContainer({ ...props, onDismiss: this._onDismiss }, this._onRenderContainer)}
         {hasErrorMessage && (
-          <div id={errorMessageId} className={this._classNames.errorMessage}>
+          <div role="alert" id={errorMessageId} className={this._classNames.errorMessage}>
             {errorMessage}
           </div>
         )}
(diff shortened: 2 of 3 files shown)


jdhuntington:
@cliffkoh Good questions.

Looking at the docs of the announced component, it doesn't feel like the goals of announced are aligned with error messages in forms. I have no problem integrating it, but announced seems very targeted at long-running operations.

Setting `role=alert` is the same behavior we currently expose in `TextField`, so I expect it to feel familiar to consumers of Fabric. To that end, I suspect a "validate" button that brings up 4 form errors would be an unpleasant experience for both users of screen readers as well as those without screen readers.

cliffkoh:
I don't have a problem with this approach if it is consistent with how the similar scenario is handled in Textfield, thanks for looking into that. On my 2nd question though, @dzearing has expressed concerns about that before and might have some thoughts. You might possibly want to run this through him for a quick sanity check.

micahgodbolt:
There seem to be a few different ways this is done in OUFR right now. Here's the comobobox example:

```jsx
<div role="region" aria-live="polite" aria-atomic="true" id={errorMessageId} className={this._classNames.errorMessage}>
          {errorMessage !== undefined ? errorMessage : ''}
</div>
```

https://github.com/OfficeDev/office-ui-fabric-react/blob/7073ea62ada078d5727734a759b874b07b149121/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx#L444


and Textfield (which is reused in DatePicker etc)

```jsx
<div role="alert">
  <DelayedRender>
    <p className={this._classNames.errorMessage}>
      <span data-automation-id="error-message">{errorMessage}</span>
    </p>
  </DelayedRender>
</div>
```


https://github.com/OfficeDev/office-ui-fabric-react/blob/d48f6dad41a8939d81d823fc94e61739eb7d1f5a/packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx#L231

jdhuntington:
Updated to include a test. Opted for an explicit test rather than a snapshot, as I think it makes the intent clearer and harder to break in the future.

PR #10940, opened 2019-10-23
Facepile: Add missing aria label to facepile list

### LINKED ISSUE #10877: Facepile Keros issue - ARIA input fields have an accessible name
### Environment Information

- **Package version(s)**: 6.204.4
- **Browser**: Chrome

### Describe the issue:
Keros Fast pass reports an issue for Facepile component - 
ARIA input fields have an accessible name

Refer the below image:
image

The issue can be found on - 
https://developer.microsoft.com/en-us/fabric#/controls/web/facepile

#### Actual behavior:
Keros reports an issue - ARIA input fields have an accessible name

#### Expected behavior:
There should be no issue reported by Keros

marygans:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #10877

#### Description of changes

This change addresses Facepile accessibility insights issue: "aria-input-field-name: ARIA input fields have an accessible name[ WCAG 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)"

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10940)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Facepile/Facepile.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Facepile/Facepile.base.tsx
@@ -35,7 +35,7 @@ export class FacepileBase extends BaseComponent<IFacepileProps, {}> {
   }
 
   public render(): JSX.Element {
-    let { overflowButtonProps } = this.props;
+    let { overflowButtonProps, ariaLabel } = this.props;
     const { chevronButtonProps, maxDisplayablePersonas, personas, overflowPersonas, showAddButton } = this.props;
 
     const { _classNames } = this;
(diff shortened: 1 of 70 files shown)


packages/office-ui-fabric-react/src/components/Facepile/Facepile.base.tsx
```diff
@@ -58,7 +58,7 @@ export class FacepileBase extends BaseComponent<IFacepileProps, {}> {
         {this.onRenderAriaDescription()}
         <div className={_classNames.itemContainer}>
           {showAddButton ? this._getAddNewElement() : null}
-          <ul className={_classNames.members} role="listbox">
+          <ul className={_classNames.members} role="listbox" aria-label={'Facepile persona list'}>
```
mtennoe: >Facepile persona list' [](start = 74, length = 22)

This needs to be a localized string, so it needs to be provided by the consumer of the component

marygans: you're right @mtennoe,  thanks!

aneeshack4:
@marygans Looks like Identifier 'ariaLabel' is never reassigned; use 'const' instead of 'let'

PR #10943, opened 2019-10-23
Added TeachingBanner Component

janjachacz:
#### Pull request checklist

#### Description of changes

Added a new TeachingBanner Component

https://www.figma.com/file/BAfsTrZ9ehIYZ0oYuyskUN2R/Web-Fluent?node-id=13858%3A47817

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10943)

### DIFF
--- a/change/@uifabric-experiments-2019-10-22-00-02-37-users-v-jajach-feature-teaching-banner.json
+++ b/change/@uifabric-experiments-2019-10-22-00-02-37-users-v-jajach-feature-teaching-banner.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "New experimental Teaching Banner Component",
+  "packageName": "@uifabric/experiments",
+  "email": "jan.jachacz@gmail.com",
+  "commit": "a6348ae648fbd5d1a0b91927857ddee106761200",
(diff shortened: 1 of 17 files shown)


micahgodbolt:
my concern with this control is that there is so much overlap with message bar. as an experiment, that's fine, but i'd want to better rationalize how this could simply be a message bar variant.

v-jajach:
@micahgodbolt, I guess you're right.

At the time of scoping this, `messageBarIconProps` was not in `MessageBar`'s source code yet. My understanding was to create a new component to match the `TeachingBubble`'s and `Coachmark` _collection_. 

I'll discuss this with my team. 

Thanks for your return

micahgodbolt:
@v-jajach any decision here?

v-jajach:
@micahgodbolt, yes finally we'll leverage the MessageBar component. 
Just need to add the task in one of our sprints.

micahgodbolt:
Thanks @v-jajach i'll close down this PR then.

PR #10954, opened 2019-10-23
Fixes: basebutton not rendering the aria label in the outer div

kenotron:
#### Pull request checklist

#### Description of changes

This addresses an issue with the basebutton not rendering the aria label in the outer div

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10954)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-23-13-51-25-master.json
+++ b/change/office-ui-fabric-react-2019-10-23-13-51-25-master.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Adds aria label to basebutton on the outer div",
+  "packageName": "office-ui-fabric-react",
+  "email": "kchau@microsoft.com",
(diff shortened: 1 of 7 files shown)


kenotron:
@SavannahBourgeois - this is the fix, it should contain a link to the deployed site soon.

SavannahBourgeois:
Looks great! Thanks for the quick fix @kenotron ! Do you think you have enough split button test examples to cover this? My only recommendation would be to consider adding a split button with no label under your split button examples since that now works properly with your change.

kenotron:
Everything seems a-ok. The example looks like it's doing the same thing as yours... I actually realized that we had removed the aria-label in the last snapshot but we accepted it anyway!

PR #10956, opened 2019-10-23
Modal: update styles

maxwellred:
#### Pull request checklist

#### Description of changes

Updating Modal styles per new spec from Chris Wymer in Sketch toolkit
- Added MIN/MAX widths/heights to the Modal plus spacing is 16px(screenshots below)
- Changed some header stylings ("X" button instead of close button, top blue border instead of entire header being blue, title size and color) in the Modal.Example.scss file simply to keep whats shown on the site identical to the toolkit.
- [Here's the link](https://app.goabstract.com/projects/82c13930-e212-11e8-a883-231e20d982e7/branches/5842cf9d-814c-44fe-821a-27f367a4482e/files/BAE218DE-4BBA-462C-8C8E-5D3E15116F78/pages/E88F0CCB-E98D-41D3-A8C4-244DE16F77A3?filter=changed) to the toolkit but you might not have permissions

#### BEFORE (Header style change)
 - note: the header styling is within the example styling
Screen Shot 2019-10-23 at 3 49 30 PM

#### AFTER (Header style change)
Screen Shot 2019-10-23 at 3 49 00 PM

#### Modal MAX-width/height
- note: screenshot doesn't show the edge radius well but it is there
Screen Shot 2019-10-23 at 3 48 52 PM

#### Modal MIN-width/height
Screen Shot 2019-10-23 at 3 47 55 PM

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10956)


maxwellred:
Hi @JasonGore and @marygans I've sent Chris Wymer (designer) the deployed link of this feature and he gives it the greenlight. So when you're ready you can have a look and let me know if there's any tweaks you'd like to be put in as well, thanks!

packages/office-ui-fabric-react/src/components/Modal/Modal.styles.ts
```diff
@@ -67,9 +68,18 @@ export const getStyles = (props: IModalStyleProps): IModalStyles => {
...
         outline: '3px solid transparent',
-        maxHeight: '100%',
+        maxHeight: 'calc(100% - 32px)',
+        maxWidth: 'calc(100% - 32px)',
+        minHeight: '176px',
+        minWidth: '288px',
         overflowY: 'auto',
-        zIndex: isModeless ? ZIndexes.Layer : undefined
+        zIndex: isModeless ? ZIndexes.Layer : undefined,
+        selectors: {
```
micahgodbolt: i'm fine with the other additions, though i could see this media query styling causing issues for modal users, as it's higher specificity, and more difficult to override. Is there a reason these more opinionated styles aren't on the dialog vs modifying this base control?

maxwellred: @micahgodbolt Actually, nicely enough, there are styles I put in DialogContent.styles that cover this base. Just spoke with Chris and he said having 16px all around would be fine so I have kept that styling and removed the breakpoints.

packages/office-ui-fabric-react/src/components/Modal/Modal.styles.ts
```diff
@@ -67,7 +67,10 @@ export const getStyles = (props: IModalStyleProps): IModalStyles => {
         position: 'relative',
         textAlign: 'left',
         outline: '3px solid transparent',
-        maxHeight: '100%',
+        maxHeight: 'calc(100% - 32px)',
+        maxWidth: 'calc(100% - 32px)',
```
micahgodbolt: would vh be a better value here, rather than calc.

maxwellred: The calc is to act as a constant max space between the the Modal and the edge of the viewport (16px).

maxwellred:
Hi @JasonGore I was hoping you could approve these changes when you get a chance, thanks!

packages/office-ui-fabric-react/src/components/Modal/examples/Modal.Basic.Example.tsx
```diff
@@ -30,19 +30,42 @@ const styles = mergeStyleSets({
...
+
+const iconButtonStyles = mergeStyleSets({
+  root: {
+    color: theme.palette.neutralPrimary,
+    marginLeft: 'auto',
+    marginTop: '4px',
+    marginRight: '2px'
+  },
+  rootHovered: {
+    color: theme.palette.neutralDark
```
JasonGore: 👍 Thanks for doing the changes!

packages/office-ui-fabric-react/src/components/Modal/examples/Modal.Basic.Example.tsx
```diff
@@ -64,22 +87,32 @@ export class ModalBasicExample extends React.Component<{}, IModalBasicExampleSta
...
           subtitleAriaId={this._subtitleId}
           isOpen={this.state.showModal}
           onDismiss={this._closeModal}
           isBlocking={false}
-          containerClassName={styles.container}
+          containerClassName={modalStyles.container}
           dragOptions={this.state.isDraggable ? this._dragOptions : undefined}
         >
-          <div className={styles.header}>
+          <div className={modalStyles.header}>
```
JasonGore: This is still a bit weird. It's called `modalStyles` but it's not used for the `Modal`? Bit of a nitpick, but maybe it would be better as `styles` or `contentStyles` to minimize confusion.

maxwellred: You bet!

PR #10967, opened 2019-10-25
Option to use checkbox role in Toggle component

### LINKED ISSUE #10966: Toggle state is not read by screen readers in IE 11.
### Environment Information

- **Package version(s)**:  master
- **Browser and OS versions**: IE 11, Windows 10

### Describe the issue:

Toggle state (on, off, changes in state) is not read by screen readers in IE 11.

https://developer.microsoft.com/en-us/fabric#/controls/web/toggle
To repro, open the fabric controls demo page in IE 11 and use Narrator while navigating to the toggle buttons. If you change the role to "checkbox" in DOM, Narrator will announce state 

#### Actual behavior:

Narrator (or NVDA) won't announce the state of the toggle button or the changes in state. 

#### Expected behavior:

Narrator (or NVDA) should announce the state of the toggle button or the changes in state. 

Toggles are not accessible if the user doesn't know if it's on or off.
https://www.w3.org/TR/wai-aria-1.1/#switch

sabintim:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #10966 

#### Description of changes

Since the `switch` ARIA role is not supported by all user agents (e.g, IE 11), this PR provides an option to use the `checkbox` role for backwards compatibility.

### DIFF
--- a/change/office-ui-fabric-react-2019-10-24-17-56-34-sabin-toggle-compat.json
+++ b/change/office-ui-fabric-react-2019-10-24-17-56-34-sabin-toggle-compat.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Option to use checkbox role in Toggle component",
+  "packageName": "office-ui-fabric-react",
+  "email": "satimals@microsoft.com",
+  "commit": "db4a33a8a686d180db5f64e7417b0ea279d9fdd7",
+  "date": "2019-10-25T00:56:34.058Z"
+}
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -7674,6 +7674,7 @@ export interface IToggle {
 // @public
 export interface IToggleProps extends React.HTMLAttributes<HTMLElement> {
     ariaLabel?: string;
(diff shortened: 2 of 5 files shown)


kenotron:
Thanks for the contrib, I think a better approach here is to have the `{...toggleNativeProps}` moved to the bottom (so incoming props can override the `role` prop). This way, if you wanted to change the role to "checkbox" you can do it like this:

```jsx
<Toggle role="checkbox" />
```

sabintim:
Interesting, I hadn't thought about that. I'll modify

kenotron:
Now, that's a bigger behavioral change. If you want to be conservative, you could do a one off override of the "role" prop :)

sabintim:
Does this sound about right? If "toggleNativeProps" has a role, use that, otherwise keep current pattern

kenotron:
> Does this sound about right? If "toggleNativeProps" has a role, use that, otherwise keep current pattern

Yeah I think so, that's safest - my original suggestion might change a LOT of behaviors

sabintim:
@kenotron I found adding "role" as a component property is easier to ensure type safety. Updated PR.

sabintim:
I still need to modify a test to include the new button in Toggle examples

sabintim:
@kenotron does the PR look good now? I also need to apply this to Fabric 6.

PR #10972, opened 2019-10-26
Removing circular import for IPoint in positioning.types.d.ts

mehanig:
Looks like this file is failing with TypeScript3.7-rc  because of 
ERROR in ..office-ui-fabric-react/lib/utilities/positioning/positioning.types.d.ts(2,10):
TS2440: Import declaration conflicts with local declaration of 'IPoint'.

#### Pull request checklist

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10972)

### DIFF
--- a/packages/office-ui-fabric-react/src/utilities/positioning/positioning.types.ts
+++ b/packages/office-ui-fabric-react/src/utilities/positioning/positioning.types.ts
@@ -1,5 +1,4 @@
 import { DirectionalHint } from '../../common/DirectionalHint';
-import { IPoint } from './positioning.types';
 import { IRectangle } from '../../Utilities';
 
 export enum RectangleEdge {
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/utilities/positioning/positioning.types.ts
```diff
@@ -1,5 +1,4 @@
 import { DirectionalHint } from '../../common/DirectionalHint';
-import { IPoint } from './positioning.types';
```
mehanig: This line is conflicting with this export https://github.com/OfficeDev/office-ui-fabric-react/pull/10972/files#diff-7bc6de4d67a6feec646ff9ce256a7f67R103

LeopoldLerch:
can this also be fixed in lower versions? especiall Version 5?

PR #10977, opened 2019-10-28
[Breaking change] Move @uifabric/example-app to devDependency and remove re-exports in office-ui-fabric-react

KevinTCoughlin:
#### Pull request checklist

#### Description of changes

The refactor of test related code into its own module `@uifabric/example-data` in https://github.com/OfficeDev/office-ui-fabric-react/pull/10280 has a downstream effect that devs specify `@uifabric/example-data` in their `package.json` when consuming Fabric 7. This can result in the module's code being bundled in production assets if not tree-shaken properly.

Although we avoid breaking changes in the same major release, I believe this case to be exceptional. If consumers require the exports from this dependency then they can rely on the package itself in their app. This will also benefit non-consumers of the module who may be pulling in the bits in their production assets.

#### Focus areas to test

- CI to pass.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10977)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-28-14-20-36-keco-remove-example-data-dep.json
+++ b/change/office-ui-fabric-react-2019-10-28-14-20-36-keco-remove-example-data-dep.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Remove re-export of @uifabric/example-data from office-ui-fabric-react",
+  "packageName": "office-ui-fabric-react",
+  "email": "KevinTCoughlin@users.noreply.github.com",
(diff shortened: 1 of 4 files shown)


KevinTCoughlin:
> Talked with Jon and now that we have another option for importing the example data, and he's okay with removing the old imports, I'm okay with it too.

Awesome, I'll let the bot merge this later today.

dzearing:
This is the right fix, even though this is a breaking change. Somehow some test code leaked into the public API. This should never have happened.

PR #10981, opened 2019-10-29
Update API doc generation to handle deprecated messages

ecraig12345:
#### Pull request checklist

#### Description of changes

I noticed that the API doc tables didn't display property-specific deprecation messages, and for some reason I decided to make fixing that into a weekend project.

**Before example** 

**After example**

Page JSON generation improvements:
- Include deprecated messages in the output
- Include deprecated information (and messages) for all item types--anything can be deprecated, not just interface/class members, and it's in our interest to ensure users know about all deprecated items
- Code de-duplication to make adding new info (like deprecated messages) easier
- Improve "rendering" (translation from doc nodes to markdown) of `@link` and code elements inside comments.
- Add the constructor to the list of methods in a class (and always render it first)

API table rendering improvements:
- Show the deprecated message (if provided) after the default "Warning: this API is now obsolete."
- Format property names and defaults as code for easier reading
- Switch the font for all code in the API tables to the nicer monospace font used elsewhere in the docs
- Render a default "(No properties)" message for classes and interfaces with no properties/methods, to avoid having a totally empty heading
- Always use `bodyText` for the table text (don't change color on hover)
- Un-escape `@<>{}` characters: the escapes are needed for API Extractor, but not in the UI

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10981)


packages/example-app-base/src/components/ApiReferencesTable/ApiReferencesTable.tsx
```diff
@@ -512,3 +441,47 @@ export class ApiReferencesTable extends React.Component<IApiReferencesTableProps
...
+      codeIndex++;
+      textIndex += codeBlock.text.length;
+    }
+    key++;
+  }
+  if (textIndex < text.length) {
+    textElements.push(text.substring(textIndex));
+  }
+
+  return textElements;
```
jdhuntington: Regex seems like the wrong tool here; I think it would be more efficient just to walk the text. 
I believe this does pretty much the right thing:

```
const _extractCodeBlocks = (text: string): React.ReactNode[] => {
  const result: React.ReactNode[] = [];
  if (!text || text.length === 0) {
    return result;
  }
  let index = 0;
  let inCodeBlock = false;
  while (index < text.length) {
    let sectionEnd = text.indexOf("`", index);
    if (sectionEnd === -1) sectionEnd = text.length;
    const elementType = inCodeBlock ? "code" : "span";
    result.push(<elementType key={index}>{text.substring(index, sectionEnd).trim()}</elementType>
    inCodeBlock = !inCodeBlock;
    index = sectionEnd + 1;
  }
  return result;
};
```

packages/example-app-base/src/components/ApiReferencesTable/ApiReferencesTable.tsx
```diff
@@ -457,7 +386,7 @@ export class ApiReferencesTable extends React.Component<IApiReferencesTableProps
     return undefined;
   }
 
-  private _parseILinkTokensHelper = (linkTokens: ILinkToken[], extend: boolean): JSX.Element | undefined => {
```
jdhuntington: yay for getting rid of boolean params!

packages/example-app-base/src/components/ApiReferencesTable/ApiReferencesTable.tsx
```diff
@@ -14,61 +15,82 @@ import { SelectionMode } from 'office-ui-fabric-react/lib/Selection';
...
+      }
+    }
+  },
+  isMultiline: {
+    wordBreak: 'break-word'
+  }
+};
 
+// When the text is passed to the function using this regex, it has had newline characters removed,
+// so the regex will match backtick sequences that span multiple lines.
```
jdhuntington: This comment is confusing; how are there multiple lines if there are no newline characters?

packages/api-docs/src/PageJsonGenerator.tsx
```diff
@@ -182,98 +194,141 @@ function extractText(node: DocNode): string {
...
+    }
+  }
+  return '';
+}
 
+/**
+ * Generate an ITableJson for any top-level API item (interface, class, enum, type alias)
+ * with the name, description, deprecated message, and optionally extends tokens pre-filled.
+ */
+function getTableJson(
```
jdhuntington: This function (and a few more in this file that are similar) are fine, but they contain enough logic that I wonder if `ITableJson` and `ITableRowJson` would be better implemented as full-on classes instead of interfaces. Then getters could be defined for things like `deprecated` and the logic for that bit would be obvious.

packages/example-app-base/src/components/ApiReferencesTable/ApiReferencesTable.tsx
ecraig12345: I think it means sequences that originally spanned multiple lines? (I didn't write this, just moved it)

packages/example-app-base/src/components/ApiReferencesTable/ApiReferencesTable.tsx
ecraig12345: The regex logic was just moved from another part of the file. But I could change to this since I'm guessing it would be more efficient.

packages/api-docs/src/PageJsonGenerator.tsx
ecraig12345: Interesting idea. I'm not going to change it now but I'll keep that in mind.

PR #10987, opened 2019-10-29
Update office-fabric-ui-core dev dependency

maxwellred:
#### Pull request checklist

- [ ] Addresses an existing request in email to update this dep

#### Description of changes

Updates the office-ui-fabric-core dev dependency in the packages/office-ui-fabric-react/package.json from version 10 to version 11 adding ~ 350 new icons

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10987)


ecraig12345:
Is there any way this upgrade could possibly cause a breaking change for consumers? (Not in the sense of a code break probably, but some unexpected styling or bundle size impact.)

jdhuntington:
@maxwellred Any chance you can find numbers of how much it increases the bundle size? Would be nice to communicate that proactively.

maxwellred:
@jdhuntington Sorry for the delay. It looks like v11.0.0 is 411.42 kB and v10.0.0 is 369.17 kB so an increase in 42.25 kBs

dzearing:
Hmm. I'm pretty confused by questions and answers here. This increases the fabric core css, which does not impact the bundle size of fabric react or fabric icons.

If you include fabric core css in your page, then yes, the css would increase the page size. No bundles are created from the css though.

maxwellred:
Hi @dzearing So I think there's two mistakes on my part that can clear all this up. 1) I mistook package size for bundle size in my latest comment 2) I misquoted Peter (double checked with him now) and this wont affect bundle size (as you have stated in your comment). Sorry everyone!

ecraig12345:
That makes sense now. One other issue I noticed though is that you should also update the core version in fabric-website, fabric-website-resources, and example-app-base.

ecraig12345:
And related to @KevinTCoughlin's comment in https://github.com/OfficeDev/office-ui-fabric-react/issues/10449#issuecomment-558827876, could you please update:
1. The fabric core getting started link in FabricIconsUsage.md -- it should be `#/get-started/web#fabric-core`
2. The current version numbers in GetStartedDevelopCore.md under "Adding Fabric Core to your site"

ecraig12345:
Oh and in Fabric7Overview.md, you should probably update the part where it mentions version 10. In that context it's probably better to say "10 and later" so something like this:
> For apps built with Fabric Core, versions 10 and later include the new Fluent colors, type sizes, animations, and shadows.

ecraig12345:
Looks like you need change files for the additional packages.

PR #10988, opened 2019-10-30
Fix index import for Stack in TeachingBubbleContent

KevinTCoughlin:
#### Pull request checklist

#### Description of changes

This `office-ui-fabric-react` root import is causing a downstream partner's AMD bundle size to increase by ~160kB. 

#### Focus areas to test

- CI to pass

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10988)


ecraig12345:
By "lint rule" I mean another condition in the lint-imports script probably

KevinTCoughlin:
> If this is such a problem for AMD bundle size, we should look into adding a lint rule to prevent it from happening again. (Auto-added imports will often be in this format.)

I agree. It would be extremely useful for AMD consumers. I don't have cycles to do the work until I finish internal upgrade work, but will consider it after that. I'm also willing to review a PR that adds such a rule :).

PR #10994, opened 2019-10-30
10936 toggle wrapping

### LINKED ISSUE #10936: Toggle content can render outside of its bounds if label is unbreakable and long
### Environment Information

- **Package version(s)**: "office-ui-fabric-react": "7.6.0"
- **Browser and OS versions**: Chrome Version 77.0.3865.120 (Official Build) (64-bit) - Windows 10 Enterprise 64-bit (10.0, Build 18363) (18362.19h1_release.190318-1202)

https://codepen.io/dPJinjo/pen/QWWpyGX

CSS
```
.skinny {
  width: 240px;
  background-color: cyan;
}
```

JS
```
let { Toggle } = window.Fabric;

const Content = () =>   
  <Toggle className="skinny" label="thisisareallylonglabelwithnobreakswhichforcesthetoggleout" onText="on" offText="off" inlineLabel defaultChecked />;

ReactDOM.render(
  <Content />, 
  document.getElementById("content")
);
```

Result
image

#### Actual behavior:

It's possible to force parts of a Toggle outside of the bounds of the Toggle if the label is a very long unbreakable string

#### Expected behavior:

All Toggle contents should be rendered within the Toggle bounds and not overflow off of the end.

Requested priority: (Normal)

Products/sites affected:

jdhuntington:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #10936

#### Description of changes

Wraps long labels using `overflow: hidden;` (causing the existing word-wrap to take effect, no text is actually hidden.)

Before:
Screen Shot 2019-10-30 at 3 29 39 PM

After:
Screen Shot 2019-10-30 at 3 29 29 PM

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10994)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Toggle/Toggle.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Toggle/Toggle.styles.ts
@@ -60,7 +60,8 @@ export const getStyles = (props: IToggleStyleProps): IToggleStyles => {
         inlineLabel && {
           order: 1,
           marginLeft: 16
-        }
+        },
+      inlineLabel && { overflow: 'hidden' }
     ],
 
     container: [
--- a/packages/office-ui-fabric-react/src/components/Toggle/Toggle.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Toggle/Toggle.styles.ts
@@ -61,7 +61,7 @@ export const getStyles = (props: IToggleStyleProps): IToggleStyles => {
           order: 1,
           marginLeft: 16
         },
-      inlineLabel && { overflow: 'hidden' }
+      inlineLabel && { wordBreak: 'break-all' }
     ],
 
     container: [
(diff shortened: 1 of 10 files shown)


packages/office-ui-fabric-react/src/components/Toggle/Toggle.styles.ts
```diff
@@ -60,7 +60,8 @@ export const getStyles = (props: IToggleStyleProps): IToggleStyles => {
         inlineLabel && {
           order: 1,
           marginLeft: 16
-        }
+        },
+      inlineLabel && { overflow: 'hidden' }
```
micahgodbolt: shouldn't this be word-break instead? is there a reason to use overflow instead?

https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

jdhuntington: Yeah, that seems to work too. I'll switch it out.

PR #10996, opened 2019-10-30
Adds a hotfix pipeline definition

kenotron:
#### Description of changes

This adds a pipeline definition for hotfixes - this hotfix can be used to generate a tarball drop which will be deployed if approved in the release pipeline

*Using this build pipeline requires a variable*

If `targetNpmVersion` isn't set, it will fail:
https://uifabric.visualstudio.com/UI%20Fabric/_build/results?buildId=35924

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10996)

### DIFF
--- a/azure-pipelines.hotfix.yml
+++ b/azure-pipelines.hotfix.yml
@@ -23,7 +23,7 @@ steps:
 
   - script: |
       npm version $(targetNpmVersion)
-    workingDir: 'packages/office-ui-fabric-react'
+    workingDirectory: 'packages/office-ui-fabric-react'
     displayName: 'Manually set fabric version'
 
   - script: |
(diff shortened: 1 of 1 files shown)


azure-pipelines.hotfix.yml
```diff
@@ -0,0 +1,54 @@
...
+  - script: |
+      yarn buildto office-ui-fabric-react --production
+    displayName: yarn build to OUFR (Create production build)
+
+  - script: |
+      npm pack
+    displayName: 'Create tarball'
+    workingDirectory: 'packages/office-ui-fabric-react'
+
+  - task: nkdagility.variablehydration.variabledehydration-task.variabledehydration@0
```
KevinTCoughlin: Is this task used?

kenotron: I deleted it - seems like blindly copying the old one will yield strange results.

PR #10885, opened 2019-10-17
Fix dropdown textfield combobox spin button wcag issues

betrue-final-final:
These fixes addresses WCAG 2.1 non-text contrast guidelines https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast

I've updated the focus and pressed states of these components:

Textfield
Combobox
Dropdown
Spinbutton
Before and after
image
https://user-images.githubusercontent.com/20363037/64815800-42256d00-d55b-11e9-810e-57d0d9648a51.png

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10885)

PR #11035, opened 2019-10-31
Searchbox/Textfield: Fix high contrast borders when disabled

### LINKED ISSUE #10624: Searchbox/Textfield: No visual difference when disabled in high contrast mode
pulled from #7500

A standard `<input disabled />` is not visually different from an enabled one in high contrast mode (not something we broke in Fabric). Disabled labels look different, but the inputs (borders/background) and place holder text are identical.

image

TextField

image

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10624

Before:

65621535-cbd03400-df78-11e9-9a8b-5a19032b22e5
image

After:

image

image

image

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11035)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/TextField/TextField.styles.tsx
+++ b/packages/office-ui-fabric-react/src/components/TextField/TextField.styles.tsx
@@ -156,7 +156,12 @@ export function getStyles(props: ITextFieldStyleProps): ITextFieldStyles {
           width: '100%'
         },
         disabled && {
-          borderBottomColor: semanticColors.disabledBackground
+          borderBottomColor: semanticColors.disabledBackground,
+          selectors: {
+            [HighContrastSelector]: {
+              borderColor: 'GrayText'
+            }
+          }
         },
         !disabled && {
           selectors: {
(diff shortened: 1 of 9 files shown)


ecraig12345:
I think we should either integrate this into @betrue-final-final's PR #10885 or at least wait until that merges, to avoid making Ben deal with even more conflicts.

dzearing:
@micahgodbolt the above PR was merged. Can you rebase this and get it merged? Thank you!

micahgodbolt:
All set

packages/office-ui-fabric-react/src/components/TextField/TextField.styles.tsx
```diff
@@ -185,16 +190,7 @@ export function getStyles(props: ITextFieldStyleProps): ITextFieldStyles {
...
-        selectors: {
-          ':hover': {
-            selectors: {
-              [HighContrastSelector]: {
-                borderColor: 'Highlight'
-              }
-            }
-          }
-        }
+        position: 'relative'
```
ecraig12345: Why was this removed?

micahgodbolt: I moved it town to 206 inside of `!disabled` otherwise disabled controls get a hover border 
textfield

PR #11038, opened 2019-10-31
Combobox: hidden option fix

### LINKED ISSUE #11032: [ComboBox] Hidden options are visible using Up/Down keys for selection
### Environment Information

- **Browser and OS versions**: All browsers

<https://codepen.io/neeraj-khanna/pen/WNNXxXv>

#### Actual behavior:
Set focus on ComboBox and use Down key to change values. You will find 10 options (Option A to Option J).

Now expand ComboBox using mouse and you will see only 6 options (Option A, Option B, Option G, Option H, Option I, Option J) as rest 4 are hidden.

#### Expected behavior:
Hidden options should not be visible even when selection is made using Up/Down keys.

No

High

maxwellred:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #11032

#### Description of changes

Adds extra OR conditional clause, `option.hidden === true`, when determining the next selectable index in order to skip option items holding the hidden:true key value

#### Focus areas to test

Click Combobox Autofill input (Having set some options hidden: true)
Arrow up and down options
Any items that have been marked "hidden: true" should not be shown/skipped

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11038)

### DIFF
--- a/change/office-ui-fabric-react-2019-10-31-12-25-57-v-mare-Combobox-hidden-option-fix.json
+++ b/change/office-ui-fabric-react-2019-10-31-12-25-57-v-mare-Combobox-hidden-option-fix.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Combobox: Fixed issue where hidden options are shown when arrowing through combobox option list",
+  "packageName": "office-ui-fabric-react",
+  "email": "v-mare@microsoft.com",
+  "commit": "f2f8fc5b251c053fb2d9d1f212ec81e7e427840b",
+  "date": "2019-10-31T19:25:57.970Z"
+}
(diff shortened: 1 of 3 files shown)


maxwellred:
Just noticed that this also wasn't skipping disabled items as well so i added that conditional clause in as well

joschect:
@maxwellred I believe that disabled items should be focusable. That way a screen reader user can still get information about it. @betrue-final-final or @micahgodbolt might be able to chime in about it for sure.

micahgodbolt:
i believe the consensus was that disabled options in a list should be focusable, but not standalone elements.

maxwellred:
@micahgodbolt Sounds good! I'll remove the disabled conditional.

PR #11040, opened 2019-10-31
moved to webpack-cli instead of calling webpack directly via function

kenotron:
#### Description of changes

This PR will get us onto using webpack-cli instead (using a new just-scripts webpackCliTask). It adds back the deleted ts checker during development. I have made it so we don't do typechecking again in TF_BUILD situation (for minor speed gains)

From chatting with some folks, it seems that people DO use the typechecker for during inner loop development, so I'm adding that back for now.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11040)


kenotron:
Thx

Issue #6075, opened 2018-08-24
Remove CSS styling from examples


kkjeer:
**Describe the feature that you would like added**

Clean up the Fabric examples so that they use JS styling instead of CSS styling.

**What component or utility would this be added to**

Examples

**Have you discussed this feature with our team, and if so, who**

@micahgodbolt 

**Additional context/screenshots**

Users are having trouble modifying the examples to fit their own codebase when the examples use CSS styling, e.g. in #6072

cliffkoh:
This is going to be huge and probably requires splitting up the work across the team...

micahgodbolt:
yeah, this is certainly more of a meta issue. lets be sure to update as we make changes.

ecraig12345:
Reopening since this isn't fully fixed

JustSlone:
Closing this out as it's very old, mostly done, and examples will likely be updated soon anyway

PR #11044, opened 2019-11-01
Remove SCSS From Callout

pandasa123:
#### Pull request checklist

- [x] Addresses an existing issue: Part of #6075

#### Description of changes

- Codepen Export 

Steps taken:
- Removing SCSS file
- Added Callout Example Styles for all Callout examples using SCSS
- Updated snapshots
- Verified functionality

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11044)


packages/office-ui-fabric-react/src/components/Callout/Callout.doc.tsx
```diff
@@ -25,26 +34,31 @@ export const CalloutPageProps: IDocPageProps = {
     {
       title: 'Default Callout',
       code: CalloutBasicExampleCode,
+      codepenJS: CalloutBasicExampleCodepen,
```
ecraig12345: These aren't needed now (and will actually cause compile errors if you update from master)

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
```diff
@@ -1,8 +1,9 @@
 import * as React from 'react';
-import './CalloutExample.scss';
 import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
 import { Callout, DirectionalHint } from 'office-ui-fabric-react/lib/Callout';
 import { Dropdown, IDropdownOption } from 'office-ui-fabric-react/lib/Dropdown';
+import { getId, classNamesFunction, css } from 'office-ui-fabric-react/lib/Utilities';
+import { calloutExampleStyle, ICalloutExampleStyle } from './Callout.Examples.Styles';
```
ecraig12345: Putting the example styles in their own file still prevents export from working, even if it's a TS file rather than SCSS. Also, we typically recommend using `mergeStyleSets` for examples (rather than `classNamesFunction`) for simplicity. (Same comment applies to other examples.)

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
```diff
@@ -48,31 +55,34 @@ export class CalloutCoverExample extends React.Component<{}, ICalloutCoverExampl
     // Large beak will disable some position to avoid beak over the callout edge.
     return (
       <div className="ms-CalloutExample">
-        <div className="ms-CalloutExample-configArea">
+        <div className={css(classNames.msCalloutExampleConfigArea)}>
```
ecraig12345: If there's just one class name, you don't need to use the css helper (same comment applies to all example files)

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Directional.Example.tsx
```diff
@@ -31,8 +32,14 @@ const DIRECTION_OPTIONS = [
...
 ];
 
+const getClassNames = classNamesFunction<{}, ICalloutExampleStyle>();
+const classNames = getClassNames(calloutExampleStyle, {});
+
 export class CalloutDirectionalExample extends React.Component<{}, ICalloutDirectionalExampleState> {
   private _menuButtonElement: HTMLElement | null;
+  // Use getId() to ensure that the callout title ID is unique on the page.
+  // (It's also okay to use a plain string without getId() and manually ensure its uniqueness.)
+  private _titleId: string = getId('callout-label');
```
ecraig12345: What's the reasoning behind adding this ID? I see it being attached to an element but not referenced anywhere else.

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Examples.Styles.ts
```diff
@@ -0,0 +1,95 @@
+import { FontSizes, FontWeights, DefaultPalette } from 'office-ui-fabric-react/lib/Styling';
+import { IStyle } from '@uifabric/styling';
+
+interface ICalloutExampleStyle {
```
ecraig12345: These styles need to go in the same file as the example that uses them, even if that means duplication (though I don't believe all these styles are used for all examples). Also I'd prefer that you not make interfaces for example styles--when using `mergeStyleSets` as recommended in another comment, an interface isn't needed to get accurate typings, so effectively it just adds visual clutter to the file.

packages/office-ui-fabric-react/src/components/Callout/Callout.doc.tsx
pandasa123: Removed!

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
pandasa123: Got it. I think fixed it (following the example in the Basic file)

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
pandasa123: Fixed!

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Directional.Example.tsx
pandasa123: I saw this in other file and was confused but removed it!

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Examples.Styles.ts
pandasa123: Got it and updated to mergeStyleSets pattern!

pandasa123:
Updated to the `mergeStyleSets` pattern. Sorry about that!

pandasa123:
I just went through the errors on Azure DevOps and it's outputing `Error: Cannot find module '@uifabric/experiments'`. Any advice what I should do to fix this one?

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
```diff
@@ -24,6 +32,63 @@ const DIRECTION_OPTIONS = [
   { key: DirectionalHint.rightBottomEdge, text: 'Right Bottom Edge' }
 ];
 
+const theme = getTheme();
+const styles = mergeStyleSets({
+  buttonArea: {
+    verticalAlign: 'top',
+    display: 'inline-block',
+    textAlign: 'center'
```
ecraig12345: Was there a reason for getting rid of the `margin: 0 100px` and `min-width: 130px` from `.ms-CalloutCoverExample-buttonArea`?

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Directional.Example.tsx
```diff
@@ -27,6 +37,63 @@ const DIRECTION_OPTIONS = [
   { key: DirectionalHint.rightBottomEdge, text: 'Right Bottom Edge' }
 ];
 
+const theme = getTheme();
+const styles = mergeStyleSets({
+  buttonArea: {
+    verticalAlign: 'top',
```
ecraig12345: Was there a reason for getting rid of some of the styles from `.ms-CalloutExample-buttonArea`? Specifically:
```
    min-width: 130px;
    margin: 0 100px;
    height: 32px;
```

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
```diff
@@ -24,6 +32,63 @@ const DIRECTION_OPTIONS = [
...
+  },
+  subtext: [
+    theme.fonts.small,
+    {
+      margin: 0,
+      color: theme.palette.neutralPrimary,
+      fontWeight: FontWeights.semilight
+    }
+  ],
+  link: [
```
ecraig12345: I don't see `link` being used in this example or the Directional one.

Same with `buttons`.

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
```diff
@@ -46,20 +111,20 @@ export class CalloutCoverExample extends React.Component<{}, ICalloutCoverExampl
     // Large beak will disable some position to avoid beak over the callout edge.
     return (
       <div className="ms-CalloutExample">
```
ecraig12345: Can you also remove the remaining classes like this which weren't doing anything? (in all the files)

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
```diff
@@ -24,6 +32,63 @@ const DIRECTION_OPTIONS = [
...
+    textAlign: 'center'
+  },
+  configArea: {
+    minWidth: '300px',
+    display: 'inline-block'
+  },
+  callout: {
+    maxWidth: 300
+  },
+  calloutExampleButton: {
```
ecraig12345: Not seeing `calloutExampleButton` used in this example.

Same with `subtext`.

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
```diff
@@ -24,6 +32,63 @@ const DIRECTION_OPTIONS = [
...
+      margin: 0,
+      color: theme.palette.neutralPrimary,
+      fontWeight: FontWeights.semilight
+    }
+  ],
+  inner: {
+    height: '100%',
+    padding: '0 24px 20px'
+  },
+  actions: {
```
ecraig12345: Not seeing this used in any of the examples

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
pandasa123: Removed! I replaced some with fragments but I don't know if it's a recommended practice in this codebase?

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
pandasa123: Removed link and buttons

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
pandasa123: Removed!

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
pandasa123: Removed!

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Directional.Example.tsx
pandasa123: Re-added! I missed those styles somehow

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Cover.Example.tsx
pandasa123: Fixed!

PR #11045, opened 2019-11-01
Remove SCSS from ChoiceGroup Examples

pandasa123:
#### Pull request checklist

#### Description of changes

- Codepen Export

Steps taken:

- Removing SCSS file
- Added ChoiceGroup Example Styles for ChoiceGroup Custom Example using SCSS
- Updated snapshots
- Verified functionality

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11045)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroup.doc.tsx
+++ b/packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroup.doc.tsx
@@ -12,7 +12,9 @@ const ChoiceGroupBasicExampleCodepen = require('!@uifabric/codepen-loader!office
 const ChoiceGroupLabelExampleCode = require('!raw-loader!office-ui-fabric-react/src/components/ChoiceGroup/examples/ChoiceGroup.Label.Example.tsx') as string;
 const ChoiceGroupLabelExampleCodepen = require('!@uifabric/codepen-loader!office-ui-fabric-react/src/components/ChoiceGroup/examples/ChoiceGroup.Label.Example.tsx') as string;
 const ChoiceGroupCustomExampleCode = require('!raw-loader!office-ui-fabric-react/src/components/ChoiceGroup/examples/ChoiceGroup.Custom.Example.tsx') as string;
+const ChoiceGroupCustomExampleCodepen = require('!@uifabric/codepen-loader!office-ui-fabric-react/src/components/ChoiceGroup/examples/ChoiceGroup.Custom.Example.tsx') as string;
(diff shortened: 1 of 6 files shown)


packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroup.doc.tsx
```diff
@@ -12,7 +12,9 @@ const ChoiceGroupBasicExampleCodepen = require('!@uifabric/codepen-loader!office
 const ChoiceGroupLabelExampleCode = require('!raw-loader!office-ui-fabric-react/src/components/ChoiceGroup/examples/ChoiceGroup.Label.Example.tsx') as string;
 const ChoiceGroupLabelExampleCodepen = require('!@uifabric/codepen-loader!office-ui-fabric-react/src/components/ChoiceGroup/examples/ChoiceGroup.Label.Example.tsx') as string;
 const ChoiceGroupCustomExampleCode = require('!raw-loader!office-ui-fabric-react/src/components/ChoiceGroup/examples/ChoiceGroup.Custom.Example.tsx') as string;
+const ChoiceGroupCustomExampleCodepen = require('!@uifabric/codepen-loader!office-ui-fabric-react/src/components/ChoiceGroup/examples/ChoiceGroup.Custom.Example.tsx') as string;
```
ecraig12345: These imports aren't needed any more.

packages/office-ui-fabric-react/src/components/ChoiceGroup/examples/ChoiceGroup.Custom.Example.Styles.ts
```diff
@@ -0,0 +1,19 @@
+import { IStyle } from '@uifabric/styling';
+
+interface IChoiceGroupCustomExampleStyles {
+  root?: IStyle;
+  dropdown: IStyle;
+}
+
+const choiceGroupCustomExampleStyle: IChoiceGroupCustomExampleStyles = {
```
ecraig12345: Example styles should go in the same file as the example (otherwise export still won't work), and to simplify things we typically use just `mergeStyleSets` rather than the `getStyles/classNamesFunction` approach. In this case:

```
const classNames = mergeStyleSets({
  root: {
    display: 'flex',
    alignItems: 'baseline'
  },
  dropdown: {
    marginBottom: '0px',
    marginLeft: '5px'
  }
});
```

ecraig12345:
Wait, actually now that I look again, this SCSS was already removed sometime between the previous version of this PR and now. So I'm going to close this since the necessary change is already done.

PR #11046, opened 2019-11-01
Remove SCSS from DatePicker Examples

pandasa123:
#### Pull request checklist

#### Description of changes

- Codepen Export

Steps taken:
- Removed SCSS file
- Added DatePicker Example Styles for DatePicker examples that used SCSS
- Updated snapshots
- Verified functionality

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11046)


pandasa123:
Not sure why `yarn change` is failing since we do have a change file added (last commit)

ecraig12345:
Sorry that approval was on the wrong PR somehow...I had some additional comments on this one.

packages/office-ui-fabric-react/src/components/DatePicker/examples/DatePicker.Basic.Example.tsx
```diff
@@ -20,6 +19,17 @@ const DayPickerStrings: IDatePickerStrings = {
   closeButtonAriaLabel: 'Close date picker'
 };
 
+const styles = mergeStyleSets({
```
ecraig12345: Since both of these sets of styles are the same, it's better to use the same class for both of them and call it something generic (like `control` maybe). Same comment on the other examples.

Alternatively, you could go even simpler and just use `const controlClass = mergeStyles({...})` since there's only one class name.

pandasa123: As in just renaming `styles` to `controlClass`?

pandasa123: Got it! Will fix now

ecraig12345: That's actually kind of odd...typically `___Class` implies that it's *one* class name, not a set. It's clearer to stick with `styles` or `classNames` for sets of class names. I suggested the name `controlClass` assuming that you'd switch to `mergeStyles` which returns a single class (since there's not really a reason to use a set here).

PR #11067, opened 2019-11-04
CommandBar: hidden attribute fix

### LINKED ISSUE #11034: Fabric 7: Button hidden:true does not hide button in IE11
Please see: #10419

It still doesn't work with the latest Fabric 7 version, and no I can't provide a CodePen- CodePen does not work with IE11 - it would work with Chrome but that is again the point of this issue!

I am pretty sure this is a real bug, it all works fine in Chrome but NOT in IE11 (exactly the same code).

Did anyone ever look into this or did you just close #10419???

maxwellred:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #11034

#### Description of changes

Adding hidden:true to CommandBar item array items hides items. Also, I've confirmed this change works in Win10 IE11.

#### Before (New Item Button)
Screen Shot 2019-11-04 at 1 29 30 PM

#### Add Code (New Item Button)
Screen Shot 2019-11-04 at 1 30 54 PM

#### After (New Item Button)
Screen Shot 2019-11-04 at 1 29 08 PM

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11067)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
@@ -144,16 +144,18 @@ export class OverflowSetBase extends BaseComponent<IOverflowSetProps, {}> implem
   }
 
   private _onRenderItems = (items: IOverflowSetItemProps[]): JSX.Element[] => {
-    return items.map((item, i) => {
-      const wrapperDivProps: React.HTMLProps<HTMLDivElement> = {
-        className: this._classNames.item
-      };
-      return (
-        <div key={item.key} {...wrapperDivProps}>
-          {this.props.onRenderItem(item)}
-        </div>
-      );
-    });
+    return items
+      .filter(item => item.hidden !== true)
+      .map((item, i) => {
+        const wrapperDivProps: React.HTMLProps<HTMLDivElement> = {
+          className: this._classNames.item
+        };
+        return (
+          <div key={item.key} {...wrapperDivProps}>
+            {this.props.onRenderItem(item)}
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -144,16 +144,18 @@ export class OverflowSetBase extends BaseComponent<IOverflowSetProps, {}> implem
...
-        className: this._classNames.item
-      };
-      return (
-        <div key={item.key} {...wrapperDivProps}>
-          {this.props.onRenderItem(item)}
-        </div>
-      );
-    });
+    return items
+      .filter(item => item.hidden !== true)
```
maxwellred: Added .filter method to completely remove element from being rendered

micahgodbolt:
i'm not too keen on this change because:

1. hidden is not actually a type safe prop (that I can tell) of the command bar item. It works in codepen because we pass everything through, but hidden is not part of the existing interface. So we're adding functionality that isn't listed in the interface
2. even if we did add 'hidden' to the item, the expected behavior of hidden is that changing the `display` value of hidden should take priority, so we aren't following the HTML spec (though this isn't a HUGE deal considering we use hidden elsewhere

the better question would be, what is the user trying to do, and can we accomplish this without changing the control?

maxwellred:
Per our Teams conversation I'll close this PR. Thanks again for the help!

Issue #10952, opened 2019-10-23
Office UI Fabric site does not work well in High contrast


jgage176:
Switching to high contrast (I used the black one) the site had many accessibility issues, failing MS accessibility standards. Here are some of them I have seen and captured:

1. Pictures on left navigation for iOS and Android do not have enough color contrast:
 
![image](https://user-images.githubusercontent.com/29739628/67426580-3c299f80-f58f-11e9-9455-d183db68b08e.png)

2. Active and focus buttons on navigation have background and font color as black:
 
![image](https://user-images.githubusercontent.com/29739628/67426587-4055bd00-f58f-11e9-9f97-5d845b1e30f0.png)

3. When moving around to select a component from a menu on the left the focus is invisible:
 
![image](https://user-images.githubusercontent.com/29739628/67426598-451a7100-f58f-11e9-879f-9359443b4809.png)

Notice how the link at the bottom shows what I am currently selecting but there is no indication visually what I have selected.

4. On the individual component pages they have color contrast issues as well:
Both the active and the focus have font color and background

![image](https://user-images.githubusercontent.com/29739628/67426610-4c417f00-f58f-11e9-9218-050a0ea36f27.png)

5. The button color is white and the font color is white
 
![image](https://user-images.githubusercontent.com/29739628/67426619-519ec980-f58f-11e9-915c-600698b73077.png)

jdhuntington:
Thanks for the report. Will dig into this this week.

ecraig12345:
@jdhuntington Getting rid of the background gradient in high contrast might help fix some of these issues, and is something we should do regardless. The change is as follows (and you should probably do the same in `example-app-base` `_base.scss`)--the important part is the `@include high-contrast`.

```diff
diff --git a/apps/fabric-website/src/styles/_base.scss b/apps/fabric-website/src/styles/_base.scss
index 7e99df4ef..eeeeed78d 100644
--- a/apps/fabric-website/src/styles/_base.scss
+++ b/apps/fabric-website/src/styles/_base.scss
@@ -6,15 +6,21 @@
   html, body {
     -webkit-font-smoothing: antialiased;
     -webkit-tap-highlight-color: transparent;
-    background: $ms-color-gray10; // Match header/footer color for browsers with momentum scroll
+    // Match header/footer color for browsers with momentum scroll
+    background: $ms-color-gray10;
+    // Used to give the nav a white background and the body a gray background
[...]

PR #11070, opened 2019-11-04
10952 website high contrast

jdhuntington:
#### Pull request checklist

- [X] Addresses an existing issue: #10952 (Issues 2-5)

#### Description of changes

- Fix the left-hand navigation by omitting the gradient background
- Fix several buttons by fielding the high contrast selector

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11070)

### DIFF
--- a/apps/fabric-website/src/styles/_base.scss
+++ b/apps/fabric-website/src/styles/_base.scss
@@ -78,7 +78,7 @@
     @include ms-fontColor-white;
 
     @include high-contrast {
-      color: window;
+      color: Window;
     }
   }
 
--- a/apps/fabric-website/src/styles/_base.scss
+++ b/apps/fabric-website/src/styles/_base.scss
@@ -1,9 +1,11 @@
 @import '~office-ui-fabric-core/dist/sass/References';
 @import './variables';
+@import './mixins';
 
 // Root elements
 :global {
-  html, body {
+  html,
+  body {
     -webkit-font-smoothing: antialiased;
     -webkit-tap-highlight-color: transparent;
     background: $ms-color-gray10; // Match header/footer color for browsers with momentum scroll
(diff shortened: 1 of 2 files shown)


apps/fabric-website/src/styles/_base.scss
```diff
@@ -62,8 +73,14 @@
   }
 
   // Feedback section styles
-  .FeedbackList-button .ms-Button-label { // Feedback button color overridden by UHF
+  .FeedbackList-button .ms-Button-label {
+    // Feedback button color overridden by UHF
     @include ms-fontColor-white;
+
+    @include high-contrast {
+      // The gradient doesn't behave well in high contrast, so get rid of it.
```
ecraig12345: Different comment?

jdhuntington: good catch

apps/fabric-website/src/styles/_base.scss
```diff
@@ -62,8 +73,13 @@
   }
 
   // Feedback section styles
-  .FeedbackList-button .ms-Button-label { // Feedback button color overridden by UHF
+  .FeedbackList-button .ms-Button-label {
+    // Feedback button color overridden by UHF
     @include ms-fontColor-white;
+
+    @include high-contrast {
+      color: window;
```
maxwellred: @jdhuntington This is potentially super-nit but aren't system colors usually pascal case? If this doesn't cause an issue then disregard.

jdhuntington: Good call; that does appear to be correct. (TIL!)

PR #11083, opened 2019-11-05
Contextual menu preventdefault

### LINKED ISSUE #11077: Can't close contextual menu via KB (expected 'esc' to close the menu)
### Environment Information

- **Browser and OS versions**: chrome win10

### Describe the issue:

Can't close contextual menu with keyboard

#### Actual behavior:

go to fabric site https://developer.microsoft.com/en-us/fabric#/controls/web/contextualmenu 
using only the keyboard open a contextual menu, and try close it without selecting any option 

#### Expected behavior:

being able to close the menu using 'esc' same as popup component behavior

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11077

#### Description of changes

Contextual menu's keyHandler is firing 'preventDefault' on every click, which means that the menuButton now can't close with any key clicks.

I believe the preventDefault is unnecessary as the function is already stopping propagation. I understand this might cause unexpected outcome, so I'm pushing this out to see what tests/feedback have to say.

#### Focus areas to test

What keyboard keypress might have unintended consequence now that the default isn't prevented.

Here's the boolean causing preventDefault and stopProp to fire:

https://github.com/OfficeDev/office-ui-fabric-react/blob/63cde726a39c9ce5bc10be55da0c97fb9633c1a3/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx#L786-L788

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11083)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -838,7 +838,6 @@ export class ContextualMenuBase extends BaseComponent<IContextualMenuProps, ICon
 
     if (shouldHandleKey(ev)) {
       this._isFocusingPreviousElement = true;
-      ev.preventDefault();
       ev.stopPropagation();
       this.dismiss(ev, dismissAllMenus);
       handled = true;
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -838,7 +838,6 @@ export class ContextualMenuBase extends BaseComponent<IContextualMenuProps, ICon
 
     if (shouldHandleKey(ev)) {
       this._isFocusingPreviousElement = true;
-      ev.preventDefault();
```
joschect: I don't believe this is the right fix, it should still prevent default to prevent the key press from propagating up and out. For instance a contextualmenu in a panel would dismiss the panel as well.

PR #11088, opened 2019-11-05
Production App - make sure the fabric-website also builds production given the --production flag

kenotron:
#### Pull request checklist

#### Description of changes

Apparently, fabric-website has a vastly different webpack.config.js so that the one fix for all other packages wouldn't work for this when respecting the --production flag. This PR will add this to the fabric-website app as well.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11088)

### DIFF
--- a/apps/fabric-website/webpack.config.js
+++ b/apps/fabric-website/webpack.config.js
@@ -1,12 +1,14 @@
 // @ts-check
 
-module.exports = function(env) {
+module.exports = function(env, argv) {
   const path = require('path');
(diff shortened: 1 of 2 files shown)


kenotron:
I'm merging this because it's passed the build steps - just waiting on screener which will not be affected by this change.

PR #11089, opened 2019-11-06
Platform picker high contrast

### LINKED ISSUE #10952: Office UI Fabric site does not work well in High contrast

jdhuntington:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #10952

#### Description of changes

Adds high contrast support to platform picker.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11089)

### DIFF
--- a/packages/example-app-base/src/styles/_mixins.scss
+++ b/packages/example-app-base/src/styles/_mixins.scss
@@ -34,3 +34,11 @@
     pointer-events: none;
   }
 }
+
+// High contrast mode mixins
+// @todo: Should be defined in common base
+@mixin high-contrast {
+  @media screen and (-ms-high-contrast: active) {
+    @content;
+  }
+}
--- a/change/@uifabric-example-app-base-2019-11-05-17-20-24-platform-picker-high-contrast.json
+++ b/change/@uifabric-example-app-base-2019-11-05-17-20-24-platform-picker-high-contrast.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "PlatformPicker: High Contrast support",
+  "packageName": "@uifabric/example-app-base",
+  "email": "jdh@microsoft.com",
+  "commit": "64d918168ddc62de51799cb4f02f05b0d22df9b5",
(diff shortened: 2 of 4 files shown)


packages/example-app-base/src/styles/_mixins.scss
```diff
@@ -34,3 +34,11 @@
     pointer-events: none;
   }
 }
+
+// High contrast mode mixins
```
ecraig12345: Could you add a reference to `~office-ui-fabric-react/lib/common/_highContrast.scss` instead? (that may not be quite the right path formatting)

jdhuntington: That doesn't seem to work. Looks like `lib/common` has an already-javascript version, so can't be imported by the sass loader.

ecraig12345: Ah it's dist. Similar example: `@import '~office-ui-fabric-core/dist/sass/References';`

ecraig12345: Wait sorry that's an import from core...

ecraig12345: Interesting, for OUFR imports I see other places using both src and dist. So...pick one?
```
@import '~office-ui-fabric-react/src/common/common';
@import '~office-ui-fabric-react/dist/sass/References';
```

PR #11095, opened 2019-11-06
Remove SCSS from List Scrolling Example

pandasa123:
#### Pull request checklist

#### Description of changes
- Removed SCSS file
- Added `mergeStyleSets` pattern similar to #11044 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11095)


packages/office-ui-fabric-react/src/components/List/examples/List.Scrolling.Example.tsx
```diff
@@ -123,7 +162,7 @@ export class ListScrollingExample extends React.Component<IListScrollingExampleP
   private _onRenderCell = (item: IExampleItem, index: number): JSX.Element => {
     return (
       <div data-is-focusable={true}>
-        <div className={css(styles.itemContent, index % 2 === 0 ? styles.itemContentEven : styles.itemContentOdd)}>
+        <div className={index % 2 === 0 ? styles.itemContentEven : styles.itemContentOdd}>
```
KevinTCoughlin: Can even and odd styles be applied using pure CSS e.g. `nth-child(even)`? It would be more performance since it would offload the calculation to CSS layer and not JS runtime.

pandasa123: Got it! I'll refactor it using the selector arg

packages/office-ui-fabric-react/src/components/List/examples/List.Scrolling.Example.tsx
```diff
@@ -1,13 +1,46 @@
...
+    theme.fonts.medium,
+    normalize,
+    {
+      position: 'relative',
+      display: 'block',
+      borderLeft: '3px solid ' + theme.palette.themePrimary,
+      paddingLeft: 27
+    }
+  ],
+  selected: {
```
ecraig12345: I don't see `selected` being used in this example.

ecraig12345:
This looks okay to me in general. To @KevinTCoughlin's question, I don't remember any specific reason for not converting these examples--we probably just never got to it.

packages/office-ui-fabric-react/src/components/List/examples/List.Scrolling.Example.tsx
pandasa123: Fixed!

dzearing:
@KevinTCoughlin in this example all of the classnames are pre computed outside of the cell rendering, so I don't see any perf concern other than perhaps nth-child selectors but I don't think those are problematic.

KevinTCoughlin:
> @KevinTCoughlin in this example all of the classnames are pre computed outside of the cell rendering, so I don't see any perf concern other than perhaps nth-child selectors but I don't think those are problematic.

nth-child selectors in CSS are better re: perf than the run-time JS modulo solution so this is probably fine for example code.

ecraig12345:
> nth-child selectors in CSS are better re: perf than the run-time JS modulo solution so this is probably fine for example code.

I'm curious why the JS solution of applying different classes would be bad for perf in this case? It's not inside a style function, so there aren't any of the concerns that can come up around that.

KevinTCoughlin:
> > nth-child selectors in CSS are better re: perf than the run-time JS modulo solution so this is probably fine for example code.
> 
> I'm curious why the JS solution of applying different classes would be bad for perf in this case? It's not inside a style function, so there aren't any of the concerns that can come up around that.

With the JS solution, you're executing code each time that DOM needs to be rendered which involves the main thread vs. adding two rules _once_ to the browser's CSSOM that the browser evaluates when painting.

Browsers should be optimized for handling CSS rules (especially basic ones like these). So if we can rely on that and remove work from the main thread, that seems ideal. Caveat being that I haven't profiled this and its so inconsequential it likely wouldn't be noticeable either way.

The reason I advocated for its use here is that I think certain scenarios are best showcased to still be handled via CSS than JS.

ecraig12345:
Re: @KevinTCoughlin's comment, I see your point in general, but I'm still not entirely convinced of the perf impact in this case (assuming the two-class version got rid of the unnecessary `css()` helper call...don't know why that was there). I wouldn't expect adding one more class to have significant impact, and the cost of the modulo condition itself (minus `css` call) is negligible. Regardless, I think it's fine to leave this as-is now.

PR #11096, opened 2019-11-06
Remove SCSS From SearchBox Examples

pandasa123:
#### Pull request checklist

- [x] Addresses an existing issue: Related to #6075 

#### Description of changes

- Removed SCSS file
- Added mergeStyleSets pattern similar to #11044

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11096)


maxwellred:
Looks good to me thanks, @pandasa123 !

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.CustomIcon.Example.tsx
```diff
@@ -1,12 +1,18 @@
 import * as React from 'react';
 import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
-import './SearchBox.Examples.scss';
+import { mergeStyleSets } from '@uifabric/styling';
+
+const styles = mergeStyleSets({
```
ecraig12345: Since this example just has one SearchBox, it doesn't need any extra styles--the margin style was just added to make space between SearchBoxes.

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.Disabled.Example.tsx
```diff
@@ -1,12 +1,18 @@
 import * as React from 'react';
 import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
-import './SearchBox.Examples.scss';
+import { mergeStyleSets } from '@uifabric/styling';
+
+const styles = mergeStyleSets({
+  searchBox: {
```
ecraig12345: The new styles aren't doing quite the same thing. This is the old version, which applies the margin to **any SearchBox inside the example** (not the example root itself).
```
  .ms-SearchBoxExample .ms-SearchBox {	
    margin: 0 0 10px 0;
  }
```

Typically what we do now for cases like this is rather than using `mergeStyleSets`, replace the wrapper `div` with a `Stack` that has `childrenGap` set (I think it should be 20 instead of 10 to look less cramped). Basically like this:

```
import { Stack } from 'office-ui-fabric-react/lib/Stack';

const stackTokens = { childrenGap: 20 };

...
return (
  <Stack tokens={stackTokens}>
    <SearchBox
...
```

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.FullSize.Example.tsx
```diff
@@ -1,13 +1,19 @@
 import * as React from 'react';
 import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
-import './SearchBox.Examples.scss';
+import { mergeStyleSets } from '@uifabric/styling';
```
ecraig12345: Same comment as disabled example

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.CustomIcon.Example.tsx
pandasa123: Understood! Removed!

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.Disabled.Example.tsx
pandasa123: That makes sense. I updated it to your spec

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.FullSize.Example.tsx
pandasa123: Updated

maxwellred:
Hi @pandasa123 when you get a chance please update the snapshots and I believe that will resolve the build issue, thanks!

pandasa123:
@maxwellred Just updated! Sorry about that!

maxwellred:
@ecraig12345 It looks like he put all your changes in.

PR #11107, opened 2019-11-07
CommandBar: a11y role fix

### LINKED ISSUE #11039: OverflowSet inside of CommandBar uses aria-orientation, even though role is set to 'presentation'
### Environment Information

- **Package version(s)**: office-ui-fabric-react v7.42.0
- **Browser and OS versions**: Microsoft Edge v79.0.305.0

### Describe the issue:

CommandBar currently creates an OverflowSet and passes in the role of "presentation", you can see this [here](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.base.tsx#L108).

OverflowSet though will set a aria-orientation tag of either vertical or horizontal, you can see this [here](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx#L59).

Using aria-orientation with an element who's role is "presentation" is incorrect, as per this W3 document: https://www.w3.org/TR/wai-aria-1.1/#aria-orientation

aria-orientation should only be used with the following roles:

- scrollbar
- select
- separator
- slider
- tablist
- toolbar
- combobox
- listbox
- menu
- menubar
- radiogroup
- tree
- treegrid

This error get's flagged by axe-core, as breaking the aria-allowed-attr rule (https://dequeuniversity.com/rules/axe/3.3/aria-allowed-attr).

#### Actual behavior:

aria-orientation is used with role="presentation".

#### Expected behavior:

role should probably be the default role for an OverflowSet, i.e. "menubar".

https://www.w3.org/TR/wai-aria-1.1/#aria-orientation

maxwellred:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #11039

#### Description of changes

Removal of role={'presentation'} in CommandBar implementation of Overflow 

#### Focus areas to test

Tested with Windows + Narrator and showed no adverse effects

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11107)


micahgodbolt:
ok, that makes sense. added orientation a while ago, but i see how presentation cancels that out

maxwellred:
Hi @micahgodbolt I've tried deleting and re-running repo wide snapshots as well as running update snapshots (and a11y snapshots) in the experiments package and can't seem to get generate any other snapshots than the ones pushed here. Am I misinterpreting the build error readout?

joschect:
@maxwellred I'm not sure why this is happening. It could be that you need to merge from master and rebuild, it's possible that a snapshot was changed or added at some point that your change impacts.

maxwellred:
Closing in favor of #11224. Thanks @micahgodbolt !

PR #11112, opened 2019-11-07
Microfeedback: (a11y) Added arial labels to component rating buttons

kelkhadiri:
#### Pull request checklist

#### Description of changes

(a11y) Added ariaLabel to the rating buttons in microfeedback component and exposed them as props so that screen readers can announce the appropriate meaning/behavior of the buttons.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11112)

### DIFF
--- a/packages/experiments/src/components/MicroFeedback/MicroFeedback.view.tsx
+++ b/packages/experiments/src/components/MicroFeedback/MicroFeedback.view.tsx
@@ -12,6 +12,8 @@ export const MicroFeedbackView: IMicroFeedbackComponent['view'] = props => {
     sendFollowUpIndex,
     likeIconTitle,
     dislikeIconTitle,
+    likeIconAriaLabel,
+    dislikeIconAriaLabel,
     likeQuestion,
     dislikeQuestion,
     vote,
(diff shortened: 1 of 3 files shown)


packages/experiments/src/components/MicroFeedback/MicroFeedback.view.tsx
```diff
@@ -12,6 +12,8 @@ export const MicroFeedbackView: IMicroFeedbackComponent['view'] = props => {
     sendFollowUpIndex,
     likeIconTitle,
     dislikeIconTitle,
+    likeIconAriaLabel,
```
dzearing: I don't have a problem with this at all, but I think in our next release, we could approach these through slotProps:

```
// Option 1: provide a consistent way to send these sorts of things to particular slots without creating a huge contract.
<MicroFeedback slotProps={{ likeIcon: { ariaLabel: '' } }} />

// Option 2: consolidate string labels into a single prop:
<MicroFeedback strings={{ likeIconTitle,. likeIconAriaLabel, etc }} />

// Option 3: both. User can default to the `strings` prop for providing a specific set of strings, but use `slotProps` as a fallback mechanism.
```

The benefit of the first is that it's a predictable and all encompassing way to pass things to sub components. You can use the same patern to pass along `title`, a data attribute, or any other specific requirement.

The benefit of the later is that for the user, to make the component accessible, they know exactly what to do. They get intellisense for the composite strings object.

This is just a discussion point, not actionable here. Again, because we don't have either of these patterns, this is a fine change, but in the next major, we're leaning towards the first or third option.

@aneeshack4 @ecraig12345 @JasonGore @jdhuntington @joschect @khmakoto @xugao

dzearing:
Also you should make this change in master (Fabric 7) first, really, and cherry pick it to a prior release. Otherwise you're creating an upgrade breaking change.

maxwellred:
@kelkhadiri I'll merge this in but please cherry pick these changes into Fabric 7 master, thanks!

PR #11122, opened 2019-11-07
SearchBox: Remove unsupported autocomplete documentation

### LINKED ISSUE #11121: SearchBox documentation describes unsupported autocomplete suggestions
Docs here describe autocomplete behavior in the "Autocomplete suggestions" and "Do/Don't" sections, but that functionality isn't supported:
https://developer.microsoft.com/en-us/fabric/#/controls/web/searchbox

JasonGore:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11121

#### Description of changes

Remove autocomplete documentation as feature is not supported.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11122)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/SearchBox/docs/SearchBoxBestPractices.md
+++ b/packages/office-ui-fabric-react/src/components/SearchBox/docs/SearchBoxBestPractices.md
@@ -1,14 +1,8 @@
-## Zero input state
+### Zero input state
 
 When the user has clicked into the SearchBox, but has not entered any text, there is an opportunity to display "hint text" within the input field, explaining what a user can do next. This could prompt a user to search for specific type content, or explain the scope of the search. Examples include "type to search", "try searching for \<x\>", "search for a place" or "type to search in \<x location\>".
 
-## Full search
-
-If a user hits "enter" after entering input, a full search is executed. Full searches often go to another "results" page, or change/filter the content of the current page to show only applicable content. The results can appear in any form that best communicates the content.
-
-As a general guideline, results should be displayed in context with the query that was typed, with immediate access to edit the query or enter a new one. One method to enable efficient access to both edit the previous query and enter a new query is to highlight the previous query when the field is reactivated. This way, any keystroke will replace the previous string, but the string is maintained so that the user can position a cursor to edit or append the previous string.
-
-## Search scopes
+### Search scopes
 
 Although search entry points tend to be similarly visualized, they can provide access to results that range from broad to narrow. By effectively communicating the scope of a search, you can help to ensure that the user expectation will be met by the capabilities of the search you are performing, which will reduce the possibility of frustration. The search entry point should be juxtaposed with the content being searched.
 
@@ -18,6 +12,6 @@ Some common search scopes include:
 - **Web:** Search a web index. Results include pages, entities, and answers.
 - **My stuff:** Search across device(s), cloud, social graphs, and more. Results are varied, but are constrained by the connection to user account(s).
 
-## SearchBox with no parent container
+### SearchBox with no parent container
 
 Use a SearchBox without a parent container when it is not restricted to a certain width to accommodate other content. This search box will span the entire width of the space it's in.
--- a/packages/office-ui-fabric-react/src/components/SearchBox/docs/SearchBoxBestPractices.md
+++ b/packages/office-ui-fabric-react/src/components/SearchBox/docs/SearchBoxBestPractices.md
@@ -2,10 +2,6 @@
 
 When the user has clicked into the SearchBox, but has not entered any text, there is an opportunity to display "hint text" within the input field, explaining what a user can do next. This could prompt a user to search for specific type content, or explain the scope of the search. Examples include "type to search", "try searching for \<x\>", "search for a place" or "type to search in \<x location\>".
 
-## Autocomplete suggestions
-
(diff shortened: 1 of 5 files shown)


cliffkoh:
Actually I just wonder, since usually people use searchbox together with some form of auto-complete (in the form described, i.e. in a dropdown), maybe there is still some value in having these be on the page.. but just clearer that searchbox does not ship out of box with it. Just my 2 cents.

JasonGore:
I think these pages in general are overly burdened without going beyond their component's scope to give product guidance. It may be useful somewhere but I don't think this page is the best place for it. Also, I think including this type of guidance would grow all pages in general, not just this one.

packages/office-ui-fabric-react/src/components/SearchBox/docs/SearchBoxBestPractices.md
```diff
@@ -2,10 +2,6 @@
...
-
-As the user enters a query string, they are provided with a dropdown of autocomplete suggestions or disambiguation options. This will help them expedite the input process and formulate an effective query. Recent search history, trending searches, contextual search suggestions, hints and tips are all good candidates for autocomplete content. In general, autocomplete suggestions have the user's input highlighted in some way (generally bolded) to indicate why it's being displayed. As the user enters more keystrokes, the suggestions update continuously/in real time. To see autocomplete suggestions, the user does not need to hit enter (execute a full search), as it is a lightweight way to get quick suggestions or results. If there are mixed result types within the autocomplete suggestions, provide visual indicators or grouping to help organize the information, making it easier to parse.
-
 ## Full search
 
 If a user hits "enter" after entering input, a full search is executed. Full searches often go to another "results" page, or change/filter the content of the current page to show only applicable content. The results can appear in any form that best communicates the content.
```
ecraig12345: This one is also a detail of specific usage patterns that aren't implemented out of the box.

JasonGore: I agree on "Full search". What do you think about "Search scopes"? Also seems a bit out of scope (pun actually not intended.)

ecraig12345: That one I have less of an issue with since it doesn't seem to imply that the SearchBox has some OOB capability which it doesn't actually have. 

I wonder if instead of removing these sections, we should put all of them under another section which makes it clear that these are recommendations for additional behavior you can add and **not** things supported out of the box?

(Incidentally, while you're here, the headings in this file ought to start at h3 instead of h2, to nest under the h2 the website uses as a card header. And if you added a recommendation section, the points under it should be h4.)

JasonGore: Honestly this doesn't appear to be much more than a glorified TextField. I think we should consider removing SearchBox entirely rather than insinuate a component being more than it is and burdened with providing a bunch of guidance for behavior it doesn't provide.

packages/office-ui-fabric-react/src/components/SearchBox/docs/SearchBoxBestPractices.md
```diff
@@ -1,18 +1,8 @@
-## Zero input state
+### Zero input state
 
 When the user has clicked into the SearchBox, but has not entered any text, there is an opportunity to display "hint text" within the input field, explaining what a user can do next. This could prompt a user to search for specific type content, or explain the scope of the search. Examples include "type to search", "try searching for \<x\>", "search for a place" or "type to search in \<x location\>".
 
-## Autocomplete suggestions
```
micahgodbolt: good example of design component documentation the explains how you SHOULD implement auto complete, even if we haven't wired it into the react control.

JasonGore: I'm swinging the other direction on this one. It's confusing users (per the issue), implying the `SearchBox` has autocomplete functionality. In fact, looking at `SearchBox` implementation it barely seems to add value over a regular `TextField`. I'm questioning the need for `SearchBox` entirely, per my other comment on this PR. I think this is a good example of where Fabric could narrow its scope a bit.

packages/office-ui-fabric-react/src/components/SearchBox/docs/SearchBoxDos.md
```diff
@@ -2,8 +2,4 @@
 - Example: "Search"; "Search files"; "Search site"
 - Once the user has clicked into the SearchBox but hasn’t entered input yet, use "hint text" to communicate search scope.
 - Examples: "Try searching for a PDFs"; "Search contacts list"; "Type to find \<content type\>"
-- Provide autocomplete suggestions to help the user search quickly. These suggestions can be from past searches or auto-completions of the user's query text.
-- Provide autocomplete suggestions where there are strong matches to the user's query that the user may want to view immediately.
-- Use a visual separator to define a group of a similar or conceptually aligned autocomplete suggestions.
-- If possible, provide a preview (e.g. image, title, etc.) for autocomplete suggestions to help the user quickly determine if the suggested result is what they were searching for.
-- Use the Underlined SearchBox for CommandBars.
\ No newline at end of file
+- Use the Underlined SearchBox for CommandBars.
```
ecraig12345: While we're here?
```suggestion
- If the SearchBox is inside a CommandBar, use the underlined variant.
```

PR #2878, opened 2017-09-19
Button: Add ability to dismiss menu from IButton interface

christiango:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [x] Include a change request file using `$ npm run change`

#### Description of changes
When implementing onRender for items rendered inside IContextualMenu, it is often desirable to have the click handler also dismiss the menu. This is currently not possible and can be seen for the 'charms' submenu in [ContextualMenu customization example]:(https://dev.office.com/fabric#/components/contextualmenu). 
dismissbug

This PR adds a dismissMenu function on the IButton interface to allow click handlers to dismiss the menu associated with a button. The functionality is integrated into the Contextual Menu customization example on the demo site. 

This PR also updates the ContextualMenu examples to use MenuProps on BaseButton instead of managing the showMenu state manually.

PR #11124, opened 2019-11-08
ContextualMenu: Add example using ContextualMenu directly

### LINKED ISSUE #11113: ContextualMenu for other components

Add an ability to use `ContextualMenu`s outside a buttons, like in `Link`s.

`Link`, `TextField` (for "mention" tagging)

No

Honestly I would expect to get a standalone `ContextualMenu` component even for overriding Right Click event

JasonGore:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11113

#### Description of changes

PR #2878 removed all uses of `ContextualMenu` from the examples, leading to some confusion in #11113 as to the availability of a generic `ContextualMenu` component. This PR just adds an example using `ContextualMenu` directly.

Also fix some inconsistency in basic vs. default naming in the existing example.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11124)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Basic.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Basic.Example.tsx
@@ -1,9 +1,39 @@
 import * as React from 'react';
 import { ContextualMenu, ContextualMenuItemType, IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
+import { useConstCallback } from '@uifabric/react-hooks';
 
 export interface IContextualMenuBasicExampleState {
   showContextualMenu?: boolean;
 }
+
+export const ContextualMenuBasicExample: React.FunctionComponent = () => {
+  const linkRef = React.useRef(null);
+  const [showContextualMenu, setShowContextualMenu] = React.useState(false);
+  const onShowContextualMenu = useConstCallback(() => setShowContextualMenu(true));
+  const onHideContextualMenu = useConstCallback(() => setShowContextualMenu(false));
+
+  return (
+    <div>
+      This example directly uses ContextualMenu to show how it can be attached to arbitrary elements. The remaining examples use
+      ContextualMenu through Fabric Button components.
+      <p>
+        <b>
+          <a ref={linkRef} onClick={onShowContextualMenu}>
+            Click for ContextualMenu
+          </a>
+        </b>
+      </p>
+      <ContextualMenu
+        items={menuItems}
+        hidden={!showContextualMenu}
+        target={linkRef}
+        onItemClick={onHideContextualMenu}
+        onDismiss={onHideContextualMenu}
+      />
+    </div>
+  );
+};
+
 const menuItems: IContextualMenuItem[] = [
   {
     key: 'newItem',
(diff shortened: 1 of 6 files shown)


packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Basic.Example.tsx
```diff
@@ -1,79 +1,103 @@
...
-              },
-              {
-                key: 'disabled',
-                text: 'Disabled item',
-                disabled: true,
-                onClick: () => console.error('Disabled item should not be clickable.')
-              }
-            ]
-          }}
+        This example directly uses ContextualMenu to show how it can be attached to arbitrary elements. The remaining examples use
```
maxwellred: Just thinking out loud but should this explanation also be in overview? Maybe in something like a "Notes" section?

JasonGore: Hmm, I think it's better with the example, in case someone wants to change the example, or wonders why it's there. I think having this specific example NOT be a button helps call it out, too.

ecraig12345: Probably not. The overview is really just meant for a high-level explanation (the components using it as a dumping ground for extra info ideally shouldn't be doing that). Right now the best place for content which is this developer-specific is next to the relevant example.

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Basic.Example.tsx
```diff
@@ -1,79 +1,103 @@
 import * as React from 'react';
-import { ContextualMenuItemType } from 'office-ui-fabric-react/lib/ContextualMenu';
-import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
-import './ContextualMenuExample.scss';
+import { ContextualMenu, ContextualMenuItemType, IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
+
+export interface IContextualMenuBasicExampleState {
+  showContextualMenu?: boolean;
+}
+const menuItems: IContextualMenuItem[] = [
```
ecraig12345: Nitpicky thing but I've started to put item list definitions like this *after* the main example component, so people viewing the code can more quickly understand the "meat" of how it works. (This is safe from a declaration order perspective as long as you're only using the items within a function/method.)

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Basic.Example.tsx
```diff
@@ -1,79 +1,103 @@
...
+  },
+  {
+    key: 'disabled',
+    text: 'Disabled item',
+    disabled: true,
+    onClick: () => console.error('Disabled item should not be clickable.')
+  }
+];
+
+export class ContextualMenuBasicExample extends React.Component<{}, IContextualMenuBasicExampleState> {
```
ecraig12345: Another thing you can take or leave, I've been trying to convert the examples to function components since that's the pattern we want to encourage now. So if you don't mind converting these ones, that would be nice (but not essential).

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Basic.Example.tsx
maxwellred: Sounds good, thanks everybody!

PR #11125, opened 2019-11-08
Refining typography and signal icon on top of Folder Icon, filetype mapping updates

bigbadcapers:
#### Description of changes

- [x] Replaces the filetype icon salt to ensure new Fluid icon is visible
- [x] Makes the signal icon and the itemcount in the folder tile view accessible and better laid out
- [x] Adds bigbadcapers to codeowners for the foldercover because masochism

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11125)

### DIFF
--- a/packages/experiments/src/components/FolderCover/FolderCover.scss
+++ b/packages/experiments/src/components/FolderCover/FolderCover.scss
@@ -1,5 +1,7 @@
 @import '~office-ui-fabric-react/src/common/common';
 
+@import '~office-ui-fabric-react/src/common/ThemingSass.scss';
+
 $folder-accent: #bf5712;
 
 .root {
@@ -115,12 +117,12 @@ $folder-accent: #bf5712;
 
   &,
   &.isSmall {
-    font-size: $ms-font-size-16;
+    font-size: $ms-font-mediumPlusFontSize;
     margin-bottom: 1px;
   }
 
   .isLarge & {
-    font-size: $ms-font-size-16;
+    font-size: $ms-font-mediumPlusFontSize;
     margin-bottom: 4px;
     @include margin-right(7px);
   }
--- a/packages/experiments/src/components/FolderCover/FolderCover.scss
+++ b/packages/experiments/src/components/FolderCover/FolderCover.scss
@@ -115,7 +115,8 @@ $folder-accent: #bf5712;
   @include margin-left(8px);
   @include margin-right(4px);
 
-  .isFluent & {
+  // Added :first-child to CSS selector because a rule from Signal.scss affecting the Icon will otherwise trample this color change.
+  .isFluent & :first-child {
     color: $folder-accent;
   }
 
--- a/packages/experiments/src/components/FolderCover/FolderCover.scss
+++ b/packages/experiments/src/components/FolderCover/FolderCover.scss
@@ -115,6 +115,10 @@ $folder-accent: #bf5712;
   @include margin-left(8px);
   @include margin-right(4px);
 
+  .isFluent & {
+    color: $folder-accent;
+  }
+
   &,
   &.isSmall {
     font-size: $ms-font-mediumPlusFontSize;
--- a/packages/experiments/src/components/FolderCover/FolderCover.scss
+++ b/packages/experiments/src/components/FolderCover/FolderCover.scss
@@ -1,5 +1,6 @@
-
 @import '~office-ui-fabric-react/src/common/common';
+
+$folder-accent: #bf5712;
 
 .root {
   display: inline-flex;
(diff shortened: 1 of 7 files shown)


packages/experiments/src/components/FolderCover/FolderCover.scss
```diff
@@ -88,19 +91,19 @@
 
   .isFluent & {
     @include text-align(left);
-    color: $ms-color-yellowDark;
+    color: $folder-accent;
```
phkuo: we should make this a theme color, otherwise it'll break/be unchangeable if it ever gets used on SharePoint (and when we do theming in OneDrive, like for dark mode)

bigbadcapers: I was thinking of the same thing so i consulted with @Jahnp on the best approach and where to declare the color value, here's what we resolved: This color value is for use on text and icons that will only ever appear on top of the folder SVG itself, never set against a theme-controlled background, so it actually shouldnt be tied to the theme, but rather only ever updated if we change out the folder plate svgs like this one: https://spoprod-a.akamaihd.net/files/fabric/office-ui-fabric-react-assets/foldericons-fluent/lg-fg.svg

bigbadcapers:
@betrue-final-final or @dzearing could you help me to declare the change the screener changed as intentional? The shared people icon should appear smaller and with better contrast against the folder plate. I checked out the pics in screener and the change it detected is the actual intent of the PR.

bigbadcapers:
> @betrue-final-final or @dzearing could you help me to declare the change that Screener found as intentional? The shared people icon should appear smaller and with better contrast against the folder plate. I checked out the pics in screener and the change it detected is the actual intent of the PR.

@maxwellred are you able to do the blessing?

maxwellred:
Sorry @bigbadcapers, I don't have the ability to accept screener changes.

bigbadcapers:
hey @maxwellred peter suggested that i can get the size auditor to re-run by closing and reopening the PR, is that the recommended way? im happy to do it unless its going to cause anything else to potentially get lost, like the current discussion etc.

phkuo:
@bigbadcapers closing and re-opening preserves all conversations and comments.

maxwellred:
So I did a little digging and It looks like the bot is failing on node-sass install. I checked some node-sass issues and they set some version minimums per Node version. Perhaps since the build definition is using Node v12.13.0 we need to bump the node-sass package to the latest to get around this error? Heres the Node-sass issue and a screenshot
 https://github.com/sass/node-sass/pull/2633/files
Screen Shot 2019-11-12 at 7 05 03 PM

bigbadcapers:
> @phkuo So I did a little digging and It looks like the bot is failing on node-sass install. I checked some node-sass issues and they set some version minimums per Node version. Perhaps since the build definition is using Node v12.13.0 we need to bump the node-sass package (currently 4.10) to the latest to get around this error? Heres the Node-sass issue and a screenshot
> https://github.com/sass/node-sass/pull/2633/files
> Screen Shot 2019-11-12 at 7 05 03 PM

interesting! is this happening to other PRs? or is there something i can do with this one to overcome that?

bigbadcapers:
Woohoo! checks passed, thank you Max! @ThomasMichon et al, can you take a peek and lmk if this is ready for merging?

maxwellred:
@Jahnp or @KatherineThayerMicrosoft It appears Thomas Michon is out of office so I was wondering if one of you could take a look at this PR and approve, thanks!

packages/file-type-icons/src/FileTypeIconMap.ts
```diff
@@ -249,7 +248,7 @@ export const FileTypeIconMap: { [key: string]: { extensions?: string[] } } = {
   contact: {
     extensions: ['vcf']
   },
-  css: {}, // we dont have the icon yet, but i believe we want it, snapping to 'code' for now
+  /*  css: {},  not broken out yet, snapping to 'code' for now */
```
ecraig12345: Technically it's a breaking change to remove the `css` and `channelfolder` properties, even if they weren't being used. I'm not familiar with how the file type icons work--would it cause any problems for those keys to continue to be present but empty?

ecraig12345: Should have said, it's a breaking change if this const is exported from the package. If it's not exported from the root, the change is probably fine.

bigbadcapers: it's a good question. channelfolder is never used. css is covered in another row, so these changes wont affect any given filetype icon rendering in the current product.

bigbadcapers: @ecraig12345 @maxwellred let me know if this is cool and we can merge.

PR #11126, opened 2019-11-08
API doc generation refactor

ecraig12345:
#### Pull request checklist

#### Description of changes

Simplify and modularize the API doc generation code and tables for displaying API docs. This will make it easier to add features or reuse parts of the code for doc generation in other formats. (The latter part was the immediate motivation of this change.)

All existing UI elements are retained, plus a new feature: detect required props and mark them as required in the UI.

The schema of the page.json files and some props on the API table components have changed, but this should be safe because it doesn't appear to be used elsewhere. (Both the api-docs package and the API tables are tightly tied to the structure of the Fabric project and website, further decreasing the likelihood of them being used elsewhere.)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11126)


maxwellred:
Hi @kenotron I was hoping you could take a look at this PR when you have a chance, thanks!

packages/example-app-base/package.json
```diff
@@ -46,6 +46,7 @@
     "@microsoft/load-themed-styles": "^1.7.13",
     "@uifabric/example-data": "^7.0.1",
     "@uifabric/fluent-theme": "^7.1.4",
+    "@uifabric/react-hooks": "^7.0.1",
```
kenotron: @KevinTCoughlin - do you think there's any implications on any of your consumption of this package?

ecraig12345: example-app-base should only be referenced at development time (also tsx-editor already depends on react-hooks), so it shouldn't have an impact on production bundles.

PR #11155, opened 2019-11-12
List size

### LINKED ISSUE #10939: List items aren't announcing position in set
### Environment Information

- **Package version(s)**: All

### Describe the issue:

List items don't announce where they are in the set. E.g. "2 of 10".

1. Load any list. The one in the Fabric demo page will do.
2. Turn on Narrator.
3. Navigate to a list item.

#### Actual behavior:

Narrator doesn't announce the list items position in the set.

If you look at it's aria attributes, these details are not set.

#### Expected behavior:

The list item that has focus should announce its position in the set. E.g. "2 of 10".

jdhuntington:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #10939

#### Description of changes

Specify `aria-posinset` and `aria-setsize` for `List` elements.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11155)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/DetailsList/__snapshots__/DetailsRow.test.tsx.snap
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/__snapshots__/DetailsRow.test.tsx.snap
@@ -779,6 +779,8 @@ exports[`DetailsRow renders details list row correctly 1`] = `
                   style={Object {}}
                 >
                   <div
+                    aria-posinset={1}
+                    aria-setsize={5}
                     className="ms-List-cell"
                     data-automationid="ListCell"
                     data-list-index={0}
@@ -789,6 +791,8 @@ exports[`DetailsRow renders details list row correctly 1`] = `
                     </div>
                   </div>
                   <div
+                    aria-posinset={2}
+                    aria-setsize={5}
                     className="ms-List-cell"
                     data-automationid="ListCell"
                     data-list-index={1}
@@ -799,6 +803,8 @@ exports[`DetailsRow renders details list row correctly 1`] = `
                     </div>
                   </div>
                   <div
+                    aria-posinset={3}
+                    aria-setsize={5}
                     className="ms-List-cell"
                     data-automationid="ListCell"
                     data-list-index={2}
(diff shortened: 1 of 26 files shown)


packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -473,7 +473,8 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
   }
 
   private _onRenderPage = (pageProps: IPageProps<T>, defaultRender?: IRenderFunction<IPageProps<T>>): any => {
-    const { onRenderCell, role } = this.props;
+    const { onRenderCell, role, items: allItems } = this.props;
+    const totalItemCount = allItems ? allItems.length : 0;
```
jdhuntington: I don't love accessing `props.items` here, but it doesn't feel like the value belongs on `pageProps` either. Since we're already accessing `props` it feels the least bad.

jdhuntington:
https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships#aria-posinset_aria-setsize

packages/office-ui-fabric-react/src/components/DetailsList/__snapshots__/DetailsRow.test.tsx.snap
```diff
@@ -799,6 +803,8 @@ exports[`DetailsRow renders details list row correctly 1`] = `
                     </div>
                   </div>
                   <div
+                    aria-posinset={3}
```
dzearing: I'm a bit confused here on this one. This might be wrong, but I'm not sure. Did you test how this reads?

I would have assumed that these aria attributes should be on a focusable entity.

JasonGore:
Is this for virtualized list scenarios? According to the [ARIA `aria-posinset` property documentation](https://www.w3.org/TR/wai-aria-1.1/#aria-posinset), `aria-posinset` should only need to be set if all required elements aren't present in the DOM. I want to make sure this isn't a browser issue. If these DOM elements are already in the tree, then the browsers and screen readers should have the information they need to narrate row information.

jdhuntington:
> Is this for virtualized list scenarios?
Yes, that's exactly what this is for.

KevinTCoughlin:
> > Is this for virtualized list scenarios?
> 
> Yes, that's exactly what this is for.

Should we avoid adding the `aria-*` attributes when `onShouldVirtualize` returns `false` then?

jdhuntington:
Closing, will reopen with a more targeted approach that won't impact things like details list.

PR #11156, opened 2019-11-12
Nav: fix hover styles

### LINKED ISSUE #10565: Nav: Chevron button doesn't need backgrounds in any state
Nav's chevron button's background should be permanently set to transparent because it should always have the same bg color as the item. (This is just code cleanup, this does not currently manifest visually at all.)
Same with command bar menu button chevrons.

phkuo:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10565

#### Description of changes

When hovering on a collapsible item in the Nav, the arrow chevron gets a different and incorrect background color. Instead, just have it permanently be transparent, so it picks up the background color set by the Link it's on top of.
Also remove some unnecessary hover styles that overwrote rest state styles with the same value.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11156)

### DIFF
--- a/change/office-ui-fabric-react-2019-11-11-18-27-18-phkuo-fixNavArrow.json
+++ b/change/office-ui-fabric-react-2019-11-11-18-27-18-phkuo-fixNavArrow.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Nav: remove incorrect and unnecessary hover styles",
+  "packageName": "office-ui-fabric-react",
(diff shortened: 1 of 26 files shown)


phkuo:
@betrue-final-final can you review screener changes?

phkuo:
@dzearing for screener changes

PR #11159, opened 2019-11-12
Apply accessibility changes to main package

betrue-final-final:
#### Pull request checklist

Updated the main package Calendar to use the styles from #10980 

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11159)


lorejoh12:
Something seems off in two of the examples:
"Inline Calendar with week selection and overlaid year picker when month header is clicked":
- hover state looks a shade darker than the hover state in day view
- there's no difference between the hover state and the pressed/selected state

"Inline Calendar with month selection and overlaid year picker when month header is clicked":
- the hover state is too dark. It's supposed to be the same hover color (probably neutralLighter) as the hover state in day view, right?
- the pressed state looks really weird. The individual day stays the same color, but the rest of the month gets way too dark.

packages/office-ui-fabric-react/src/components/Calendar/Calendar.scss
```diff
@@ -197,14 +199,16 @@ $Calendar-dayPicker-margin: 10px;
 .dayIsFocused:hover,
 .dayIsUnfocused:hover {
   cursor: pointer;
-  background: $ms-color-neutralLight;
+  background: $ms-color-neutralLighter;
```
lorejoh12: I believe you also need to make this change in .dayHover class as well

packages/office-ui-fabric-react/src/components/Calendar/Calendar.scss
```diff
@@ -228,14 +232,14 @@ $Calendar-dayPicker-margin: 10px;
...
+  background-color: $ms-color-themeLight;
 }
 
 .dayIsFocused:active,
 .dayIsHighlighted {
   &.day {
-    color: $ms-color-neutralPrimary;
-    font-weight: $ms-font-weight-semibold;
+    color: $ms-color-neutralDark;
+    background-color: $ms-color-neutralLight;
```
lorejoh12: I think you need to add this to .dayPress as well to fix the month selection issue

packages/office-ui-fabric-react/src/components/Calendar/Calendar.scss
betrue-final-final: Done

packages/office-ui-fabric-react/src/components/Calendar/Calendar.scss
```diff
@@ -746,26 +750,25 @@ $Calendar-dayPicker-margin: 10px;
...
-    background-color: $ms-color-neutralLight;
+    background-color: $ms-color-neutralLighter;
     color: $ms-color-neutralDark;
     @include high-contrast {
       outline: 1px solid highlight;
     }
   }
   &:active {
-    background-color: $ms-color-themeLight;
+    background-color: 'red'; //$ms-color-neutralTertiaryAlt;
```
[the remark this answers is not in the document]
betrue-final-final: Oops, removing the test.

PR #11162, opened 2019-11-12
Saving one extra mount in ResizeGroup

whykushal93:
#### Pull request checklist

#### Description of changes
ResizeGroup has two divs - one for measurement and one for actually rendering the resize able contents. When it receives new props, it uses the measurement div that is hidden from the user to perform measurements. After it is able to scale the new contents to fit, it uses the render div to render the scaled new content. As a result, React, completely unmounts everything under the measurement div, and re-mounts these new contents in the render div. As a result, we pay the cost to mount these new components twice. 

Instead this PR, switches which div is the measurement div so that new contents can be measured and rendered in the same div. And, instead of unmounting and remounting, we can mount only once. In the second render, where we make the new contents visible, we still need to update them. But, this is generally cheaper than a new mount. In fact, if you have PureComponents, this almost has zero cost.

This is going to give us at least 30%-40% improvement in Tab Switching Perf. when no scaling steps are applied (fairly common scenario).

As an example, a slightly modified version of the Fabric example, has the following results (not to scale). 
**Before:**
image

**After:**
image

Also added a test to preserve this behavior in the future. 
#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11162)


dzearing:
It might be a great scenario to add a flamegrill test for. Maybe you could take a stab to think about how this would work? Just an idea. It would be ideal to catch regressions in this space as soon as we can.

packages/office-ui-fabric-react/src/components/ResizeGroup/ResizeGroup.base.tsx
```diff
@@ -29,6 +31,16 @@ export interface IResizeGroupState {
...
+
+  /**
+   * When we get new props, we typically first create a measure div and in the next render put this data into a new rendered div
+   * (See the render function below)
+   *
+   * When we do this, React unmounts the measured div and its children fully and remounts all of them in the rendered div.
+   * To avoid this, we flip the keys to "fool" React so that it updates the previously measure div to be the render div
+   * instead of remounting.
+   */
+  measureDivKey?: ResizeGroupDivKey;
```
dzearing: That's pretty interesting. :)

packages/office-ui-fabric-react/src/components/ResizeGroup/ResizeGroup.base.tsx
```diff
@@ -419,3 +449,15 @@ export class ResizeGroupBase extends BaseComponent<IResizeGroupProps, IResizeGro
     }
   }
 }
+
+/**
+ * Helper function to switch the keys so that we can "fool" React into thinking
+ * we are measuring and rendering in the same div.
+ * @param key Key to switch
+ */
+function otherKey(key?: ResizeGroupDivKey) {
```
dzearing: I don't see any case where key can be undefined.

dzearing: You could just have a number and simplify this...

```tsx
// key=0:
key={ state.index };

// key=1:
key={ state.index^1 }
```

JasonGore:
This would occur on initial render, correct? If so, it can be added pretty easily to the [scenario directory](https://github.com/OfficeDev/office-ui-fabric-react/tree/master/apps/perf-test/src/scenarios) where it will get automatically picked up.

In fact, we could add the scenario in another PR, merge it and update this branch and see if the perf gains come through. 😄  If I can help here let me know. Thanks!

christiango:
Long term, I wonder if we can avoid a lot of these shennanigans by moving away from ResizeGroup and instead using [useLayoutEffect](https://reactjs.org/docs/hooks-reference.html#uselayouteffect) to let us do everything on the same div without fear of the user seeing the intermediate scaling steps

packages/office-ui-fabric-react/src/components/ResizeGroup/ResizeGroup.base.tsx
christiango: This is only for the initial render that we do this key swap right? If not I fear we will end up doing weird things like losing focus with this trick

christiango: nevermind, it looks like it is for every render. I'm guessing the Ribbon would lose focus if it had it in this case?

whykushal93: Yeah I think this might lead to focus bugs. Did not verify it but makes sense.

whykushal93: I can this would lead enough inconsistency that I think the perf. gains might not be worth it. 
Would it even make sense to handhold focus when we do this switching ? Like keep track of the previousFocusedElement inside the ResizeGroup's contents and see it still is present after the new contents are rendered. It will however raise superfluous "blur" and "focus" event on that element.

whykushal93:
Christian pointed out a focus issue that I do not have an easy solution for.

The issue is because new contents will be rendered in a different div every time, focus can be lost.
The solution could have been to keep track of any focused element in the ResizeGroup's contents and try to focus the corresponsing element when new contents are rendered. However, this is hard to do given that the element focused might not have any selectors on it etc. 
I am closing this PR because of this.

PR #11178, opened 2019-11-14
Optmizing ResizeGroup when it gets new props which do not change the layout

whykushal93:
#### Pull request checklist

#### Description of changes
I have made previous changes that target ResizeGroup unsuccessfully :). This is another try.
This PR tries to solve the following problems - 
1. `ResizeGroup` still uses `UNSAFE_componentWillReceiveProps`. Changed that to be `getDerivedStateFromProps`.

2. `ResizeGroup` puts a lot of strain of memoization of data provided to the `ResizeGroup`. Suppose we have to apply a lot of scaling steps before our data fits. Now, if that data changes slightly (without affecting layout), `ResizeGroup` will start from original unscaled data and ask us to recompute new data after every scaling step. A lot times, this exceeds our cache limit for memoized data generation and leads to unnecessary re-rendering in downstream components. 

I thought it would nice to implement and use batch scaling of data for `ResizeGroup`. In this PR, I use to scale down new props with the same cache key as the previous props. If the cache key matches even after scaling down, we can directly render this data. This saves us a render also !

Note that for batch scaling of data, `ResizeGroup` needs to communicate how many total scaling steps to apply. This information will be read from the data's `__scalingStepsCount` property. The weird name is for avoiding collisions.

PS: With batchScaling, we can potentially in future make improvements to how many measurements we make by binary searching the scalingStepCount space instead of doing a linear search. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11178)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ResizeGroup/ResizeGroup.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ResizeGroup/ResizeGroup.base.tsx
@@ -41,16 +41,6 @@ export interface IResizeGroupState {
    * instead of remounting.
    */
   measureDivKey?: ResizeGroupDivKey;
-
-  /**
-   * Integer which represents all the scaling steps applied for the data to render.
-   * If 0, data in props was rendered as is.
-   * If positive, we grew the data in props before we rendered it.
-   * If negative, we shrunk the data in props before we rendered it.
-   */
-  renderedDataScalingSteps?: number;
-
-  dataToMeasureScalingSteps?: number;
 }
 
 /**
@@ -119,7 +109,6 @@ export const getNextResizeGroupStateProvider = (measurementCache = getMeasuremen
    */
   function _shrinkContentsUntilTheyFit(
     data: any,
-    appliedScalingStepsCount: number,
     onReduceData: (prevData: any) => any,
     getElementToMeasureDimension: () => number
   ): IResizeGroupState {
@@ -170,7 +159,6 @@ export const getNextResizeGroupStateProvider = (measurementCache = getMeasuremen
    */
   function _growDataUntilItDoesNotFit(
     data: any,
-    appliedScalingStepsCount: number,
     onGrowData: (prevData: any) => any,
     getElementToMeasureDimension: () => number,
     onReduceData: (prevData: any) => any
@@ -228,7 +216,7 @@ export const getNextResizeGroupStateProvider = (measurementCache = getMeasuremen
       if (onGrowData) {
         nextState = {
           resizeDirection: 'grow',
-          dataToMeasure: renderedData
+          dataToMeasure: onGrowData(renderedData)
         };
       } else {
         nextState = {
@@ -279,12 +267,7 @@ export const getNextResizeGroupStateProvider = (measurementCache = getMeasuremen
       if (currentState.resizeDirection === 'grow' && props.onGrowData) {
         nextState = {
           ...nextState,
-          ..._growDataUntilItDoesNotFit(
-            currentState.dataToMeasure,
-            currentStaprops.onGrowData,
-            getElementToMeasureDimension,
-            props.onReduceData
-          )
+          ..._growDataUntilItDoesNotFit(currentState.dataToMeasure, props.onGrowData, getElementToMeasureDimension, props.onReduceData)
         };
       } else {
         nextState = {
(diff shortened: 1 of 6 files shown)


christiango:
Can you describe the scenario in a little more detail here? What exactly is so expensive from the current implementation?

If we are using cache keys, shouldn't this just still have a minimal set of renders? Do you have a scenario and some data on what this improves? Would be helpful in determining if the improvement justifies the added complexity.

packages/office-ui-fabric-react/src/components/ResizeGroup/ResizeGroup.base.tsx
```diff
@@ -314,15 +320,65 @@ export class ResizeGroupBase extends BaseComponent<IResizeGroupProps, IResizeGro
...
   private _hasRenderedContent = false;
 
+  public static getDerivedStateFromProps(props: IResizeGroupProps, state: IResizeGroupState) {
+    if (state.prevProps !== props) {
+      const { data: prevData } = state.prevProps!; // We set initial prevProps in getInitialState.
+      const { data, onBatchScaleData } = props;
+      const { renderedData } = state;
+
+      // If cache key is defined and equal to cache key of previous props' data we can do an optimization
+      if (prevData.cacheKey && data.cacheKey === prevData.cacheKey) {
```
christiango: What happens if this prop update also changed the scaling steps? Say a new scaling step was added or one was removed. The scalingStepCount would be incorrect

whykushal93:
Adding a bit more explanation to what we are seeing in case of the Office Online Ribbon when using the ResizeGroup. I think the cacheKey works as expected and we see minimal renders at the ResizeGroup component level. 

The Ribbon (uses ResizeGroup) has a lot of scaling steps it can apply (say around 25). It is also a composite component that constructs the props for down stream components (groups). It depends on good memoization (we use Fabric's `memoizeFunction`) of props to avoid unnecessary renders in its child components.

The Ribbon props look like this - 
```
interface IRibbonProps {
groupProps: IGroupProps[];
}
```

Suppose we are in a scaled down state (20 scaling steps applied) and there is a non-layout altering change in one of these groups. The best case scenario here would be re-render only that particular group and not update any other groups. What currently happens is....

The ResizeGroup starts from the original data (0 scaling steps applied) and calls `onReduceData` 20 times. For each of these `onReduceData` calls, we have to generate new props for each group. 
The scaled props for each group are a function of the scalingSteps and original group props. 
```
scaledPropsForGroup: (unscaledGroup: IGroupProps, scalingSteps: IScalingStep[]) => IGroupProps
```

Since we do not know at which scaling step we will stop, we have to cache/memoize all 20 versions of these arguments for each group. This results in exceeding our cache limits quite frequently and causing new references to props of groups which have not changed. 

This causes us to re-render groups which have not changed also.

packages/office-ui-fabric-react/src/components/ResizeGroup/ResizeGroup.base.tsx
whykushal93: This will definitely create bugs if scaling steps are added or removed. However, in my opinion, that should be rare scenario and we can add it as a caveat to using the batch scaling of data.

christiango: I don't think it's something that rare in the grand scheme of things (especially for a shared component like this), think of add-ins being inserted at runtime or if we enable user customization of the Ribbon.

I'm in favor of improving performance, but not at the expense of correctness. This will be hard to debug when it goes wrong in the future.

christiango: Rather than making things in the shared component more complex, is it possible to keep the existing API but add these smarts in the Ribbon's implementation of onGrowData  onReduceData?

whykushal93: So, when we batchScaleData, at least in the current implmentation, I check to verify that the final cacheKey after batchScaling is the same before rendering. If the scalingSteps change, I would expect the cacheKey to change. If it does not, it should be ok to still render, because the current rendered data has the same key.

micahgodbolt:
@christiango leaning on you and @whykushal93 on this one. Is this still the best approach, or are you looking at alternatives?

ecraig12345:
Echoing @micahgodbolt's question:

> @christiango leaning on you and @whykushal93 on this one. Is this still the best approach, or are you looking at alternatives?

PR #11213, opened 2019-11-14
DropdownOption title consistency for multiSelect case

### LINKED ISSUE #11187: Option to disable tooltip for DropdownOption

Currently for `multiSelect` mode in [`Dropdown.base.tsx`](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx) Checkbox.title is set to `title={item.title ? item.title : item.text}`. This title is used as tooltip for the dropdown option. This means that when not specifying `title`, `text` is used as tooltip. This is redundant. It would be good to have an option to disable that tooltip.

Note: For `!multiSelect ` CommandButton.title is set like: `title={title}`. Can we have the same for `multiSelect`?

Dropdown

Redundant tooltip:
image

Putting a space is only way to get rid of it, but not nice at all: 
image

kenotron:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11187

#### Description of changes

Probably for historical reasons, the DropdownOption render hands down title prop differently between multiselect and non-multiselect cases. These behavior almost the same - the difference is in how the code treats null / undefined.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11213)

### DIFF
--- a/change/office-ui-fabric-react-2019-11-14-12-40-37-dropdown-title.json
+++ b/change/office-ui-fabric-react-2019-11-14-12-40-37-dropdown-title.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "DropDown: make dropdown option title prop to be assigned consistently for the multiselect case",
+  "packageName": "office-ui-fabric-react",
+  "email": "kchau@microsoft.com",
+  "commit": "745782f7e6e0238a70a63feff60658dd62da79c5",
+  "date": "2019-11-14T20:40:37.817Z"
+}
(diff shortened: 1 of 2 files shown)


joschect:
@kenotron I'm not positive but doesn't this [line](https://github.com/OfficeDev/office-ui-fabric-react/pull/11213/files#diff-ea483296e1c9a31579fb983113704f10R590) do the same thing?

joschect:
@kenotron nvm I see how it is handled! Good call.

PR #11218, opened 2019-11-15
Dropdown: fix item hover/keyboard nav state in high contrast mode

### LINKED ISSUE #10953: Dropdown keyboard selection in high contrast selection issues
Dropdown support for high contrast, specifically for selecting using the keyboard to select items does not make the selection clear. Here are some examples:

On the main site (https://developer.microsoft.com/en-us/fabric#/controls/web/dropdown) selection and current focus look like this:
image

On https://fabricweb.z5.web.core.windows.net/oufr/7.54.1/#/examples/dropdown it has a bit different look, but also has issues which we have seen on our own site:

image

Notice how it is very hard to see what item is selected. For a non multi-select example, using arrow keys to switch between Carrot and Lettuce look like this:

image

image

@gouthamigangu

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10953

#### Description of changes
This change only affect single-select dropdown (multi-select dropdown remains same).

image

On keyboard nav/hover with one item selected:
image

Also reverted fix from https://github.com/OfficeDev/office-ui-fabric-react/issues/5238 as it's no longer reproing without it.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11218)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.styles.ts
@@ -45,12 +45,7 @@ const highContrastItemAndTitleStateMixin: IRawStyle = {
     [HighContrastSelector]: {
       backgroundColor: 'Highlight',
       borderColor: 'Highlight',
-      color: 'HighlightText',
-      selectors: {
-        ':hover': {
-          color: 'HighlightText' // overrides the hover styling for buttons that are also selected
-        }
-      }
+      color: 'HighlightText'
     },
     ...highContrastAdjustMixin
   }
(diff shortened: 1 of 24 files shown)


packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.styles.ts
```diff
@@ -45,12 +45,7 @@ const highContrastItemAndTitleStateMixin: IRawStyle = {
     [HighContrastSelector]: {
       backgroundColor: 'Highlight',
       borderColor: 'Highlight',
-      color: 'HighlightText',
-      selectors: {
-        ':hover': {
-          color: 'HighlightText' // overrides the hover styling for buttons that are also selected
```
xugao: removed since it doesn't seem to be needed. original change: https://github.com/OfficeDev/office-ui-fabric-react/issues/5238

joschect:
Awesome! Thanks @xugao

PR #11224, opened 2019-11-15
CommandBar: a11y remove role = presentation

### LINKED ISSUE #11039: OverflowSet inside of CommandBar uses aria-orientation, even though role is set to 'presentation'

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11039

#### Description of changes

Removal of role={'presentation'} in CommandBar implementation of Overflow 

#### Focus areas to test

Tested with Windows + Narrator and showed no adverse effects

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11224)


micahgodbolt:
created this PR to see if I had the same snapshot issues as #11107

micahgodbolt:
snaps sadly look the same. will keep an eye on the build

maxwellred:
Interesting and thanks for trying this as I was about to close my PR and see if opening a fresh branch would've done the trick....guess not

micahgodbolt:
so frustrating. this builds just fine locally

khmakoto:
Can we close #11107 if this goes in?

micahgodbolt:
turning off a bunch of experiment tests, and seeing if this will go thorugh

micahgodbolt:
@maximus12793 it's passing, it's passing!

micahgodbolt:
@khmakoto do you think we need to keep those experimental button tests around? I can just remove them vs commenting them out. they were causing flickers

khmakoto:
@micahgodbolt since we're not planning on shipping the experimental Button anymore I guess it's fine if we remove them. Worst case scenario we can look at the history and bring them back.

maxwellred:
@khmakoto I'll close mine and let this one handle it. Woooo!

PR #11226, opened 2019-11-15
PickerSuggestions: Styles were missing from interface even though they were being passed through

### LINKED ISSUE #11222: Picker personaWrapper style is not applied when setting
### Environment Information

- **Package version(s)**: Office UI Fabric React 7.61.0
- **Browser and OS versions**: Chrome Version 78.0.3904.97 (Official Build) (64-bit) + Windows 10

#### Actual behavior:

When setting _IPeoplePickerItemSuggestionStyles_ for a _NormalPeoplePicker_ everything in _personaWrapper_ is ignored, while the style in _root_ is applied.

**Here is a codepen showing the problem:** https://codepen.io/apollon7/pen/wvvQGNR

**How to reproduce:**

1. Export sample from https://developer.microsoft.com/en-us/fabric#/controls/web/peoplepicker with the button "Export to CodePen"
2. Replace value of _const suggestionProps_ with the following:
```
const suggestionProps: IPeoplePickerItemSuggestionProps = {
    styles: {
        personaWrapper: {
            width: "100%",
            backgroundColor: "green",
            borderStyle: "solid"
        },
        root: {
            borderStyle: "solid"
        }
    },
    suggestionsProps: {
        suggestionsHeaderText: 'Suggested People',
        mostRecentlyUsedHeaderText: 'Suggested Contacts',
        noResultsFoundText: 'No results found',
        loadingText: 'Loading',
        showRemoveButtons: false,
        suggestionsAvailableAlertText: 'People Picker Suggestions available',
        suggestionsContainerAriaLabel: 'Suggested contacts'
    }
};
```

image

#### Expected behavior:

Style from personaWrapper should be applied.

What I want to do is set the width for personaWrapper to 100% to remove the space on the right side reserved for the Remove-Button which I'm not using (showRemoveButtons: false). Currently longer names are truncated with ... even though there is enough space left.
image

micahgodbolt:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #11222

#### Description of changes

Styles were being passed through to picker suggestions via `pickerSuggestionsProps` but the interface didn't include them. The styles prop was also missing typings making it difficult to know what the styles props and targets are of the control. 

#### Focus areas to test

@joschect do you have any issues with tightening of the styles interface? This will certainly throw errors for anyone passing in invalid style targets. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11226)

### DIFF
--- a/change/office-ui-fabric-react-2019-11-15-09-46-34-picker-suggestion-typings.json
+++ b/change/office-ui-fabric-react-2019-11-15-09-46-34-picker-suggestion-typings.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "PickerSuggestions: Fixed typings to include typed styles function",
+  "packageName": "office-ui-fabric-react",
+  "email": "mgodbolt@microsoft.com",
+  "commit": "a6e2ea7d73d0e9af2021bb90f983180c55495096",
+  "date": "2019-11-15T17:46:34.883Z"
+}
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -686,7 +686,7 @@ export const CompactPeoplePicker: React.FunctionComponent<IPeoplePickerProps>;
 export class CompactPeoplePickerBase extends BasePeoplePicker {
     static defaultProps: {
         onRenderItem: (props: IPeoplePickerItemSelectedProps) => JSX.Element;
-        onRenderSuggestionsItem: (personaProps: IPersonaProps, suggestionsProps?: IBasePickerSuggestionsProps<any> | undefined) => JSX.Element;
+        onRenderSuggestionsItem: (personaProps: IPersonaProps, suggestionsProps?: ISuggestionItemProps<IPersonaProps> | undefined) => JSX.Element;
         createGenericItem: typeof createGenericItem;
     };
 }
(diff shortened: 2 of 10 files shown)


micahgodbolt:
talked to @joschect and it seems this typing issue is a much larger can of worms. I'll close this down until he figures out a good course of action.

micahgodbolt:
ok, worked with @joschect to get these typings worked out. updated and reopened.

joschect:
@micahgodbolt I don't have any problems with this change.

@dzearing or @cliffkoh are there any problems with making a change like this? It's a sort of breaking change, but in reality is just bringing everything in line with the way that it should be.

micahgodbolt:
update api snapshots

micahgodbolt:
This is happy and passing now. @joschect can you think of some primary users of this control that we could valid this change with?

joschect:
@micahgodbolt yeah. There are probably some users who we could ask about it. I'll follow up with you offline.

ecraig12345:
Tightening the type of a prop is technically a breaking change, so unfortunately we should probably avoid it until a major release (adding a TODO and a work item would be good). It's safest to just add a note the doc comment for each prop about what the types are, and use them internally. A contrived example to demonstrate what I mean about using the types internally:
```tsx
interface IFooProps {
  thing: any;
}
// in Foo component
const thing: IActualThing = this.props.thing;
```

ecraig12345:
Also you have merge conflicts now.

micahgodbolt:
I don't want to fully close this PR down. When @joschect and I went through this we found a number of completely wrong typings. I'm hoping we can capture some of those fixes while accepting that we can't clamp down on the interface without a major release. 

Jon is Shield next week, so I'll schedule some time with him and talk about what changes in this PR we CAN keep, and then document the workaround to apply correct styling.

micahgodbolt:
I'll close this for now.

Issue #11144, opened 2019-11-11
Calendar: Name is not defined for the dialog which appears on pressing enter on click for calendar button.


dzearing:
Tracking all of the issues brought up with the 03/15 accessibility pass on Calendar fabric control
Details on individual bugs can be found on [Sharepoint](https://microsoft-my.sharepoint.com/:x:/r/personal/v-amaver_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B8cfe67c5-ce76-4083-b60e-4190dbd38206%7D&action=default&uid=%7B8CFE67C5-CE76-4083-B60E-4190DBD38206%7D&ListItemId=1872&ListId=%7B395E1ECB-5002-440E-81C5-6A090E98077C%7D&odsp=1&env=prodbubble)

lorejoh12:
I'm unsure about what the correct fix is here. What name should we be using? I'm guessing by "name", this bug is about there being no aria-label on the dialog that pops up in the examples. Technically that's a bug with the example itself, not the Calendar component, because the example is what wraps the calendar in the Popup, but if this is something that needs to be fixed, it'll need to be fixed in the DatePicker component, so we should get some clarity on expected behavior here.

lorejoh12:
For some reason the PR merging didn't auto close the issues... closing, this should be fined now.

Issue #11142, opened 2019-11-11
Calendar: Information about the change of week is not narrated by screen reader whenever user press Enter on previous/next button.


dzearing:
Tracking all of the issues brought up with the 03/15 accessibility pass on Calendar fabric control
Details on individual bugs can be found on [Sharepoint](https://microsoft-my.sharepoint.com/:x:/r/personal/v-amaver_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B8cfe67c5-ce76-4083-b60e-4190dbd38206%7D&action=default&uid=%7B8CFE67C5-CE76-4083-B60E-4190DBD38206%7D&ListItemId=1872&ListId=%7B395E1ECB-5002-440E-81C5-6A090E98077C%7D&odsp=1&env=prodbubble)

lorejoh12:
This is a bug with the example page, not the control, although we can mitigate this issue with a change to the control.

Right now, the control only announces the navigated state. Month and year. So when the month and year change (either by clicking the arrows within the control itself, or by clicking the example's buttons below the control enough times to change the month), the data gets announced. However, there's no aria-live for the selected date itself.

The simplest fix is to just put aria-live on the selected date in the label for the control. In terms of actually using this component, that means we're expecting the consumer to keep track of when a date gets selected and tell the user about it with their own aria states.

A better fix might be to add an "invisible" aria-live label at the root of the control with the selected date information in it. That way, the immediately obvious visual information of the blue highlight is given to screen reader users as soon as they focus into the control, and when the selection changes, it gets announced.

lorejoh12:
For some reason the PR merging didn't auto close the issues... closing, this should be fixed now.

Issue #11141, opened 2019-11-11
Calendar: Two months in the month Picker calendar is having aria-selected property as true when only one is selected by user.


dzearing:
Tracking all of the issues brought up with the 03/15 accessibility pass on Calendar fabric control
Details on individual bugs can be found on [Sharepoint](https://microsoft-my.sharepoint.com/:x:/r/personal/v-amaver_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B8cfe67c5-ce76-4083-b60e-4190dbd38206%7D&action=default&uid=%7B8CFE67C5-CE76-4083-B60E-4190DBD38206%7D&ListItemId=1872&ListId=%7B395E1ECB-5002-440E-81C5-6A090E98077C%7D&odsp=1&env=prodbubble)

lorejoh12:
We currently have two different "selected" dates. "Today" is auto-highlighted with dark blue, and the "selected date" (selected by the user) is light blue. I guess semantically we only want the user selected date, but doing so means we're losing out on informing the user about "today" highlighting. I guess we could just toss that information into the aria label for that cell.

lorejoh12:
Upon further inspection, the real bug here is that for month view specifically, we're marking "today" as selected even when it's not highlighted. This has been fixed in date-time, but needs to be fixed in OUFR Calendar as well.

lorejoh12:
For some reason the PR merging didn't auto close the issues... closing, this should be fixed now.

Issue #10908, opened 2019-10-18
[Accessibility] Date pickers are not accessible using shortcut keys (PageUp, PageDown, Ctrl+PageUp & Ctrl+PageDown


SyedRehmanRazvi:
User Experience:

Users who rely on screen reader are getting affected as Date pickers should be accessible using  shortcut keys.
 
Repro Steps:
Step 1: Turn on narrator and Open latest fabric Ui URL in Edge browser.
Step 2: Navigate to Date picker page
Step 3: Navigate to any date picker control and verify whether shortcut keys are working.

Actual:
Date pickers are not accessible using  shortcut keys (PageUp, PageDown, Ctrl+PageUp & Ctrl+PageDown) .

Expected:
Date pickers should be accessible using  shortcut keys (PageUp, PageDown, Ctrl+PageUp & Ctrl+PageDown) .
i.e. 
 PageUp: To navigate to Next month
 PageDown: To navigate to Previous month
 Ctrl+PageUp: To navigate to Next year
Ctrl+PageDown To navigate to Previous year.
 
![image](https://user-images.githubusercontent.com/53210862/67101442-41799980-f1df-11e9-98e2-5cf940eceb3b.png)

marygans:
Hi @SyedRehmanRazvi thank you for submitting this issue!

I was able to repro this issue. Im assuming the expectation on behavior should match the [aria example](http://w3c.github.io/aria-practices/examples/dialog-modal/datepicker-dialog.html). 

Will ensure this gets added, thanks!

ecraig12345:
It's probably best if @lorejoh12 or another calendar team member works on this, rather than taking it on ourselves. They would also be better able to answer whether it's something they intend to support.

lorejoh12:
It's a feature request, but not one that I think would be particularly difficult to add. I'll add a note to our backlog to see if/when we would be able to pick this up.

lorejoh12:
For some reason the PR merging didn't auto close the issues... closing, this should be fixed now.

PR #11254, opened 2019-11-20
[Date-time] date time accessibility pass

### LINKED ISSUE #11143: Calendar: Upon Selecting the date, calendar dialog box blinks and focus moves to the arrow.
Tracking all of the issues brought up with the 03/15 accessibility pass on Calendar fabric control
Details on individual bugs can be found on [Sharepoint](https://microsoft-my.sharepoint.com/:x:/r/personal/v-amaver_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B8cfe67c5-ce76-4083-b60e-4190dbd38206%7D&action=default&uid=%7B8CFE67C5-CE76-4083-B60E-4190DBD38206%7D&ListItemId=1872&ListId=%7B395E1ECB-5002-440E-81C5-6A090E98077C%7D&odsp=1&env=prodbubble)

lorejoh12:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11143, #11144, #11142, #11141, #10908, #10352, #8305

#### Description of changes

#11143 Upon selecting the date, calendar dialog box blinks and focus moves to the arrow
the part with the focus moving to the arrow was already fixed in date-time, but here we also fix the issue of the focus not automatically moving to the day picker after selecting a month in the month picker

#11144 Name is not defined for the dialog which appears on pressing enter on click for calendar button
#8305 Today/Selected date is not read by screen reader
fixed by adding an aria-label to the root of the control which contains information for today/selected date. Will be read when first focusing into the control.

#11142 Information about the change of week is not narrated by screen reader whenever user press Enter on previous/next button
Added an aria-live region to the root Calendar component with the selected date information, which announces whenever date selection is changed, even from external sources like the previous/next button

#11141 Two months in the month Picker calendar is having aria-selected property as true when only one is selected by the user
Removed the duplicate aria-selected which was being applied to the current month (rather than selected month, which I left). This was only in OUFR, date-time was already correct.

#10908 Date pickers are not accessible using shortcut keys (PageUp, PageDown, Ctrl+PageUp & Ctrl+PageDown)
Added key listeners for the above keys to navigate forwards/backwards by a month/year

#10352 Name is not defined for the selected year button
Added an aria-label for the month picker header and year picker header buttons, telling you that clicking on the button will switch you to the other picker

In addition, consolidates all of the various default strings throughout date-time Calendar into a single defaults.ts file (and another one for DatePicker, which reads from it for the common strings). Now new strings only need to be added in 1 place instead of many.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11254)

### DIFF
--- a/change/@uifabric-date-time-2019-11-19-16-58-04-lorejoh12-dateTimeAccessibility.json
+++ b/change/@uifabric-date-time-2019-11-19-16-58-04-lorejoh12-dateTimeAccessibility.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Accessibility fixes for date-time Calendar, adding new strings for new aria labels and fixing aria-live regions",
+  "packageName": "@uifabric/date-time",
+  "email": "jolore@microsoft.com",
+  "commit": "2e06318622eba26ed36ded82a07969a71e966e6e",
+  "date": "2019-11-20T00:57:34.064Z"
+}
--- a/change/office-ui-fabric-react-2019-11-19-16-58-04-lorejoh12-dateTimeAccessibility.json
+++ b/change/office-ui-fabric-react-2019-11-19-16-58-04-lorejoh12-dateTimeAccessibility.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "fixing double selection in Calendar month picker",
+  "packageName": "office-ui-fabric-react",
+  "email": "jolore@microsoft.com",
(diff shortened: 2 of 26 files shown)


micahgodbolt:
@evlevy can you give a thumbs up on this?

PR #11257, opened 2019-11-20
align color system

### LINKED ISSUE #11256: Tooltip should use same color system to describe its color
### Environment Information

- **Package version(s)**: 7.x

### See code here:
https://github.com/OfficeDev/office-ui-fabric-react/blob/13872671d45d1793700081d4cfab3e8f5b7fc5c3/packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.styles.ts#L17

### and here:
https://github.com/OfficeDev/office-ui-fabric-react/blob/13872671d45d1793700081d4cfab3e8f5b7fc5c3/packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.styles.ts#L41

in a single component, it's using a different color system to describle.

In one color system, it should be fine, however when you are mixing two color system, color contrast may not be enough.

in fabric 6, it's correct.

### why would i raise the question
I got some questions on the theme:

what's the relationship between palette and semantic color?

If palette is fixed, did you think consumer can modify part of the semantic color without modifying the palette?

from the code and its comment here:
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/styling/src/styles/theme.ts#L182

the mapping only existed when semantic color is not provided.

And my guesses are you will allow consumer to define their semantic color based on their understanding of the mapping between palette and semantic color, right?

If this is true, the questions is that if you mixed using palette and semantic color, how can customer define their own semantic color based on their understanding?

We can't analyze every component to make sure every combination of semantic color and palette work well. The only thing that we care is the combination in semantic color is good.

So i filed this PR to raise the problem. We should not mix-use the two color system.

Why did i use palette?

Because it was used in fabric 6.

qianghuang94:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11256

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11257)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.styles.ts
@@ -3,7 +3,7 @@ import { AnimationClassNames } from '../../Styling';
 
 export const getStyles = (props: ITooltipStyleProps): ITooltipStyles => {
   const { className, beakWidth = 16, gapSpace = 0, maxWidth, theme } = props;
-  const { palette, fonts, effects } = theme;
+  const { semanticColors, fonts, effects } = theme;
 
   // The math here is done to account for the 45 degree rotation of the beak
   const tooltipGapSpace = -(Math.sqrt((beakWidth * beakWidth) / 2) + gapSpace);
@@ -14,7 +14,7 @@ export const getStyles = (props: ITooltipStyleProps): ITooltipStyles => {
       theme.fonts.medium,
       AnimationClassNames.fadeIn200,
       {
-        background: palette.white,
+        background: semanticColors.menuBackground,
         boxShadow: effects.elevation8,
         padding: '8px',
         maxWidth: maxWidth,
@@ -38,7 +38,7 @@ export const getStyles = (props: ITooltipStyleProps): ITooltipStyles => {
       {
         position: 'relative',
         zIndex: 1,
-        color: palette.neutralPrimary,
+        color: semanticColors.menuItemText,
         wordWrap: 'break-word',
         overflowWrap: 'break-word',
         overflow: 'hidden'
--- a/packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.styles.ts
@@ -3,7 +3,7 @@ import { AnimationClassNames } from '../../Styling';
 
 export const getStyles = (props: ITooltipStyleProps): ITooltipStyles => {
   const { className, beakWidth = 16, gapSpace = 0, maxWidth, theme } = props;
-  const { palette, semanticColors, fonts, effects } = theme;
+  const { palette, fonts, effects } = theme;
 
   // The math here is done to account for the 45 degree rotation of the beak
   const tooltipGapSpace = -(Math.sqrt((beakWidth * beakWidth) / 2) + gapSpace);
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.styles.ts
```diff
@@ -14,7 +14,7 @@ export const getStyles = (props: ITooltipStyleProps): ITooltipStyles => {
       theme.fonts.medium,
       AnimationClassNames.fadeIn200,
       {
-        background: semanticColors.menuBackground,
+        background: palette.white,
```
micahgodbolt: rather than moving to palette color here, how about changing the text color menuItemText.

qianghuang94: @micahgodbolt if we moved, it can be a break change to migrating from 6.

micahgodbolt: menuItemText is based off of palette.neutralPrimary....so updating neutralPrimary would still update menuItemText.

If there's hard problem with this, thats okay. I just wanted to bring it up.

kenotron: Fabric prefers semanticColors when possible. We should change both to semanticColors instead.

dzearing: I'm unclear about the purpose of this change in tooltip for Fabric 7. The description of the PR is vague, and there are no screenshots to illustrate what's going on. 

Is `menuBackground` the wrong color? As others have said, switching to a palette color brings no semantic meaning, so flipping white out to black might not be correct for tooltip backgrounds.

qianghuang94: @dzearing @kenotron @micahgodbolt 

I got some questions on the theme:

1. what's the relationship between palette and semantic color?

2. If palette is fixed, did you think consumer can modify part of the semantic color without modifying the palette?

from the code and its comment here:
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/styling/src/styles/theme.ts#L182

the mapping only existed when semantic color is not provided.

And my guesses are you will allow consumer to define their semantic color based on their understanding of the mapping between palette and semantic color, right?

If this is true, the questions is that if you mixed using palette and semantic color, how can customer define their own semantic color based on their understanding?

We can't analyze every component to make sure every combination of semantic color and palette work well. The only theme we care is the combination in semantic color is good (used by it's meaning).

So i filed this PR to raise the problem. We should not mix-use the two color system.

Why would i use palette?

Because it was used in fabric 6.

qianghuang94:
I got some questions on the theme:

1. what's the relationship between palette and semantic color?

2. If palette is fixed, did you think consumer can modify part of the semantic color without modifying the palette?

from the code and its comment here:
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/styling/src/styles/theme.ts#L182

the mapping only existed when semantic color is not provided.

And my guesses are you will allow consumer to define their semantic color based on their understanding of the mapping between palette and semantic color, right?

If this is true, the questions is that if you mixed using palette and semantic color, how can customer define their own semantic color based on their understanding?

We can't analyze every component to make sure every combination of sematic color and palette work well. The only theme we care is the combination in semantic color is good (used by it's meaning).

So i filed this PR to raise the problem. We should not mix-use the two color system.

Why would i use palette?

Because it was used in fabric 6.

packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.styles.ts
ecraig12345: The use of `palette` in Fabric 6 is a legacy thing, not an example to follow. We should use semantic colors whenever possible moving forward, meaning the tooltip text color should be changed to `semanticColors.menuItemText` (and revert this line).

qianghuang94:
@kenotron @aneeshack4 I have updated the color, please help to review it

PR #11258, opened 2019-11-20
Fix dropdown misalignment

betrue-final-final:
#### Pull request checklist

- [ ] Addresses an issue found by @dzearing 

Added focus rectangle as an after element to fix shifting control up and left by 1px.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11258)


jdhuntington:
Looks like this is missing a change file (and potentially snapshots); will push those momentarily.

joschect:
@betrue-final-final or @dzearing do we know what happened to the focus mixin? Are we no longer using it?

kenotron:
Is that screener diff a click on a disabled item? I don't think we would be doing a outline on a disabled item?

ecraig12345:
Is this waiting on anything or is it ready to go? (I just triggered a new build.)

PR #11270, opened 2019-11-22
Pivots move to overflow set when limited space - Initial PR for design review

### LINKED ISSUE #4066: Pivot: Support overflowing pivots
Hi all,

I think this would be nice. Do we have any plan to handle Pivots are overflowed.?

Thank in advance,
Quang Le

msnyder-msft:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #4066

#### Description of changes
Today, pivots are not very responsive. When many tabs are shown and take up a lot of horizontal space, they don't adapt when the page/container size changes. 
This change set is intended to add a resize group around the pivot and pushes items into an overflow set when there isn't enough room to show all the tabs.

This is just an initial change set to get some input on designs and things to consider. I'd love to hear your thoughts.
I also haven't touched the tests, I know there's going to be a bunch of work to do there.

#### Focus areas to test
Pivots

image
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11270)

### DIFF
--- a/change/office-ui-fabric-react-2019-11-22-14-30-15-msnyder-PivotOverflow.json
+++ b/change/office-ui-fabric-react-2019-11-22-14-30-15-msnyder-PivotOverflow.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Enabling pivots to move to overflow set based on available space",
+  "packageName": "office-ui-fabric-react",
+  "email": "msnyder@microsoft.com",
+  "commit": "58165b3f75cf267fa1bfddd679f4471c6850ecd7",
+  "date": "2019-11-22T22:30:15.528Z"
+}
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -6128,11 +6128,20 @@ export interface IPickerItemProps<T> extends React.AllHTMLAttributes<HTMLElement
 // @public (undocumented)
 export interface IPivot {
     focus(): void;
+    remeasure(): void;
+}
+
+// @public (undocumented)
+export interface IPivotData {
+    cacheKey: string;
+    items: IPivotItemProps[];
+    overflowItems: IOverflowSetItemProps[];
 }
 
 // @public (undocumented)
 export interface IPivotItemProps extends React.HTMLAttributes<HTMLDivElement> {
     ariaLabel?: string;
+    cacheKey?: string;
     componentRef?: IRefObject<{}>;
     headerButtonProps?: {
         [key: string]: string | number | boolean;
@@ -6145,12 +6154,14 @@ export interface IPivotItemProps extends React.HTMLAttributes<HTMLDivElement> {
     // @deprecated
     linkText?: string;
     onRenderItemLink?: IRenderFunction<IPivotItemProps>;
+    renderedInOverflow?: boolean;
 }
 
 // @public (undocumented)
 export interface IPivotProps extends React.ClassAttributes<PivotBase>, React.HTMLAttributes<HTMLDivElement> {
     className?: string;
     componentRef?: IRefObject<IPivot>;
+    dataDidRender?: (renderedData: any) => void;
     defaultSelectedIndex?: number;
     defaultSelectedKey?: string;
     getTabId?: (itemKey: string, index: number) => string;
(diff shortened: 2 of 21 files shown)


micahgodbolt:
@msnyder-msft can you add a changefile? Just run `yarn change` from the root

micahgodbolt:
i know this just an early code review, but it'd be great to get it passing so that we can see the rendered output as well. Just need to add a changefile to get ci kicked off

msnyder-msft:
> i know this just an early code review, but it'd be great to get it passing so that we can see the rendered output as well. Just need to add a changefile to get ci kicked off

Sorry for the delay. It's pushed now. Didn't realize that was a problem for design review.

micahgodbolt:
@msnyder-msft its not really REQUIRED for a review, but we currently fail builds that don't have one, and its great to see the finished build in our PR deploy site.

msnyder-msft:
> @msnyder-msft its not really REQUIRED for a review, but we currently fail builds that don't have one, and its great to see the finished build in our PR deploy site.

That's really neat! I didn't realize it would do that. Will it still deploy if snapshots are failing? Because I haven't updated the tests yet.

micahgodbolt:
Looks like it didn't resolve all of the snap issues. cd into packages/office-ui-fabric-react and run `yarn update-snapshots`.

msnyder-msft:
> Looks like it didn't resolve all of the snap issues. cd into packages/office-ui-fabric-react and run `yarn update-snapshots`.

Thanks! I pushed the snapshots now.

micahgodbolt:
@msnyder-msft can you push those updated snaps so we can get a PR preview?

micahgodbolt:
oh, snaps seem done. there are some linting issues popping up about your imports

image

msnyder-msft:
@micahgodbolt I've pushed the fixes for those 2 files. We'll see if it's happier now. The output from that lint command is so noisy that it's really hard to see where the actual issues are at so thanks for highlighting the important stuff.
It also keeps complaining about having to force stop jest. Is that normal?

micahgodbolt:
I saw issues about physical file imports. Don't include the file extension in your imports.

msnyder-msft:
> I saw issues about physical file imports. Don't include the file extension in your imports.

@micahgodbolt I just pushed an update to fix the errors about the API. I think everything should be happy now except for jest which is complaining about force stopping.
image

micahgodbolt:
Looks like we've got a passing build :D

We are getting visual test errors, and I can tell you why. 

image

Now that we're using resizeGroup to fit the pivot into the given space......this means that if the parent is not a block level element, it'll keep shrinking to fit the contents.....and the contents will keep shrinking to fit the parent. 

So you get a fully collapsed pivot.

This will be a concern going forward, as this change could be considered breaking as pivots no longer automatically take up all the space expected.

msnyder-msft:
@micahgodbolt I see you started the review thread over email. I had seen those failing tests but before investing too much in fixing them I wanted to be sure the major design pattern was acceptable to folks. Once folks are happy with that, I'll get those discrepancies resolved.

micahgodbolt:
@msnyder-msft trying to remember if you were on this thread about our thoughts here, but to sum up:

Making this change is probably going to be too disruptive for anyone expecting pivots to take up 100% of the space that it can.

What we'd like to recommend is to update the pivots to allow for an onRender function to replace the existing pivots with custom pivots that you could customize in the way that you did. We can even include an example of how you can do it (depending on complexity).

So basically, take the work you did into an onRenderPivotItems (or something similar), and update the pivot to support that custom render.

Happy to give you a hand with how this would work. I know it has come up a few times.

msnyder-msft:
@micahgodbolt thanks for the summary here. I was waiting for an outcome on that thread before moving forward. I understand the general idea but not sure I am completely clear where this logic would live now. 
The first part of your solution (expose `onRender`) sounds like you are expecting the consumer to do all the work I've created here. But then the second part sounds like we have this logic defined in the pivot component and they can somehow access that function `onRenderPivotItems` and pass that into `onRender`'?
Maybe it would be easiest to have a quick call to discuss unless there's a simpler way to explain here.

micahgodbolt:
@msnyder-msft it'd look something like this

```jsx
    
        <Pivot
          onRenderPivotLinks={(linkCollection, selectedKey) => {
            return <ResizeablePivotLinks linkCollections={linkCollection} selectedKey={selectedKey} />;
          }}
          linkFormat={PivotLinkFormat.tabs}
        >
          <PivotItem headerText="Foo">
            <Label>Pivot #1</Label>
          </PivotItem>
          <PivotItem headerText="Bar">
            <Label>Pivot #2</Label>
          </PivotItem>
          <PivotItem headerText="Bas">
            <Label>Pivot #3</Label>
          </PivotItem>
          <PivotItem headerText="Biz">
            <Label>Pivot #4</Label>
          </PivotItem>
        </Pivot>

```

micahgodbolt:
Also, you don't necessarily need to write the <ResizeablePivotLinks> control in Fabric. We just need the Pivot set up to be flexible enough to support that custom render function

msnyder-msft:
@micahgodbolt ok that makes sense. So you're saying to make limited changes in Fabric and that `ResizeablePivotLinks` component is something I write in my own repo?

micahgodbolt:
Bingo :)

How can we modify the current pivot without breaking existing API to support rendering a customized set of links. Then you can hack on that in your own repo, and if you come up with a really solid, stable solution we can look at adding that to the OUFR repo

micahgodbolt:
@msnyder-msft is take a new approach that will make the change less disruptive. I'm going to close this PR down until he's ready for a new review.

PR #11280, opened 2019-11-22
Combobox disabled options

### LINKED ISSUE #11193: [Accessibility][Keyboard Navigation]: Disabled Combo box option 'Option J' is selectable using keyboard.
Split out of issue #6920

### LINKED ISSUE #11203: [Accessibility][Keyboard navigation]: Arrow keys navigation is not working properly on 'Basic uncontrolled example (allowFreeform: F, AutoComplete: F)' combo box.
Split out of issue #6920

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11193 Fixes #11203

#### Description of changes

Keyboard navigation and freeform were able to select disabled options. Now we check if the option is disabled right before setting the new value. If the option is disabled we noop and return to the previously selected value.

Also, seems prettier config changed a bit, so just note line 872

I also added in a fix to show the correct value for non freeform/autocomplete scenarios, which was dependent on the disabled fix.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11280)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
@@ -596,6 +596,10 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
           // the matched option's index
           index = currentPendingValueValidIndex;
           displayValues.push(this._normalizeToString(currentPendingValue));
+        } else if (!this.state.isOpen && currentPendingValue) {
+          displayValues.push(
+            this._indexWithinBounds(currentOptions, index) ? currentPendingValue : this._normalizeToString(suggestedDisplayValue)
+          );
         } else {
           displayValues.push(
             this._indexWithinBounds(currentOptions, index) ? currentOptions[index].text : this._normalizeToString(suggestedDisplayValue)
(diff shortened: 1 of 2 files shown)


micahgodbolt:
@msft-github-bot wait for @jspurlin before merging this in

micahgodbolt:
@jspurlin i was going to do #11203 separately but it didn't work properly without the disabled fix.

My one concern moving forward is that the styling of the disabled option isn't obvious. 

image

Not sure how design would want to designate that the current value is disabled and removing focus will revert to the prev selected value.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -868,7 +872,8 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
     // there is nothing to do
     if (this.props.multiSelect || selectedIndices.length < 1 || (selectedIndices.length === 1 && selectedIndices[0] !== index)) {
       const option: IComboBoxOption = { ...currentOptions[index] };
-      if (!option) {
+      // if option doesn't existing, or option is disabled, we noop
```
jspurlin: >existing [](start = 27, length = 8)

NIT: `exist`

PR #11296, opened 2019-11-25
Add Data Prop to INavLinkGroup

nuh1234:
#### Pull request checklist

- [ X] Include a change request file using `$ yarn change`

#### Description of changes

Added an additional field to INavLinkGroup, this is now on par with INavLink

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11296)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Nav/Nav.types.ts
+++ b/packages/office-ui-fabric-react/src/components/Nav/Nav.types.ts
@@ -154,6 +154,11 @@ export interface INavLinkGroup {
    * ARIA label when group is collapsed and can be expanded.
    */
   collapseAriaLabel?: string;
+
+  /**
+   * (Optional) Any additional properties to apply to the rendered groups.
+   */
+  [propertyName: string]: any;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/Nav/Nav.types.ts
```diff
@@ -154,6 +154,11 @@ export interface INavLinkGroup {
    * ARIA label when group is collapsed and can be expanded.
    */
   collapseAriaLabel?: string;
+
+  /**
+   * (Optional) Any additional properties to apply to a group.
+   */
+  [propertyName: string]: any;
```
jdhuntington: Would you mind adding some examples of scenarios this enables? Prefer to avoid opening up the types with an `any` like this.

ThomasMichon: I would advise against allowing direct attached props, in favor of forcing them through a `data?: any` prop, similar to what we've done with other components. Adding an indexer winds up causing TypeScript to degrade the type to `any`, which can mask typos and other bugs when using this interface.

nuh1234: Agreed I prefer not having to use `any` as well, I have updated it to groupData @ThomasMichon 

@jdhuntington An example scenario would be attaching additional props for user later on in the onRender function of a group header.

PR #11299, opened 2019-11-25
Use closeButtonAriaLabel as   dialog component close button tool tip

RajeshGoriga:
#### Pull request checklist

#### Description of changes
Expose prop for Dialog component for close button tool tip

#### Focus areas to test
Dialog close button 

#### Screenshot

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11299)

### DIFF
--- a/change/office-ui-fabric-react-2019-11-25-18-02-32-user-v-satgar-ExposePropForDialog.json
+++ b/change/office-ui-fabric-react-2019-11-25-18-02-32-user-v-satgar-ExposePropForDialog.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Expose prop for dialog component close button tooltip",
+  "packageName": "office-ui-fabric-react",
+  "email": "email not defined",
+  "commit": "dbcbaed1f306bbe5f18433ef54165b21b1600fad",
+  "date": "2019-11-25T12:32:32.409Z"
+}
--- a/change/office-ui-fabric-react-2019-11-25-18-02-32-user-v-satgar-ExposePropForDialog.json
+++ b/change/office-ui-fabric-react-2019-11-25-18-02-32-user-v-satgar-ExposePropForDialog.json
@@ -1,8 +1,8 @@
 {
-  "type": "minor",
-  "comment": "Expose prop for dialog component close button tooltip",
+  "type": "patch",
+  "comment": "Dialog: Use closeButtonAriaLabel for close button tooltip",
   "packageName": "office-ui-fabric-react",
-  "email": "email not defined",
+  "email": "v-ragor@microsoft.com",
   "commit": "dbcbaed1f306bbe5f18433ef54165b21b1600fad",
   "date": "2019-11-25T12:32:32.409Z"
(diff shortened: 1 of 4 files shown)


packages/office-ui-fabric-react/src/components/Dialog/DialogContent.types.ts
```diff
@@ -42,6 +42,11 @@ export interface IDialogContentProps extends React.ClassAttributes<DialogContent
    */
   showCloseButton?: boolean;
 
+  /**
+   * Add tooltip for 'x' close button in the upper-right corner
+   */
+  closeButtonToolTip?: string;
```
jdhuntington: This is likely to change the API and the snapshots, which need to be updated for the build to pass.

Are you able to run these in the `packages/office-ui-fabric-react` directory?

```sh
yarn update-api
yarn update-snapshots
```

ecraig12345: Instead of adding a new prop, we should probably just use the existing `closeButtonAriaLabel` for the title as well. Not completely ideal but other components (Panel, TeachingBubbleContent) already do this.

RajeshGoriga: updated as you suggested using closeButtonAriaLabel as title

change/office-ui-fabric-react-2019-11-25-18-02-32-user-v-satgar-ExposePropForDialog.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
```
ecraig12345: This should be patch now
```suggestion
  "type": "patch",
```

change/office-ui-fabric-react-2019-11-25-18-02-32-user-v-satgar-ExposePropForDialog.json
```diff
@@ -0,0 +1,8 @@
...
+  "comment": "Expose prop for dialog component close button tooltip",
```
ecraig12345: This should be updated too
```suggestion
  "comment": "Dialog: Use closeButtonAriaLabel for close button tooltip",
```

change/office-ui-fabric-react-2019-11-25-18-02-32-user-v-satgar-ExposePropForDialog.json
RajeshGoriga: done

change/office-ui-fabric-react-2019-11-25-18-02-32-user-v-satgar-ExposePropForDialog.json
RajeshGoriga: done

PR #11303, opened 2019-11-25
Only update ExtendedPicker state on input events that are not mid-composition

Adjective-Object:
#### Pull request checklist

- ~~[ ] Addresses an existing issue: Fixes #0000~~

#### Description of changes

This is for a bug coming from outlook.

when composing input to the extendedpicker, we react to inputChanged events in order to notify the floating picker of its new search string, and query for new results. However, that doesn't align with the intent of a change coming from mid-IME composition. Pickers should not search with mid-composing inputs, since those change under the composition box & are usually not relevant until composed.

#### Focus areas to test
Autofill:
- run the test suite

ExtendedPickers.
Test plan:
1. run `yarn start` in `packages/office-ui-fabric-react`
2. confirm existing behaviour for non-composing input is the same(example: english language)
3. Switch to a composing input (e.g. hiragana jp)
4. compose a search
5. observe that, while still composing, the search results do not update,
6. complete the composition
7. observe the updated search results.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11303)


Adjective-Object:
ah, my bad. forgot to push a changefile. One moment.

Adjective-Object:
@aneeshack4 the yarn build is failing with this error which doesn't _seem_ related. Any advice?

image

aneeshack4:
@Adjective-Object It looks like it is actually related:
image
Did you make sure to update your branch with latest master? Also make sure to update the api file (`yarn update-api`) since it looks like you've updated the types file also. I don't think `yarn update-snapshots` is required because there are no stories related changes and all the snapshots are passing so I think if you merge master & run a local build it should pass.

packages/office-ui-fabric-react/src/components/Autofill/Autofill.tsx
```diff
@@ -267,28 +270,28 @@ export class Autofill extends BaseComponent<IAutofillProps, IAutofillState> impl
     }
   }
 
-  private _notifyInputChange(newValue: string): void {
+  private _notifyInputChange(newValue: string, wasComposing: boolean): void {
     if (this.props.onInputValueChange) {
-      this.props.onInputValueChange(newValue);
+      this.props.onInputValueChange(newValue, wasComposing);
```
joschect: Could we name this variable just `composing` I think it clarifies what it means and then consuming functions can choose to consume the value when `composing` is false.

Adjective-Object: sure. I don't generally like using non (is/was/should) names for booleans, but since the is/was is ambiguous I'm fine with it

Adjective-Object:
@joschect @ecraig12345  is there anything left here on my side?

packages/office-ui-fabric-react/src/components/Autofill/Autofill.tsx
```diff
@@ -160,21 +160,24 @@ export class Autofill extends BaseComponent<IAutofillProps, IAutofillState> impl
...
     this._tryEnableAutofill(inputValue, this.value, false, true);
     this._isComposing = false;
     // Due to timing, this needs to be async, otherwise no text will be selected.
     this._async.setTimeout(() => {
-      // Call getCurrentInputValue here again since there can be a race condition where this value has changed during the async cal
-      this._updateValue(this._getCurrentInputValue());
+      // it's technically possible that the value of _isComposing is reset during this timeout,
+      // so explicitly trigger this with composing=true here, since it is supposed to be the
+      // update for composition end
+      this._updateValue(this._getCurrentInputValue(), true);
```
joschect: Since it's the end, isn't composing false at this point? Composition has completed and this is final value that should be used?

Adjective-Object: Anyway, I think this is fine, but I'm going to move from `this._isComposing` to `(ev.nativeEvent as any).isComposing` where possible, since I think in the above edge case _isComposing can't be trusted.

PR #11311, opened 2019-11-26
Joou/embedded panel

### LINKED ISSUE #10918: Panel can be embedded to parent container and inherit parent's height

Hi, currently Panel component overlays the whole DOM and its top height is always the top of window, this does matter when we have dynamic message bars at top. We want the panel as part of content instead of whole page, but so it's diffcult to calculate and update the latest height of panel, just like this.

image

My code is also the one of all components, besides message bars there are also other components at top, so can you make a type of Panel component which can be embedded to parent container?

[Panel](https://developer.microsoft.com/en-us/fabric#/controls/web/panel)

No discuss yet.

Null.

Joouis:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10918

#### Description of changes

Implementing a new type named `smallRelativeFar`, this type can make the panel showed under its father element container.

demo

#### Focus areas to test

It's my first time to contribute to Fabric UI project, I have ran unit test locally and did some manual tests, maybe A11Y test should be focused because I'm new to this area.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11311)

### DIFF
--- a/change/office-ui-fabric-react-2019-11-26-14-46-49-joou-embedded-panel.json
+++ b/change/office-ui-fabric-react-2019-11-26-14-46-49-joou-embedded-panel.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Fix feature request #10918 by implementing a new panel type",
+  "packageName": "office-ui-fabric-react",
+  "email": "kuzhou@microsoft.com",
+  "commit": "8efd18b6ae4457a588b96641a1fd961496478ca3",
+  "date": "2019-11-26T06:46:49.680Z"
+}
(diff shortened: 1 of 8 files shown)


Joouis:
@micahgodbolt  Hi Micah, thanks for your reply in parent issue, it's really useful. We can hold on this request while I'm trying to use `layerProps` to resolve my problem.

micahgodbolt:
closing for now. can re-open if layerProps doesn't solve issue.

PR #11314, opened 2019-11-26
Coachmark beak persistence

### LINKED ISSUE #8221: Enable option to continue showing beak when Coachmark is expanded

We're using a Coachmark with TeachingBubbleContent, but our designers want to keep showing the beak when the teaching bubble expands. As of #6158, the beak is only rendered when collapsed. There should be an option for this behavior.

Coachmark

**Have you discussed this feature with our team, and if so, who**

From Vitalie Braga, "This is actually something that needs to be fixed in the default behavior because the design redlines are leaving the beak in place after expansion. So feel free to file an issue on Github and the Shield engineer will take care of it)))"

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #8221

#### Description of changes

coachmark

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11314)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Coachmark/Coachmark.types.ts
+++ b/packages/office-ui-fabric-react/src/components/Coachmark/Coachmark.types.ts
@@ -178,7 +178,7 @@ export interface ICoachmarkProps extends React.ClassAttributes<CoachmarkBase> {
   preventDismissOnLostFocus?: boolean;
 
   /**
-   * If true then the Coachmark beak will always be visible as long as Coachmark is visible
+   * If true then the Coachmark beak (arrow pointing towards target) will always be visible as long as Coachmark is visible
    * @defaultvalue false
    */
   persistentBeak?: boolean;
--- a/packages/office-ui-fabric-react/src/components/Coachmark/Coachmark.types.ts
+++ b/packages/office-ui-fabric-react/src/components/Coachmark/Coachmark.types.ts
@@ -178,6 +178,12 @@ export interface ICoachmarkProps extends React.ClassAttributes<CoachmarkBase> {
   preventDismissOnLostFocus?: boolean;
 
   /**
+   * If true then the Coachmark beak will always be visible as long as Coachmark is visible
+   * @defaultvalue false
+   */
+  persistentBeak?: boolean;
+
+  /**
    * If true then focus will not be set to the Coachmark when it mounts. Useful in cases where focus on coachmark
    * is causing other components in page to dismiss upon losing focus.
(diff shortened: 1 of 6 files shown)


packages/office-ui-fabric-react/src/components/Coachmark/Coachmark.types.ts
```diff
@@ -177,6 +177,12 @@ export interface ICoachmarkProps extends React.ClassAttributes<CoachmarkBase> {
    */
   preventDismissOnLostFocus?: boolean;
 
+  /**
+   * If true then the Coachmark beak will always be visible as long as Coachmark is visible
+   * @defaultvalue false
+   */
+  persistentBeak?: boolean;
```
jdhuntington: Nit: I realize this is prior art, but it might be nice to clarify what we mean by `beak`, as it's not clearly defined in the existing docs for coachmark.

packages/office-ui-fabric-react/src/components/Coachmark/examples/Coachmark.Basic.Example.tsx
```diff
@@ -97,7 +102,14 @@ export class CoachmarkBasicExample extends BaseComponent<{}, ICoachmarkBasicExam
             ]}
             onChange={this._onDropdownChange}
           />
-        </div>
+          <Toggle
```
jdhuntington: 👍 for the feature, but do we gain much by adding this example? It feels like this might confuse the a11y story for this example; would be nice to make sure the example only shows the recommended usage.

micahgodbolt: i'm good either way if we want to remove it.

packages/office-ui-fabric-react/src/components/Coachmark/Coachmark.types.ts
micahgodbolt: updated

packages/office-ui-fabric-react/src/components/Coachmark/examples/Coachmark.Basic.Example.tsx
aneeshack4: I think the example is helpful - the visual treatment being applied with the toggle clarifies what this newly added boolean is for.

PR #11315, opened 2019-11-26
Update Modal KeyUp to not only register in componentWillReceiveProps

SavannahBourgeois:
#### Pull request checklist

- [x ] Include a change request file using `$ yarn change`

#### Description of changes
Not all modals and dialogs mount when the page loads and show based on updating isOpen in the props form false (onMount) to true (gets it to show). Some dialogs (like the ones in Office) mount when they are ready to show and pass in isOpen is true.

However, the onKeyUp event is added in UNSAFE_componentWillReceiveProps which is not always called when the modal mounts. This means any modal that does not update the props to show and shows when they mount the component, will not register for onKeyUp events. To fix this, I added componentDidMount which checks to see if the modal is open and visible before registering for onKeyUp. Additionally, I added some checks to make sure that we only register for the event once by updating the state.

Unfortunately all of Fabric's dialog and modal examples are mounted when the page loads so they do not repro this issue. The dialogs in the office repo do not all mount when the page loads so some experience this issue when they are created/renders based on a button click.

#### Focus areas to test

Dialogs and Modals.

I have verified the dialogs/modals in fabric still open the move/close menu as before. I also verified this works in the office repo to ensure it fixes the issue we were seeing.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11315)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
@@ -28,6 +28,7 @@ export interface IDialogState {
   isInKeyboardMoveMode?: boolean;
   x: number;
   y: number;
+  hasRegisteredKeyUp?: boolean;
 }
 
 const getClassNames = classNamesFunction<IModalStyleProps, IModalStyles>();
@@ -79,8 +80,11 @@ export class ModalBase extends BaseComponent<IModalProps, IDialogState> implemen
           isOpen: true
         });
         // Add a keyUp handler for all key up events when the dialog is open
-        if (newProps.dragOptions) {
+        if (newProps.dragOptions && !this.state.hasRegisteredKeyUp) {
           this._events.on(window, 'keyup', this._onKeyUp, true /* useCapture */);
+          this.setState({
+            hasRegisteredKeyUp: true
+          });
         }
       } else {
         // Modal has been opened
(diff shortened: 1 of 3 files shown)


SavannahBourgeois:
@jspurlin as FYI

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
```diff
@@ -79,8 +80,11 @@ export class ModalBase extends BaseComponent<IModalProps, IDialogState> implemen
           isOpen: true
         });
         // Add a keyUp handler for all key up events when the dialog is open
-        if (newProps.dragOptions) {
+        if (newProps.dragOptions && !this.state.hasRegisteredKeyUp) {
```
jspurlin: Should `hasRegisteredKeyUp` be state, could it just be a private member?

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
```diff
@@ -112,6 +116,17 @@ export class ModalBase extends BaseComponent<IModalProps, IDialogState> implemen
     }
   }
 
+  public componentDidMount() {
+    // Not all modals show just by updating their props. Some only render when they are mounted and pass in
+    // isOpen as true. We need to add the keyUp handler in componentDidMount if we are in that case.
+    if (this.state.isOpen && this.state.isVisible && !this.state.hasRegisteredKeyUp) {
+      this._events.on(window, 'keyup', this._onKeyUp, true /* useCapture */);
```
jdhuntington: We end up calling this in 2 places with similar conditional logic beforehand, but not identical. I can imagine a case in the future where someone will edit one of these locations but not the other. Would you be open to moving the body of `componentDidMount` to a private method that's called unconditionally from both `componentDidMount` and `UNSAFE_componentWillReceiveProps` ? That way future changes will be sure to set `hasRegisteredKeyUp` correctly and any other boolean logic will be encapsulated.

SavannahBourgeois: yeah sure, I just saw this now so let me do one more iteration

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
SavannahBourgeois: Good call, I updated it

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
SavannahBourgeois: updated now, thanks for the suggestion!

maxwellred:
Hi @SavannahBourgeois it looks like you need to update the api with `yarn update-api` to pass the build, thanks!

PR #11316, opened 2019-11-26
[Website] Update mobile docs

lynamemi:
#### Pull request checklist

#### Description of changes

- New Bottom Navigation for Android
- New Tab Bar (aka bottom nav) and Shimmer for iOS
- Android text fix
- Additional List Cell example for iOS

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11316)


apps/fabric-website/src/SiteDefinition/SiteDefinition.pages/Controls/android.tsx
```diff
@@ -87,6 +87,15 @@ export const controlsPagesAndroid: INavPage[] = [
     url: '#/controls/android/popupmenu',
     isCategory: true,
     pages: [
+      {
+        title: 'Bottom Navigation',
```
ecraig12345: Typically (at least for now) we use the name of the control itself here, so BottomNavigation (same with TabBar)

lynamemi: I think we've actually been doing the opposite so far for the mobile controls :). What's the goal for using the control name?

ecraig12345: Ah I see that now...ideally we should be doing the same for both, but that doesn't need to change right now. The main advantage I can see of using control names is to make it easier for developers to find docs on a particular control, but honestly the reason it's that way on the website now may be more "historical reasons" (it's always been that way) than anything else.

apps/fabric-website/src/pages/Controls/ListCellsPage/docs/ios/ListCellsOverview.md
```diff
@@ -1,7 +1,7 @@
 When displaying fundamental content types like files or people, it's important to provide a consistent appearance across every screen. These list cells have been designed to represent commonly used entities using a consistent system for representing their icons, metadata and secondary states.
 
 <!-- prettier-ignore-start -->
-| Single, Double, and Triple Line Cells | Other Cell Variants |
-| --- | --- |
-| <img src="https://static2.sharepointonline.com/files/fabric/fabric-website/images/controls/ios/listcells/demo-table-view-cells.png" alt="Examples of Various List Cells" style="width: 60%;" /> | <img src="https://static2.sharepointonline.com/files/fabric/fabric-website/images/controls/ios/listcells/other-cells.png" alt="Examples of Various List Cells" style="width: 100%;" /> |
+| Other Cell Variants | Cells with metadata and Grouped style | Single, Double, and Triple Line Cells |
```
ecraig12345: This should either be all title case or all sentence case. Last I remember from looking at the official style guide, we're actually supposed to use sentence case in headings, but since title case is so widely used in this section of the site, it's probably fine to stick with that for now.
```suggestion
| Other Cell Variants | Cells with Metadata and Grouped Style | Single, Double, and Triple Line Cells |
```

lynamemi: You're right, sentence case is the way to go. I'll try to keep that in made and make incremental fixes for that as I seem them.

apps/fabric-website/src/SiteDefinition/SiteDefinition.pages/Controls/android.tsx
lynamemi: Yeah that makes sense - I'll leave for now to be consistent in this section, but we can discuss holistically at some point.

PR #11317, opened 2019-11-27
Pickers forcecomplete

### LINKED ISSUE #11255: People Picker: Expose a way to force resolve text

For the people picker, when emails are entered that aren't in the suggested list, users have to press 'Enter' or 'Tab' to force resolve the e-mail. This is confusing as some users will save the form, and the text in the field is not persisted.

people-picker-commit

People Picker

Yes, Jon Schectman

[8:13 AM] Jon Schectman (WEAVER OF FABRICS)
    David Liu (FLOW) There isn't a way that you can do this right now. You could take a look at the floating pickers instead, they provide A little more fine grain control. Alternatively, we could maybe expose a force resolve function. You can force the picker to complete a suggestion if there's one available with something like this https://codepen.io/slimemold/pen/pooGWaZ?editors=0010

joschect:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11255

#### Description of changes

There were several public methods and existing props in the pickers that weren't exposed. This pr just enables users to have more control over pickers by accessing api's that, for the most part, already existed but weren't exposed.

Also cleaned up the pickers tests, just a bit.

Also added test

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11317)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
@@ -3,10 +3,10 @@ import * as ReactDOM from 'react-dom';
 import * as ReactTestUtils from 'react-dom/test-utils';
 import * as renderer from 'react-test-renderer';
 
-import { IBasePickerProps, IBasePicker } from './BasePicker.types';
+import { IBasePickerProps, IBasePicker, ValidationState } from './BasePicker.types';
 import { BasePicker } from './BasePicker';
 import { IPickerItemProps } from './PickerItem.types';
-import { resetIds } from '@uifabric/utilities';
+import { resetIds, KeyCodes } from '@uifabric/utilities';
 
 function onResolveSuggestions(text: string): ISimple[] {
   return [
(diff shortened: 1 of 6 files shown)


packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -193,11 +193,13 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends BaseComponent<
     this.setState({ suggestionsVisible: false });
   };
 
-  public completeSuggestion() {
```
khmakoto: If the signature for this is changed why are we adding the method in the `.types.ts` file instead of changing the signature?

packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
```diff
@@ -110,12 +115,103 @@ describe('BasePicker', () => {
...
+    const suggestionOptions = document.querySelectorAll('.ms-Suggestions-itemButton');
+    expect(suggestionOptions.length).toEqual(0);
+    ReactTestUtils.Simulate.keyDown(input, { which: KeyCodes.enter });
+
+    const currentPicker = picker.current!.items;
+    expect(currentPicker).toHaveLength(1);
+    expect(currentPicker![0].name).toEqual('asdff');
+  });
+
+  it('has force suggestions button', () => {
```
jdhuntington: 👍

dliu120:
Thanks! Could we cherry pick this change to V6?

PR #11326, opened 2019-11-27
Slider: getAriaValueText should return value if ariaValueText isn't supplied

### LINKED ISSUE #10858: [Accessibility][Usable]: Narrator doesn't announce the slider value along with the control name while changing the 'Number of persona' slider using arrow keys. (Bug 4)
Tracking all of the issues brought up with the 1/4 accessibility pass on Facepile control 
Details on individual bugs can be found on [SharePoint](https://microsoft-my.sharepoint.com/:x:/r/personal/v-amaver_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7Bfd43fcc5-515d-47d8-946d-0d4862a56278%7D&action=default&uid=%7BFD43FCC5-515D-47D8-946D-0D4862A56278%7D&ListItemId=1647&ListId=%7B395E1ECB-5002-440E-81C5-6A090E98077C%7D&odsp=1&env=prodbubble) 

Breakout of issue #7534

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10858

#### Description of changes

Turns out we were never returning a value to ariaValue unless user supplied the optional ariaValueText function.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11326)


packages/office-ui-fabric-react/src/components/__snapshots__/Slider.Basic.Example.tsx.shot
```diff
@@ -80,6 +80,7 @@ exports[`Component Examples renders Slider.Basic.Example.tsx correctly 1`] = `
         aria-valuemax={5}
         aria-valuemin={1}
         aria-valuenow={2}
+        aria-valuetext="2"
```
jdhuntington: This seems redundant, as `aria-valuenow` is already defined. What value is this adding?

For reference, https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuenow_attribute suggests this is only required when `valuenow` does not accurately describe the slider.

micahgodbolt: well, narrator wasn't reading aria-valuenow. Maybe it's expecting a string?

micahgodbolt: aria-valuenow is typed to be a number, but it seems narrator doesn't read that number

micahgodbolt: yeah, without the valuetext all i get is "Basic example, slider at..." and never says the number

image

Not sure if this is a narrator bug then, but the easy solution for us is to be explicit with the aria-valuetext

micahgodbolt: same experience can be found here:

http://www.oaa-accessibility.org/examplep/slider1/

narrator won't read out the numeric 'valuenow' value.

packages/office-ui-fabric-react/src/components/Slider/Slider.base.tsx
```diff
@@ -173,6 +173,8 @@ export class SliderBase extends BaseComponent<ISliderProps, ISliderState> implem
   private _getAriaValueText = (value: number | undefined): string | undefined => {
     if (this.props.ariaValueText && value !== undefined) {
       return this.props.ariaValueText(value);
+    } else if (value) {
```
dzearing: This seems wrong. First, you're creating a redundant check. Do not repeat condition checks. Second, you repeated it wrong, I believe. If the value is 0, this will not work.

Instead, do this:

```tsx
const { ariaValueText } = this.props;

if (value !== undefined) {
  return ariaValueText ? ariaValueText(value) : value.toString();
}

return undefined;
```

micahgodbolt: haha...i rewrote the code with the ternary before I looked at your comment. Felt good about coming up with a better solution...only to realize I wrote exactly what you suggested.

micahgodbolt:
pushing my fix is always helpful. @dzearing concerns addressed.

micahgodbolt:
resolved david's comment. got signoffs. merging.

PR #11329, opened 2019-11-27
Theme Designer: use Customizations.applySettings instead of loadTheme

### LINKED ISSUE #11289: Theming functions needs to be resilient across versions
We've historically had this issue come up several times now.

First example:

- The page is running Fabric v7.
- A component on the page loads Fabric v5.
- The component calls loadTheme(), creating a v5-style pre-Fluent theme without the `effects` prop.
- v7 components on the page break because they can't find `effects`.

Second example:

- The page is running Fabric v7.
- A component on the page incorrectly loads a copy of Fabric v7.
- The page sets the theme via `loadTheme()`.
- The component calls `getTheme()` but gets the wrong theme, because it's calling into a different copy of theming than the page.

While both of these are slightly in the realm of "stop doing that", Fabric should still be resilient against it, and can throw a console warning if it detects these issues.

Suggested fix: Add a version number to theme.ts. When the file loads, if the version number is the highest, register the functions to the global context. Instead of e.g. `loadTheme()` calling the code directly, it calls the version that is registered to the global context. If a second registration is detected, log a console warning.
This would need to be backported to several major versions of Fabric.

This would have been nice to have 2 years ago. Now we're just trying to find a workaround for Example 1.

aneeshack4:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11289 

#### Description of changes

Update theming documentation (theme designer's export panel & the exported codepen) to not refer developers to `loadTheme` as it is meant to be an internal utility and not encouraged to use publicly. Use `Customizations` API instead.
image
https://codepen.io/aneeshack4/pen/vYYoXzZ

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11329)


packages/office-ui-fabric-react/src/components/ThemeGenerator/ThemeGenerator.ts
```diff
@@ -81,11 +81,15 @@ export class ThemeGenerator {
   /*
    * Gets code-formatted load theme blob that can be copy and pasted.
    */
-  public static getThemeAsCode(slotRules: IThemeRules): any {
+  public static getThemeAsCode(slotRules: IThemeRules, forNewThemeDesigner: boolean): any {
```
jdhuntington: Would be nice not to add boolean flags, since it suggests the caller knows something about the implementation of this method already. Can this be a separate method, like `getThemeAsCodeForThemeDesigner` ?

aneeshack4: Done.

packages/office-ui-fabric-react/src/components/ThemeGenerator/ThemeGenerator.ts
```diff
@@ -80,23 +80,22 @@ export class ThemeGenerator {
...
-        const camelCasedName = rule.name.charAt(0).toLowerCase() + rule.name.slice(1);
-        const outputColor = rule.color ? '#' + rule.color.hex : rule.value || '';
-        output += format(attributeTemplate, camelCasedName, outputColor);
-      }
-    }
-    output += '  }\n});';
-    return output;
+  /**
+   * Gets code-formatted load theme blob, specifically for the new theme designer,
+   * aka.ms/themedesigner. Can't use loadtheme like the old theme designer.
```
ecraig12345: Might be good to clarify why the new theme designer can't use loadTheme (I guessed at the explanation in the suggestion).
```suggestion
   * aka.ms/themedesigner. Shouldn't use loadTheme like the old theme designer since it's deprecated.
```

packages/office-ui-fabric-react/src/components/ThemeGenerator/ThemeGenerator.ts
```diff
@@ -179,4 +178,29 @@ export class ThemeGenerator {
       }
     }
   }
+
+  /* Makes the rest of the code that's used for the load theme blob in the theme designers' codepens.
```
ecraig12345: Super nitpicky thing
```suggestion
  /* Makes the rest of the code that's used for the load theme blob in the theme designer's codepens.
```

packages/office-ui-fabric-react/src/components/ThemeGenerator/ThemeGenerator.ts
```diff
@@ -80,23 +80,22 @@ export class ThemeGenerator {
...
-      }
-    }
-    output += '  }\n});';
-    return output;
+  /**
+   * Gets code-formatted load theme blob, specifically for the new theme designer,
+   * aka.ms/themedesigner. Can't use loadtheme like the old theme designer.
+   * We want to use the theme object from createTheme and use the Customizations.applySettings API instead.
+   */
+  public static getThemeAsCodeForThemeDesigner(slotRules: IThemeRules): any {
```
ecraig12345: One other comment, I wonder if it would be better to give this a name less specific to theme designer? Like `getCreateThemeCode` or `getThemeAsCodeWithCreateTheme` maybe. (I'd actually have preferred adding an extra prop `useCreateTheme` or something like that to the existing `getThemeAsCode`, contrary to JD's feedback. :P ) But it's not a big deal since I don't think this is a public API.

ecraig12345:
I just noticed that ThemesOverview.md still talks about `loadTheme`, so that should probably be updated too (either in this PR or separately).

packages/office-ui-fabric-react/src/components/ThemeGenerator/ThemeGenerator.ts
aneeshack4: There are two theme designers hence the plural - I'll make that more clear.

ecraig12345: Ahhhh right :)

packages/office-ui-fabric-react/src/components/ThemeGenerator/ThemeGenerator.ts
aneeshack4: Done.

apps/theming-designer/src/components/Header.tsx
```diff
@@ -52,10 +52,10 @@ const headerStackStyles = (p: IStackProps, theme: ITheme) => ({
   }
 });
 
-const codeHeader = "import { loadTheme } from 'office-ui-fabric-react';\n\n";
-const codepenHeader = 'const { loadTheme, DefaultButton, PrimaryButton, Toggle, TooltipHost } = Fabric;\n\n';
+const codepenHeader = 'const { createTheme, Customizations, DefaultButton, PrimaryButton, Toggle, TooltipHost } = Fabric;\n\n';
 const codepenSamples =
-  '\n\nclass Content extends React.Component {\n  public render() {\n    return (<div>' +
+  '\n\nclass Content extends React.Component {\n  public render() {\n  Customizations.applySettings({ theme: _theme });  ' +
```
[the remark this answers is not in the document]
aneeshack4: Done!

ecraig12345:
It's here https://developer.microsoft.com/en-us/fabric/#/controls/web/themes

aneeshack4:
Ahhh ok @ecraig12345 thanks! I'll do that in a subsequent PR to not hold up the theme designer update since that's where a lot of our users seem to already be going - don't want to delay giving them the correct info.

PR #11332, opened 2019-11-28
Preserve modal selection state when navigating between selected items

ThomasMichon:
Fixed an issue where modal selection state would get cleared when navigating selected items using the arrow keys. Since `_clearAndSelectIndex` invokes `setAllSelected(false)` and that clears modal state internally, the `SelectionZone` needs to restore `isModal` once it is done modifying selection.
This is already done elsewhere in `SelectionZone` but this pathway was missed.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11332)

### DIFF
--- a/packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
+++ b/packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
@@ -99,7 +99,14 @@ export interface ISelectionZoneProps extends React.ClassAttributes<SelectionZone
 /**
  * {@docCategory Selection}
  */
-export class SelectionZone extends BaseComponent<ISelectionZoneProps, {}> {
+export interface ISelectionZoneState {
+  isModal: boolean | undefined;
+}
+
+/**
+ * {@docCategory Selection}
+ */
+export class SelectionZone extends BaseComponent<ISelectionZoneProps, ISelectionZoneState> {
   public static defaultProps = {
     isSelectedOnFocus: true,
     selectionMode: SelectionMode.multiple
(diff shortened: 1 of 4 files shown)


packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
```diff
@@ -577,12 +577,17 @@ export class SelectionZone extends BaseComponent<ISelectionZoneProps, {}> {
...
 
     if (!isAlreadySingleSelected) {
+      const isModal = selection.isModal && selection.isModal();
       selection.setChangeEvents(false);
       selection.setAllSelected(false);
       selection.setIndexSelected(index, true, true);
-      if (this.props.enterModalOnTouch && this._isTouch && selection.setModal) {
-        selection.setModal(true);
-        this._setIsTouch(false);
+      if (isModal || (this.props.enterModalOnTouch && this._isTouch)) {
```
aneeshack4: How is `isModal` being restored here?

ThomasMichon: Since `isModal` is set to `true` within the body, then if it was already `true` it winds up being 'restored' to `true`.

packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
```diff
@@ -139,12 +169,22 @@ export class SelectionZone extends BaseComponent<ISelectionZoneProps, {}> {
...
       >
         {this.props.children}
       </div>
     );
   }
 
+  public componentDidUpdate(previousProps: ISelectionZoneProps): void {
+    const { selection } = this.props;
+
+    if (selection !== previousProps.selection) {
```
aneeshack4: A comment explaining this would add some context

packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
```diff
@@ -155,6 +195,16 @@ export class SelectionZone extends BaseComponent<ISelectionZoneProps, {}> {
     this._handleNextFocus(false);
   };
 
+  private _onSelectionChange = (): void => {
```
aneeshack4: Adding a test or 2 to validate this new behavior would be useful.

Issue #8855, opened 2019-04-26
Overlay: Prevents scrolling in contained components.


abhishek1214:
### Environment Information 
 
- __Package version(s)__: (fill this out) 
- __Browser and OS versions__: (fill this out if relevant)

### Please provide a reproduction of the bug in a codepen:
  https://codepen.io/anon/pen/QPZQLP

![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/7584903/56786486-c4b0c880-6816-11e9-91e4-a41f927e8222.png)

#### Actual behavior:
 1. The menu of the dropdown is not scrollable on touch device, (for ex: I am using a surface pro).

This happens because Overlay calls "[disableBodyScroll](https://github.com/OfficeDev/office-ui-fabric-react/blob/2894062b1d0515055c29061dfc6f586f918ca41c/packages/office-ui-fabric-react/src/components/Overlay/Overlay.base.tsx#L9)" when mounted. This fixes an iOS bug that occurs anytime a dialog/modal is open. 
 

 
#### Expected behavior:
  1. One should be able to scroll the menu on a touch screen, for example i have a surface pro device.

#### Priorities and help requested:
 
Are you willing to submit a PR to fix? (Yes, No)
Yes

Requested priority: (Blocking, High, Normal, Low) 
 
Products/sites affected: (if applicable)

apneer:
Thank you for bringing this to our attention. I confirm missing feature. We'll work on getting this prioritized.

dzearing:
This sounds like a bug more than a feature request. @joschect can you please take a look?

joschect:
@abhishek1214 can you confirm that this is still an issue? I tried to reproduce using my phone and I was able to scroll fine.

v-dhiraj:
@joschect I work with @abhishek1214 , and we are still not able to make the dropdown work on mobile phone browsers. Please let us know if you have any fix.

abhishek1214:
@joschect we recently upgraded to office-fabric v6.176.0. We are still facing this issue on touch devices. We have Panel with a dropdown to be more specific.

joschect:
@v-dhiraj and @abhishek1214 Are you using a stylus or something like this? I tried using the codepen you provided and it worked without issue on my android device. You mentioned using a Surface, are there any other devices you've tested with that don't work?

v-dhiraj:
[...]

PR #11339, opened 2019-11-28
Panel, Modal, Overlay: Optional prop that disables bodyscroll block on touch devices

### LINKED ISSUE #10982: Panel, allowScrollOnElement on mobile issue.
### Environment Information

- **Package version(s)**: "office-ui-fabric-react": "^6.210.2"
- **Browser and OS versions**: Safari on IPhone

### Please provide a reproduction of the bug in a codepen
https://codepen.io/laiprorus/full/pooWjZw

#### Actual behavior:
The Panel and Overlay components use allowScrollOnElement function from https://github.com/OfficeDev/office-ui-fabric-react/blob/09d5abf9a89b1f2ae31de5427e24279537eb9a4c/packages/utilities/src/scroll.ts.
This function creates touchmove events that prevent the user from paning or zooming document body. 
In my codepen, if you use a page on mobile, and use touch pinch to zoom in for example on the button, and then click, panel will open. But since you are zoomed in, you will only see part of it. And since "touchmove" events are prevented, you cant scroll to panels content.
There is no way for my code, that uses Panel component to disable those blocking events and it is also impossible to ensure the user is fully zoomed out before panel opens. And even if user is zoomed out, default behaviour on mobile for elements such as text input field, is to zoom in on that input field, and then the user has hard time to zoom out back.
Im not sure if its a bug or missing feature, but it is pretty breaking behavior for my current project. ~~My workaround was to render this component inside Panel content to enable zoom and paning. `const EnableScrollFix: React.FunctionComponent<{}> = () => {
    React.useEffect(() => {
      Fabric.enableBodyScroll();
      $("div[data-is-scrollable]").each((index, element) => {
        const contents = $(element).contents();
        $(element).replaceWith($("<div></div>").append(contents));
      });
      return () => {
      };
    }, []);
    return (null);
  };`
this components is empty, and is just used to get to that "componentDidMount" callback of Panel content. It replaces div that blocks events, with empty div, since i couldnt remove event otherwise. But its very cheesy solution, and especially risky since i am modifying dom elements created by react, with out its knowledge.~~

#### Expected behavior:

After Panel opens, user should be able to pan and zoom on mobile devices.

Requested priority: 
Blocking

Products/sites affected: corporate internal web application.

laiprorus:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10982

#### Description of changes

Adds optional prop "allowTouchBodyScroll" to Panel, Modal and Overlay components. This prop allows body scroll events on touch devices. 
This prop disables overscroll blocks in [allowScrollOnElement](https://github.com/OfficeDev/office-ui-fabric-react/blob/2894062b1d0515055c29061dfc6f586f918ca41c/packages/utilities/src/scroll.ts#L84) and doesnt call [disableBodyScroll](https://github.com/OfficeDev/office-ui-fabric-react/blob/2894062b1d0515055c29061dfc6f586f918ca41c/packages/utilities/src/scroll.ts#L95) on overlay mount.
As this prop is optional, without it being set to true, components work exactly as they were before.
This change is a simple workaround for breaking behaviour described in issue: https://github.com/OfficeDev/office-ui-fabric-react/issues/10982

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11339)


laiprorus:
Besides my original issue #10982, I found more issues that are related to those 2 functions (allowScrollOnElement and disableBodyScroll). A recent one #11343 and older one #8855.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -5531,6 +5531,7 @@ export interface IModal {
 
 // @public (undocumented)
 export interface IModalProps extends React.ClassAttributes<ModalBase>, IWithResponsiveModeState, IAccessiblePopupProps {
+    allowIosBodyScroll?: boolean;
```
aneeshack4: Rename this to `allowTouchBodyScroll` or something non-platform specific.

packages/office-ui-fabric-react/src/components/Overlay/Overlay.base.tsx
```diff
@@ -5,12 +5,21 @@ import { IOverlayProps, IOverlayStyleProps, IOverlayStyles } from './Overlay.typ
...
+
+  constructor(props: IOverlayProps) {
+    super(props);
+
+    const { allowIosBodyScroll = false } = this.props;
+    this._allowIosBodyScroll = allowIosBodyScroll;
+  }
+
   public componentDidMount(): void {
-    disableBodyScroll();
```
aneeshack4: So disableBodyScroll() is just being deprecated?

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
```diff
@@ -270,7 +281,7 @@ export class ModalBase extends BaseComponent<IModalProps, IDialogState> implemen
   // Allow the user to scroll within the modal but not on the body
   private _allowScrollOnModal = (elt: HTMLDivElement | null): void => {
     if (elt) {
-      allowScrollOnElement(elt, this._events);
+      allowScrollOnElement(elt, this._events, this._allowIosBodyScroll);
```
aneeshack4: What's missing from the current allowScrollOnElement that misses the touch capability? Instead of making a separate, new method to handle touch, can we fill in the missing gaps to handle touch in allowScrollOnElement?

packages/utilities/etc/utilities.api.md
```diff
@@ -18,7 +18,7 @@ export function addDirectionalKeyCode(which: number): void;
 export function addElementAtIndex<T>(array: T[], index: number, itemToAdd: T): T[];
 
 // @public
-export const allowScrollOnElement: (element: HTMLElement | null, events: EventGroup) => void;
+export const allowScrollOnElement: (element: HTMLElement | null, events: EventGroup, allowIosOverscroll?: boolean) => void;
```
aneeshack4: Adding a boolean arg is usually not recommended - can we circumvent this without a boolean arg?

aneeshack4: See: https://github.com/OfficeDev/office-ui-fabric-react/wiki/Pull-request-reviewing-guidance
> Avoid using boolean params in method definitions. Boolean params usually mean a conditional if statement will be used which means the caller probably knows too much about the implementation of the method it's calling. It also makes it hard to change the method later. Make 2 methods that do different things instead.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
laiprorus: this is good point, i will change it to allowTouchBodyScroll, i think i took the name from [_disableIosBodyScroll](https://github.com/OfficeDev/office-ui-fabric-react/blob/2894062b1d0515055c29061dfc6f586f918ca41c/packages/utilities/src/scroll.ts#L86)

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
laiprorus: allowScrollOnElement allows scrolling of content within the element and when top or botton of content is reached, events are blocked, this way it prevents overscrolling [_preventOverscrolling](https://github.com/OfficeDev/office-ui-fabric-react/blob/2894062b1d0515055c29061dfc6f586f918ca41c/packages/utilities/src/scroll.ts#L34) where body would be scrolled. But this creates an issue, because _preventOverscrolling blocks all touch events . Example is my original issue #10982 where if you are zoomed in on mobile device, you cant pan around because of _preventOverscrolling event handler. We need to somehow allow the user to scroll the content, but also be able to pan around while zoomed in.
My easiest and minimal to implement solution was, to just make blocking happening in _preventOverscrolling optional, that what boolean argument allowIosOverscroll is used for.

laiprorus: With this PR, i could create Panel and Modal components that dont block touch events. This is not perfect solution, as it allows body scrolling behind the overlay, which is a bit irritating, but UI is still intuitively usable. Where in current state, user is stuck in not usable UI.

packages/utilities/etc/utilities.api.md
laiprorus: Originaly i tought of this, but this would create duplicate code. It would be just 2 slighty different versions of [_makeElementScrollAllower](https://github.com/OfficeDev/office-ui-fabric-react/blob/2894062b1d0515055c29061dfc6f586f918ca41c/packages/utilities/src/scroll.ts#L21). But now i see if you remove the code that is needed for _preventOverscrolling handler, not much is left and duplicate code will be minimal.

laiprorus:
i want to investigate allowScrollOnElement more, especially the single vs multiple finger touch event. It is not clear to me, that [_preventOverscrolling](https://github.com/OfficeDev/office-ui-fabric-react/blob/09d5abf9a89b1f2ae31de5427e24279537eb9a4c/packages/utilities/src/scroll.ts#L36) says that it only responds to single finger touch events, but zooming (2 finger event) is still not quite possible on Panel content.

maxwellred:
@laiprorus I have pulled down this branch and attempted to use the functionality but haven't been able to observe the fix. I have attempted using the allowTouchBodyScroll={true} in chrome dev tools mobile emulator and on my samsung s7 phone using a codepen but panning and pinch zooming are still locked when the Panel is shown. Perhaps my testing environment or tools are insufficient for testing this feature? Can you provide a gif or video showing the functionality in action, thanks!

laiprorus:
panel with allowTouchBodyScroll
panelwithallow

laiprorus:
@maxwellred here 2 examples without allowTouchBodyScroll (how it is current fabric).
in first one, i open panel fully zoomed out, then i manage to zoom in and get stuck, not being able zoom out to reach close button.
panelwithoutallow1

in second, i zoom in into button, open panel, scroll to left until no panel is in view point, and stuck again unable to zoom out or reach close button. 
panelwithoutallow2

in both examples, some touch actions (inside textfield for example) could not be blocked by fabric and are executed. you can see errors in console saying fabric tried to block but it was intervented by browser.

laiprorus:
here is the component code is used for this example:-
`import * as React from 'react';
import { Component, Fragment } from 'react';
import {
  Fabric,
  Text,
  TextField,
  DefaultButton,
  Panel,
  PanelType
} from 'office-ui-fabric-react';

interface DemoComponentProps { }

interface DemoComponentState {
  isOpenWith: boolean
  isOpenWithout: boolean
}

export class DemoComponent extends Component<DemoComponentProps, DemoComponentState> {
  constructor(props: DemoComponentProps) {
    super(props);
    this.state = {
      isOpenWith: false,
      isOpenWithout: false
    }
  }

  _openPanelWith = () => {
    this.setState({ isOpenWith: true });
  }

  _closePanelWith = () => {
    this.setState({ isOpenWith: false });
  }

  _openPanelWithout = () => {
    this.setState({ isOpenWithout: true });
  }

  _closePanelWithout = () => {
    this.setState({ isOpenWithout: false });
  }

  _panelContent = () => {
    return [...Array(20)].map((_, i) => <Fragment key={i}>
      <Text>
        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
              </Text>
      <TextField label={`TextField ${i}`} />
    </Fragment>);
  }

  render = () => {
    const { isOpenWith, isOpenWithout } = this.state;
    return <Fragment>
      <Fabric>
        <DefaultButton
          text={"Open Panel with allowTouchBodyScroll"}
          onClick={this._openPanelWith}
        />
        <br />
        <DefaultButton
          styles={{ root: { marginTop: 9 } }}
          text={"Open Panel without allowTouchBodyScroll"}
          onClick={this._openPanelWithout}
        />
        <Panel
          type={PanelType.medium}
          isOpen={isOpenWith}
          allowTouchBodyScroll
          onDismiss={this._closePanelWith}
        >
          {this._panelContent()}
        </Panel>
        <Panel
          type={PanelType.medium}
          isOpen={isOpenWithout}
          onDismiss={this._closePanelWithout}
        >
          {this._panelContent()}
        </Panel >
      </Fabric>
    </Fragment >;
  }
}
`

laiprorus:
`
import { Component, Fragment } from 'react';
import {
  Fabric,
  Text,
  TextField,
  DefaultButton,
  Panel,
  PanelType
} from 'office-ui-fabric-react';

interface DemoComponentProps { }

interface DemoComponentState {
  isOpenWith: boolean
  isOpenWithout: boolean
}

export class DemoComponent extends Component<DemoComponentProps, DemoComponentState> {
  constructor(props: DemoComponentProps) {
    super(props);
    this.state = {
      isOpenWith: false,
      isOpenWithout: false
    }
  }

  _openPanelWith = () => {
    this.setState({ isOpenWith: true });
  }

  _closePanelWith = () => {
    this.setState({ isOpenWith: false });
  }

  _openPanelWithout = () => {
    this.setState({ isOpenWithout: true });
  }

  _closePanelWithout = () => {
    this.setState({ isOpenWithout: false });
  }

  _panelContent = () => {
    return [...Array(20)].map((_, i) => <Fragment key={i}>
      <Text>
        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
              </Text>
      <TextField label={`TextField ${i}`} />
    </Fragment>);
  }

  render = () => {
    const { isOpenWith, isOpenWithout } = this.state;
    return <Fragment>
      <Fabric>
        <DefaultButton
          text={"Open Panel with allowTouchBodyScroll"}
          onClick={this._openPanelWith}
        />
        <br />
        <DefaultButton
          styles={{ root: { marginTop: 9 } }}
          text={"Open Panel without allowTouchBodyScroll"}
          onClick={this._openPanelWithout}
        />
        <Panel
          type={PanelType.medium}
          isOpen={isOpenWith}
          allowTouchBodyScroll
          onDismiss={this._closePanelWith}
        >
          {this._panelContent()}
        </Panel>
        <Panel
          type={PanelType.medium}
          isOpen={isOpenWithout}
          onDismiss={this._closePanelWithout}
        >
          {this._panelContent()}
        </Panel >
      </Fabric>
    </Fragment >;
  }
}
`

laiprorus:
Here is code of component i used in this example:

```
import { Component, Fragment } from 'react';
import {
  Fabric,
  Text,
  TextField,
  DefaultButton,
  Panel,
  PanelType
} from 'office-ui-fabric-react';

interface DemoComponentProps { }

interface DemoComponentState {
  isOpenWith: boolean
  isOpenWithout: boolean
}

export class DemoComponent extends Component<DemoComponentProps, DemoComponentState> {
  constructor(props: DemoComponentProps) {
    super(props);
    this.state = {
      isOpenWith: false,
      isOpenWithout: false
    }
  }

  _openPanelWith = () => {
    this.setState({ isOpenWith: true });
  }

  _closePanelWith = () => {
    this.setState({ isOpenWith: false });
  }

  _openPanelWithout = () => {
    this.setState({ isOpenWithout: true });
  }

  _closePanelWithout = () => {
    this.setState({ isOpenWithout: false });
  }

  _panelContent = () => {
    return [...Array(20)].map((_, i) => <Fragment key={i}>
      <Text>
        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
              </Text>
      <TextField label={`TextField ${i}`} />
    </Fragment>);
  }

  render = () => {
    const { isOpenWith, isOpenWithout } = this.state;
    return <Fragment>
      <Fabric>
        <DefaultButton
          text={"Open Panel with allowTouchBodyScroll"}
          onClick={this._openPanelWith}
        />
        <br />
        <DefaultButton
          styles={{ root: { marginTop: 9 } }}
          text={"Open Panel without allowTouchBodyScroll"}
          onClick={this._openPanelWithout}
        />
        <Panel
          type={PanelType.medium}
          isOpen={isOpenWith}
          allowTouchBodyScroll
          onDismiss={this._closePanelWith}
        >
          {this._panelContent()}
        </Panel>
        <Panel
          type={PanelType.medium}
          isOpen={isOpenWithout}
          onDismiss={this._closePanelWithout}
        >
          {this._panelContent()}
        </Panel >
      </Fabric>
    </Fragment >;
  }
}
```

maxwellred:
@laiprorus This is great, thank you so much! Everything looks good to me

laiprorus:
i restricted my solution to only mounting of components. to eliminate possible side effects and bugs, like allowTouchBodyScroll prop being changed during touchmove event. I think respecting prop change after mount can implemented later? 
Also i find "allow" doesnt describe well what is happening. By default those 3 components will set event blocks. With allowTouchBodyScroll it just does not set those blocks. But if you open Panel with allowTouchBodyScroll set, and then open for example another Modal without that prop, it will block all events, even panel beneath it does not. Maybe you could help change api documentation to describe the behavior better?

mrphannguyen:
Hi @maxwellred 
Can we have the fix in the version 6.5x.x? I have the same issue with that version.

laiprorus:
Thank you for your feedback and aprroval. Is there anything i could for this PR or should i just wait?

soryy708:
Why is it `false` by default?

laiprorus:
> Why is it `false` by default?

This was the behaviour before this PR.

PR #11349, opened 2019-12-02
Explicit checks for RTL based on theme context.

jdhuntington:
#### Pull request checklist

#### Description of changes

Follow up to #11295 , adds an explicit reference to the theme context in several calls to `getRTL()`. There are still a few open work items to finish this, but this is a good checkpoint and releasable as-is.

For the components affected, they will now work when they are rendered in this style:

```tsx
<Fabric dir="rtl">
  <Slider {...} />
</Fabric>
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11349)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/DetailsList/DetailsList.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/DetailsList.base.tsx
@@ -270,7 +270,7 @@ export class DetailsListBase extends React.Component<IDetailsListProps, IDetails
       shouldForceUpdates = true;
     }
 
-    if (isAllGroupsCollapsed === undefined && (newProps.groupProps && newProps.groupProps.isAllGroupsCollapsed !== undefined)) {
+    if (isAllGroupsCollapsed === undefined && newProps.groupProps && newProps.groupProps.isAllGroupsCollapsed !== undefined) {
       this.setState({
         isCollapsed: newProps.groupProps.isAllGroupsCollapsed,
         isSomeGroupExpanded: !newProps.groupProps.isAllGroupsCollapsed
@@ -468,7 +468,7 @@ export class DetailsListBase extends React.Component<IDetailsListProps, IDetails
               componentRef={this._focusZone}
               className={classNames.focusZone}
               direction={FocusZoneDirection.vertical}
-              isInnerZoneKeystroke={isRightArrow}
+              isInnerZoneKeystroke={this.isRightArrow}
               onActiveElementChanged={this._onActiveRowChanged}
               onBlur={this._onBlur}
             >
@@ -1082,6 +1082,10 @@ export class DetailsListBase extends React.Component<IDetailsListProps, IDetails
       onRenderHeader
     };
   }
+
+  private isRightArrow = (event: React.KeyboardEvent<HTMLElement>) => {
+    return event.which === getRTLSafeKeyCode(KeyCodes.right, this.props.theme);
+  };
 }
 
 export function buildColumns(
@@ -1122,10 +1126,6 @@ export function buildColumns(
   }
 
   return columns;
-}
-
-function isRightArrow(event: React.KeyboardEvent<HTMLElement>): boolean {
-  return event.which === getRTLSafeKeyCode(KeyCodes.right);
 }
 
 function getPaddedWidth(column: IColumn, isFirst: boolean, props: IDetailsListProps): number {
--- a/packages/office-ui-fabric-react/src/components/ProgressIndicator/ProgressIndicator.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/ProgressIndicator/ProgressIndicator.styles.ts
@@ -29,7 +29,7 @@ const IndeterminateProgressRTL = keyframes({
 });
 
 export const getStyles = (props: IProgressIndicatorStyleProps): IProgressIndicatorStyles => {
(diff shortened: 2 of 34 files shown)


packages/office-ui-fabric-react/src/components/DetailsList/DetailsList.base.tsx
```diff
@@ -1082,6 +1082,10 @@ export class DetailsListBase extends React.Component<IDetailsListProps, IDetails
       onRenderHeader
     };
   }
+
+  private isRightArrow = (event: React.KeyboardEvent<HTMLElement>) => {
```
khmakoto:
```suggestion
  private _isRightArrow = (event: React.KeyboardEvent<HTMLElement>) => {
```

packages/office-ui-fabric-react/src/components/ProgressIndicator/ProgressIndicator.styles.ts
```diff
@@ -29,7 +29,7 @@ const IndeterminateProgressRTL = keyframes({
 });
 
 export const getStyles = (props: IProgressIndicatorStyleProps): IProgressIndicatorStyles => {
-  const isRTL = getRTL();
+  const isRTL = getRTLWithTheme(props.theme);
```
khmakoto: You're already getting `theme` from props below, you could move this line and have this instead:

```suggestion
  const isRTL = getRTLWithTheme(theme);
```

packages/office-ui-fabric-react/src/utilities/rtl.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  it('returns ltr when called theme specifying ltr', () => {
+    const theme: ITheme = { ...createTheme({}), rtl: false };
+    expect(getRTLWithTheme(theme)).toBeFalsy();
+  });
+
+  it('returns rtl when called theme specifying rtl', () => {
+    const theme: ITheme = { ...createTheme({}), rtl: true };
+    expect(getRTLWithTheme(theme)).toBeTruthy();
+  });
+});
```
khmakoto: You could have `const theme = createTheme({})` on top and use that instead of calling `createTheme` in every test.

packages/office-ui-fabric-react/src/utilities/rtl.ts
```diff
@@ -0,0 +1,13 @@
+import { getRTL, getRTLSafeKeyCode } from './../Utilities';
+import { ITheme } from '@uifabric/styling';
+
+export function getRTLWithTheme(theme?: ITheme): boolean {
```
ecraig12345: Couldn't we just add an optional `theme` argument to the existing `getRTL`?

packages/office-ui-fabric-react/src/utilities/rtl.test.ts
joschect: Plus one for using beforeEach and afterEach

packages/office-ui-fabric-react/src/utilities/rtl.ts
jdhuntington: I considered it, but I don't want to have the utilities package take a dependency on styling.

packages/office-ui-fabric-react/src/utilities/rtl.test.ts
jdhuntington: It's minor, but it does feel like a tradeoff. The benefits of moving the theme creation up in context saves a few milliseconds, but in my opinion it introduces a few risks:
- tests become DRY instead of [DAMP](http://blog.jayfields.com/2006/05/dry-code-damp-dsls.html)
- easy for someone to accidentally modify the `theme` object

packages/office-ui-fabric-react/src/utilities/rtl.ts
ecraig12345: Makes sense. You could get around that by using a "theme-like" object `theme?: { rtl?: boolean }`, though that's not ideal either.

packages/office-ui-fabric-react/src/components/ProgressIndicator/ProgressIndicator.styles.ts
```diff
@@ -29,7 +29,7 @@ const IndeterminateProgressRTL = keyframes({
 });
 
 export const getStyles = (props: IProgressIndicatorStyleProps): IProgressIndicatorStyles => {
-  const isRTL = getRTL();
+  const isRTL = getRTLWithTheme(theme);
```
khmakoto: You have to move this below line 33 in order for it to build.

khmakoto:
@jdhuntington any updates here?

packages/office-ui-fabric-react/src/utilities/rtl.ts
dzearing: I expected what @ecraig12345 suggested. I was going to look at fixing a shimmer issue, and thought that this PR might have been merged so I tried just passing theme into `getRTL` and it didn't work.

I would prefer that we just have 1 API in utilities; it's weird to have 2.

jdhuntington:
Closing, as some of @micahgodbolt 's changes are in the PR incorrectly. Will reopen.

PR #11353, opened 2019-12-03
ComboBox: empty errorMessage shouldn't create white space.

### LINKED ISSUE #11277: ComboBox empty errorMessage creates whitespace
### Environment Information

- **Package version(s)**:    "office-ui-fabric-react": "^7.56.3",
- **Browser and OS versions**: chrome 78.0

### Describe the issue:

ComboBox is rendering an empty div for errorMessage under the box, creating a space so it does not align with other components.

#### Actual behavior:

`<div role="region" aria-live="polite" aria-atomic="true" id="ComboBox91-error" class="css-180"></div>`  as seen on https://developer.microsoft.com/en-us/fabric#/controls/web/combobox#feedback

#### Expected behavior:

Hide the div if errorMessage is false or empty.

{errorMessage && <ErrorMessage/>}

aneeshack4:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11277 

#### Description of changes

If the div for the error message doesn't have any text, the div still needs to be there to let ATs like Narrator know to read out the text when it changes (as it is an `aria-live` region) but it doesn't need to have a margin.
Before:
image

After:
image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11353)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
@@ -339,6 +339,7 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
     } = this.props;
     const { isOpen, focused, suggestedDisplayValue } = this.state;
     this._currentVisibleValue = this._getVisibleValue();
+    const hasText = this._currentVisibleValue === '';
 
     const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties, ['onChange', 'value']);
 
@@ -441,7 +442,13 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
             },
             this._onRenderContainer
           )}
-        <div role="region" aria-live="polite" aria-atomic="true" id={errorMessageId} className={this._classNames.errorMessage}>
+        <div
+          role="region"
+          aria-live="polite"
+          aria-atomic="true"
+          id={errorMessageId}
+          className={hasText ? this._classNames.errorMessage : ''}
+        >
           {errorMessage !== undefined ? errorMessage : ''}
         </div>
       </div>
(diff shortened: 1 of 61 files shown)


packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -441,7 +442,13 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
             },
             this._onRenderContainer
           )}
-        <div role="region" aria-live="polite" aria-atomic="true" id={errorMessageId} className={this._classNames.errorMessage}>
+        <div
+          role="region"
+          aria-live="polite"
+          aria-atomic="true"
+          id={errorMessageId}
+          className={hasText ? this._classNames.errorMessage : ''}
```
jdhuntington: This seems to hinge adding a class name based on whether or not the combobox itself has text; is that the intent? It seems like an error when the combo is empty is a legit case.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -339,6 +339,7 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
     } = this.props;
     const { isOpen, focused, suggestedDisplayValue } = this.state;
     this._currentVisibleValue = this._getVisibleValue();
+    const hasText = this._currentVisibleValue === '';
```
micahgodbolt: Why are we keying off of `_currentVisibleValue` here? Wouldn't we only want to show the error message is  or not?

can't we 
```js
{ hasErrorMessage && <div> error message div </div> }
```

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
aneeshack4: Yes this is the intent - right now it's adding a margin even when there's no text making it look like the div has empty space.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
aneeshack4: You're right, thanks Micah! Don't need to make a new variable

aneeshack4: I just double checked it works the same: 
image

PR #11357, opened 2019-12-03
Added Non Focusable Callout Example and made FocusTrapCallout more explicit

### LINKED ISSUE #1655: [Accessibility] Callout is not readable by windows narrator when callout is not having any focusable element
All callouts mentioned here https://dev.office.com/fabric#/components/callout are not readable by windows narrator. For example callout mentioned in the "Callout directional example" however "Nested callout..." content is readable by windows narrator.

Repro steps:

1) Launch edge and go to https://dev.office.com/fabric#/components/callout
2)  Press windows+Enter key to start narrator
3) Now make tab and reach to "Nested callout example..." and press enter
4) Narrator can read content by pressing CAPS + right/left arrow
5) Now move to "Callout directional example" and do the same...

Here narrator are not able to read the content and we can't move inside with CAPS + left/right arrow. The reason is that it doesn't have any focusable element. But ideally we should give way to read the content by Narrator, currently there is no way to do the same.

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1655

#### Description of changes

Created new example of the callout that has specific role and aria-live values, as well as a defered render, similar to the Announced control. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11357)


packages/office-ui-fabric-react/src/components/Callout/AnnouncedCallout.tsx
```diff
@@ -0,0 +1,24 @@
...
+ */
+export const AnnouncedCallout: React.StatelessComponent<IAnnouncedCalloutProps> = (props: IAnnouncedCalloutProps): JSX.Element => {
+  // announced scenario requires specific role and ariaLive values regardless of user provided values
+  const announcedProps = {
+    role: 'status',
+    ariaLive: 'assertive'
+  };
+  return (
+    <Callout {...announcedProps} {...props}>
+      <DelayedRender>
```
ecraig12345: Can we not just use an Announced inside a Callout? (Or add an example of doing that to the Callout docs?)

micahgodbolt: Yes, this is 100% doable without this control. I have a demo right here:
https://codepen.io/micahgodbolt/pen/abbxvPN?editors=0010

The goal is to provide a consistent solution for a very common scenario. It's very similar to the FocusTrapCallout. 

The advantage of this approach is that we provide solutions for common problems without increasing the weight of the original callout (would be silly to ship FocusTrapZone and DelayedRender with every callout).

I agree, there is technical debt in creating this control vs just writing an example, but seeing as it solves a major accessibility issues, I feel that it is a worthy addition.

khmakoto:
@micahgodbolt any updates on this?

micahgodbolt:
yeah...i started in on a PR to remove the extra control and just update docs to show how you can do this on your own. I'll get that PR in today or tomorrow.

packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -321,7 +321,8 @@ export class CalloutContentBase extends React.Component<ICalloutProps, ICalloutS
       (ev.target !== this._targetWindow &&
         isEventTargetOutsideCallout &&
         ((this._target as MouseEvent).stopPropagation ||
-          (!this._target || (target !== this._target && !elementContains(this._target as HTMLElement, target)))))
+          !this._target ||
+          (target !== this._target && !elementContains(this._target as HTMLElement, target))))
```
micahgodbolt: linting update

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Directional.Example.tsx
```diff
@@ -79,12 +81,29 @@ const styles = mergeStyleSets({
...
+    whiteSpace: 'nowrap'
+  }
 });
 
 export class CalloutDirectionalExample extends React.Component<{}, ICalloutDirectionalExampleState> {
   private _menuButtonElement: HTMLElement | null;
 
+  // Use getId() to ensure that the callout label and description IDs are unique on the page.
+  // (It's also okay to use plain strings without getId() and manually ensure their uniqueness.)
+  private _labelId: string = getId('callout-label');
```
micahgodbolt: Added IDs to give example label/description

micahgodbolt:
@ecraig12345 this PR has been updated to simply add an example of the non focusable callout

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Basic.Example.tsx
```diff
@@ -74,7 +74,7 @@ export class CalloutBasicExample extends React.Component<{}, ICalloutBasicExampl
     return (
       <>
         <div className={styles.buttonArea} ref={this._menuButtonElement}>
-          <DefaultButton onClick={this._onShowMenuClicked} text={isCalloutVisible ? 'Hide callout' : 'Show callout'} />
+          <DefaultButton onClick={this._onShowMenuClicked} text={isCalloutVisible ? 'Hide Callout' : 'Show Callout'} />
```
ecraig12345: This should stay sentence-case rather than title case. MS style is to use sentence case, even though we're not consistent about that right now.

packages/office-ui-fabric-react/src/components/Callout/docs/CalloutOverview.md
```diff
@@ -1,5 +1,7 @@
...
+
 Use a Callout for displaying additional contextual information about an item on the screen. Callouts also have a tail that identifies their source. A common use for Callout is the introduction of a new feature or capability of an app or site. Alternate usages include pairing the Callout with a button or clickable element for on-demand presentation of additional or supporting content.
 
-Real-world examples of this implementation can be seen in administrative interfaces where a particularly difficult-to-understand concept is paired with the iconClassNames.info "i" icon. In this example, Callout - with its tip text - is opened when the user clicks on or hovers over the icon.
\ No newline at end of file
+Real-world examples of this implementation can be seen in administrative interfaces where a particularly difficult-to-understand concept is paired with the iconClassNames.info "i" icon. In this example, Callout - with its tip text - is opened when the user clicks on or hovers over the icon.
```
ecraig12345: I know you didn't add it but this paragraph is unnecessarily complex and specific. I'd suggest something like this instead (also we ideally ought to have an example showing this pattern)...don't totally love this but I think it's an improvement.

```suggestion
One example of Callout usage is when an "i" icon is displayed in the label of a form element which might require additional explanation. A Callout is opened to display tip text when the user selects or hovers over the icon.
```

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.FocusTrap.Example.tsx
```diff
@@ -95,14 +95,14 @@ export class CalloutFocusTrapExample extends React.Component<{}, ICalloutFocusTr
               <div className={styles.inner}>
                 <div>
                   <p className={styles.subtext}>
-                    Message body is optional. If help documentation is available, consider adding a link to learn more at the bottom.
+                    Content is wrapped in a Focus Trap Zone so that user cannot accidently tab out of this callout.
```
ecraig12345: Probably either use the literal control name or lowercase the words
```suggestion
                    Content is wrapped in a FocusTrapZone so that user cannot accidently tab out of this callout.
```

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.FocusTrap.Example.tsx
```diff
@@ -95,14 +95,14 @@ export class CalloutFocusTrapExample extends React.Component<{}, ICalloutFocusTr
...
                   </p>
                 </div>
               </div>
               <FocusZone>
                 <Stack className={styles.buttons} gap={8} horizontal>
-                  <PrimaryButton onClick={this._onDismiss}>Agree</PrimaryButton>
-                  <DefaultButton onClick={this._onDismiss}>Cancel</DefaultButton>
+                  <PrimaryButton onClick={this._onDismiss}>Button 1</PrimaryButton>
```
ecraig12345: Curious why this was changed?

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Directional.Example.tsx
```diff
@@ -131,13 +152,18 @@ export class CalloutDirectionalExample extends React.Component<{}, ICalloutDirec
...
-              <div>
-                <p className={styles.subtext}>
-                  Message body is optional. If help documentation is available, consider adding a link to learn more at the bottom.
-                </p>
+              <p className={styles.subtext} id={this._descriptionId}>
+                Message body is optional. If help documentation is available, consider adding a link to learn more at the bottom.
+              </p>
+              <div className={styles.actions}>
+                <Link className={styles.link} href="http://microsoft.com" target="_blank">
+                  Go to microsoft
```
ecraig12345:
```suggestion
                  Go to Microsoft
```

packages/office-ui-fabric-react/src/components/Callout/docs/CalloutOverview.md
ecraig12345: Umm actually we should remove this paragraph...this is TooltipHost behavior, not something we should mention in the Callout article.

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Status.Example.tsx
```diff
@@ -0,0 +1,83 @@
...
+            className={styles.callout}
+            target={this._menuButtonElement.current}
+            onDismiss={this._onCalloutDismiss}
+            role="status"
+            aria-live="assertive"
+          >
+            <DelayedRender>
+              <>
+                <p className={styles.subtext}>
+                  This message is treated as an aria-live assertive status message, and will be read by narrator regardless of focus.
```
ecraig12345:
```suggestion
                  This message is treated as an aria-live assertive status message, and will be read by a screen reader regardless of focus.
```

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Status.Example.tsx
```diff
@@ -0,0 +1,83 @@
...
+        </div>
+        {this.state.isCalloutVisible && (
+          <Callout
+            className={styles.callout}
+            target={this._menuButtonElement.current}
+            onDismiss={this._onCalloutDismiss}
+            role="status"
+            aria-live="assertive"
+          >
+            <DelayedRender>
```
ecraig12345: Is there a reason for using DelayedRender instead of Announced here? (Also for the fragment wrapping the p?)

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Directional.Example.tsx
```diff
@@ -79,12 +81,29 @@ const styles = mergeStyleSets({
...
+    width: '100%',
+    whiteSpace: 'nowrap'
+  }
 });
 
 export class CalloutDirectionalExample extends React.Component<{}, ICalloutDirectionalExampleState> {
   private _menuButtonElement: HTMLElement | null;
 
+  // Use getId() to ensure that the callout label and description IDs are unique on the page.
+  // (It's also okay to use plain strings without getId() and manually ensure their uniqueness.)
```
khmakoto:
```suggestion
  // It's also okay, though not recommended, to use plain strings without getId() and manually ensure their uniqueness.
```

micahgodbolt: was a copy/paste from other examples :D but truth

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Status.Example.tsx
micahgodbolt: Fragment is required by DelayedRender. It's a bit picky.

Announced speaks the string but doesn't display it. To display an announced string, this is basically exactly what Announced does.

ecraig12345: Oh interesting, I forgot about that. Almost makes me wonder if we should have an Announced variant or prop which DOES show the content? I suspect it's a common enough scenario.

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Basic.Example.tsx
micahgodbolt: I capped it because "Callout" is a proper noun.

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Status.Example.tsx
micahgodbolt: Variant, or just base control. Would be good to abstract this usecase

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Basic.Example.tsx
ecraig12345: In some sense that's true, but the style we're moving towards (per editorial board discussions) is using sentence case names rather than CodeNames for components unless it's necessary for clarity. Though that's still very inconsistent in the current docs.

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.FocusTrap.Example.tsx
micahgodbolt: > Don’t use Callout to ask the user to confirm an action, use a Dialog instead.

packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Status.Example.tsx
ecraig12345: `<Announced show>`? And maybe reconsider the default in the Fluent version.

ecraig12345: I feel like this has come up for panel content or other cases too.

PR #11360, opened 2019-12-03
GroupedList _getGroupHeight is now calculated with recursion for children

### LINKED ISSUE #11359: GroupedList - Default getGroupHeight ignores collapsed children, causes render issue
### Environment Information

- **Package version(s)**: `office-ui-fabric-react@7.68.0`

The group height is miscalculated in GroupedList if the provided group is not collapsed but has children that are collapsed. This causes a rendering issue, see the following codepen example:
https://codepen.io/marwankhalili/pen/VwYYjWX

#### Actual behavior:
Expanding a group on level 0 with collapsed children causes issues with the virtualization / scrolling and subsequent groups on level 0 will not to be displayed.

#### Expected behavior:
Expanding a group on level 0 with collapsed children displays subsequent groups on level 0 if there is enough space. The scrolling / virtualization should end where the last item or collapsed group is displayed.

This can be fixed by providing a custom getGroupHeight function to the DetailsList / GroupedList. In the codepen example above [(link)](https://codepen.io/marwankhalili/pen/VwYYjWX), uncomment line 19 to see the expected behavior.

Requested priority: Normal

marwankhalili:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11359

#### Description of changes

GroupedList was not calculating the height correctly if the group was not collapsed but had children that were collapsed. `_getGroupHeight` now calculates the height using recursion if the group is not collapsed and has at least one children.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11360)

### DIFF
--- a/change/office-ui-fabric-react-2019-12-03-20-40-54-grouped-list-recursive-height.json
+++ b/change/office-ui-fabric-react-2019-12-03-20-40-54-grouped-list-recursive-height.json
@@ -1,0 +1,8 @@
+{
+  "type": "major",
+  "comment": "GroupedList _getGroupHeight is now calculated with recursion for children",
+  "packageName": "office-ui-fabric-react",
+  "email": "marwankhalili@hotmail.com",
+  "commit": "a34d714a150b0c02f7ea95e57aa84acf0f0414f1",
+  "date": "2019-12-03T19:40:54.780Z"
+}
--- a/change/office-ui-fabric-react-2019-12-03-20-40-54-grouped-list-recursive-height.json
+++ b/change/office-ui-fabric-react-2019-12-03-20-40-54-grouped-list-recursive-height.json
@@ -1,5 +1,5 @@
 {
-  "type": "major",
+  "type": "patch",
   "comment": "GroupedList _getGroupHeight is now calculated with recursion for children",
   "packageName": "office-ui-fabric-react",
   "email": "marwankhalili@hotmail.com",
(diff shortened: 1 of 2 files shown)


maxwellred:
Hi @marwankhalili can you change your if statements to use curly braces in order to pass the build, thanks!

change/office-ui-fabric-react-2019-12-03-20-40-54-grouped-list-recursive-height.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "major",
```
KevinTCoughlin: This change type should be `patch` since no API signature was changed.

KevinTCoughlin:
Thanks for working to address this issue @marwankhalili. The recursive approach may regress deeply nested Lists commonly used in SharePoint Online.

Should the return value of this function be cached? I notice it is called repeatedly on-scroll which may be problematic with the recursive solution, see:

image

cc: @ThomasMichon @soettl @aditima @dzearing

change/office-ui-fabric-react-2019-12-03-20-40-54-grouped-list-recursive-height.json
marwankhalili: Got it, I'm changing it to `patch` .

I was not sure whether to classify this as a breaking change, in theory there could be applications that are dependent on the flawed behavior of `_getGroupHeight`. It seems highly unlikely though.

micahgodbolt:
If this change is back into discussion, can we close the PR until we're ready to address it again?

PR #11361, opened 2019-12-03
Made input border semantic slot accessible

betrue-final-final:

Updated input rest states to have an accessible border by changing the inputborder semantic slot from neutralSecondaryAlt to neutralSecondary.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11361)

### DIFF
--- a/change/@uifabric-date-time-2019-12-03-13-10-37-fix_textfield_dropdown_combo_spin_restStates.json
+++ b/change/@uifabric-date-time-2019-12-03-13-10-37-fix_textfield_dropdown_combo_spin_restStates.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "updated snapshots",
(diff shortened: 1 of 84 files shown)


ecraig12345:
Looks like you need to run `yarn change` again for datetime and experiments (you can use change type `none` if it's just snapshot changes).

betrue-final-final:
I ran yarn change again on the root, which said was unnecessary, and it won't let me run it on date-time or experiments.
image

betrue-final-final:
Hey @aneeshack4, can you take a look at this? Very simple.

betrue-final-final:
Hey @lorejoh12, can you approve this?

PR #11365, opened 2019-12-04
DetailList: Screen Reader is reading wrong column header with associated cell value for single-select list

### LINKED ISSUE #11348: [Accessibility] Screen Reader is not reading ShimmeredDetailsList's column header with associated cell value
### Environment Information

- **Package version(s)**: 7.31.2
- **Browser and OS versions**: [Edge, Chrome, FireFox] + Narrator and NVDA

### Describe the issue:

When using ShimmeredDetailsList component from office-ui-fabric-react, notice when an item is selected (using single selection mode), screen reader's scan mode is not reading the correct header information associated with cell value. This was addressed previously: 

https://github.com/OfficeDev/office-ui-fabric-react/issues/5164#issuecomment-409370385

(copied from https://developer.microsoft.com/en-us/fabric#/controls/web/detailslist/shimmer)

https://codepen.io/kejatura/pen/GRgRxQe?editors=1010

- click toggle, and select an item in the list.
- Turn on screen reader's scan mode for selected item and read through the selection.

#### Actual behavior:

- User will notice that the column header reading is ahead of the current cell value.

#### Expected behavior:

- Column header should be read with associated cell value.

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11348

#### Description of changes
`aria-colindex` is not being respected when narrator is reading the header name for each row value. (the multi-line cross-reference scenario). 
This issue seems to something lacking by narrator. Here is a minimal repro in codepen to better demo this bug & my fix: https://codepen.io/xugao/pen/WNbNaVW
The fix I can find is to __not__ hide the empty checkbox header column. Then narrator will count it as a column by default, and header name will be correct since header row has same number of columns as other rows.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11365)

### DIFF
--- a/change/office-ui-fabric-react-2019-12-02-18-41-57-master.json
+++ b/change/office-ui-fabric-react-2019-12-02-18-41-57-master.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "fix scan mode for list item",
+  "packageName": "office-ui-fabric-react",
+  "email": "xgao@microsoft.com",
+  "commit": "153203d03b2612359bdcd61ae4fd45ef71be680e",
+  "date": "2019-12-03T02:41:57.509Z"
+}
--- a/change/office-ui-fabric-react-2019-12-02-18-41-57-master.json
+++ b/change/office-ui-fabric-react-2019-12-02-18-41-57-master.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "fix scan mode for list item",
+  "comment": "Screen Reader is reading wrong column header with associated cell value for single-select detail list",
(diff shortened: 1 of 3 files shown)


ecraig12345:
Can you share screenshots of what it looks like before and after? If there's a visual difference this may be something we need to run by design.

xugao:
@ecraig12345 - there is no visual diff caused by this change. the element i removed style from is empty. 
https://codepen.io/xugao/pen/QWwbNBB if you want to play real quick

PR #11366, opened 2019-12-04
Button: fix size mismatch between primary and default buttons

### LINKED ISSUE #11355: Primary buttons are smaller normal buttons because they have no border
### Environment Information

- **Package version(s)**: latest commit 34c2785efba9f52609bc80cca77a6d37d1d43ae3
- **Browser and OS versions**: -

https://codepen.io/johannes-z/pen/QWwWXeE

#### Actual behavior:

Even though both compound buttons have the same content, their size is different, because the primary button has `border: none`, instead of setting the `borderColor` to `primaryButtonBackground`.

This is the line where this issues occurs: https://github.com/OfficeDev/office-ui-fabric-react/blob/34c2785efba9f52609bc80cca77a6d37d1d43ae3/packages/office-ui-fabric-react/src/components/Button/ButtonThemes.ts#L156

#### Expected behavior:

Both buttons should be the same size.

The line should read:

```ts
border: `1px solid ${s.primaryButtonBackground}`
```

Are you willing to submit a PR to fix? Yes - *I think this applies to all normal/primary button styles though*.

Requested priority: Low

xugao:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #11355

#### Description of changes
Fix bug where primary & default buttons with same child have different sizes due to primary one has no border but default one does.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11366)

### DIFF
--- a/change/office-ui-fabric-react-2019-12-03-18-38-53-xgao-button-style-fix.json
+++ b/change/office-ui-fabric-react-2019-12-03-18-38-53-xgao-button-style-fix.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Button: fix size mismatch between primary and default buttons",
+  "packageName": "office-ui-fabric-react",
+  "email": "xgao@microsoft.com",
+  "commit": "768d5b2be63e2c8e9fd58e4eb91c3268f85820bf",
+  "date": "2019-12-04T02:38:53.362Z"
+}
--- a/packages/office-ui-fabric-react/src/components/Button/ButtonThemes.ts
+++ b/packages/office-ui-fabric-react/src/components/Button/ButtonThemes.ts
@@ -152,8 +152,8 @@ export function primaryStyles(theme: ITheme): IButtonStyles {
   return {
     root: {
       backgroundColor: s.primaryButtonBackground,
-      color: s.primaryButtonText,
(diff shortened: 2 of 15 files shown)


micahgodbolt:
this is causing a small change with high contrast mode

before

image

after

image

Can you make sure that borders are picking up the high contrast background as well

micahgodbolt:
Also, the change in border is affecting the appearance of the focus state

before: 

image

after

image

Not 100% sure the 'after' there is wrong. but it's a change. @betrue-final-final what's your take on the focus styles with that extra pixel border?

betrue-final-final:
It actually makes it a little clearer. I'm good with this. Would Default also look like this with the extra padding?

micahgodbolt:
better view of side by side 
image

micahgodbolt:
Ben is good with the new focus rec......i'd say just clean up high contrast and we're good

betrue-final-final:
Forgot to hit send. Yes this is great.

khmakoto:
@xugao I think we also need to adjust the border of primary buttons in High Contrast when the buttons are hovered and pressed.

micahgodbolt:
high contrast is looking great now 

image

PR #11368, opened 2019-12-04
SpinButton: remove default min/max values

### LINKED ISSUE #11358: SpinButton default max of 100, misleading optional property
### Environment Information

- **Package version(s)**: `office-ui-fabric-react@7.59.0`
- **Browser and OS versions**: Google Chrome is up to date
Version 78.0.3904.108 (Official Build) (64-bit)

There is misleading documentation.
The docs claim the default is 10, but it appears to be 100.
The types also make the property optional implying that if you don't supply a `max` it will have no max value other than the assumed MAX_INTEGER or something.

The issue is that not giving a `max` property means the `max` is 100. 

See codepen. Taken from https://developer.microsoft.com/en-us/fabric#/controls/web/spinbutton then removed all extra unecessary code related to bug.

Just type in 2343245 into the input and press tab. It sets to 100.

https://codepen.io/mattmazzola-the-reactor/pen/KKwwgKN?editors=1010

#### Actual behavior:

Spin button limits input to 100 when it shouldn't.  There is other weird behavior where if you type a large number and click outside of the input onto the webpage background the number stays unmodified, but if you enter 2342345 and press tab, enter, or click on another Office Fabric control the input is reset to 100.

This change in behavior based on type of Dismissal reminds me of another bug related to picker dimissal changing. https://github.com/OfficeDev/office-ui-fabric-react/issues/10892

My work around seems to be forcing highest max possible.
`max={Number.MAX_SAFE_INTEGER}`

Update: looks like issue is same with `min` property also limiting to 0

#### Expected behavior:

Number should be not have a `max` if the `max` property is not provided.

Requested priority: (Blocking, High, **Normal**, Low)

Conversation Learner

xugao:
#### Pull request checklist
- [X] Addresses an existing issue: Fixes #11358

#### Description of changes
We should not set default min/max values on spin buttons as they don't make sense. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11368)


packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
```diff
@@ -112,23 +110,19 @@ export class SpinButton extends React.Component<ISpinButtonProps, ISpinButtonSta
...
     this._lastValidValue = this.state.value;
-    let value: string = newProps.value !== undefined ? newProps.value : String(newProps.min);
-    if (newProps.defaultValue) {
-      value = String(Math.max(newProps.min as number, Math.min(newProps.max as number, Number(newProps.defaultValue))));
-    }
 
     if (newProps.value !== undefined) {
       this.setState({
-        value: value
+        value: newProps.value
```
jdhuntington: This behavior makes sense (not coercing to min/max), but it might trip someone up who is fresh to the code. Would be nice to have a test that explicitly highlights the fact that this is an allowed case.

packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
```diff
@@ -47,7 +47,7 @@ export interface ISpinButtonState {
 }
 
 export type DefaultProps = Required<
-  Pick<ISpinButtonProps, 'step' | 'min' | 'max' | 'disabled' | 'labelPosition' | 'label' | 'incrementButtonIcon' | 'decrementButtonIcon'>
```
aneeshack4: So we're not actually removing min & max as props just from the default props? Also I checked out the issue - do we know why this was causing the max to be 10? Seems random

xugao: I am removing their default values, but not removing these as props. We were setting default values:
image

xugao:
@ecraig12345 , if i have to make a guess, most users of spinbutton today need to explicitly set max and min just to override our defaults, rather than actually relying on them. But this is just my assumption of course, as a user. I definitely also can see this can be a breaking change (your point)

I think I agree more on I should revert this change and just update the doc with correct default value.

However, before I do, could you tell me what is the general guidance on when we can make such change? Is it when we do our next major release?

ecraig12345:
> However, before I do, could you tell me what is the general guidance on when we can make such change? Is it when we do our next major release?

It's generally done in a major release, and explicitly called out in the breaking change notes. You can add a comment like `// TODO (Fabric 8): remove default min/max values--see (issue link)` and tag the issue as Fabric Next. That doesn't guarantee that we'll get to it as part of the release, but it's better than nothing.

xugao:
Created a separate PR to address @ecraig12345's comment: https://github.com/OfficeDev/office-ui-fabric-react/pull/11380

Going to close this one for now

PR #11374, opened 2019-12-04
Dropdown: Items associated with headers

### LINKED ISSUE #10921: Dropdown : 'Fruits' and same for 'Vegetables' not associated with the items under the combo box.
**User Experience:**
Screen reader users will face difficulty if labels are not associated with the items under combo box control.

**Test Environment:**
OS: Version: RS5_Prerelease(1903 OS Build 19005.1)
URL: https://developer.microsoft.com/en-us/fabric#/controls/web/dropdown
Edge Browser Version:
Microsoft Edge: 44.19005.1.0
Microsoft EdgeHTML: 18.19005
Chrome Browser Version:
Chrome: 77.0.3865.120

**Usages:**
All Usages

**Repro Steps:**
1.https://developer.microsoft.com/en-us/fabric#/controls/web/dropdown in Edge and turn on Narrator.
2.Navigate to 'Basic input' button and activate it.
3.Navigate to 'Dropdown' link and activate it.
4.'Dropdown' pane will be appeared.
5.Navigate to 'Basic uncontrolled example' combo box control and activate it.
6.Navigate by using up and down arrow keys and verify the issue.

**Actual Result:**
Screen reader does not convey any information about 'Fruits' and 'Vegetables' labels.

**Expected Result:**
'Apple', 'Banana' are not associated with the 'Fruits' and same for 'Vegetables'

**MAS Reference:**
https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={4b95a70d-258c-474a-820e-5c8b79d5c0bf}

**Attachment:**
image
['Fruits' and same for 'Vegetables' are not associated with items.zip](https://github.com/OfficeDev/office-ui-fabric-react/files/3750696/Fruits.and.same.for.Vegetables.are.not.associated.with.items.zip)

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10921

#### Description of changes

Place items with header into a single group with proper aria-labelledby so that narrator associates items with the header as you arrow down the list. 

#### Focus areas to test

Narrator for custom scenarios

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11374)


packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -539,7 +539,43 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
...
 
-    return <>{props.options.map((item: any, index: number) => onRenderItem({ ...item, index }, this._onRenderItem))}</>;
+    const groups: IDropdownOption[][] = [[]];
+    let index = -1;
+
+    for (const item of props.options) {
+      const groupLength = groups.length;
+      const currentGroup = groups[groupLength - 1];
+      if (item.itemType === SelectableOptionMenuItemType.Header) {
+        if (currentGroup.length === 0) {
```
jspurlin: >currentGroup [](start = 12, length = 12)

I'm not seeing the usage of this except for as a marker for if you are already processing a group (e.g. have seen a header). Does this need to be an array?

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -539,7 +539,43 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
...
+      // check if first item in group is a header
+      const hasHeader = group[0].itemType && group[0].itemType === SelectableOptionMenuItemType.Header;
+      if (hasHeader) {
+        group[0].id = groupId;
+      }
+      const items = group.map((item: any) => {
+        index++;
+        return onRenderItem({ ...item, index }, this._onRenderItem);
+      });
+      renderedGroups.push(hasHeader ? <div aria-labelledby={groupId}>{items}</div> : items);
```
jspurlin: ><div aria-labelledby={groupId}> [](start = 38, length = 31)

consider adding `role="group"` to this div to help make sure screen readers give a consistent experience

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
micahgodbolt: my initial thought was to transform the data into an array of arrays....one for each group, then render them all out.
```
[
  [header1, item2, item3],
  [header4, item5, item6]
]
```
Then I can loop through each array and render out the group down below. 

I could consider refactoring this to simply be a queue....stack a bunch of items in until i hit a header, then spit them out and start the queue over again. Wasn't sure if that refactoring was adding any value.

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
micahgodbolt: good catch. i'd added and removed it a few times in testing. seems i forgot to add it back

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -633,13 +675,18 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
...
+        {...ariaPosProps}
       />
     );
   };
 
   /** Render content of item (i.e. text/icon inside of button) */
   private _onRenderOption = (item: IDropdownOption): JSX.Element => {
-    return <span className={this._classNames.dropdownOptionText}>{item.text}</span>;
+    return (
+      <span role="presentation" className={this._classNames.dropdownOptionText}>
```
jspurlin: >role="presentation" [](start = 12, length = 19)

This shouldn't need `role="presentation"`

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
jspurlin: If you did that (batched all items associated with a header together and flushing them when encountering another header) you should end up having to iterate over the items less, yeah?

---
In reply to: [353948752](https://github.com/OfficeDev/office-ui-fabric-react/pull/11374#discussion_r353948752) [](ancestors = 353948752)

jspurlin:
This is great! The question that comes to mind is: what about doing this for `ContextMenu` and `ComboBox` as well?

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
micahgodbolt: hmmm, yeah, instead of pushing the options to the array, i could push the rendered options....then render the header at the end with the associated wrapping parent. I'll give that a shot tomorrow to see if I can reduce the processing.

micahgodbolt: Ok, i got this working with a single forEach loop, and a single temp cache. I'm also breaking after a divider, so items after the divider, but before the next header, will not be in a group.

micahgodbolt:
@jspurlin yes, doing this for ContextMenu and ComoboBox is also important, though definitely a different PR

jspurlin:
agreeed, just wanted to make sure that comment didn't get lost and have those other components potentially fall through the cracks

hatpick:
Shouldn't this be added to ComboBox as well? @micahgodbolt

PR #11379, opened 2019-12-05
Add alignItems props to <LayoutGroup />

yumikohey:
#### Pull request checklist

#### Description of changes
There is no way to align items cross axis currently in LayoutGroup; thus, added `alignItems` props to provide more flexibility.

#### Focus areas to test
<LayoutGroup />

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11379)


micahgodbolt:
LayoutGroup is an old experiment that has been replaced by Stack. Can you use Stack instead? LayoutGroup should probably be removed from experiments.

PR #11390, opened 2019-12-05
OverflowSet: Vertical layout now uses menu instead of menubar

### LINKED ISSUE #11389: OverflowSet: Vertical is using role="menubar"
Vertical overflow sets are using default menubar role, which is specifically for horizontal menus. Role menu should be used instead.

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11389

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11390)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
@@ -10,11 +10,6 @@ import { IOverflowSet, IOverflowSetItemProps, IOverflowSetProps, IOverflowSetSty
 const getClassNames = classNamesFunction<IOverflowSetStyleProps, IOverflowSetStyles>();
 
 export class OverflowSetBase extends BaseComponent<IOverflowSetProps, {}> implements IOverflowSet {
-  public static defaultProps: Pick<IOverflowSetProps, 'vertical' | 'role'> = {
-    vertical: false,
-    role: 'menubar'
-  };
-
   private _focusZone = React.createRef<IFocusZone>();
   private _persistedKeytips: { [uniqueID: string]: IKeytipProps } = {};
   private _keytipManager: KeytipManager = KeytipManager.getInstance();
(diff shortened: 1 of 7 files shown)


packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -32,10 +27,12 @@ export class OverflowSetBase extends BaseComponent<IOverflowSetProps, {}> implem
   }
 
   public render(): JSX.Element {
-    const { items, overflowItems, className, focusZoneProps, styles, vertical, role, doNotContainWithinFocusZone } = this.props;
+    const { items, overflowItems, className, focusZoneProps, styles, vertical, doNotContainWithinFocusZone } = this.props;
 
     this._classNames = getClassNames(styles, { className, vertical });
 
+    const role = this.props.role || vertical ? 'menu' : 'menubar';
```
khmakoto: Isn't it weird that you apply `role='menu'` when people specify their own role?

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.types.ts
```diff
@@ -49,6 +49,7 @@ export interface IOverflowSetProps extends React.ClassAttributes<OverflowSetBase
 
   /**
    * Change item layout direction to vertical/stacked.
+   * Setting vertical to true also changes default role to "menu"
```
khmakoto:
```suggestion
   * Setting vertical to true also changes default role to "menu".
```

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
micahgodbolt: if this.props.role, then that role is used, otherwise we use menu or menubar depending on vertical prop.

PR #9578, opened 2019-06-25
Panel: Fluent Updates

marygans:
#### Pull request checklist

- [ ] Addresses an existing issue: #7920
- [x] Include a change request file using `$ npm run change`

#### Description of changes

Updates panel spacing and header & button alignment to reflect Fabric design spec in toolkit.

**Before**
before

**After**
after

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/9578)

PR #11391, opened 2019-12-06
Panel: header and navigation style updates

maxwellred:
#### Pull request checklist

- [ ] Addresses a style update provided through a toolkit from Chris Wymer

#### Description of changes

This incorporates some of the changes proposed in #9578.

The main change was allowing for the header to be on the same level as the close button (and the flex styles that went with it). To accomplish this I simply placed the header inside the classname "commands" container found in Panel.base.tsx.

I removed breakpoints as since in a previous PR it was recommended that I don't add breakpoint determined style changes since it introduces an extra level of specificity for the user to overcome when custom styling.

Conditionalized close IconButton back to its native 32x32 height/width.

Added new IPanelStyleProps `isHeaderInContentInner` in order to control where the header component is rendered as well as conditionalizing relevant styles.

#### Example:
Screen Shot 2019-12-05 at 4 36 09 PM

#### Instead of:
Screen Shot 2019-11-25 at 3 41 19 PM

#### Header and close button alignment (BEFORE)
Screen Shot 2019-11-25 at 4 42 22 PM

#### Header and close button alignment (AFTER)
Screen Shot 2019-11-25 at 3 41 43 PM

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11391)


packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -144,6 +144,7 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
     const isLeft = type === PanelType.smallFixedNear || type === PanelType.customNear ? true : false;
     const isRTL = getRTL();
     const isOnRightSide = isRTL ? isLeft : !isLeft;
+    const isHeaderInContentInner = !!(this.props.onRenderNavigation || this.props.onRenderNavigationContent);
```
micahgodbolt: how about we use a more semantic variable here. What we're really checking is if we have custom navigation or not. How about `hasCustomNavigation` or `navigationIsCustom`

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -210,7 +212,7 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
                 {onRenderNavigation(this.props, this._onRenderNavigation)}
               </div>
               <div className={_classNames.contentInner}>
-                {header}
+                {(isHeaderInContentInner || !hasCloseButton) && header}
```
micahgodbolt: then this will read hasCustomNavigation or !hasCloseButton - which is more descriptive of the state. 

like why would we ever check hasCloseButton if we already have a variable stating if the header is in the content?

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -295,28 +297,39 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
       // const iconButtonStyles = this._classNames.subComponentStyles
       // ? (this._classNames.subComponentStyles.iconButton as IStyleFunctionOrObject<IButtonStyleProps, IButtonStyles>)
       // : undefined;
+      const headerTextId = this.props.headerText && this.state.id + '-headerText';
+      const header = this._onRenderHeader(this.props, this._onRenderHeader, headerTextId);
+      const isHeaderInContentInner = !!(this.props.onRenderNavigation || this.props.onRenderNavigationContent);
```
micahgodbolt: can't we use the global one of these?

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -295,28 +297,39 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
       // const iconButtonStyles = this._classNames.subComponentStyles
       // ? (this._classNames.subComponentStyles.iconButton as IStyleFunctionOrObject<IButtonStyleProps, IButtonStyles>)
       // : undefined;
+      const headerTextId = this.props.headerText && this.state.id + '-headerText';
+      const header = this._onRenderHeader(this.props, this._onRenderHeader, headerTextId);
```
micahgodbolt: is this necessary? seems we could have a single function that determines headerTextID and renders the control, and share that between both render locations.

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
maxwellred: Thank you for these explanations!

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
maxwellred: added `_hasCustomNavigation` as a private member variable whose value is set by the expression you highlight above.

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
maxwellred: added `_header` as a private member whose value is set within the `render` function but is conditionally rendered either in the render function in its original location or within `_onRenderNavigationContent`

micahgodbolt:
This looks great. Just want to get this visual change communicated to partners before merging.

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
```diff
@@ -153,8 +153,8 @@ const getPanelBreakpoints = (type: PanelType): { [x: string]: IStyle } | undefin
 const commandBarHeight = '44px';
 
 const sharedPaddingStyles = {
-  paddingLeft: '16px',
-  paddingRight: '16px'
+  paddingLeft: '24px',
+  paddingRight: '24px'
```
Vitalius1: This change is the only one that is worrisome that it's not messing layouts. It's a 16px reduction in the overall width that was available to products to fit things in the panel and the probability that the entire width was taken into account by designers is high.

Vitalius1: @micahgodbolt partners need to know about this surface area reduction. It might easily break some layouts if the styling of the content inside was always counting on a specific width.

micahgodbolt: Yeah, part of this change is partner communication on Max's part.

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -166,9 +168,12 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
       isOnRightSide,
       isOpen,
       isHiddenOnDismiss,
-      type
+      type,
+      hasCustomNavigation: this._hasCustomNavigation
     });
 
+    this._header = onRenderHeader(this.props, this._onRenderHeader, headerTextId);
```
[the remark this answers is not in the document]
maxwellred: So what I've done in the most recent commit is remove this._header altogether and simply use the onRenderHeader() function in the two different locations: the render() and onRenderNavigationContent() functions

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -288,35 +290,43 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
...
-          }}
-          className={this._classNames.closeButton}
-          onClick={this._onPanelClick}
-          ariaLabel={closeButtonAriaLabel}
-          title={closeButtonAriaLabel}
-          data-is-visible={true}
-          iconProps={{ iconName: 'Cancel' }}
-        />
+        <React.Fragment>
+          {!this._hasCustomNavigation && onRenderHeader(this.props, this._onRenderHeader, this._headerTextId)}
```
aneeshack4: So when there's custom navigation, are we sure the textbook shouldn't be aligned - I think design wise, the 2nd looks more together:
Screen Shot 2019-12-05 at 4 36 09 PM

Instead of:
Screen Shot 2019-11-25 at 3 41 19 PM

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
```diff
@@ -153,8 +153,8 @@ const getPanelBreakpoints = (type: PanelType): { [x: string]: IStyle } | undefin
 const commandBarHeight = '44px';
 
 const sharedPaddingStyles = {
```
aneeshack4: Is there where the breakpoints you're referring to in the descriptions used to be? I'm a bit confused on what you mean by "breakpoint determined style changes"

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
maxwellred: replied in Teams

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
maxwellred: replied in Teams

maxwellred:
Closing reopening to see if that solves the strange license/cla check issue

Vitalius1:
Make sure you have the partners green light on this king of change. We don't want to have breaking visual changes that will not be caught by their build pipelines and with a very small chance actually getting into production.

aneeshack4:
@maxwellred Status on this? Is this blocked on getting the green light from partners?

micahgodbolt:
we're waiting until people get back from vacation to give them a chance to respond before merging

phkuo:
@maxwellred Do you know if this will work in SharePoint, which is currently using the mdl2-theme project style overrides in MSIT and prod? IMO, choose the least-work path forward. The font styles need to match, but the other styles with how things are positioned can be updated to the latest Fluent styles.

maxwellred:
Just for everyone's info per a DM in Teams between @phkuo and I, I checked to ensure that nothing looks out of place with the MDL2-theme applied plus in my updates I don't touch font styles so we should be good from that end.

Issue #11403, opened 2019-12-09
Cannot override autocomplete in PeoplePicker


dliu120:
### Environment Information

- **Package version(s)**: 6.195.6
- **Browser and OS versions**: Chrome Version 78.0.3904.108 (Official Build) (64-bit)

### Bug Description:

The autocomplete attribute is set to "off" in the PeoplePicker, but Chrome ignores it and overlays its own autofill picker in some cases. See this [stack overflow post](https://stackoverflow.com/questions/15738259/disabling-chrome-autofill). 

We want to override the inputProps attribute for autoComplete to be set to "no", but it does not seem to work: https://codepen.io/dliu120/pen/GRgJoJL (see line 74)

There is a similar issue here [#10376](https://github.com/OfficeDev/office-ui-fabric-react/issues/10376) that was closed due to inactivity. 

#### Actual behavior:

![image](https://user-images.githubusercontent.com/2090090/70482638-da6bb900-1a9a-11ea-9365-23fbad765764.png)

#### Expected behavior:

![image](https://user-images.githubusercontent.com/2090090/70482663-f3746a00-1a9a-11ea-94ff-9dcfae3742af.png)

### Priorities and help requested:

Are you willing to submit a PR to fix? No

Requested priority: Normal

Products/sites affected:  https://flow.microsoft.com

Notes: We have a workaround, but it's not ideal. We manually patch the input's autocomplete attribute after the component renders.

micahgodbolt:
Auto complete appears to be added to the input correctly in chrome

![image](https://user-images.githubusercontent.com/1434956/70484372-7ea42e80-1aa0-11ea-8579-84744822026c.png)


Are you seeing the same thing?

Also,  you mentioned 6.195.6, but your codepen example is in 7.x. Which version are you having trouble with?

micahgodbolt:
Ok, I can confirm the bug is in 6.x from looking at the code. Should be a quick fix.

dliu120:
> Auto complete appears to be added to the input correctly in chrome
> 
> ![image](https://user-images.githubusercontent.com/1434956/70484372-7ea42e80-1aa0-11ea-8579-84744822026c.png)
> 
> Are you seeing the same thing?
> 
> Also, you mentioned 6.195.6, but your codepen example is in 7.x. Which version are you having trouble with?

[...]

PR #11409, opened 2019-12-10
Autocomplete: Allow value to be overriden

micahgodbolt:
#### Pull request checklist

#### Description of changes

Moved autoCapitalize (and other static values) above the nativeInput props to allow those values to be overriden. 

Issue came up #11403. Will cherry pick this back to 6 to resolve that specific issue. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11409)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
@@ -272,9 +272,6 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends BaseComponent<
               {this.canAddItems() && (
                 <Autofill
                   spellCheck={false}
-                  autoCapitalize="off"
-                  autoComplete="off"
-                  aria-autocomplete={'both'}
                   {...inputProps as any}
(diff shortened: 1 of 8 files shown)


packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -283,11 +286,8 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends BaseComponent<
                   aria-describedby={items.length > 0 ? this._ariaMap.selectedItems : undefined}
                   aria-controls={`${suggestionsAvailable} ${selectedSuggestionAlertId}` || undefined}
                   aria-activedescendant={this.getActiveDescendant()}
-                  autoCapitalize="off"
```
joschect: Given that autofill has this already, we can probably remove it from both here and extended picker. Same with autocomplete.

PR #11418, opened 2019-12-10
Card: Adding keyboard navigation and interaction on clickable Cards and fixing example styling

### LINKED ISSUE #11406: Card: Missing keyboard navigation on clickable Cards
The `Card` component currently has an `onClick` property that makes it possible to interact with it via mouse and touch controls. However, the `Card` itself is not reachable via keyboard navigation, which is an omission since it breaks accessibility within the component. When the `onClick` property is provided, the `Card` should be reachable via keyboard.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11406

#### Description of changes

This PR adds tab stops and `keydown` implementations on clickable `Cards` so that we have correct keyboard interaction on them.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11418)

### DIFF
--- a/packages/react-cards/src/components/Card/Card.view.tsx
+++ b/packages/react-cards/src/components/Card/Card.view.tsx
@@ -90,12 +90,12 @@ export const CardView: ICardComponent['view'] = props => {
 
   return (
     <Slots.root
-      {...nativeProps}
-      horizontal={horizontal}
-      horizontalAlign={horizontal ? 'start' : 'stretch'}
       onClick={onClick}
       onKeyDown={_onKeyDown}
       tabIndex={onClick ? 0 : -1}
+      {...nativeProps}
+      horizontal={horizontal}
+      horizontalAlign={horizontal ? 'start' : 'stretch'}
       tokens={tokens as IStackComponent['tokens']}
       verticalAlign="start"
       verticalFill
--- a/packages/react-cards/src/components/Card/Card.view.tsx
+++ b/packages/react-cards/src/components/Card/Card.view.tsx
@@ -92,6 +92,7 @@ export const CardView: ICardComponent['view'] = props => {
     <Slots.root
       onClick={onClick}
       onKeyDown={_onKeyDown}
+      role={onClick ? 'button' : 'presentation'}
       tabIndex={onClick ? 0 : -1}
       {...nativeProps}
       horizontal={horizontal}
--- a/packages/react-cards/src/components/Card/Card.view.tsx
+++ b/packages/react-cards/src/components/Card/Card.view.tsx
@@ -1,7 +1,7 @@
 /** @jsx withSlots */
 import * as React from 'react';
 import { withSlots, getSlots } from '@uifabric/foundation';
-import { getNativeProps, htmlElementProperties, warn } from '@uifabric/utilities';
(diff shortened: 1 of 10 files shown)


packages/react-cards/src/components/Card/Card.view.tsx
```diff
@@ -81,14 +81,24 @@ export const CardView: ICardComponent['view'] = props => {
     }
   );
 
+  const _onKeyDown = (ev: React.KeyboardEvent): void => {
+    if (onClick && (ev.which === KeyCodes.enter || ev.which === KeyCodes.space)) {
+      onClick(ev as any);
```
xugao: 1) i think we should add a `onKeyDown` prop and invoke that instead `onClick`
2) we should not be casing to `any` unless we have to (have a good reason to)

packages/react-cards/src/components/Card/Card.view.tsx
```diff
@@ -81,14 +81,24 @@ export const CardView: ICardComponent['view'] = props => {
...
+      ev.preventDefault();
+    }
+  };
+
   return (
     <Slots.root
+      onClick={onClick}
+      onKeyDown={_onKeyDown}
```
Anschur: > onKeyDown={_onKeyDown} [](start = 5, length = 23)

If I understood this PR correctly, this should fix the issue for keyboard users. But what about screen reader users? How will they know that this is a clickable item? Is there a role being set? Or is there a clicktarget inside the card that will convey this information?

khmakoto: @Anschur I fixed this issue, please review again.

packages/react-cards/src/components/Card/Card.view.tsx
khmakoto: @xugao I've made some changes, please review again when you get the chance.

packages/react-cards/src/components/Card/examples/Card.Vertical.Example.tsx
```diff
@@ -99,13 +99,13 @@ export class CardVerticalExample extends React.Component<{}, {}> {
...
       <Stack horizontal tokens={sectionStackTokens}>
-        <Card>
+        <Card aria-label="Basic vertical card">
           <Card.Item>
             <Text>Basic vertical card</Text>
           </Card.Item>
         </Card>
 
-        <Card onClick={alertClicked} tokens={cardTokens}>
+        <Card aria-label="Clickable vertical card with image bleeding at the center of the card" onClick={alertClicked} tokens={cardTokens}>
```
Anschur: In a real world example you probably would include some information from the card in the aria-label. Otherwise this description would be the same for every card, and the user would have to navigate inside the card with scan-mode in order to get any information that is in the card.

khmakoto: That's true, however, because that's the concern of the developer using the controllers. Since these examples are meant to exemplify controllers with dummy data I don't think there is any issue with the `aria-label` being set as is.

PR #11422, opened 2019-12-10
CI: Add Parallelization

JasonGore:
#### Pull request checklist

#### Description of changes

This PR modifies ADO configuration to support parallelization.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11422)


JasonGore:
I had contemplated merging this before further improvements, but the benefit seems negligible.

Before: 34 minutes total build time
After: 33 minutes total build time

Since the benefit with this alone is negligible, I'll wait to incorporate this change until build and tested are separated, at which point we should get more benefits from parallel pipelines.

image

PR #11430, opened 2019-12-11
ColorPicker: added aria labels for hue and alpha sliders. Added strings to example

### LINKED ISSUE #11415: ColorPicker: Alpha slider does not have a dedicated aria label string
### Environment Information

- **Package version(s)**: 7.69.3 (or latest)
- **Browser and OS versions**:  Microsoft Edge 44.18362.449.0  and  Windows 10 Enterprise version 1903 OS build 18362.476

### Describe the issue:

User Impact:
People with Visual impairment will not be able to know how to access the sliders. 
In the Color picker control, Narrator should provide the instruction to the users how to access the sliders. 
For the horizontal sliders, users can access only using Right/Left arrow keys. So Narrator should provide the instruction. 

https://codepen.io/16f4cae9-74df-4efe-af3b-485541d4fa53/pen/wvBGByO?&editable=true

#### Actual behavior:
Narrator should provide the instruction to the users how to access the sliders. 
For the horizontal sliders, users can access only using Right/Left arrow keys. So Narrator should provide the instruction. 

#### Expected behavior:

Narrator should provide the instruction as 'Use right/left arrow keys to change the value' for horizontal sliders and 'Use Arrow keys to change the value' for slider.

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11415

#### Description of changes

Prior to this, the hue string was used as an aria label, and the alpha string was used as aria label and visible text.

Now `hue` is deprecated, and we have a consistent `hueAriaLabel` and `alphaAriaLabel`.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11430)


ecraig12345:
@dzearing you're supposed to be on vacation!! 🏖 

@micahgodbolt I have some comments but need to work on a demo right now, so please don't merge until I get to finish reviewing.

micahgodbolt:
@ecraig12345 have you had a chance to look at this?

packages/office-ui-fabric-react/src/components/ColorPicker/examples/ColorPicker.Basic.Example.tsx
```diff
@@ -41,6 +41,10 @@ export class ColorPickerBasicExample extends React.Component<{}, IBasicColorPick
           styles={colorPickerStyles}
           // The ColorPicker provides default English strings.
           // If your app is localized, you MUST provide the `strings` prop with localized strings.
+          strings={{
+            alphaAriaLabel: 'Use left and right arrow keys to change alpha value. Hold shift for a larger jump',
```
ecraig12345: I'm not convinced by the argument in #11415 that we need to provide instructions--the screen reader should announce that the control is a slider, which is a standard control that the user should already know how to interact with. (I only included instructions by default for the color rectangle because it's non-standard.) 

Or if we do want to provide an example of overrides it should be something like "Alpha slider. Use left and right arrow keys to change value. Hold shift for a larger jump." (lead with the important part)

micahgodbolt: The spec says arrow left/right for increment, and pageup/down for ~10% jumps. Since we deviate a bit from the spec, it's probably good to point out. I agree that leading with the slider name is a good idea. I'll check and make sure it's not redundant. 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role

micahgodbolt: @ecraig12345 updated the string to have slider name first, with instructions after.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -151,7 +151,7 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
             <div className={classNames.flexSlider}>
               <ColorSlider
                 className="is-hue"
-                ariaLabel={strings.hue}
+                ariaLabel={strings.hueAriaLabel || strings.hue}
```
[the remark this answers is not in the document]
micahgodbolt: ok, I can flip those around. I'll give hueAriaLabel the default value, and use the deprecated value if provided.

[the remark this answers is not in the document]
micahgodbolt: yuck...this gets messy quick :)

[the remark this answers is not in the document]
micahgodbolt: Ok, moved defaultValue to non deprecated hueAriaLabel and will check for strings.hue before using hueAriaLabel.

Removed default from strings.hue and strings.alphaAriaLabel.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -151,7 +151,7 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
             <div className={classNames.flexSlider}>
               <ColorSlider
                 className="is-hue"
-                ariaLabel={strings.hue}
+                ariaLabel={strings.hue || strings.hueAriaLabel}
```
ecraig12345: Should probably flip this
```suggestion
                ariaLabel={strings.hueAriaLabel || strings.hue}
```

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
```diff
@@ -131,15 +131,26 @@ export interface IColorPickerStrings {
   blue?: string;
 
   /**
-   * Label for the alpha text field and slider.
+   * Label for the alpha text field and slider aria-label.
```
ecraig12345:
```suggestion
   * Label for the alpha text field.
   * Also used as the aria label for the alpha slider if `alphaAriaLabel` is not provided.
```

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
```diff
@@ -131,15 +131,26 @@ export interface IColorPickerStrings {
...
+
   /**
    * Aria label for the hue slider.
    * @defaultvalue Hue
    */
+  hueAriaLabel?: string;
+
+  /**
+   * Aria label for the hue slider.
+   * @deprecated use 'hueAriaLabel'
```
ecraig12345: Nitpicky thing but it's generally good to use backticks for prop names so they get formatted as code in the docs.
```suggestion
   * @deprecated Use `hueAriaLabel`
```

packages/office-ui-fabric-react/src/components/ColorPicker/examples/ColorPicker.Basic.Example.tsx
```diff
@@ -41,6 +41,10 @@ export class ColorPickerBasicExample extends React.Component<{}, IBasicColorPick
           styles={colorPickerStyles}
           // The ColorPicker provides default English strings.
           // If your app is localized, you MUST provide the `strings` prop with localized strings.
+          strings={{
+            alphaAriaLabel: 'Alpha Slider: Use left and right arrow keys to change value, hold shift for a larger jump',
```
ecraig12345: Probably good to add something to the comment above explaining why only these strings are here, like: "You can also override specific strings to provide more detailed instructions."

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
micahgodbolt: already mentioned that hueAriaLabel now has the default string, so strings.hue should be used if provided

PR #11443, opened 2019-12-12
Panel: onLightDismissClick now properly prevents light dismiss

### LINKED ISSUE #11437: Panel: Confirm light dismiss- No button in the Diaog doesnt work as expected
In the case of Panel in office-ui-fabric-react example provided at following adress, The dialog doesnt behaive as expected
[Office UI Fabric](https://developer.microsoft.com/en-us/fabric/#/controls/web/panel)

#### Actual behavior:

When No Button on Dialog pressed, both of Dialog and Panel hide which is similar to Yes button but it is not as expected

#### Expected behavior:

When No Button on Dialog pressed, the dialog should hide but panel remains visible

micahgodbolt:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #11437

#### Description of changes

onLightDismissClick was billed as "Optional custom function to handle clicks outside the panel in lightdismiss mode", but clicks outside the panel were still calling dismiss even with that function passed through. This issue was evident in our light dismiss with 'Confirm light dismiss' example. 

My solution was to check for a preventDefault on the event so that button clicks in modal buttons (like the example) wouldn't always close the panel. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11443)

### DIFF
--- a/change/office-ui-fabric-react-2019-12-12-09-06-23-panel-lightdismiss-fix.json
+++ b/change/office-ui-fabric-react-2019-12-12-09-06-23-panel-lightdismiss-fix.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Panel: onLightDismissClick now properly prevents light dismiss",
+  "packageName": "office-ui-fabric-react",
+  "email": "mgodbolt@microsoft.com",
+  "commit": "b291e1361a914558025560c9193296202cd95d1d",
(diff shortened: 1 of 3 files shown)


micahgodbolt:
hmmm....thinking this through again

PR #11462, opened 2019-12-13
PersonaCoin: make divs role presentation

### LINKED ISSUE #11461: PersonaCoin: Div elements make it unusable inside buttons
Breakout of #11431

Personas are often used inside of buttons, but due to persona coins being built with divs they violate the role of buttons only containing phrasing content.

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11461

#### Description of changes

Persona coin is often used inside of buttons, but buttons are only supposed to have phrasing content (not flow). We can either change all of the divs to spans, make the elements configurable, or simply add role="presentation" because they are only there for layout.

This PR adds role="presentation" to all of the non aria-hidden divs. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11462)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
@@ -36,7 +36,7 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
       isPlaceholder
     });
 
-    const RootType = isImage ? 'div' : 'i';
+    const RootType = isImage ? 'span' : 'i';
     const nativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, htmlElementProperties, ['aria-label']);
     const { imageLoadError } = this.state;
     const imageProps: IImageProps = {
--- a/packages/office-ui-fabric-react/src/components/Persona/PersonaCoin/PersonaCoin.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Persona/PersonaCoin/PersonaCoin.base.tsx
@@ -100,10 +100,10 @@ export class PersonaCoinBase extends BaseComponent<IPersonaCoinProps, IPersonaSt
     );
 
     return (
-      <div {...divProps} className={classNames.coin}>
+      <div role="presentation" {...divProps} className={classNames.coin}>
         {// Render PersonaCoin if size is not size8. size10 and tiny need to removed after a deprecation cleanup.
         size !== PersonaSize.size8 && size !== PersonaSize.size10 && size !== PersonaSize.tiny ? (
-          <div {...divCoinProps} className={classNames.imageArea} style={coinSizeStyle}>
+          <div role="presentation" {...divCoinProps} className={classNames.imageArea} style={coinSizeStyle}>
             {shouldRenderInitials && (
               <div
                 className={mergeStyles(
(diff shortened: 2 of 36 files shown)


packages/office-ui-fabric-react/src/components/Persona/PersonaCoin/PersonaCoin.base.tsx
```diff
@@ -100,10 +100,10 @@ export class PersonaCoinBase extends BaseComponent<IPersonaCoinProps, IPersonaSt
     );
 
     return (
-      <div {...divProps} className={classNames.coin}>
+      <div role="presentation" {...divProps} className={classNames.coin}>
         {// Render PersonaCoin if size is not size8. size10 and tiny need to removed after a deprecation cleanup.
         size !== PersonaSize.size8 && size !== PersonaSize.size10 && size !== PersonaSize.tiny ? (
-          <div {...divCoinProps} className={classNames.imageArea} style={coinSizeStyle}>
+          <div role="presentation" {...divCoinProps} className={classNames.imageArea} style={coinSizeStyle}>
```
mtennoe: Do we need to set this for the PersonaPresence or Icon components as well in the tiny case?

micahgodbolt: presence is already aria-hidden i believe. 

icons are automatically aria-hidden unless they have aria-labeled by. So i think we should be good.

micahgodbolt: oh THIS isn't a good idea. If icon is an image it uses a div wrapper. That'll make problems if icons are in buttons. I can add that fix too

image

micahgodbolt: not 100% related to persona (doubt the icon would ever be an image), but i guess it's possible.

micahgodbolt: @mtennoe i think i got all of the divs. Do you see any I missed? The only one without 'presentation' is the one that is already aria-hidden

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
```diff
@@ -36,7 +36,7 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
       isPlaceholder
     });
 
-    const RootType = isImage ? 'div' : 'i';
+    const RootType = isImage ? 'span' : 'i';
```
mtennoe: Wont this be considered a breaking change?

packages/office-ui-fabric-react/src/components/Persona/PersonaCoin/PersonaCoin.base.tsx
mtennoe: Cool! Looks good to me. I left one small comment with regards to the div->span change

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
dzearing: if the root style is explicit about `display` css setting being `inline-block` or `block`, I *think* it would be mostly ok to have this. I think it is a larger majority of cases that icons are hosted within `button` elements, and given this html standards violation would come up any time image icons are used, it seems like a reasonable idea to do this.

mtennoe: Makes sense!

PR #11463, opened 2019-12-13
List: always render last page in list so user can 'End' key nav to the real last item in the list

### LINKED ISSUE #11346: [Keyboard Navigation] 'End' key does not focus the last item in List
### Environment Information

- **Package version(s)**: 7.59.3
- **Browser and OS versions**: Chrome 78.0.3904.108

### Describe the issue:
'End' key does not focus the last item in the list: it focuses the last item of the last page that is available in the DOM, and this is inconsistent with the behavior of the 'Home' button that always focuses the first item of the list.

https://codepen.io/avenezia/pen/jOEOVJg  (it corresponds to the codepen for your DetailsList - Basic page)

#### Actual behavior:
'End' key focuses the last item of the last page that is available in the DOM

#### Expected behavior:
'End' key focuses the last item of the list

xugao:
#### Pull request checklist
- [X] Addresses an existing issue: Fixes #11346

#### Description of changes
in case of rendering virtualized list, `End` key doesn't navigate to the __real__ last item in the list because if we didn't render the last item inside the DOM. the fix here is to always render the last page inside `List`.

#### Focus areas to test
- Made sure no unexpected unit test failures
- Manually tested all `DetailList` examples and didn't spot any issues
- Didn't find any assumptions we previously made for not having last item rendered

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11463)

### DIFF
--- a/apps/vr-tests/src/stories/ScrollablePane.GroupedDetailsList.stories.tsx
+++ b/apps/vr-tests/src/stories/ScrollablePane.GroupedDetailsList.stories.tsx
@@ -239,6 +239,7 @@ storiesOf('ScrollablePane Grouped Details List', module)
   .addDecorator(story => (
     <Screener
       steps={new Screener.Steps()
+        .wait(100)
         .snapshot('default: scrollbars should be visible', cropTo)
         .executeScript(`${getElement}.scrollTop = 100`)
         .snapshot('Scrollbars visibility when header is sticky', cropTo)
--- a/change/office-ui-fabric-react-2019-12-13-12-28-37-xgao-detaillist-2.json
+++ b/change/office-ui-fabric-react-2019-12-13-12-28-37-xgao-detaillist-2.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "List: always render last page in list so user can 'End' key nav to the real last item in the list",
+  "packageName": "office-ui-fabric-react",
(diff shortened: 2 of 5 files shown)


xugao:
will look into the screener test failures

xugao:
Regarding the screener test change:
The scroll bar takes a bit time to get to the final scroll length. this is __not__ a behavior introduced by my change. Adding a bit wait time upon initial render seems to make test to pass consistently. 
However I am a bit concerned why my change would require adding the wait time but not previously.
Happy to hear any guidance here on what i should check further. 
grouplist_screener_after

xugao:
Concerns over this change: 
1. it increases render time since now it always render the last page 
2. it kicks off API requests for the last page (depending on the consumer), no matter whether the user will ever scroll to the end.

dzearing:
Might be a fix for focuszone instead.
You could have opt in focusable bumpers on beginning and end which asynchronously pushes focus to next or previous. That way list is unaffected and stuff doesn’t get rendered or fetched unnecessarily.

xugao:
Thanks @dzearing for the suggestions. I am going to close this for now and let ODSP team to triage again

PR #11469, opened 2019-12-13
FocusZone: Adding tab handle for DomOrder direction

### LINKED ISSUE #11336: FocusZoneDirection.domOrder will break handleTabKey
In the FocusZone component I can use handleTabKey to get it to also control tab key actions. There is also direction attribute to control the direction of the arrow keys. These don't seem to affect each other with one exception: If I set direction={FocusZoneDirection.domOrder}, the arrow keys will follow the dom order int he same way as tab would usually do. But tab will no longer navigate inside the FocusZone.

### Environment Information

- **Browser and OS versions**: Chrome

https://codepen.io/svartn/pen/pooMZzE

#### Actual behavior:

Tab will not move between the list items when direction={FocusZoneDirection.domOrder}

#### Expected behavior:

Tab will not move between the list items in the same way as arrowign does.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11336

#### Description of changes

An issue recently came that specified that the `handleTabKey` prop was not working when `FocusZoneDirection.domOrder` was specified. This PR fixes this issue and adds a test so it doesn't regress in the future.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11469)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
+++ b/packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
@@ -581,7 +581,7 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
               !this._shouldWrapFocus(this._activeElement as HTMLElement, NO_HORIZONTAL_WRAP)
             ) {
               focusChanged = ev.shiftKey ? this._moveFocusUp() : this._moveFocusDown();
-            } else if (direction === FocusZoneDirection.horizontal || direction === FocusZoneDirection.bidirectional) {
+            } else {
               const tabWithDirection = getRTL() ? !ev.shiftKey : ev.shiftKey;
               focusChanged = tabWithDirection ? this._moveFocusLeft() : this._moveFocusRight();
             }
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -581,7 +581,7 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
               !this._shouldWrapFocus(this._activeElement as HTMLElement, NO_HORIZONTAL_WRAP)
             ) {
               focusChanged = ev.shiftKey ? this._moveFocusUp() : this._moveFocusDown();
-            } else if (direction === FocusZoneDirection.horizontal || direction === FocusZoneDirection.bidirectional) {
+            } else {
```
dzearing: is there a missing test case?

khmakoto: @dzearing just added it.

khmakoto:
I think the screener issues are unrelated to this change.

PR #11483, opened 2019-12-16
Sticky: Always render narrator readable content in place

### LINKED ISSUE #9597: ScrollablePane-Accessibility: Narrator can't read the Sticky headers in Scan Mode
## Codepen demonstrating the issue
https://developer.microsoft.com/en-us/fabric#/controls/web/scrollablepane
See Default example under 'Usage'.

## Actual behavior:
In Scan Mode, Narrator skips the header and reads the content immediately instead.

## Expected behavior:
In Scan Mode, Narrator reads the header and then the content.

## Documentation describing expected behavior
The actual non-Sticky headers don't render until they're visible in the Pane... that's why Narrator can't read them unless the user has scrolled up/down.

image

micahgodbolt:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #9597

#### Description of changes

Scrollable pane was moving sticky content out of flow. This change ensures that there is already readable text in the sticky area. 

#### Focus areas to test

This solution assumes that the nonStickyContent is not rendered in place. Will this break users who aren't using scrollable pane?

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11483)


micahgodbolt:
hmm...size auditor seems odd. There shouldn't be much more code to the file (though the output of a rendered sticky might appear larger)

micahgodbolt:
@leddie24 @Jahnp do you have time to review? This is one of our last remaining P1 bugs.

PR #11493, opened 2019-12-17
FocusZone: Adding paging down/up feature

### LINKED ISSUE #11507: [FocusZone][Accessibility][[Keyboard Navigation]: Adding paging down/up feature
Paging down/Up (update the focus element when we are paging Up/paging Down)

FocusZone

yes , with David Zearing

Adding paging down/up feature to the FocusZone for accessibility purposes (changing the page view without changing the focus element is not expected), also to make the FocusZone consistent with the other components like HTML <select> Tag and  file explorer which abstract arrow key navigation behaviors including directional arrow key ,page down/up, home and end.

nasabek:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11507

#### Description of changes
Adding paging down/up feature to the focusZone for accessibility purposes (changing the page view without changing the focus element is not expected), also to make the focusZone consistent with the other components like HTML Select Tag and  file explorer which abstract arrow key navigation behaviors including directional arrow key ,page down/up, home and end.

#### Focus areas to test
Manually tested the focusZone in 2 web applications:
1) _odsp-common\items-view\addon-photos_ which has a large collection of photos and didn't 
    spot any issues
2) Fabric UI simple starter demo


jspurlin:
Can you give a better description of your change (e.g. what's the issue and what change are you making to fix it?) Also, can you add validation you did for this change?

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -567,6 +568,16 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
             break;
           }
           return;
+        case KeyCodes.pageDown:
+          if (direction !== FocusZoneDirection.horizontal && this._movePageDown()) {
```
jspurlin: >(direction !== FocusZoneDirection.horizontal [](start = 13, length = 44)

Do we not want pageUp/PageDown to have the behavior you're proposing when we are in a horizontal FocusZone? Having pageUp/PageDown work on horizontal FocusZones seems like a valid thing that a user may want to do.

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -924,6 +935,208 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
 
     return false;
   }
+  private _movePageDown(): boolean {
```
jspurlin: I think there's a larger question on if we think pageUp/pageDown should always update focus if/when it scrolls.

If we want this functionality to be introduced, it should probably be done in a way that it doesn't break/change the current behavior.

Also, fyi @joschect , @micahgodbolt , and @dzearing .

nasabek: i discussed the idea before with @dzearing and he likes it :)

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
nasabek: yes , you are right.
i think we need it in the all cases(horizontal ,vertical)

JasonGore:
Size auditor failed due to an Azure issue. Merging in master (#11504) should fix this PR.

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
JasonGore: It's not clear to me if there's an action needed here. Does the PR need to be updated?

nasabek: yes, i update it now jason :)

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -924,6 +935,150 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
...
+          return LARGE_DISTANCE_FROM_CENTER;
+        } else {
+          if (leftAlignment >= targetRect.left && leftAlignment <= targetRect.left + targetRect.width) {
+            distance = 0;
+          } else {
+            distance = Math.abs(targetRect.left + targetRect.width / 2 - leftAlignment);
+          }
+        }
+        return distance;
+      })
```
jspurlin: This function looks almost exactly like the one in `_movePageDown` (outside of `targetRectTop < activeRectBottom` vs `targetRectBottom > activeRectTop`. It'd be great if this was pulled out into a helper function that took a function for this `if` check, and returned a function (e.g `(activeRect: ClientRect, targetRect: ClientRect) => number`) to pass here... You can also pull that signature into a type to make it easier to reason about, like: `type GetDistanceFromCenter = (activeRect: ClientRect, targetRect: ClientRect) => number;`

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -924,6 +935,150 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
...
+    if (!element || !this._root.current) {
+      return false;
+    }
+    if (this._isElementInput(element)) {
+      if (!this._shouldInputLoseFocus(element as HTMLInputElement, isForward)) {
+        return false;
+      }
+    }
+    const scrollableParent = findScrollableParent(element);
+    if (scrollableParent) {
```
jspurlin: Consider changing this to `if (!scrollableParent) { return false; }` and then you could return true where you're setting `changedFocus = true` currently and the last `return changedFocus` could be `return false`. I'm suggesting this because currently that `if(scrollableParent)` check  encompasses most of the body of the function

jspurlin: Also, it looks like most of the member variables you define at the beginning of this function could be scoped down because currently there's cases where you create those variables but they're never used

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -925,6 +936,126 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
...
+    // ClientRect values can be floats that differ by very small fractions of a decimal.
+    // If the difference between top and bottom are within a pixel then we should treat
+    // them as equivalent by using Math.floor. For instance 5.2222 and 5.222221 should be equivalent,
+    // but without Math.Floor they will be handled incorrectly.
+    const targetRectTop = Math.floor(targetRect.top);
+    const activeRectBottom = Math.floor(activeRect.bottom);
+    const targetRectBottom = Math.floor(targetRect.bottom);
+    const activeRectTop = Math.floor(activeRect.top);
+    // for paging down
+    if (isForward && targetRectTop < activeRectBottom) {
```
jspurlin: This check and the `else if` check could be pulled into one variable and then you could check that in the `if` and remove the `else if`... And it looks like at that point the `if` could return and then the current body of the `else` doesn't need to be in an else anymore

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -925,6 +936,126 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
...
+        const elementDistance = this._getHorizontalDistanceFromCenter(isForward, activeRect, targetRect);
+
+        // for paging down
+        if (isForward && targetRectTop > activeRectBottom + pagesize) {
+          break;
+        }
+        // for paging up
+        if (!isForward && targetRectBottom < activeRectTop - pagesize) {
+          break;
+        }
```
jspurlin: Same comment here, pull the body of the `if` checks into a variable and check that variable with one `if` statement to decide if it should break

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -947,26 +946,21 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
...
-      if (!this._shouldWrapFocus(this._activeElement as HTMLElement, NO_VERTICAL_WRAP)) {
-        return LARGE_NEGATIVE_DISTANCE_FROM_CENTER;
+    const isValidCandidateOnpagingDown = isForward && targetRectTop > activeRectBottom;
+    const isValidCandidateOnpagingUp = !isForward && targetRectBottom < activeRectTop;
+
+    if (isValidCandidateOnpagingDown || isValidCandidateOnpagingUp) {
+      if (leftAlignment >= targetRect.left && leftAlignment <= targetRect.left + targetRect.width) {
+        return 0;
+      } else {
+        return Math.abs(targetRect.left + targetRect.width / 2 - leftAlignment);
```
jspurlin: Look like this returns in every case so if the outer if is true then there doesn't need to be an else after it

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -925,6 +936,117 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
...
+      this.focusElement(candidateElement);
+      this._setFocusAlignment(candidateElement as HTMLElement, false, true);
+    } else if (this.props.isCircularNavigation && useDefaultWrap) {
+      if (isForward) {
+        return this.focusElement(getNextElement(
+          this._root.current,
+          this._root.current.firstElementChild as HTMLElement,
+          true
+        ) as HTMLElement);
+      } else {
```
jspurlin: Looks like this else is unneeded as well

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
nasabek: 👍

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
nasabek: 👍

JasonGore:
It's too bad that size auditor was broken during the review period. The main PR is showing [how big of a hit](https://github.com/OfficeDev/office-ui-fabric-react/pull/11661#issuecomment-572792689) components are taking, particularly components like `Rating` which will never have a use for this feature. We should probably consider re-examining how this is implemented so it can be opt-in somehow.

packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -925,6 +936,115 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
...
+      return Math.abs(targetRect.left + targetRect.width / 2 - leftAlignment);
+    } else {
+      if (!this._shouldWrapFocus(this._activeElement as HTMLElement, NO_VERTICAL_WRAP)) {
+        return LARGE_NEGATIVE_DISTANCE_FROM_CENTER;
+      }
+      return LARGE_DISTANCE_FROM_CENTER;
+    }
+  };
+
+  private _moveFocusPaging(isForward: boolean, useDefaultWrap: boolean = true): boolean {
```
JasonGore: This is nicely isolated. I think we could move this to an external utility and components that need `FocusZone` with paging could do something like:

```tsx
<FocusZone handlePaging={_moveFocusPaging} />
```

And the components like `Rating` and `Calendar` that will never have paging won't take the bundle hit.

PR #11494, opened 2019-12-17
GroupedList: Added built in keyboard support

### LINKED ISSUE #6318: GroupedList: Right/left arrow no longer expand and collapse groups
Mac OSX High Sierra, Chrome browser
 
Requested priority: (High) 
 
 OneDrive, SharePoint, etc...lots

There's no way to expand or collapse a group with keyboard.
Expected: With focus on the group header, right or left arrow should expand or collapse the group.

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #6318

#### Description of changes

(give an overview)

glist

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11494)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
@@ -1,12 +1,11 @@
 import * as React from 'react';
 import { IProcessedStyleSet, ITheme } from '../../Styling';
-import { classNamesFunction, getRTL, composeRenderFunction } from '../../Utilities';
+import { composeRenderFunction, classNamesFunction, getRTL, getRTLSafeKeyCode, KeyCodes } from '../../Utilities';
 import { SelectionMode } from '../../utilities/selection/index';
 import { Check } from '../../Check';
 import { Icon } from '../../Icon';
 import { GroupSpacer } from './GroupSpacer';
 import { Spinner } from '../../Spinner';
-import { FocusZone, FocusZoneDirection } from '../../FocusZone';
 import {
   IGroupHeaderStyleProps,
   IGroupHeaderStyles,
@@ -53,7 +52,7 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
   public render(): JSX.Element | null {
     const {
       group,
-      groupLevel,
+      groupLevel = 0,
       viewport,
       selectionMode,
       loadingText,
@@ -107,16 +106,18 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
         className={this._classNames.root}
         style={viewport ? { minWidth: viewport.width } : {}}
         onClick={this._onHeaderClick}
-        aria-expanded={!group.isCollapsed}
         aria-label={group.ariaLabel || group.name}
-        aria-level={groupLevel !== undefined ? groupLevel + 1 : undefined}
         aria-setsize={ariaSetSize}
         aria-posinset={ariaPosInSet}
         data-is-focusable={true}
+        onKeyUp={this._onKeyUp}
+        aria-expanded={!this.state.isCollapsed}
+        aria-level={groupLevel + 1}
       >
-        <FocusZone className={this._classNames.groupHeaderContainer} direction={FocusZoneDirection.horizontal}>
+        <div className={this._classNames.groupHeaderContainer}>
           {isSelectionCheckVisible ? (
             <button
+              data-is-focusable={false}
               type="button"
               className={this._classNames.check}
               role="checkbox"
@@ -137,10 +138,11 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
             <Icon iconName="Tag" />
           </div>
           <button
+            data-is-focusable={false}
             type="button"
             className={this._classNames.expand}
-            onClick={this._onToggleCollapse}
-            aria-expanded={!group.isCollapsed}
+            onClick={this._onToggleClick}
+            aria-expanded={group ? !group.isCollapsed : undefined}
             aria-controls={group && !group.isCollapsed ? groupedListId : undefined}
             {...expandButtonProps}
           >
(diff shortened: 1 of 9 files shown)


packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
```diff
@@ -148,7 +153,20 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
     if (onToggleCollapse) {
       onToggleCollapse(group!);
     }
+  };
+
+  private _onKeyUp = (ev: React.KeyboardEvent<HTMLElement>): void => {
+    const shouldOpen = this.state.isCollapsed && ev.which === getRTLSafeKeyCode(KeyCodes.right);
```
jdhuntington: This should pass theme to handle contextual RTL.

micahgodbolt: is that not what `getRTLSafeKeyCode` does?

micahgodbolt:
```
export function getRTLSafeKeyCode(key: number, theme: { rtl?: boolean } = {}): number {
  if (getRTL(theme)) {
    if (key === KeyCodes.left) {
      key = KeyCodes.right;
    } else if (key === KeyCodes.right) {
      key = KeyCodes.left;
    }
  }

  return key;
}
```

jdhuntington: Yes, it will use theme if passed. However, this call to `getRTLSafeKeyCode` is not passing theme, so it will fall back to global detection.

micahgodbolt: Ok that makes sense. it won't look at contextual theme unless it is passed in. Can do

micahgodbolt: pushed

packages/office-ui-fabric-react/src/components/GroupedList/GroupedListSection.tsx
```diff
@@ -367,7 +367,7 @@ export class GroupedListSection extends React.Component<IGroupedListSectionProps
         groupIndex={subGroupIndex}
         groupNestingDepth={nestingDepth}
         groupProps={groupProps}
-        headerProps={headerProps}
+        headerProps={{ ...headerProps, groupLevel: nestingDepth }}
```
KevinTCoughlin: This in-line Object spread will be a new object reference each render pass and cause a React re-render.

micahgodbolt: better to drop it into a const and then pass it in?

micahgodbolt: changed

micahgodbolt:
This is a pretty significant behavior change. I'm going to wait until the new year to solicit review from the greater community.

packages/office-ui-fabric-react/src/components/GroupedList/GroupedListSection.tsx
```diff
@@ -354,6 +354,8 @@ export class GroupedListSection extends React.Component<IGroupedListSectionProps
 
     const nestingDepth = subGroup.level ? subGroup.level + 1 : groupNestingDepth;
 
+    const groupHeaderProps = { ...headerProps, groupLevel: nestingDepth };
```
KevinTCoughlin: Unfortunately, changing `groupHeaderProps` to a `const` will not resolve the re-render of `GroupedListSection` issue mentioned in https://github.com/OfficeDev/office-ui-fabric-react/pull/11494#discussion_r359576553. With this change, `groupHeaderProps` is being assigned to a new object reference in memory. When `GroupedListSection` does a referential equality check in `shouldComponentUpdate` e.g. `oldProp === newProp` it will evaluate to `false` always, causing a re-render.

To resolve, either the object needs to not be recreated in-line each render pass when `groupLevel` does not change or special handling needs to be added to `GroupedListSection` given the current prop signature.

micahgodbolt: turns out this wasn't necessary afterall. Not sure why I even added it.

micahgodbolt:
@aneeshack4 i'm kinda curious why this is marked up as a grid in the first place. Seems a tree is more the experience that @betrue-final-final is referring to:

https://www.w3.org/WAI/GL/wiki/Using_ARIA_trees

> **Up**: Select the previous visible tree item.
> **Down**: Select next visible tree item.
> **Left**: Collapse the currently selected parent node if it is expanded. Move to the previous parent node (if possible) when the current parent node is collapsed.
> **Right**: Expand the currently selected parent node and move to the first child list item.
> **Enter**: Toggle the expanded or collapsed state of the selected parent node.
> **Home**: Select the root parent node of the tree.
> **End**: Select the last visible node of the tree.
> **Tab**: Navigate away from tree.

Though 'enter' would be different than our design

micahgodbolt:
@KevinTCoughlin mentioned scenarios with custom content in the header that needs to be accessible. Waiting for a design/example to try and mock up and see how it might work.

I'm worried to say the least. This interaction model doesn't really leave much room for accessible controls in the custom header. So either we can't support this interaction model, or we can't support actionable content in the header....or we need some other UI affordance to get focus into the header content.

dzearing:
Curious what the status is here and if we want to merge this or not.

micahgodbolt:
I haven't heard any compelling reason not to. @KevinTCoughlin have you come across any?

micahgodbolt:
Going to merge on Friday January 17th if I don't hear any compelling reasons not to.

micahgodbolt:
not merged yet....wanting to get some code signoff first

khmakoto:
@micahgodbolt do you think this will eventually be in a state where we can merge it? Is there any way I can help you with that? Or should we just close the PR for the time being?

micahgodbolt:
@khmakoto i'm ok to close it. the changes are good, but the possibility of breaking people is an issue. I'll keep the branch around and maybe we can tackle it in 8x, or when the need for a11y improvement outweight the need to not break people :D

fecuadri:
We are kinda blocked on this issue, since we can't ship our product without this being completely accessible. If not fixed, we are going to have to build our own.

micahgodbolt:
@fecuadri you could use the grouped-list code found in my branch?

cschlechty:
I am okay with this. We're going to have to revisit the semantics of this control at some point. Nested grids isn't working well.

micahgodbolt:
@khmakoto you did some aria updates on this control recently...Can you take a quick look over the code and make sure I didn't revert anything new/novel that you might have added.

micahgodbolt:
Ok, these changes are signed off. Merging in.

PR #11497, opened 2019-12-17
SearchBox, Pickers: update focus border styles

### LINKED ISSUE #11147: TextField/Dropdown border width on focus inconsistent with Picker components
### Environment Information

- **Package version(s)**: 7.59.3

The issue has been introduced in #10885

See the picker example page https://developer.microsoft.com/en-us/fabric#/controls/web/peoplepicker

#### Actual behavior:

Notice how the border width for the picker is 1px wide while TextField and Dropdown components have a 2px border width when the field has focus.
image

image

#### Expected behavior:

The border width should be consistent across the different input components.

Requested priority: Low

xugao:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #11147

#### Description of changes
- Apply border style update to SearchBox, TagPicker, PeoplePicker similar to [this fix](https://github.com/OfficeDev/office-ui-fabric-react/pull/10885) by Ben
- Noticed and fixed a bug with underlined `SearchBox` in HC, that it was showing border in all directions rather than just the bottom border
- Minor refactor to use 1 `getInputFocusBorder` everywhere

#### Screenshots
SearchBox:
image
image

Picker:
image
image
image
image

#### Focus areas to test
- Components: TextField, DocumentCard, Dropdown, Combobox, Searchbox, Pickers
- Modes: High contrast, focused 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11497)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts
@@ -13,6 +13,7 @@ import { IComboBoxOptionStyles, IComboBoxStyles } from './ComboBox.types';
 
 import { IButtonStyles } from '../../Button';
 import { memoizeFunction } from '../../Utilities';
+import { getInputFocusBorder } from '../TextField/TextField.styles';
 
 const ComboBoxHeight = 32;
 const ComboBoxLineHeight = 30;
@@ -301,27 +302,7 @@ export const getStyles = memoizeFunction(
       MsHighContrastAdjust: 'none'
     };
 
-    const getFocusBorder = (color: string): IStyle => ({
-      borderColor: color,
-      selectors: {
-        ':after': {
-          pointerEvents: 'none',
-          content: "''",
-          position: 'absolute',
-          left: -1,
-          top: -1,
-          bottom: -1,
-          right: -1,
-          border: '2px solid ' + color,
-          borderRadius: effects.roundedCorner2,
-          selectors: {
-            [HighContrastSelector]: {
-              borderColor: 'Highlight'
-            }
-          }
-        }
-      }
-    });
+    const focusBorderStyles: IStyle = getInputFocusBorder(root.borderPressedColor, effects.roundedCorner2);
 
     const styles: IComboBoxStyles = {
       container: {},
@@ -393,7 +374,7 @@ export const getStyles = memoizeFunction(
             [HighContrastSelector]: ComboBoxRootHighContrastFocused
           }
         },
-        getFocusBorder(root.borderPressedColor)
+        focusBorderStyles
       ],
 
       rootFocused: [
@@ -408,7 +389,7 @@ export const getStyles = memoizeFunction(
             [HighContrastSelector]: ComboBoxRootHighContrastFocused
           }
         },
-        getFocusBorder(root.borderFocusedColor)
+        focusBorderStyles
       ],
 
       rootDisabled: getDisabledStyles(theme),
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts
@@ -7,13 +7,13 @@ import {
   HighContrastSelector,
   IStyle,
   getPlaceholderStyles,
-  hiddenContentStyle
+  hiddenContentStyle,
+  getInputFocusStyle
 } from '../../Styling';
 import { IComboBoxOptionStyles, IComboBoxStyles } from './ComboBox.types';
 
 import { IButtonStyles } from '../../Button';
 import { memoizeFunction } from '../../Utilities';
-import { getInputFocusBorder } from '../TextField/TextField.styles';
 
 const ComboBoxHeight = 32;
 const ComboBoxLineHeight = 30;
@@ -302,7 +302,7 @@ export const getStyles = memoizeFunction(
       MsHighContrastAdjust: 'none'
     };
 
-    const focusBorderStyles: IStyle = getInputFocusBorder(root.borderPressedColor, effects.roundedCorner2);
+    const focusBorderStyles: IStyle = getInputFocusStyle(root.borderPressedColor, effects.roundedCorner2);
 
     const styles: IComboBoxStyles = {
       container: {},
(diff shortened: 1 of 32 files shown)


micahgodbolt:
can you drop in some before/after screenshots?

packages/office-ui-fabric-react/src/components/TextField/TextField.styles.tsx
```diff
@@ -61,6 +61,33 @@ function getLabelStyles(props: ITextFieldStyleProps): IStyleFunctionOrObject<ILa
   });
 }
 
+export const getInputFocusBorder = (
```
ecraig12345: I think rather than using the version from TextField everywhere, we should move this to the styling package. Importing TextField styles from controls which otherwise don't import TextField isn't ideal for bundle size reasons.

JasonGore:
Size auditor failed due to an Azure issue. Merging in master (#11504) should fix this PR.

micahgodbolt:
i got through 24/25 screener diffs with no problem......but then the last one was this:

image


Seems focus of underlined Textfield needs some tweaking

xugao:
@betrue-final-final , i found [combobox](https://developer.microsoft.com/en-us/fabric#/controls/web/combobox) has 2 borders w/ different colors when it's open :
image

I think we should remove line
https://github.com/OfficeDev/office-ui-fabric-react/blob/d911adae4a551db00e0316e9a0de4dd173e59464/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts#L304

can you pls confirm ?

ecraig12345:
Ben is out right now, but @maxwellred just added that style just recently and can probably confirm what's intended. I just looked at the design toolkit and it appears that the border is supposed to return to its initial color (not the focused color) when the menu is open, meaning the extra-wide focus border should also go away. 

We should probably also add a screener test for this scenario.

maxwellred:
@xugao Nice catch! I've looked at the most recent toolkit and @ecraig12345 is correct in that the border should turn back to the non-focused border upon menu expansion. The simplest way I found was to add the below to the '&.is-open' rule you highlighted.
```
':after': {
     content: 'none'
 }
```

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts
```diff
@@ -301,27 +302,7 @@ export const getStyles = memoizeFunction(
...
-          borderRadius: effects.roundedCorner2,
-          selectors: {
-            [HighContrastSelector]: {
-              borderColor: 'Highlight'
-            }
-          }
-        }
-      }
-    });
+    const focusBorderStyles: IStyle = getInputFocusStyle(root.borderPressedColor, effects.roundedCorner2);
```
joschect: Woooooooohhh!

micahgodbolt:
Make sure to take a look at #11565 as well. It seems you've already noticed the combobo box border issue.

micahgodbolt:
**Combobox**

image


**vs Dropdown**

image

micahgodbolt:
underlined and high contrast both look good.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts
```diff
@@ -360,6 +341,9 @@ export const getStyles = memoizeFunction(
             '&.is-open': {
               borderColor: root.borderColor,
               selectors: {
+                ':after': {
```
micahgodbolt: @betrue-final-final and @jspurlin  just to verify: when the combobox is open, should focus rect be around the input or not? Hypothetically the input still has focus (you can type there), though if we follow Dropdown's example, there is no focus border when it is open.

If we ARE removing the focus border when it is open, it seems the chevron button has a bit of a gap on the top:

image

xugao: good catch on the style of the chevron button. once i get confirmation from design, i will address accordingly

jspurlin: The focus rect should still be around the whole combobox since "it" (e.g. the input) has focus

betrue-final-final: Ok, that makes sense, sort of. Combo is always a little confusing. I assume that means that Dropdown remains as it is today without focus as the menu has focus?

xugao: hmm. In dropdown case, menu has the focus. In combobox case, the entire combobox has focus (both the input and the menu has focus).

betrue-final-final: Sounds good.

xugao: thinking a bit more and checked with Ben, will add focus border (remove the redundant black border) back for ComboBox, dropdown remain the same

ecraig12345: Per @maxwellred's earlier comment https://github.com/OfficeDev/office-ui-fabric-react/pull/11497#issuecomment-567603193, the current design in the toolkits is that ComboBox should NOT have the focus border when open.

xugao: Latest: 
image

xugao: @micahgodbolt - on the chevron button style issue you pointed out, it's unrelated to this PR so im going to issue a separate PR to fix instead of adding more changes to this one

betrue-final-final: The toolkit should be updated @maxwellred

micahgodbolt:
I still have some concerns with combobox focus outline looking odd at some screen resolutions (i'm assuming it's sub pixel rendering).

Currently its REALLY broken (this is prod)

image

And while this PR improves the border, it's still not quite right

image


This happens on chrome and chromium edge

Then here's Firefox:

image

xugao:
@micahgodbolt - nice catch! I opened a separate issue to track the couple rendering issue with ComboBox: https://github.com/OfficeDev/office-ui-fabric-react/issues/11631

the 2 issues are kinda related, and i will fix them in a separate change after this one

PR #11506, opened 2019-12-18
Size Auditor: Re-enable

JasonGore:
#### Pull request checklist

#### Description of changes

Size auditor had to be temporarily disabled (#11504) due to [an Azure issue](https://developercommunity.visualstudio.com/content/problem/859511/buildsourceversionmessage-no-longer-works-to-fetch.html
). This PR is being created in draft state to re-enable it as soon as possible, either by waiting for status on the Azure issue or changing size auditor to accommodate changed Azure behavior.

We should also do a sanity check on bundle sizes when it's re-enabled to make sure nothing substantial crept in while size auditor was disabled.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11506)

### DIFF
--- a/azure-pipelines.bundlesize.yml
+++ b/azure-pipelines.bundlesize.yml
@@ -47,13 +47,13 @@ jobs:
           PathtoPublish: 'packages/office-ui-fabric-react/dist/office-ui-fabric-react.js'
           ArtifactName: oufrdrop
 
-  # - job: lightrail
-  #   pool: server
-  #   dependsOn: build
-  #   steps:
-  #     - task: odefun.odsp-lightrail-tasks-partner.odsp-lightrail-tasks-SizeAuditorWorker.SizeAuditorWorker@0
-  #       displayName: 'Size Auditor Check on LightRail'
-  #       inputs:
-  #         connectedServiceName: lowimpact
-  #         sourceVersionMessage: '$(Build.SourceVersionMessage)'
-  #         sourceRepositoryUrl: 'https://github.com/OfficeDev/office-ui-fabric-react'
+  - job: lightrail
+    pool: server
+    dependsOn: build
+    steps:
+      - task: odefun.odsp-lightrail-tasks-partner.odsp-lightrail-tasks-SizeAuditorWorker.SizeAuditorWorker@0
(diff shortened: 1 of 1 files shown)


azure-pipelines.bundlesize.yml
```diff
@@ -47,13 +47,13 @@ jobs:
...
-  #         sourceRepositoryUrl: 'https://github.com/OfficeDev/office-ui-fabric-react'
+  - job: lightrail
+    pool: server
+    dependsOn: build
+    steps:
+      - task: odefun.odsp-lightrail-tasks-partner.odsp-lightrail-tasks-SizeAuditorWorker.SizeAuditorWorker@0
+        displayName: 'Size Auditor Check on LightRail'
+        inputs:
+          connectedServiceName: lowimpact
+          sourceVersionMessage: '$(Build.SourceVersionMessage)'
```
aftab-hassan: @JasonGore, from the [link](https://developercommunity.visualstudio.com/content/problem/857818/cannot-read-buildsourceversionmessage-when-check-i.html) you mentioned in PR, looks like adding double quotes could work

So, `sourceVersionMessage: "$(Build.SourceVersionMessage)"`

We should see the result when pushing the commit, so low-risk. I don't have a dev environment with me.

JasonGore: Yes, I tried without quotes and with double quotes in #11504 and reported my findings in the Azure issue. None of that has worked so far.

ecraig12345:
Still failing due to missing Build.SourceVersionMessage :(

JasonGore:
I'm subscribed to the issues in the PR desc. They say they have a fix but is not yet deployed. I'll rerun the jobs when the issue changes state.

JasonGore:
A recent [6.0 PR](https://github.com/OfficeDev/office-ui-fabric-react/pull/11661#issuecomment-572792689) just reported size auditor results so I'm retesting this PR.

JasonGore:
We're back in business!

PR #11523, opened 2019-12-19
Adding pointer down support for dismissing the fabric base layer.

### LINKED ISSUE #11396: FabricBase not setting isFocusVisible for pointer events of type 'touch' or 'pen'
### Environment Information

### Describe the issue:

On pen down or touch down the callouts are not dismissed because isFocusVisible state of FabricBase is not set to false due to it not listening to other pointer events besides mousedown.

#### Actual behavior:

Callout remains visible.

#### Expected behavior:

Callout is dismissed.

Change https://github.com/OfficeDev/office-ui-fabric-react/blob/ae503099db790f5df2f35c8969efb4f47bc88898/packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx

onComponentDidMount to handle pointerdown instead of mousedown:

    const win = getWindow(this._rootElement.current);
    if (win) {
      this._disposables.push(on(win, '**pointerdown**', this.**_onPointerDown**, true), on(win, 'keydown', this._onKeyDown, true));
    }
    this._addClassNameToBody();
  }

Yes

Normal

PowerPoint Online. We did a workaround but it is hacky.

albertborges:
#### Pull request checklist

Fixes #11396

#### Description of changes

Adding pointer down support for dismissing the fabric base layer.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11523)


ecraig12345:
Thanks for the contribution! There are a couple issues though, which you'll need to [set up a local copy of Fabric](https://github.com/OfficeDev/office-ui-fabric-react/wiki/Setup) to fix:

1. Need to add a change file using `yarn change`

2. Rather than adding a separate `_onPointerDown`, I'd recommend reusing the existing `_onMouseDown` since it does the same thing.

micahgodbolt:
pinged the author over teams.

albertborges:
Responded to all feedback.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -87,7 +91,7 @@ export class FabricBase extends React.Component<
     }
   }
 
-  private _onMouseDown = (ev: MouseEvent): void => {
+  private _onPointerDown = (ev: PointerEvent): void => {
```
khmakoto: I don't know if @ecraig12345 preferred using the `_onMouseDown` name that was already there, but since you're at least not using different methods this seems fine to me.

ecraig12345: Either is fine with me.

albertborges:
@dzearing

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
dzearing: Shouldn't this be both? Does onPointerDown get called when mouse down is fired?

albertborges: onPointerDown does get called when mouse down is fired.

xugao: just checked, `onPointerDown` does fire on mouse down and all browsers support pointer down. we should remove `on(win, 'mousedown', this._onPointerDown, true),`

albertborges: https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onpointerdown

^According to that, Safari and Opera do not.

albertborges: so I'm a bit hesitant to remove that.

xugao: indeed. i am not sure how i saw "all browsers support" :)

xugao: Ah, this PR has been hanging for a while, we should close on this :)

re-reading the comments, I think here is what we need to do / what David was originally suggesting:
we should have 2 handlers - 
```
private _onMouseDown = (ev: MouseEvent): void => {
   this.setState({ isFocusVisible: false });
};

private _onPointerDown = (ev: PointerEvent): void => {
if (event.pointerType !== 'mouse') { // mouse type is handled by _onMouseDown
   this.setState({ isFocusVisible: false });
}
};
```

xugao:
Just a note here after chatting with David: 
we probably want to use `initializeFocusRects` in `Fabric` to handle this similar to other components. a problem we have today with `initializeFocusRects` is that it does not clean up the event listeners properly. Let's proceed with this change for now and  I will take this work as a follow up.

PR #11524, opened 2019-12-20
Ensure icon font family isn't overridden by Fabric Core

### LINKED ISSUE #10449: IconButton is broken with icons that are not supported by fabric-core
### Environment Information

- office-ui-fabric-react 7@latest
  Fabric-Core css @10.1.0
- Chrome@latest, Windows 10

`IconButton` is broken with icon that is not supported by Fabric Core.
Seems one additional `ms-Icon` class is added.
Is there any workaround for this?
https://codepen.io/chnldw/pen/jONpKGe?editors=1010

**Repro:**
1. include fabric-core css.
2. Create an IconButton with `PageList` Icon.

**Expected:**
PageList icon rendered correctly.

**Actual:**
Icon is not rendered. It was working on previous version. around version 7.11

image

Requested priority: (High)

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10449

#### Description of changes

In `FontIcon`, apply the proper Fabric React icon `fontFamily` using the element's `style` prop (rather than a className) to ensure that the correct font always takes precedence. This means that even if a mismatched Fabric Core version is present on the page, its `ms-Icon` font family won't override the one defined for Fabric React icons.

#### Background

When I added the optimized `FontIcon` and `ImageIcon`, I gave them the className `ms-Icon` to provide a way to target them for customization (as an optimization, the new icons can't be targeted the standard way through `scopedSettings`). 

This unexpectedly created problems because while Fabric React doesn't directly apply any styling using the `ms-Icon` class, Fabric Core does, creating potential `font-family` conflicts. If an icon was available in Fabric React's icon font, but some version of Fabric Core which did **not** have that icon yet was loaded on the page first, the outdated `font-family` defined for `ms-Icon` in Fabric Core would be applied to the Fabric React icon and prevent the icon character from displaying.

Potential downside of applying the proper `font-family` using `style` is that it can't be overridden using normal CSS styles. But I don't think that's a big concern in this case because if someone wants to change the font family for Fabric React icons, they should be doing that through `registerIcons`/`initializeIcons`.

This workaround is only necessary in `FontIcon`. Traditional `Icon` doesn't have the issue because it doesn't use `ms-Icon`, and `ImageIcon` doesn't practically have the issue because it doesn't depend on fonts.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11524)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.test.tsx
@@ -522,7 +522,7 @@ describe('DetailsHeader', () => {
     expect(header._currentDropHintIndex).toBe(1);
     let dropHintElement = component.find('#columnDropHint_1').getDOMNode();
     let dropHintElementChildren = dropHintElement.children;
-    expect(dropHintElementChildren.item(0)!.getAttribute('style')).toEqual('display: inline-block;');
+    expect(dropHintElementChildren.item(0)!.getAttribute('style')).toContain('display: inline-block;');
     expect(dropHintElementChildren.item(1)!.getAttribute('style')).toEqual('display: inline-block;');
 
     // dragover b/w b&c and c&d -> dead zone b/w 370 and 810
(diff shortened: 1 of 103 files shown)


packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.test.tsx
```diff
@@ -645,7 +645,7 @@ describe('DetailsHeader', () => {
 
     dropHintElement = component.find('#columnDropHint_1').getDOMNode();
     dropHintElementChildren = dropHintElement.children;
-    expect(dropHintElementChildren.item(0)!.getAttribute('style')).toEqual('display: none;');
+    expect(dropHintElementChildren.item(0)!.getAttribute('style')).toContain('display: none;');
```
aneeshack4: why this change `toContain`? It's not an exact match anymore?

khmakoto: I guess not, since `style` now has `fontFamily` in it from my understanding.

ecraig12345: Yes that's why

PR #11526, opened 2019-12-20
SwatchColorPicker: properly apply ID; naming and doc updates

### LINKED ISSUE #11520: Property 'Id' not working for SwatchColorPicker
### Environment Information

- **Package version(s)**: 7.74.2

https://codepen.io/LukeO/pen/gObgVLp

#### Actual behavior:

`<SwatchColorPicker id={"someid"} ... ` has no effect. There is no component with id 'someid' (only the color fields will receive an id starting with 'someid-...'.

#### Expected behavior:

There is should be a component with id 'someid' which can be used to reference the picker, e.g. for [adding a label for the picker](https://github.com/OfficeDev/office-ui-fabric-react/issues/11480#event-2885588433).

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11520

#### Description of changes

SwatchColorPicker's `id` prop should be applied to the root of the grid.

Also "renamed" some SwatchColorPicker-related props (add new prop, deprecate old one) which were badly named:
- `IColorPickerGridCellProps`:
  - `id` to `idPrefix`, because it's actually just used as a prefix for a generated ID
- `ISwatchColorPickerProps` and `IGridProps`:
  - `positionInSet` to `ariaPosInSet` to match standard naming
  - `setSize` to `ariaSetSize` to match standard naming

And assorted doc updates in SwatchColorPicker and related components.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11526)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
@@ -12,13 +12,15 @@ import { ISwatchColorPickerProps, ISwatchColorPickerStyleProps, ISwatchColorPick
 import { Grid } from '../../utilities/grid/Grid';
 import { IColorCellProps } from './ColorPickerGridCell.types';
 import { ColorPickerGridCell } from './ColorPickerGridCell';
-import { memoizeFunction } from '@uifabric/utilities';
+import { memoizeFunction, warnDeprecations } from '@uifabric/utilities';
 
 export interface ISwatchColorPickerState {
   selectedIndex?: number;
 }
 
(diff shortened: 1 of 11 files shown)


packages/office-ui-fabric-react/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -116,21 +118,21 @@ export class SwatchColorPickerBase extends React.Component<ISwatchColorPickerPro
     return (
       <Grid
         {...this.props}
+        id={this._id}
```
ecraig12345: The actual bug fix

PR #11527, opened 2019-12-20
Removed circular import (cherry-pick to 5.0)

LeopoldLerch:
#### Pull request checklist

Cherry-pick of #10972 

#### Description of changes

Removed the circular import in positioning.types.ts in Utilities. Did import from itself, which crashed with Typescript 3.7 (does not allow importing names that are locally defined)

However, in order to to a npm install I had to remove the dev-dependency 

"office-ui-fabric-react-tslint": ">=5.0.0 <6.0.0",

which seems to be not available in npm registry any more. This is not checked in, as I did not want to. If it is necessary, pls have someone a look at it if there are alternatives or if it can be removed safely.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11527)

### DIFF
--- a/common/changes/office-ui-fabric-react/2019-12-31-02-43.json
+++ b/common/changes/office-ui-fabric-react/2019-12-31-02-43.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "office-ui-fabric-react",
+      "comment": "Removed the circular import in positioning.types.ts",
+      "type": "patch"
+    }
+  ],
+  "packageName": "office-ui-fabric-react",
+  "email": "not specified"
+}
--- a/packages/office-ui-fabric-react/src/utilities/positioning/positioning.types.ts
+++ b/packages/office-ui-fabric-react/src/utilities/positioning/positioning.types.ts
@@ -1,6 +1,5 @@
 
 import { DirectionalHint } from '../../common/DirectionalHint';
-import { IPoint } from './positioning.types';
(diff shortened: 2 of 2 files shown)


ecraig12345:
Thanks for doing this! Looks like you need to run `npm run change` to generate a change file  (that's why the build is failing).

> However, in order to to a npm install I had to remove the dev-dependency
> 
> "office-ui-fabric-react-tslint": ">=5.0.0 <6.0.0",
> 
> which seems to be not available in npm registry any more. This is not checked in, as I did not want to. If it is necessary, pls have someone a look at it if there are alternatives or if it can be removed safely.

This is a local/private package which exists in the 5.0 branch of OUFR. It shouldn't affect external installs of `office-ui-fabric-react@5` since they don't install dev dependencies. If you're using the [build commands for 5.0](https://github.com/OfficeDev/office-ui-fabric-react/wiki/Build-Commands-(5.0,-6.0-Branches)) (and ideally [put your local copy of 5.0 in a separate clone](https://github.com/OfficeDev/office-ui-fabric-react/wiki/Legacy-Branches#setting-up-for-legacy-branch-development)), it shouldn't cause problems locally either. The build commands are slightly different for 5.0 due to some tooling changes made since then.

ecraig12345:
I pushed a commit with a change file, so this should be ready to go now.

PR #11535, opened 2019-12-21
Modal ignoreExternalFocusing prop should also affect Popup

### LINKED ISSUE #11514: Modal ignoreExternalFocusing props doesn't work
### Environment Information

- **Browser and OS versions**: chrome windows 10

https://codepen.io/ipip2005/pen/JjoExPj

Click "Open dialog" button
Then click "Close" button in the dialog.

#### Actual behavior:

It focuses on the "Open dialog" button.

#### Expected behavior:

While I'm specifying `ignoreExternalFocusing` to `true`, I expect focus will be lost.

Modal ignoreExternalFocusing props doesn't work because it's not passing the corresponding props to Popup
Should it pass 
```
            shouldRestoreFocus={!this.props.ignoreExternalFocusing}
```
To Popup?

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11514

#### Description of changes

Modal/Dialog have a (misleadingly named) `ignoreExternalFocusing` prop which if true means that when the modal closes, it shouldn't re-focus the element which opened the modal. Popup (underlying component) *also* has the capability to re-focus elements, controlled by its `shouldRestoreFocus` prop. #11514 pointed out that even if re-focusing the opening element was disabled at the Modal level, that wasn't being passed through to the Popup--now fixed.

I also updated the docs for various `IAccessiblePopupProps` (Modal/Dialog shared) and `IFocusTrapZoneProps` members to be clearer/more accurate, and gave `IAccessiblePopupProps` a `docCategory` to make it actually show up on the website.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11535)

### DIFF
--- a/change/office-ui-fabric-react-2019-12-20-13-17-02-modal-focus.json
+++ b/change/office-ui-fabric-react-2019-12-20-13-17-02-modal-focus.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Modal ignoreExternalFocusing should also affect Popup",
+  "packageName": "office-ui-fabric-react",
+  "email": "elcraig@microsoft.com",
+  "commit": "30f3293eb58cb16866a2005a6ef54fceef24dd5d",
+  "date": "2019-12-20T21:17:02.143Z"
+}
--- a/change/office-ui-fabric-react-2019-12-20-13-17-02-modal-focus.json
+++ b/change/office-ui-fabric-react-2019-12-20-13-17-02-modal-focus.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Modal ignoreExternalFocusing should also affect Popup",
+  "comment": "Modal ignoreExternalFocusing prop should also affect Popup",
   "packageName": "office-ui-fabric-react",
   "email": "elcraig@microsoft.com",
   "commit": "30f3293eb58cb16866a2005a6ef54fceef24dd5d",
--- a/packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
+++ b/packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
@@ -1,5 +1,5 @@
 /**
- * {@docCategory Modal}
+ * {@docCategory IAccessiblePopupProps}
  */
 export interface IAccessiblePopupProps {
   /**
--- a/packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
+++ b/packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
@@ -1,3 +1,6 @@
+/**
+ * {@docCategory Modal}
+ */
 export interface IAccessiblePopupProps {
   /**
    * Sets the HTMLElement to focus on when exiting the FocusTrapZone.
--- a/packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
+++ b/packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
@@ -3,35 +3,36 @@
  */
 export interface IAccessiblePopupProps {
   /**
-   * Sets the HTMLElement to focus on when exiting the FocusTrapZone.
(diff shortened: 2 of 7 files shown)


change/office-ui-fabric-react-2019-12-20-13-17-02-modal-focus.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
JasonGore: Should this be minor due to a new interface being exposed?

packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
```diff
@@ -1,34 +1,38 @@
+/**
+ * {@docCategory Modal}
```
JasonGore: Is `Modal` correct here? Curious since this is in a `common` directory.

packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
```diff
@@ -1,34 +1,38 @@
...
-   * Sets the HTMLElement to focus on when exiting the FocusTrapZone.
-   * @defaultvalue The element.target that triggered the Panel.
+   * Sets the element to focus on when exiting the control's FocusTrapZone.
+   * @defaultvalue The `element.target` that triggered the control opening.
    */
   elementToFocusOnDismiss?: HTMLElement;
 
   /**
-   * Indicates if this dialog will ignore keeping track of HTMLElement that activated the Zone.
+   * If false (the default), the control's FocusTrapZone will restore focus to the element which
```
JasonGore: I wonder if `FocusTrapZone` is an implementation detail here that may not necessarily apply and noticed some interface elements that are duplicated in `FocusTrapZone` below. Should this interface be a `FocusTrapZone` type, possibly implemented by `FocusTrapZone` props?

packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
ecraig12345: The options I thought of were either Modal, Dialog, or a separate docCategory which would land under References.

packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
ecraig12345: I'm not sure of the background on the duplicated interface elements (or of why IAccessiblePopupProps exists), though I agree that it's weird. Could you clarify what you mean by the second sentence of your comment?

JasonGore: I'm wondering if it makes more sense to have in `FocusTrapZone.types.ts`:

```tsx
export interface IAccessiblePopupProps { ... };

export interface IFocusTrapZoneProps extends IAccessiblePopupProps, ... { ... };
```

I know it's a bit outside of the scope of this PR, but I wanted to raise the question before we 
 start exporting `IAccessiblePopupProps`. Although I now see `IDialogProps` is already exporting it, so maybe the point is moot.

ecraig12345: That makes sense. Since I'm already doing this I'd like to fix it the right way.

packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
ecraig12345: I decided to change to a separate category which will go under References

packages/office-ui-fabric-react/src/common/IAccessiblePopupProps.ts
ecraig12345: After some discussion we're leaving this as-is for now.

PR #11538, opened 2019-12-21
Selection: accept generic item type

### LINKED ISSUE #11484: Selection types should not expect items to have a `key` field, since `getKey()` is customizable
### Environment Information

- **Package version(s)**: Latest

TypeScript errors don't show up there for some reason

#### Actual behavior:

The `getKey` callback in the `Selection` constructor should allow me to specify my own field to use as a key, but the TypeScript definitions expect a field called `key` to always be there. 

For example, I tried to make it so that `getKey` would return my object's `Id` field. The code looks like this:

```ts
this.selection = new Selection({
      getKey: item => item.Id,
      ...
})
```

I get a TypeScript error that says that: 

> "Property 'Id' does not exist on type 'IObjectWithKey'.'. 

The issue appears to be that `getKey` expects the item passed in to be of type `IObjectWithKey`, which you can see [here](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/utilities/src/selection/Selection.ts#L9).

This also applies to all of `Selection`'s public methods, like `setItems` (a similar error occurs):

```ts
selection.setItems(items);
```

> Argument of type 'AccountDetails[]' is not assignable to parameter of type 'IObjectWithKey[]'.
  Type 'AccountDetails' has no properties in common with type 'IObjectWithKey'.

#### Expected behavior:

A TypeScript error should not appear in `getKey` when passing an item without a `key` field, or in any of `Selection`'s public methods, like `setItems`. 

Products/sites affected: (if applicable) N/A

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11484

#### Description of changes

Since Selection supports fully customizing how keys are generated using `getKey`, it makes sense to allow a generic item type instead of just `IObjectWithKey`. I kept `IObjectWithKey` as the default, so this shouldn't be a breaking change.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11538)

### DIFF
--- a/packages/utilities/src/selection/Selection.ts
+++ b/packages/utilities/src/selection/Selection.ts
@@ -4,26 +4,26 @@ import { EventGroup } from '../EventGroup';
 /**
  * {@docCategory Selection}
  */
-export interface ISelectionOptions {
+export interface ISelectionOptions<TItem = IObjectWithKey> {
   onSelectionChanged?: () => void;
-  getKey?: (item: IObjectWithKey, index?: number) => string | number;
-  canSelectItem?: (item: IObjectWithKey, index?: number) => boolean;
+  getKey?: (item: TItem, index?: number) => string | number;
+  canSelectItem?: (item: TItem, index?: number) => boolean;
   selectionMode?: SelectionMode;
 }
 
 /**
  * {@docCategory Selection}
  */
-export class Selection implements ISelection {
+export class Selection<TItem = IObjectWithKey> implements ISelection<TItem> {
   public count: number;
   public readonly mode: SelectionMode;
 
-  private _getKey: (item: IObjectWithKey, index?: number) => string | number;
-  private _canSelectItem: (item: IObjectWithKey, index?: number) => boolean;
+  private _getKey: (item: TItem, index?: number) => string | number;
+  private _canSelectItem: (item: TItem, index?: number) => boolean;
 
   private _changeEventSuppressionCount: number;
-  private _items: IObjectWithKey[];
-  private _selectedItems: IObjectWithKey[] | null;
+  private _items: TItem[];
+  private _selectedItems: TItem[] | null;
   private _selectedIndices: number[] | undefined;
   private _isAllSelected: boolean;
   private _exemptedIndices: { [index: string]: boolean };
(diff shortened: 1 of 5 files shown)


ecraig12345:
Marking as do not merge because I'd like to ensure Thomas gets to review this.

packages/utilities/src/selection/Selection.ts
```diff
@@ -36,8 +36,8 @@ export class Selection implements ISelection {
   private _unselectableCount: number;
   private _isModal: boolean;
 
-  constructor(options: ISelectionOptions = {}) {
-    const { onSelectionChanged, getKey, canSelectItem = (item: IObjectWithKey) => true, selectionMode = SelectionMode.multiple } = options;
+  constructor(options: ISelectionOptions<TItem> = {}) {
```
ThomasMichon: Will the compiler let us do:
```ts
constructor(options: TItem extends IObjectWithKey ? ISelectionOptions<TItem> : ISelectionOptionsWithOptionalGetKey<TItem> & ISelectionOptionsWithRequiredGetKey<TItem>)
```
This will ensure that if the supplied type doesn't have `key`, then the caller must supply the implementation.

ecraig12345: I'm not sure how to express that `options` is optional unless `TItem` doesn't extend `IObjectWithKey`.

ThomasMichon: Alternatively, to continue to enforce the case even when `options` is not passed, do:
```ts
constructor(
  // If the item type has a built-in key...
  ...args: TItem extends IObjectWithKey ?
    // Then the arguments can be empty or have the options without `getKey`
    ([ISelectionOptions<TItem>] | []) :
    // Otherwise, arguments require options with `getKey`.
    [ISelectionOptionsWithRequiredGetKey<TItem>]
) {
  const [options] = args;
  ...
}
```

ecraig12345: Updated the comments a bit more. Let me know what you think--it's fairly verbose in the interest of being clear to consumers who may not be as good with reading TS types.

PR #11548, opened 2019-12-24
OverflowSet: add menuitem role to child

Ravishankar9-dev:
#### Pull request checklist

#### Description of changes

Added role as 'menuitem' to the child div which already has role as 'menubar' in 'OverflowSet'

#### Focus areas to test

OverflowSet

#### Before Fix

image

#### After Fix

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11548)

### DIFF
--- a/change/office-ui-fabric-react-2019-12-24-12-39-09-user-v-satgar-addedRoleAsMenuItem.json
+++ b/change/office-ui-fabric-react-2019-12-24-12-39-09-user-v-satgar-addedRoleAsMenuItem.json
@@ -2,7 +2,7 @@
   "type": "patch",
   "comment": "Added role as menuitem to the child div",
   "packageName": "office-ui-fabric-react",
-  "email": "email not defined",
+  "email": "gantaravishankar@gmail.com",
   "commit": "8295ecd9f838cd45fa9432a21d0c24edfc116551",
   "date": "2019-12-24T07:09:09.120Z"
 }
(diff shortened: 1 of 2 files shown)


change/office-ui-fabric-react-2019-12-24-12-39-09-user-v-satgar-addedRoleAsMenuItem.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Added role as menuitem to the child div",
+  "packageName": "office-ui-fabric-react",
+  "email": "email not defined",
```
ecraig12345: Can you update this file to have your email? And set it in git config for the future? (run `git config --global user.email you@microsoft.com`)

micahgodbolt:
My main concern here is that command bar already uses role="menuitem" on the command bar buttons. I agree though that we should be consistent on if overflow set is 100% a menu abstraction (therefor role="menu/menubar" and items being "menuitem"). 

i.e. i'm fine with this fix (i already added the parent role)...but we need to make sure components using overflow set are then using it properly.

change/office-ui-fabric-react-2019-12-24-12-39-09-user-v-satgar-addedRoleAsMenuItem.json
Ravishankar9-dev: Added email address

micahgodbolt:
Ok...new concern here. Sometimes you need to be able to customize the role of the item, like if the button is actually toggleable, and therefore it should be a role="menuitemcheckbox". 

I'm not 100% sure the solution here. Should overflowSet specify role=menu/menubar? Or should that be up to the control using it.

And if it DOES speciffy that role, how do we support custom item role scenarios? check for a role on the item? leave it up to the user to provide item roles?

PR #11556, opened 2019-12-26
Icon: update cached icon if null/undefined icon value is stored in cache

vishakha116:
#### Description of changes

If a call to load the icon is made and the icon is not available, the icon is cached as undefined. For all further calls to load that icon, undefined is returned. Because of this if a call to load any icon is made before the icon is available, that icon always remains in the undefined state. 
With the changes in the PR, the cached icon can be updated if it is undefined based on a Boolean field(doNotCacheNullIcon). If the field, doNotCacheNullIcon, is set to true, the icon content will be re-fetched every time till the time the icon content is not null.

Undefined value for items is stored in cache if callback method returned undefined. Now, if the flag, doNotCacheIfCallbackResultIsNull, is passed true to the memorize method, the callback result will be recomputed every time till the time the callback result is not null.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11556)


packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
```diff
@@ -11,21 +11,56 @@ export const getIconContent = memoizeFunction((iconName?: string) => {
...
+  }
+
   return {
     children: code,
     iconClassName: subset.className,
     fontFamily: subset.fontFace && subset.fontFace.fontFamily
   };
 });
 
+export const getIconContentUnCached = memoizeFunction(
```
sachinma: Should you rather name it something like getIconContentWithValidIconCaching, since it does still cache the valid icons.

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
```diff
@@ -26,7 +26,21 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
     const { className, styles, iconName, imageErrorAs, theme } = this.props;
     const isPlaceholder = typeof iconName === 'string' && iconName.length === 0;
     const isImage = this.props.iconType === IconType.image || this.props.iconType === IconType.Image || !!this.props.imageProps;
-    const { iconClassName, children } = getIconContent(iconName);
+    let iconContent = undefined;
+
+    if (this.props.doNotCacheNullIcon) {
```
sachinma: You can use tertiary operator for single line assignment and declaration: let iconContent = this.props.don. ? <> : <>. 
You can also keep the variable assignment as: const { iconClassName, children } = iconContent || {}

packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
```diff
@@ -11,21 +11,56 @@ export const getIconContent = memoizeFunction((iconName?: string) => {
...
+      return null;
+    }
+    return {
+      children: code,
+      iconClassName: subset.className,
+      fontFamily: subset.fontFace && subset.fontFace.fontFamily
+    };
+  },
+  undefined,
+  true
```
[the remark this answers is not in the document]
vishakha116: Done

packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
```diff
@@ -11,21 +11,56 @@ export const getIconContent = memoizeFunction((iconName?: string) => {
...
 
+export const getIconContentUnCached = memoizeFunction(
+  (iconName?: string) => {
+    const { code, subset }: Pick<IIconRecord, 'code'> & { subset: Partial<IIconSubsetRecord> } = getIcon(iconName) || {
+      subset: {},
+      code: undefined
+    };
+
+    const shouldReturnNull: boolean = code ? false : true;
+    if (shouldReturnNull) {
```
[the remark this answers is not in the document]
vishakha116: Done

packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
```diff
@@ -11,21 +11,56 @@ export const getIconContent = memoizeFunction((iconName?: string) => {
...
  */
 export const FontIcon: React.FunctionComponent<IFontIconProps> = props => {
   const { iconName, className, style = {} } = props;
-  const { iconClassName, children, fontFamily } = getIconContent(iconName);
+  const iconContent = getIconContent(iconName);
+  let iconClassName = undefined;
+  let children = undefined;
+  let fontFamily = undefined;
+
+  if (iconContent) {
```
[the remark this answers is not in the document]
vishakha116: Reverted the changes from getIconContent  method

packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
vishakha116: Done

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
vishakha116: Done

packages/office-ui-fabric-react/src/components/Icon/Icon.types.ts
```diff
@@ -70,6 +70,12 @@ export interface IIconProps extends IBaseProps, React.HTMLAttributes<HTMLElement
    */
   styles?: IStyleFunctionOrObject<IIconStyleProps, IIconStyles>;
   theme?: ITheme;
+
+  /**
+   * Flag to decide whether to cache null icon or not. If the flag is set to true, the icon is recomputed every time till
+   * the icon is not null for the first time and then the non-null version gets cached.
+   */
```
micahgodbolt:
```suggestion
   * @defaultvalue false
   */
```

micahgodbolt:
Is there a scenario where we want to cache null icons? Should not chaching be the default (and only) behavior?

ecraig12345:
@vishakha116 Could you please update the title of the PR to be more descriptive?

packages/utilities/src/memoize.ts
```diff
@@ -80,9 +80,16 @@ export function memoize<T extends Function>(
  * @public
  * @param cb - The function to memoize.
  * @param maxCacheSize - Max results to cache. If the cache exceeds this value, it will reset on the next call.
+ * @param doNotCacheIfCallbackResultIsNull - Flag to decide whether to cache callback result if it is null.
```
ecraig12345: Couple issues with this parameter:

1. Should it skip caching for `null`, `undefined`, or both? It sounds like the specific case you care about involves `undefined`, not `null`. (Either way, the comment should be updated to specify which value(s) are handled specially.)
2. I'd prefer a more concise name like `ignoreUndefinedResult` or `ignoreNullResult` (as appropriate).

packages/utilities/src/memoize.ts
```diff
@@ -120,6 +127,10 @@ export function memoizeFunction<T extends (...args: any[]) => RET_TYPE, RET_TYPE
       cacheSize++;
     }
 
+    if (doNotCacheIfCallbackResultIsNull && !currentNode.value) {
```
ecraig12345: The `!currentNode.value` check is too loose. It will prevent caching of *any falsey value*: `null`, `undefined`, `false`, `0`, `NaN`, and `''` (empty string). Instead you need to check for *specifically* `!== null` and/or `!== undefined` (depending on what you decide the behavior should be).

You should also add tests for this new parameter, including that it does *not* prevent caching of other falsey values.

packages/office-ui-fabric-react/src/components/Icon/Icon.types.ts
```diff
@@ -70,6 +70,12 @@ export interface IIconProps extends IBaseProps, React.HTMLAttributes<HTMLElement
    */
   styles?: IStyleFunctionOrObject<IIconStyleProps, IIconStyles>;
   theme?: ITheme;
+
+  /**
+   * Flag to decide whether to cache null icon or not. If the flag is set to true, the icon is recomputed every time till
```
ecraig12345: I think not caching undefined results should be the default behavior (instead of adding a flag).

vishakha116: I have introduced a flag just to ensure backward compatibility, just in case any component using cached icons might want to keep it as the first value stored and don't want to update the value.

vishakha116: Please let me know if keeping backward compatibility is not required in this case as the existing behavior is not correct, then I will remove the flag.

micahgodbolt: We always want the 'exception' to be the boolean prop. So if you expect most people will want to `doNotCacheNullIcon`, then we should make `CacheNullIcons` the prop, so that others can opt back into that behavior.

On the other note, backwards compatibility for us means not breaking existing prop contracts. If we make a functional change that improves the control (like not caching null icons) we don't necessarily have to support the less efficient approach as well. If anything, I'd rather leave this prop out unless someone specifically asks for it to be added back in. You can always add props later. You can never take them away :)

ecraig12345: Agreed with Micah. Since the new behavior should generally be considered an improvement, we shouldn't add a flag, to avoid API bloat. Then if someone files an issue with a compelling case for caching null icons (which I doubt will happen), we can consider whether to add a flag at that point.

change/@uifabric-utilities-2019-12-02-18-10-48-dev-vishgup-memoizeissue.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "The items for which callback method returned null/undefined value, null/undefined was being stored in cache for these items. With the changes introduced, this undefined/null value can be updated (if available later) based on a flag.",
```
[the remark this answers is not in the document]
vishakha116: Done

change/office-ui-fabric-react-2019-12-02-18-10-48-dev-vishgup-memoizeissue.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "The items for which callback method returned null/undefined value, null/undefined was being stored in cache for these items. With the changes introduced, this undefined/null value can be updated (if available later) based on a flag.",
```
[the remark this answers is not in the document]
vishakha116: Done

packages/utilities/src/memoize.ts
vishakha116: changed the name to ignoreNullOrUndefinedResult as I want to handle for undefined and null

packages/utilities/src/memoize.ts
vishakha116: Modified the check  and I have also added a new test to check that it does not prevent caching of falsey values other that null and undefined.

packages/office-ui-fabric-react/src/components/Icon/Icon.types.ts
vishakha116: I have removed the prop from Icon

packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
```diff
@@ -11,12 +11,18 @@ export const getIconContent = memoizeFunction((iconName?: string) => {
     code: undefined
   };
 
+  if (!code) {
```
ecraig12345: Nitpicky thing, but the formatting on this file is off now (indentation of this function's parameters is non-standard, and there's an extra semicolon later), which is odd because Prettier should be running automatically on commit to standardize the formatting. This makes me suspect that your local environment isn't set up properly and/or you're skipping/disabling pre-commit hooks.

packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
```diff
@@ -25,7 +31,11 @@ export const getIconContent = memoizeFunction((iconName?: string) => {
  */
 export const FontIcon: React.FunctionComponent<IFontIconProps> = props => {
   const { iconName, className, style = {} } = props;
-  const { iconClassName, children, fontFamily } = getIconContent(iconName);
+  const { iconClassName, children, fontFamily } = getIconContent(iconName) || {
+    iconClassName: undefined,
```
ecraig12345: You shouldn't need to specify all of these, you can just do `getIconContent(iconName) || {}`. (If the compiler complains about that, we can add an interface for the getIconContent return type.)

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
```diff
@@ -26,7 +26,12 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
     const { className, styles, iconName, imageErrorAs, theme } = this.props;
     const isPlaceholder = typeof iconName === 'string' && iconName.length === 0;
     const isImage = this.props.iconType === IconType.image || this.props.iconType === IconType.Image || !!this.props.imageProps;
-    const { iconClassName, children } = getIconContent(iconName);
+    const iconContent = getIconContent(iconName);
+
+    const { iconClassName, children } = iconContent || {
```
ecraig12345: Similar to above, just do `const { iconClassName, children } = iconContent || {};`

packages/utilities/src/memoize.test.ts
```diff
@@ -98,6 +98,60 @@ describe('memoizeFunction', () => {
...
   });
+
+  it('caches and preserves falsey values other than null and undefined', () => {
+    let returnFalseyValue = true;
+
+    // tslint:disable-next-line:no-any
+    let falseyValue: any = null;
+
+    let callback = (): number | null => {
+      if (returnFalseyValue) { return falseyValue; }
```
ecraig12345: This formatting (and other places in this file) is also non-standard and should have been fixed by Prettier.

packages/utilities/src/memoize.test.ts
```diff
@@ -98,6 +98,60 @@ describe('memoizeFunction', () => {
...
+      return 1;
+    };
+    let func = memoizeFunction((a: string) => callback(), undefined, true/*ignoreNullOrUndefinedResult */);
+
+    // If falsey value returned by the callback method is null, update the cached value.
+    expect(func('nullFqalseyValue')).toEqual(falseyValue);
```
ecraig12345: Can you just remove the function parameter if it's not doing anything?

packages/utilities/src/memoize.test.ts
```diff
@@ -98,6 +98,60 @@ describe('memoizeFunction', () => {
...
+    returnFalseyValue = false;
+    expect(func('nullFalseyValue')).toEqual(1);
+
+    // If falsey value returned by the callback method is undefined, update the cached value.
+    returnFalseyValue = true;
```
ecraig12345: It would be preferable to split this into multiple tests.

packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
vishakha116: Added a new interface IIconContent as simply doing getIconContent(iconName) || {} was giving compiler error.

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
vishakha116: Made the change

packages/utilities/src/memoize.test.ts
vishakha116: There was some issue with the setup. Now Prettier is taking care of formatting.

packages/utilities/src/memoize.test.ts
vishakha116: Done

packages/utilities/src/memoize.test.ts
vishakha116: Done

packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
vishakha116: There was some issue with the setup. Now Prettier is taking care of formatting.

packages/office-ui-fabric-react/src/components/Icon/FontIcon.tsx
```diff
@@ -5,18 +5,32 @@ import { classNames, MS_ICON } from './Icon.styles';
...
+
+export const getIconContent = memoizeFunction(
+  (iconName?: string): IIconContent => {
+    const { code, subset }: Pick<IIconRecord, 'code'> & { subset: Partial<IIconSubsetRecord> } = getIcon(iconName) || {
+      subset: {},
+      code: undefined
+    };
+
+    if (!code) {
+      return {};
```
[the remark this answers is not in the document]
vishakha116: Returned null.

packages/utilities/src/memoize.test.ts
```diff
@@ -98,6 +98,101 @@ describe('memoizeFunction', () => {
     expect(func('a')).toEqual('a4');
     expect(func('a')).toEqual('a4');
   });
+
+  it('updates the cache if the cached value is null', () => {
+    let returnFalseyValue = true;
+    let falseyValue: null = null;
+    let callback = (): number | null => {
+      if (returnFalseyValue) {
+        return falseyValue;
```
ecraig12345: It would be clearer to just return null here instead of introducing an extra `falseyValue` variable.

packages/utilities/src/memoize.test.ts
```diff
@@ -98,6 +98,101 @@ describe('memoizeFunction', () => {
     expect(func('a')).toEqual('a4');
     expect(func('a')).toEqual('a4');
   });
+
+  it('updates the cache if the cached value is null', () => {
+    let returnFalseyValue = true;
```
ecraig12345: It would be clearer to call this `returnNull`.

packages/utilities/src/memoize.test.ts
```diff
@@ -98,6 +98,101 @@ describe('memoizeFunction', () => {
...
+      return 1;
+    };
+    let func = memoizeFunction(() => callback(), undefined, true /*ignoreNullOrUndefinedResult */);
+    expect(func()).toEqual(falseyValue);
+    returnFalseyValue = false;
+    expect(func()).toEqual(1);
+  });
+
+  it('updates the cache if the cached value is undefined', () => {
+    let returnFalseyValue = true;
```
ecraig12345: Similar comments to above: rename this to `returnUndefined` and get rid of the `falseyValue` variable. Same thing in the subsequent tests.

packages/utilities/src/memoize.test.ts
vishakha116: Done

packages/utilities/src/memoize.test.ts
vishakha116: Done

packages/utilities/src/memoize.test.ts
vishakha116: Done

PR #11560, opened 2019-12-26
Allow RTL to be passed into createTheme and picked up by Fabric and Layer

### LINKED ISSUE #11543: Dropdown isn't RTL compatible

https://codepen.io/jenian/pen/xxbdBXE?&editable=true#anon-signup

#### Actual behavior:
Dropdown control's caret doesn't change direction when dir is RTL.

#### Expected behavior:
In LTR - the caret should stick to the right (as it is now).
In RTL - the caret should stick to the left (currently it's to the right).

Requested priority: Normal

micahgodbolt:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #11543

#### Description of changes

Here's an example of how to use contextual RTL once this is released.

```tsx
       const rtlTheme = createTheme({
         rtl: true
       });

      // Customizer creates an RTL theme that is passed
      // to all child components style functions, 
      // even those rendered outside of the rtl div
      <Customizer settings={{ theme: rtlTheme }}>
        {/* 
            RTL is still required because Customizer does modify the DOM
            You can also use `Fabric` component instead of RTL div, 
            as Fabric will pick up RTL from theme 
        */}
        <div dir="rtl">
          <Dropdown
            placeholder="Select an option"
            label="Basic uncontrolled example"
            options={options}
            styles={dropdownStyles}
          />
        </div>
      </Customizer>

```

https://codepen.io/micahgodbolt/pen/NWPaGoB?editors=0010

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11560)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Fabric/__snapshots__/Fabric.test.tsx.snap
+++ b/packages/office-ui-fabric-react/src/components/Fabric/__snapshots__/Fabric.test.tsx.snap
@@ -505,6 +505,7 @@ exports[`Fabric renders a Fabric component in RTL 1`] = `
               -moz-osx-font-smoothing: grayscale;
               -webkit-font-smoothing: antialiased;
               color: #323130;
+              direction: rtl;
               font-family: 'Segoe UI', 'Segoe UI Web (West European)', 'Segoe UI', -apple-system, BlinkMacSystemFont, 'Roboto', 'Helvetica Neue', sans-serif;
               font-size: 14px;
               font-weight: 400;
--- a/packages/office-ui-fabric-react/src/components/Layer/Layer.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Layer/Layer.styles.ts
@@ -16,10 +16,12 @@ export const getStyles = (props: ILayerStyleProps): ILayerStyles => {
     root: [
       classNames.root,
       theme.fonts.medium,
+      {
+        direction: theme.rtl ? 'rtl' : undefined
+      },
       isNotHost && [
         classNames.rootNoHost,
         {
-          direction: theme.rtl ? 'rtl' : undefined,
           position: 'fixed',
           zIndex: ZIndexes.Layer,
           top: 0,
--- a/packages/office-ui-fabric-react/src/components/Layer/Layer.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Layer/Layer.styles.ts
@@ -16,9 +16,6 @@ export const getStyles = (props: ILayerStyleProps): ILayerStyles => {
     root: [
       classNames.root,
       theme.fonts.medium,
-      {
-        direction: theme.rtl ? 'rtl' : undefined
-      },
       isNotHost && [
         classNames.rootNoHost,
         {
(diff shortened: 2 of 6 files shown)


packages/office-ui-fabric-react/src/components/Layer/Layer.styles.ts
```diff
@@ -19,6 +19,7 @@ export const getStyles = (props: ILayerStyleProps): ILayerStyles => {
       isNotHost && [
         classNames.rootNoHost,
         {
+          direction: theme.rtl ? 'rtl' : undefined,
```
JasonGore: Are you sure this should only apply when `isNotHost` is `true`?

packages/office-ui-fabric-react/src/components/Fabric/__snapshots__/Fabric.test.tsx.snap
```diff
@@ -505,6 +505,7 @@ exports[`Fabric renders a Fabric component in RTL 1`] = `
               -moz-osx-font-smoothing: grayscale;
               -webkit-font-smoothing: antialiased;
               color: #323130;
+              direction: rtl;
```
JasonGore: I feel like I'm missing something but it seems `rtl` defaults to `undefined` in this PR and am not sure why it's showing up in this snapshot.

packages/office-ui-fabric-react/src/components/Layer/Layer.styles.ts
micahgodbolt: oh...totally didn't see that

packages/office-ui-fabric-react/src/components/Fabric/__snapshots__/Fabric.test.tsx.snap
micahgodbolt: this snapshot is for an RTL direction Fabric control

JasonGore: That's what I was missing. :P

packages/office-ui-fabric-react/src/components/Layer/Layer.styles.ts
micahgodbolt: oh funny. I don't actually need this because Layer renders content inside of a Fabric component.

JasonGore: What happens for consumers who don't use the `Fabric` component?

micahgodbolt: problem was:

```jsx
<Customizer settings={{theme: RTLTheme}}>
<div dir="rtl">
  <Dropdown .../>
</div>
</Customizer>
```

The above fixes all CSS calculations inside of Dropdown, but all of the layout flipping (flex direction etc) is caused by the `dir="rtl"`. And since the Dropdown options are rendered outside of that div, we need to make sure that Layers pull that RTL from the theme if present. 

And since Layer renders all its contents in a Fabric control, updating Fabric will be enough (i.e. layer style is redundant). 

https://github.com/OfficeDev/office-ui-fabric-react/blob/f6af0479a54f17fd16b9616a62d6be38686a5c1e/packages/office-ui-fabric-react/src/components/Layer/Layer.base.tsx#L56-L58

micahgodbolt:
I thought about that, but why Dropdown? All controls will work this way. Feels odd to have our example on a specific one.

Possible that we could have a theming page with demo of RTL in theming

micahgodbolt:
Ok...merging this in. Going to push docs off to another issue. Added example to the PR itself.

PR #11563, opened 2019-12-27
DetailsList: Update examples to work in strictFunctionTypes

### LINKED ISSUE #11546: Documentation: Failed to compile the example of DetailsList - Custom Item Rows
### Environment Information

- **Package version(s)**: 7.64.1

### Describe the issue:

Failed to compile the example code from [DetailsList - Custom Item Rows](https://developer.microsoft.com/en-us/fabric#/controls/web/detailslist/customitemrows)

#### Actual behavior:
Type check failed
```
Type '(props: IDetailsRowProps) => JSX.Element' is not assignable to type 'IRenderFunction<IDetailsRowProps>'.
  Types of parameters 'props' and 'props' are incompatible.
    Type 'IDetailsRowProps | undefined' is not assignable to type 'IDetailsRowProps'.
      Type 'undefined' is not assignable to type 'IDetailsRowProps'.ts(2322)
```

#### Other information
This issue was reported in https://github.com/OfficeDev/office-ui-fabric-react/issues/10012 and also relates to https://github.com/OfficeDev/office-ui-fabric-react/pull/10225. As @KevinTCoughlin suggested in https://github.com/OfficeDev/office-ui-fabric-react/issues/10012 I am able to solve the issue by setting `strictFunctionTypes: false`.

micahgodbolt:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #11546

#### Description of changes

Previous examples listed onRenderFunctions too loosely, stating that props would always be defined, when in fact the function only passes them optionally.

Without strictFunctionTypes typescript has no problem with this because the defined function is a [supertype of the original.](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html) 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11563)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.CustomFooter.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.CustomFooter.Example.tsx
@@ -6,9 +6,9 @@ import {
   IDetailsFooterProps,
   DetailsRow,
   SelectionMode,
-  IDetailsRowCheckProps,
   DetailsRowCheck
 } from 'office-ui-fabric-react/lib/DetailsList';
+import { IDetailsRowBaseProps } from '../DetailsRow';
 
 export interface IDetailsListCustomFooterExampleItem {
   key: number;
@@ -71,14 +71,17 @@ export class DetailsListCustomFooterExample extends React.Component<{}, {}> {
   }
 }
 
-function _renderDetailsFooterItemColumn(item: IDetailsListCustomFooterExampleItem, index: number, column: IColumn) {
-  return (
-    <div>
-      <b>{column.name}</b>
-    </div>
-  );
-}
+const _renderDetailsFooterItemColumn: IDetailsRowBaseProps['onRenderItemColumn'] = (item, index, column) => {
+  if (column) {
+    return (
+      <div>
+        <b>{column.name}</b>
+      </div>
+    );
+  }
+  return undefined;
+};
 
-function _onRenderCheckForFooterRow(props: IDetailsRowCheckProps): JSX.Element {
+const _onRenderCheckForFooterRow: IDetailsRowBaseProps['onRenderCheck'] = (props): JSX.Element => {
   return <DetailsRowCheck {...props} styles={{ root: { visibility: 'hidden' } }} selected={true} />;
-}
+};
--- a/packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.CustomFooter.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.CustomFooter.Example.tsx
@@ -6,9 +6,9 @@ import {
   IDetailsFooterProps,
   DetailsRow,
   SelectionMode,
-  DetailsRowCheck
+  DetailsRowCheck,
+  IDetailsRowBaseProps
 } from 'office-ui-fabric-react/lib/DetailsList';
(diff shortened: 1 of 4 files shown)


packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.CustomFooter.Example.tsx
```diff
@@ -71,14 +71,17 @@ export class DetailsListCustomFooterExample extends React.Component<{}, {}> {
...
 }
 
-function _renderDetailsFooterItemColumn(item: IDetailsListCustomFooterExampleItem, index: number, column: IColumn) {
-  return (
-    <div>
-      <b>{column.name}</b>
-    </div>
-  );
-}
+const _renderDetailsFooterItemColumn: IDetailsRowBaseProps['onRenderItemColumn'] = (item, index, column) => {
```
aneeshack4: Just to clarify, changing this to const is making this usage in the examples such that the props are required and not optional like before?

packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.CustomGroupHeaders.Example.tsx
```diff
@@ -68,32 +75,43 @@ export class DetailsListCustomGroupHeadersExample extends React.Component<{}, {}
...
-      <div className={classNames.headerAndFooter}>
-        <div className={classNames.headerTitle}>{`Custom header for ${props.group!.name}`}</div>
-        <div className={classNames.headerLinkSet}>
-          <Link className={classNames.headerLink} onClick={this._onToggleSelectGroup(props)}>
-            {props.isSelected ? 'Remove selection' : 'Select group'}
-          </Link>
-          <Link className={classNames.headerLink} onClick={this._onToggleCollapse(props)}>
-            {props.group!.isCollapsed ? 'Expand group' : 'Collapse group'}
-          </Link>
+  private _onRenderGroupHeader: IDetailsGroupRenderProps['onRenderHeader'] = props => {
```
aneeshack4: Is this the new usage of the strictFunctionTypes? The `onRenderHeader`

packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.CustomFooter.Example.tsx
micahgodbolt: the const is just to be consistent in our examples of using const + lambdas for internal functions. Also makes it easier to show the typing.

But yes, the big change is that I refer to the exact signature for onRenderItemColumn vs simply trying to recreate it

packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.CustomGroupHeaders.Example.tsx
micahgodbolt: yes, instead of defining the function type via  `(props: IGroupDividerProps): JSX.Element`, I'm refering directly to the `onRenderHeader` from `IDetailsGroupRenderProps`. This ensures that the types are identical, therefor not causing issues when using strictFunctionTypes.

It also exposed a bug in that `props` are actually optional, and null should be returned if props don't exist.

PR #11571, opened 2019-12-27
Layer: add touch events to filtering to stop those that are bubbling up.

### LINKED ISSUE #11264: React DND Drag and Drop Touch Events not working inside Dialog and Modal components
### Environment Information

- **Package version(s)**: (6.189.0)
- **Browser and OS versions**: (Touch supported devices)

https://github.com/jyothsnag29/ReactDnDMultiBackEndDialog

#### Use case about the mentioned repository
image
If we drag the red/green/blue color box and drop it to the gray box we get an alert about which color has dropped. Open Pop up link opens the Dialog component having same drag and drop logic. Open
Model link open the Modal component having same drag and drop logic.

#### Actual behavior:
I was using [React dnd-multi-backend](https://github.com/LouisBrunner/dnd-multi-backend/tree/master/packages/react-dnd-multi-backend/examples) to make it work for both touch and mouse events. But unfortunately Drag and Drop Touch events are not working inside Dialog or Modal window even after setting the eventBubblingEnabled to true. However It works in normal screen.
Mouse events work in all scenarios.

#### Expected behavior:

When we try to do the color drag and drop from normal screen or inside the Dialog or inside Modal window it should work for both Mouse events and Touch events

aneeshack4:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11264 

#### Description of changes

Drag and Drop stopped working inside Dialog and Modal components because Layer which both of these are built on doesn't catch touch events that are bubbling up and stop them.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11571)

### DIFF
--- a/change/office-ui-fabric-react-2019-12-27-13-39-03-reactdnd.json
+++ b/change/office-ui-fabric-react-2019-12-27-13-39-03-reactdnd.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Layer: add touch events to filtering to stop those that are bubbling up",
+  "packageName": "office-ui-fabric-react",
+  "email": "email not defined",
+  "commit": "992537473a22e8d7f7ef217e17ab4da31ceef3db",
+  "date": "2019-12-27T21:39:03.474Z"
+}
--- a/change/office-ui-fabric-react-2019-12-27-13-39-03-reactdnd.json
+++ b/change/office-ui-fabric-react-2019-12-27-13-39-03-reactdnd.json
@@ -2,7 +2,7 @@
   "type": "patch",
   "comment": "Layer: add touch events to filtering to stop those that are bubbling up",
   "packageName": "office-ui-fabric-react",
-  "email": "email not defined",
+  "email": "aneeshak@microsoft.com",
   "commit": "992537473a22e8d7f7ef217e17ab4da31ceef3db",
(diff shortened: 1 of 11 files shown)


aneeshack4:
Tested this change with `yarn codepen` - I added the fabric bundles that created to the codesandbox the issue author provided. [Forked that](https://codesandbox.io/s/reactdndmultibackendcolordrop-6eel1) as a result and made sure that drag and drop works on this new codesandbox: 
image

change/office-ui-fabric-react-2019-12-27-13-39-03-reactdnd.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Layer: add touch events to filtering to stop those that are bubbling up",
+  "packageName": "office-ui-fabric-react",
+  "email": "email not defined",
```
JasonGore: A bit odd, is email set up in your git config?

aneeshack4: I thought I correctly set it up I'll redo git config.

micahgodbolt:
Considering the two approvals, i think this is good to go.

PR #11573, opened 2019-12-27
Add custom Icon renderers to Calendar component

### LINKED ISSUE #11572: Unable to set custom Icons on Calendar component
### Environment Information

- **Package version(s)**: Current

#### Actual behavior:

If the default icon library is not loaded, this results into not having any icon rendered on the Calendar component:

Screen Shot 2019-12-27 at 3 24 57 PM

We generally avoid this issue by using an icon custom render function, but in the case of the Calendar component there is none.

#### Expected behavior:

Being able to set a render function for the icon.

Requested priority: High

Products/sites affected: Teams for Edu

matidomingues:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11572

#### Description of changes

- Added an Icon renderer prop for the Calendar component while maintaining the old Icon name string property due for backwards compatibility.
- Fixed a small bug where Icons where not being rotated for RTL in CalendarDay component

#### Focus areas to test

Calendar component

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11573)

### DIFF
--- a/change/office-ui-fabric-react-2019-12-27-15-14-05-master.json
+++ b/change/office-ui-fabric-react-2019-12-27-15-14-05-master.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Add custom renderers to Calendar Icons",
+  "packageName": "office-ui-fabric-react",
+  "email": "madomi@microsoft.com",
+  "commit": "dfb2dc019f02b1108fda68d867484495c1df6d45",
+  "date": "2019-12-27T23:14:05.203Z"
+}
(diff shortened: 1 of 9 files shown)


micahgodbolt:
@lorejoh12 what do you think of this change? It seems like a large change to support custom icons. 

@matidomingues can you accomplish this using `registerIcons` instead?

https://codepen.io/micahgodbolt/pen/PdeKEy?editors=0010

matidomingues:
@micahgodbolt registerIcons worked perfectly, I can delete this PR or merge it in case you like it, up to you :)

micahgodbolt:
I think we're probably good without the addition seeing as others could easily do the same thing. Appreciate the feedback!

PR #11574, opened 2019-12-28
CI: Break up Build & Test, Add Parallelization

JasonGore:
#### Pull request checklist

#### Description of changes

This PR builds off the work started in #11422 to speed up local and CI build times by breaking up the existing `build` task into `build`, `test`, `lint` and `bundle` and then using the new tasks in parallel in a new parallelization pipeline.

#### Command Changes

| Monorepo Build Command | Prereqs | Important Notes                                                                                      | Description                                                                                                                        |
| ---------------------- | ------- | ---------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
| build                  |         | Now TypeScript only. No longer runs bundle, lint or test. No longer production build by default.     | Generates output using TypeScript.                                                                                                 |
| build:prod             |         | New.                                                                                                 | Production version of build command.                                                                                               |
| buildci                |         | No change. This command approximates PR build steps and is useful for validating your build locally. | Reproduction of PR build steps: build, lint, test, bundle.                                                                         |
| buildci:prod           |         | New                                                                                                  | Production version of buildci command.                                                                                             |
| buildto                |         | Now TypeScript only. No longer runs bundle, lint or test. No longer production build by default.     | Builds to specified target package.                                                                                                |
| buildfast              |         | Removed (redundant with 'build' and 'buildto')                                                       |                                                                                                                                    |
| bundle                 | build   | New                                                                                                  | Runs webpack. Webpack configs use lib output as entry point, requiring build as prereq.                                            |
| bundle:prod            | build   | New                                                                                                  | Production version of bundle command.                                                                                              |
| clean                  |         | New                                                                                                  | Cleans package output.                                                                                                             |
| lint                   | build   | New                                                                                                  | Runs lint and lint-imports tasks. lint-imports requires build artifacts to detect that physical file exist for imports from 'lib'. |
| test                   | build   | New (replaces 'validate')                                                                            | Runs jest. Build artifacts required for dependencies of packages under test.                                                       |
| validate               |         | Removed (replaced by 'test')                                                                         |                                                                                                                                    |

#### Pipeline Strategy

Based on build times and prerequisites, this is the parallelization and job strategy defined by the pipeline:

| Job      | Prereq Job | Runs                               | Description                                                                     |
| -------- | ---------- | ---------------------------------- | ------------------------------------------------------------------------------- |
| Build    |            | TypeScript compliation only        |
| Validate | Build      | Test (jest) and lint               | Requires build artifacts for internal package deps.                             |
| Deploy   | Build      | Bundle (webpack)                   | Requires build artifacts since webpack configs are set to use lib entry points. |
| PerfTest | Build      | Perf test script                   | Requires build artifacts for bundling and running perf-test script.             |
| VRTest   | Build      | Visual Regression tests (screener) | Requires build artifacts for internal package deps.                             |

#### TODO (before merge):
* ~~Finalize command structure.~~
* ~~Finish testing (command output, release pipeline.)~~
* ~~Resolve TODOs in PR code.~~
* ~~Communicate build command changes via email.~~
* ~~Document command changes in wiki.~~
* ~~Coordinate timing of merge to deal with any possible release pipeline and publishing issues.~~
* ~~Try `lerna --parallel` for all post-build jobs to increase performance.~~
* ~~Fix `PerfTest` (currently needs `Deploy` as prereq which mostly negates build time savings)~~
* ~~Rename `azure-pipelines.parallel.yml` pipeline to `azure-pipelines.yml`.~~

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11574)


package.json
```diff
@@ -12,15 +12,16 @@
...
     "builddemo": "node ./scripts/monorepo/buildTo.js fabric-website-resources --min",
+    "bundle": "lerna run bundle --stream --",
     "scrub": "node ./scripts/scrub.js",
     "vrtest": "cd apps && cd vr-tests && npm run screener",
     "a11ytest": "cd apps && cd a11y-tests && npm run test",
     "start": "yarn workspace @uifabric/fabric-website-resources start",
     "start-exp": "yarn workspace @uifabric/experiments start",
-    "build": "lerna run build --stream -- --production --lint",
-    "buildci": "lerna run build --stream -- --lint",
+    "build": "lerna run build --stream --",
```
jdhuntington: Is `--` necessary here?

azure-pipelines.parallel.yml
```diff
@@ -0,0 +1,155 @@
...
+
+    # TODO: uncomment, temporarily disabled to speed debugging
+    - template: azure-pipelines.tools.yml
+
+    - script: |
+        yarn
+      displayName: yarn
+
+    - script: |
+        npm run checkchange
```
jdhuntington: nit: `yarn run`

azure-pipelines.tools.yml
```diff
@@ -0,0 +1,12 @@
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '10.x'
+  displayName: 'Install Node.js'
+
+- task: geeklearningio.gl-vsts-tasks-yarn.yarn-installer-task.YarnInstaller@3
```
jdhuntington: Do we need this dependency? Can we just directly install it with a shell command instead?

package.json
JasonGore: This is in here for now because I'm not sure how else to pass the `--production` flag to `build`. It implicitly worked with `buildci` before because the arg comes after `--` there as well.

azure-pipelines.parallel.yml
JasonGore: This is copied code from the existing pipeline.

azure-pipelines.tools.yml
JasonGore: Ken advised me on this approach. I'm open to other methods, as well.

azure-pipelines.parallel.yml
JasonGore: (I'm not sure why `yarn` wasn't already used for much of this... was it just because of the newer `yarn` dep?)

azure-pipelines.tools.yml
kenotron: @jdhuntington - we could install it through shell, I've not tested how fast that is. This is just what we've been using for the other tasks, seems to "install" yarn pretty quickly and made available nicely without having to worry about installing something globally on a CI box. That's the only difference to me...

azure-pipelines.parallel.yml
kenotron: actually do:

```suggestion
        yarn checkchange
```

azure-pipelines.parallel.yml
```diff
@@ -0,0 +1,155 @@
...
+      displayName: check change
+
+    - script: |
+        npm run build
```
kenotron:
```suggestion
        yarn build
```

JasonGore: Do you know why it was this way on master? Was it purely because of the yarn dep?

scripts/package.json
```diff
@@ -11,7 +11,7 @@
     "bundlesize": "bundlesize --debug",
     "clean": "",
     "code-style": "node ./just-scripts.js code-style",
-    "test": "jest"
+    "test": ""
```
kenotron: You should delete build, test, and clean here.

apps/pr-deploy-site/just.config.js
```diff
@@ -8,6 +8,9 @@ const findGitRoot = require('@uifabric/build/monorepo/findGitRoot');
 const gitRoot = findGitRoot();
 let instructions = [];
 
+// TODO: this doesn't appear to be working compared to master. why?
+// TODO: this works when 'yarn build' is called in 'apps/pr-deploy-site', but not when called from repo root.
+// TODO: is this meant to run purely as a side effect? when should it run? what triggers it? seems unreliable.
```
JasonGore: @kenotron I'm looking into this now. I'm not sure why it's not working from root. Just wanted to tag you in case you see anything obvious. Thanks!

JasonGore: I think I have to disable use of `.cached()` for all commands other than ` build`. The inputs to the other commands (`lint`, `test`, `bundle`) include library and dist output, and `.cached()` won't be reliable for detecting changes to those files.

apps/perf-test/tasks/perf-test.js
```diff
@@ -188,7 +189,11 @@ const urlForDeployPath = process.env.BUILD_SOURCEBRANCH
   ? `http://fabricweb.z5.web.core.windows.net/pr-deploy-site/${process.env.BUILD_SOURCEBRANCH}/perf-test`
   : 'file://' + path.resolve(__dirname, '../dist/');
 
-const urlForDeploy = urlForDeployPath + '/index.html';
+// Temporarily comment out deploy site usage to speed up CI build time and support parallelization.
+// At some point perf test should be broken out from CI default pipeline entirely and then can go back to using deploy site.
+// For now, use local perf-test bundle so that perf-test job can run ASAP instead of waiting for the perf-test bundle to be deployed.
+// const urlForDeploy = urlForDeployPath + '/index.html';
+const urlForDeploy = 'file://' + path.resolve(__dirname, '../dist/') + '/index.html';
```
ecraig12345: Does this need to be `file:///`? And does it work on both Mac and Windows?

apps/pr-deploy-site/just.config.js
```diff
@@ -32,7 +35,7 @@ repoDeps.forEach(dep => {
 preset();
 
 task(
-  'build',
+  'bundle',
```
ecraig12345: Maybe a silly question but why is this `bundle` now?

apps/ssr-tests/just.config.js
```diff
@@ -10,4 +10,7 @@ task('mocha', () => {
   return spawn(process.execPath, [mochaCmd, 'dist/ssr-tests.js'], { stdio: 'inherit' });
 });
 
-task('build', series('webpack', 'mocha')).cached();
+// TODO: add typescript task? shouldn't have to bundle for typechecking
```
ecraig12345: Does webpack even do type-checking in this project's configuration? (IIRC we typically only do type-checking in webpack serve config)

azure-pipelines.tools.yml
```diff
@@ -0,0 +1,12 @@
+steps:
```
ecraig12345: Can you add a comment about what this pipeline does?

packages/tsx-editor/just.config.js
```diff
@@ -11,15 +11,4 @@ task('copy-types', copyTypes);
...
-    'clean',
-    'copy',
-    'copy-types',
-    parallel(
-      condition('validate', () => !argv().min),
-      series('ts', parallel(condition('webpack', () => !argv().min), condition('lint-imports', () => !argv().min)))
-    )
-  )
-).cached();
+task('build', series('clean', 'copy', 'copy-types', 'ts')).cached();
```
ecraig12345: Nice side effect...making it easier to re-define `build`

change/@uifabric-api-docs-2019-12-27-17-16-21-jg-build-and-test.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
ecraig12345: Shouldn't most/all of these be `none` since there's no impact to the published packages?

apps/ssr-tests/just.config.js
JasonGore: I'll check. I just assumed typechecking is something we'd want in this package, unless there's a reason not to that I don't know of. (Is ssr-test 100% auto-generated? Perhaps that's why.)

apps/pr-deploy-site/just.config.js
JasonGore: Is there a `yarn start` that will fail anywhere if this command is not executed? I interpreted this command as a 'bundle' step of sorts before `pr-deploy-site` is deployed.

JasonGore: There also isn't any TS or typechecking that I see here, so I'm not sure this is really a build command in any case.

apps/perf-test/tasks/perf-test.js
JasonGore: Per my comments, I need to do this in order to prevent PerfTest from being a super long pole. If I don't it also needs to wait for the `Deploy` job to complete, which pushes total build time past 30 minutes (same as master), negating all of the perf gains. This is the same URL that has been used when running perf-test locally all along for both Mac and Windows (see line 189.)

azure-pipelines.tools.yml
JasonGore: The pipeline strategy is in the PR description, but I can add something to this file as well.

apps/ssr-tests/just.config.js
JasonGore: Yeah, `ssr-tests` has no TypeScript. I'll remove the comment.

apps/pr-deploy-site/just.config.js
JasonGore: After further investigation, this command actually needs to copy the `bundle` output of other packages, and additionally, I needed to make `bundle` run in `--stream` rather than `--parallel` so as not to break this task. For these reasons it can't run in the `build` phase.

package.json
JasonGore: I've removed stray instances of trailing `--` and added a few `:prod` convenience commands. Funny enough I found in testing that `yarn build -- --production` does not work, but `yarn build -- -- --production` DOES. I've decided to add the `:prod` as convenience commands, which are only slightly less awkward.

kenotron:
Please ignore that draft build... this build is green!

apps/perf-test/tasks/perf-test.js
kenotron: resolve will add that `/` for you

micahgodbolt:
image

PR #11581, opened 2019-12-30
ChoiceGroupOption: pass original renderer to on render label

### LINKED ISSUE #11575: Pass the original renderer to onRenderLabel in ChoiceGroupOption

Currently, the ChoiceGroup component allows you to pass "onRenderLabel" property that is a function used to render the labels instead of the fabric function. This function currently is provided with the original properties but not the original renderer.
If we want to keep the original design but wrap it with another functionality/design - we will have to generate it on our own and even give it an internal class name (ms-ChoiceFieldLabel).

The expected behavior is a 2nd parameter sent to onRenderLabel that is the original renderer (_onRenderLabel) so you will be able to call it and wrap the original rendered label.

ChoiceGroup / ChoiceGroupOption

No, but this functionality is present in many onRender* functions in fabric. The best example for this is "**onRenderField**" this specific component (ChoiceGroupOption)

FYI @maelgami

dagoltz:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11575 

#### Description of changes

Pass the defaultRender as a second parameter to onRenderLabel. Also added tests to cover the existing and new usages of onRenderLabel.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11581)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.test.tsx
@@ -1,4 +1,4 @@
-/* tslint:disable:no-unused-variable */
+/* tslint:disable:no-unused-variable jsx-no-lambda */
 import * as React from 'react';
 /* tslint:enable:no-unused-variable */
 import * as renderer from 'react-test-renderer';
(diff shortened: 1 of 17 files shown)


dagoltz:
FYI @jdhuntington

jdhuntington:
Awesome; thank you!

packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.test.tsx
```diff
@@ -18,6 +19,33 @@ describe('ChoiceGroupOption', () => {
         <ChoiceGroupOption iconProps={{ iconName: 'CalendarWeek' }} key="G" text="Option G" focused />
         <ChoiceGroupOption iconProps={{ iconName: 'CalendarDay' }} key="H" text="Option H" checked />
         <ChoiceGroupOption iconProps={{ iconName: 'CalendarWeek' }} key="I" text="Option I" disabled />
+        <ChoiceGroupOption
```
ecraig12345: Might be better to put this in its own test, like "custom renders correctly" (and then update snapshots again)

packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.test.tsx
```diff
@@ -18,6 +19,33 @@ describe('ChoiceGroupOption', () => {
...
+          key="J"
+          text="Option J"
+          onRenderLabel={(option: IChoiceGroupOptionProps): JSX.Element => {
```
ecraig12345: To get the build to pass you'll need to add a comment line `// tslint:disable:jsx-no-lambda` (either around here or near the top of the file). That lint rule generally doesn't matter for tests, so disabling it is fine.

dagoltz:
Thanks, @ecraig12345, I have resolved all issues - waiting for the build.
Who has the right to merge this PR once it is re-approved (if needed)?

PR #11582, opened 2019-12-31
ComboBox: support non latin characters

### LINKED ISSUE #5834: Combobox doesn't autocomplete with non-english characters
### Bug Report

- __Package version(s)__: 6.39.2
- __Browser and OS versions__: latest Chrome, Windows Server 2016

#### Priorities and help requested (not applicable if asking question):

Requested priority: (Blocking) 

#### Describe the issue:

I have Combobox with options containing non-english (russian) words. 
`
<ComboBox
    allowFreeform={false}
    autoComplete="on"
    options={this.props.store!.budgetArticles.map(x => ({ key: x.Id, text: x.Title }))}
/>
`
By typing I can choose only options with english text. Russian text doesn't appears and options not selected.
#### Actual behavior:

Can't type russian characters and therefore select options with russian text

#### Expected behavior:

Can use any language for selection in combobox.

micahgodbolt:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #5834

#### Description of changes

`String.fromCharCode(ev.which)` returns the latin character mapped to the given keycode, which means even with a Russian keyboard, you are always firing latin characters.

image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11582)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.test.tsx
@@ -241,7 +241,7 @@ describe('ComboBox', () => {
   it('Can insert non latin text in uncontrolled case with autoComplete on and allowFreeform off', () => {
     wrapper = mount(<ComboBox defaultSelectedKey="0" options={RUSSIAN_OPTIONS} autoComplete="on" allowFreeform={false} />);
 
-    wrapper.find('input').simulate('keyup', { key: 'п' });
+    wrapper.find('input').simulate('input', { target: { value: 'п' } });
     wrapper.update();
     expect(wrapper.find('input').props().value).toEqual('папа');
   });
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.test.tsx
@@ -20,6 +20,13 @@ const DEFAULT_OPTIONS3: IComboBoxOption[] = [
   { key: '3', text: 'Bar' }
 ];
 
+const RUSSIAN_OPTIONS: IComboBoxOption[] = [
+  { key: '0', text: 'сестра' },
+  { key: '1', text: 'брат' },
+  { key: '2', text: 'мама' },
+  { key: '3', text: 'папа' }
+];
+
 const returnUndefined = () => undefined;
 
 type InputElementWrapper = ReactWrapper<React.InputHTMLAttributes<any>, any>;
@@ -229,6 +236,14 @@ describe('ComboBox', () => {
     wrapper.find('input').simulate('input', { target: { value: 'f' } });
     wrapper.update();
     expect(wrapper.find('input').props().value).toEqual('Foo');
+  });
+
+  it('Can insert non latin text in uncontrolled case with autoComplete on and allowFreeform off', () => {
+    wrapper = mount(<ComboBox defaultSelectedKey="0" options={RUSSIAN_OPTIONS} autoComplete="on" allowFreeform={false} />);
+
+    wrapper.find('input').simulate('keyup', { key: 'п' });
+    wrapper.update();
+    expect(wrapper.find('input').props().value).toEqual('папа');
   });
 
   it('Can insert text in uncontrolled case with autoComplete off and allowFreeform on', () => {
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -1849,14 +1849,6 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
           return;
         }
 
-        // If we are not allowing freeform and
-        // allowing autoComplete, handle the input here
-        // since we have marked the input as readonly
-        if (!allowFreeform && autoComplete === 'on') {
-          this._onInputChange(String.fromCharCode(ev.which));
```
jspurlin: >ev.which [](start = 50, length = 8)

why couldn't this be updated to `ev.key`?

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -1896,6 +1888,15 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
       return;
     }
 
+    // If we are not allowing freeform and
+    // allowing autoComplete, handle the input here
+    // since we have marked the input as readonly
+
+    if (!allowFreeform && autoComplete === 'on') {
+      this._onInputChange(ev.key);
```
joschect: I'm not sure that ev.key can just be used like this. It looks like ev.key will map to things like `Enter`, `Escape` and other "named" keys like `F1`

See the console in this [codepen](https://codepen.io/slimemold/pen/zYxEPdX?editors=1111) for a demonstration, enter is the most obvious one. Also more info here on [mdn](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key)

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.test.tsx
```diff
@@ -231,6 +238,14 @@ describe('ComboBox', () => {
     expect(wrapper.find('input').props().value).toEqual('Foo');
   });
 
+  it('Can insert non latin text in uncontrolled case with autoComplete on and allowFreeform off', () => {
+    wrapper = mount(<ComboBox defaultSelectedKey="0" options={RUSSIAN_OPTIONS} autoComplete="on" allowFreeform={false} />);
+
+    wrapper.find('input').simulate('keyup', { key: 'п' });
+    wrapper.update();
+    expect(wrapper.find('input').props().value).toEqual('папа');
```
jdhuntington: 👍

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
micahgodbolt: Here's the challenge. Take a look at these results using a Russian alphabet keyboard

image

micahgodbolt: Also, the only valid keys when `!allowFreeform && autoComplete` are characters in the list items. Enter still enters, escape still escapes.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
micahgodbolt: hmmm...for some reason i thought ev.key wasn't available on mouse down

micahgodbolt: ok...reduced the change of this PR to simply using ev.key instead.

micahgodbolt:
@jspurlin what do you feel about this change? Are there scenarios where you think this would not work?

jspurlin:
@micahgodbolt, @joschect suggested that `event.key` needs to be investigated more before that's the solution we land on. Depending on what comes out from that investigation, assuming things look good, I'm good to sign off

micahgodbolt:
@jspurlin is there any way to try this change out in the ribbon? want to make sure that "needs to be investigated" does not means "let the PR sit forever".

jspurlin:
Sure, there's a way to try it out in the ribbon, but I think the concern was more from a technical standpoint if we could/should trust what `event.key` gives us in all cases (for example if a modifier key is pressed at the same time as a normal key comes to mind as a potentially interesting case). @joschect may have more concerns... If not, I'm good to sign off

micahgodbolt:
This is only used in the autocomplete scenario, so we only add characters that match the string in the options. If user presses a random key, it won't match any option value, so won't have any effect.

PR #11584, opened 2019-12-31
Combobox: Clicks to headers and scrollbars no longer cause onBlur

### LINKED ISSUE #6392: ComboBox invokes onBlur handler unexpectedly when the user clicks the scroll bar of the dropdown menu
### Bug Report 

For a ComboBox that is configured with "allowFreeForm" being true, if the user clicks on the scroll bar of the dropdown menu in order to scroll to an option to select, it invokes onBlur handler unexpectedly (technically the user's focus is still on the ComboBox as a whole).  In comparison, Fabric Dropdown component doesn't have this issue.
 
 
- __Package version(s)__: 6.61.2 
- __Browser and OS versions__: Chrome on Windows 10/2018 Server
 
Requested priority: High
 
Products/sites affected: SPO
 
#### Describe the issue:
 
Add a ComboBox that is configured with "allowFreeForm" to be true and large number of options so that a scroll bar is shown in the dropdown menu.  The user clicks on the arrow to open the dropdown menu.  Then the user clicks on the scroll bar in the dropdown and try to drag it down or up.

 
#### Actual behavior:
 
Actual: onBlur handler is invoked unexpectedly.
 
#### Expected behavior:
 
Expected: onBlur handler is not invoked because technically the interaction is still on the ComboBox. (Drowdown control doesn't invoke onBlur in this situation)

micahgodbolt:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #6392

#### Description of changes

Clicks outside of the input were causing the input to lose focus and fire 'onblur'. With this change clicks to headers, dividers and scrollbars will have no effect, but clicks to menu items work properly.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11584)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
@@ -1386,6 +1386,9 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
       : this._getFirstSelectedIndex();
   }
 
+  /**
+   * Mouse clicks to headers, dividers and scrollbar should not make input lose focus
+   */
   private _onCalloutMouseDown: ICalloutProps['onMouseDown'] = ev => {
     ev.preventDefault();
   };
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
@@ -27,6 +27,7 @@ import { KeytipData } from '../../KeytipData';
 import { Label } from '../../Label';
 import { SelectableOptionMenuItemType, getAllSelectedOptions } from '../../utilities/selectableOption/index';
 import { BaseButton, Button } from '../Button/index';
+import { ICalloutProps } from '../Callout';
 
 export interface IComboBoxState {
   /** The open state */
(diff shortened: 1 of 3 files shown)


micahgodbolt:
really? 
image

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -1173,6 +1174,7 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
         className={css(this._classNames.callout, calloutProps ? calloutProps.className : undefined)}
         target={this._comboBoxWrapper.current}
         onDismiss={this._onDismiss}
+        onMouseDown={this._onCalloutMouseDown}
```
aneeshack4: What part of this change is getting rid of the onBlur invoking/removing the effect of clicks to headers, dividers and scrollbars?

micahgodbolt: by adding `ev.preventDefault();` to the onMouseDown that has the effect to literally nullify that click. It never happened. So focus is never removed from the input. 

With the Dropdown, we control focus via state, so we have control over when focus is lost. Combobox uses natural HTML input focus, so we have to stop the click at the point of clicking, not later on in the process.

khmakoto:
> really?
> image

It's interesting how it is saying that `lib_initializeIcons` is a new function. I wonder what part of the code you've added could cause this.

Issue #11291, opened 2019-11-22
Tile: Text illegible on dark theme


phkuo:
Broken code:
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/experiments/src/components/Tile/Tile.scss

The text on a tile in a dark theme is illegible:
![image](https://user-images.githubusercontent.com/20691641/69466193-32cf5680-0d38-11ea-9dcd-ef46f7d776d6.png)

The issue is that the text color uses `$ms-color-white`, while the background gradient uses a hardcoded black color (line 424). Themed colors and hardcoded colors should not be mixed. In this case, either we should find a way to make the gradient themable (which sounds hard), or also hardcode #fff as the font color (which sounds like the sane path forward).

aneeshack4:
@phkuo Where is setting the text color? I see `ms-color-white` for `name` on line 476 - is that what you're referring to? Hardcoding the #fff as the font color sounds reasonable to me as well. Do you think some design input here could be useful?

phkuo:
Not sure which line it is exactly. Design intent here is clear, i vote we just hardcode it.

Sent by phone

________________________________

aneeshack4:
@phkuo I'm following up online.

aneeshack4:
It seems either the `$ms-color-white` for [`name`](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/experiments/src/components/Tile/Tile.scss#L477) or inside [this `hasForegroundFrame`](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/experiments/src/components/Tile/Tile.scss#L148) are the most appropriate places for this change - I tried changing both of these separaetly,  Tile text looks just as visible in dark mode as in regular:
![image](https://user-images.githubusercontent.com/4161791/69768704-765e0200-1136-11ea-8675-a1d2f7f1ef36.png)
![image](https://user-images.githubusercontent.com/4161791/69768709-7958f280-1136-11ea-8244-1f9041b48836.png)
Tagging component owner @ThomasMichon - Thomas, do you have any additional insights into this?

phkuo:
@aneeshack4 it's specifically for tiles on top of picture thumbnails, sorry if I didn't make that clear! In those cases, it adds a gradient at the bottom, and the text is on top of that gradient. The gradient is hardcoded to white, but the text is themed.

aneeshack4:
will need some input from @ThomasMichon who is the component owner

[...]

PR #11587, opened 2020-01-02
Tile: Fix text over image color

Pratik-Chhajer:
#### Pull request checklist

#### Description of changes

The text appearing over image use's $ms-color-white to set it's color styling. In case of dark mode, it gets translated to some shade of black which is not desired behavior. Below screenshots are from OneDrive for Consumers web.

Before:
MicrosoftTeams-image (1)

After:
MicrosoftTeams-image (2)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11587)

### DIFF
--- a/change/@uifabric-experiments-2020-01-02-09-47-28-pratik-chhajer-text-over-image-color-fix.json
+++ b/change/@uifabric-experiments-2020-01-02-09-47-28-pratik-chhajer-text-over-image-color-fix.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "fix color of text over image in dark mode",
+  "comment": "Tile: fix color of text over image in dark mode",
   "packageName": "@uifabric/experiments",
   "email": "prchhaje@microsoft.com",
   "commit": "964a018662ecadc2db4f6ee9cb049bc9543b1ec3",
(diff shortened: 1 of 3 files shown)


aneeshack4:
@Pratik-Chhajer Is this a fix for #11291? It looks like it - thanks for making it! I was investigating it and talking with @ThomasMichon a while ago about where exactly in Tile.scss to make this change as that was unclear but I think this looks good. Though clarification question - why Signal.scss is being changed too?

aneeshack4:
Adding @phkuo in case you're not looped in on this already - I believe this will fix your problem in sharepoint?

Pratik-Chhajer:
@aneeshack4 I recently implemented dark mode for OneDrive PWA, I was facing issue similar to #11291. 

For my use case in OneDrive web, the tiles of video uses `signal` class for play icon. So even after changes in `Tile.scss` the play icon for videos was coming in black shade in case of dark mode, to fix that I updated `Signal.scss` too, I am not sure if that would break any other component using `Signal.scss`, will be helpful if any of @aneeshack4  @phkuo or @ThomasMichon can confirm this once.


play_button_web

micahgodbolt:
Is this ready to be merged?

aneeshack4:
@micahgodbolt might be good to have @ThomasMichon look at it since he's component owner but since it's just a scss change maybe it's not needed - I think the Signal change is a bit unclear

PR #11594, opened 2020-01-02
Fix event listener and debounced function memory leaks in List component

KevinTCoughlin:
#### Pull request checklist

#### Description of changes

There are a number of event listener callbacks that are not unsubscribed when the List unmounts. This leads to run-time memory leaks (**~15 kB** in this example) captured in the below _before_ and _after_ heap snapshots.

---

The fix is to terminate the event listeners both via `EventGroup#off` and invoke the debounced / throttled callbacks in `componentWillUnmount` e.g. `this._onAsyncResize()`.

To debug and verify, I performed the following steps:

1. Rendered "Shimmer" List example in isolation locally.
1. Invoked garbage collection and captured the initial heap "Snapshot 1".
1. Toggled Shimmer "Off", invoked garbage collection, and captured heap "Snapshot 2".
1. Compared "Snapshot 2" against "Snapshot 1" and filtered by "Detached".

#### Before fix

In the "Shimmer" List example, ~15 kB of DOM Nodes are reported as "Detached" and therefore cannot be reclaimed by GC.

before_heap

Here are the object retainers:

before_heap_retainers

#### After fix

after_heap

#### Focus areas to test

- CI passes
- List examples render as expected

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11594)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/List/List.tsx
+++ b/packages/office-ui-fabric-react/src/components/List/List.tsx
@@ -316,6 +316,14 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
   }
 
   public componentWillUnmount(): void {
+    this._events.off(window, 'resize', this._onAsyncResize);
+    this._events.off(this._scrollElement, 'scroll', this._onScroll);
+    this._events.off(this._scrollElement, 'scroll', this._onAsyncScroll);
+
+    this._onAsyncIdle();
+    this._onAsyncResize();
+    this._onAsyncScroll();
+
     this._async.dispose();
     this._events.dispose();
   }
--- a/change/office-ui-fabric-react-2020-01-02-13-37-03-keco-memory-leaks.json
+++ b/change/office-ui-fabric-react-2020-01-02-13-37-03-keco-memory-leaks.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Terminate List event listeners on unmount to fix memory leaks",
+  "packageName": "office-ui-fabric-react",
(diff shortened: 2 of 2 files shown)


packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -316,6 +316,14 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
   }
 
   public componentWillUnmount(): void {
+    this._events.off(window, 'resize', this._onAsyncResize);
```
ecraig12345: Doesn't disposing the EventGroup automatically do this for each registered listener?

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -316,6 +316,14 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
...
+    this._events.off(this._scrollElement, 'scroll', this._onScroll);
+    this._events.off(this._scrollElement, 'scroll', this._onAsyncScroll);
+
+    this._onAsyncIdle();
```
ecraig12345: Can you add a comment inline explaining why these must be called?

KevinTCoughlin:
I need to take some more time with debugging this. There are `FocusZone` and `styled` DOM Node leaks due to retained closures.

I still believe that the debounced functions need to be cleared on List's `componentWillUnmount`. However, I need to find the heap snapshot that proves that.

PR #11596, opened 2020-01-03
Pickers: fix issue where same option would be selected repeatedly.

### LINKED ISSUE #8451: PeoplePicker: same person can be selected multiple times when using keyboard.
### Environment Information 
 
- __Package version(s)__: Package version(s): Latest Fabric UI React

#### Actual behavior:
 
When you quickly press the return or tab key in a short time while the suggested people callout window is active, it adds the same person as many times as you pressed the key.

developer microsoft com_en-us_fabric
 
#### Expected behavior:
 
Selected person with keyboard can only be add once.

Requested priority: Normal

joschect:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #8451

#### Description of changes

Consolidate some of the picker completion logic. It brings keyboard selection in line with mouse selection of items. It also fixes a bug where pressing enter rapidly would result in the same item being selected many times.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11596)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
@@ -195,9 +195,7 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends BaseComponent<
 
   public completeSuggestion(forceComplete?: boolean) {
     if (this.suggestionStore.hasSelectedSuggestion() && this.input.current) {
-      this.addItem(this.suggestionStore.currentSuggestion!.item);
-      this.updateValue('');
-      this.input.current.clear();
+      this.completeSelection(this.suggestionStore.currentSuggestion!.item);
     } else if (forceComplete) {
       this._completeGenericSuggestion();
     }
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -746,12 +743,17 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends BaseComponent<
     );
   };
 
-  protected addItemByIndex = (index: number): void => {
-    this.addItem(this.suggestionStore.getSuggestionAtIndex(index).item);
+  protected completeSelection = (item: T) => {
+    this.addItem(item);
```
aneeshack4: Just checking my understanding - this addition of checking current input prevents it from clearing selection if it's already been selected? If so, shouldn't the updateValue call with empty string happen after the if?

joschect: Do you mean `if(this.input.current)` ? That's just checking to make sure that the input.current already exists before clearing it out. That's a separate thing from clearing the currently entered value. Both things should happen but are independent of one another. `updateValue` triggers the clearing of the suggested items that are available to the user. That should always get cleared out/get called. 

It's possible that the `updateValue` call should be contained within the if statement. But I'm not sure it needs to be.

Issue #11561, opened 2019-12-26
Support LTR context in a RTL page


micahgodbolt:
Contextual RTL is usually what we need to do, but sometimes you might have an RTL page that you need LTR in, and it doesn't appear to work right now.

Hard to repro right now because I found the bug while fixing rtl in createTheme. Will update once #11560 is merged.

PR #11597, opened 2020-01-03
Fabric: Render wrapper container correctly if dir prop changes.

jdhuntington:
#### Pull request checklist

- [X] Addresses an existing issue: #11561 (this does not fix *all* LTR in RTL issues, but does fix several)

#### Description of changes

When nested in a context with `dir` specified in theme, this control would not correctly render the container with the opposite direction specified in the outer `div`, causing confusing results.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11597)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
@@ -34,18 +34,16 @@ export class FabricBase extends React.Component<
   }
 
   public render() {
-    const { as: Root = 'div', theme, dir } = this.props;
-    const classNames = this._getClassNames();
-    const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
+    const { dir, theme } = this.props;
     const isRTL = dir === 'rtl';
-    let renderedContent = <Root {...divProps} className={classNames.root} ref={this._rootElement} />;
+    const needDirectionChange = dir !== undefined && theme && (theme.rtl === undefined || theme.rtl !== isRTL);
 
-    // Expose an rtl based theme if dir is specified and it doesn't agree with the theme setting.
-    if (dir !== undefined && theme && (theme.rtl === undefined || theme.rtl !== isRTL)) {
-      renderedContent = <Customizer settings={{ theme: getRTLTheme(theme, isRTL) }}>{renderedContent}</Customizer>;
+    if (needDirectionChange) {
+      const switchedTheme = getRTLTheme(theme!, isRTL);
+      return <Customizer settings={{ theme: switchedTheme }}>{this._renderRoot(switchedTheme)}</Customizer>;
+    } else {
+      return this._renderRoot(theme!);
     }
-
-    return renderedContent;
   }
 
   public componentDidMount(): void {
@@ -63,10 +61,17 @@ export class FabricBase extends React.Component<
     }
   }
 
-  private _getClassNames(): IProcessedStyleSet<IFabricStyles> {
-    const { className, theme, applyTheme } = this.props;
+  private _renderRoot(theme: ITheme) {
+    const { as: Root = 'div' } = this.props;
+    const classNames = this._getClassNames(theme);
+    const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
+    return <Root {...divProps} className={classNames.root} ref={this._rootElement} />;
+  }
+
+  private _getClassNames(theme: ITheme): IProcessedStyleSet<IFabricStyles> {
+    const { className, applyTheme } = this.props;
     const classNames = getClassNames(getStyles, {
-      theme: theme!,
+      theme,
       applyTheme: applyTheme,
       className,
       isFocusVisible: this.state.isFocusVisible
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -35,17 +35,23 @@ export class FabricBase extends React.Component<
...
-    const classNames = this._getClassNames();
-    const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
     const isRTL = dir === 'rtl';
-    let renderedContent = <Root {...divProps} className={classNames.root} ref={this._rootElement} />;
 
-    // Expose an rtl based theme if dir is specified and it doesn't agree with the theme setting.
+    // Adjust dir if prop is specified and it doesn't agree with the theme setting.
     if (dir !== undefined && theme && (theme.rtl === undefined || theme.rtl !== isRTL)) {
-      renderedContent = <Customizer settings={{ theme: getRTLTheme(theme, isRTL) }}>{renderedContent}</Customizer>;
+      const newTheme = getRTLTheme(theme, isRTL);
```
jdhuntington: There is some duplication in the `if` and `else` branches, but I think pulling out the common parts and making completely DRY would lead to a brittle result.

micahgodbolt: @JasonGore and I talked about this idea when I added RTL to Fabric. Adding customizer to Fabric might be overloading the control a bit. 

Should we be creating a new control that provides theme and Fabric together? For instance, our layer uses `Fabric`, so we're loading customizer with just about every control now.

JasonGore: Oh boy, I didn't realize (or remember) when we talked that `Layer` was actually using `Fabric` as well. I wonder what the rationale is for `Layer` to use `Fabric`? If it's simply to provide context, I don't think that's needed any more with React portals. If it's to watch for keypresses and use focus outlining, I think that would also propagate through React portals now, as well. We should explore removing `Fabric` from `Layer` entirely.

In any case, it seems not good to have `Fabric` and `Customizer` for every layered component (ContextualMenu, Callout, Dropdown, DatePicker, DetailsHeader, KeytipLayer, Dialog, Modal, Panel, etc.)

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -35,17 +35,23 @@ export class FabricBase extends React.Component<
...
+      const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
+      return (
+        <Customizer settings={{ theme: newTheme }}>
+          <Root {...divProps} className={classNames.root} ref={this._rootElement} />
+        </Customizer>
+      );
+    } else {
+      const classNames = this._getClassNames(this.props.theme!);
+      const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
+      return <Root {...divProps} className={classNames.root} ref={this._rootElement} />;
```
JasonGore: Isn't `divProps` the same in both branches? It only seems to depend on `this.props`, and I'm not seeing how that would differ in each conditional branch.

If you do have to repeat yourself, could you do:
```tsx
      const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
      const rootContent = <Root {...divProps} className={classNames.root} ref={this._rootElement} />;
      return (
        <Customizer settings={{ theme: newTheme }}>
          {rootContent}
        </Customizer>
      );
    } else {
      const classNames = this._getClassNames(this.props.theme!);
      return rootContent;
    }
```

JasonGore: Err, I see now `classNames` differs for each, but it could also be a function that returns the same content applying `classNames`. (I also missed an `if` statement there.)

JasonGore: I edited the original code comment to show what I meant. :)

jdhuntington: Yes, this would work, but I think the creation of a lambda just to take in `rootClassName` makes the intent less clear, rather than more. As written, everything that's happening is pretty straightforward.

JasonGore: Well, I think at least `divProps` could be centralized. I'm not sure why that's duplicated and actually obfuscated intent when I reviewed it as I could not see why there were two instances of it.

micahgodbolt:
didn't realize that this was still open when I did this work: #11626.

#11626 seems to handle LTR in RTL properly. Is that a preferred approach? 

Also, I didn't realize that @dzearing had added Customizer to Fabric component a late last year. Are we confident this is what we want to do moving forward?

jdhuntington:
Closing in favor of #11626

PR #11604, opened 2020-01-03
Search focus

### LINKED ISSUE #8261: With ContextualMenu open, clicking in SearchBox does not properly focus SearchBox
### Environment Information 
 
- __Package version(s)__: 6.143.1
- __Browser and OS versions__: All (different behaviors in each)

1. Open this [CodeSandbox](https://codesandbox.io/s/8llkop3wo2).
2. Click on the button to open the context menu
3. Click on the SearchBox

 
#### Actual behavior:
 
**Chrome:** Search box is focused, but does not show focus borders and the 'search' icon does not disappear. However, user is able to type in search box still.
**Edge:** Search box does not show focus borders and 'search' icon does not disappear. However, use is NOT able to type in the search box, despite the browser showing a blinking cursor.
**Firefox:** Search box does not show focus borders, 'search' icon does not disappear, no blinking cursor, User is unable to type.
 
#### Expected behavior:
 
Search box focuses (and shown with focus borders and search icon disappearing) and user is able to type

Are you willing to submit a PR to fix? Yes, but may not have time
 
Requested priority: Normal
 
micahgodbolt:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #8261

#### Description of changes

To put search more in line with Textfield, i removed the event group and put onBlur back on the input. The event group has been there since before 2017, so i'm not 100% why someone thought it was necessary. 

If you know why this approach wouldn't work, please do tell, otherwise it fixes #8261

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11604)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
@@ -3,7 +3,6 @@ import { ISearchBoxProps, ISearchBoxStyleProps, ISearchBoxStyles } from './Searc
 import {
   initializeComponentRef,
   warnDeprecations,
-  EventGroup,
   getId,
   KeyCodes,
   classNamesFunction,
@@ -32,7 +31,6 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
   private _inputElement = React.createRef<HTMLInputElement>();
   private _latestValue: string;
   private _fallbackId: string;
-  private _events: EventGroup | undefined;
 
   public constructor(props: ISearchBoxProps) {
     super(props);
@@ -62,12 +60,6 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
       this.setState({
         value: newProps.value || ''
       });
-    }
-  }
-
-  public componentWillUnmount() {
-    if (this._events) {
-      this._events.dispose();
     }
   }
 
@@ -115,7 +107,6 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
         <input
           {...nativeProps}
           id={id}
-          onFocusCapture={() => console.log('focus capture')}
           className={classNames.field}
           placeholder={placeholderValue}
           onChange={this._onInputChange}
@@ -126,6 +117,7 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
           role="searchbox"
           aria-label={ariaLabel}
           ref={this._inputElement}
+          onBlur={this._onBlur}
         />
         {value!.length > 0 && (
           <div className={classNames.clearButton}>
@@ -185,11 +177,6 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
       hasFocus: true
     });
 
-    if (!this._events) {
-      this._events = new EventGroup(this);
-    }
-    this._events.on(ev.currentTarget, 'blur', this._onBlur, true);
-
     if (this.props.onFocus) {
       this.props.onFocus(ev as React.FocusEvent<HTMLInputElement>);
     }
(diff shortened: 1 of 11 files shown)


micahgodbolt:
Tests run locally. Good sign.

micahgodbolt:
hmm...build failing but not sure why. nothing obvious (to me) in the logs

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
```diff
@@ -65,13 +63,6 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
     }
   }
 
-  // tslint:disable-next-line function-name
-  public UNSAFE_componentWillMount() {
```
ecraig12345: Umm.........what on earth was this doing..............

micahgodbolt:
I found the PR where this was changed:

https://github.com/OfficeDev/office-ui-fabric-react/pull/523

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
ecraig12345: Typo (by me) 🙄 ...and then the UNSAFE_ prefix was added in a general pass.

micahgodbolt:
@joschect is also looking into another possible solution here

ecraig12345:
> @joschect is also looking into another possible solution here

If the fix continues to use the EventGroup, please move the disposal to `componentWillUnmount`. 🤣

micahgodbolt:
@ecraig12345 Jon is good here. Are you ready to sign off?

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
```diff
@@ -100,7 +100,13 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
     ]);
 
     return (
-      <div role="search" ref={this._rootElement} className={classNames.root} onFocusCapture={this._onFocusCapture}>
+      <div
+        role="search"
+        ref={this._rootElement}
+        className={classNames.root}
+        onBlurCapture={this._onBlur}
```
ecraig12345: Why the capture version?

micahgodbolt: we need to capture blur whether it happens on the input or the button. I could add onBlur to both, but that seems inconsistent.

ecraig12345: No I mean I'm curious why you needed to catch it specifically in the capture phase

micahgodbolt: oh, i didn't think about the different phases in this case. Added it to be consistent with focus being at that level, and reduce redundant onBlur I would have had to put onto the button. I can go either way.

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
```diff
@@ -119,6 +111,7 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
           placeholder={placeholderValue}
           onChange={this._onInputChange}
           onInput={this._onInputChange}
+          onBlur={this._onBlur}
```
ecraig12345: Will the input blur when the button gets focus?

micahgodbolt: This behavior isn't changing. when button gets focus, onBlur and onFocus get called again. The blur is just moved from an event listener to an input callback

https://github.com/OfficeDev/office-ui-fabric-react/pull/11604/files#diff-c1170007a89461079892a4ffc0886f74L190

ecraig12345: Oh, got it. I'm not sure that's desirable, but if it's the same as the previous behavior, it's fine for now.

micahgodbolt: yeah, i toyed around with a number of things to see if I could remove the double onFocus, but didn't have any luck. As you said, it's not a regression.

PR #11605, opened 2020-01-03
FocusZone: clear disposable function references on unmount to avoid memory leak

KevinTCoughlin:
#### Pull request checklist

#### Description of changes

Currently, `FocusZone` maintains references to event handlers pushed on to `this._disposables` after unmount. Therefore, the callback context and element reference cannot be GC'ed.

This fix is similar to the pre-existing solution in `BaseComponent`, see https://github.com/OfficeDev/office-ui-fabric-react/blob/498eb74150d4aa2b724c178aa38b8a55cea1ce95/packages/utilities/src/BaseComponent.ts#L93

#### Heap snapshot

image

#### Focus areas to test

- CI should pass.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11605)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
+++ b/packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
@@ -199,6 +199,9 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
     // Dispose all events.
     this._disposables.forEach(d => d());
 
+    // Clear function references.
+    delete this._disposables;
+
     // If this is the last outer zone, remove the keydown listener.
     if (_outerZones.size === 0 && _disposeGlobalKeyDownListener) {
       _disposeGlobalKeyDownListener();
--- a/packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
+++ b/packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
@@ -199,7 +199,7 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
     // Dispose all events.
     this._disposables.forEach(d => d());
 
-    // Clear function references.
+    // Clear function references so their closures can be garbage-collected.
     delete this._disposables;
 
     // If this is the last outer zone, remove the keydown listener.
--- a/change/office-ui-fabric-react-2020-01-03-13-05-22-keco-focuszone-leaks.json
+++ b/change/office-ui-fabric-react-2020-01-03-13-05-22-keco-focuszone-leaks.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "FocusZone: Delete function references on unmount",
+  "packageName": "office-ui-fabric-react",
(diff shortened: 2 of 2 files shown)


packages/office-ui-fabric-react/src/components/FocusZone/FocusZone.tsx
```diff
@@ -199,6 +199,9 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
     // Dispose all events.
     this._disposables.forEach(d => d());
 
+    // Clear function references.
```
ecraig12345: Might be good to be even more specific:
```suggestion
    // Clear function references so their closures can be garbage-collected.
```

JasonGore:
Is there a larger issue here? I'm wondering if it's a problem that `BaseComponent` and `FocusZone` instances aren't getting garbage collected rather than just some of their class members. I'm not sure how these class members would otherwise be keeping instances of `BaseComponent` and `FocusZone` around.

KevinTCoughlin:
> Is there a larger issue here? I'm wondering if it's a problem that `BaseComponent` and `FocusZone` instances aren't getting garbage collected rather than just some of their class members. I'm not sure how these class members would otherwise be keeping instances of `BaseComponent` and `FocusZone` around.

@JasonGore, the larger issue is that there are a number of script references throughout Fabric UI that reference DOM which do not clean-up their references. As a result, DOM Nodes referenced become "Detached" because V8 (as an example) does not know it can safely GC those Nodes. Only after fixing those script references will the closure and DOM Nodes be GC'ed, relinquishing browser memory. That will require time and a number of fixes such as this one.

To test locally, capture a memory heap snapshot of base state. Then perform an operation, force GC, and capture another heap snapshot. Then set the view to "Compare" from "Snapshot 1" to "Snapshot 2". Filter by "Detached" to see the DOM that cannot be GC'ed.

For this example, I tested by mounting and unmounting `DetailsList` on a blank page as it includes the most Fabric components. There are additional issues to fix in other areas of code to free-up leaked `DetailsList` DOM.

I can follow-up with some of those candidates if you wish.

https://v8.dev/blog/tracing-js-dom

KevinTCoughlin:
Additional note that `BaseComponent` is unaffected by this problem because it null's out `this.__disposables` on unmount. `FocusZone` no longer extends from `BaseComponent` and implements its own "disposables" logic that retains the closures until this fix is merged.

ecraig12345:
@KevinTCoughlin In general, would the suspect components be those that have their own EventGroup, Async, or disposables using `on`?

KevinTCoughlin:
> @KevinTCoughlin In general, would the suspect components be those that have their own EventGroup, Async, or disposables using `on`?

Yes, usage of the above utilities are prone to this type of closure leak. Particularly `Async` and `EventGroup`. I have a few examples from SPO of those errors as well as heap snapshots of existing problems in Fabric that I will share offline with the team to triage.

PR #11608, opened 2020-01-04
Fix editor bugs

### LINKED ISSUE #11489: Editor is running workers on UI thread
The live code editor on the website is supposed to use web workers on separate threads for TypeScript and other purposes, to prevent this resource-intensive code from running on the UI thread and causing lag. However, there's a built-in browser security restriction that code from a different origin/domain than the original page (such as a CDN) can't create workers ([details here, sort of](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers)). It's supposed to be possible to work around this by using blob URLs or data URIs to load the code, and we tried to do this, but at least in Chrome it's currently falling back to running workers on the main thread. ~~Suspect this has to do with changing `configureEnvironment` to use blob URLs instead of data URIs.~~

~~We may need to vary the approach based on what works per-browser--data URIs seemed to work in Chrome but not old Edge, and blobs may work in old Edge but not Chrome.~~

**EDIT:** it still runs on the UI thread in Chrome/new Edge with data URIs, at least per the console warnings. 😞

### LINKED ISSUE #11490: Editor not working properly in demo app
The live code editor isn't currently working in the demo app, at least in PR deploy.

It's also including significant parts of Monaco in the bundle used on initial page load, which isn't ideal.

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11489, fixes #11490

#### Description of changes

Switch back to loading worker files using data-uri, except in Edge. I'd previously switched to loading workers with blob URLs to make Edge work, but that seems to have caused workers to load on the UI thread in Chrome.

Switch fabric-website-resources to build esnext by default instead of commonjs, to prevent the editor from being bundled (and add notes about this requirement various places).

Remove the UMD bundle config from fabric-website-resources to prevent chunks from the UMD and serve configs from stomping on each other.

Update Monaco version, and add a build step to get rid of problems in Monaco's .d.ts files: new `get` syntax which isn't supported in TS 3.5, and references to a global ambient `monaco`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11608)


packages/tsx-editor/src/components/TsxEditor.tsx
```diff
@@ -138,15 +145,17 @@ function _loadTypes(supportedPackages: IPackageGroup[]): Promise<void> {
...
-  // React must be loaded first
+  // React types must be loaded first (don't use import() to avoid potential bundling issues)
   promises.push(
-    // @ts-ignore: this import is handled by webpack
-    import('!raw-loader!@types/react/index.d.ts')
-      // raw-loader 0.x exports a single string, and later versions export a default.
-      // The package.json specifies 0.x, but handle either just in case.
-      .then((result: string | { default: string }) => {
+    new Promise<void>(resolve =>
+      require.ensure([], require => {
```
kenotron: Why don't we make sure that this is bundled as ESM instead of CommonJS so that this switch to require.ensure isn't needed?

ecraig12345:
Before merging, need to verify that the editor isn't getting pulled into the main website bundle.

PR #11614, opened 2020-01-06
[Migration] Update test fixtures to handle windows path issue.

### LINKED ISSUE #11464: Migration: Update Snapshots Fails in Windows
### Environment Information

- **Package version(s)**: 7.73.0

`yarn buildto migrations` - ensure migrations package was built
`yarn update-snapshots` - this can be run either from repo root or `packages/migration` to repro the issue

Update snapshots command fails on multiple tests:
```
  ● warnAutobindRemoved › warns when autobind is used

    ENOENT, no such file or directory '\\?\C:\Users\Jason\src\oufr-jg-i\packages\migration\_root\7.0.0\autobind.tsx'

      53 |   try {
      54 |     migration.step({ dryRun: false, warn: (msg: string) => result.warnings.push(msg) });
    > 55 |     const contents = await readFileAsync(path.join('_root', filename));
         |                            ^
      56 |     teardown();
      57 |     result.contents = contents.toString();
      58 |     return result;

      at Object.<anonymous> (../../node_modules/mock-fs/lib/binding.js:520:15)
      at maybeCallback (../../node_modules/mock-fs/lib/binding.js:68:18)
      at Object.<anonymous>.Binding.open (../../node_modules/mock-fs/lib/binding.js:493:10)
      at Object.<anonymous> (../../node_modules/mock-fs/lib/index.js:33:39)
      at Object.<anonymous> (src/tests/mods/utils/index.ts:55:28)
      at fulfilled (../../node_modules/tslib/tslib.js:107:62)
```
image

#### Actual behavior:

`yarn update-snapshots` fails on jest tests.

#### Expected behavior:

`yarn update-snapshots` should not fail on jest tests without local changes.

jdhuntington:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #11464

#### Description of changes

Update migration tests to deal with node's path treatment on windows.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11614)


change/@uifabric-migration-2020-01-06-10-02-55-master.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
JasonGore: I think this could actually be "none" since this is a test-only change.

JasonGore: Actually, maybe it's not test-only? I'm not sure based on the path, but it does seem to be in execution. I guess you'll let me know. :)

packages/migration/src/tests/mods/utils/index.ts
```diff
@@ -15,7 +15,7 @@ const fileSystem: any = {};
 async function recordFile(filename: string): Promise<void> {
   const contents = await readFileAsync(filename);
   const trimmedName = filename.substring(fixturesRoot.length);
-  const parts = path.dirname(trimmedName).split(path.sep);
+  const parts = path.dirname(trimmedName).split('/'); // Intentionally not using path.sep, as node is giving '/' for separator instead of os-correct version.
```
ecraig12345: Umm....per the docs, it should be OS-correct?? https://nodejs.org/api/path.html#path_path_sep

JasonGore:
One thing I forgot: as part of this PR, can you remove [the build task redefined here](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/migration/just.config.js#L6) and verify it works? (The only reason it was redefined was to exclude tests originally.)

We may have to deal with a merge conflict with #11574 depending on whether you or I merge first.

JasonGore:
Actually, the task looks like it simply needs to be defined as this (sorry for the confusion):

```tsx
task('build', 'build:node-lib').cached();
```

packages/migration/src/tests/mods/utils/index.ts
jdhuntington: Well, `path.sep` is correct, but filename globbing returns os-independent pathnames so it ends up being moot.

ecraig12345: Got it. In that case can you update the comment?

JasonGore:
I've merged #11574 which means you'll have to add `test` as a command to `packages/migration/package.json` [like so](https://github.com/OfficeDev/office-ui-fabric-react/blob/f0beb2605e7154f9ed6948de88209b783540f133/packages/office-ui-fabric-react/package.json#L23) in order for unit tests to run. (I would have added this myself but couldn't since the tests weren't working.)

PR #11618, opened 2020-01-07
FocusTrapZone: Fix event handler and element reference memory leaks

KevinTCoughlin:
#### Pull request checklist

#### Description of changes

Fixes memory leaks in `FocusTrapZone` due to element references and event listeners in script that cannot be GC'ed.

Discovered this testing SPFx WebPart "PropertyPane" in SharePoint Online. Currently, we leak the focused DOM Node(s) when PropertyPane(s) are opened then closed.

#### Heap snapshot

Highlighted are the script references I am cleaning-up in this PR. Note the `+2` DOM Nodes that are not GC'ed when the Panel is closed (the snapshot I have selected). I have a similar `Popup` PR to follow.

image

To capture the above, I opened and closed the "A Dialog nested in a Panel" example on http://localhost:4322/#/examples/focustrapzone capturing heap snapshots before and after for comparison and filtering by "Detached".

#### Focus areas to test

- CI should pass.

cc: @patmill 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11618)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
+++ b/packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
@@ -72,6 +72,21 @@ export class FocusTrapZone extends React.Component<IFocusTrapZoneProps, {}> impl
     ) {
       this._returnFocusToInitiator();
     }
+
+    // Dispose of event handlers so their closures can be garbage-collected
+    if (this._disposeClickHandler) {
+      this._disposeClickHandler();
+      this._disposeClickHandler = undefined;
+    }
+
+    if (this._disposeFocusHandler) {
+      this._disposeFocusHandler();
+      this._disposeFocusHandler = undefined;
+    }
+
+    // Dispose of element references so the DOM Nodes can be garbage-collected
+    delete this._previouslyFocusedElementInTrapZone;
+    delete this._previouslyFocusedElementOutsideTrapZone;
   }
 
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -72,6 +72,21 @@ export class FocusTrapZone extends React.Component<IFocusTrapZoneProps, {}> impl
     ) {
       this._returnFocusToInitiator();
     }
+
+    // Dispose of event handlers so their closures can be garbage-collected
+    if (this._disposeClickHandler) {
```
ecraig12345: Should you get rid of root, firstBumper, and lastBumper too, or is removing DOM node references from refs handled automatically?

KevinTCoughlin: I haven't looked at React's source, but the fields utilizing React's Ref seem to be cleaned-up automatically. I considered your suggestion, but have not seen any of those refs in the heap snapshots. More motivation to use Refs wherever possible vs. manual property assignment.

ecraig12345: Interesting, that's good to know!

PR #11619, opened 2020-01-07
Popup: Fix memory leak due to element reference in callback closure

KevinTCoughlin:
#### Pull request checklist

#### Description of changes

Continuation of #11618 this time in `Popup`. This change fixes memory leaks in `Popup` due to script reference to element in callback handler closure.

#### Heap snapshot

image

#### Focus areas to test

(optional)

cc: @patmill 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11619)


KevinTCoughlin:
BTW the DOM Nodes are still not able to be GC'ed after both PRs. It looks like there are a few more dangling references in various utilities and components that need to be addressed. The candidates are highlighted:

image

packages/office-ui-fabric-react/src/components/Popup/Popup.tsx
```diff
@@ -66,6 +66,9 @@ export class Popup extends React.Component<IPopupProps, IPopupState> {
         this._originalFocusedElement.focus();
       }
     }
+
+    delete this._originalFocusedElement;
```
ecraig12345: For good measure I wonder if you should also dispose and de-reference `this._async`. Might also be good to add a comment about why this is needed (at least until the pattern is prevalent enough throughout the codebase that people just follow it).

KevinTCoughlin: Done! However, I opted not to de-reference `this._async` as I don't think its necessary until I see it pop-up in snapshots. The disposables array is the problematic pattern because it allows for one to store closures without an explicit clean-up call on unmount.

packages/office-ui-fabric-react/src/components/Popup/Popup.tsx
```diff
@@ -53,6 +53,8 @@ export class Popup extends React.Component<IPopupProps, IPopupState> {
 
   public componentWillUnmount(): void {
     this._disposables.forEach((dispose: () => void) => dispose());
+    this._async.dispose();
```
dzearing: I think you can just add async to disposables and that's all you need to do. I don't think having a reference to the element would keep things alive.

KevinTCoughlin: It is the closure in the `this._disposables` array that is retaining the DOM element so `delete this._disposables` or what you recommended should suffice. I'll revise and verify in a few.

KevinTCoughlin: In my testing we need to explicitly de-reference `this._originalFocusedElement` because it gets captured via `_this` in the closure of the transpiled `_onKeyDown` function.

image

image

I've reverted my changes to reflect a fix for the above.

PR #11620, opened 2020-01-07
Fix bugs with API reference tables

ecraig12345:
#### Pull request checklist

#### Description of changes

Fix bugs with API reference table generation, including one which prevented class members and methods from showing up on the doc site.

For API reference pages, show all property tables on page load. It makes sense to hide the extra tables under "See more" on a control page, but that's not really necessary/helpful on a reference page which only contains API tables.

Restyle the "Members" and "Methods" headers on the class reference tables to have better spacing and stand out more.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11620)

### DIFF
--- a/packages/api-docs/src/tableRowJson.ts
+++ b/packages/api-docs/src/tableRowJson.ts
@@ -29,6 +29,7 @@ export function createTableRowJson(collectedData: ICollectedData, apiItem: ApiIt
         apiProperty.excerptTokens,
         apiProperty.propertyTypeExcerpt.tokenRange
       );
+      break;
     }
 
     case ApiItemKind.Constructor:
@@ -41,6 +42,7 @@ export function createTableRowJson(collectedData: ICollectedData, apiItem: ApiIt
         // The constructor is similar to a method, but we have to manually add the name.
         tableRowJson.name = 'constructor';
       }
+      break;
     }
 
     case ApiItemKind.Function:
--- a/packages/api-docs/src/types.ts
+++ b/packages/api-docs/src/types.ts
@@ -50,7 +50,7 @@ export interface ITableJson {
    * `[{ text: 'Readonly<' }, { text: 'IFoo', linkedPage: 'Foo', linkedPageGroup: 'components' }, { text: '>' }]`
    */
   extendsTokens?: ILinkToken[];
-  members?: ITableRowJson[] | IEnumTableRowJson[];
+  members: ITableRowJson[] | IEnumTableRowJson[];
   deprecated?: boolean;
   deprecatedMessage?: string;
 }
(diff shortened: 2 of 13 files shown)


packages/api-docs/src/tableRowJson.ts
```diff
@@ -29,6 +29,7 @@ export function createTableRowJson(collectedData: ICollectedData, apiItem: ApiIt
         apiProperty.excerptTokens,
         apiProperty.propertyTypeExcerpt.tokenRange
       );
+      break;
```
jdhuntington: This seems like a function that would benefit from some tests. Was the lack of `break;` causing issues, or just here for consistency?

packages/api-docs/src/types.ts
```diff
@@ -50,7 +50,7 @@ export interface ITableJson {
    * `[{ text: 'Readonly<' }, { text: 'IFoo', linkedPage: 'Foo', linkedPageGroup: 'components' }, { text: '>' }]`
    */
   extendsTokens?: ILinkToken[];
-  members?: ITableRowJson[] | IEnumTableRowJson[];
+  members: ITableRowJson[] | IEnumTableRowJson[];
```
khmakoto: Just want to confirm, we're making the `members` required because there is no reason to have a table without them right?

packages/example-app-base/src/components/ApiReferencesTable/ApiReferencesTableSet.tsx
```diff
@@ -162,8 +167,10 @@ function _generateTableProps(jsonDocs: IPageJson | undefined): IApiReferencesTab
       });
     }
 
-    // to ensure that I{componentName}Props comes first
-    if (kind === 'interface' && propsName === table.name) {
+    if (jsonDocs.group === 'references' && jsonDocs.name === name) {
+      results.unshift(tableProps);
+    } else if (kind === 'interface' && propsName === name) {
```
khmakoto: I'm very surprised we weren't using `name` instead of `table.name` before given that we had already extracted it from above.

packages/api-docs/src/tableRowJson.ts
ecraig12345: It was making properties get parsed/output as methods. 🤦‍♀ It would be nice if there was a lint rule for "no switch case fall-through except on consecutive lines"...

ecraig12345: While I totally agree it would be good to have tests, that would require more time/thought to set up due to the reliance on API Extractor models. So I'd rather not do that right now.

packages/api-docs/src/types.ts
ecraig12345: The main goal was to address the mismatch between this interface and the mirrored one in DocPage.types.ts. Thinking about it more, I'm going to revert this and make members optional in the other interface (and update the spot in ApiReferencesTableSet which assumed members was defined), because type aliases don't have members.

PR #11626, opened 2020-01-07
Fabric: Properly handle dir being passed into control

micahgodbolt:
#### Description of changes

Earlier Fabric component update set `direction: rtl` based off of the theme alone. This update will look first to the `dir` attribute (if provided), and then look to the theme.rtl (if provided) before setting `dir` attribute of the Fabric component. 

Updated tests as well. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11626)


micahgodbolt:
@dzearing are you good with this change?

packages/office-ui-fabric-react/src/components/Fabric/Fabric.base.tsx
```diff
@@ -36,9 +36,12 @@ export class FabricBase extends React.Component<
   public render() {
     const { as: Root = 'div', theme, dir } = this.props;
     const classNames = this._getClassNames();
-    const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
+    const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties, ['dir']);
+    // If dir is set via props, use that value, otherwise if theme is set, use theme RTL setting, otherwise, set undefined.
+    const htmlDir =
+      dir !== undefined ? dir : theme !== undefined && theme.rtl !== undefined ? (theme.rtl === true ? 'rtl' : 'ltr') : undefined;
```
dzearing: minor nit but I believe just:

```tsx
const themeDir = theme.rtl ? 'rtl' : 'ltr';
const htmlDir = dir !== undefined ? dir : themeDir;
```

micahgodbolt:
per david's suggestion, the dir logic is a bit more simple and will always return RTL or LTR for a Fabric component.

micahgodbolt:
had to do a bit more work to account for page level RTL (which does not affect theme). Reworked a bit of my and Davids Customizer logic to be a bit easier to follow.

PR #11627, opened 2020-01-07
Button: Use aria-checked for checkbox and menuitemcheckbox roles

micahgodbolt:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11627)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
+++ b/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
@@ -194,6 +194,10 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
       (this.props as any)['data-is-focusable'] === false || (disabled && !allowDisabledFocus) || this._isSplitButton ? false : true;
 
     const isCheckboxTypeRole = role === 'menuitemcheckbox' || role === 'checkbox';
+    // if isCheckboxTypeRole, always return a checked value, otherwise only return checked value if toggle is set to true
+    // This is because role="checkbox" always needs to have an aria-checked value
+    // but our checked prop only sets aria-pressed if we mark the button as a toggle="true"
+    const checkedOrPressedValue = isCheckboxTypeRole ? !!checked : toggle === true ? !!checked : undefined;
 
     const buttonProps = assign(nativeProps, {
       className: this._classNames.root,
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
```diff
@@ -207,7 +210,9 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
       'aria-describedby': ariaDescribedBy,
       'aria-disabled': isPrimaryButtonDisabled,
       'data-is-focusable': dataIsFocusable,
-      'aria-pressed': toggle ? !!checked : undefined // aria-pressed attribute should only be present for toggle buttons
+      // aria-pressed attribute should only be present for toggle buttons
+      // aria-checked attribute should only be present for toggle buttons with checkbox type role
```
khmakoto: I was very confused at this PR until I read this comment, this is why inline documentation is helpful 😃

packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
```diff
@@ -194,6 +194,10 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
       (this.props as any)['data-is-focusable'] === false || (disabled && !allowDisabledFocus) || this._isSplitButton ? false : true;
 
     const isCheckboxTypeRole = role === 'menuitemcheckbox' || role === 'checkbox';
+    // if isCheckboxTypeRole, always return a checked value, otherwise only return checked value if toggle is set to true
+    // This is because role="checkbox" always needs to have an aria-checked value
+    // but our checked prop only sets aria-pressed if we mark the button as a toggle="true"
+    const checkedOrPressedValue = isCheckboxTypeRole ? !!checked : toggle === true ? !!checked : undefined;
```
khmakoto: Do we really need the `toggle === true` check? Does `toggle` alone not work?

micahgodbolt: I did that to make it obvious it was a condition and not a value. sometimes those nested ternaries can be annoying to read.

micahgodbolt:
```jsx
const checkedOrPressedValue = isCheckboxTypeRole ? !!checked : toggle ? !!checked : undefined;
```
if isCheckboxTypeRole is false then return toggle? no..., first check for toggle===true then return !!checked or undefined

PR #11629, opened 2020-01-07
add icon support to combobox

makopch-ms:
#### Description of changes

Currently, you cannot add an icon to be shown before the combobox's input box.  This change adds IIconProps to the Combobox props so this is now an option.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11629)


makopch-ms:
@jspurlin - please review :)

jspurlin:
Can you add some screenshots (both LTR and RTL)? I'm curious how this looks

joschect:
@makopch-ms Looks like you might need to run `yarn update-snapshots`

makopch-ms:
@jspurlin 
image

micahgodbolt:
Can we avoid adding custom 1 off icons and just support 'onRenderLabel' or change 'label' to JSX?

packages/office-ui-fabric-react/src/components/__snapshots__/ComboBox.Basic.Example.tsx.shot
```diff
@@ -4638,5 +4638,6 @@ exports[`Component Examples renders ComboBox.Basic.Example.tsx correctly 1`] = `
       
     </div>
   </div>
+  /&gt;
```
jdhuntington: This seems out of place.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -1207,6 +1206,21 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
     }
   };
 
+  private _onRenderLabel = (props: IComboBoxProps): JSX.Element | null => {
```
jdhuntington: Since the rendering depends on `this._multiselectAccessibleText`, it feels like it should be a parameter to this function and not an instance variable. That way when a user overrides this function they can access all necessary state.

makopch-ms:
@jdhuntington @micahgodbolt @jspurlin  could you please take another look

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.types.ts
```diff
@@ -246,6 +246,15 @@ export interface IComboBoxProps extends ISelectableDroppableTextProps<IComboBox,
    * Optional iconButton props on combo box
    */
   iconButtonProps?: IButtonProps;
+
+  /**
+   * Custom render function for the label text.
+   */
+  onRenderLabel?: (
+    props?: IComboBoxProps,
+    multiselectAccessibleText?: string,
```
joschect: I think put this into a single prop object rather than having IComboBoxProps and muktiselectAccessibleText be separate. We might want to keep things consistent with the [IRenderFunction](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/utilities/src/IRenderFunction.ts).

micahgodbolt:
I feel that updating a single input control to support icons before the label is a bad idea. If combobox has them, why not toggle, checkbox, dropdown, searchbox, and every other input that we have? We either need a holistic approach to support this for every form control with a label, or we need to find a 1 off way to accomplish this without adding a ton of extra opinionated props to the control.

micahgodbolt:
Is there anything wrong with not supplying a label to your Combobox, and using the Label control with an icon instead? This is an approach you can do today with zero code change.

https://codepen.io/micahgodbolt/pen/zYxLXVR?editors=0010


```jsx
        <Label htmlFor="myComboBox">
          <Icon iconName="Airplane" /> My Label
        </Label>
        <ComboBox
          id="myComboBox"
          options={INITIAL_OPTIONS}
        />
```

jspurlin:
@micahgodbolt this change is design driven (at least for the Combobox). Additionally, the label is directly tied to the combobox and they should be treated as such.

I thought the earlier proposed onRenderLabel approach allowed that to be achieved while also not changing any existing behavior

micahgodbolt:
Ok, if this is actually design driven, then I'm okay with an interface change, as we want to promote consistent use/setup of label icons.

Is there a plan for updating other controls? Would it be better to add icon functionality into the Label and expose it on each control?

micahgodbolt:
Ok, noticing that PR has been changed to onRender label. I have no problem with this type of update. I would be curious if this design change is something we'd want to support across the board, and should eventually be added to the Label control....or at least consistently across all form controls.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.types.ts
micahgodbolt: If we could use IRenderFunction that'd keep the interface consistent with other `onRender` functions.

jspurlin:
I think that's a great question and should be followed up separate of this PR. I think at this point it's not an across the board approach, but again we should follow up on this.

packages/office-ui-fabric-react/src/components/ComboBox/examples/ComboBox.Basic.Example.tsx
```diff
@@ -78,7 +78,6 @@ export class ComboBoxBasicExample extends React.Component<{}, IComboBoxBasicExam
             }}
           />
         </div>
```
jspurlin: >div [](start = 10, length = 3)

There's only whitespace changes in this file, is this needed?

makopch-ms:
@joschect could you please review again

makopch-ms:
@jdhuntington @micahgodbolt I think I am blocked my the previous requested changes.  Jon signed off so let me know if there is anything else.

PR #11638, opened 2020-01-08
ComboBox: Fix focus border and chevron button styling issues

### LINKED ISSUE #11631: ComboBox style issues
Issues:
1. Chevron down button is not centered vertically
image

2. Focus border rendering is bad in certain resolution, this is caused by setting `overflow:hidden` on `ms-ComboBox`
image

xugao:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #11631

#### Description of changes
1. Fixed chevron down button miss alignment in combobox
2. Fix focus border style bug under certain resolution:
before:
image
after:
image

#### Train of thoughts on the changes I made:
The reason the focus border was rendering incorrectly was due to `overflow: hidden` on the `ms-ComboBox`. I believe the reason that it was added is that to make the chevron button doesn't overflow out of the comboBox like this:
image
After correcting the positioning of the chevron button to be correctly sized and aligned, it's still cutting off the border under certain resolutions, like:
image
This leads me to make the change to remove the border styles we had in ComboBox and moved to instead setting border using `:after` content. And some styles were updated due to this change.

#### Focus areas to test
- Browsers: Edge, Chrome, FF
- Different zoom levels
- High contrast mode

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11638)

### DIFF
--- a/change/@uifabric-experiments-2020-01-07-17-49-22-xgao-fix-combobox-styles.json
+++ b/change/@uifabric-experiments-2020-01-07-17-49-22-xgao-fix-combobox-styles.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "ComboBox: fix chevron button and focus border styling",
+  "packageName": "@uifabric/experiments",
+  "email": "xgao@microsoft.com",
+  "commit": "b1f23673efec6c713219c81d02be736f3dc7a373",
+  "date": "2020-01-08T01:48:44.522Z"
+}
--- a/change/@uifabric-styling-2020-01-07-18-22-34-xgao-fix-combobox-styles.json
+++ b/change/@uifabric-styling-2020-01-07-18-22-34-xgao-fix-combobox-styles.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "add borderPosition param to getInputFocusStyle",
+  "packageName": "@uifabric/styling",
+  "email": "xgao@microsoft.com",
+  "commit": "927aefb4447731b6eb0aa84f1fb57e3adc638bdb",
+  "date": "2020-01-08T02:22:34.442Z"
+}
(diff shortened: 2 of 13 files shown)


micahgodbolt:
@xugao can you give a quick explination of how this approach solves the problem where the past one did not? You changed a number of things, what was the underlined problem/solution?

xugao:
@micahgodbolt - good call! updated PR description and added comment inline

ecraig12345:
Thanks for working on this! One potential issue with setting all borders using :after (when most components use real element borders) is that it makes them harder to override, and makes the overriding approach vary even more between components. Not to say that means we can't do this approach, just something to consider.

xugao:
@ecraig12345 - these are super valid points. let me see if i can find better way

xugao:
@ecraig12345 - i tried another alternative, which is adding the focus border using `border` css instead of `after`. i believe that can fix the issue too, but it's not a pretty solution either. fundamentally, this is issue with browser sub-pixel rendering. a demo here: https://codesandbox.io/s/sub-pixel-rendering-bonel

I think i still overall like my current solution. i do wish i can keep combobox and dropdown styles more alike, but the hover style on the chevron button on combox is a stopper.

ecraig12345:
@xugao Maybe you could see if @micahgodbolt or @dzearing or maybe @phkuo have any ideas? I'm not the best person to help with subtle CSS issues unfortunately.

micahgodbolt:
I'm happy to take a crack at it. I'm curious how far we'd get following the approach used on the dropdown.

xugao:
@micahgodbolt  - feel free! what do you mean by following the approach used on the dropdown? it's not really comparable with dropdown since dropdown don't have the chevron as stand-alone button with hover affect. fundamentally we are walking around the issue of this: https://codesandbox.io/s/sub-pixel-rendering-bonel

xugao:
@micahgodbolt shared a good idea with me offline. i will try it out

xugao:
After some digging, the problem still boils down to the sub-pixel rendering issue. current solution is still the best i can find, which works visually correct in all browsers.  will merge this and communicate this change with our partners

PR #11642, opened 2020-01-08
Change focus class behavior: Alternate between 'hidden' and 'visible'

paflakst:
This change aims to enable an approach where **visible focus indication** is the default/baseline treatment, and hidden focus indication is the special treatment.

The current Fabric approach is the opposite, and this encourages disabling focus indication as the baseline. In short: Fabric wants focus indication to depend on Fabric. Accessibility-wise this feels wrong – and perhaps rightly so?

Reasons why visible focus indication as the baseline might be better:
- It seems like a better fail-safe than hidden, as it doesn’t entail obstructing users or breaking the law. (_Think f. ex. if a version update changed the class name, or if the <Fabric> component was removed or replaced. Temporary situations, sure, but still._)
- The best-practice principle would probably favor it. (_For one, all major browsers have visible focus indication as their default._)
- We don’t know what kind of interface a non-mouse user might be using. We do however know it won’t necessarily be a keyboard, so keyboard interaction should not be a prerequisite for focus indication. Even though WCAG only talks about a requirement for visible keyboard focus, we might want to keep this in mind.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11642)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Fabric/Fabric.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Fabric/Fabric.styles.ts
@@ -16,10 +16,12 @@ export interface IFabricClassNames {
 export const getStyles = (props: IFabricStyleProps): IFabricStyles => {
   const { theme, className, isFocusVisible, applyTheme } = props;
   const classNames = getGlobalClassNames(GlobalClassNames, theme);
+  const focusVisibility = isFocusVisible ? 'Visible' : 'Hidden';
   return {
     root: [
       classNames.root,
-      isFocusVisible && 'is-focusVisible ms-Fabric--isFocusVisible',
+      // keywords for search: is-focusVisible ms-Fabric--isFocusVisible ms-Fabric--isFocusHidden
+      `is-focus${focusVisibility} ms-Fabric--isFocus${focusVisibility}`,
       theme.fonts.medium,
       {
         direction: theme.rtl ? 'rtl' : undefined,
--- a/packages/utilities/src/initializeFocusRects.ts
+++ b/packages/utilities/src/initializeFocusRects.ts
@@ -1,7 +1,8 @@
 import { getWindow } from './dom/getWindow';
 import { isDirectionalKeyCode } from './keyboard';
+import { setFocusVisibility } from './setFocusVisibility';
 
-export const IsFocusVisibleClassName = 'ms-Fabric--isFocusVisible';
+export { IsFocusVisibleClassName } from './setFocusVisibility';
 
 /**
  * Initializes the logic which:
@@ -9,11 +10,12 @@ export const IsFocusVisibleClassName = 'ms-Fabric--isFocusVisible';
  * 1. Subscribes keydown and mousedown events. (It will only do it once per window,
  *    so it's safe to call this method multiple times.)
  * 2. When the user presses directional keyboard keys, adds the 'ms-Fabric--isFocusVisible' classname
- *    to the document body.
- * 3. When the user clicks a mouse button, we remove the classname if it exists.
+ *    to the document body, removes the 'ms-Fabric-isFocusHidden' classname.
+ * 3. When the user clicks a mouse button, adds the 'ms-Fabric-isFocusHidden' classname to the
+ *    document body, removes the 'ms-Fabric--isFocusVisible' classname.
  *
- * This logic allows components on the page to conditionally render focus treatments only
- * if the global classname exists, which simplifies logic overall.
+ * This logic allows components on the page to conditionally render focus treatments based on
+ * the existence of global classnames, which simplifies logic overall.
  *
  * @param window - the window used to add the event listeners
  */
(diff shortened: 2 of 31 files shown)


micahgodbolt:
@paflakst this is a pretty significant change which might be better started as an issue. We've spent a good deal of energy on our keyboard focus solution and any change to that solution would take a good deal of discussion, prototyping, and deliberation. I don't believe that a PR is the right place to start that discussion.

paflakst:
> @paflakst this is a pretty significant change which might be better started as an issue. We've spent a good deal of energy on our keyboard focus solution and any change to that solution would take a good deal of discussion, prototyping, and deliberation. I don't believe that a PR is the right place to start that discussion.

@micahgodbolt it actually started as an email exchange between @dzearing and myself. He was very positive to this change, and asked me if was willing to create a PR for it.

I get how this might come across as a pretty significant change, but is it really though...? The "change" is really just an _addition_ of something new that was not there before. It should not affect anything already existing.

Please see the points mentioned in the description. If you think something there doesn't make sense, then let's discuss :)

micahgodbolt:
@paflakst Ah, I didn't know you'd already discussed this with David. That certainly changed this from an 'out of the blue PR' to something we asked you to propose. Thanks for the FYI.

My primary questions will be around backwards compatibility and if users will need to do anything to continue using Fabric in the way they are used to.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.styles.ts
```diff
@@ -16,10 +16,12 @@ export interface IFabricClassNames {
 export const getStyles = (props: IFabricStyleProps): IFabricStyles => {
   const { theme, className, isFocusVisible, applyTheme } = props;
   const classNames = getGlobalClassNames(GlobalClassNames, theme);
+  const focusVisibility = isFocusVisible ? 'Visible' : 'Hidden';
   return {
     root: [
       classNames.root,
-      isFocusVisible && 'is-focusVisible ms-Fabric--isFocusVisible',
+      // keywords for search: is-focusVisible ms-Fabric--isFocusVisible ms-Fabric--isFocusHidden
+      `is-focus${focusVisibility} ms-Fabric--isFocus${focusVisibility}`,
```
micahgodbolt: ok, this makes sense. instead of adding `is-focusVisible ms-Fabric--isFocusVisible` to the Fabric control when focus is visible, we toggle between:

```
is-focusVisible ms-Fabric--isFocusVisible

// and 

is-focusHidden ms-Fabric--isFocusHidden
```

micahgodbolt:
Code looks sound. Would like to get a passing build to make sure it is acting as expected.

packages/utilities/src/initializeFocusRects.ts
```diff
@@ -28,25 +30,9 @@ export function initializeFocusRects(window?: Window): void {
 }
 
 function _onMouseDown(ev: MouseEvent): void {
-  const win = getWindow(ev.target as Element);
```
jdhuntington: 👍 for removing the duplicate logic

paflakst:
> @paflakst Ah, I didn't know you'd already discussed this with David. That certainly changed this from an 'out of the blue PR' to something we asked you to propose. Thanks for the FYI.
> 
> My primary questions will be around backwards compatibility and if users will need to do anything to continue using Fabric in the way they are used to.

Yeah, sorry, I should probably have mentioned that I'd already been in touch. As far as backwards compatibility goes, it should not affect anyone. All users should be able to continue using Fabric as they are used to.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.styles.ts
paflakst: That's correct.

paflakst: ...so with respect to the 'visible' class, this PR should change nothing.

packages/office-ui-fabric-react/src/components/Fabric/Fabric.styles.ts
```diff
@@ -16,10 +16,12 @@ export interface IFabricClassNames {
 export const getStyles = (props: IFabricStyleProps): IFabricStyles => {
   const { theme, className, isFocusVisible, applyTheme } = props;
   const classNames = getGlobalClassNames(GlobalClassNames, theme);
+  const focusVisibility = isFocusVisible ? 'Visible' : 'Hidden';
```
aneeshack4: Just to clarify, we're currently just disabling focus altogether in the alternate case but now instead of that we're toggling to hidden?

paflakst: Currently, the "visible" class is added on keystroke detection (of a defined set of "directional" keys). That same class is then removed on mouse-click detection. That's it. So, there is actually no "hidden" concept in Fabric currently; it is only _implicitly implied_ by the absence of the "visible" class. This is a bit unfortunate, as that absence could also be due to other factors. More on this in the PR description. Alternating between the "hidden" and "visible" eliminates ambiguity and gives users more flexibility in how to define focus indication. (Mainly: No more need to disable focus indication by default.)

jdhuntington:
Kicking off new CI run to get this unblocked.

jdhuntington:
I believe the [CI run has passed successfully](https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=46829&view=results), despite github's status report.

paflakst:
Thanks for accepting & completing the PR 😃

PR #11645, opened 2020-01-08
switching to using self-hosted agent pool for the CPU intensive jobs

kenotron:
#### Pull request checklist

#### Description of changes

With this change, we should be bringing our builds from ~20minutes (thanks @JasonGore!) to ~12-15 minutes. 

Obviously with this kind of change, we are now the masters of our own destiny (and burden is on us to manintain these boxes). However, I have automated the process to scale up or down our VMs in use so that it isn't a huge burden. I'm separately sending out an email about how to maintain the Azure DevOps agent pool to an internal audience.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11645)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,135 +12,137 @@ pool:
 
 jobs:
   - job: Build
+    pool: 'Self Host Ubuntu'
     steps:
-    - template: azure-pipelines.tools.yml
+      - template: azure-pipelines.tools.yml
 
-    - script: |
-        yarn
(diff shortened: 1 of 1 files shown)


kenotron:
Per offline discussion, we think that this perf regression is a false alarm.

PR #11647, opened 2020-01-08
Fix yarn start

ecraig12345:
#### Pull request checklist

#### Description of changes

Changing the `fabric-website-resources` `tsconfig.json` to use `module: esnext` for some reason causes spurious errors during `yarn start`. The most obvious fix is to revert this change (go back to `commonjs`) and make updates in `tsx-editor` instead to work around the issue that motivated the change.

The change to `module: esnext` was done to prevent dynamic `import()` statements in tsx-editor from breaking code splitting, which happens because for `commonjs`, `import()` is transformed into `require` (see https://github.com/webpack/webpack/issues/5703#issuecomment-357512412). To keep us or consumers using commonjs from running into this issue, I switched all dynamic imports to use `require.ensure` instead. This isn't as clean or modern, but it's effective at preventing subtle bundling issues for either us or consumers, which I think is more important for now.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11647)

### DIFF
--- a/packages/tsx-editor/src/components/TsxEditor.tsx
+++ b/packages/tsx-editor/src/components/TsxEditor.tsx
@@ -81,8 +81,15 @@ function _useGlobals(supportedPackages: IPackageGroup[]): boolean {
     Promise.all(
       supportedPackages.map(group => {
         if (!win[group.globalName]) {
-          // tslint:disable-next-line:no-any
-          return group.loadGlobal().then((globalModule: any) => (win[group.globalName] = globalModule));
+          // tslint:disable:no-any
+          return new Promise<any>(resolve => {
+            // handle either promise or callback function
+            const globalResult = group.loadGlobal(resolve);
+            if (globalResult && (globalResult as PromiseLike<any>).then) {
+              globalResult.then(resolve);
+            }
+          }).then((globalModule: any) => (win[group.globalName] = globalModule));
+          // tslint:enable:no-any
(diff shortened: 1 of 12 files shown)


packages/tsx-editor/src/components/TsxEditor.tsx
```diff
@@ -138,15 +145,17 @@ function _loadTypes(supportedPackages: IPackageGroup[]): Promise<void> {
...
-  // React must be loaded first
+  // React types must be loaded first (don't use import() to avoid potential bundling issues)
   promises.push(
-    // @ts-ignore: this import is handled by webpack
-    import('!raw-loader!@types/react/index.d.ts')
-      // raw-loader 0.x exports a single string, and later versions export a default.
-      // The package.json specifies 0.x, but handle either just in case.
-      .then((result: string | { default: string }) => {
+    new Promise<void>(resolve =>
+      require.ensure([], require => {
```
kenotron: This now places a requirement on our consumers to use webpack which is kind of what we had already with that `!raw-loader!` syntax...

ecraig12345: Yeah the editor only ever worked with webpack

PR #11655, opened 2020-01-09
Use default semantic html instead of aria attributes and roles

### LINKED ISSUE #10860: [Accessibility][Screen reader]: Narrator announces same position number for all facepile controls. (Bug 6)
Tracking all of the issues brought up with the 1/4 accessibility pass on Facepile control 
Details on individual bugs can be found on [SharePoint](https://microsoft-my.sharepoint.com/:x:/r/personal/v-amaver_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7Bfd43fcc5-515d-47d8-946d-0d4862a56278%7D&action=default&uid=%7BFD43FCC5-515D-47D8-946D-0D4862A56278%7D&ListItemId=1647&ListId=%7B395E1ECB-5002-440E-81C5-6A090E98077C%7D&odsp=1&env=prodbubble) 

Breakout of issue #7534

Markionium:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10860

#### Description of changes

Fixes accessibility  bug #10860 in Facepile

#### Focus areas to test

Facepile accessibility

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11655)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Facepile/Facepile.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Facepile/Facepile.base.tsx
@@ -58,7 +58,7 @@ export class FacepileBase extends BaseComponent<IFacepileProps, {}> {
         {this.onRenderAriaDescription()}
         <div className={_classNames.itemContainer}>
           {showAddButton ? this._getAddNewElement() : null}
-          <ul className={_classNames.members} role="listbox" aria-label={ariaLabel}>
+          <ul className={_classNames.members} aria-label={ariaLabel}>
             {this._onRenderVisiblePersonas(personasPrimary, personasOverflow.length === 0 && personas.length === 1)}
           </ul>
           {overflowButtonProps ? this._getOverflowElement(personasOverflow) : null}
@@ -90,12 +90,7 @@ export class FacepileBase extends BaseComponent<IFacepileProps, {}> {
         ? onRenderPersona(persona, this._getPersonaControl)
         : onRenderPersonaCoin(persona, this._getPersonaCoinControl);
       return (
-        <li
-          role="option"
-          key={`${singlePersona ? 'persona' : 'personaCoin'}-${index}`}
-          className={this._classNames.member}
-          aria-label={persona['aria-label'] || persona.personaName}
-        >
+        <li key={`${singlePersona ? 'persona' : 'personaCoin'}-${index}`} className={this._classNames.member}>
           {persona.onClick
             ? this._getElementWithOnClickEvent(personaControl, persona, index)
             : this._getElementWithoutOnClickEvent(personaControl, persona, index)}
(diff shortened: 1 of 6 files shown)


packages/office-ui-fabric-react/src/components/Facepile/Facepile.base.tsx
```diff
@@ -58,7 +58,7 @@ export class FacepileBase extends BaseComponent<IFacepileProps, {}> {
         {this.onRenderAriaDescription()}
         <div className={_classNames.itemContainer}>
           {showAddButton ? this._getAddNewElement() : null}
-          <ul className={_classNames.members} role="listbox" aria-label={ariaLabel}>
```
aneeshack4: Why is this role being removed? It's needed for ATs to understand groups of Facepiles when they are in a list of people right? ARIA says we can use HTML `<select>` instead but it doesn't support images.

packages/office-ui-fabric-react/src/components/Facepile/Facepile.base.tsx
```diff
@@ -90,12 +90,7 @@ export class FacepileBase extends BaseComponent<IFacepileProps, {}> {
         ? onRenderPersona(persona, this._getPersonaControl)
         : onRenderPersonaCoin(persona, this._getPersonaCoinControl);
       return (
-        <li
-          role="option"
```
aneeshack4: This is being removed because each child of listbox needs to have `role="option"` right? But if we do that aren't we fully removing support of a parent-child relationship with Personas that are used in Facepile?

packages/office-ui-fabric-react/src/components/Facepile/Facepile.base.tsx
aneeshack4: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role

Markionium: The link you pasted says the following.

> The listbox role is used to identify an element that creates a list from which a user may select one or more static items

I think the Facepile is just a "list" of people, and not a picker of any sort, where the user can **select** one.

Currently it will announce it as a "group"/"list" which might make more sense. Especially since the list can be a mix of clickable and non-clickable items.

packages/office-ui-fabric-react/src/components/Facepile/Facepile.base.tsx
Markionium: The `<ul />` `<li />` relationship should be sufficient to indicate the parent child relationship between the list and the items in the list. 

Doing this will also make the AT do the native thing with regards to list and announce item 1 or 2 etc. correctly.

Markionium:
Please hold off merging this as I'll need to verify it using Narrator instead of just Voice over.

PR #11658, opened 2020-01-09
OverflowSet: Don't assume role or orientation. 

### LINKED ISSUE #11646: ARIA attribute is not allowed: aria-orientation="horizontal" when role="group"
### Environment Information

- **Package version(s)**: all

### Describe the issue:
OverflowSet allows items to have role="group" within them. However, OverflowSet always sets the aria-orientation on the items. According to the documentation on aria-orientation [(link)](https://www.w3.org/TR/wai-aria/#aria-orientation), it should not be set alongside role="group".

Not in codepen but in onenote online...
Step 1: Open Onenote Online using edge browser.
Step 2: Navigate to “Edit Notebook” button by using tab key.
Step 3: Navigate to “Edit Note in browser field” menu-item by using tab key and select it.
Step 4: Navigate to Home tab by using ctrl+F6 and select it.
Step 5: Navigate to Insert tab by using arrow key and activate it.
Step 6: Navigate to “Table” button by using tab key and expand it.”
Step 7: Select row and column and press enter key.
Step 8: Navigate to “Table” tab by using “ctrl+F6” and select it.
Step 9: Open Developer tool bar (F12)
Step 10: Verify whether Elements must only use allowed ARIA attributes.

This is the line that sets the aria-orientation without checking the role it is setting [OverflowSetBase](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx)

```
 <Tag aria-orientation={vertical ? 'vertical' : 'horizontal'} {...uniqueComponentProps} className={this._classNames.root} role={role}>

        {items && this._onRenderItems(items)}

        {overflowItems && overflowItems.length > 0 && this._onRenderOverflowButtonWrapper(overflowItems)}

      </Tag>
```

#### Actual behavior:

If you pass an item with role="group", aria-orientation is incorrectly set.

#### Expected behavior:

If you pass an item with role="group", aria-orientation should not be set. There should be a check to ensure we are not passing role="group" before setting the property.

[Aria-orientation](https://www.w3.org/TR/wai-aria/#aria-orientation)

micahgodbolt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11646 Fixes #11548

#### Description of changes

OverflowSet is a utility, not a menu. It is often used in menus, but it also used as a group or tablist or a way to render non interactive content. 

So rather than assume menu and orientation (and therefore needing menuitem as well #11548), this PR strips all role/aria assumptions from the overflowset, and add docs to inform users that these values will be necessary. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11658)


micahgodbolt:
chatting with @jspurlin about if a default behavior is desired or not.

micahgodbolt:
Ok, a few tweaks. After talking with Jeremy, the default behavior of an OverflowSet is simply a group, similar to [this example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_group_role):

```html
<div role="menu">
  <ul role="group">
    <li role="menuitem">Inbox</li>
    <li role="menuitem">Archive</li>
    <li role="menuitem">Trash</li>
  </ul>
  <ul role="group">
    <li role="menuitem">Custom Folder 1</li>
    <li role="menuitem">Custom Folder 2</li>
    <li role="menuitem">Custom Folder 3</li>           
  </ul>
  <ul role="group">
    <li role="menuitem">New Folder</li>
  </ul>
</div>
```
 
So we'll leave the role generic unless otherwise noted, and we'll only set orientation if role="menubar" is provided.

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -104,8 +105,9 @@ export class CheckboxBase extends React.Component<ICheckboxProps, ICheckboxState
     return (
       <KeytipData keytipProps={keytipProps} disabled={disabled}>
         {(keytipAttributes: any): JSX.Element => (
-          <div className={this._classNames.root} title={title}>
+          <div role="presentation" className={this._classNames.root} title={title}>
```
jspurlin: >title={title} [](start = 69, length = 13)

Why does this have a title? Having both title and role=presentation doesn't make sense... Really, what is this element?

jspurlin:
>         role={'menubar'}

Consider updating this be a toolbar (then the contents don't have to be "menuitem'-based)

---
Refers to: packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.base.tsx:103 in c7b575d. [](commit_id = c7b575d630e7fe5891f80384d1dbff7b3c3092b4, deletion_comment = False)

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -142,11 +145,8 @@ export class OverflowSetBase extends BaseComponent<IOverflowSetProps, {}> implem
 
   private _onRenderItems = (items: IOverflowSetItemProps[]): JSX.Element[] => {
     return items.map((item, i) => {
-      const wrapperDivProps: React.HTMLProps<HTMLDivElement> = {
-        className: this._classNames.item
-      };
       return (
-        <div key={item.key} {...wrapperDivProps}>
+        <div role="presentation" key={item.key} className={this._classNames.item}>
```
jspurlin: >presentation [](start = 19, length = 12)

Note, you don't have to put role=presentation here. role=presentation just makes the HTML element not have the default role (<li> is an example, it has role=listitem) and div doesn't have one unless it is made an accessible node

packages/office-ui-fabric-react/src/components/Tooltip/TooltipHost.base.tsx
```diff
@@ -82,6 +82,7 @@ export class TooltipHostBase extends React.Component<ITooltipHostProps, ITooltip
 
     return (
       <div
+        role="presentation"
```
jspurlin: >presentation [](start = 14, length = 12)

same here, this isn't needed

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -104,8 +105,9 @@ export class CheckboxBase extends React.Component<ICheckboxProps, ICheckboxState
...
             <input
+              role={role}
```
joschect: Isn't this already included as part of `inputProps`

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
micahgodbolt: the element is just a wrapper. I added 'presentation' because I wanted to ensure that the input was representing the proper role.

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
micahgodbolt: good question.

jspurlin: If inputProps either has it explicitly or is utilizing nativeProps then it should have it

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
jspurlin: but this is going to make it show read out as "<title> group" This doesn't need role=presentation because screen readers are smart enough now that they don't add positioning elements to the accessiblity tree unless they have an accessible name (which this element is currently getting because it has a title... hence that also needs to be removed and then the element won't be represented in the accessiblity tree as expected)

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
micahgodbolt: ok...so it's only to change something that'd have a default value applied/assumed for it.

micahgodbolt: wait...but, doesn't the menubar expect its children to be menuitems?

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.base.tsx
micahgodbolt: fixed

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -53,7 +51,12 @@ export class OverflowSetBase extends BaseComponent<IOverflowSetProps, {}> implem
     }
 
     return (
-      <Tag aria-orientation={vertical ? 'vertical' : 'horizontal'} {...uniqueComponentProps} className={this._classNames.root} role={role}>
+      <Tag
+        role={role || 'group'}
+        aria-orientation={role === 'menubar' ? (vertical === true ? 'vertical' : 'horizontal') : undefined}
```
xugao: `role === 'menu' || role === 'menubar` here ?

micahgodbolt: Overflow set probably won't be used for a role="menu" (which is usually a contextual menu), so menubar is the primary role we're accounting for.

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
jspurlin: Yes, the focusable elements should be a menuitem-type role

micahgodbolt: the focusable element will be the control rendered inside of this wrapper div (button, link etc). So if that button has role="menuitem", then this div should have role="presentation", since it is the direct child of the root, when it has role="menu".

I could make the above role conditional to role="menubar" on the root, but is there harm in having it here?

jspurlin: see example 8 in the aria spec for [role=presentation](https://www.w3.org/TR/wai-aria-1.1/#presentation), `<span>` is an example of what I'm talking about (e.g. `<div>` does not have an explicit role (unless it is given an accessible name) and will be ignored). Historically, there's been an overuse of role presentation when it's not needed.

For example, I put together a [jsfiddle](https://jsfiddle.net/6a0ubhw1/) showing two button, one wrapped in a div. Note the resulting accessibility tree:
image

micahgodbolt: Thanks for the clarification. I agree, I have seen 'presenation' used a ton (details list for example), so I expected it was needed here. I can remove it.

packages/office-ui-fabric-react/src/components/OverflowSet/docs/OverflowSetOverview.md
```diff
@@ -2,4 +2,4 @@ The OverflowSet is a flexible container component that is useful for displaying
 
 ### Accessibility
 
-For an accessible OverflowSet, ensure that all surface level items (main items and overflow button) use `role="menuitem"`
+By default, the OverflowSet is simply `role=group`. If you used as a menu, you will need to add `role="menubar"` and add proper aria roles to each rendered item (`menuitem`, `menuitemcheckbox`, `menuitemradio`)
```
khmakoto:
```suggestion
By default, the OverflowSet is simply `role=group`. If used as a menu, then `role="menubar"` should be added, as well as the proper aria roles for each rendered item (`menuitem`, `menuitemcheckbox`, `menuitemradio`).
```

PR #11659, opened 2020-01-09
Text: Fixing children check so it can render {0} value

### LINKED ISSUE #11641: <Text>component cannot display 0 inside jsx but can display other numbers
### Environment Information

- **Package version(s)**: 7.78.3
- **Browser and OS versions**:Chrome Browser:79.0.3945.88 (Official Build) (64-bit), OS: Windows 10 Pro v1903

import {Text} from "office-ui-fabric-react"
const x = 0;
<pre><code>
&lt;Text&gt;{x}&lt;/Text &gt;
</pre></code>
#### Actual behavior:
Does'nt display anything
#### Expected behavior:

Should display 0

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11641 

#### Description of changes

This PR fixes an issue with `Text` that returned `null` on render if its `children` was `{0}` because it evaluated the condition to false. I searched for similar issues in the repo and couldn't find any.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11659)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Text/Text.view.tsx
+++ b/packages/office-ui-fabric-react/src/components/Text/Text.view.tsx
@@ -1,10 +1,11 @@
 /** @jsx withSlots */
+import * as React from 'react';
 import { withSlots, getSlots } from '../../Foundation';
 import { getNativeProps, htmlElementProperties } from '../../Utilities';
 import { ITextComponent, ITextProps, ITextSlots } from './Text.types';
 
 export const TextView: ITextComponent['view'] = props => {
-  if (!props.children) {
+  if (React.Children.count(props.children) === 0) {
     return null;
   }
 
--- a/change/office-ui-fabric-react-2020-01-09-11-08-17-textChildrenCheck.json
+++ b/change/office-ui-fabric-react-2020-01-09-11-08-17-textChildrenCheck.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Text: Fixing children check so it can render {0} value.",
+  "packageName": "office-ui-fabric-react",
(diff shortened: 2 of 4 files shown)


packages/office-ui-fabric-react/src/components/Text/Text.view.tsx
```diff
@@ -1,10 +1,11 @@
 /** @jsx withSlots */
+import * as React from 'react';
 import { withSlots, getSlots } from '../../Foundation';
 import { getNativeProps, htmlElementProperties } from '../../Utilities';
 import { ITextComponent, ITextProps, ITextSlots } from './Text.types';
 
 export const TextView: ITextComponent['view'] = props => {
-  if (!props.children) {
+  if (React.Children.count(props.children) === 0) {
```
JasonGore: Can children explicitly be `null` or `undefined`? What does the code do in these cases? What should we do in this case?

khmakoto: `React.Children.count` returns 0 in case `null` or `undefined` are passed. You can see that in [this codepen](https://codepen.io/khmakoto/pen/ZEYxqME?editors=0010)

JasonGore:
I should have thought of this when I reviewed the PR, but I think this is a perfect candidate for a simple snapshot test to make sure it doesn't regress in the future.

khmakoto:
@JasonGore I'll add the snapshot test.

PR #11662, opened 2020-01-09
Remove onBlur prop exclusion in Grid.Base

chang47:
#### Pull request checklist

#### Description of changes

In this PR https://github.com/OfficeDev/office-ui-fabric-react/pull/11526, we unintentionally fixed a bug where we didn't exclude an onBlur event from being included in the Grid component, however office has existing behaviors that relied on the onBlur event that the change caused to be excluded.

#### Focus areas to test

This is mostly relevant around the Swatch Color Picker area where we want events to be fired if focus is lost on the grid.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11662)

### DIFF
--- a/change/office-ui-fabric-react-2020-01-09-15-24-23-master.json
+++ b/change/office-ui-fabric-react-2020-01-09-15-24-23-master.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Remove onBlur from being excluded in grid",
+  "packageName": "office-ui-fabric-react",
+  "email": "chiechan@microsoft.com",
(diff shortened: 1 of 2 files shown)


chang47:
Looks like I might have jumped the gun for this, I realized where onBlur is being used, going back to re-iterate on the solution

chang47:
Updated thanks to Elizabeth!

chang47:
@ecraig12345 @jdhuntington Can either of you merge the branch? I would like this change make it to the release tomorrow so I can unblock office online tomorrow. Thanks!

PR #11682, opened 2020-01-11
Split out screener and perf-test jobs into different pipelines (cleaner PR)

kenotron:
#### Pull request checklist

#### Description of changes

UPDATE: more like ~12 minutes with Screener back inside the job; but still worth doing because perf-test isn't a gate - just info.

With this change, we can achieve **~8.5 min PR builds**!

image

#### Things that got changed:
1. Replaced the use of Jobs with Stages to have the better visualization as show in the screenshot
2. Changed the shorthand of "download" task to a full blown "DownloadPipelineArtifact" task because of the extra needed parameters
3. Added the concept of parameters in `azure-pipelines.artifacts.yml` because it is the only way to pass variables into a template (this is needed to accommodate (4) below)
4. Added 2 new pipeline yml match new build definitions for Perf & Screener - both using that artifact yml
5. The triggers of Perf Test & Screener are no longer part of the main PR Gate

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11682)


azure-pipelines.perf-test.yml
```diff
@@ -0,0 +1,61 @@
...
+  - task: GithubPRStatus@0
+    displayName: 'Update Github Pull Request Status'
+    inputs:
+      githubOwner: OfficeDev
+      githubRepo: 'office-ui-fabric-react'
+      githubContext: 'Pull Request Deployed Site'
+      githubDescription: 'Click "Details" to go to the Deployed Site'
+      githubTargetLink: 'http://fabricweb.z5.web.core.windows.net/pr-deploy-site/$(Build.SourceBranch)/'
+
+  - task: DeleteFiles@1
```
ecraig12345: This is because we're managing our own agents, right? What happens if the job fails before this step?

azure-pipelines.perf-test.yml
```diff
@@ -0,0 +1,61 @@
...
+pool:
+  vmImage: 'Ubuntu 18.04'
+
+steps:
+  # No need for checkout since we're using build artifacts.
+  - checkout: none
+
+  - template: azure-pipelines.artifacts.yml
+    parameters:
+      artifact: $(artifact)
```
ecraig12345: Where do these come from in this context?

azure-pipelines.yml
```diff
@@ -10,168 +10,120 @@ variables:
...
+          - publish: $(Build.ArtifactStagingDirectory)
+            artifact: Build-PR-$(Build.BuildNumber)
+
+          - script: |
+              az --version
+              az extension add -n azure-devops
+              echo $(buildQueuePat) | az devops login --organization https://uifabric.visualstudio.com
+              az devops configure --defaults organization=https://uifabric.visualstudio.com project="fabricpublic" --use-git-aliases true
+            env:
+              AZURE_DEVOPS_EXT_PAT: $(buildQueuePat)
```
ecraig12345: Does this have any security implications?

azure-pipelines.perf-test.yml
kenotron: these come from the perf-test build definition variables (sent thru the az pipeline build queue command)

azure-pipelines.perf-test.yml
kenotron: If the job fails before this step, it's kind of nice - that means we can examine the failed build artifacts via ssh if need be. Although we can definitely make this run even if the previous steps failed (delete files task was added before, I'd make that change as part of another change if need be)

azure-pipelines.yml
kenotron: We have a token here that triggers builds. So someone could potentially kick off more build jobs within the fabricpublic project which they already can by forking & submitting PRs.

azure-pipelines.perf-test.yml
JasonGore: Does this affect other PR jobs or only any given PR's pipeline? If this could pollute other PRs I definitely feel it should run at all times.

azure-pipelines.tools.yml
```diff
@@ -1,13 +1,13 @@
...
-  inputs:
-    versionSpec: 1.19.x
-    checkLatest: true
-    includePrerelease: false
+  - task: geeklearningio.gl-vsts-tasks-yarn.yarn-installer-task.YarnInstaller@3
+    displayName: 'Use Yarn 1.19.x'
+    inputs:
+      versionSpec: 1.19.x
+      checkLatest: true
+      includePrerelease: false
```
JasonGore: What is doing all of this formatting? I was following existing formatting **and** [Azure docs formatting](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml) when I made these changes.

JasonGore:
I'm concerned we should be a bit more disciplined about cleaning up perf-test and screener tasks, particularly if the artifacts hang around for all master and PR builds.

azure-pipelines.tools.yml
JasonGore: The indentation in other files seems excessive. The `-` serves as an indent marker. It'd be kinda like writing code like this:

```tsx
if (test)
  {
    some code;
  }
else
  {
    more code;
  }
```

kenotron: meh - this is prettier yaml stuff - I could add this stuff to the prettierignore?

kenotron: One thing I'm not sure you caught is that I'm moving the steps one level DOWN - so that will make all of it indent one more. So it used to be jobs -> steps, and I made it to be stages -> jobs -> steps to get that visualization showing up on the builds

azure-pipelines.perf-test.yml
kenotron: I'm adding a `condition: always()` to always clean up

kenotron:
Update: more like ~12 minutes still... but this is more solidly inside 12 minutes rather than the fluctuating stuff with perf-test involved.

image

azure-pipelines.perf-test.yml
```diff
@@ -0,0 +1,62 @@
...
+      githubRepo: 'office-ui-fabric-react'
+      githubContext: 'Pull Request Deployed Site'
+      githubDescription: 'Click "Details" to go to the Deployed Site'
+      githubTargetLink: 'http://fabricweb.z5.web.core.windows.net/pr-deploy-site/$(Build.SourceBranch)/'
+
+  - task: DeleteFiles@1
+    inputs:
+      SourceFolder: $(Build.SourcesDirectory)
+      Contents: '**/*'
+    condition: always()
```
JasonGore: nit: I see this repeated 4 times. Could have an `azure-pipelines.cleanup.yml` maybe?

JasonGore:
Another nit, this seems a bit obtuse now:

image

azure-pipelines.perf-test.yml
```diff
@@ -0,0 +1,62 @@
...
+  - template: azure-pipelines.artifacts.yml
+    parameters:
+      artifact: $(artifact)
+      buildId: $(buildId)
+
+  - template: azure-pipelines.tools.yml
+
+  - script: |
+      yarn
+    displayName: yarn
```
JasonGore: I thought I took this out in #11574. If this pipeline is downloading artifacts, it shouldn't need to run `yarn`. None of the other jobs do.

kenotron: I'll take it out and then merge if green.

azure-pipelines.perf-test.yml
kenotron: We're exploding in the azure-pipeline files! I want to have a ./devops folder first before adding a lot more of these files. To do that, we need to have DUPLICATES for a couple of days and jobs that are updated to use either. This takes some coordination, so I'm keeping it like this for now.

PR #11689, opened 2020-01-13
FocusZone: removing keydown listener correctly.

dzearing:
In nested FocusZones, a global keydown listener is added but not removed. This should address this and only add it on the first outer zone and remove it on the last outer zone.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11689)

### DIFF
--- a/change/office-ui-fabric-react-2020-01-13-09-02-40-focuszone-listener-fix.json
+++ b/change/office-ui-fabric-react-2020-01-13-09-02-40-focuszone-listener-fix.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "FocusZone: remove keydown listener correctly to avoid a leak with inner zones.",
+  "packageName": "office-ui-fabric-react",
+  "email": "dzearing@microsoft.com",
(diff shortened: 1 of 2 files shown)


JasonGore:
The [Related PR](https://github.com/microsoft/fluent-ui-react/pull/2227/files) also had a fix to `componetWillUnmount`, is that not needed here?

KevinTCoughlin:
> The [related PR](https://github.com/microsoft/fluent-ui-react/pull/2227/files) also had a fix to `componentWillUnmount`, is that not needed here?

As far as I can tell that change was only white-space related.

KevinTCoughlin:
> > The [related PR](https://github.com/microsoft/fluent-ui-react/pull/2227/files) also had a fix to `componentWillUnmount`, is that not needed here?
> 
> As far as I can tell that change was only white-space related.

Nevermind, @JasonGore you're right it is nested inside the `if` in both places in the original fix!

dzearing:
@JasonGore For symmetry, I moved the same dispose logic into the !innerzone block in unmount.

Issue #11072, opened 2019-11-05
Pivot does not have an accessible name 


taparikh:
### Environment Information

- **Package version(s)**: 7.8.2
- **Browser and OS versions**: Chrome Version 77.0.3865.120 (Official Build) (64-bit)

> Note that our screen reader support is limited to Edge + Narrator. Issues that do not reproduce in Edge + Narrator, and aren't caused by obvious invalid aria values, should be filed with the respective screen reading software, not the Fabric repo.

### Describe the issue:
Pivot does not have an accessible name when tested using the Accessibility Insights tool under the Custom Wdgets > Design Pattern Test (Assessment) :: We see <no value> 
When I check in the inspect element I do see the aria-label .

### Please provide a reproduction of the issue in a codepen:
https://codepen.io/pen?template=rEjZeV

#### Actual behavior:
The accessible name does not appear. 

#### Expected behavior:
The accessible name should appear

### Documentation describing expected behavior
This information is from the Accessibility Insights tool : A widget is an interactive interface component, such as a link, button, or combo box.  A custom widget is an interactive interface component other than a link or native HTML element. Custom widgets can be simple (e.g., a link or a button) or complex (e.g., a text field, listbox, and button that together function as a combo box). Each custom widget should follow the ARIA design pattern that best describes its function.

Below is a screenshot for the issue. 
![acc2](https://user-images.githubusercontent.com/51679982/68171154-7c8e0380-ff27-11e9-84e5-c6e8284746c1.JPG)

Below is a screenshot showing that aria-label gets applied to the component but still does not show up in the Accessibility insights test. 
![acc_1](https://user-images.githubusercontent.com/51679982/68171179-9c252c00-ff27-11e9-8c97-9de72a2930d5.JPG)

JasonGore:
The CodePen you linked seems to be a template.

However, I was still able to take a look at this and confirm what you are seeing. Even when the `Pivot` element containing the `role` of `tablist` has an `aria-label`, Accessibility Insights still shows `Accessible name: (no value)`. This means there isn't a fix I can identify from Fabric's side (at least not yet.) I suggest filing an issue with Accessibility Insights repos (repo links are in the [wiki](https://github.com/OfficeDev/office-ui-fabric-react/wiki/Accessibility-Troubleshooting#troubleshooting).)

![image](https://user-images.githubusercontent.com/26070760/68230677-59f2fd80-ffae-11e9-8e6a-797d11486b18.png)

[...]

PR #11690, opened 2020-01-13
Revert " Pivot: Move ARIA role to align with props assignment"

chang47:
Talking to the accessibility driver on our team and looking at the aria guidelines for [tab panels](https://www.w3.org/TR/2010/WD-wai-aria-practices-20100916/#tabpanel)

We believe that the tablist role should be on the div that contains the tab and not the container element  that holds the tabpanel

image

I'm reverting this change, because we believe the role was correct in its previous location.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11690)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Pivot/examples/Pivot.Basic.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/Pivot/examples/Pivot.Basic.Example.tsx
@@ -9,7 +9,7 @@ const labelStyles: Partial<IStyleSet<ILabelStyles>> = {
 
 export const PivotBasicExample: React.FunctionComponent = () => {
   return (
-    <Pivot>
+    <Pivot aria-label="Basic Pivot Example">
       <PivotItem
         headerText="My Files"
         headerButtonProps={{
--- a/packages/office-ui-fabric-react/src/components/Pivot/examples/Pivot.Basic.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/Pivot/examples/Pivot.Basic.Example.tsx
@@ -9,7 +9,7 @@ const labelStyles: Partial<IStyleSet<ILabelStyles>> = {
 
 export const PivotBasicExample: React.FunctionComponent = () => {
   return (
-    <Pivot aria-label="Basic Pivot Example">
+    <Pivot>
       <PivotItem
         headerText="My Files"
         headerButtonProps={{
--- a/change/office-ui-fabric-react-2020-01-13-11-05-09-revert-11465-11072-fix-pivot.json
+++ b/change/office-ui-fabric-react-2020-01-13-11-05-09-revert-11465-11072-fix-pivot.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Reverting moving tablist to be a parent of tabpanel",
+  "packageName": "office-ui-fabric-react",
+  "email": "chiechan@microsoft.com",
+  "commit": "e68a3b6aedc1d1a497b0015c938d0ce058804791",
+  "date": "2020-01-13T19:05:09.083Z"
+}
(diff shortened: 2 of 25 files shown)


packages/office-ui-fabric-react/src/components/Pivot/examples/Pivot.Basic.Example.tsx
```diff
@@ -9,7 +9,7 @@ const labelStyles: Partial<IStyleSet<ILabelStyles>> = {
 
 export const PivotBasicExample: React.FunctionComponent = () => {
   return (
-    <Pivot aria-label="Basic Pivot Example">
```
xugao: could you pls keep the `aria-label` added in the examples

xugao: and run `yarn update-snapshots` to update the tests

xugao:
i think we should also move `aria-label`  down to the element which has `role=tablist`
reference here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role
 
@JasonGore thoughts?

JasonGore:
Just to reiterate some of what we chatted about offline: The main issue here is that there is no singular `tab` role root, there are two roots: `tablist` and `tabpanel`. Most of our components apply top level native attributes to the root container. As we are finding, root component in this case can't have either tab role.

I'd still like to keep the top level native attributes applied to the root container for consistency with our other components. I also don't think it's good to split those top level native attributes and apply some of them elsewhere. I think the best thing to do here is add a prop (or set of props) that can be applied to the `tablist` element separately from the top level native attributes.

I'll tie up and reopen the associated issue for now.. if we resolve it in this PR it will again close the issue.

xugao:
@chang47 - to be explicit on the action here, after discussing with Jason offline.

We should add `tabListProps` as a new prop to the `<Pivot />` component, the type of the new prop should simply be `interface ITabListProps extends React.HTMLAttributes<HTMLDivElement>` and the prop should be spread on `ms-Pivot` element. Pls let me know if you can take this change along with this PR. thanks!

chang47:
@xugao I've talked with @JasonGore offline, we agreed to just revert the change because it's blocking word online and to not apply any forward fixes. 

I'm sorry to ask, but can we continue the check-in process?

xugao:
Since the fix we are purposing is for #11072 requires a code change from our partner to use the new prop anyway, I am okay with doing the revert by itself

chang47:
@xugao Thank you! Can we proceed to merging in the revert now? Thanks!

chang47:
@jdhuntington Can you merge this PR when you have the chance? Thanks!

jspurlin:
Looks like the element that had `role="tablist "` should be updated to `role="toolbar"`

chang47:
Change pushed, now we wait for everything to build.

JasonGore:
> Looks like the element that had `role="tablist "` should be updated to `role="toolbar"`

The [ARIA spec](https://www.w3.org/TR/wai-aria-1.1/#toolbar) doesn't seem to show any relation between the `toolbar` role and `tablist` / `tabpanel` roles. It does suggest some relation to `menubar`, which may or may not apply here. Why do you propose this change?

jspurlin:
I'm proposing the change because a Pivot has been represented as a toolbar, menubar doesn't really apply for to type of UI in my mind. Across office the ribbon, for example, is represented as a toolbar for all endpoints. If you use role="menubar" it needs to contain menuitem-type children which doesn't really fit with this type of UI in my mind. If we wanted to explore the possibility of using role=menubar by default, that should be a separate conversation.

A consumer can override the role if they so choose, and having a default role allows ATs to announce something better than "<label> group"

JasonGore:
Just to clarify, I wasn't trying to indicate that we should use `menubar` at all. I was just questioning the default application of `toolbar` role with `tab` roles, particularly in the absence of `menubar` usage.

PR #11692, opened 2020-01-13
Added prop to restrict the max length in TextField

### LINKED ISSUE #11691: TextField support for maximum length

Currently, users of the TextField component need to restrict the input length manually on their events (see Additional context/screenshots section). The TextField component should expose a prop to support this restriction so users can leverage this and not have to add their own support.

Additionally, this should work with clipboard. For example, if the maxLength is 3 characters, and the user copies "Lorem", the TextField input should still have its value set to "Lor".

TextField

Yes, discussed this with @chang47. 

See the example in this CodePen: https://codepen.io/katiearriagam/pen/oNgydjy?editable=true - the TextField labeled "Controlled TextField limiting length of value to 5" has the length restriction functionality on its `onChange` even. Additionally, it doesn't work with clipboard if the text exceeds the maximum length, so other users might accidentally introduce this bug when they use the TextField component.

katiearriagam:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11691 

#### Description of changes

Exposed the maxLength prop which users can leverage to restrict the length of the input in the TextField component. This works with manually typed inputs as well as text copied from the clipboard.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11692)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx
@@ -481,11 +481,17 @@ export class TextFieldBase extends React.Component<ITextFieldProps, ITextFieldSt
     // tests (ours and maybe consumers' too), so it seemed best to do the switch in a major bump.
 
     const element = event.target as HTMLInputElement;
-    const value = element.value;
+
     // Ignore this event if the value is undefined (in case one of the IE bugs comes back)
-    if (value === undefined || value === this._lastChangeValue) {
+    if (element.value === undefined || element.value === this._lastChangeValue) {
       return;
     }
+
+    const value =
+      this.props.maxLength !== undefined && element.value.length >= this.props.maxLength
+        ? element.value.substring(0, this.props.maxLength)
+        : element.value;
+
     this._lastChangeValue = value;
 
     // This is so developers can access the event properties in asynchronous callbacks
--- a/packages/office-ui-fabric-react/src/components/TextField/examples/TextField.Controlled.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/TextField/examples/TextField.Controlled.Example.tsx
@@ -23,8 +23,9 @@ export class TextFieldControlledExample extends React.Component<{}, ITextFieldCo
         <TextField
           label="Controlled TextField limiting length of value to 5"
           value={this.state.value2}
-          onChange={this._onChange2}
           styles={{ fieldGroup: { width: 100 } }}
+          maxLength={5}
+          onChange={this._onChange2}
         />
       </Stack>
     );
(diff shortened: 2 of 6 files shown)


xugao:
@katiearriagam - `maxLength` prop is available today since it's a `InputHTMLAttribute`: https://codepen.io/xugao/pen/abzKjoO

I think "Controlled TextField limiting length of value to 5" example we have is quite confusing, and I believe it's meant to demo using `onChange` prop to build **controlled** TextField. I think the next step here is simply making a better code example

packages/office-ui-fabric-react/src/components/TextField/examples/TextField.Controlled.Example.tsx
```diff
@@ -1,50 +1,41 @@
...
+        />
+      </Stack>
+    );
+  }
+
+  private _onChange1 = (ev: React.FormEvent<HTMLInputElement>, newValue?: string) => {
+    this.setState({ value1: newValue || '' });
+  };
+
+  private _onChange2 = (ev: React.FormEvent<HTMLInputElement>, newValue?: string) => {
```
ecraig12345: This example as it was previously implemented is important because it shows the pattern for using a controlled TextField to restrict input to valid values. Instead of switching it to use the maxLength prop, we should either:
a) switch the example to put a different restriction on the input, or 
b) add a comment above the second TextField indicating that this is just an example of how to restrict input using a controlled TextField, and in real code, you can more easily restrict the length using the `maxLength` prop.

It might also be good to add a `maxLength` example in TextField.Basic.Example.tsx.

packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx
```diff
@@ -1,579 +1,585 @@
...
-}
-
-/**
- * If `validateOnFocusIn` or `validateOnFocusOut` is true, validation should run **only** on that event.
- * Otherwise, validation should run on every change.
- */
-function _shouldValidateAllChanges(props: ITextFieldProps): boolean {
-  return !(props.validateOnFocusIn || props.validateOnFocusOut);
-}
+import * as React from 'react';
```
ecraig12345: It looks like you've somehow changed the check-in newlines in all the edited files, which needs to be reverted to avoid causing a huge amount of history noise.

How to do this varies by editor. In VS Code, towards the right of the bottom status bar, there's a little `CRLF` or `LF` indicator. Clicking that will give you the option to change the line endings. Change it to `LF` in each edited file, then save and commit.

To avoid this happening in the future, you should also configure `core.autocrlf` (either globally, or local to this repo if you prefer a different setting in general): on Windows `git config core.autocrlf true`, or on Mac `git config core.autocrlf input`, and add `-g` if you want to set globally. It's probably best to do this after fixing the existing newline issues.

PR #11702, opened 2020-01-14
ColorPicker: Fix click on color rectangle not working in Safari/Firefox browsers

### LINKED ISSUE #11680: Click by touch on touchscreens on color rectangle is not working in Safari/Firefox browsers.
### Environment Information

- **Package version(s)**: All
- **Browser and OS versions**: Safari, Firefox

#### Actual behavior:
For color picker component (https://developer.microsoft.com/en-us/fabric#/controls/web/colorpicker), Click by touch on touchscreens on color rectangle does not change the color in Safari/Firefox browsers, while it is working for Chrome/Edge Chromium browsers.

#### Expected behavior:
Click by touch on touch screens on color rectangle should change the color as this is doing in chrome/edge chromium browsers.

#### Expected fix:
This is because the way the handler (_onMouseMove ) for mouse event is implemented for color rectangle.There is below if condition in _onMouseMove handler code, where ev.buttons is coming as 1 (i.e. primary button usually left key is pressed) for touch event in browsers like Chrome/Edge Chromium whereas it is coming as 0 for Safari/Firefox browsers. 
It seems like this if condition is needed for debugging purpose and this if check should not be there.

Code:
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/ColorPicker/ColorRectangle/ColorRectangle.base.tsx
    // If the primary button (1) isn't pressed, the user is no longer dragging, so turn off the
    // event handlers and exit. (this may only be relevant while debugging)
    // tslint:disable-next-line:no-bitwise
    if (!(ev.buttons & 1)) {
      this._disableEvents();
      return;
    }

Requested priority: High

Products/sites affected: Microsoft Word/Powerpoint Online

bnslanuj:
#### Pull request checklist

- [x] Fixes #11680 

#### Description of changes

There was a check for ev.buttons in mouse event handler method, which i believe was added for debugging purpose. Also, according to the notes mentioned here ([MouseEvent.Buttons](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons)) ev.buttons always return 0 in case of safari browsers. So, removing that if condition.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11702)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ColorPicker/ColorRectangle/ColorRectangle.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ColorPicker/ColorRectangle/ColorRectangle.base.tsx
@@ -168,14 +168,6 @@ export class ColorRectangleBase extends React.Component<IColorRectangleProps, IC
       return;
     }
 
-    // If the primary button (1) isn't pressed, the user is no longer dragging, so turn off the
-    // event handlers and exit. (this may only be relevant while debugging)
-    // tslint:disable-next-line:no-bitwise
-    if (!(ev.buttons & 1)) {
-      this._disableEvents();
-      return;
-    }
-
     const newColor = _getNewColor(ev, this.state.color, this._root.current);
     if (newColor) {
       this._updateColor(ev, newColor);
--- a/packages/office-ui-fabric-react/src/components/ColorPicker/ColorRectangle/ColorRectangle.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ColorPicker/ColorRectangle/ColorRectangle.base.tsx
@@ -168,6 +168,15 @@ export class ColorRectangleBase extends React.Component<IColorRectangleProps, IC
       return;
     }
 
+    // Leaving the following commented code which is sometimes necessary for debugging:
+    // If the primary button (1) isn't pressed, the user is no longer dragging, so turn off
+    // the event handlers and exit.
+    // tslint:disable-next-line:no-bitwise
+    // if (!(ev.buttons & 1)) {
+    //   this._disableEvents();
+    //   return;
+    // }
+
     const newColor = _getNewColor(ev, this.state.color, this._root.current);
     if (newColor) {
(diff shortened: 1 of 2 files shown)


xugao:
Need @ecraig12345 or @cliffkoh to review this one. It's not clear to me why the condition was added in first place

bnslanuj:
> Need @ecraig12345 or @cliffkoh to review this one. It's not clear to me why the condition was added in first place

@ecraig12345 / @cliffkoh please review, when you get time.

cliffkoh:
I have never edited ColorPicker and am pretty sure I lack context on this change.. will have to defer to @ecraig12345

packages/office-ui-fabric-react/src/components/ColorPicker/ColorRectangle/ColorRectangle.base.tsx
```diff
@@ -168,14 +168,6 @@ export class ColorRectangleBase extends React.Component<IColorRectangleProps, IC
       return;
     }
 
-    // If the primary button (1) isn't pressed, the user is no longer dragging, so turn off the
```
ecraig12345: If you could leave this there but commented in case it's needed in the future for debugging purposes, that would potentially be helpful. Something like this:
```
    // Leaving the following commented code which is sometimes necessary for debugging:
    // If the primary button (1) isn't pressed, the user is no longer dragging, so turn off
    // the event handlers and exit.
    // tslint:disable-next-line:no-bitwise
    // if (!(ev.buttons & 1)) {
    //   this._disableEvents();
    //   return;
    // }
```

bnslanuj: Done!

PR #11703, opened 2020-01-14
Dialog/Panel: allow user to specify heading aria level

### LINKED ISSUE #11666: [Dialog][Panel][Accessibility] The heading is always set at level 2.
Environment Details:​​
Browser Versions:Chrome Version 78.0.3904.97 
Windows Version: Windows10​​​

Repro Steps:​​
1. Hit the application URL
2. Repeat the same steps as ( Select application as "Sales hub" from the list of application--> From the change area which is present in the bottom select "App Settings"-->Navigate to "Site Map" and then Navigate to "Goal Management" and select "Forecast Configuration")
3. Fast Track Your Forecast Screen will appears.
4. Click on "Org chart forecast"
5. Forecast Configuration screen will appears
6. Fill all the mandatory field and click on "next button".
7. Again click on next button.
8. Layout and columns screen will appears and then click on +Add from option set button and from the options select "Forecast category" and click on "Choose selected"
9. Auto Configure popup will appears
10.Launch accessibility insights for web and ad hoc tools and then select heading.

Actual:​
The heading for the pop-up is displayed as h2.

Same issue is observed in below screens:
1)Confirm delete popup
2)Delete the column.

Expected:​
The heading should be at level h1, to maintain the structure.

User Impact:​
Screen reader users will think there might be a hidden h1 heading that as not been announced , as well as the structure will not be defines well for the users.

1640412_A11y_D365ForecastConfiguration_Layoutandcolumns_Auto-configurecolumns_A14W

xugao:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #11666

#### Description of changes
For the header with `role=heading` in Dialog/Panel, we set `aria-level` to 2 and there is no way to override that for user. Adding `AriaLevel` prop to enable this. I don't like adding this type of `Aria` prop, because it doesn't feel ergonomic. But for the short/mid-term, this is more consistent with what we have else where.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11703)

### DIFF
--- a/change/office-ui-fabric-react-2020-01-13-19-08-10-xgao-dialog-heading-level.json
+++ b/change/office-ui-fabric-react-2020-01-13-19-08-10-xgao-dialog-heading-level.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Dialog: allow prop to set aria-level for title heading",
+  "packageName": "office-ui-fabric-react",
+  "email": "xgao@microsoft.com",
+  "commit": "bd5f04f01ecf5a16079e9765b9e4b462c2b34832",
+  "date": "2020-01-14T03:08:10.815Z"
+}
--- a/change/office-ui-fabric-react-2020-01-13-19-08-10-xgao-dialog-heading-level.json
+++ b/change/office-ui-fabric-react-2020-01-13-19-08-10-xgao-dialog-heading-level.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Dialog: allow prop to set aria-level for title heading",
+  "comment": "Dialog/Panel: allow prop to set aria-level for the heading",
   "packageName": "office-ui-fabric-react",
   "email": "xgao@microsoft.com",
   "commit": "bd5f04f01ecf5a16079e9765b9e4b462c2b34832",
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -3759,6 +3759,7 @@ export interface IDialogContentProps extends React.ClassAttributes<DialogContent
     subTextId?: string;
     theme?: ITheme;
     title?: string | JSX.Element;
+    titleAriaLevel?: number;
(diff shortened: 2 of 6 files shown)


JasonGore:
Hmm, did you consider adding something like `titleProps` that would allow modification of not just ARIA attributes, but also the `role` and native attributes as well?

I question the existing design in general. I wonder if it'd be better to have a render header prop.

xugao:
@JasonGore,
On option to add `titleProps` which allow user to set any valid html attributes - The cons of this are:
1) It opens the interface to accept way more props than compared to just enable adding `aria-level`. No knowing for sure what the plan is for FUR, i hesitate to change our existing prop interface to allow more than it currently needs/
2) I looked through all the components, currently we don't have a single component that allows `blahProps` which simply extend `HTMLAttributes`. I think consistency is more important here.

On the option to add render prop:
It would become an option for user to pass the attribute, and possibly a lot more other things. but I simply don't feel render prop is the right way to enable this specific scenario. thoughts?

JasonGore:
I actually thought this was a similar approach our discussion of add `tablistProps` ala #11690. I'm curious what the differences are here.

My primary concern is that this is a half-step for a particular use case and that later we'll find another reason to replace the `role` on heading or allow even more attributes to be applied to it, which will either grow the top level props further or then create a need for `titleProps`, at which point we will have redundant and obsolete top-level props.

xugao:
@JasonGore -  I actually changed my mind since our discussion, to add `blahAriaLabel` instead of `blahProps`. reasons are being the same as i listed above. On your concern, i actually could not think of a reason the role not being `heading`, that's why i went with adding AriaLevel prop in first place.

With this being said, i'm not totally against adding `blahProps` for the reasons you said. if we do that, i think we should not extend HTMLAttributes, rather we start from cherry-pick the must-need ones.

xugao:
discussed with @JasonGore further offline (thanks Jason!) Will close this PR and issue a new one to add a single (grouped) `blahProps` prop as the approach

PR #11706, opened 2020-01-14
Updates persona primary text default line height

helizabethwhite:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes https://github.com/OfficeDev/office-ui-fabric-react/issues/11719

#### Description of changes

Text is being clipped in certain Persona scenarios.

Screenshot:

image

_(This was grabbed from the Search suggestions list in the latest version of OWA.)_

Changes:
The current Persona styles enforce a line-height on the `primaryText` that can be too small depending on the text's font-size and height. Rather than upping the `const` value to a larger number, it seems that removing the use of an explicit value and instead defaulting to `line-height: normal` is a more lasting fix to the bug.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11706)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Persona/Persona.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Persona/Persona.styles.ts
@@ -224,7 +224,6 @@ export const getStyles = (props: IPersonaStyleProps): IPersonaStyles => {
 
       showSecondaryText && {
         height: showSecondaryTextDefaultHeight,
-        lineHeight: showSecondaryTextDefaultHeight,
         overflowX: 'hidden'
       },
 
@@ -264,7 +263,6 @@ export const getStyles = (props: IPersonaStyleProps): IPersonaStyles => {
       showSecondaryText && {
         display: 'block',
         height: showSecondaryTextDefaultHeight,
-        lineHeight: showSecondaryTextDefaultHeight,
         overflowX: 'hidden'
       },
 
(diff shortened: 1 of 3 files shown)


helizabethwhite:
Aware of the test failures and will update shortly.

philipzloh:
Enforcing a `line-height` was bad practice, good to remove. Explicitly-specified `line-height` tends to be smaller than what browsers would naturally use, so chances are very high that clipping will be observed (as mentioned).

packages/office-ui-fabric-react/src/components/Persona/Persona.styles.ts
```diff
@@ -224,7 +224,6 @@ export const getStyles = (props: IPersonaStyleProps): IPersonaStyles => {
 
       showSecondaryText && {
         height: showSecondaryTextDefaultHeight,
-        lineHeight: showSecondaryTextDefaultHeight,
```
xugao: removing `lineHeight` will affect the vertical alignment of the text

xugao: I don't repro the issue in our example page https://developer.microsoft.com/en-us/fabric#/controls/web/persona, which i believe uses default theme. in the case you have the repro, what are the `font-size`, `height` and `lineHeight`?

helizabethwhite: hey @xugao, thanks for pointing out that it isn't reproducible on the demo page. that led me to do some more digging and while i think this change wouldn't hurt (particularly for the comment left by @philipzloh), i'll close this PR as the OWA-specific problem can be fixed within OWA code.

PR #11716, opened 2020-01-15
Revert "ContextualMenu: Fix disabled item being focusable"

xugao:
Reverts OfficeDev/office-ui-fabric-react#11693

Discussed with @jspurlin , we should keep disabled item focusable. Dropdown and ComboBox should be the same. For context, here are some conversations I had offline with @jspurlin :

[9:16 AM] Jeremy Spurlin
    There's a been a lot of discussion around this in the past with internal/external accessible users and fabric stakeholders. If the items are not focusable, a blind user potentially will never know that the control is there in the first place. Additionally, if disabled items are not focusable, it breaks the muscle memory that accessible users (and power users for that matter) build for executing controls
[9:18 AM] Jeremy Spurlin
    For example, users learn to press the downArrow key x times when this menu is open to execute a specific control, if that changes because an item above it is disabled then they will potentially accidentally execute an incorrect control
[9:20 AM] Jeremy Spurlin
    There was an agreement that they should be focusable, but the existing behavior was to not have them focusable which aligns with traditional HTML behavior. Note, traditional HTML behavior was created for much simpler HTML pages and does not align with web application behaviors (which is to focus disabled controls). Fabric added the ability to have disabled items be focusable
​[9:20 AM] Jeremy Spurlin
    in general and office online uses that
​[9:21 AM] Jeremy Spurlin
    for menus, there's no native HTML behavior and it's unique to modern web applications and hence we decided to make those controls focusable by default when disabled

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11716)

### DIFF
--- a/change/office-ui-fabric-react-2020-01-13-15-11-26-xgao-fix-contextual-menu-keyboard.json
+++ b/change/office-ui-fabric-react-2020-01-13-15-11-26-xgao-fix-contextual-menu-keyboard.json
@@ -1,8 +1,0 @@
-{
-  "type": "patch",
-  "comment": "ContextualMenu: Fix disabled item being focusable",
-  "packageName": "office-ui-fabric-react",
-  "email": "xgao@microsoft.com",
-  "commit": "e07cba4de908eb62656cd8263f2a68782e73f59e",
-  "date": "2020-01-13T23:11:26.708Z"
-}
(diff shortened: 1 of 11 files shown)


xugao:
wrong repo to push this revert from. here is the correct PR: #11718

PR #11722, opened 2020-01-15
hide checkmark if canCheck is false

yumikohey:
#### Pull request checklist

- [x ] Addresses an existing issue: Fixes #0000
- [x ] Include a change request file using `$ yarn change`

#### Description of changes
1. Update `renderCheckMarkIcon` to provide flexibility to hide checkmark for toggle contextual command with `canCheck` prop.
2. Add an example under toggleable contextualMenu
a) When command is not toggled
Screen Shot 2020-01-14 at 7 02 29 PM

b) When command is toggled
Screen Shot 2020-01-14 at 7 02 39 PM

#### Focus areas to test
Contextual menu

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11722)

### DIFF
--- a/change/office-ui-fabric-react-2020-01-15-13-12-22-huaxi-customize-checkmark.json
+++ b/change/office-ui-fabric-react-2020-01-15-13-12-22-huaxi-customize-checkmark.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "hide checkmark if canCheck is false",
+  "packageName": "office-ui-fabric-react",
+  "email": "huaxi@microsoft.com",
+  "commit": "7a3f3bd39492f60ed9edfd9b8c8e51ad4ec891fa",
+  "date": "2020-01-15T21:12:22.849Z"
+}
--- a/change/office-ui-fabric-react-2020-01-15-13-12-22-huaxi-customize-checkmark.json
+++ b/change/office-ui-fabric-react-2020-01-15-13-12-22-huaxi-customize-checkmark.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "hide checkmark if canCheck is false",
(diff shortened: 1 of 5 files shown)


packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenuItem.base.tsx
```diff
@@ -26,7 +26,7 @@ const renderCheckMarkIcon = ({ onCheckmarkClick, item, classNames }: IContextual
     // Ensures that the item is passed as the first argument to the checkmark click callback.
     const onClick = (e: React.MouseEvent<HTMLElement>) => onCheckmarkClick(item, e);
 
-    return <Icon iconName={isItemChecked ? 'CheckMark' : ''} className={classNames.checkmarkIcon} onClick={onClick} />;
+    return <Icon iconName={!!item.canCheck && isItemChecked ? 'CheckMark' : ''} className={classNames.checkmarkIcon} onClick={onClick} />;
```
joschect: To preserve current implementation, I think that it should only if item.canCheck === false. Currently if it's undefined for one item, then it will be visible. That way we don't break people who are relying on check being visible in the current case where canCheck only needs to be true for one item.

change/office-ui-fabric-react-2020-01-15-13-12-22-huaxi-customize-checkmark.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "hide checkmark if canCheck is false",
```
khmakoto:
```suggestion
  "comment": "ContextualMenu: Hide checkmark in ContextualMenuItems if canCheck is false.",
```

PR #11677, opened 2020-01-10
TypeScript 3.7 Upgrade

JasonGore:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [ ] Include a change request file using `$ yarn change`

#### Description of changes

This PR helps prep the Fabric repo for repo consolidation and [TypeScript project references](https://www.typescriptlang.org/docs/handbook/project-references.html) by upgrading to TS3.7.

I've created this PR so it can be reviewed, but I am creating it in draft state due to a potential blocker described below that needs to be mitigated. (Kevin also offered to help run tests against this PR. Thanks Kevin!)

#### Code Changes
  * Add TS postprocessing to maintain TS3.3 minbar compatibility.
    * API readonly -> get/set changes: This is a [breaking change for TS3.5 and earlier](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#class-field-mitigations).
  * Fixed improper uses of `assign` with some basic unit tests to make sure they don't resurface.
    * TS3.6 [changed the way spreads were emitted in TS3.6](https://www.typescriptlang.org/docs/handbook/release-notes/overview.html#more-accurate-array-spread), and these improper uses of `assign` were causing `RangeError: Invalid array length` runtime exceptions with tslib's `__spreadArrays` helper.
  * API Extractor dependencies needed to be updated to fix type mismatches.
  * Added use of `!!` to indicate intentional test of function presence per [Uncalled Function Checks](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#uncalled-function-checks).
  * Casts to `DOMRect` per [DOM Correctness Changes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#dom-changes).

#### New TS3.7 Features

I wanted to call out some new TypeScript features that we may be able to use going forward:
  * [Optional Chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)
  * [Nullish Coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing)
  * [(More) Recursive Type Aliases](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#more-recursive-type-aliases)

[...]

PR #11723, opened 2020-01-15
Website: don't make routes for page categories

ecraig12345:
#### Pull request checklist

#### Description of changes

Fix some bugs around routing and page categories, discovered in #11677:
- Previously in `createSite` we were creating routes for categories of pages. This doesn't make sense because the category itself doesn't actually have anything to render. So stop doing that.
- Router should not use a "route" which has neither `component` nor `getComponent` defined (since there's nothing to render in that case)
- `INavPage.url` should be optional to reflect that categories don't have a URL. (and remove the `url` from all categories)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11723)


packages/example-app-base/src/utilities/router/Router.tsx
```diff
@@ -114,7 +116,7 @@ export class Router extends React.Component<IRouterProps, IRouterState> {
         if (component) {
           const componentChildren = this._resolveRoute(route.props.children || []);
           return React.createElement(component, { key: route.key! }, componentChildren);
-        } else if (getComponent) {
+        } else if (!!getComponent) {
```
JasonGore: You could probably revert this and save yourself a merge conflict since I have to make the change in #11677 .

ecraig12345: I think it will "just work" since this is the same change

PR #11724, opened 2020-01-15
MessageBar: Fix High Contrast Mode issues

### LINKED ISSUE #10880: Links in Message Bar do not respond to High Contrast
This reproes in all High Contrast themes
image

You can repro at https://aka.ms/fabricdemo
Change to any high contrast theme, and go to MessageBar. Turn on one of them with a link and observe.

The color should use link color for the theme.

Requested priority: (High)

Products/sites affected: OneDrive, SharePoint, Outlook, various Azure properties

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10880

#### Description of changes

This PR overhauls the entire experience when using `MessageBars` in High Contrast Mode.

__Before:__
image

__After:__
image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11724)


ecraig12345:
@betrue-final-final Mak and I noticed that too, but we were having a terrible time figuring out how to override the default darker Link-in-MessageBar color, because there's not an actual system color name for the link color. :( Underlining could be a good alternative.

ecraig12345:
I'm also concerned that the default system link color may not have enough contrast against some of our custom backgrounds if they've customized their color scheme (I know @betrue-final-final  checked it on the default color schemes), which might be another argument for underlining.

khmakoto:
@betrue-final-final I was telling Elizabeth about a really weird thing that is happening where, if we use `inherit` to unset the color on High Contrast Mode on the stylesheet, then the browser unsets it to `WindowText`. However, if we then apply the same `inherit` color directly on the browser, it takes the correct `Link` color. My suspicion is that this is a browser related issue where unsetting the color from the stylesheet is being done before the page render, so the actual `Link` color is not being applied when High Contrast Mode is turned on, whereas it is correct if we unset the color later.

Setting the `color` to `inherit` on the stylesheet:
image

Setting the `color` to `inherit` on the page after everything renders:
image

khmakoto:
@ecraig12345's point about customized high contrast themes is also very valid, maybe the better idea after all would be to just have `Links` have `WindowText` color with underlined text decoration. What do you think @betrue-final-final?

betrue-final-final:
I think underlining is probably our best option.

khmakoto:
@betrue-final-final @ecraig12345 alright, I went for the 'underlining' option and now all checks have passed and links look distinguishable in high contrast mode.

packages/office-ui-fabric-react/src/components/Link/Link.styles.ts
```diff
@@ -92,6 +92,16 @@ export const getStyles = (props: ILinkStyleProps): ILinkStyles => {
           },
           '&:focus': {
             color: linkColor
+          },
+          [HighContrastSelector]: {
```
ecraig12345: I'm not sure if this is something we want for all links?

packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
```diff
@@ -49,6 +49,20 @@ const getRootBackground = (messageBarType: MessageBarType | undefined, palette:
   }
   return palette.neutralLighter;
 };
+// Returns the background color of the MessageBar root element based on the type of MessageBar when in High Contrast Mode.
```
ecraig12345: Maybe add a note that not varying based on the theme is intentional.

packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
```diff
@@ -134,11 +148,18 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
         selectors: {
           '& .ms-Link': {
             color: palette.themeDark,
-            ...fonts.small
+            ...fonts.small,
+            selectors: {
+              [HighContrastSelector]: {
```
ecraig12345: Probably add a note about why it's inheriting the color here

xugao:
@khmakoto - setting `inherit` seems wrong to me. since it's behaving correctly in https://developer.microsoft.com/en-us/fabric#/controls/web/link , we should understand what made browser to skip adjusting the link color in the message bar. did we set `-ms-high-contrast-adjust: none` somewhere on an ancestor node of the link? i think the fix here might be set `-ms-high-contrast-adjust: auto` on the link in message bar

khmakoto:
This might actually work @xugao! I'll try it out.

khmakoto:
@xugao that did the trick! Will push updates shortly.

khmakoto:
@ecraig12345 @xugao just updated the PR with the latest feedback.

change/@uifabric-example-app-base-2020-01-15-14-08-56-messageBarHighContrastLinks.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "MessageBar: Fix High Contrast Mode issues.",
+  "packageName": "@uifabric/example-app-base",
```
xugao: i dont think there is any change in this package?

khmakoto: There was before the latest changes, but afterwards there are not anymore.

PR #11725, opened 2020-01-16
Use standard postprocess task for monaco-editor and fix TS compat issue

ecraig12345:
#### Pull request checklist

#### Description of changes

Use @JasonGore's new postprocess task for removing get/set from .d.ts files in monaco-editor.

The custom transformDtsTask is still needed to deal with a different issue. I also added basic handling for removing `readonly` modifiers from parameters, because I just learned today that this breaks TS 3.4. (This could also be added to the main postprocess task, but that would require figuring out how to do it with the TS API...)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11725)

### DIFF
--- a/packages/monaco-editor/just.config.js
+++ b/packages/monaco-editor/just.config.js
@@ -2,6 +2,7 @@
 const { just, preset } = require('@uifabric/build');
 const { task, series, parallel, copyInstructions, copyInstructionsTask, cleanTask } = just;
 const { ts } = require('@uifabric/build/tasks/ts');
+const { postprocessTask, defaultLibPaths } = require('@uifabric/build/tasks/postprocess');
 const path = require('path');
 const { transformCssTask } = require('./tasks/transformCssTask');
 const { transformDtsTask } = require('./tasks/transformDtsTask');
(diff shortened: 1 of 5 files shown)


packages/monaco-editor/just.config.js
```diff
@@ -23,6 +24,7 @@ task('transform-css', transformCssTask);
 task('transform-dts', transformDtsTask);
 task('ts:esm', ts.esm);
 task('ts:commonjs', ts.commonjs);
-task('ts', parallel('ts:esm', 'ts:commonjs'));
+task('ts:postprocess', postprocessTask([...defaultLibPaths, 'esm/**/*.d.ts']));
+task('ts', series(parallel('ts:esm', 'ts:commonjs'), 'ts:postprocess'));
```
JasonGore: It'd be nice if we could just redefine `ts:compile` and let the preset `ts` use the modified `ts:postprocess`. However, that won't work here since `ts:postprocess` is also redefined and the preset `ts` won't pick it up. If the task names instead worked as references somehow (scoped to a project), then projects could redefine them and still have those new definitions apply in the preset tasks.

JasonGore:
What usage of `readonly` breaks TS3.4? When testing #11677 I tested against TS3.3 and didn't find any issues... we've had `readonly` in our API for a while with a TS3.3 minbar.

packages/monaco-editor/just.config.js
ecraig12345: In this case I'm not inheriting from the basic config at all, so the task inheritance isn't so relevant, but I agree in general. (Also, this is why I wanted extensible task configs @kenotron. 😉 )

ecraig12345:
> What usage of `readonly` breaks TS3.4? When testing #11677 I tested against TS3.3 and didn't find any issues... we've had `readonly` in our API for a while with a TS3.3 minbar.

This from monaco-editor. `setSelections(selections: readonly string): void;`
http://www.typescriptlang.org/play/index.html?ts=3.3.3#code/JYOwLgpgTgZghgYwgAgJIDED2nkG8BQyRyAzhGAMoQA2ECYwmIJAFGbfY8wFzJQRwAJk2oBPUmCigA5gEpeAN0zBBAbnwBfIA

PR #11734, opened 2020-01-16
Forward supportedPackages to TsxEditor

jimkyndemeyer:
#### Description of changes

`<EditorWrapper>` exposes a `supportedPackages` prop, but doesn't pass that along to the wrapped `<TsxEditor>`.

This PR forwards the missing prop to `<TsxEditor>` such that additional packages can be used in the editor.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11734)

### DIFF
--- a/change/@uifabric-tsx-editor-2020-01-17-10-51-44-patch-1.json
+++ b/change/@uifabric-tsx-editor-2020-01-17-10-51-44-patch-1.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Forward supportedPackages from EditorWrapper to TsxEditor",
+  "packageName": "@uifabric/tsx-editor",
+  "email": "jimkyndemeyer@gmail.com",
+  "commit": "98ecc14cae342612339373063ce7d0ab078f20c0",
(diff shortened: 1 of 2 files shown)


ecraig12345:
@jimkyndemeyer The change should be a patch since it's not modifying any public APIs (just changing how an existing API is called).

If you run into any other issues using the editor or have questions, feel free to file an issue and tag me. It's cool to see someone else is actually using this! Only caveat is that there's still a possibility of breaking changes (for a 0.x package this is allowed in a minor release)--probably no huge changes at this point, but there could be some refinement of props if more people start using this and it turns out that the existing API doesn't work well, or I forgot something that's needed to make it work outside of Fabric.

jimkyndemeyer:
@ecraig12345 Thanks for the quick review. With the latest `yarn change` commit, I assume this is good to go.

I saw the code editor on the Fabric website, and wanted to have something similar for our design system, and the EditorWrapper component turned out to be super useful, so thanks for making it :)

PR #11736, opened 2020-01-16
ContextualMenu: Transforming examples to Function Components and cleaning them up

### LINKED ISSUE #9535: Clean up Contextual Menu examples
Examples in ContextualMenu. possibly other components as well, can benefit from cleanup.

Specifically, I see

1. Unused state variables, like for example, [this](https://github.com/OfficeDev/office-ui-fabric-react/blob/659b57cc432a37173324eb06f22ea7a918ecff36/packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Basic.Example.tsx#L10) one
2. Unnecessary class components
3. Unused css class names

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #9535

#### Description of changes

This PR transforms all `ContextualMenu` example to be Function Components and cleans them up.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11736)


packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Checkmarks.Example.tsx
```diff
@@ -25,188 +24,176 @@ const keys: string[] = [
...
+  };
 
-  public render(): JSX.Element {
-    const { selection } = this.state;
+  return (
+    <DefaultButton
+      text="Click for ContextualMenu"
+      menuProps={{
+        shouldFocusOnMount: true,
+        items: [
```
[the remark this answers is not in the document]
khmakoto: I'll move them out of the return statement, and out of the render function if possible, but in this specific case there is state logic inside the items, so I can't move them out of the render function.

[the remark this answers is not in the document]
khmakoto: @ecraig12345 I've moved the item definitions to not be inline 😄

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Checkmarks.Example.tsx
```diff
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { IContextualMenuItem, ContextualMenuItemType, DirectionalHint } from 'office-ui-fabric-react/lib/ContextualMenu';
 import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
-import './ContextualMenuExample.scss';
 
 export interface IContextualMenuMultiselectExampleState {
```
ecraig12345: You can get rid of this now

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Checkmarks.Example.tsx
```diff
@@ -25,198 +24,188 @@ const keys: string[] = [
...
+    if (item) {
+      console.log(selection);
+      setSelection({ ...selection, ...{ [item.key]: selection[item.key] === undefined ? true : !selection[item.key] } });
+      console.log(selection);
+    }
+  };
 
-  public render(): JSX.Element {
-    const { selection } = this.state;
+  const menuItems: IContextualMenuItem[] = [
```
ecraig12345: Even though this is no longer inline it will still mutate every render, so you should really do useMemo. `useMemo(() => /*items*/, [selection, onToggleSelect])` (pretty sure those are the only two dependencies which can change)

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Checkmarks.Example.tsx
```diff
@@ -25,198 +24,188 @@ const keys: string[] = [
...
 ];
 
-export class ContextualMenuCheckmarksExample extends React.Component<{}, IContextualMenuMultiselectExampleState> {
-  constructor(props: {}) {
-    super(props);
+export const ContextualMenuCheckmarksExample: React.FunctionComponent = () => {
+  const [selection, setSelection] = React.useState<{ [key: string]: boolean }>({});
 
-    this._onToggleSelect = this._onToggleSelect.bind(this);
+  const onToggleSelect = (ev?: React.MouseEvent<HTMLButtonElement>, item?: IContextualMenuItem): void => {
```
ecraig12345: You should probably use useCallback here. `useCallback(/*callback*/, [selection])`

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Checkmarks.Example.tsx
```diff
@@ -25,198 +24,188 @@ const keys: string[] = [
...
+    ev && ev.preventDefault();
 
-    this.state = {
-      selection: {}
-    };
-  }
+    if (item) {
+      console.log(selection);
+      setSelection({ ...selection, ...{ [item.key]: selection[item.key] === undefined ? true : !selection[item.key] } });
```
ecraig12345: The second spread isn't really needed here.
```suggestion
      setSelection({ ...selection, [item.key]: selection[item.key] === undefined ? true : !selection[item.key] });
```

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Checkmarks.Example.tsx
```diff
@@ -25,198 +24,188 @@ const keys: string[] = [
...
-
-    this.setState({
-      selection: selection
-    });
-  }
-}
+  return (
+    <DefaultButton
+      text="Click for ContextualMenu"
+      menuProps={{
```
ecraig12345: If we're being super pedantic (which it sounded like we wanted to in examples last time it was discussed), this shouldn't be defined inline either--basically **no** inline object or array definitions, since even though it's not likely to cause problems in examples, we want to demonstrate good patterns that real apps should use. Instead define above with `useMemo(() => /*obj*/, [menuItems])`

ecraig12345:
Caveat to my comment above is that there is some discussion about whether it's *actually* good to do useMemo/useCallback all the time or whether that actually makes perf worse (due to the overhead of the hooks)...but I brought this up in the conversation about example patterns and IIRC the conclusion was that we still want to use them.

khmakoto:
@ecraig12345 made those changes as well 👍

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.CustomMenuItem.Example.tsx
```diff
@@ -1,55 +1,46 @@
...
-    this.state = {
-      showCallout: false
-    };
-  }
+export const ContextualMenuWithCustomMenuItemExample: React.FunctionComponent = () => {
+  return (
+    <div>
+      <DefaultButton
+        text="Click for ContextualMenu"
+        menuProps={{
```
ecraig12345: Again being pedantic, you should probably do `useMemo` for these props. (same in other files)

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.CustomMenuItem.Example.tsx
```diff
@@ -1,55 +1,46 @@
...
-export class ContextualMenuWithCustomMenuItemExample extends React.Component {
-  constructor(props: {}) {
-    super(props);
-    this.state = {
-      showCallout: false
-    };
-  }
+export const ContextualMenuWithCustomMenuItemExample: React.FunctionComponent = () => {
+  return (
+    <div>
```
ecraig12345: Is this div doing anything? (also check in other files)

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.CustomMenuList.Example.tsx
```diff
@@ -2,11 +2,78 @@ import * as React from 'react';
...
 import { Icon } from 'office-ui-fabric-react/lib/Icon';
-import './ContextualMenuExample.scss';
 import { IContextualMenuListProps, IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
 import { IRenderFunction } from 'office-ui-fabric-react/lib/Utilities';
 
-const ITEMS: IContextualMenuItem[] = [
+export const ContextualMenuWithCustomMenuListExample: React.FunctionComponent = () => {
+  const [items, setItems] = React.useState(menuItems);
+
+  const onAbort = () => {
```
ecraig12345: Should probably wrap `onAbort` and `onChange` in `useConstCallback` from `@uifabric/react-hooks` to prevent the callback from mutating.

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.CustomMenuList.Example.tsx
```diff
@@ -2,11 +2,78 @@ import * as React from 'react';
...
+      });
+    }
+
+    setItems(filteredItems);
+  };
+
+  const renderMenuList = (menuListProps: IContextualMenuListProps, defaultRender: IRenderFunction<IContextualMenuListProps>) => {
+    return (
+      <div>
+        <div style={{ borderBottom: '1px solid #ccc' }}>
```
ecraig12345: Again being really pedantic, this `style` and the `styles` below should be in consts at the top or bottom of the file.

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Customization.Example.tsx
```diff
@@ -1,219 +1,218 @@
...
+    <IconButton
+      iconProps={{ iconName: item.name }}
+      className="ms-ContextualMenu-customizationExample-icon ms-ContextualMenu-link"
+      data-is-focusable={true}
+      onClick={dismissMenu}
+    />
+  );
+}
+
+function renderCategoriesList(item: any): JSX.Element {
```
ecraig12345: What's the type of `item`(here and elsewhere in the file, and a couple others)?

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Customization.Example.tsx
```diff
@@ -1,219 +1,218 @@
 import * as React from 'react';
-import { DirectionalHint, ContextualMenuItemType } from 'office-ui-fabric-react/lib/ContextualMenu';
+import { ContextualMenuItemType, DirectionalHint, IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
 import { DefaultButton, IconButton } from 'office-ui-fabric-react/lib/Button';
 import { FocusZoneDirection } from 'office-ui-fabric-react/lib/FocusZone';
```
ecraig12345: Unless you're replacing the styling with CSS-in-JS (which would be nice but not required right now), you need to keep the scss import.

Same in any other example that uses classnames.

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Directional.Example.tsx
```diff
@@ -33,110 +22,86 @@ const DIRECTION_OPTIONS = [
...
+  const [isBeakVisible, setIsBeakVisible] = React.useState(false);
+  const [useDirectionalHintForRTL, setUseDirectionalHintForRTL] = React.useState(false);
+  const [directionalHint, setDirectionalHint] = React.useState<DirectionalHint>(DirectionalHint.bottomLeftEdge);
+  const [directionalHintForRTL, setDirectionalHintForRTL] = React.useState<DirectionalHint>(DirectionalHint.bottomLeftEdge);
 
-  private _onShowBeakChange = (event: React.FormEvent<HTMLElement>, isVisible: boolean): void => {
-    this.setState({
-      isBeakVisible: isVisible
-    });
+  const onShowBeakChange = (event: React.FormEvent<HTMLElement>, isVisible: boolean): void => {
```
ecraig12345: These should probably use `useConstCallback` too.

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Icon.Example.tsx
```diff
@@ -8,99 +8,92 @@ import * as stylesImport from './ContextualMenuExample.scss';
...
-    this.state = {
-      showCallout: false
-    };
-  }
+export const ContextualMenuIconExample: React.FunctionComponent = () => {
+  const [showCallout, setShowCallout] = React.useState(false);
 
-  public render(): JSX.Element {
-    const { showCallout } = this.state;
+  const menuItems: IContextualMenuItem[] = [
```
ecraig12345: For this you can use `useConst` from `@uifabric/react-hooks`

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Icon.Example.tsx
```diff
@@ -8,99 +8,92 @@ import * as stylesImport from './ContextualMenuExample.scss';
...
+        menuProps={{
+          shouldFocusOnMount: true,
+          items: menuItems
+        }}
+      />
+      {showCallout && (
+        <Callout
+          setInitialFocus={true}
+          // tslint:disable-next-line:jsx-no-lambda
+          onDismiss={() => setShowCallout(false)}
```
ecraig12345: Can you get rid of these lambdas? (`useConstCallback`)

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Submenu.Example.tsx
```diff
@@ -1,143 +1,135 @@
...
+  };
+
+  return (
+    <div>
+      <TextField
+        value={String(hoverDelay)}
+        label="Hover delay (ms)"
+        type="number"
+        onChange={onHoverDelayChanged}
+        styles={{
```
ecraig12345: Ideally should be a const too

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Submenu.Example.tsx
```diff
@@ -1,143 +1,135 @@
...
-  constructor(props: any) {
-    super(props);
+export const ContextualMenuSubmenuExample: React.FunctionComponent = () => {
+  const [hoverDelay, setHoverDelay] = React.useState(250);
 
-    this.state = {
-      hoverDelay: 250
-    };
-  }
+  const onHoverDelayChanged = (ev: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, newValue: string) => {
```
ecraig12345: useConstCallback

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Customization.Example.tsx
khmakoto: It is really `any` in the types file but doesn't matter if we change it to be `IContextualMenuItem` here. I'm also enforcing what category list is by creating an interface in the file.

ecraig12345: Oh okay. In that case I'm fine with either typed or any.

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.CustomMenuList.Example.tsx
```diff
@@ -1,12 +1,79 @@
...
 import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
-import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
+import { ISearchBoxStyles, SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
 import { Icon } from 'office-ui-fabric-react/lib/Icon';
-import './ContextualMenuExample.scss';
 import { IContextualMenuListProps, IContextualMenuItem } from 'office-ui-fabric-react/lib/ContextualMenu';
 import { IRenderFunction } from 'office-ui-fabric-react/lib/Utilities';
 
-const ITEMS: IContextualMenuItem[] = [
+const wrapperStyle: React.CSSProperties = { borderBottom: '1px solid #ccc' };
```
ecraig12345: Most often what we've done for these styles instead of `React.CSSProperties` is `mergeStyleSets` with a class name for each thing, which I think was to try and discourage people from using the native `style` prop? But I'm also not sure how solid the reasoning behind that is. (I've also started usually putting styles below the main example since they're usually not the main point of the example.)

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.CustomMenuList.Example.tsx
```diff
@@ -1,12 +1,79 @@
...
+            <span>No actions found</span>
+          </div>
+        )
+      });
+    }
+
+    setItems(filteredItems);
+  });
+
+  const renderMenuList = (menuListProps: IContextualMenuListProps, defaultRender: IRenderFunction<IContextualMenuListProps>) => {
```
ecraig12345: useConstCallback

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.CustomMenuList.Example.tsx
```diff
@@ -1,12 +1,79 @@
...
+        </div>
+        {defaultRender(menuListProps)}
+      </div>
+    );
+  };
+
+  return (
+    <DefaultButton
+      text="Click for ContextualMenu"
+      menuProps={{
```
ecraig12345: useMemo

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Customization.Example.tsx
```diff
@@ -1,219 +1,224 @@
...
-          directionalHint: DirectionalHint.bottomLeftEdge,
-          className: 'ms-ContextualMenu-customizationExample',
-          items: [
+export const ContextualMenuCustomizationExample: React.FunctionComponent = () => {
+  return (
+    <DefaultButton
+      className="ContextualMenuButton3"
+      text="Click for ContextualMenu"
+      menuProps={{
+        shouldFocusOnMount: true,
```
ecraig12345: useConst for these menu props

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.CustomizationWithNoWrap.Example.tsx
```diff
@@ -1,230 +1,230 @@
...
+  return (
+    <DefaultButton
+      className="ContextualMenuButton3"
+      text="Click for ContextualMenu"
+      menuProps={{
+        shouldFocusOnMount: true,
+        directionalHint: DirectionalHint.bottomLeftEdge,
+        className: 'ms-ContextualMenu-customizationExample',
+        items: menuItems
+      }}
```
ecraig12345: useConst for menuProps

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Default.Example.tsx
```diff
@@ -1,72 +1,69 @@
...
-            ]
-          }}
-        />
-      </div>
-    );
+export const ContextualMenuDefaultExample: React.FunctionComponent = () => {
+  return (
+    <DefaultButton
+      text="Click for ContextualMenu"
+      menuProps={{
```
ecraig12345: useConst

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Directional.Example.tsx
```diff
@@ -33,110 +24,96 @@ const DIRECTION_OPTIONS = [
...
-
-  private _onDirectionalChanged = (event: React.FormEvent<HTMLDivElement>, option: IDropdownOption): void => {
-    this.setState({
-      directionalHint: option.key as DirectionalHint
-    });
-  };
+      <div className="ms-ContextualMenuDirectionalExample-buttonArea">
+        <DefaultButton
+          text="Show context menu"
+          menuProps={{
```
ecraig12345: `useMemo(/*stuff*/, [isBeakVisible, directionalHint, useDirectionalHintForRTL])`

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Header.Example.tsx
```diff
@@ -1,94 +1,93 @@
...
-              iconProps: {
-                iconName: 'Print'
-              },
-              text: 'Print'
-            },
+export const ContextualMenuHeaderExample: React.FunctionComponent = () => {
+  return (
+    <DefaultButton
+      text="Click for ContextualMenu"
+      menuProps={{
```
ecraig12345: useConst

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Icon.Example.tsx
```diff
@@ -1,106 +1,93 @@
...
-        )}
-      </div>
-    );
-  }
-}
+  return (
+    <div>
+      <DefaultButton
+        text="Click for ContextualMenu"
+        menuProps={{
```
ecraig12345: useConst (in this case you could actually get rid of the separate const for the menu items, and maybe even onShowCallout, since they're only used within the menu props)

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Icon.SecondaryText.Example.tsx
```diff
@@ -1,58 +1,50 @@
...
-    super(props);
-    this.state = {
-      showCallout: false
-    };
-  }
+export const ContextualMenuIconSecondaryTextExample: React.FunctionComponent = () => {
+  return (
+    <DefaultButton
+      text="Click for ContextualMenu"
+      menuProps={{
```
ecraig12345: useConst

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Persisted.Example.tsx
```diff
@@ -1,96 +1,86 @@
...
-    this.state = {
-      showCallout: false
-    };
-  }
+export const ContextualMenuPersistedExample: React.FunctionComponent = () => {
+  return (
+    <DefaultButton
+      text="Click for ContextualMenu"
+      persistMenu={true}
+      menuProps={{
```
ecraig12345: useConst

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.ScrollBar.Example.tsx
```diff
@@ -1,59 +1,44 @@
...
-    super(props);
-    this.state = {
-      showCallout: false
-    };
-  }
+export const ContextualMenuWithScrollBarExample: React.FunctionComponent = () => {
+  return (
+    <DefaultButton
+      text="Click for ContextualMenu"
+      menuProps={{
```
ecraig12345: useConst

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Section.Example.tsx
```diff
@@ -1,75 +1,72 @@
...
-            ]
-          }}
-        />
-      </div>
-    );
+export const ContextualMenuSectionExample: React.FunctionComponent = () => {
+  return (
+    <DefaultButton
+      text="Click for ContextualMenu"
+      menuProps={{
```
ecraig12345: useConst

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.Submenu.Example.tsx
```diff
@@ -1,143 +1,138 @@
...
+      <TextField
+        value={String(hoverDelay)}
+        label="Hover delay (ms)"
+        type="number"
+        onChange={onHoverDelayChanged}
+        styles={textFieldStyles}
+      />
+      <DefaultButton
+        text="Click for ContextualMenu"
+        menuProps={{
```
ecraig12345: useMemo

packages/office-ui-fabric-react/src/components/ContextualMenu/examples/ContextualMenu.CustomMenuList.Example.tsx
khmakoto: Huh, weird, I've usually heard that we shouldn't use `mergeStyleSets` for menial things like this because it contributes to perf hits. I'll move the styles below the example though.

ecraig12345: Interesting, I'd be curious about more background on that. I'm guessing that applies mainly when it's done within a component's render method? Though if we're trying to model best practices, maybe we shouldn't be showing something similar to a bad practice in examples...the trouble is that using what I *think* is the current best practice (though now about to be deprecated...) of styled components makes examples way too complicated.

khmakoto: Yeah, might be good knowing what @dzearing things about this, since "best practice" is very subjective.

khmakoto:
@ecraig12345 just updated the PR to address the latest feedback.

PR #11765, opened 2020-01-22
Creating react-focus package and copying FocusZone code there

dzearing:
This PR introduces an initial breakout of the focus components into a `react-focus` package.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11765)


packages/react-focus/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@fluentui/react-focus",
+  "version": "0.0.1",
```
ecraig12345: Is this intentional? create-package usually does 0.1.0 to avoid the weirdness of 0.0.x semver.

packages/react-focus/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "@types/webpack-env": "1.13.9",
+    "@uifabric/build": "^7.0.0",
+    "@uifabric/example-app-base": "^7.11.1",
+    "@uifabric/jest-serializer-merge-styles": "^7.0.6",
+    "@uifabric/prettier-rules": "^7.0.4",
+    "@uifabric/tslint-rules": "^7.0.5",
+    "enzyme": "~3.9.0",
+    "enzyme-adapter-react-16": "^1.2.0",
+    "es6-weak-map": "^2.0.2",
+    "office-ui-fabric-react": "^7.81.0",
```
ecraig12345: Should this be depending on OUFR at all? Will that create a circular link in graphs?

packages/react-focus/README.md
```diff
@@ -0,0 +1,11 @@
+# @fluentui/react-focus
+
+**React Focus components for [Fluent UI](https://fluent.microsoft.com/)**
```
ecraig12345: Can we update the create-package template to do this (if it is what we want for other packages too)?

packages/react-focus/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@fluentui/react-focus",
```
ecraig12345: Will beachball support publishing to two scopes @kenotron ?

packages/react-focus/src/components/FocusZone/FocusZone.tsx
```diff
@@ -0,0 +1,1201 @@
+import * as React from 'react';
```
ecraig12345: Are these files just a copy of the ones in OUFR?

packages/react-focus/src/components/FocusZone/FocusZonePage.tsx
```diff
@@ -0,0 +1,73 @@
+import * as React from 'react';
```
ecraig12345: We don't have to change this now but we should consider whether we actually want these to live inside the relevant package (which will definitely create a circular dep just for example stuff if OUFR eventually depends on this new package) or inside fabric-website or something. 

If we made that change everywhere, it would also let us reduce the size of published packages. It would be a big change but definitely something that could be worth doing (and which I'd be interested in doing).

packages/react-focus/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "react": "16.8.6",
+    "react-app-polyfill": "~1.0.1",
+    "react-dom": "16.8.6",
+    "react-test-renderer": "^16.3.0"
+  },
+  "dependencies": {
+    "@uifabric/set-version": "^7.0.2",
+    "@uifabric/utilities": "^7.11.0",
+    "@uifabric/merge-styles": "^7.8.1",
+    "tslib": "^1.7.1"
```
JasonGore: This has to be `^1.10.0` to prevent `__spreadArray` exceptions per #11739. Did this come from `create-package` or a template? If so please let me know, it has to be fixed.

yarn.lock
```diff
@@ -14404,7 +14404,7 @@ ts-loader@^4.1.0:
     micromatch "^3.1.4"
     semver "^5.0.1"
 
-tslib@^1.10.0, tslib@^1.8.0, tslib@^1.8.1, tslib@^1.9.0:
+tslib@^1.10.0, tslib@^1.7.1, tslib@^1.8.0, tslib@^1.8.1, tslib@^1.9.0:
```
JasonGore: Per my other comment, we shouldn't see any changes in this file.

packages/react-focus/package.json
dzearing: i wasn't sure what the weirdness was. I can certainly revert but what's the weirdness?

packages/react-focus/package.json
dzearing: no it should not, but i'm a little unsure about the docs solution.

The docs refer to components within oufr. So i will need to figure out a way to address keeping the examples somewhere else.

packages/react-focus/package.json
dzearing: i don't see this being a problem.

packages/react-focus/src/components/FocusZone/FocusZone.tsx
dzearing: yes, for now.

dzearing: This is just a scaffolding package. Once it is in a good shape, we can do a separate PR to move the other one to new location.

packages/react-focus/src/components/FocusZone/FocusZonePage.tsx
dzearing: yes! I was thinking... should these be in umm, react-focus-docs

packages/react-focus/package.json
dzearing: Yes I believe this was from the template.

yarn.lock
dzearing: weird!

packages/react-focus/package.json
ecraig12345: I think it's mainly that `^0.0.1` will only pick up **specifically** that version (essentially `^` or `~` is ignored).

Also IIRC `^` behaves like `~` for `0.(>0).x`, but that's not quite as surprising.

packages/react-focus/package.json
ecraig12345: Yeah we should potentially consider moving the examples to fabric-website or maybe even a dedicated examples package. That would fix a substantial number of circular dep issues.

packages/react-focus/src/components/FocusZone/FocusZonePage.tsx
ecraig12345: A dedicated docs package for each published package seems like a bit much. Though if you put them all in the same package, once we multiple components by the same name, you have to figure out how to distinguish them, so...not sure of the best approach here.

yarn.lock
JasonGore: Well, this change is caused by the tslib `^1.7.1` usage above. When you change it to `^1.10.0`, this file should revert to having no changes. :)

packages/react-focus/package.json
JasonGore: I ran `yarn create-package` and it did use tslib `^1.10.0` correctly, maybe this was created before my PR was merged. If you used `create-package` and ran into any issues, let me know and I'll queue them up. I'll have to update it anyways per upcoming CI changes (project references.)

packages/react-focus/package.json
khmakoto: +1 for separating examples

scripts/just.config.ts
```diff
@@ -1,29 +1,31 @@
...
-import prettier from './tasks/prettier';
-import bundleSizeCollect from './tasks/bundle-size-collect';
-import checkForModifiedFiles from './tasks/check-for-modified-files';
-import generateVersionFiles from './tasks/generate-version-files';
-import generatePackageManifestTask from './tasks/generate-package-manifest';
-import { postprocessTask } from './tasks/postprocess';
-import { postprocessAmdTask } from './tasks/postprocess-amd';
-import { postprocessCommonjsTask } from './tasks/postprocess-commonjs';
-import { startStorybookTask, buildStorybookTask } from './tasks/storybookTask';
+// @ts-check
```
kenotron: Is there a reason this is turned back from typescript syntax back to javascript?

ecraig12345: The change to TS syntax was actually just in a particular commit of this PR, not in master. It broke some other stuff, so I turned it back for now for simplicity.

ecraig12345:
Made some updates to this PR after discussion with @kenotron:
- Delete OUFR FocusZone and re-export the one from react-focus
- The examples and docs stay in OUFR for now to avoid a circular dependency
- Addressed outstanding comments aside from ones pertaining to create-package and a separate long-term example location, which will be handled separately
- Add API Extractor to react-focus
- Update api-docs and tsx-editor to support react-focus

yarn.lock
ecraig12345: I fixed this.

packages/react-focus/src/components/FocusZone/FocusZonePage.tsx
ecraig12345: Thinking about this again, the simplest solution would be to have the examples in a docs or website package, with sub-folders per package (which solves the issue of how to have examples for both e.g. OUFR button and experiments button). Definitely not happening in this PR though.

packages/react-focus/package.json
ecraig12345: Before this merges, @kenotron needs to make sure that the account we use for publishing has permission to publish to `@fluentui` as well.

packages/react-focus/package.json
ecraig12345: Left the examples in OUFR for now to avoid the circular dep and will move elsewhere long-term

packages/react-focus/src/components/FocusZone/FocusZone.test.tsx
```diff
@@ -2,11 +2,11 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import * as renderer from 'react-test-renderer';
 import * as ReactTestUtils from 'react-dom/test-utils';
-import { setRTL, KeyCodes } from '../../Utilities';
+import { setRTL, KeyCodes } from '@uifabric/utilities';
```
JasonGore: 👍

packages/react-focus/package.json
kenotron: I need the owner of uifabricteam on npm js.org to accept my invitation. Likely @dzearing?

ecraig12345:
The mysterious test failure was due to a missing path mapping causing ComponentExamples.test.tsx to fail, but it was really hard to debug because the test was eating error messages (displaying a custom error instead of the real error). That has now been substantially improved, though I'm still not sure why the error messages weren't showing up in the CI build when they did show up locally.

ecraig12345:
ComponentExamples stuff has been split out into #11849

ecraig12345:
Tests are failing with this error, **on both this and an unrelated PR**...not sure what's going on.

```
office-ui-fabric-react:   ● DatePicker › should reflect the correct date in the input field when selecting a value
office-ui-fabric-react:     TypeError: Cannot read property '__reactInternalInstance$lwv38jf7tt' of null
office-ui-fabric-react: 
office-ui-fabric-react:       226 |     datePicker.setState({ isDatePickerShown: true });
office-ui-fabric-react:       227 |     const todayButton = document.querySelector('.ms-DatePicker-day--today') as HTMLButtonElement;
office-ui-fabric-react:     > 228 |     ReactTestUtils.Simulate.click(todayButton);
office-ui-fabric-react:           |                             ^
office-ui-fabric-react:       229 | 
office-ui-fabric-react:       230 |     const selectedDate = datePicker
office-ui-fabric-react:       231 |       .find('input')
office-ui-fabric-react: 
office-ui-fabric-react:       at getInstanceFromNode$1 (../../node_modules/react-dom/cjs/react-dom.development.js:888:18)
office-ui-fabric-react:       at Object.click (../../node_modules/react-dom/cjs/react-dom-test-utils.development.js:1209:22)
office-ui-fabric-react:       at Object.<anonymous> (src/components/DatePicker/DatePicker.test.tsx:228:29)
office-ui-fabric-react:   ● DatePicker › should reflect the correct date in the input field when selecting a value and a different format is given
office-ui-fabric-react:     TypeError: Cannot read property '__reactInternalInstance$lwv38jf7tt' of null
office-ui-fabric-react: 
office-ui-fabric-react:       250 |     datePicker.setState({ isDatePickerShown: true });
office-ui-fabric-react:       251 |     const todayButton = document.querySelector('.ms-DatePicker-day--today') as HTMLButtonElement;
office-ui-fabric-react:     > 252 |     ReactTestUtils.Simulate.click(todayButton);
office-ui-fabric-react:           |                             ^
office-ui-fabric-react:       253 | 
office-ui-fabric-react:       254 |     const selectedDate = datePicker
office-ui-fabric-react:       255 |       .find('input')
office-ui-fabric-react: 
office-ui-fabric-react:       at getInstanceFromNode$1 (../../node_modules/react-dom/cjs/react-dom.development.js:888:18)
office-ui-fabric-react:       at Object.click (../../node_modules/react-dom/cjs/react-dom-test-utils.development.js:1209:22)
office-ui-fabric-react:       at Object.<anonymous> (src/components/DatePicker/DatePicker.test.tsx:252:29)
```

JasonGore:
Test should be fixed with the merge of #11856.

packages/react-focus/src/components/FocusZone/FocusZone.tsx
```diff
@@ -427,8 +430,8 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
    * Handle global tab presses so that we can patch tabindexes on the fly.
    */
   private _onKeyDownCapture = (ev: KeyboardEvent): void => {
-    if (ev.which === KeyCodes.tab) {
-      _outerZones.forEach(zone => zone._updateTabIndexes());
+    if (ev.key === 'Tab') {
```
aneeshack4: why this change to not use KeyCodes?

khmakoto: `ev.which` and `ev.keyCode` are deprecated:

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

aneeshack4: Ok would be good to remove the rest of this usage too but can be done in subsequent PRs as well

ecraig12345: What's the recommended alternative?

ecraig12345: Can we update the key codes enum or add a new one so we're not stuck using strings which aren't type safe?

dzearing: wait all. don't make a single code change in this pr. separate ones.

khmakoto: Ok, I'll revert `which` and just add a tslint exception so that the linter doesn't complain. For reference, [these](https://github.com/OfficeDev/office-ui-fabric-react/pull/11765/commits/a34ca1ebef4c5f1248c5fa10e219fa8ce34b45d2) are the only things I changed.

dzearing: Here's what I was thinking:

Step 1. Single PR. Break out FZ into package. Don't remove old.
Step 2. Get it deployed and published. Make sure it's consumable and safe.
Step 3. Separate PR, now make OUFR depend on it.

I think we might be doing too many steps here.

khmakoto: For sure, I'll remove what I did and just add a rule at the top to disable the linter in that file for now, since it won't pass the build if I don't disable it.

dzearing: BTW @khmakoto ev.key doesn't seem to be supported in android browser.

https://caniuse.com/#search=KeyboardEvent.key

dzearing: Also, we should adopt this eventually https://github.com/levithomason/keyboard-key/blob/master/src/keyboardKey.js

dzearing: Let's get the package to a v1 state, so that we can ship it in current without breaking everyone, then remove deprecations and stage that for a next release.

packages/office-ui-fabric-react/package.json
```diff
@@ -71,6 +71,7 @@
     "sinon": "^4.1.3"
   },
   "dependencies": {
+    "@fluentui/react-focus": "^0.1.0",
```
dzearing: Can we not do this dependency in production? The docs literally say: no production code should use a v0.

Let's stage this. Leave the old code. Publish. Then in a separate PR do this and validate that nothing breaks.

khmakoto:
@dzearing, @ecraig12345 I've now left a copy of `FocusZone` inside the oufr package until we publish the `react-focus` one. I think this PR should be ready to be merged without contention as nothing official is dependent on `react-focus` now.

packages/office-ui-fabric-react/package.json
khmakoto: Done!

packages/react-focus/src/components/FocusZone/FocusZone.tsx
khmakoto: Reverted the deprecation-related changes!

packages/react-focus/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "@types/jest": "~24.9.0",
+    "@types/react": "16.8.11",
+    "@types/react-dom": "16.8.4",
+    "@types/react-test-renderer": "^16.0.0",
+    "@types/webpack-env": "1.13.9",
+    "@uifabric/build": "^7.0.0",
+    "@uifabric/jest-serializer-merge-styles": "^7.0.6",
+    "@uifabric/prettier-rules": "^7.0.4",
+    "@uifabric/tslint-rules": "^7.0.5",
+    "enzyme": "~3.9.0",
```
JasonGore: I think this got out of sync with other enzyme versions in the repo, causing master to fail due to a required yarn.lock update.

PR #11778, opened 2020-01-23
Storybook upgraded to 5.3, SB also enabled for Fabric behind a specific just-task

kenotron:
#### Pull request checklist

#### Description of changes

In the effort of enabling a better innerloop experience, we're moving to Storybook. The idea is to take advantage of the Common Story Format (CSF) to describe our examples as well as an ever growing ecosystem of addons for the demo app.

We can then export these examples / docs for another website repo / package as a consumable bit (e.g. a GatsbyJS app). This will allow us to get out of the business of maintaining two separate & slightly different web applications. 

This change is an initial checkin that:
1. enables SB in Fabric as a just-task (`yarn just storybook:start`) - we're not quite ready to pull the full trigger to replace our inner loop just yet (need to gather a slight bit of feedback of this first)
2. upgrades the entire repo to use the new Storybook 5.3 that has most of its addons ecosystem providing their own typings

#### Focus areas to test

VR-Tests project is heavily modified here to upgrade to use the latest SB 5.3 APIs (using the storiesOf() api still, but addStory got renamed to add).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11778)


apps/vr-tests/src/stories/ActivityItem.stories.tsx
```diff
@@ -17,7 +17,7 @@ storiesOf('ActivityItem', module)
       {story()}
     </Screener>
   )
-  .addStory(
+  .add(
```
kenotron: >add [](start = 3, length = 3)

apparently SB 5.3 changed all its storiesOf() api!

kenotron: This accounts for most of the changes under "stories/"

kenotron: Nvm - this is my misunderstanding - `addStory` is an API that we invented.

scripts/webpack/webpack-resources.js
```diff
@@ -154,6 +157,10 @@ module.exports = {
           extensions: ['.ts', '.tsx', '.js']
         },
 
+        resolveLoader: {
+          modules: ['node_modules', path.join(__dirname, '../node_modules'), path.join(__dirname, '../../node_modules')]
```
JasonGore: Could this automatically resolve deps for packages that didn't explicitly include them? I'm worried with this change that a package could resolve to a dep of another package but then fail when consumed as an NPM package.

kenotron: All webpack resources / devdeps reside in scripts. There could be phantom webpack loaders that we reference in theory, but that's kind of a problem of having a tooling package inside a yarn workspace monorepo in general. Given that this is a single source of truth (all our webpack configs are in /scripts), we should be okay

JasonGore: Ok, I see this is for resolving loaders, not deps, so my concern doesn't apply.

scripts/config/storybook/webpack.config.js
```diff
@@ -0,0 +1,76 @@
+const HardSourceWebpackPlugin = require('hard-source-webpack-plugin');
+const IgnoreNotFoundExportWebpackPlugin = require('ignore-not-found-export-webpack-plugin');
+const path = require('path');
+
+module.exports = ({ config }) => {
```
ecraig12345: I know this is merged already but can we share any of this with the regular webpack config?

PR #11789, opened 2020-01-24
Use aria-current for active nav links

### LINKED ISSUE #11788: Active Nav component links override accessible name
### Environment Information

- **Package version(s)**: latest / 7.84.1

### Describe the issue:

The `selectedAriaLabel` property adds an `aria-label` with the defined text to the currently selected link. This means that the selected link will have its accessible name overridden by generic "selected" text, making it inaccessible. Even if all links have individual `ariaLabel` properties and the "selected" text is appended to that `ariaLabel`, including the active state of a link in its accessible name is not good practice.

Ideally the active link should have `aria-current="page"` set, and the accessible name should not be modified.

### Reproduction:
https://fabricweb.z5.web.core.windows.net/oufr/7.84.1/#/examples/nav

The issue repros on the first example in the Nav controls page. The highlighted "Documents" link has its accessible name overridden with the text "Selected", so a screen reader user would be unaware of what the text of the link actually is.

#### Actual behavior:

The "Documents" link has the accessible name "Selected"

#### Expected behavior:

The "Documents" link should get its accessible name from the text content (i.e. "Documents"), and have `aria-current="page"` set.

https://w3c.github.io/aria/#aria-current
https://tink.uk/using-the-aria-current-attribute/

smhigley:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11788

#### Description of changes

Add `aria-current="page"` to active links, and remove logic that modifies the accessible name of selected links.

#### Choosing "page"
Though `aria-current` takes multiple token values, the only two that could potentially be used with the Nav component are `page` and `step`. The overwhelming majority of Nav component use cases will likely fit best with "page", and even for the minority where "step" may be a good choice, the use of "step" over "page" would be a relatively minor augment. It's hard to imagine a use of the Nav component where `aria-current="page"` would be entirely inappropriate.

#### Removing `selectedAriaLabel` behavior

The current `selectedAriaLabel` property has two possible behaviors: it replaces the accessible name of links without an `ariaLabel` property, or it appends text to the label of links with an `ariaLabel` property.

The first behavior is a serious bug, and the second is not broken, but is less than ideal. Using an ARIA state attribute instead of a modified accessible name has the benefits of creating a more uniform, accessible experience across the web, and keeps the visible name matching the programmatic name.

This PR removes the `aria-label` logic associated with `selectedAriaLabel`, but keeps the property in the Nav API for backward compat (though I'd be happy to remove it entirely if desired).

___

Thanks for looking at this, and please let me know if I've missed anything 😊

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11789)

### DIFF
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -5592,6 +5592,7 @@ export interface INavButtonProps extends IButtonProps {
 // @public (undocumented)
 export interface INavLink {
     [propertyName: string]: any;
+    ariaCurrent?: 'page' | 'step' | 'location' | 'date' | 'time' | 'true';
     ariaLabel?: string;
     automationId?: string;
     collapseAriaLabel?: string;
@@ -5639,6 +5640,7 @@ export interface INavProps {
     onLinkExpandClick?: (ev?: React.MouseEvent<HTMLElement>, item?: INavLink) => void;
     onRenderGroupHeader?: IRenderFunction<INavLinkGroup>;
     onRenderLink?: IRenderFunction<INavLink>;
+    // @deprecated
     selectedAriaLabel?: string;
     selectedKey?: string;
     styles?: IStyleFunctionOrObject<INavStyleProps, INavStyles>;
--- a/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
@@ -92,6 +92,7 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
     const { styles, groups, theme } = this.props;
     const isLinkWithIcon = link.icon || link.iconProps;
     const isSelectedLink = this._isLinkSelected(link);
+    const { ariaCurrent = 'page' } = link;
     const classNames = getClassNames(styles!, {
       theme: theme!,
       isSelected: isSelectedLink,
@@ -118,7 +119,7 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
         target={link.target}
         rel={rel}
         disabled={link.disabled}
-        aria-current={isSelectedLink ? 'page' : undefined}
+        aria-current={isSelectedLink ? ariaCurrent : undefined}
         aria-label={link.ariaLabel ? link.ariaLabel : undefined}
         link={link}
       >
--- a/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
@@ -89,7 +89,7 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
   };
 
   private _renderNavLink(link: INavLink, linkIndex: number, nestingLevel: number): JSX.Element {
-    const { styles, groups, theme, selectedAriaLabel } = this.props;
+    const { styles, groups, theme } = this.props;
     const isLinkWithIcon = link.icon || link.iconProps;
     const isSelectedLink = this._isLinkSelected(link);
     const classNames = getClassNames(styles!, {
@@ -103,7 +103,6 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
 
     // Prevent hijacking of the parent window if link.target is defined
     const rel = link.url && link.target && !isRelativeUrl(link.url) ? 'noopener noreferrer' : undefined;
-    const selectedStateAriaLabel = isSelectedLink && selectedAriaLabel ? selectedAriaLabel : undefined;
 
     const LinkAs = this.props.linkAs ? composeComponentAs(this.props.linkAs, ActionButton) : ActionButton;
     const onRenderLink = this.props.onRenderLink ? composeRenderFunction(this.props.onRenderLink, this._onRenderLink) : this._onRenderLink;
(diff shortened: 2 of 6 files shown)


HalcyonPride:
I can't speak for the code changes. As far as the functional changes go, however, this is an excellent Accessibility PR with a thorough explanation - thank you!

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
```diff
@@ -119,15 +118,8 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
...
-        aria-label={
-          link.ariaLabel && selectedStateAriaLabel
-            ? `${link.ariaLabel} ${selectedStateAriaLabel}`
-            : selectedStateAriaLabel
-            ? selectedStateAriaLabel
-            : link.ariaLabel
-            ? link.ariaLabel
-            : undefined
-        }
+        aria-current={isSelectedLink ? 'page' : undefined}
```
jdhuntington: We should introduce a new property on `link` that can be used to set this value; perhaps `ariaCurrentType` and set it to be an element in type `'page' | 'step' | 'location' | 'date' | 'time' | 'true' | 'false'`. There are several scenarios where a `Nav` can be used but doesn't actually control page navigation.

jdhuntington: That said, I think `page` is the perfect default if not otherwise provided.

smhigley: @jdhuntington is that something you'd like within this PR? I'd be happy to add `ariaCurrentType`, though I'd suggest limiting it to `'page' | 'step' | 'true'`.

Granted, I'm assuming the Nav component would not be used as a date grid, a list of times (where the times are not links), or a flowchart, (and that `undefined` is preferable to `false`). If those seem possible enough, or it doesn't seem like a good idea to restrict allowed values, the full list makes sense too :)

dzearing: @smhigley I think adding `ariaCurrent` prop to `INavLink` with the restrictions makes sense. If not provided defaults to 'page'.

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
```diff
@@ -119,15 +118,8 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
...
+        aria-label={link.ariaLabel ? link.ariaLabel : undefined}
```
dzearing: `selectedStateAriaLabel` is no longer respected here. This is documented in the `INavLink.types.ts` file as a property but it no longer is applied.

I think that probably makes sense; judging from this crazy logic, it doesn't key off any state and seems to serve no purpose. So a deprecation probably makes sense.

Please modify the comment of the property and replace the description with something like:

```ts
@deprecated Use `ariaLabel` instead.
```

packages/office-ui-fabric-react/src/components/__snapshots__/Nav.Basic.Example.tsx.shot
```diff
@@ -638,7 +638,7 @@ exports[`Component Examples renders Nav.Basic.Example.tsx correctly 1`] = `
               name="Documents"
             >
               <a
-                aria-label="Selected"
+                aria-current="page"
```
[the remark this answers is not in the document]
inateeg: OK got it after reading aria-current

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
inateeg: reviewed, looks good to me.

inateeg: just wonder if  'page' is enum, does it mean it will work for all languages?

smhigley: @dzearing not allowing authors to create a separate screen-reader-only label was purposeful, since I would consider it an easy-to-misunderstand antipattern. On the other hand, I don't know what the original motivation was; if there's a reasonable need, perhaps hint text/`aria-describedby` could be an alternate place that `selectedStateAriaLabel` could go?

@inateeg it does work for all languages, the localization is handled at the accessibility API and screen reader level. Thanks for checking that :)

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
smhigley: @dzearing done!

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -5629,6 +5629,7 @@ export interface INavButtonProps extends IButtonProps {
 // @public (undocumented)
 export interface INavLink {
     [propertyName: string]: any;
+    ariaCurrent?: 'page' | 'step' | 'location' | 'date' | 'time' | 'true';
```
xugao: shouldn't this be `'page' | 'true'` . trying to think why would user set the other values for a nav link?

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
```diff
@@ -119,15 +119,8 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
...
-          link.ariaLabel && selectedStateAriaLabel
-            ? `${link.ariaLabel} ${selectedStateAriaLabel}`
-            : selectedStateAriaLabel
-            ? selectedStateAriaLabel
-            : link.ariaLabel
-            ? link.ariaLabel
-            : undefined
-        }
+        aria-current={isSelectedLink ? ariaCurrent : undefined}
+        aria-label={link.ariaLabel ? link.ariaLabel : undefined}
```
xugao: can just simply be `aria-label={link.ariaLabel}`

PR #11792, opened 2020-01-24
Nav: Fixing regular expression used to determine if URL is protocol relative or not.

### LINKED ISSUE #11785: rel="noopener noreferrer" is not added for external link for Nav
### Environment Information

- **Package version(s)**: 7.84.0

#### Codepen Demo
https://codepen.io/dance2die/pen/ZEYwdRX?editors=0010

#### Reproduction Steps

1. Go to https://developer.microsoft.com/en-us/fabric#/controls/web/nav
1. Scroll down to "Nav similar to the one in this demo app" section
1. Export to Codepen.
1. Replace internal links with external links.
e.g.) 
```javascript
<Nav
      styles={{ root: { width: 300 } }}
      ariaLabel="Nav example similiar to one found in this demo page"
      groups={[
        {
          name: 'External Links',
          expandAriaLabel: 'Expand Basic components section',
          collapseAriaLabel: 'Collapse Basic components section',
          links: [
            {
              key: 'r/reactjs',
              name: 'r/reactjs',
              url: 'https://www.reddit.com/r/reactjs/',
              target: "__blank"
            },
            {
              key: 'Nav',
              name: 'Fabric Nav',
              url: 'https://developer.microsoft.com/en-us/fabric#/controls/web/nav'
              ,target: "__blank"
            }
          ]
        }
      ]}
    />
```
5. Open a new preview window.
image
6. View the generated source on links.
7. You can see that `rel="noopener noreferrer"` is not added for the external links.
image

#### Actual behavior:

`rel="noopener noreferrer"` is not added for external links with target specified for [Nav](https://developer.microsoft.com/en-us/fabric#/controls/web/nav) component.

#### Expected behavior:

Generated anchor tags must have `rel="noopener noreferrer"`.

[Nav.base.tsx](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx#L105) implements it but `rel` attribute isn't added.

Requested priority: High (security issue)

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11785

#### Description of changes

The regular expression we were using to determine if a URL was protocol relative or not was wrong, which was causing `Nav` links to not have `rel="noopener noreferrer"` set up, causing a potential security risk. This PR corrects this by fixing the regular expression.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11792)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
@@ -18,7 +18,7 @@ let _urlResolver: HTMLAnchorElement | undefined;
 
 export function isRelativeUrl(url: string): boolean {
   // A URL is relative if it has no protocol.
-  return !!url && !/^[a-z0-9+-.]:\/\//i.test(url);
+  return !!url && !/^[a-z0-9+-.]+:\/\//i.test(url);
 }
 
 const getClassNames = classNamesFunction<INavStyleProps, INavStyles>();
--- a/change/office-ui-fabric-react-2020-01-24-11-45-34-navRelLink.json
+++ b/change/office-ui-fabric-react-2020-01-24-11-45-34-navRelLink.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Nav: Fixing regular expression used to determine if URL is protocol relative or not.",
+  "packageName": "office-ui-fabric-react",
+  "email": "Humberto.Morimoto@microsoft.com",
+  "commit": "eadbfa6f02f782e8252c37dfe577d4000ff741dd",
+  "dependentChangeType": "patch",
+  "date": "2020-01-24T19:45:33.986Z"
+}
--- a/packages/office-ui-fabric-react/src/components/Nav/Nav.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/Nav/Nav.test.tsx
@@ -129,4 +129,40 @@ describe('Nav', () => {
     expect(nav.getDOMNode().querySelectorAll('.ms-Nav-compositeLink.is-selected').length).toBe(1);
     expect(nav.getDOMNode().querySelectorAll('.ms-Nav-compositeLink.is-selected')[0].textContent).toEqual(linkOne.name);
   });
+
+  it('places the correct values on rel depending on the url and target specified', () => {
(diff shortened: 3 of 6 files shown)


packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
```diff
@@ -18,7 +18,7 @@ let _urlResolver: HTMLAnchorElement | undefined;
 
 export function isRelativeUrl(url: string): boolean {
   // A URL is relative if it has no protocol.
-  return !!url && !/^[a-z0-9+-.]:\/\//i.test(url);
+  return !!url && !/^[a-z0-9+-.]+:\/\//i.test(url);
```
ecraig12345: Is `.` valid in protocols?

khmakoto: Good question! It makes me question if any of those characters are actually permitted:

image

khmakoto:
@jdhuntington the snapshots updated because the regular expression was always returning false, which was setting the `rel` to `undefined` in line 105:

```tsx
const rel = link.url && link.target && !isRelativeUrl(link.url) ? 'noopener noreferrer' : undefined;
```

Regarding the test you mention, any idea how to make an exhaustive test about this and not only make the function run through different cases?

jdhuntington:
> Regarding the test you mention, any idea how to make an exhaustive test about this and not only make the function run through different cases?

I think different cases is a fine way to do it; they can be amended if necessary.

As for implementation, what about this approach to leverage the browser's URL parsing instead?

```js
function isRelativeUrl(href) {
  const loc = window.location;
  const url = new URL(href, loc);
  return loc.origin === url.origin;
}

console.log(isRelativeUrl("/foo")); // => true
console.log(isRelativeUrl("https://cdpn.io/bar")); // => true
console.log(isRelativeUrl("http://cdpn.io/bar")); // => false
console.log(isRelativeUrl("https://example.com")); // => false
```

ecraig12345:
@jdhuntington This is why we can't use URL yet :(
https://caniuse.com/#feat=url
image

PR #11796, opened 2020-01-25
Panel: Calling onOpen and on non-imperative scenarios

### LINKED ISSUE #11782: Panel onOpen prop does not work
### Environment Information

- **Package version(s)**: 7.84.0

### Describe the issue:
The onOpen callback never gets called for Panels.

#### Actual behavior:
Other callbacks, like onDismiss and onOpened, do get called but not onOpen

#### Expected behavior:
onOpen prop gets called

https://codepen.io/kevinjkim/pen/ExarzXK

Requested priority: Normal/Low

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11782 

#### Description of changes

The description for the `onOpen` prop in the `Panel` component states that it should be called when the `Panel` opens but was only being called on imperative scenarios. This PR fixes this so that `onOpen` is also called on non-imperative scenarios and adds a test so that this doesn't regress in the future.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11796)


khmakoto:
I'm not even touching `Text` here so I don't think there is a perf regression in this PR.

khmakoto:
@xugao would you mind merging the PR? I don't have admin privileges and Aneesha is on vacation.

PR #11797, opened 2020-01-25
Button: Adding support for onPointerDown and onPointerUp

### LINKED ISSUE #11767: onPointerDown not being called when passing it to Button

https://codepen.io/khmakoto/pen/OJPdXGR

#### Actual behavior:

When passing `onPointerDown` to the `Button` component, this never gets called because of business logic that is making use of this call.

#### Expected behavior:

When passing `onPointerDown` to the `Button` component, both business and user logic should be executed.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11767 

#### Description of changes

The `Button` component was swallowing user provided `onPointerDown` events and was not forwarding the correct `onPointerUp` events to the `MenuButton` part of `SplitButton` components. This PR fixes both of these issues.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11797)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
+++ b/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
@@ -240,8 +240,13 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
     // For split buttons, touching anywhere in the button should drop the dropdown, which should contain the primary action.
     // This gives more hit target space for touch environments. We're setting the onpointerdown here, because React
     // does not support Pointer events yet.
-    if (this._isSplitButton && this._splitButtonContainer.current && 'onpointerdown' in this._splitButtonContainer.current) {
-      this._events.on(this._splitButtonContainer.current, 'pointerdown', this._onPointerDown, true);
+    if (this._isSplitButton && this._splitButtonContainer.current) {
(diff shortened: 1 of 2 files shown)


makopch-ms:
Looks good to me.  Thanks!

packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
```diff
@@ -240,8 +240,13 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
     // For split buttons, touching anywhere in the button should drop the dropdown, which should contain the primary action.
     // This gives more hit target space for touch environments. We're setting the onpointerdown here, because React
     // does not support Pointer events yet.
```
xugao: i don't feel these comments are relevant. i think the code is pretty self-explain here. pls remove them if you agree

khmakoto: @xugao we've had issues before with people complaining about this behavior, which is why the comment was added in the first place. With subsequent issues we've just pointed them to this comment, so I wouldn't feel completely comfortable with removing it.

PR #11798, opened 2020-01-25
Dialog/Panel: allow to set aria-level on header title

### LINKED ISSUE #11666: [Dialog][Panel][Accessibility] The heading is always set at level 2.

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11666

#### Description of changes
Added "props" prop which allow user to set `aria-level` as well as other div html attributes on the container element of the title.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11798)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Dialog/DialogContent.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dialog/DialogContent.base.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { BaseComponent, classNamesFunction } from '../../Utilities';
+import { BaseComponent, classNamesFunction, getNativeProps, divProperties, css } from '../../Utilities';
 import { DialogType, IDialogContentProps, IDialogContentStyleProps, IDialogContentStyles } from './DialogContent.types';
 import { IconButton } from '../../Button';
 import { DialogFooter } from './DialogFooter';
@@ -21,6 +21,12 @@ export class DialogContentBase extends BaseComponent<IDialogContentProps, {}> {
 
   constructor(props: IDialogContentProps) {
     super(props);
+
+    if (process.env.NODE_ENV !== 'production') {
+      this._warnDeprecations({
+        titleId: 'titleProps.id'
+      });
+    }
   }
 
   public render(): JSX.Element {
@@ -31,6 +37,7 @@ export class DialogContentBase extends BaseComponent<IDialogContentProps, {}> {
       onDismiss,
       subTextId,
       subText,
+      titleProps = {},
       titleId,
       title,
       type,
@@ -57,12 +64,20 @@ export class DialogContentBase extends BaseComponent<IDialogContentProps, {}> {
       );
     }
 
+    const titleNativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(titleProps, divProperties, ['id', 'className']);
+
     return (
       <div className={classNames.content}>
         <div className={classNames.header}>
-          <p className={classNames.title} id={titleId} role="heading" aria-level={2}>
+          <div
+            className={css(classNames.title, titleProps.className)}
+            id={titleProps.id || titleId}
+            role="heading"
+            aria-level={2}
+            {...titleNativeProps}
+          >
             {title}
-          </p>
+          </div>
           <div className={classNames.topButton}>
             {this.props.topButtonsProps!.map((props, index) => (
               <IconButton key={props.uniqueId || index} {...props} />
--- a/packages/office-ui-fabric-react/src/components/Dialog/DialogContent.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dialog/DialogContent.base.tsx
@@ -67,13 +67,7 @@ export class DialogContentBase extends BaseComponent<IDialogContentProps, {}> {
     return (
       <div className={classNames.content}>
         <div className={classNames.header}>
-          <div
-            role="heading"
-            aria-level={2}
-            {...titleProps}
-            id={titleProps.id || titleId}
-            className={css(classNames.title, titleProps.className)}
-          >
+          <div id={titleId} role="heading" aria-level={2} {...titleProps} className={css(classNames.title, titleProps.className)}>
             {title}
           </div>
           <div className={classNames.topButton}>
--- a/packages/office-ui-fabric-react/src/components/Dialog/DialogContent.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dialog/DialogContent.base.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { BaseComponent, classNamesFunction, getNativeProps, divProperties, css } from '../../Utilities';
+import { BaseComponent, classNamesFunction, css } from '../../Utilities';
 import { DialogType, IDialogContentProps, IDialogContentStyleProps, IDialogContentStyles } from './DialogContent.types';
 import { IconButton } from '../../Button';
 import { DialogFooter } from './DialogFooter';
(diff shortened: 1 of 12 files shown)


packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -347,14 +348,21 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
...
   ): JSX.Element | null => {
-    const { headerText } = props;
+    const { headerText, headerTextProps = {} } = props;
+    const headerTextNativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(headerTextProps, divProperties, ['id', 'className']);
 
     if (headerText) {
       return (
         <div className={this._classNames.header}>
-          <p className={this._classNames.headerText} id={headerTextId} role="heading" aria-level={2}>
+          <div
```
JasonGore: Why the switch to `div`? Could this cause subtle styling regressions?

packages/office-ui-fabric-react/src/components/Panel/Panel.types.ts
```diff
@@ -77,6 +77,11 @@ export interface IPanelProps extends React.HTMLAttributes<PanelBase> {
    */
   headerText?: string;
 
+  /**
+   * The props for header text container.
+   */
+  headerTextProps?: IPanelHeaderTextProps;
```
JasonGore: Is there a reason not to just make this `React.AllHTMLAttributes<HTMLDivElement>` now? If it needs to be expanded later, it could be broken out while still maintaining backwards interface compatibility. Less interfaces is more, I think.

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -347,14 +348,21 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
     defaultRender?: (props?: IPanelProps) => JSX.Element | null,
     headerTextId?: string | undefined
   ): JSX.Element | null => {
-    const { headerText } = props;
+    const { headerText, headerTextProps = {} } = props;
+    const headerTextNativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(headerTextProps, divProperties, ['id', 'className']);
```
JasonGore: I think our pattern on most components is only to strip out native props when they are pulled out from the top level props object. When it comes to nested (subcomponent) props objects we spread them while preserving all properties. (Related to my `IPanelHeaderTextProps` comment, I think we should be clear about the type and what is actually preserved in the spread.)

packages/office-ui-fabric-react/src/components/Dialog/DialogContent.types.ts
```diff
@@ -77,6 +79,11 @@ export interface IDialogContentProps extends React.ClassAttributes<DialogContent
    */
   title?: string | JSX.Element;
 
+  /**
+   * The props for title container.
+   */
+  titleProps?: IDialogContentTitleProps;
```
JasonGore: Same comments as for `IPanelHeaderTextProps`.

packages/office-ui-fabric-react/src/components/Dialog/DialogContent.base.tsx
```diff
@@ -57,12 +64,20 @@ export class DialogContentBase extends BaseComponent<IDialogContentProps, {}> {
       );
     }
 
+    const titleNativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(titleProps, divProperties, ['id', 'className']);
```
JasonGore: Same comments as for `Panel.base`.

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
xugao: it's not semantically correct to use `p` here. Since i am adding props to spread on this element, I decided to change the element type here so i don't need to type it to be `React.AllHTMLAttributes<HTMLParagraphElement>`.

There is no visual regression from this change. But there is down side - if user is styling it using element type selector, this change is breaking. thoughts?

packages/office-ui-fabric-react/src/components/Panel/Panel.types.ts
xugao: i don't have a strong reason. i will update this as you suggested

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
JasonGore: The only hesitation I have is that this is actually an element with a styles classname. I'm not sure how practical a concern it is, but most of my fear would come from exposure. How many components does `DialogContent` get reused in? I'm not sure...

packages/office-ui-fabric-react/src/components/Dialog/DialogContent.types.ts
```diff
@@ -69,6 +69,8 @@ export interface IDialogContentProps extends React.ClassAttributes<DialogContent
 
   /**
    * The Id for title container
+   *
+   * @deprecated use titleProps.id instead.
```
khmakoto:
```suggestion
   * @deprecated use the `id` attribute in `titleProps` instead.
```

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
xugao: The stat report Ken made: https://msit.powerbi.com/groups/cbe7890b-e262-4c36-97c2-d6efc21c2373/reports/0747523d-9ce3-4676-8525-d9673e54cab6/ReportSection?ctid=72f988bf-86f1-41af-91ab-2d7cd011db47

In general , the usage of DialogContentStyles prop is really low. also, since we have 2 better ways for user to target the style of the title:
1) Style prop
2) Static class name
The chance that user is targeting the element using its type is quite low.

Overall, I still lean towards updating the element type here

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -347,14 +348,20 @@ export class PanelBase extends BaseComponent<IPanelProps, IPanelState> implement
...
     if (headerText) {
       return (
         <div className={this._classNames.header}>
-          <p className={this._classNames.headerText} id={headerTextId} role="heading" aria-level={2}>
+          <div
+            role="heading"
+            aria-level={2}
+            {...headerTextProps}
+            className={css(this._classNames.headerText, headerTextProps.className)}
+            id={headerTextProps.id || headerTextId}
```
JasonGore: If `headerTextProps.id` is given priority couldn't this line just move above 359 and be `id={headerTextId}`?

packages/office-ui-fabric-react/src/components/Dialog/DialogContent.types.ts
```diff
@@ -77,6 +79,11 @@ export interface IDialogContentProps extends React.ClassAttributes<DialogContent
    */
   title?: string | JSX.Element;
 
+  /**
+   * The props for title container.
+   */
+  titleProps?: React.AllHTMLAttributes<HTMLDivElement>;
```
JasonGore: Sorry for not noticing this earlier but I think this should just be `React.HTMLAttributes` for correctness and consistency. I'll approve for now

xugao: ah good catch. my bad. will fix

PR #11799, opened 2020-01-25
Storybook for more packages; yarn start for Fabric is switched to storybook

kenotron:
#### Pull request checklist

#### Description of changes

1. upgrade a11y-addon to an alpha version for now because it allows for manual runs
2. storybook now has all the examples and becomes the default innerloop experience
3. storybook is yarn start for OUFR, Cards, Lists, Date Time, Charting, and Experiments
4. yarn builddemo or build is no longer necessary before one starts the inner loop (git clone -> yarn -> yarn start)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11799)


packages/charting/.storybook/preview.js
```diff
@@ -0,0 +1,30 @@
+import { initializeIcons } from '@uifabric/icons';
```
ecraig12345: Can we get rid of the duplication (especially this larger file) between the projects?

kenotron: I think I will refactor into a file inside /scripts and then call this logic (we still need preview.js physically present)

kenotron: Actually, i've already refactored everything I wanted to refactor into the generateStoriesFromExample() function. I kept the duplication for these:

1. addons should be customizable according to the needs of the package itself
2. require.context should to be called in that exact file for relative import reasons
3. some package in the future might not need to initializeIcons() so I'm leaving it there.

So, actually a little duplication allows for good amount of flexibility. The meat of the logic was to convert our .Example.js to stories anyway, so that's already factored into a separated place. The rest of the code is simply config.

apps/vr-tests/package.json
```diff
@@ -35,7 +35,7 @@
     "@types/react-dom": "16.8.4",
     "@types/webpack-env": "1.13.9",
     "@uifabric/build": "^7.0.0",
-    "@uifabric/date-time": "^7.8.9",
+    "@uifabric/date-time": "^7.8.8",
```
JasonGore: The changes in this file don't seem right

kenotron: This file has been reverted - there's no need to touch this file.

packages/charting/.storybook/preview.js
ecraig12345: Okay. In either case you should really add the new files to the create-package script (and maybe delete anything we don't need any more while you're at it).

PR #11812, opened 2020-01-28
TeachingBubble: Applying Callout style-related props like calloutWidth and calloutMaxWidth.

### LINKED ISSUE #11780: TeachingBubble: "calloutWidth" in "calloutProps" is not working
### Environment Information

- **Package version(s)**: 6.210.6

https://codepen.io/alexsunms/pen/xxbMeOB?editable=true

#### Actual behavior:

Passing `calloutWidth` in `calloutProps` is not working, but "beakWidth" is.

#### Expected behavior:

Passing `calloutWidth` in `calloutProps` should change the width of TeachingBubble.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11780

#### Description of changes

The `Callout` related styles of the `TeachingBubble` were being set in `TeachingBubble.styles.ts` without any regard for style-related `calloutProps` like `calloutWidth` and `calloutMaxWidth`. This PR fixes this by taking these props into account in the `TeachingBubble` styles.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11812)


packages/office-ui-fabric-react/src/components/TeachingBubble/TeachingBubble.base.tsx
```diff
@@ -85,8 +85,8 @@ export class TeachingBubbleBase extends BaseComponent<ITeachingBubbleProps, ITea
       <Callout
         target={target || targetElement}
         onDismiss={onDismiss}
-        {...calloutProps}
         className={classNames.root}
+        {...calloutProps}
```
micahgodbolt: wouldn't it be better to mix the calloutProps.className in with the className above?

khmakoto: That makes sense, I'll make the change. I think the rest of the PR is still valid though.

khmakoto: Actually, this is already done in line 79, where you send `stylesProps` to the `getClassNames` function, so that `calloutProps.className` is already part of `classNames.root`.

PR #11813, opened 2020-01-28
Adding istrusted !== false check in the dismiss on scroll handler

### LINKED ISSUE #11800: isTrusted check to the dismiss on scroll handler
### Environment Information

#### Actual behavior:

Requirement is to add isTrusted check to the dismiss on scroll handler in calloutContent.base
The need to be able to stop the drop down menus getting closed when in a coauthoring environment a scroll event is received due to adjustment of the canvas in word online

Proposed change is to add a check to the if statement to not dismiss is ev.isTrusted == false on the file below at line# 264
https://github.com/OfficeDev/office-ui-fabric-react/blob/a5bbc76367eb01e993f20796955569f422a538b2/packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx

#### Expected behavior:

The scroll handler should be able to define if the closing of the dropdowns is required or not. Also this should be handled for scroll per say and rather on change of focus

Are you willing to submit a PR to fix? (Yes, No) YES

Requested priority: (Blocking, High, Normal, Low) Blocking (Hi-Pri customer issue)

Word Online

sabans-dev:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11800

#### Description of changes

Users in word online have been experiencing this behavior where when multiple author on a document are typing and one of those users had a drop down open from the simplified ribbon and another co-author writes something and enters a new line the drop down for the co-author gets closed automatically. Now this on investigation was found due to the adjustment word online has to do to make sure the co-authors do not get a scrolling experience from the fact that new text got added and the IP moves. In order to achieve this the auto scroll event is fired when the changes from co-authors are merged in the session. This event is handled by the office-ui code that makes sure all the drop downs get closed when scroll happens. i added a fix on that side of code to ensure this is not done via a flag check but that looked like a hack and this was suggested that this information to be given via the office-fabric code. Hence this change.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11813)

### DIFF
--- a/change/office-ui-fabric-react-2020-01-28-12-15-29-sabans-dismissonscrollbugfix.json
+++ b/change/office-ui-fabric-react-2020-01-28-12-15-29-sabans-dismissonscrollbugfix.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "Adding ev.isTrusted == false check for the diss dropdown callout handler.",
+  "packageName": "office-ui-fabric-react",
+  "email": "sabans@microsoft.com",
+  "commit": "a727628ea9e1ff750da418ab57b8c39c0c763c68",
+  "dependentChangeType": "patch",
+  "date": "2020-01-28T06:45:29.628Z"
+}
--- a/packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
@@ -269,7 +269,7 @@ export class CalloutContentBase extends React.Component<ICalloutProps, ICalloutS
 
   protected _dismissOnScroll = (ev: Event) => {
     const { preventDismissOnScroll } = this.props;
-    if (this.state.positions && !preventDismissOnScroll && ev.isTrusted == false) {
+    if (this.state.positions && !preventDismissOnScroll && ev.isTrusted !== false) {
       this._dismissOnClickOrScroll(ev);
     }
   };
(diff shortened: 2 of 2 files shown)


khmakoto:
Hi @sabans-dev, thanks for the contribution! In the overview, could you write a more detailed description of why you need this change and link it to its corresponding issue? Thanks!

sabans-dev:
Sure @khmakoto & thanks for correction on the title, missed it completely.
Detailed description:
Users in word online have been experiencing this behavior where when multiple author on a document are typing and one of those users had a drop down open from the simplified ribbon and another co-author writes something and enters a new line the drop down for the co-author gets closed automatically. Now this on investigation was found due to the adjustment word online has to do to make sure the co-authors do not get a scrolling experience from the fact that new text got added and the IP moves. In order to achieve this the auto scroll event is fired when the changes from co-authors are merged in  the session. This event is handled by the office-ui code that makes sure all the drop downs get closed when scroll happens. i added a fix on that side of code to ensure this is not done via a flag check but that looked like a hack and this was suggested that this information to be given via the office-fabric code. Hence this change.

sabans-dev:
@khmakoto Requesting some attention to this as this is blocking our CR. Apologies for the trouble.

joschect:
@sabans-dev Hey! Is this happening in Fabric dropdowns? Or do you have your own version? I'm tempted to say that this should be in the onDismiss callback that is passed to the callout from the dropdown. I'm a little worried about potentially changing the behavior for all consumers of callout by adding a check for isTrusted. If you're using the fabric dropdown the change might wind up being more involved.

I understand that this is a blocker for you, so if you add a test with a comment about why it's necessary, I'll approve and we can get things moving.

sabans-dev:
@joschect Thanks for your comments but then I still believe this should not affect anyone else per say as we are not checking if its undefined and just false. Having said that the scenario here is more related to a coauthoring environment which anyways has no affects when there is a single user per say. Requesting you to kindly help this unblock.

joschect:
@sabans-dev I understand that we don't believe it will impact other users, but it's adding some behavior that we'd want to capture. Per my comment if you add some tests for it I'll go ahead and approve. I want to make sure that if we do a refactor or something that this behavior is taken into account.

PR #11822, opened 2020-01-28
Fix key up causing scrolling issue of whole page

### LINKED ISSUE #11821: [Accessibility-Keyboard-Navigation] Dropdown, key up causes scrolling of whole page
Environment Information
Package version(s): Latest version
Browser and OS versions: Chrome, Edge, Firefox. Windows 10.
Describe the issue:
Pressing the up key, when the first option is highlightable/selectable, when the dropdown is not open, causes the whole page to scroll.

Actual behavior:

So I've created a codepen to make it easier to reproduce.  Please make sure that the scrollbar is visible and there is some space for the scrollbar to scroll upwards.  Tab to the dropdown, then press the up key.  You will see that the screen scrolls upwards.  The reason is that in the Dropdown.base.tsx, for the function _onDropdownKeyDown there is a check for newIndex !== selectedIndex.  Now in the function _moveIndex there is a check for going out of bounds but it puts the index back to the selectedIndex, thus causing the issue.  This is the same issue as https://github.com/OfficeDev/office-ui-fabric-react/issues/11768  but is only found out because of removing the header option ({ key: 'fruitsHeader', text: 'Fruits', itemType: DropdownMenuItemType.Header },) and thus allowing the first item in the dropdown to be selectable/highlightable.  

Expected behavior:
I expect the up key to cycle through all the options.

If applicable, please provide a codepen repro:
https://codepen.io/woofton/pen/GRgapGw

Priorities and help requested (not applicable if asking question):
Are you willing to submit a PR to fix? (Yes, No) Yes, about to create a PR to fix the issue.

woofton:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #11821

#### Description of changes

// If the user is pressing the up or down key we want to make
    // sure that the dropdown cycles through the options without
    // causing the screen to scroll. In _onDropdownKeyDown
    // at the very end is a check to see if newIndex !== selectedIndex.
    // If the index is less than 0 and we set it back to 0, then
    // newIndex will equal selectedIndex and not stop the action
    // of the key press happening and vic versa for index greater
    // than or equal to the options length.

#### Focus areas to test

Dropdowns.  Please see existing issue for how to reproduce the issue.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11822)

### DIFF
--- a/change/office-ui-fabric-react-2020-01-28-13-42-15-fixKeyUpCausingScrollingIssue.json
+++ b/change/office-ui-fabric-react-2020-01-28-13-42-15-fixKeyUpCausingScrollingIssue.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Fix an issue where pressing the key up can cause scrolling of the whole page.",
+  "packageName": "office-ui-fabric-react",
+  "email": "jehawley@microsoft.com",
+  "commit": "ec4f9e60183ddd572046820cea04c0f29c508a68",
+  "dependentChangeType": "patch",
+  "date": "2020-01-28T21:42:15.529Z"
+}
--- a/change/office-ui-fabric-react-2020-01-28-13-42-15-fixKeyUpCausingScrollingIssue.json
+++ b/change/office-ui-fabric-react-2020-01-28-13-42-15-fixKeyUpCausingScrollingIssue.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fix an issue where pressing the key up can cause scrolling of the whole page.",
+  "comment": "Dropdown: Fix issue where pressing key up can cause scrolling of the whole page.",
   "packageName": "office-ui-fabric-react",
   "email": "jehawley@microsoft.com",
   "commit": "ec4f9e60183ddd572046820cea04c0f29c508a68",
--- a/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
@@ -420,18 +420,20 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
       return selectedIndex;
     }
 
-    // Set starting index to 0 if index is < 0
-    if (index < 0) {
-      index = 0;
-    }
-    // Set starting index to the first option index if greater than options.length.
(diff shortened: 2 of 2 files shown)


packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -420,18 +420,20 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
...
-    // the dropdown, the dropdown will cycle through the options. If
-    // the index is set as the last option, then pressing the down key will
-    // cause the whole window to scroll, if the window can be scrolled.
+    // If the user is pressing the up or down key we want to make
+    // sure that the dropdown cycles through the options without
+    // causing the screen to scroll. In _onDropdownKeyDown
+    // at the very end is a check to see if newIndex !== selectedIndex.
+    // If the index is less than 0 and we set it back to 0, then
+    // newIndex will equal selectedIndex and not stop the action
+    // of the key press happening and vic versa for index greater
```
khmakoto:
```suggestion
    // of the key press happening and vice versa for indexes greater
```

change/office-ui-fabric-react-2020-01-28-13-42-15-fixKeyUpCausingScrollingIssue.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Fix an issue where pressing the key up can cause scrolling of the whole page.",
```
khmakoto:
```suggestion
  "comment": "Dropdown: Fix issue where pressing key up can cause scrolling of the whole page.",
```

packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx
woofton: Fixed.  Thank you.

change/office-ui-fabric-react-2020-01-28-13-42-15-fixKeyUpCausingScrollingIssue.json
woofton: Fixed.  Thank you.

PR #11824, opened 2020-01-29
Text: Inheriting parent style definitions when no variant is supplied

### LINKED ISSUE #11801: Text component does not inherit parent style definitions when no variant is supplied
### Environment Information

- **Package version(s)**: 6.176.0
- **Browser and OS versions**: Firefox 72.0.2, MacOS Mojave 10.14.6

https://codepen.io/singhshashi/pen/xxbBmzN?editable=true

#### Actual behavior:
Text component wrapped around by a div is not inheriting font weight bold from the div. 

#### Expected behavior:

Text component wrapped around a by a div and with no variant prop should inherit text styles from the parent as documented at https://developer.microsoft.com/en-us/fabric#/controls/web/text#overview

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11801

#### Description of changes

The `Text` component documentation establishes that the component inherits parent style definitions when no `variant` is supplied, but this wasn't happening and instead the `variant="medium"` styles were being applied. This PR fixes this behavior to accurately reflect the behavior in the component overview.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11824)

### DIFF
--- a/apps/vr-tests/src/stories/Card.stories.tsx
+++ b/apps/vr-tests/src/stories/Card.stories.tsx
@@ -189,7 +189,7 @@ storiesOf('Card', module)
     <Fabric>
       <Card>
         <Card.Item>
-          <Text>Basic vertical card</Text>
+          <Text variant="medium">Basic vertical card</Text>
         </Card.Item>
       </Card>
     </Fabric>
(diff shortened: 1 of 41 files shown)


khmakoto:
@natalieethell thanks for the info, I'll close this PR and create another one that removes that from the component description and the unnecessary logic from styles.

natalieethell:
Thanks @khmakoto!

PR #11831, opened 2020-01-29
Text: Updating docs and style logic to remove 'inherit' by default behavior

### LINKED ISSUE #11801: Text component does not inherit parent style definitions when no variant is supplied

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11801

#### Description of changes

The `Text` component had some outdated documentation about the component inheriting its parent font styles while this is no longer the case. This PR updates this documentation and fixes some logic in the styles that still accounted for this old behavior.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11831)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Text/Text.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Text/Text.styles.ts
@@ -12,12 +12,12 @@ export const TextStyles: ITextComponent['styles'] = (props: ITextProps, theme: I
       theme.fonts.medium,
       {
         display: block ? (as === 'td' ? 'table-cell' : 'block') : 'inline',
-        fontFamily: (variantObject && variantObject.fontFamily) || 'inherit',
-        fontSize: (variantObject && variantObject.fontSize) || 'inherit',
-        fontWeight: (variantObject && variantObject.fontWeight) || 'inherit',
-        color: (variantObject && variantObject.color) || 'inherit',
-        mozOsxFontSmoothing: variantObject && variantObject.MozOsxFontSmoothing,
-        webkitFontSmoothing: variantObject && variantObject.WebkitFontSmoothing
+        fontFamily: variantObject.fontFamily,
+        fontSize: variantObject.fontSize,
+        fontWeight: variantObject.fontWeight,
+        color: variantObject.color,
+        mozOsxFontSmoothing: variantObject.MozOsxFontSmoothing,
+        webkitFontSmoothing: variantObject.WebkitFontSmoothing
       },
       nowrap && {
         whiteSpace: 'nowrap',
--- a/change/@uifabric-experiments-2020-01-29-15-00-12-textDocumentation.json
+++ b/change/@uifabric-experiments-2020-01-29-15-00-12-textDocumentation.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Updating snapshots in experiments package.",
+  "packageName": "@uifabric/experiments",
+  "email": "Humberto.Morimoto@microsoft.com",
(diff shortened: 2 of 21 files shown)


packages/office-ui-fabric-react/src/components/Text/Text.styles.ts
```diff
@@ -12,12 +12,12 @@ export const TextStyles: ITextComponent['styles'] = (props: ITextProps, theme: I
...
-        fontFamily: (variantObject && variantObject.fontFamily) || 'inherit',
-        fontSize: (variantObject && variantObject.fontSize) || 'inherit',
-        fontWeight: (variantObject && variantObject.fontWeight) || 'inherit',
-        color: (variantObject && variantObject.color) || 'inherit',
-        mozOsxFontSmoothing: variantObject && variantObject.MozOsxFontSmoothing,
-        webkitFontSmoothing: variantObject && variantObject.WebkitFontSmoothing
+        fontFamily: variantObject.fontFamily,
+        fontSize: variantObject.fontSize,
+        fontWeight: variantObject.fontWeight,
+        color: variantObject.color,
```
natalieethell: Right now, I don't believe any of the default font styles include color. I wonder if this should instead default to a color from the theme.

natalieethell: Not sure if @betrue-final-final would still be the right person to ask about this? Maybe @dzearing?

khmakoto: Maybe, but I think having it like this will at least have the same effect as before since `undefined` colors will just inherit the colors of the parent.

betrue-final-final: David is an engineer, and I’m a designer. I don’t think they are interchangeable perspectives. 

In any event, what are we suggesting?

natalieethell: Of course! By tagging you and David, I more so meant that I'm not sure if this is an engineering or a design question.

Should the Text component inherit its styles from its parent? Or should it default to a semantic slot?

natalieethell: @khmakoto good point!

khmakoto: @natalieethell after speaking with @betrue-final-final we've leaned towards just letting the color be undefined and have the default inherited color so that we keep consistency before/after the PR instead of introducing the color here because, while it might be the right thing to do in the long term, it might muddle this PR with unintended things breaking.

mtennoe: Does this mean the PR is ready-to-go as is?

khmakoto: Yes, it should be, @natalieethell has already approved.

PR #11833, opened 2020-01-30
GroupedList: Adding aria-expanded state to the focus target of groups

### LINKED ISSUE #11816: Details List: Group header doesn't announce Expand/Collapse state
### Environment Information

### Describe the issue:

When Narrator is turned on and the user uses keyboard navigation, when keyboard focus lands on the group header its expand/collapse state isn't announced.

1. Navigate to [this](https://developer.microsoft.com/en-us/fabric#/controls/web/detailslist/grouped) link.
2. Turn on Narrator.
3. Using keyboard navigation go to the group header.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11816

#### Description of changes

When using keyboard navigation on the `GroupedList` and arriving on a group, the tab stop was on the entire group itself, which was not announcing the collapsed/expanded state of the group because this information was only being added to the chevron button. This PR adds this information on the top-level tab stop so that screen reader users can know this when navigating through `GroupedLists`.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11833)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
@@ -90,7 +90,7 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
         className={this._classNames.root}
         style={viewport ? { minWidth: viewport.width } : {}}
         onClick={this._onHeaderClick}
-        aria-expanded={group ? !group.isCollapsed : undefined}
+        aria-expanded={!group.isCollapsed}
         aria-label={group.ariaLabel || group.name}
         data-is-focusable={true}
       >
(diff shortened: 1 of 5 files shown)


packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
```diff
@@ -90,6 +90,7 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
         className={this._classNames.root}
         style={viewport ? { minWidth: viewport.width } : {}}
         onClick={this._onHeaderClick}
+        aria-expanded={group ? !group.isCollapsed : undefined}
```
KevinTCoughlin:
```suggestion
        aria-expanded={!group.isCollapsed}
```

We're guaranteed to have a `group` at this point because of the early return at https://github.com/OfficeDev/office-ui-fabric-react/blob/6fcf7c822401e334ba3bbc05a2ed5dd32936708b/packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx#L86

Otherwise LGTM!

PR #11836, opened 2020-01-30
ColorPicker: Adding transparency slider

navkuma:
#### Pull request checklist

#### Description of changes

Added logic to render alpha slider as transparency slider in the color picker control while maintaining backword compatible changes. This is needed for the color picker dialog being used in the PowerPoint online  

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11836)


packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
```diff
@@ -137,6 +147,12 @@ export interface IColorPickerStrings {
    */
   alpha?: string;
 
+  /**
+  Label for the transparency text field.
+  * @defaultvalue Transparency
+  */
+  transparency?: string;
```
sachinma: I think you should not have a separate prop for transparency label here as well as above. We can repurpose alpha label as transparency label. Or Alternatively, you should also take transparencyAriaLabel as an input

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -244,13 +253,14 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
...
   private _onAChanged = (ev: React.MouseEvent<HTMLElement>, a: number): void => {
-    this._updateColor(ev, updateA(this.state.color, Math.round(a)));
+    this._updateColor(ev, updateA(this.state.color, this._getSliderValue(Math.round(a))));
   };
 
   private _onTextChange(component: keyof IRGBHex, event: React.FormEvent<HTMLInputElement>, newValue?: string): void {
     const color = this.state.color;
     const isHex = component === 'hex';
     const isAlpha = component === 'a';
+    newValue = isAlpha ? String(this._getSliderValue(Number(newValue))) : newValue;
```
sachinma: You should do it later in code since we have not validated that newValue is a valid number

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -357,6 +367,10 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
       this.setState({ color: newColor, editingColor: undefined });
     }
   }
+
+  private _getSliderValue(alpha: number): number {
```
sachinma: nit: the method name is getSliderValue, but this is used both to get A value from Slider and slider value from A. See if you can name better

packages/office-ui-fabric-react/src/utilities/color/interfaces.ts
```diff
@@ -1,37 +1,40 @@
...
+}
+
+export interface IColor extends IRGB, IHSV {
+  /** Hex string for the color (excluding alpha component), *not* prefixed with #. */
+  hex: string;
+
+  /** CSS color string. If a hex value, it must be prefixed with #. */
+  str: string;
+
+  /* transparency*/
```
ecraig12345: I would prefer not to have a separate value in IColor for transparency because it makes a lot of internal logic unnecessarily complicated. Instead, just use the alpha value and if the consumer wants to display it as transparency they can do the conversion themselves. Or we can add a helper utility `getTransparency(color: Pick<IRGB, 'a'>)` .

Also in general, you need to use `/** doc comments */` (with two leading `**`) to make this show up properly in the documentation, and be more descriptive with the comment.

ecraig12345: Never mind, I see this is removed now.

packages/office-ui-fabric-react/src/components/ColorPicker/examples/ColorPicker.Basic.Example.tsx
```diff
@@ -1,75 +1,89 @@
...
-      color = updateA(this.state.color, 100);
-    }
-    this.setState({ alphaSliderHidden: !!checked, color });
-  };
-
-  private _onShowPreviewBoxClick = (ev: React.MouseEvent<HTMLElement>, checked?: boolean) => {
-    this.setState({ showPreview: !!checked });
-  };
-}
+import * as React from 'react';
```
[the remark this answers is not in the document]
ecraig12345: Never mind, this was fixed after I wrote the comment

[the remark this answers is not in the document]
ecraig12345: But please still ensure that your `core.autocrlf` setting is configured properly, as described in that comment.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
```diff
@@ -41,6 +41,10 @@ export interface IColorPickerProps {
    */
   alphaSliderHidden?: boolean;
 
+  /**
+  whether require a transparency slider.
```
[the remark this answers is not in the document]
ecraig12345: This prop name was also updated since I wrote the comment--`showAlphaAsTransparencySlider` is clearer, but I'd prefer something less verbose if possible.

The issue with the comment formatting still needs to be fixed, and the description needs more detail about what "transparency" means vs "alpha." (And my overarching comment *please discuss changes like this with the team beforehand* certainly still applies.)

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
ecraig12345: I could go either way on this. The benefit of a separate label is that it's clearer how to properly make a transparency slider, whereas the benefit of reusing the alpha label is that it's a smaller API surface overall.

If you keep the separate label, you should add a comment that it's only used when the certain prop (whatever we decide to call it) is true. Or if you combine the labels, you should update the comments on `a` and `alphaLabel` to indicate that they're also used for transparency when the prop is true.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -118,8 +119,11 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
     const props = this.props;
     const strings = this._strings;
     const textLabels = this._textLabels;
-    const { theme, className, styles, alphaSliderHidden } = props;
+    const { theme, className, styles, alphaSliderHidden, showAlphaAsTransparencySlider } = props;
     const { color } = this.state;
+    const alphaValue = color.a;
+    const sliderValue = alphaValue ? this._getSliderValue(alphaValue) : 100;
+    textLabels.a = showAlphaAsTransparencySlider ? props.transparencyLabel || strings.transparency : props.alphaLabel || strings.alpha;
```
ecraig12345: Please don't mutate `textLabels`! (And in general, you shouldn't be doing side effects like this in the render method.)

Instead, update the type of `textLabels` like this, and set `transparency` along with the other values in the constructor. 
`private _textLabels: { [K in keyof IRGBHex]: string; transparency: string; };`

Then update the logic here to decide which of `textLabels` to use.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
ecraig12345: If you do keep the separate `transparency` label, you should add an additional one `transparencyAriaLabel` in case anyone would like to set the aria label differently from the visual label.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
navkuma: Done.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -118,8 +118,10 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
     const props = this.props;
     const strings = this._strings;
     const textLabels = this._textLabels;
-    const { theme, className, styles, alphaSliderHidden } = props;
+    const { theme, className, styles, alphaSliderHidden, useTransparencySlider } = props;
     const { color } = this.state;
+    const alphaValue = color.a;
+    const sliderValue = alphaValue ? this._getAlphaOrTransValue(alphaValue) : 100;
```
sachinma: why do you have this condition, doesn't look correct. If existing alpha is zero - then that is a valid value

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -191,7 +194,7 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
                 <td>{textLabels.r}</td>
                 <td>{textLabels.g}</td>
                 <td>{textLabels.b}</td>
-                {!alphaSliderHidden && <td>{textLabels.a}</td>}
+                {!alphaSliderHidden && <td>{useTransparencySlider ? props.alphaLabel : textLabels.a}</td>}
```
sachinma: Lets not change here. Lets update textLabel.a in the constructor as required

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
```diff
@@ -41,6 +41,13 @@ export interface IColorPickerProps {
    */
   alphaSliderHidden?: boolean;
 
+  /**
+   * When useTransparencySlider is true, this will change the alpha slider to transparency slider.
+   * Alpha value represents the opacity of the color whereas transparency reperesnts the tranparentness of the color,
+   * i.e. 30% transparent color has 70% opaqueness.
+   */
+  useTransparencySlider?: boolean;
```
sachinma: I still feel showAlphaAsTransparency

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
```diff
@@ -41,6 +41,13 @@ export interface IColorPickerProps {
    */
   alphaSliderHidden?: boolean;
 
+  /**
+   * When useTransparencySlider is true, this will change the alpha slider to transparency slider.
+   * Alpha value represents the opacity of the color whereas transparency reperesnts the tranparentness of the color,
```
sachinma: Typo in tranparentness

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
navkuma: can't change textLabel.a in the constructor based on props value since its initialized only once.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
navkuma: accepted the @ecraig12345  suggestion to make it less verbose, i believe useTransparencySlider is less verbose and precise.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
```diff
@@ -132,13 +140,15 @@ export interface IColorPickerStrings {
 
   /**
    * Label for the alpha text field.
+   * Also used as transparency when useTransparencySlider is true.
    * Also used as the aria label for the alpha slider if `alphaAriaLabel` is not provided.
    * @defaultvalue Alpha
```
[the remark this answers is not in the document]
navkuma: not sure if defaultValue can have multiple value's. also by default useTransparencyslider is false by default and alpha label will have "alpha" value by default

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
navkuma: re-purposing alpha label for transparency as well now!!

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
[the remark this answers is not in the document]
navkuma: done.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
navkuma: fixed.

navkuma:
@ecraig12345 Sure will keep in mind about the process of submitting new feature request in future.

navkuma:
@ecraig12345  npm pre-commit hook lint-staged is adding space to the end of this file **change/office-ui-fabric-react-2020-01-24-16-29-51-dev-vishgup-groupchanges.json** leading to merge conflict. Any suggestion to stop this?

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.types.ts
navkuma: @ecraig12345 i have added separate label "transparency" and separate aria label "transparencyAriaLabel" for better clarity.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -244,7 +258,7 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
   };
 
   private _onAChanged = (ev: React.MouseEvent<HTMLElement>, a: number): void => {
-    this._updateColor(ev, updateA(this.state.color, Math.round(a)));
+    this._updateColor(ev, updateA(this.state.color, this._getAlphaOrTransValue(Math.round(a))));
```
ecraig12345: When using a transparency slider, is the alpha value in the resulting color meant to be interpreted as transparency? If so, you should update the docs for IColor with a note about that.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorSlider/ColorSlider.types.ts
```diff
@@ -82,6 +82,12 @@ export interface IColorSliderProps {
    * Call to provide customized styling that will layer on top of the variant rules.
    */
   styles?: IStyleFunctionOrObject<IColorSliderStyleProps, IColorSliderStyles>;
+
+  /**
+   * If true, the alpha slider represents a transparency slider
+   * @defaultvalue false
+   */
+  useTransparencySlider?: boolean;
```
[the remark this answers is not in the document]
navkuma: In my opinion, useTransparencySlider is quite verbose and gives clear indication caller what to expect.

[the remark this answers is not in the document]
ecraig12345: In this case it's *too* verbose (the component is a slider, so we don't need that in the prop name--whereas it does make sense to include "slider" in the color picker prop name) and doesn't match the existing similar prop name `isAlpha`.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -118,8 +121,10 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
     const props = this.props;
     const strings = this._strings;
     const textLabels = this._textLabels;
-    const { theme, className, styles, alphaSliderHidden } = props;
+    const { theme, className, styles, alphaSliderHidden, useTransparencySlider } = props;
     const { color } = this.state;
+    const alphaValue = color.a;
+    const sliderValue = alphaValue == null ? 100 : this._getAlphaOrTransValue(alphaValue);
```
ecraig12345: Why the new default value logic here? Also this var should be called alphaSliderVa

ecraig12345: Why the new default value logic here? Also this var should be called alphaSliderValue to be clearer since there are two sliders.

packages/office-ui-fabric-react/src/utilities/color/interfaces.ts
```diff
@@ -6,7 +6,9 @@ export interface IRGB {
   g: number;
   /** Blue, range 0-255. */
   b: number;
-  /** Alpha, range 0 (transparent)-100. Usually assumed to be 100 if not specified. */
+  /** Alpha, range 0 (transparent)-100. Usually assumed to be 100 if not specified.
+   * When useTransparencySlider is true, alpha value represents the transparentness
```
ecraig12345: Thinking about this more, it could confuse consumers to have a value "alpha" which is sometimes transparency. Also, none of the color conversion functions will handle that correctly.

Instead, I think it would be better to store the value as actual alpha, and add a public conversion function (in this folder) for getting transparency from a color. (The color slider can convert the alpha value to transparency only when displaying it.) Sorry for not providing this feedback sooner.

ecraig12345:
Sorry for all the delays here, but I really want to make sure that we do this the right way for our consumers, and in a way that we will be OK with supporting for the long term. This is the trouble with adding an API to a library: once it's added, you can't remove or change anything in a breaking way except in a major release, so we have to be very careful to do things right the first time.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorSlider/ColorSlider.types.ts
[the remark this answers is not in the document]
navkuma: Ahh Got it, Fixed.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
navkuma: fixed.

packages/office-ui-fabric-react/src/utilities/color/interfaces.ts
navkuma: So i have removed this comment, i misinterpreted your earlier comment regarding putting a comment in the IColor Interface. So at all times the value is alpha only, we are just showing transparency on UI end, but at all times the "alpha" variable actually holds the alpha value only and not the transparency.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
navkuma: at all times whether using transparency slider or not, the "alpha" hold the actual alpha value only and not the transparency. So no need of updating the docs for IColor.

dzearing:
This PR has gone stale. But also I am unsure of the point. What is the problem being solved? What is the difference in behavior? How is the user expected to use this?

ecraig12345:
This is a valid feature but closing in favor of my implementation in PR #12153.

PR #11841, opened 2020-01-30
Upgrade prettier to latest 1.19.1

xugao:
#### Pull request checklist

#### Description of changes
- Upgrade prettier which includes [fix for optional chaining error](https://github.com/prettier/prettier/issues/6595)
- Fixed minor console log mistake in `prettier.js`
- Run `prettier --all` to fix all files in the repo

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11841)

### DIFF
--- a/apps/fabric-website-resources/src/components/pages/DetailsList/DetailsListBasicPage.tsx
+++ b/apps/fabric-website-resources/src/components/pages/DetailsList/DetailsListBasicPage.tsx
@@ -2,6 +2,4 @@ import * as React from 'react';
 import { DemoPage } from '../../DemoPage';
 import { DetailsListBasicPageProps } from 'office-ui-fabric-react/lib/components/DetailsList/DetailsList.doc';
 
-export const DetailsListBasicPage = (props: { isHeaderVisible: boolean }) => (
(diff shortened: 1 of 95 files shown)


ecraig12345:
There will probably be some formatting updates with this new version, so it might be good to run prettier on all the files and check in any changes as part of this PR, to reduce noise in other PRs. (IIRC the command is `yarn prettier`.)

JasonGore:
@ecraig12345 Doesn't prettier only run against files actually changed instead of against the whole repo? I'm not sure if it's worthwhile adding all of that noise to this PR.

xugao:
@JasonGore @ecraig12345 - I kinda agree with Jason here. we can let people fix the error as we touch the files later. But honestly this change is not tooo bad/large. I am okay either way

JasonGore:
I'm pretty sure most of these changes would have occurred regardless of the prettier bump. And some people just don't seem to have pre-commit hooks working for one reason or another, so the issue of files not being prettiered consistently will persist either way.

PR #11844, opened 2020-01-31
Website: Fixing Card API reference tables not appearing on the component docs page

### LINKED ISSUE #11843: Card API docs missing from website
Noticed while trying to show someone the Card docs that the API reference tables are missing. (I can potentially help with this, mainly logging issue as reminder.) 
https://developer.microsoft.com/en-us/fabric#/controls/web/card

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11843

#### Description of changes

The API reference tables for the `Card` component weren't appearing on the website because it was looking only at pages generated from the `office-ui-fabric-react` package whereas the `Card` component is in the `@uifabric/react-cards` package. This PR fixes this issue by making the website look in both places for components.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11844)

### DIFF
--- a/apps/fabric-website/src/pages/Controls/ControlsAreaPage.tsx
+++ b/apps/fabric-website/src/pages/Controls/ControlsAreaPage.tsx
@@ -6,21 +6,16 @@ import { IPageJson } from 'office-ui-fabric-react/lib/common/DocPage.types';
 
 export interface IControlsPageProps extends IPageProps<Platforms> {}
 
-const apiRequireContext = [
-  require.context('@uifabric/api-docs/lib/pages/office-ui-fabric-react'),
-  require.context('@uifabric/api-docs/lib/pages/react-cards')
-];
+const apiRequireContext = require.context('@uifabric/api-docs/lib/pages/', true, /^(?!references).*/);
 
 const ControlsAreaPageBase: React.StatelessComponent<IControlsPageProps> = props => {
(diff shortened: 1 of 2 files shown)


apps/fabric-website/src/pages/Controls/ControlsAreaPage.tsx
```diff
@@ -6,16 +6,21 @@ import { IPageJson } from 'office-ui-fabric-react/lib/common/DocPage.types';
 
 export interface IControlsPageProps extends IPageProps<Platforms> {}
 
-const apiRequireContext = require.context('@uifabric/api-docs/lib/pages/office-ui-fabric-react');
+const apiRequireContext = [
+  require.context('@uifabric/api-docs/lib/pages/office-ui-fabric-react'),
```
ecraig12345: Oh actually, I have another idea which will scale better: go back to a single context and use negative lookahead! (I confirmed it works in all browsers.) 
`require.context('@uifabric/api-docs/lib/pages/', true, /^(?!references).*/)`

require.context reference: https://webpack.js.org/api/module-methods/#requirecontext

PR #11848, opened 2020-01-31
Dialog: Fix DialogContentProps titleProps bugs

xugao:
#### Pull request checklist

#### Description of changes
Fixing couple issues with `dialogContentProps.titleProps` which were missed in [this change](https://github.com/OfficeDev/office-ui-fabric-react/pull/11798):
1. deprecated `titleId` field is not being honored regardless whether `titleProps` is passed. because `titleProps.id` is passed with default Id from `Dialog` and it overwrites `titleId` inside `DialogContent`
 2. when `titleProps` is passed, it overwrites `titleProps={{ id: this._defaultTitleTextId }}` completely regardless if `titleProps` contains `id` or not

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11848)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Dialog/Dialog.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Dialog/Dialog.base.tsx
@@ -121,7 +121,11 @@ export class DialogBase extends React.Component<IDialogProps, {}> {
     const dialogContentProps: IDialogContentProps = {
       ...DefaultDialogContentProps,
       ...this.props.dialogContentProps,
-      draggableHeaderClassName: dialogDraggableClassName
+      draggableHeaderClassName: dialogDraggableClassName,
+      titleProps: {
+        id: this.props.dialogContentProps?.titleId || this._defaultTitleTextId,
+        ...this.props.dialogContentProps?.titleProps
+      }
     };
 
     const classNames = getClassNames(styles!, {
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/Dialog/Dialog.base.tsx
```diff
@@ -121,7 +121,11 @@ export class DialogBase extends React.Component<IDialogProps, {}> {
     const dialogContentProps: IDialogContentProps = {
       ...DefaultDialogContentProps,
       ...this.props.dialogContentProps,
-      draggableHeaderClassName: dialogDraggableClassName
+      draggableHeaderClassName: dialogDraggableClassName,
+      titleProps: {
+        id: this.props.dialogContentProps?.titleId || this._defaultTitleTextId,
+        ...this.props.dialogContentProps?.titleProps
+      }
```
JasonGore: Some snapshot tests may be good here with varying values of `titleId` and `titleProps`, including `undefined` values. That should help prevent regressions in the future.

JasonGore: (Also be good to test the tests by temporarily commenting out the change, verifying they fail, and then are fixed with your change. 😄 )

xugao: sure, let me add more coverage

PR #11850, opened 2020-01-31
GroupedList: Announcing position in set

### LINKED ISSUE #11818: Details List: Group header doesn't announce number of children
### Environment Information

### Describe the issue:

When a Narrator user navigates to a group header, the item doesn't announce how many children it has.

1. Navigate to [this](https://developer.microsoft.com/en-us/fabric#/controls/web/detailslist/grouped) page.
2. Turn Narrator on.
3. Using keyboard navigation go to the group header.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11818

#### Description of changes

Group headers in `GroupedLists` were not announcing their position relative to other groups at the same level. This PR adds the corresponding logic to do this, now announcing their position.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11850)

### DIFF
--- a/change/office-ui-fabric-react-2020-01-31-15-24-29-groupedListPosInSet.json
+++ b/change/office-ui-fabric-react-2020-01-31-15-24-29-groupedListPosInSet.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "GroupedList: Announcing position in set.",
+  "packageName": "office-ui-fabric-react",
+  "email": "Humberto.Morimoto@microsoft.com",
+  "commit": "666329aad1b362ee3d57ed8134f8398bab2347aa",
+  "dependentChangeType": "patch",
+  "date": "2020-01-31T23:24:29.313Z"
+}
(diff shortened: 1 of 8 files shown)


KevinTCoughlin:
@khmakoto should / can this value be inferred vs. passed-in via props?

khmakoto:
@KevinTCoughlin good question! This is what the aria specs says for each:

## aria-posinset

>Defines an element's number or position in the current set of listitems or treeitems. Not required if all elements in the set are present in the DOM. See related aria-setsize.
>
>If all items in a set are present in the document structure, it is not necessary to set this attribute, as the user agent can automatically calculate the set size and position for each item. However, if only a portion of the set is present in the document structure at a given moment, this property is needed to provide an explicit indication of an element's position.

## aria-setsize

>Defines the number of items in the current set of listitems or treeitems. Not required if all elements in the set are present in the DOM. See related aria-posinset.
>
>This property is marked on the members of a set, not the container element that collects the members of the set. To orient the user by saying an element is "item X out of Y," the assistive technologies would use X equal to the aria-posinset attribute and Y equal to the aria-setsize attribute.
>
>If all items in a set are present in the document structure, it is not necessary to set this property, as the user agent can automatically calculate the set size and position for each item. However, if only a portion of the set is present in the document structure at a given moment (in order to reduce document size), this property is needed to provide an explicit indication of set size.
>
>Authors MUST set the value of aria-setsize to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of aria-setsize to -1.

PR #11851, opened 2020-02-01
Update Jest and remove unused config files

ecraig12345:
#### Pull request checklist

#### Description of changes

Update Jest to ~24.9.0. Only a couple files needed to be updated as a result (see 33b1c3474b1b8e899a1c24a4c1ee973908686df1).

Update Node types to `^10.0.0`. 10 and 12 are both LTS now, so dropping 8 is reasonable.

Remove unused config files and deps:
- All `jsconfig.json` (doesn't seem to serve any purpose for our packages)
- `jest.config.js`, `config/tests.js`, `test` script, and test-related deps in packages which don't actually have tests
- `prettier.config.js` without any overrides (and deps on prettier-rules)

Now that most things don't depend on prettier-rules, I updated `invalidate-just-cache` to look at tslint-rules instead.

(Next up: entirely remove the prettier-rules package and use prettier's built-in overrides capability; [preview here](https://github.com/ecraig12345/office-ui-fabric-react/pull/13).)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11851)


scripts/invalidate-just-cache.js
```diff
@@ -1,7 +1,7 @@
 const fs = require('fs-extra');
 const path = require('path');
 
-const topLevelDep = path.join(__dirname, '../packages/prettier-rules/node_modules/.just');
+const topLevelDep = path.join(__dirname, '../packages/tslint-rules/node_modules/.just');
```
ecraig12345: @kenotron Will this work correctly?

JasonGore:
What is driving this change? We seem to have a lot of instability with testing in general over the past week and would like to try and make sure we stabilize that before we merge this (just to not introduce more variables.)

ecraig12345:
@JasonGore IIRC it was partly repo convergence plus generally trying to stay up to date (Jest 25 is out now, so going to latest 24 seemed like a reasonable intermediate step)

scripts/invalidate-just-cache.js
kenotron: I think this just picks a random cache to make sure to invalidate cache for all. This will work okay.

PR #11853, opened 2020-02-03
TagItem title fix

### LINKED ISSUE #11845: TagItem title renders `[object Object]` when children is not a string
### Environment Information

- **Package version(s)**: 7.85.0
- **Browser and OS versions**: All browsers

https://codepen.io/onlyann/pen/qBEzmQB

#### Actual behavior:

Hovering on the tag item displays _[object Object]_ when using a non string `children` prop to `TagPicker`.

#### Expected behavior:

Hovering on the tag item should either display the item name, or TagItem should allow title and aria-label to be passed in to override this behaviour.

The issue occurs because TagItem assumes that the `children` prop is a string and doesn't offer any way to specify our own title/aria-label

Are you willing to submit a PR to fix? Yes, if requested

Requested priority: Low

onlyann:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11845 

#### Description of changes

The TagItem component uses the `item.name` as the title and aria-label value instead of the `children` prop given it is not guaranteed to be a string.
`title` prop is now exposed to allow custom title / label.

#### Focus areas to test

TagPicker when `onResolveItem` uses the `TagItem` with a non string `children` prop. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11853)

### DIFF
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -7522,6 +7522,7 @@ export interface ITagItemProps extends IPickerItemProps<ITag> {
     enableTagFocusInDisabledPicker?: boolean;
     styles?: IStyleFunctionOrObject<ITagItemStyleProps, ITagItemStyles>;
     theme?: ITheme;
+    title?: string;
 }
 
 // @public
--- a/packages/office-ui-fabric-react/src/components/pickers/TagPicker/TagItem.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/TagPicker/TagItem.test.tsx
@@ -1,0 +1,41 @@
+import * as React from 'react';
+import * as renderer from 'react-test-renderer';
+
+import { TagItem } from './TagItem';
+import { resetIds } from '@uifabric/utilities';
+
+describe('TagItem', () => {
+  beforeEach(() => {
+    resetIds();
+  });
+
+  it('renders correctly', () => {
+    const component = renderer.create(
+      <TagItem item={{ name: 'Red', key: 'red' }} index={0}>
+        Red
+      </TagItem>
+    );
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders custom children correctly', () => {
+    const component = renderer.create(
+      <TagItem item={{ name: 'Red', key: 'red' }} index={0}>
+        <span>Red</span>
+      </TagItem>
+    );
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('accepts title override', () => {
+    const component = renderer.create(
+      <TagItem item={{ name: 'Red', key: 'red' }} title="Red color" index={0}>
+        Red color
+      </TagItem>
+    );
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+});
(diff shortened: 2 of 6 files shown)


onlyann:
2 DatePicker tests unrelated to this PR are broken on `master`, blocking the CI build.
`TypeError: Cannot read property '__reactInternalInstance$ekmc6a8uul7' of null`

packages/office-ui-fabric-react/src/components/pickers/TagPicker/TagItem.tsx
```diff
@@ -40,7 +41,7 @@ export const TagItemBase = (props: ITagItemProps) => {
       data-selection-index={index}
       data-is-focusable={(enableTagFocusInDisabledPicker || !disabled) && true}
     >
-      <span className={classNames.text} aria-label={children as string} title={children as string}>
+      <span className={classNames.text} aria-label={title} title={title}>
```
joschect: For consistency, children should still be passed into aria-label. That's where the string is expected to come in from and I think it helps keep that inline.

packages/office-ui-fabric-react/src/components/pickers/TagPicker/TagItem.test.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+import * as renderer from 'react-test-renderer';
+
+import { TagItem } from './TagItem';
+import { resetIds } from '@uifabric/utilities';
+
+describe('TagItem', () => {
```
joschect: Thanks for adding tests! :D

packages/office-ui-fabric-react/src/components/pickers/TagPicker/TagItem.tsx
onlyann: aria-label has the same constraint than the title where only a string should be passed in. 
Should `children` be passed if they represent a string as a default value, falling back to `props.item.name`, while still being overridable by `props.title`?

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -7520,6 +7520,7 @@ export interface ITagItemProps extends IPickerItemProps<ITag> {
     enableTagFocusInDisabledPicker?: boolean;
     styles?: IStyleFunctionOrObject<ITagItemStyleProps, ITagItemStyles>;
     theme?: ITheme;
+    title?: string;
```
aneeshack4: An example using this new title prop would be helpful

onlyann:
> @onlyann Any update on this PR? Looks like the DatePicker test needs to be updated and is causing the build to fail. Let us know if we can help you get this to a mergeable state.

I see that the DatePicker test has been fixed. I'll rebase from master to get it in a mergeable state.
I am still waiting on @joschect to reply to my latest suggestion.

packages/office-ui-fabric-react/src/components/pickers/TagPicker/TagItem.tsx
onlyann: I have changed the default value logic to keep the previous behaviour. What do you think?

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
onlyann: @aneeshack4 I added an example as part of the tests. Does that suit you?

aneeshack4: Yes since there aren't any existing examples on the TagPicker page on the docs site this is fine.

packages/office-ui-fabric-react/src/components/pickers/TagPicker/TagItem.tsx
joschect: Looks good!

packages/office-ui-fabric-react/src/components/pickers/TagPicker/TagItem.test.tsx
cliffkoh: Agree, thank you!!

Issue #11540, opened 2019-12-23
Can't tell if checkbox is checked in high-contrast-mode


omers4:
### Environment Information

- **Package version(s)**: 6.162.0
- **Browser and OS versions**: Edge 79, Windows 8

### Describe the issue:

When setting high contrast mode to "On" on Windows (black on white, white on black), it's impossible to tell if the checkbox is checked or not (it remains the same color and background). However, on hover the check mark is visible due to usage of -ms-high-contrast property.

### Please provide a reproduction of the issue in a codepen:

It's impossible to give a reproduction in codepen, since the high-contrast mode should be activated from windows settings. once you turn on high contrast mode, you will be able to see immediately that checkbox remains blank white/black when checking it, instead of showing the check mark.
For example, here
https://developer.microsoft.com/en-us/fabric#/controls/web/checkbox

#### Actual behavior:

When checking a checkbox in high contrast mode, it remains blank instead of showing the checkmark. (hover works, but the checking itself isn't)

#### Expected behaviour:

On high contrast mode, and when checkbox is checked, the check mark should be visible and the background should be inverted from what it was.

### Documentation describing expected behavior

https://developer.microsoft.com/en-us/fabric#/controls/web/checkbox

cliffkoh:
Hi @omers4, I cannot repro with the traditional Microsoft Edge (I can repro with Chromium Edge - but Chromium Edge isn't quite at parity with high contrast handling yet). This is what I see..

![image](https://user-images.githubusercontent.com/1003246/71378235-faf84080-257b-11ea-91be-f9595879fb1b.png)

omers4:
Hi, I did use Edge Chromium. Thought it will behave the same as Edge.
Thank you

cliffkoh:
Hi @omers4, Edge Chromium does not have support for High Contrast at the same parity as Edge yet. It is something they're working on though... You could use file feedback to Edge Chromium for this issue you observed :)

xugao:
possibly related to #11742

PR #11859, opened 2020-02-03
Fix high contrast in new Edge: TextField, Checkbox, Shimmer

### LINKED ISSUE #11735: Shimmer: High contrast issue
### Environment Information

- **Package version(s)**: office-ui-fabric-react@7.78.0
- **Browser and OS versions**: Microsoft Chromium Edge, Windows 10

### Describe the issue:
When high contrast mode is on, shimmer is not working properly. 

#### Actual behavior:

Repro steps:
1- Turn on high contrast mode
2- Open Chromium Edge
3- Go to fabric UI and search for the shimmer (https://developer.microsoft.com/en-us/fabric#/controls/web/shimmer)
4- Go to any Shimmer example

ShimmerBug

#### Expected behavior:

Same as in Microsoft Edge (no Chromium):
ShimmerNoBug

### LINKED ISSUE #11742: High contrast in Edge Chromium
## Issue
Styles are broken in High contrast mode in the new chromium Edge browser, but works in old Edge browser. 

## Root cause
In old Edge, the `-ms-high-contrast` media query implies `-ms-high-contrast-adjust: none;`.
In new/chromium Edge, this is no longer the case and `-ms-high-contrast-adjust: none;` needs to be set explicitly.

## Fix
Setting `-ms-high-contrast-adjust: none` is to ask the browser to not override those styles when high contrast is enabled. We need to make sure we set `-ms-high-contrast-adjust: none;` everywhere we set `-ms-high-contrast` from now on.

xugao:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #11742
- [X] Addresses an existing issue: Fixes #11735, #11540

#### Description of changes
**Recap of the issue:**
In old Edge, the `-ms-high-contrast` media query implies `-ms-high-contrast-adjust: none` on each property inside the media query. In new/chromium Edge, this is no longer the case and `-ms-high-contrast-adjust: none` needs to be set explicitly for the properties inside HC media query to take effect.

**Heads-up:** 
The solution to this problem is not ideal. Ideally, this can be somehow addressed/improved at browser-level. However, after talking to folks in Edge, I don't see them offering anything concrete for us any time soon. This PR fixes couple issues raised by users and demo how the fix looks like. I also looking forward hearing alternative ways to fix.

**Principal of the fix:**
1. Increase bundle size as little as possible. Add minimal amount of new css to resolve the issue.
2. No breaking change for styles set by existing use cases: 
Turing off high contrast adjustment can be a breaking change to the styles user set today on our component. For the styles user set today which are expected to be adjusted, this change potentially turns off adjustments for those case and user will need to add styles to set colors explicitly for HC.

**The fix:**
1. Added a new css selector added which turns off the HC adjustment in new Edge, and **only** new Edge (this is ensure principal No.2 above). You can read more about `forced-color-adjust` [here](https://www.w3.org/TR/css-color-adjust-1/#forced). it serves the same purpose as `-ms-high-contrast-adjust: none` in the context of this change.
2. Applies the new css selector in places where set high contrast styles using `HighContrastSelector` so our high contrast styles work in new Edge.
3. Find places where colors were previously expected to get adjusted but no longer are. Explicitly set HC styles for those cases

**Prioritization of this fix / next steps:**
Going forward, we know we need to continue provide HC styles in new Edge. However, ATM I'd like to make fixes on demand (if user opens an issue on a particular components). Reasons being:
1. Developers are still realizing this breaking change made by Edge. I still think Edge team likely will come up with something that ease developers' life. But it will certainly take time. 
2. Based on my conversations with partner teams, a lot of them have not realized this issue yet or they have not started on fixing this in their apps yet.
 
I don't like how messy this fix is if it comes from us, so I think it's better for us to postpone the fix as long as possible in case better solutions pop up later.

Other related problem we have today is we don't have designs for HC. It's hard to tell in some cases what the expected experience should be. I am following up with Ben True more on this. I also think in the long-term, we should move away from using media query to set HC styles. Instead, we provide HC styles through theming.

#### Focus areas to test
- Old Edge
- New Chromium Edge

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11859)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.styles.ts
@@ -1,7 +1,7 @@
 import { ICheckboxStyleProps, ICheckboxStyles } from './Checkbox.types';
 import { HighContrastSelector, getGlobalClassNames, IStyle } from '../../Styling';
 import { IsFocusVisibleClassName } from '../../Utilities';
-import { getEdgeChromiumForcedStylesOffSelector } from '@uifabric/styling';
+import { getEdgeChromiumNoHighContrastAdjustSelector } from '@uifabric/styling';
 
 const GlobalClassNames = {
   root: 'ms-Checkbox',
@@ -234,7 +234,7 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
           [HighContrastSelector]: {
             borderColor: 'WindowText'
           },
-          ...getEdgeChromiumForcedStylesOffSelector()
+          ...getEdgeChromiumNoHighContrastAdjustSelector()
         }
       },
       indeterminate && {
--- a/packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.styles.ts
@@ -1,6 +1,7 @@
 import { ICheckboxStyleProps, ICheckboxStyles } from './Checkbox.types';
 import { HighContrastSelector, getGlobalClassNames, IStyle } from '../../Styling';
 import { IsFocusVisibleClassName } from '../../Utilities';
+import { getEdgeChromiumForcedStylesOffSelector } from '@uifabric/styling';
 
 const GlobalClassNames = {
   root: 'ms-Checkbox',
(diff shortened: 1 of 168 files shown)


packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.styles.ts
```diff
@@ -54,7 +54,12 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
       borderColor: disabled ? checkboxBorderColorDisabled : checkboxBorderIndeterminateColor,
       transitionProperty: 'border-width, border, border-color',
       transitionDuration: MS_CHECKBOX_TRANSITION_DURATION,
-      transitionTimingFunction: MS_CHECKBOX_TRANSITION_TIMING
+      transitionTimingFunction: MS_CHECKBOX_TRANSITION_TIMING,
+      selectors: {
+        [HighContrastSelector]: {
```
aneeshack4: Just to clarify, this is keeping the use of high contrast selectors to avoid a breaking change while adding the usage of the new css selector `forced-color-adjust` as a better alternative option?

xugao:
Update: I am going to hold off this change. Currently we/a11y testers are filing similar bugs to Anaheim Edge team. Hopefully browser will own and address this

PR #11870, opened 2020-02-04
Convert withViewport decorator to use hooks

ThomasMichon:
Converted the `withViewport` decorator component to use hooks, which ensures that it better-manages subscriptions to `resize` events using `ResizeObserver` or `window`. Also, added memoization of the render output against the `props` change to try to prevent 'thrashing'. Leveraged the new `useEffect` to move the layout measurement out of the critical phase of the React render pipeline as another attempt to cut down on re-renders.

Converted `withResponsiveMode` to use hooks for similar performance improvements, and added an `initializeResponsiveMode()` function which can be used to set the default responsive mode based on the current window size. This avoids having the default `responsiveMode` be used during the first render pass.

Unfortunately, in order to get access to the current `window` for the instance of `withResponsiveMode`, we need to insert a dummy `<div />` (with `display: none`) into the DOM. This triggers a round of snapshot updates.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11870)

### DIFF
--- a/packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
+++ b/packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
-import { findDOMNode } from 'react-dom';
 import { BaseDecorator } from './BaseDecorator';
-import { getWindow, hoistStatics, EventGroup } from '../../Utilities';
+import { EventGroup, getWindow, css } from '../../Utilities';
+import * as WithResponsiveModeStyles from './withResponsiveMode.scss';
 
 export interface IWithResponsiveModeState {
   responsiveMode?: ResponsiveMode;
@@ -53,53 +53,82 @@ export function initializeResponsiveMode(element?: HTMLElement): void {
 export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveMode }, TState>(
   ComposedComponent: new (props: TProps, ...args: any[]) => React.Component<TProps, TState>,
 ): any {
-  const resultClass = class WithResponsiveMode extends BaseDecorator<TProps, IWithResponsiveModeState> {
-    private _events: EventGroup;
+  // tslint:disable-next-line:function-name
+  function WithResponsiveModeBase(
+    props: TProps,
+    forwardedRef: React.Ref<React.Component<TProps, TState>>
+  ): JSX.Element | null {
+    const [responsiveMode = ResponsiveMode.unknown, setCurrentResponsiveMode] = React.useState<ResponsiveMode>(
+      _defaultMode || _lastMode || ResponsiveMode.large
+    );
 
-    constructor(props: TProps) {
-      super(props);
-      this._events = new EventGroup(this);
-      this._updateComposedComponentRef = this._updateComposedComponentRef.bind(this);
+    const setCurrentResponsiveModeRef = React.useRef<typeof setCurrentResponsiveMode>();
+    setCurrentResponsiveModeRef.current = setCurrentResponsiveMode;
 
-      this.state = {
-        responsiveMode: _defaultMode || _lastMode || ResponsiveMode.large,
+    React.useEffect(() => {
+      return () => {
+        setCurrentResponsiveModeRef.current = undefined;
       };
-    }
+    }, []);
 
-    public componentDidMount(): void {
-      this._events.on(window, 'resize', this._onResize);
-      this._onResize();
-    }
+    const rootRef = React.useRef<HTMLDivElement | null>(null);
 
-    public componentWillUnmount(): void {
-      this._events.dispose();
-    }
+    React.useLayoutEffect(() => {
+      // Use a layout effect instead of a newer effect to ensure compatibility with existing unit tests.
+      if (typeof window !== 'undefined') {
+        const events = new EventGroup(null);
 
-    public render(): JSX.Element | null {
-      const { responsiveMode } = this.state;
+        const onResize = () => {
+          const element = rootRef.current;
+          const currentWindow = (element && getWindow(element)) || window;
 
-      return responsiveMode === ResponsiveMode.unknown ? null : (
-        <ComposedComponent
-          ref={this._updateComposedComponentRef}
-          responsiveMode={responsiveMode}
-          {...(this.props as any)}
-        />
-      );
-    }
+          if (currentWindow) {
+            if (setCurrentResponsiveModeRef.current) {
+              setCurrentResponsiveModeRef.current(getResponsiveMode(currentWindow));
+            }
+          }
+        };
 
-    private _onResize = () => {
-      const element = findDOMNode(this) as Element;
-      const currentWindow = (element && getWindow(element)) || window;
-      const responsiveMode = getResponsiveMode(currentWindow);
+        onResize();
 
-      if (responsiveMode !== this.state.responsiveMode) {
-        this.setState({
-          responsiveMode,
-        });
+        // There is not a good way to subscribe to the component instance window.
+        events.on(window, 'resize', onResize);
+
+        return () => {
+          events.dispose();
+        };
       }
-    };
-  };
-  return hoistStatics(ComposedComponent, resultClass);
+    }, []);
+
+    const componentElement = React.useMemo(() => {
+      return responsiveMode === ResponsiveMode.unknown ? null : (
+        <>
+          <ComposedComponent {...(props as any)} ref={forwardedRef} responsiveMode={responsiveMode} />
+          {
+            // Mark the element with a class name so that it is clear in snapshots what has been added.
+          }
+          <div className={css('ms-withResponsiveMode', WithResponsiveModeStyles.root)} ref={rootRef} />
+        </>
+      );
+    }, [forwardedRef, props, responsiveMode]);
+
+    return componentElement;
+  }
+
+  const WithResponsiveMode = React.forwardRef(WithResponsiveModeBase);
+
+  /**
+   * Old-style component wrapper for consumption by existing callers.
+   * The contract of the decorator interface expects a component instance from the ref, not
+   * a function component.
+   */
+  class WithResponsiveModeComponent extends BaseDecorator<TProps, IWithResponsiveModeState> {
+    public render(): JSX.Element | null {
+      return <WithResponsiveMode {...(this.props as any)} ref={this._updateComposedComponentRef} />;
+    }
+  }
+
+  return WithResponsiveModeComponent;
 }
 
 function getResponsiveMode(currentWindow: Window | undefined): ResponsiveMode {
@@ -124,7 +153,7 @@ function getResponsiveMode(currentWindow: Window | undefined): ResponsiveMode {
     } else {
       throw new Error(
         'Content was rendered in a server environment without providing a default responsive mode. ' +
-          'Call setResponsiveMode to define what the responsive mode is.',
+        'Call setResponsiveMode to define what the responsive mode is.'
       );
     }
   }
--- a/packages/office-ui-fabric-react/src/utilities/decorators/withViewport.tsx
+++ b/packages/office-ui-fabric-react/src/utilities/decorators/withViewport.tsx
@@ -40,7 +40,6 @@ export interface IWithViewportProps {
 }
 
 const RESIZE_DELAY = 500;
-const MAX_RESIZE_ATTEMPTS = 3;
 
 /**
  * A decorator to update decorated component on viewport or window resize events.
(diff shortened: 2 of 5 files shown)


packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -40,74 +51,92 @@ export function setResponsiveMode(responsiveMode: ResponsiveMode | undefined): v
...
+    props: TProps & {
+      getWindow: () => Window | undefined;
+    },
+    forwardedRef: React.Ref<React.Component<TProps, TState>>
+  ): JSX.Element | null {
+    const [responsiveMode = ResponsiveMode.unknown, setCurrentResponsiveMode] = React.useState<ResponsiveMode>(
+      _defaultMode || _lastMode || ResponsiveMode.large
+    );
+
+    React.useLayoutEffect(() => {
```
xugao: should be `React.useEffect`

ThomasMichon: If I do that it a lot more unit tests will need a re-write, to wrap all logic with `renderer.act()`. Also, Enzyme doesn't do it synchronously, so we'd need a way to capture it.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -40,74 +51,87 @@ export function setResponsiveMode(responsiveMode: ResponsiveMode | undefined): v
...
-      this._onResize();
-    }
+    const rootRef = React.useRef<HTMLDivElement | null>(null);
 
-    public componentWillUnmount(): void {
-      this._events.dispose();
-    }
+    React.useLayoutEffect(() => {
+      if (typeof window !== 'undefined') {
+        const events = new EventGroup(null);
```
dzearing: Do you really need to use EventGroup here?

Instead, why not use add/removeEventListener?

dzearing: We want to deprecate EventGroup over time, now that hooks help with lifetime support.

Additionally we are introducing a `@fluentui/react-event-listener` package which will add `useEventListener` hook and a wrapper `EventListener` component for use in class based components.

But even then, in situations like this, i think just using native eventing works fine.

KevinTCoughlin: We only add a single event listener to `window` so they're equivalent in terms of risk of memory regression, which is my main concern. I've seen folks commonly forget to keep the two in sync. `EventGroup` is convenient in that regard for its `dispose` method.

Have we marked `EventGroup` as deprecated? We use it extensively across ODSP (sometimes without clean-up).

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
-import { findDOMNode } from 'react-dom';
 import { BaseDecorator } from './BaseDecorator';
-import { getWindow, hoistStatics } from '../../Utilities';
+import { EventGroup, getWindow, css } from '@uifabric/utilities';
```
ecraig12345: Should be this per convention:
```suggestion
import { EventGroup, getWindow, css } from '../../Utilities';
```
Also we're trying to get rid of EventGroup and use the new EventListener utility instead--see #11949 for example. (edit: I see there's a conversation about that elsewhere, but I agree with not using EventGroup here)

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -1,7 +1,7 @@
...
+import * as WithResponsiveModeStyles from './withResponsiveMode.scss';
```
ecraig12345: Since this is only one very simple style, can you just use inline styles instead of SCSS?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -30,6 +30,17 @@ let _defaultMode: ResponsiveMode | undefined;
  */
 let _lastMode: ResponsiveMode | undefined;
 
+/**
+ * Initializes the responsive mode to the current window size. This can be used to avoid
+ * a re-render during first component mount since the window would otherwise not be measures
```
ecraig12345:
```suggestion
 * a re-render during first component mount since the window would otherwise not be measured
```

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -30,6 +30,17 @@ let _defaultMode: ResponsiveMode | undefined;
...
+ * until after mounting.
+ */
+export function initializeResponsiveMode(): void {
+  if (typeof window !== 'undefined') {
```
ecraig12345: Should this use the getWindow utility?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -40,74 +51,87 @@ export function setResponsiveMode(responsiveMode: ResponsiveMode | undefined): v
...
-    constructor(props: TProps) {
-      super(props);
-      this._updateComposedComponentRef = this._updateComposedComponentRef.bind(this);
-
-      this.state = {
-        responsiveMode: _defaultMode || _lastMode || ResponsiveMode.large
-      };
-    }
+  // tslint:disable-next-line:function-name
+  function WithResponsiveModeBase(props: TProps, forwardedRef: React.Ref<React.Component<TProps, TState>>): JSX.Element | null {
```
ecraig12345: Is there a reason for using function syntax here instead of const and a lambda, like is typically done for function components?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -40,74 +51,87 @@ export function setResponsiveMode(responsiveMode: ResponsiveMode | undefined): v
...
+    const [responsiveMode = ResponsiveMode.unknown, setCurrentResponsiveMode] = React.useState<ResponsiveMode>(
```
ecraig12345: In what scenario would you expect responsiveMode to be undefined?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ecraig12345: Since this hook is not commonly used, can you add a comment explaining why you're using it here?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -40,74 +51,87 @@ export function setResponsiveMode(responsiveMode: ResponsiveMode | undefined): v
...
+    const componentElement = React.useMemo(() => {
       return responsiveMode === ResponsiveMode.unknown ? null : (
-        <ComposedComponent ref={this._updateComposedComponentRef} responsiveMode={responsiveMode} {...(this.props as any)} />
+        <>
+          <ComposedComponent {...(props as any)} ref={forwardedRef} responsiveMode={responsiveMode} />
+          <div className={css('ms-withResponsiveMode', WithResponsiveModeStyles.root)} ref={rootRef} />
+        </>
       );
-    }
+    }, [forwardedRef, props, responsiveMode]);
```
ecraig12345: Is memoizing on object identity of props the correct thing to do here?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -40,74 +51,87 @@ export function setResponsiveMode(responsiveMode: ResponsiveMode | undefined): v
...
+        };
+      }
+    }, []);
+
+    const componentElement = React.useMemo(() => {
       return responsiveMode === ResponsiveMode.unknown ? null : (
-        <ComposedComponent ref={this._updateComposedComponentRef} responsiveMode={responsiveMode} {...(this.props as any)} />
+        <>
+          <ComposedComponent {...(props as any)} ref={forwardedRef} responsiveMode={responsiveMode} />
+          <div className={css('ms-withResponsiveMode', WithResponsiveModeStyles.root)} ref={rootRef} />
```
ecraig12345: Why are you adding a class name?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -40,74 +51,87 @@ export function setResponsiveMode(responsiveMode: ResponsiveMode | undefined): v
...
-        });
-      }
-    };
+  const WithResponsiveMode = React.forwardRef(WithResponsiveModeBase);
 
-    private _getResponsiveMode(): ResponsiveMode {
-      let responsiveMode = ResponsiveMode.small;
-      const element = findDOMNode(this) as Element;
-      const win = getWindow(element);
+  class WithResponsiveModeComponent extends BaseDecorator<TProps, IWithResponsiveModeState> {
```
ecraig12345: Why does this need to be a class?

packages/office-ui-fabric-react/src/utilities/decorators/withViewport.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { BaseDecorator } from './BaseDecorator';
 import { findScrollableParent, getRect, getWindow } from '../../Utilities';
+import { EventGroup, Async } from '@uifabric/utilities';
```
ecraig12345: Combine this with the previous import. And if possible it would be better not to use event group.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ThomasMichon: I need to backport these changes to Fabric 6.0, so it's best not to get carried away with the refactor and usage of new dependencies.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ThomasMichon: Shouldn't we standardize one way or the other? If I did an inline style, I'd still wind up putting it into a module-level `const`, which is equivalent to using CSS modules, I think.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ThomasMichon: That requires an element to be provided, which the caller won't typically have. But I'll modify this function to accept one optionally.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ThomasMichon: Isn't `function` syntax better? If you use the `const` form you have to declare the `I___Props` type twice, among other things.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ThomasMichon: Do we have a 'shallow-memo' hook? For the moment, this is sufficient to guard against changes to the internal state of the Fiber node.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ThomasMichon: This way its clear why there area bunch of extra `div` elements across all the snapshots.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ThomasMichon: The existing decorator returns `any`, but is expected to return a `React.Component` and not a function component.

packages/office-ui-fabric-react/src/utilities/decorators/withViewport.tsx
ThomasMichon: I need Fabric 6.0 compatibility.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ecraig12345: With our lint rules you shouldn't have to do it twice. I also prefer the const form since it makes it more explicit that the function is a component.
`const Foo: React.FunctionComponent<IFooProps> = props => { ... }`

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ThomasMichon: There is tons of usage of `EventGroup` directly across Fabric components. I think a cleaner way to go is to have someone start replacing it for the various scenarios, making it clear what to do in cases like this.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -53,53 +53,82 @@ export function initializeResponsiveMode(element?: HTMLElement): void {
...
+    const [responsiveMode = ResponsiveMode.unknown, setCurrentResponsiveMode] = React.useState<ResponsiveMode>(
+      _defaultMode || _lastMode || ResponsiveMode.large
+    );
+
+    const setCurrentResponsiveModeRef = React.useRef<typeof setCurrentResponsiveMode>();
+    setCurrentResponsiveModeRef.current = setCurrentResponsiveMode;
+
+    React.useEffect(() => {
+      return () => {
+        setCurrentResponsiveModeRef.current = undefined;
```
xugao, line 70: what is this about? `layoutEffect` runs before `useEffect`, if you set this back to undefined, isn't this making `onResize` 
no-op after first run?

ThomasMichon: This is only ever set to `undefined` when the whole component is unmounted.

xugao: ah right. it's in return statement :)

PR #11758, opened 2020-01-22
Add focus to Nav. Porting from #10310

liuqingbo:
#### Description of changes
Add focus to Nav component and this fix was ported from d1ef1ac (#10310)

PR #10310, opened 2019-08-29
Nav Accessibility: Expose focus public method so that caller can explicitly set focus to Nav component.

inateeg:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [ ] Include a change request file using `$ yarn change`

When Nav component is rendered, there is property 'isOnTop' can by passed in when the Nav is rendered as popup.  IN SPO case, this is when in T-narrow mode, Nav is only rendered when user click on Hamburg.  Under this condition, we would like Nav get focus since it is intentionally popup opened by user action.  This PR ensure when those conditions meet, Nav focusZone get focus so user can use arrow key to navigate through nodes.

(give an overview)

#### Focus areas to test

set isOnTop to true and verify the arrow key can navigate through the nav nodes while without this property set, that is not possible.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/10310)

PR #11878, opened 2020-02-05
Add focus to Nav. Porting from #10310 (#11758)

liuqingbo:
## Description of changes

Add focus to Nav component and this fix was ported from d1ef1ac (#10310)

#### Focus areas to test

(optional)

Cherry-picker from https://github.com/OfficeDev/office-ui-fabric-react/commit/a62910a58bec8368b07fef875ac27bfeab63668c

### DIFF
--- a/common/changes/office-ui-fabric-react/hotfix-AddFocus2Nav-office-ui-fabric-react-5.0_2020-01-22-05-26.json
+++ b/common/changes/office-ui-fabric-react/hotfix-AddFocus2Nav-office-ui-fabric-react-5.0_2020-01-22-05-26.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "office-ui-fabric-react",
+      "comment": "Nav: Add focus to Nav.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "office-ui-fabric-react",
+  "email": "qingliu@microsoft.com"
+}
--- a/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
@@ -2,11 +2,12 @@ import * as React from 'react';
 import {
   BaseComponent,
   classNamesFunction,
+  createRef,
   customizable,
   divProperties,
   getNativeProps
 } from '../../Utilities';
-import { FocusZone, FocusZoneDirection } from '../../FocusZone';
+import { FocusZone, FocusZoneDirection, IFocusZone } from '../../FocusZone';
 import { ActionButton } from '../../Button';
 import { Icon } from '../../Icon';
 import { buttonStyles } from './Nav.styles';
@@ -48,6 +49,7 @@ export class NavBase extends BaseComponent<INavProps, INavState> implements INav
     groups: null
   };
 
+  private _focusZone = createRef<IFocusZone>();
   constructor(props: INavProps) {
     super(props);
 
@@ -100,7 +102,7 @@ export class NavBase extends BaseComponent<INavProps, INavState> implements INav
     const classNames = getClassNames(getStyles!, { theme: theme!, className, isOnTop, groups });
 
     return (
-      <FocusZone direction={ FocusZoneDirection.vertical }>
+      <FocusZone direction={ FocusZoneDirection.vertical } ref={ this._focusZone }>
         <nav
           role='navigation'
           className={ classNames.root }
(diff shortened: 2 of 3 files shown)


common/changes/office-ui-fabric-react/hotfix-AddFocus2Nav-office-ui-fabric-react-5.0_2020-01-22-05-26.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
```
aneeshack4: The format of this change file looks off - should be like this: https://github.com/OfficeDev/office-ui-fabric-react/pull/10310/files#diff-e60e71efa905e67cc3e13c0fcfca39b0 did you run `yarn change`?

packages/office-ui-fabric-react/src/components/Nav/Nav.types.ts
```diff
@@ -11,6 +11,14 @@ export interface INav {
...
    */
   selectedKey: string | undefined;
+
+  /**
+   * Sets focus to the first tabbable item in the zone.
+   * @param forceIntoFirstElement - If true, focus will be forced into the first element, even
+   * if focus is already in the focus zone.
+   * @returns True if focus could be set to an active element, false if no operation was taken.
+   */
+  focus?: (forceIntoFirstElement?: boolean) => boolean;
```
aneeshack4: Where is the API file change? Did you run `yarn update-api`?

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
```diff
@@ -48,6 +49,7 @@ export class NavBase extends BaseComponent<INavProps, INavState> implements INav
     groups: null
   };
 
+  private _focusZone = createRef<IFocusZone>();
   constructor(props: INavProps) {
     super(props);
```
aneeshack4: If this is meant to be an exact port over, then initializeComponentRef(this) is missing right here

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
```diff
@@ -100,7 +102,7 @@ export class NavBase extends BaseComponent<INavProps, INavState> implements INav
     const classNames = getClassNames(getStyles!, { theme: theme!, className, isOnTop, groups });
 
     return (
-      <FocusZone direction={ FocusZoneDirection.vertical }>
+      <FocusZone direction={ FocusZoneDirection.vertical } ref={ this._focusZone }>
```
aneeshack4: Again if this is meant to be an exact port over, it should componentRef not just ref

packages/office-ui-fabric-react/src/components/Nav/Nav.types.ts
ecraig12345: This is the version 5 branch which doesn't have API extractor

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
ecraig12345: This should not be an exact port, because it's to version 5 which still uses BaseComponent.

common/changes/office-ui-fabric-react/hotfix-AddFocus2Nav-office-ui-fabric-react-5.0_2020-01-22-05-26.json
ecraig12345: This is fine since it's to version 5

ecraig12345: Actually I don't think a change file is needed in this special hotfix branch because the hotfix process is manual...not totally sure how it works

liuqingbo: Do I need delete the change file ? or keeping this one is also acceptable ?

ecraig12345:
@xugao This isn't going to build properly, which is fine since the 5.0 version built. My only question is whether it needs a change file or not, and then how to cue a build once it's merged.

common/changes/office-ui-fabric-react/hotfix-AddFocus2Nav-office-ui-fabric-react-5.0_2020-01-22-05-26.json
xugao: i see changes in the past have change file : https://github.com/OfficeDev/office-ui-fabric-react/commits/office-ui-fabric-react_v5.79.1.

cliffkoh: Honestly it isn't needed but it isn't harmful to have it. There is 0 automation for publishing in v5 setup. In v5 era, @dzearing does all hotfix publishing manually.

PR #11899, opened 2020-02-07
ISemanticColors: adding clarification to Fabric site.

### LINKED ISSUE #11896: DetailsList does not use SemanticColors "listText" property
### Environment Information

- **Package version(s)**: ^7.60.0

https://codepen.io/miwlekli/pen/jOPOQoK

#### Actual behavior:

Using createTheme I set listText in the semanticColors property of a ITheme object to be a certain color and DetailsList does not use the listText property. Instead it appears that IDetailsRow is setting the non-header text to be neutralSecondary.

#### Expected behavior:

The text in DetailsList rows should use listText semanticColor. If this is not the case the documentation should make this clear that this is not the use case of listText.

Requested priority: Normal

aneeshack4:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #11896

#### Description of changes

It's become clear that users are getting confused about the status of our semantic slots/palette slots usage. Consistent adoption of one over the other hasn't happened yet across the Fabric codebase due to many reasons, one being we haven't polished the semantic slots offering. We've done some polishes and some updates to use semantic slots over palette slots but not all. And we're currently working on making a better design token system overall with the new version of Fluent. In the meantime, this clarification can help our users understand this status. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11899)

### DIFF
--- a/packages/styling/src/interfaces/ISemanticColors.ts
+++ b/packages/styling/src/interfaces/ISemanticColors.ts
@@ -53,7 +53,7 @@ import { ISemanticTextColors } from './ISemanticTextColors';
  * and have a large and complex interaction surface.
  * This category covers all kinds of lists, whether they're typical one-item-per-row lists (like DetailsList) or ones with a tiled layout.
  *
- * Note: We are aware that the usage of semantic slots vs. palette slots is not consistent across the Fabric controls code and that both
+ * Note: We are aware that the usage of semantic slots vs. palette slots is not consistent across the Fabric controls code. Both
  * offerings need updates such as adding/removing slots. We've been actively making some of these changes. And we're working on a a better
  * solution that satisfies a lot of different factors including avoiding bloat of the already sizable offering. We'll update this page when
  * that's ready. Thank you for your patience.
--- a/packages/styling/src/interfaces/ISemanticColors.ts
+++ b/packages/styling/src/interfaces/ISemanticColors.ts
@@ -6,6 +6,10 @@ import { ISemanticTextColors } from './ISemanticTextColors';
  * The collection of all semantic slots for colors used in themes.
  *
  * Note: text colors are defined in ISemanticTextColors.ts.
+ * We are aware that the usage of semantic slots vs. palette slots is not consistent across the Fabric controls code.
+ * Both offerings need updates such as adding/removing slots. We've been actively making some of these changes.
+ * And we're working on a a better solution that satisfies a lot of different factors including avoiding bloat
+ * of the already sizable offering. We'll update this page when that's ready. Thank you for your patience.
  *
  * #### Naming Convention
  *
@@ -53,10 +57,6 @@ import { ISemanticTextColors } from './ISemanticTextColors';
  * and have a large and complex interaction surface.
  * This category covers all kinds of lists, whether they're typical one-item-per-row lists (like DetailsList) or ones with a tiled layout.
  *
- * Note: We are aware that the usage of semantic slots vs. palette slots is not consistent across the Fabric controls code. Both
- * offerings need updates such as adding/removing slots. We've been actively making some of these changes. And we're working on a a better
- * solution that satisfies a lot of different factors including avoiding bloat of the already sizable offering. We'll update this page when
- * that's ready. Thank you for your patience.
  * {@docCategory ISemanticColors}
  */
 export interface ISemanticColors extends ISemanticTextColors {
--- a/packages/styling/src/interfaces/ISemanticColors.ts
+++ b/packages/styling/src/interfaces/ISemanticColors.ts
@@ -6,11 +6,12 @@ import { ISemanticTextColors } from './ISemanticTextColors';
  * The collection of all semantic slots for colors used in themes.
  *
  * Note: text colors are defined in ISemanticTextColors.ts.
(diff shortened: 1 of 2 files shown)


packages/styling/src/interfaces/ISemanticColors.ts
```diff
@@ -52,6 +52,11 @@ import { ISemanticTextColors } from './ISemanticTextColors';
  * Lists differ from menus in that they are designed to show infinite amounts of items, often scroll,
  * and have a large and complex interaction surface.
  * This category covers all kinds of lists, whether they're typical one-item-per-row lists (like DetailsList) or ones with a tiled layout.
+ *
+ * Note: We are aware that the usage of semantic slots vs. palette slots is not consistent across the Fabric controls code. Both
```
xugao: we have `Note:` at the very top as well on Line 8. i think we should move this up

aneeshack4: moved to top now!

packages/styling/src/interfaces/ISemanticColors.ts
```diff
@@ -6,6 +6,10 @@ import { ISemanticTextColors } from './ISemanticTextColors';
  * The collection of all semantic slots for colors used in themes.
  *
  * Note: text colors are defined in ISemanticTextColors.ts.
+ * We are aware that the usage of semantic slots vs. palette slots is not consistent across the Fabric controls code.
+ * Both offerings need updates such as adding/removing slots. We've been actively making some of these changes.
```
phkuo: No, we never remove slots. It would break backwards compatibility.
We should not add more palette slots (besides neutralSecondaryAlt).

packages/styling/src/interfaces/ISemanticColors.ts
```diff
@@ -6,6 +6,10 @@ import { ISemanticTextColors } from './ISemanticTextColors';
...
+ * And we're working on a a better solution that satisfies a lot of different factors including avoiding bloat
```
phkuo: Shouldn't start a new sentence with "and".

packages/styling/src/interfaces/ISemanticColors.ts
```diff
@@ -6,6 +6,10 @@ import { ISemanticTextColors } from './ISemanticTextColors';
  * The collection of all semantic slots for colors used in themes.
  *
  * Note: text colors are defined in ISemanticTextColors.ts.
+ * We are aware that the usage of semantic slots vs. palette slots is not consistent across the Fabric controls code.
```
phkuo: I wouldn't say it's not consistent, that implies that slot usage is incorrect. Rather, I'd say something about how we still use both, whereas the desired state (I hope) is to move completely to semantic slots. (We should also state our desired end goal.)

aneeshack4: Ok addressed this and the above comments with a rewrite!

packages/styling/src/interfaces/ISemanticColors.ts
```diff
@@ -6,6 +6,11 @@ import { ISemanticTextColors } from './ISemanticTextColors';
  * The collection of all semantic slots for colors used in themes.
  *
  * Note: text colors are defined in ISemanticTextColors.ts.
+ * The plan is to move completely to semantic slots and replace all usage of fabric palette slots.
```
phkuo: "The plan" is a bit vague and informal. "We plan to..." or "We hope to...".
"Fabric" should be capitalized.

packages/styling/src/interfaces/ISemanticColors.ts
```diff
@@ -6,6 +6,11 @@ import { ISemanticTextColors } from './ISemanticTextColors';
...
+ * We still have a ways to go to complete this plan and we've been actively making some of these changes.
+ * At present, we're not adding any more slots to avoid bloating the already sizeable offering.
+ * We're also working on a better solution for design tokens overall in our next iteration, Fluent.
```
phkuo: Not sure we should publicly talk about Fluent yet. (Not to mention, it's an overloaded term right now!)

packages/styling/src/interfaces/ISemanticColors.ts
```diff
@@ -6,6 +6,11 @@ import { ISemanticTextColors } from './ISemanticTextColors';
  * The collection of all semantic slots for colors used in themes.
  *
  * Note: text colors are defined in ISemanticTextColors.ts.
+ * The plan is to move completely to semantic slots and replace all usage of fabric palette slots.
+ * We still have a ways to go to complete this plan and we've been actively making some of these changes.
+ * At present, we're not adding any more slots to avoid bloating the already sizeable offering.
```
phkuo: We are still adding new slots, if necessary. Our policy hasn't changed.

aneeshack4: I have this to address your earlier comment "We should not add more palette slots (besides neutralSecondaryAlt)."

aneeshack4: Just reworded it along with some other changes to clarify!

PR #11900, opened 2020-02-07
Combobox: Fix selected text not appearing to be selected

### LINKED ISSUE #4992: Combobox checked state applied when checked, or when hovered
### Bug Report

- __Package version(s)__: Current

#### Priorities and help requested (not applicable if asking question):

Are you willing to submit a PR to fix? NO

Requested priority: High

Products/sites affected: CALENDAR

#### Describe the issue:

There should be a visual difference between a selected menu item and a menu item that has been hovered on.

#### Actual behavior:

The "selected" value in the combobox is the same color as the "hover". 

#### Expected behavior:

I expect the selected value to be separate and to look like the selected state we have for dropdowns.

image

image

joschect:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #4992

#### Description of changes

The currently selected option now appears selected.

Before:
image

After:
image

Needs design assistance with the image below. Basically the selected option (current value) and the current active item (the one activated by keyboarding) have the same style. I am uncertain as to what the expected behavior is. Pressing escape with revert to the current value, so it makes sense that it should still be highlighted, but what color should it be? Overall this is still an improvement in clarity. @betrue-final-final or @jspurlin  could you weigh in on expected colors/behavior?

image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11900)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts
@@ -71,7 +71,7 @@ export const getOptionStyles = memoizeFunction(
     theme: ITheme,
     customStylesForAllOptions?: Partial<IComboBoxOptionStyles>,
     customOptionStylesForCurrentOption?: Partial<IComboBoxOptionStyles>,
-    isPending?: boolean,
+    isSelected?: boolean,
     isHidden?: boolean
   ): Partial<IComboBoxOptionStyles> => {
     const { palette, semanticColors } = theme;
@@ -89,7 +89,7 @@ export const getOptionStyles = memoizeFunction(
       root: [
         theme.fonts.medium,
         {
-          backgroundColor: isPending ? option.backgroundHoveredColor : 'transparent',
+          backgroundColor: isSelected ? option.backgroundSelectedColor : 'transparent',
           boxSizing: 'border-box',
           cursor: 'pointer',
           display: isHidden ? 'none' : 'block',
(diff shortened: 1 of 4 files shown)


joschect:
Note: I'll update snapshots after talking about what to do with keyboard selection

joschect:
Ping @jspurlin about expected highlighted behavior during keyboarding.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.styles.ts
```diff
@@ -131,7 +132,7 @@ export const getOptionStyles = memoizeFunction(
       },
       rootChecked: [
         {
-          backgroundColor: 'transparent',
+          backgroundColor: option.backgroundSelectedColor,
```
aneeshack4: LGTM - menuItemBackgroundChecked is in line with the other menuitem states root is being set to here

aneeshack4:
Would be good to get design input on the subject of the selected item and current active having the same styles but it looks like in Dropdown.styles.ts, for the active state, `menutItemTextHovered` so maybe ComboBox can do the same for consistency?

betrue-final-final:
That's a fun one. What could happen here, and should, is that there should only be a focus rectangle and not show a selection grey inside for the "active" item.

joschect:
This is now fixed and ready to go in. @jspurlin Could you review?
image

jspurlin:
> This is now fixed and ready to go in. @jspurlin Could you review?
> image

@joschect I think we want the full "selected" look to move with focus (as it always does) since it's the "pending selected value" (e.g. background-color should move with focus and also be on Option G)

joschect:
@jspurlin This matches with how dropdown keyboard selection works and was changed per @betrue-final-final 's suggestion. 
image

jspurlin:
That feels weird, like might not know where focus is because they both look like they could be focus.

Also, the newest screenshot seems to have opposite styling of your previous screenshot (it looks like the thing currently focused has the background-color and the previously selected element has the outline... that is unless moving focus doesn't update the input in dropdown which should be another bug that's logged/fixed ([native html select behavior](https://developer.mozilla.org/en-US/docs/web/html/element/select) updates the input to show the pending value)

khmakoto:
Closing until we get more consensus on how this should work.

ecraig12345:
@joschect Can you resolve the merge conflicts here so this can merge?

joschect:
Closing because there are still unresolved questions.

Issue #6310, opened 2018-09-11
Remove TSlint overrides in OUFR


micahgodbolt:
Wanted to get these on the backlog. Can be done as a single issue or broken up into multiple	

- [x] "deprecation": false, (#12045)
- [ ] "jsx-ban-props": false,
- [x] "max-line-length": [false, 140],
- [ ] "no-any": false,
- [ ] "typedef": [false]

micahgodbolt:
you'll probably run into a lot of errors upon turning each one of these off. probably safest to do 1 at a time along with any fixes needed to support that change.

cliffkoh:
@micahgodbolt I think enabling IE 11 screener tests is far more bang for buck than some of these tslint rules. `deprecation` is a good one though.

cliffkoh:
Also, is this really a good first issue?

micahgodbolt:
that's what david suggested. It's a 'turn off the exception' and fix what breaks. Rise repeat.

cliffkoh:
Fixing the breaks may be far from trivial though... But if Eddie is working on this already, we'll see how far he gets. I still think enabling IE 11 screener is much more valuable.

leddie24:
Do we actually want to remove `jsx-ban-props`?  It complains about the 

> Use of the prop 'style' is not allowed. Use className and provide css rules instead of using inline styles.

  Don't we want to use inline styles for some components and utilities (e.g `withViewport`)?

micahgodbolt:
hey @khmakoto, is this work still on your radar?

khmakoto:
I have one more change regarding the deprecation flag. Would have to talk with David to see if I will take on the work of other overrides or not.

khmakoto:
Not applicable anymore since we use eslint and not tslint now. Closing.

PR #11925, opened 2020-02-11
Removing use of some deprecated things by replacing them.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Contributes to #6310

#### Description of changes

This PR gets rid of some of the deprecated elements in the OUFR package.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11925)


packages/office-ui-fabric-react/src/components/Announced/Announced.types.ts
```diff
@@ -1,12 +1,11 @@
 import * as React from 'react';
-import { AnnouncedBase } from './Announced.base';
 import { IStyle } from '../../Styling';
 import { IStyleFunctionOrObject } from '../../Utilities';
 
 /**
  * {@docCategory Announced}
  */
-export interface IAnnouncedProps extends React.Props<AnnouncedBase>, React.HTMLAttributes<HTMLDivElement> {
+export interface IAnnouncedProps extends React.PropsWithChildren<React.HTMLAttributes<HTMLDivElement>> {
```
ecraig12345: Will this and the other changes be fully backwards compatible? Especially concerned about the React.Props changes.

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
```diff
@@ -87,7 +87,7 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
       id = this._fallbackId
     } = this.props;
     const { value, hasFocus } = this.state;
-    const placeholderValue = labelText === undefined ? placeholder : labelText;
+    const placeholderValue = placeholder || labelText;
```
ecraig12345: Could empty string be a valid value here?

packages/office-ui-fabric-react/src/components/pickers/BasePicker.types.ts
```diff
@@ -96,12 +95,12 @@ export interface IBasePickerProps<T> extends React.Props<any> {
...
    * @deprecated Use `inputProps.onFocus` instead
    */
-  onFocus?: React.FocusEventHandler<HTMLInputElement | BaseAutoFill>;
+  onFocus?: React.FocusEventHandler<HTMLInputElement>;
 
   /**
    * A callback for when the user moves the focus away from the picker
    */
-  onBlur?: React.FocusEventHandler<HTMLInputElement | BaseAutoFill>;
+  onBlur?: React.FocusEventHandler<HTMLInputElement>;
```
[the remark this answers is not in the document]
ecraig12345: I'm also a bit concerned since it seems like this could be a breaking change.

packages/office-ui-fabric-react/src/components/Announced/Announced.types.ts
JasonGore: I'm with Elizabeth here, this looks like it could be a breaking change as it removes `key` and `ref` from typings. I think it would be ok as long as we can claim correctness for any typing breaks (i.e. `key` and `ref`, if used, have absolutely no effect and shouldn't be supplied anyways), but I also want to make sure we shouldn't look at using other types like `PropsWithRef` in each of these cases.

If and when we do introduce a change explicitly adding children to props, we should rationalize why. [The typings automatically add `children` in the component definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/267dc9f23ae19f38f414ee02f8d7ac738aabad60/types/react/index.d.ts#L518), making it available when children are used in JSX. I have a feeling any `WithChildren` types should be explicitly used in nearly all component typings or defer to React component types. I'm guessing it will be the latter. 

Overall, I think this change should be part of a larger component typing strategy suitable for component typing guidelines (ref usage, ref forwarding, children specification, key usage, etc.) and conformance testing for our next major release.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.types.ts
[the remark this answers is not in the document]
khmakoto: Yes, that's probably what we should do.

packages/office-ui-fabric-react/src/components/Announced/Announced.types.ts
khmakoto: I'll roll back on these changes.

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
khmakoto: I made this change because `labelText` is the deprecated thing here, which was taking precedence if it was not `undefined`. I can change it so that this is `placeholder !== undefined ? placeholder : labelText` but I think we should still change the order since `placeholder` should take precedence over the deprecated `labelText`. What do you think about that approach @ecraig12345?

packages/office-ui-fabric-react/src/components/Announced/Announced.types.ts
khmakoto: Reverted these changes.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.types.ts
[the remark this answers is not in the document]
khmakoto: Alright, I went ahead and used what @JasonGore suggested.

change/office-ui-fabric-react-2020-02-06-17-42-11-deprecation.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "Removing use of some deprecated things by replacing them.",
```
dzearing: Can you make this comment more useful? It goes into release notes. :)

What components are being updated"?

khmakoto: Sure, I'll update the change file.

PR #11929, opened 2020-02-12
Making "gulp docs" work

kenotron:
#### Pull request checklist

#### Description of changes

This required several things to be modified:

1. @ecraig12345 mentioned that tsconfig.json extends should use the relative path for now to generate correct .d.ts files
2. fixes to docs src that import from '../../../packages/react/package.json' to '@fluentui/react/package.json'
3. pegging typescript eslint-plugin to a specific version because our own FUI eslint-plugin somehow is dependent on some now-gone deprecated internal APIs (dangerous to take that on as a dependency)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11929)

### DIFF
--- a/packages/fluentui/docs/src/views/Introduction.tsx
+++ b/packages/fluentui/docs/src/views/Introduction.tsx
@@ -1,10 +1,10 @@
-import * as _ from 'lodash'
-import * as React from 'react'
-import { NavLink } from 'react-router-dom'
-import { Flex, Header } from '@fluentui/react'
+import * as _ from 'lodash';
+import * as React from 'react';
+import { NavLink } from 'react-router-dom';
+import { Flex, Header } from '@fluentui/react';
 
-import pkg from '../../../packages/react/package.json'
-import Logo from '../components/Logo/Logo'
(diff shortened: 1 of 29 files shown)


packages/fluentui/docs/src/views/Introduction.tsx
```diff
@@ -1,10 +1,10 @@
-import * as _ from 'lodash'
-import * as React from 'react'
-import { NavLink } from 'react-router-dom'
-import { Flex, Header } from '@fluentui/react'
+import * as _ from 'lodash';
```
ecraig12345: Did you format all the files with prettier or is that going to happen randomly as they're edited as we go along?

kenotron: I am formatting with prettier as we go along for now - but I think near at the final import, we'll prettier the whole thing

PR #11930, opened 2020-02-12
Link: text decoration should be removed regardless of link vs button.

dzearing:
Tried providing an example of react-router to a developer, found that links were being underlined if buttons.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11930)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Link/Link.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Link/Link.styles.ts
@@ -26,6 +26,8 @@ export const getStyles = (props: ILinkStyleProps): ILinkStyles => {
         outline: 'none',
         fontSize: 'inherit',
         fontWeight: 'inherit',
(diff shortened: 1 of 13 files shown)


packages/office-ui-fabric-react/src/components/Link/Link.styles.ts
```diff
@@ -67,9 +69,7 @@ export const getStyles = (props: ILinkStyleProps): ILinkStyles => {
           }
         }
       },
-      !isButton && {
-        textDecoration: 'none'
-      },
+
```
khmakoto: nit: It seems we were not leaving extra blank lines before in this file.

dzearing:
I think the screener issues were an outage of the test image service. Re-running.

PR #11937, opened 2020-02-12
Fix: Uncaught SyntaxError: Failed to execute 'querySelector' on 'Document'

### LINKED ISSUE #11936: Uncaught SyntaxError: Failed to execute 'querySelector' on 'Document'
### Environment Information

- **Package version(s)**: 7.74.1
- **Browser and OS versions**: It doesn't matter :)

Please checkout codepen link [here](https://codepen.io/joouis/pen/PoqPELj). And repro steps are as below:

1. Open your console of browser.
2. Click dropdown component and choose one or more options, then close it.
3. Click dropdown component again and you will find exception log in console.

#### Actual behavior:

image

As you can see there is an uncaught DOMException in the console, as I verified it's a bug.

#### Expected behavior:

No uncaught exception here.

Are you willing to submit a PR to fix? **Yes.**

Requested priority: High.

Products/sites affected: **Yes, but use a workaround to avoid it.**

Joouis:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11936

#### Description of changes

We found many error logs like “Uncaught SyntaxError: Failed to execute 'querySelector' on 'Document': '#03dcc99b-DataFormat-list1' is not a valid selector. at window.onerror” in our AppInsight tool. Of course the root cause is Dropdown component uses `querySelector` method with IDs which start with a digit which is not allowed in CSS3 selectors.

Then I find out how we trigger this exception after tracing code, the reason is dropdown component will use the property ID we passed as beginning of element ID, which will be queried by “querySelector”, and the ID we passed is exactly starting with number, so the exception occurred so many times.

We already change our ID to avoid this exception, but I still think the library can be more robustness, as you can't expect what kinds of ID user will pass. So I change `querySelector` to `getElementById` to fix this issue, and I verified it via local storybook.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11937)


Joouis:
@xugao  Hi, I merged master and pushed again but screener tests still failed.

image

xugao:
@Joouis - hmm i think this is another 1-off error unfortunately. could you try again?

Joouis:
@xugao Yes this time everything is ok :)

xugao:
i am going to merge this since @joschect is OOF

Issue #7160, opened 2018-11-20
Cleanup: OUFR packages should not depend on react-dom


kenotron:
### Environment Information 
 
- __Package version(s)__: master packages all have this
- __Browser and OS versions__: applies to all

#### Actual behavior:
 
`ReactDOM.findDOMNode` is present in DragAndDropHelper, for example. Generally it is a code smell if the component code has some dependency on ReactDOM. We should eliminate this dependency. Libraries built on top of Fabric is arbitrarily tied to react-dom package which is unnecessary.
 
#### Expected behavior:

packages/office-ui-fabric-react should not have `react-dom` in its dependencies list.

#### Priorities and help requested:
 
Are you willing to submit a PR to fix? (Yes, No) Yes
 
Requested priority: (Blocking, High, Normal, Low) Normal
 
Products/sites affected: (if applicable) All

KevinTCoughlin:
Also:

https://github.com/OfficeDev/office-ui-fabric-react/blob/fd8af84b7deea5e883773f565477b74831cd6c81/packages/office-ui-fabric-react/src/components/DetailsList/DetailsRow.base.tsx#L341

Curious, how would one avoid its usage in `Layer.base`?

https://github.com/OfficeDev/office-ui-fabric-react/blob/fd8af84b7deea5e883773f565477b74831cd6c81/packages/office-ui-fabric-react/src/components/Layer/Layer.base.tsx#L103

kenotron:
Shoot didn't realize portal hangs off of ReactDOM. Will have to figure out whether we can abstract that out.

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

kenotron:
This really is a sticking point because of the use of portals... not much going around this - OUFR must then depend on ReactDOM. Question is whether we could use a lint rule to disable to use of it?

https://palantir.github.io/tslint/rules/import-blacklist/

So we can at least reduce the use of it via a rule? That means we can at least disable this rule only for exceptional cases?

KevinTCoughlin:
@kenotron were you planning on doing this work or should I leave it unassigned / for shield?

KevinTCoughlin:
> 
> 
> This really is a sticking point because of the use of portals... not much going around this - OUFR must then depend on ReactDOM. Question is whether we could use a lint rule to disable to use of it?
> 
> https://palantir.github.io/tslint/rules/import-blacklist/
> 
[...]

PR #11945, opened 2020-02-13
Add `domRef` property to most components to allow access to underlying DOM elements without need for deprecated `React.findDOMNode`

MLoughry:
#### Pull request checklist

- [x] Addresses an existing issue: #7160

#### Description of changes

Given that `ReactDOM.findDOMNode()` is [deprecated](https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage) and [not compatible with React's Concurrent Mode](https://reactjs.org/docs/concurrent-mode-adoption.html#feature-comparison), developers need alternate ways to access the underlying DOMElements that make up Fabric controls.

This change adds a `domRef` to most components that exposes the top-level (or, in rare cases aroud popups, the primary) DOM Element from the control. The only components intentional skipped over for this change were those that provided the ability to custom render the top-level element; both because the typing would need to be for `IRefObject<any>`, and because a developer can use a custom rendering that provides its own ref.

As a necessary addition to the utilities package for this change, I added a function to merge multiple refs into one ref that can be passed to a JSX.Element.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11945)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
+++ b/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
@@ -292,7 +292,7 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
     }
 
     const Button = (keytipAttributes?: any): JSX.Element => (
-      <Tag {...buttonProps} {...keytipAttributes}>
+      <Tag {...buttonProps} {...keytipAttributes} ref={buttonProps.domRef}>
         <span className={this._classNames.flexContainer} data-automationid="splitbuttonprimary">
           {onRenderIcon(props, this._onRenderIcon)}
           {this._onRenderTextContents()}
--- a/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
+++ b/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
@@ -12,7 +12,8 @@ import {
   mergeAriaAttributeValues,
   portalContainsElement,
   memoizeFunction,
-  nullRender
+  nullRender,
+  mergeRefs
 } from '../../Utilities';
 import { Icon, FontIcon, ImageIcon } from '../../Icon';
 import { DirectionalHint } from '../../common/DirectionalHint';
@@ -273,7 +274,7 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
     this._openMenu(shouldFocusOnContainer, shouldFocusOnMount);
   }
 
-  private _onRenderContent(tag: any, buttonProps: IButtonProps): JSX.Element {
+  private _onRenderContent(tag: any, buttonProps: IButtonProps & { ref?: any }): JSX.Element {
     const props = this.props;
     const Tag = tag;
     const {
@@ -292,7 +293,7 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
     }
 
     const Button = (keytipAttributes?: any): JSX.Element => (
-      <Tag {...buttonProps} {...keytipAttributes} ref={buttonProps.domRef}>
+      <Tag {...buttonProps} {...keytipAttributes} ref={mergeRefs(buttonProps.domRef, buttonProps.ref)}>
         <span className={this._classNames.flexContainer} data-automationid="splitbuttonprimary">
           {onRenderIcon(props, this._onRenderIcon)}
           {this._onRenderTextContents()}
--- a/packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
@@ -14,7 +14,8 @@ import {
   getNativeProps,
   getWindow,
   on,
-  shallowCompare
+  shallowCompare,
+  mergeRefs
 } from '../../Utilities';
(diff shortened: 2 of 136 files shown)


ecraig12345:
Rather than introducing a new domRef property could we just use react's built-in ref?

packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
```diff
@@ -273,7 +274,7 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
     this._openMenu(shouldFocusOnContainer, shouldFocusOnMount);
   }
 
-  private _onRenderContent(tag: any, buttonProps: IButtonProps): JSX.Element {
+  private _onRenderContent(tag: any, buttonProps: IButtonProps & { ref?: any }): JSX.Element {
```
ecraig12345: Is there a built-in React interface for this?

packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -226,7 +228,7 @@ export class CalloutContentBase extends React.Component<ICalloutProps, ICalloutS
     const visibilityStyle: React.CSSProperties | undefined = this.props.hidden ? { visibility: 'hidden' } : undefined;
     // React.CSSProperties does not understand IRawStyle, so the inline animations will need to be cast as any for now.
     const content = (
-      <div ref={this._hostElement} className={this._classNames.container} style={visibilityStyle}>
+      <div ref={mergeRefs(this._hostElement, domRef)} className={this._classNames.container} style={visibilityStyle}>
```
ecraig12345: In cases like this it might be better to do this in the constructor and get rid of the merge refs call.
```
this._hostElement = props.domRef || React.createRef();
```
(and remove the old host element initialization)

MLoughry:
> Rather than introducing a new domRef property could we just use react's built-in ref?

No. That would break the existing contract; regardless, [ref forwarding](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components) only works on functional components:

>**Note**
>The second ref argument only exists when you define a component with React.forwardRef call. Regular function or class components don’t receive the ref argument, and ref is not available in props either.

packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
MLoughry: There is not

packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
MLoughry: `domRef` could be a function (`(ref: T | null) => void`), which would break existing usage of the internal `_hostElement`

MLoughry:
@ecraig12345 @dzearing Any other issues with the PR? I'll be OOF the rest of the week, but I'd like to get traction on this PR by sometime next week.

ecraig12345:
@MLoughry I think the main thing I'd like to see is input from more team members on whether they think this is the right API. It's probably fine, but adding a new API everywhere is a big commitment.

joschect:
I'm a bit late to the party on this one. It seems the guidance from React suggests a wrapper should be used when the dom needs to be accessed. Are there some explicit use cases as to when this will be used could help us get a little more context on what's driving this? And maybe help us understand why something more than a dom wrapper is needed?

MLoughry:
@joschect What do you mean by a `wrapper`? Another layer of DOM nodes? That can cause other issues with styling and such.

This is needed for cases when a developer needs access to the underlying DOM node for whatever reason without using `ReactDOM.findDOMNode()`, which is deprecated and not supported in React Concurrent mode.

joschect:
@MLoughry yes, wrapping it in a div or span or something. While it can make styling more difficult, it might enable the escape hatch of accessing the "element" without needing findDomNode. 

Did you have some examples of where a ref is needed over a div container? I guess I'm looking to see if it's inescapable that we'll need to have forwarded/passed through refs or if there might be another reasonable generic solution.

@dzearing or @levithomason , do either of you have any thoughts?

joschect:
@layershifter Hey! Do you have any insights into ways that we might want to approach ref support?

layershifter:
`ReactDOM.findDOMNode()` is a no way for Concurrent mode, https://reactjs.org/docs/concurrent-mode-adoption.html#feature-comparison. So only to options are remaining:

- `React.forwardRef()`
- `innerRef`/`domRef` whatever else

I am not enough familiar with OUFR codebase and I don't want to say this, but `domRef` will be a safer option for the current state of things. _Why?_

As I see almost all your components are classes (_not issue_), but a lot of them have imperative APIs via exposed public methods (oops 💣 ), so if we will try to do something like this:

```tsx
const PublicList = React.forwardRef(function PublicList(props, ref) {
  return <ClassList {...props} innerRef={ref} />;
});
```

You will loose these imperative APIs 😨 I tried an approach how we can keep both on CodeSandbox with `React.useImperativeHandle()` hook: https://codesandbox.io/s/awesome-sun-vtxeg. But I am not sure that I really this this solution.

PR #11949, opened 2020-02-13
FocusZone and List: use EventListener or native eventing, not EventGroup.

aneeshack4:
#### Pull request checklist

#### Description of changes

In our convergence effort, we want to use a common eventing utility across all components in Fluent. EventListener from `fluent-ui-react` (FUR) has been imported into `office-ui-fabric-react` (OUFR) code. Since it's more performant than EventGroup, I'm replacing OUFR's usage of EventGroup with FUR's EventListener - starting with List & FocusZone. FocusZone currently using native eventing (through the `on` wrapper) - got rid of these calls to `on` and replaced them with `addEventListener` and `removeEventListener`. 

Tested the UI interactions of these events on Storybook with List in this screenshot but will need to do additional testing - just sending this PR out to start getting feedback and more ideas on how to test this (perf, etc...) I also tested out these changes in other components like Slider & Panel - it broke eventing in Slider and I'm investigating why. Panel, and I suspect other OUFR components, have helper methods with types that are incompatible with the `listener` argument of FUR's EventListener so I foresee some rewriting around these methods in OUFR's components. `react-focus` doesn't have Storybook support yet so can't test FocusZone eventing in the UI at the moment but hoping to do that soon. We also need to make `yarn change` trigger for the `react-focus` package - these are both TODOs for us as we polish the move of FocusZone.
image

#### Focus areas to test
Perf, UI interactions in eventing

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11949)


packages/office-ui-fabric-react/package.json
```diff
@@ -79,6 +79,7 @@
     "@uifabric/set-version": "^7.0.5",
     "@uifabric/styling": "^7.10.8",
     "@uifabric/utilities": "^7.12.0",
+    "@fluentui/react-focus": "^0.1.0",
```
dzearing: Can we have this in a separate PR first and make sure nobody busts?

aneeshack4: done

aneeshack4:
Not sure why it's failing for a yarn change - I did run and this has a change file...

ecraig12345:
> Not sure why it's failing for a yarn change - I did run and this has a change file...

You're missing a change file for react-focus

ecraig12345:
@kenotron We need to either make the check change script ignore fluentui packages for now, or update yarn change to generate change files for them.

ecraig12345:
@aneeshack4 re: typing mismatches, we should also make sure that the types in the event-listener package are correct.

packages/react-focus/src/components/FocusZone/FocusZone.tsx
```diff
@@ -155,11 +151,13 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
         _outerZones.add(this);
 
         if (windowElement && _outerZones.size === 1) {
-          _disposeGlobalKeyDownListener = on(windowElement, 'keydown', this._onKeyDownCapture, true);
+          windowElement.addEventListener('keydown', this._onKeyDownCapture, true);
```
ecraig12345: We need a better pattern for ensuring that all manually added listeners are also removed.

layershifter:
> @aneeshack4 re: typing mismatches, we should also make sure that the types in the event-listener package are correct.

Types are correct if got the issue correctly, but be careful there as there are two types of events [React SyntheticEvent](https://reactjs.org/docs/events.html) and native DOM. `EventListener`/`useEventListener` uses native DOM events as it binds via [`addEventListener()`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) to DOM nodes. React and native events have incompatible signatures.

### React events
```tsx
const handleClick = (e: React.MouseEvent) => {}
// 
return <div onClick={handleClick} />
```

### DOM events
```tsx
const handleOutsideClick = (e: MouseEvent) => {}
// 
return <EventListener listener={handleOutsideClick} />
```

There are no issues in `List` because of typings:
```tsx
private _onFocus(ev: any): void {
// should be, don't use `any` 😭 
private _onFocus(ev: FocusEvent): void {
```

***

I also don't think that these things are equal:

```tsx
      this._events.on(this._scrollElement, 'scroll', this._onScroll);	
      this._events.on(this._scrollElement, 'scroll', this._onAsyncScroll);
//
        <EventListener capture targetRef={this._root} type="scroll" listener={this._onScroll} />
        <EventListener capture targetRef={this._root} type="scroll" listener={this._onAsyncScroll} />
```

As `target` is different element.

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -398,11 +386,17 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
...
+        <div ref={this._root} {...divProps} role={pageElements.length > 0 ? role : undefined} className={css('ms-List', className)}>
+          <div ref={this._surface} className={'ms-List-surface'} role="presentation">
+            {pageElements}
+          </div>
         </div>
-      </div>
+        <EventListener target={window} type="resize" listener={this._onAsyncResize} />
+        <EventListener capture targetRef={this._root} type="focus" listener={this._onFocus} />
+        <EventListener capture targetRef={this._root} type="scroll" listener={this._onScroll} />
+        <EventListener capture targetRef={this._root} type="scroll" listener={this._onAsyncScroll} />
```
KevinTCoughlin: `this._root` is not equivalent to `this._scrollElement` which is returned via `findScrollableParent` call. This seems like a change in behavior that may break folks cc: @ThomasMichon

packages/react-focus/src/components/FocusZone/FocusZone.tsx
```diff
@@ -155,11 +151,13 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
...
         }
       }
 
-      this._disposables.push(on(root, 'blur', this._onBlur, true));
+      if (windowElement) {
+        windowElement.addEventListener('blur', this._onBlur, true);
```
KevinTCoughlin: This event listener needs to be unbound or it will introduce a memory leak.

packages/react-focus/src/components/FocusZone/FocusZone.tsx
KevinTCoughlin: An API and/or lint rule would be grand. We've cleaned-up a number of these from analyzing snapshots downstream, so it would be unfortunate to reintroduce them.

packages/react-focus/src/components/FocusZone/FocusZone.tsx
aneeshack4: Ah yes sorry did a removeEventListener for `keydown` but not blur - just added it.

packages/react-focus/src/components/FocusZone/FocusZone.tsx
aneeshack4: Tracked in #11957

packages/office-ui-fabric-react/src/components/List/List.tsx
aneeshack4: I've changed this - had to use target instead of `targetRef` because its type isn't the same as `this._scrollElement`

packages/react-focus/src/components/FocusZone/FocusZone.tsx
ThomasMichon: Can we use the library to add the global one-time listener, too? We're specifically avoiding the native API in order to handle small variations between browsers, still, as we still support IE 11.

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -14,6 +13,7 @@ import {
   initializeComponentRef
 } from '../../Utilities';
 import { IList, IListProps, IPage, IPageProps, ScrollToMode } from './List.types';
+import { EventListener } from '@fluentui/react-component-event-listener';
```
ThomasMichon: Given that `EventGroup` is still in many other components, all this will do is increase the bundle size for all list consumers. Is there any way to somehow use a dependency injection scheme to have the host plug this in?

packages/react-focus/src/components/FocusZone/FocusZone.tsx
ecraig12345: Interesting idea! Any suggestions for what the API would look like?

packages/office-ui-fabric-react/src/components/List/List.tsx
ecraig12345: If we did that we'd be stuck with a dependency injection pattern semi-permanently for what should be a temporary problem (we're working on removing eventgroup and basecomponent in all our code).

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -400,11 +387,17 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
...
+        <div ref={this._root} {...divProps} role={pageElements.length > 0 ? role : undefined} className={css('ms-List', className)}>
+          <div ref={this._surface} className={'ms-List-surface'} role="presentation">
+            {pageElements}
+          </div>
         </div>
-      </div>
+        <EventListener target={window} type="resize" listener={this._onAsyncResize} />
+        <EventListener capture target={this._scrollElement} type="focus" listener={this._onFocus} />
+        <EventListener capture target={this._scrollElement} type="scroll" listener={this._onScroll} />
+        <EventListener capture target={this._scrollElement} type="scroll" listener={this._onAsyncScroll} />
```
layershifter: # TBD

packages/office-ui-fabric-react/src/components/List/List.tsx
KevinTCoughlin: Can we perform the `EventGroup` removal refactor in a single change-set to minimize the bundle size regression?

Once merged, we'd still have both `EventGroup` and `@fluentui/react-component-event-listener` in ODSP since many of our internal components use `EventGroup`, but the diff should be less.

aneeshack4:
Going to close this for now - opened #12231 to take care of just FocusZone - we're rethinking replacing EventGroup with EventListener.

PR #11959, opened 2020-02-14
FocusZone: Re-exporting FocusZone from react-focus package and removing duplication in OUFR package

khmakoto:
#### Pull request checklist

#### Description of changes

This PR deletes the `FocusZone` related files in the `office-ui-fabric-react` package in favor of re-exporting the implementation in the `@fluentui/react-focus` package.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11959)


khmakoto:
This PR is just to make sure that nothing breaks in CI for now. Please do not merge yet.

packages/office-ui-fabric-react/package.json
```diff
@@ -71,6 +71,7 @@
     "sinon": "^4.1.3"
   },
   "dependencies": {
+    "@fluentui/react-focus": "^0.0.2",
```
ecraig12345: Should we bump this to v7 before adding it as a dependency of fabric? Once it's re-exported from Fabric, we lose the option of doing any breaking changes outside of a major version, and therefore it's no longer truly a v0.

khmakoto: Done!

PR #11960, opened 2020-02-14
Searchbox: use EventListener, not EventGroup.

aneeshack4:
#### Pull request checklist

#### Description of changes

Similar to #11949, replacing usage of EventGroup with EventListener in SearchBox because it's more performant and will help with our convergence effort as it's from `fluentui/react-component-event-listener` package. Had to add `onBlur` to the SearchBox.types.ts file and have the type be `(ev?: any) => void;` to make it compatible with the `listener` argument of EventListener - tried to make it not `any` and instead `FocusEvent` but caused an error that said I need to change the whole interface's type (incompatible with `ISearchBoxProps extends React.InputHTMLAttributes<HTMLInputElement>`).

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11960)


aneeshack4:
Hmm SizeAuditor is throwing a strange webpack error

ecraig12345:
> Hmm SizeAuditor is throwing a strange webpack error

The error is `Error: Cannot find module '@uifabric/experiments'`

I think this is because `yarn buildto test-bundles` isn't building experiments (it's not listed in the log in the previous step). @kenotron  Any idea why this would suddenly happen when she didn't change any deps?

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
```diff
@@ -108,35 +101,38 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
...
+              <IconButton
+                styles={{ root: { height: 'auto' }, icon: { fontSize: '12px' } }}
+                iconProps={{ iconName: 'Clear' }}
+                {...clearButtonProps}
+                onClick={this._onClearClick}
+              />
+            </div>
+          )}
+        </div>
+        {hasFocus && <EventListener targetRef={this._rootElement} type="blur" capture listener={this._onBlur} />}
```
ecraig12345: Should this be inside the div instead?

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.types.ts
```diff
@@ -60,6 +60,11 @@ export interface ISearchBoxProps extends React.InputHTMLAttributes<HTMLInputElem
    */
   onEscape?: (ev?: any) => void;
 
+  /**
+   * Callback executed when the user moves focus away from the search box.
+   */
+  onBlur?: (ev?: any) => void;
```
ecraig12345: Do you know the type of the event?

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
```diff
@@ -108,35 +101,38 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
...
-        />
-        {value!.length > 0 && (
-          <div className={classNames.clearButton}>
-            <IconButton
-              styles={{ root: { height: 'auto' }, icon: { fontSize: '12px' } }}
-              iconProps={{ iconName: 'Clear' }}
-              {...clearButtonProps}
-              onClick={this._onClearClick}
-            />
+      <>
```
dzearing: remove fragment, unneeded.

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
dzearing: yup

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.types.ts
dzearing: I think you can remove this.

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
```diff
@@ -236,10 +227,7 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
...
   };
 
-  private _onBlur = (ev: React.FocusEvent<HTMLInputElement>): void => {
-    if (this._events) {
-      this._events.off();
-    }
+  private _onBlur = (ev: FocusEvent): void => {
     this.setState({
       hasFocus: false
     });
```
dzearing: to fix the red squiggle on `this.props.onBlur(ev)`, the best you can do here (I think) is to cast the `ev` to an `any` to use the `HTMLInputAttributes` onBlur definition. It's not technically perfect, but it's also a non breaking change that doesn't really matter a ton.

kenotron:
> > Hmm SizeAuditor is throwing a strange webpack error
> 
> The error is `Error: Cannot find module '@uifabric/experiments'`
> 
> I think this is because `yarn buildto test-bundles` isn't building experiments (it's not listed in the log in the previous step). @kenotron Any idea why this would suddenly happen when she didn't change any deps?

I think it isn't because of the experiments thing right? I think the build is failing because there's no mention of the new event package in the oufr package.json in this PR.

kenotron:
I'm adding a label for not merging here because the code is not ready to published used until repo merge has happened I believe.

ecraig12345:
LGTM but agreed with Ken that we need to wait on merging.

PR #11961, opened 2020-02-14
ComboBox: use EventListener, not EventGroup.

aneeshack4:
#### Pull request checklist

#### Description of changes

Just like #11949 and #11960, replace usage of EventGroup with EventListener.

#### Focus areas to test
StoryBook focus and dropdown interactions, bundle size should also decrease.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11961)


packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -216,12 +217,12 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
   public componentDidMount(): void {
     if (this._comboBoxWrapper.current) {
       // hook up resolving the options if needed on focus
-      this._events.on(this._comboBoxWrapper.current, 'focus', this._onResolveOptions, true);
+      <EventListener target={this._comboBoxWrapper.current} type="focus" listener={this._onResolveOptions} />;
```
dzearing: this doesn't belong here; in render.
also make sure not to drop the capture flag.

dzearing: In render;
```tsx
<EventListener targetRef={this._comboBoxWrapper} type="focus" listener={this._onResolveOptions} capture />
```

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -216,12 +217,12 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
...
       if ('onpointerdown' in this._comboBoxWrapper.current) {
         // For ComboBoxes, touching anywhere in the combo box should drop the dropdown, including the input element.
         // This gives more hit target space for touch environments. We're setting the onpointerdown here, because React
         // does not support Pointer events yet.
-        this._events.on(this._comboBoxWrapper.current, 'pointerdown', this._onPointerDown, true);
+        <EventListener target={this._comboBoxWrapper.current} type="pointerdown" listener={this._onPointerDown} />;
       }
     }
```
dzearing: delete this. The "onpointerdown" check seems erroneous.

In places where the condition needs to be considered, you can always conditionalize the target. For example, if there's a "dragging" state that conditionalizes whether we listen to mousemove, you could do this:

```
<EventListener targetRef={ dragging ? rootRef : undefined } ... />
```

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -465,6 +450,8 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
         >
           {errorMessage !== undefined ? errorMessage : ''}
         </div>
+        <EventListener targetRef={this._comboBoxWrapper} type="focus" listener={this._onResolveOptions} capture />
```
ecraig12345: Will this work since the ref won't be set on first render?

aneeshack4: I believe so because the hook checks for undefined: https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/fluentui/react-component-event-listener/src/useEventListener.ts#L38

ecraig12345: But if the ref is undefined, the event listener won't actually be added until the next time the component is rendered. So even though it won't throw, it won't do what is expected.

ecraig12345: Actually never mind, I was wrong: the actual adding of the event listener is done in useEffect, Which happens after rendering. So this is fine.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -307,9 +295,6 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
 
   public componentWillUnmount(): void {
```
layershifter:
```suggestion
```

It looks that this thing can be removed

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
aneeshack4: I think this is a problem actually because the test is currently failing: https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=53154&view=logs&j=92058b04-f16a-5035-546c-cae3ad5e2f5f&t=17e772b0-f0c3-53ab-7aac-ad6df80e30d7

> office-ui-fabric-react:       Error: Uncaught [Error: @fluentui/react-component-event-listener: Passed `element` is not valid or does not support `addEventListener()` method.]

So I think `this._comboBoxWrapper` is the `element` that doesn't support the `addEventListener()` method call inside useEventListener.ts @layershifter thoughts on how to work around this?

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
aneeshack4: done.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
layershifter:
```tsx
  {keytipAttributes => (
    <>
      <div ref={this._comboBoxWrapper} /> // Order there is important see
      <EventListener
        targetRef={this._comboBoxWrapper}
        type="focus"
        listener={this._onResolveOptions}
        capture
      />
      <EventListener
        targetRef={this._comboBoxWrapper}
        type="pointerdown"
        listener={this._onPointerDown}
        capture
      />
    </>
  )}
</KeytipData>
```

See https://github.com/OfficeDev/office-ui-fabric-react/pull/11949/files#r382090349 for more details.

Issue #2997, opened 2017-10-02
List: Support ghosting experience while scrolling


markuskr:
The `List` component should support ghosting while the user scrolls. 

In my usecase I want to show a list of people with profile pictures who are loaded once the list entry becomes visible. Loading the profile pictures requires a network call, therefore I only want to do the call if the user is not scrolling through the list. Instead I want to display a ghosting experience. 

`react-virtualized` solves this by adding the `flag` is scrolling to the `renderCell` method (https://github.com/bvaughn/react-virtualized/blob/master/docs/List.md)

erichdev:
@cschleiden Are you working on something like this?

betrue-final-final:
This is a valuable feature for slow loading and large lists. We should have a design for this to work with several configurations.

cschleiden:
No, not yet. But I think it could be a valuable addition.

PR #11969, opened 2020-02-17
Adds ignoreScrollingState to List to avoid extra scrolling renders

### LINKED ISSUE #11953: DetailsList renders twice after scrolling is finished
### Environment Information

- **Package version(s)**: 7.69.3
- **Browser and OS versions**: Chrome 80 / Windows 10

The issue has been described here, but hasn't been given its own issue:
https://github.com/OfficeDev/office-ui-fabric-react/issues/8826

It can be reproduced on the example list, by running Chrome's performance tools and scrolling. You will see it causes a render after onAsyncIdle, and another after onScrollingDone:
https://developer.microsoft.com/en-us/fabric#/controls/web/detailslist

#### Actual behavior:

onAsyncIdle and onScrollingDone are called in succession after scrolling is done, causing the list to be rendered twice

#### Expected behavior:

The list should only be rendered once. DetailsList doesn't actually expose the isScrolling boolean when rendering, which was the motivation for having isScrolling in the first place, see this PR:
https://github.com/OfficeDev/office-ui-fabric-react/pull/3547
There should either be an option to opt-out of this extra render(So onScrollingDone is not called if the parent component explicitly states it doesn't care about scrolling), or isScrolling should be exposed by DetailsList's onRenderRow and onRenderItemColumn so we can delay rendering of expensive columns. Having both would be optimal.

Requested priority: Normal
Products/sites affected: ?

chce:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11953

#### Description of changes

I added ignoreScrollingState to List, and wrapped all setting of scrolling state in an if-statement checking the prop. If it's not set, it will work as before, and when set, it will not cause an extra render.

#### Focus areas to test

I checked that it worked as intended on the List Ghosting example (Turning it on and off worked as expected with the ghosting), as well as on a DetailsList. I suggest checking the same. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11969)

### DIFF
--- a/change/office-ui-fabric-react-2020-02-17-11-21-10-master.json
+++ b/change/office-ui-fabric-react-2020-02-17-11-21-10-master.json
@@ -1,9 +1,9 @@
 {
-  "type": "patch",
+  "type": "minor",
   "comment": "Adds ignoreScrollingState to List, to avoid updates to scrolling state when it's not being used",
   "packageName": "office-ui-fabric-react",
   "email": "chce@netcompany.com",
   "commit": "75d13a8a34e064a42b99763bdc43f8d8ca668982",
-  "dependentChangeType": "patch",
+  "dependentChangeType": "minor",
   "date": "2020-02-17T10:21:10.193Z"
 }
--- a/change/office-ui-fabric-react-2020-02-17-11-21-10-master.json
+++ b/change/office-ui-fabric-react-2020-02-17-11-21-10-master.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Adds ignoreScrollingState to List, to avoid updates to scrolling state when it's not being used",
+  "packageName": "office-ui-fabric-react",
+  "email": "chce@netcompany.com",
+  "commit": "75d13a8a34e064a42b99763bdc43f8d8ca668982",
+  "dependentChangeType": "patch",
+  "date": "2020-02-17T10:21:10.193Z"
+}
--- a/packages/office-ui-fabric-react/src/components/List/List.tsx
+++ b/packages/office-ui-fabric-react/src/components/List/List.tsx
@@ -347,7 +347,7 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
     let shouldComponentUpdate = false;
 
     // Update if the page stops scrolling
-    if (!newState.isScrolling && this.state.isScrolling) {
+    if (!newState.isScrolling && this.state.isScrolling && !this.props.ignoreScrollingState) {
       return true;
     }
 
@@ -527,7 +527,7 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
    * we will call onAsyncIdle which will reset it back to it's correct value.
    */
   private _onScroll(): void {
-    if (!this.state.isScrolling) {
+    if (!this.state.isScrolling && !this.props.ignoreScrollingState) {
       this.setState({ isScrolling: true });
     }
     this._resetRequiredWindows();
@@ -583,7 +583,9 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
    * This function is debounced.
    */
   private _onScrollingDone(): void {
-    this.setState({ isScrolling: false });
+    if (!this.props.ignoreScrollingState) {
+      this.setState({ isScrolling: false });
+    }
   }
 
   private _onAsyncResize(): void {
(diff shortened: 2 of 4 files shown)


chce:
Can you run the CI analysis/tests again? I think it failed due to the NPM outage yesterday.

KevinTCoughlin:
@chce, can you run the `update-api` script to regenerate the API files to include your new changes. Then CI will succeed and we can test on the PR deploy site. Thanks!

change/office-ui-fabric-react-2020-02-17-11-21-10-master.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "patch",
```
KevinTCoughlin:
```suggestion
  "type": "minor",
```

Since we're adding to the public API.

change/office-ui-fabric-react-2020-02-17-11-21-10-master.json
```diff
@@ -0,0 +1,9 @@
...
+  "comment": "Adds ignoreScrollingState to List, to avoid updates to scrolling state when it's not being used",
+  "packageName": "office-ui-fabric-react",
+  "email": "chce@netcompany.com",
+  "commit": "75d13a8a34e064a42b99763bdc43f8d8ca668982",
+  "dependentChangeType": "patch",
```
KevinTCoughlin: Same here re: "minor" I assume.

KevinTCoughlin:
The change seems reasonable. I'm not sure that we use `isScrolling` parameter entirely cc: @ThomasMichon who likely knows for sure.

If not, then perhaps we can opt-in to this perf optimization once merged.

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -347,7 +347,7 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
     let shouldComponentUpdate = false;
 
     // Update if the page stops scrolling
-    if (!newState.isScrolling && this.state.isScrolling) {
+    if (!newState.isScrolling && this.state.isScrolling && !this.props.ignoreScrollingState) {
```
xugao: no need to add this condition here since `isScrolling` won't change

packages/office-ui-fabric-react/src/components/List/List.types.ts
```diff
@@ -184,6 +184,12 @@ export interface IListProps<T = any> extends React.HTMLAttributes<List<T> | HTML
    * An object which can be passed in as a fresh instance to 'force update' the list.
    */
   version?: {};
+
+  /**
+   * Boolean value to disable scroll state updates. WARNING: This will cause the isScrolling argument in onRenderCell to always be false.
```
xugao: i think we should pass `undefined` to `onRenderCell`  in case `ignoreScrollingState` is `true`

xugao:
> The change seems reasonable. I'm not sure that we use `isScrolling` parameter entirely cc: @ThomasMichon Thomas Michon FTE who likely knows for sure.
> 
> If not, then perhaps we can opt-in to this perf optimization once merged.

@KevinTCoughlin @ThomasMichon , for my own knowledge - why did we add the `isScrolling` logic in the first place?

packages/office-ui-fabric-react/src/components/List/List.tsx
chce: Fair enough.

packages/office-ui-fabric-react/src/components/List/List.types.ts
chce: Agreed. I actually thought isScrolling wasn't an optional argument. I'll update it.

chce:
> > The change seems reasonable. I'm not sure that we use `isScrolling` parameter entirely cc: @ThomasMichon Thomas Michon FTE who likely knows for sure.
> > If not, then perhaps we can opt-in to this perf optimization once merged.
> 
> @KevinTCoughlin @ThomasMichon , for my own knowledge - why did we add the `isScrolling` logic in the first place?
It came out of #2997

KevinTCoughlin:
> > The change seems reasonable. I'm not sure that we use `isScrolling` parameter entirely cc: @ThomasMichon Thomas Michon FTE who likely knows for sure.
> > If not, then perhaps we can opt-in to this perf optimization once merged.
> 
> @KevinTCoughlin @ThomasMichon , for my own knowledge - why did we add the `isScrolling` logic in the first place?

It was added 2 years ago in https://github.com/OfficeDev/office-ui-fabric-react/pull/3547 to render placeholder content on scroll.

PR #11980, opened 2020-02-18
Fluent: Add satisfied & syncpack

JasonGore:
#### Pull request checklist

#### Description of changes

Add `pretest` command that automatically runs `satisfied` and `syncpack` as part of `yarn test`, similar to fluent-ui-react repo.

This is somewhat of an experiment to make sure CI that tests pass, there are no significant changes to bundle output, and there are no protests over the changes in deps.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11980)


JasonGore:
@kenotron Didn't you recently look at updating webpack and analyze the bundle output difference? Seems to be affecting bundle output here as well.

kenotron:
> @kenotron Didn't you recently look at updating webpack and analyze the bundle output difference? Seems to be affecting bundle output here as well.

So, yes this is a bundle size change due to webpack version changes. It is a safe because they are internals to how webpack bundles things. It seems webpack is optimizing their output a bit for size here. We only need webpack bundles for two kinds of things in PRODUCTION: the UMD bundle and the Website. So, this difference is accounting for the tooling version change only. In production, the bundles are pretty opaque (as long as the runtime is okay, then things are fine).

All in all, as long as the PR deployed sites look okay then I think the bundle size change is fine.

apps/ssr-tests/package.json
```diff
@@ -26,7 +26,7 @@
     "react": "16.8.6",
     "react-app-polyfill": "~1.0.1",
     "react-dom": "16.8.6",
-    "webpack": "4.29.5"
+    "webpack": "^4.35.0"
```
ecraig12345: Can we lock this please? don't care to what as long as we test it, but floating seems REALLY dangerous

apps/todo-app/package.json
```diff
@@ -30,7 +30,7 @@
     "react": "16.8.6",
     "react-app-polyfill": "~1.0.1",
     "react-dom": "16.8.6",
-    "typescript": "3.7.5",
+    "typescript": "~3.7.2",
```
ecraig12345: Lock this too

scripts/package.json
```diff
@@ -86,7 +86,7 @@
     "express": "^4.15.4",
     "extract-comments": "^1.0.0",
     "find-free-port": "~2.0.0",
-    "fork-ts-checker-webpack-plugin": "1.3.3",
+    "fork-ts-checker-webpack-plugin": "^1.3.3",
```
ecraig12345: might be safer to lock this

scripts/package.json
```diff
@@ -105,7 +105,7 @@
     "jju": "^1.4.0",
     "json-loader": "^0.5.7",
     "json-stable-stringify-without-jsonify": "^1.0.1",
-    "just-scripts": "~0.34.0",
+    "just-scripts": "^0.35.0",
```
ecraig12345: Probably best to lock this down more too

apps/todo-app/package.json
JasonGore: I tried pinning to `3.7.5` but API extractor is using `~3.7.2`. I've pinned it to `3.7.2` for now.

PR #11982, opened 2020-02-18
Fluent: Perf CI

JasonGore:
#### Pull request checklist

#### Description of changes

This PR enables perf-testing for Fluent components and combines perf test reporting for both Fabric and Fluent components in the same PR comment. (Danger is not used for Fluent perf-test reporting as it is used in the main pipeline and Danger comments cannot easily be consolidated into one comment from multiple pipelines.)

In the future, the two perf-test packages will be merged into one.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11982)


azure-pipelines.yml
```diff
@@ -152,34 +152,38 @@ jobs:
...
-  #           echo "Skipping Danger JS because BUILD_REASON ${BUILD_REASON} != PullRequest"
-  #         else
-  #           echo "Skipping Danger JS because API token wasn't found"
-  #         fi
-
-  #     - script: |
-  #         yarn danger ci
-  #       env:
-  #         DANGER_GITHUB_API_TOKEN: $(DANGER_GITHUB_API_TOKEN)
+  - job: Danger
```
kenotron: This is the 4th parallelized job after build - wondering what the queuing times would look after this (mostly seeing if we need to up the count of vms)

JasonGore: I will be moving this to the `perf-test` pipeline as part of this PR, it won't affect gated CI time.

JasonGore: (This is a Draft PR not ready for review.)

azure-pipelines.perf-test.yml
```diff
@@ -12,18 +12,41 @@ steps:
   # No need for checkout since we're using build artifacts.
   - checkout: none
 
+  # TODO: why is this different from main pipeline? is this really getting the correct artifacts?
+  # I think this could be grabbing the wrong (most recent PR) artifacts rather than the one it's associated with
+  # This run is for PR 20200219.5 build #53232 but is grabbing artifacts from 2020219.6 build #53235
+  # https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=53238&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=559759ef-66f4-5335-a959-132505aae30a
+  # Last PR Build: https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=53232&view=logs&jobId=92058b04-f16a-5035-546c-cae3ad5e2f5f&j=ca395085-040a-526b-2ce8-bdc85f692774&t=25a453a0-e716-5e6e-0bee-a24e8e541880
```
JasonGore: @kenotron Are you sure this pipeline is always using the correct artifacts? I feel like it would when PR activity is low, but if you check my comments and links above I feel like it could be using the "most recent" pipeline artifacts rather than the artifacts associate with its PR.

JasonGore:
@miroslavstastny @layershifter I was hoping you could just review the report format and let me know what you think. I've tweaked it a bit to collapse all information by default to reduce noise on PRs. Thanks!

There are a lot of false potential regressions because the code imported in the repo is not live with fluent repo. I think I'm going to disable regressions until convergence.

apps/perf-test/tasks/fluentPerfRegressions.ts
```diff
@@ -0,0 +1,167 @@
+import * as _ from 'lodash';
+import * as path from 'path';
+
+// TODO: remove. temporary relative pathing until perf-test packages are consolidated.
+import config from '../../../scripts/config';
```
ecraig12345: You should be able to import this from `@uifabric/build` instead of a relative path I think?

PR #11984, opened 2020-02-18
Create examples package

ecraig12345:
#### Pull request checklist

#### Description of changes

Add a separate package for examples and documentation files, and scripts for running yarn start (storybook) out of that package for multiple projects. Also add a script for moving all the existing files but don't run it yet.

Right now the `react-focus` package's examples are included in the new package to demonstrate how it will work. I plan to run the script to move existing files in a separate PR. 

This PR does contain a couple of changes to existing examples to make the move process easier: 
- Standardize the locations of `date-time` example files so that the script works with them more easily
- Delete the separate PositioningContainer docs since this component is a half-baked fork of Callout and really should not be used separately.

One unfortunate thing about this approach is that because the examples package must be published, and it must depend on all the packages it's providing examples for, we either have to 1) start publishing all packages as soon as they have examples (not ideal for early-stage stuff), or 2) include unpublished packages as `devDependencies` instead of `dependencies` ("smells funny" and would be easy to forget to change the dependency type once the package is published). I went with 2 because it seemed less risky--at the moment this only affects the `@uifabric/lists` package, but the same scenario is likely to come up with other new packages that are early in development.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11984)


packages/examples/.storybook/preview.js
```diff
@@ -0,0 +1,39 @@
...
+addDecorator(withA11y());
+addParameters({
+  a11y: {
+    manual: true
+  }
+});
+
+initializeIcons();
+
+// const packageName = path.basename(process.cwd());
```
ecraig12345: doesn't work since tgis runs on client and I got stuck on what to do instead

kenotron: I think the package name could either be a part of a path of the URL or a parameter. Then, you can use that information to FILTER out stories inside the loadStories() func.

ecraig12345: But I don't want to build all packages when only starting one package

ecraig12345: So I think the filtering should be done "server"/build side somehow

packages/examples/.storybook/preview-loader.js
```diff
@@ -0,0 +1,23 @@
...
+  // invalidate cache when process.cwd() changes
+  this.addDependency(process.cwd());
+
+  const packageName = path.basename(process.cwd());
+  if (!fs.existsSync(path.resolve(__dirname, '../src', packageName))) {
+    this.callback(new Error(`Package ${packageName} does not have examples!`));
+    return;
+  }
+
+  return source.replace('PACKAGE_NAME', packageName);
```
kenotron: nice!

kenotron: https://www.npmjs.com/package/string-replace-loader might be a guide... maybe it would be more flexible to have this loader know nothing about cwd or __dirname? would just get options passed in ?

change/@fluentui-examples-2020-02-21-20-27-46-examples-pkg.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Create example package",
```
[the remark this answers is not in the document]
ecraig12345: I don't think we really have a consistent style for this.

PR #11985, opened 2020-02-18
React-focus: Enabling storybook in new package

khmakoto:
#### Pull request checklist

#### Description of changes

This PR enables storybook for the new `react-focus` package and ports examples from `office-ui-fabric-react` to `react-focus` in order for something to show up when doing `yarn start`. The examples have been ported to use native html elements instead of Fabric components so that the package didn't have to take dependency on `office-ui-fabric-react` but a similar experience was still delivered through the examples.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11985)


ecraig12345:
FYI I'm working on the full solution in #11984. But merging this for now makes sense.

PR #11990, opened 2020-02-19
styled: do not create new styles function if styles prop is not changed

### LINKED ISSUE #11979: useReducedRowRenderer doesn't work for default row rendering on DetailsList
### Environment Information

- **Package version(s)**: 7.93.0
- **Browser and OS versions**: Chrome 80 / Win 10

https://codepen.io/chce/pen/MWweJLw
Run a performance profiler that can profile React events, and notice the many [update] renders
#### Actual behavior:

It looks like the function "styles" in the props of DetailsRow is causing a comparison miss when shouldComponentUpdate calls shallowCompare. This causes every currently rendered row to re-render when f.ex. scrolling down the list.

#### Expected behavior:

shouldComponentUpdate should return false when no props have changed, especially when I'm scrolling.

A workaround, if you don't need to use styles, is to provide an onRenderRow-function, that returns the following:
<DetailsRow {...props} styles={undefined} />

Are you willing to submit a PR to fix? Yes, but I have no idea where to start, as this is caused by the styled-HOC and therefore probably not that simple to fix

Requested priority: Normal

Products/sites affected: DetailsList, ShimmeredDetailsList, DetailsRow

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11979

#### Description of changes
Currently, as long as `styles` prop is present in `styled`, we will create new `styles` function and pass as prop to base component. We should not create new `styles` prop for base component if `styles` passed to styled component is not changed on update, to avoid unnecessary re-renders.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11990)

### DIFF
--- a/change/@uifabric-utilities-2020-02-18-18-08-34-xgao-styled.json
+++ b/change/@uifabric-utilities-2020-02-18-18-08-34-xgao-styled.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "styled: do not create new styles function if styles prop is not changed",
+  "packageName": "@uifabric/utilities",
+  "email": "xgao@microsoft.com",
+  "commit": "33fd9733154fab69a3ed29f39fea34d8f07bc632",
+  "dependentChangeType": "patch",
+  "date": "2020-02-19T02:08:34.722Z"
+}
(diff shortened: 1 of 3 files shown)


dzearing:
nice catch!

PR #11999, opened 2020-02-19
Added FUI changelog checker for Fabric

kenotron:
#### Pull request checklist

#### Description of changes

The change here is that all the packages/fluentui need to be accompanied by a change in the packages/fluentui/CHANGELOG.md (since this repo now has two libraries, we cannot put this in the root).

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/11999)

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -2,6 +2,7 @@ import { danger, fail, warn, markdown, message } from 'danger';
 
 import detectNonApprovedDependencies from './scripts/dangerjs/detectNonApprovedDependencies';
 // import checkPerfRegressions from './scripts/dangerjs/checkPerfRegressions';
+import checkChangelog from './scripts/dangerjs/checkChangelog';
 
 /**
  * This trick (of explicitly passing Danger JS utils as function arg, instead of importing them at places where needed)
@@ -13,7 +14,7 @@ export default async () => {
   // TODO: Make sure scripts don't trigger warnings on Fabric-only PRs. Don't confuse users.
 
   // TODO: Consolidate approach with Fabric's change/checkchange functionality. Remove script implementation if unused.
-  // await checkChangelog(dangerJS);
+  await checkChangelog(dangerJS);
 
   await detectNonApprovedDependencies(dangerJS);
 
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -16,7 +16,7 @@ export default async () => {
   // TODO: Consolidate approach with Fabric's change/checkchange functionality. Remove script implementation if unused.
   await checkChangelog(dangerJS);
 
-  await detectNonApprovedDependencies(dangerJS);
+  //await detectNonApprovedDependencies(dangerJS);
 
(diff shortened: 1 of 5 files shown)


dangerfile.ts
```diff
@@ -13,9 +14,9 @@ export default async () => {
   // TODO: Make sure scripts don't trigger warnings on Fabric-only PRs. Don't confuse users.
 
   // TODO: Consolidate approach with Fabric's change/checkchange functionality. Remove script implementation if unused.
-  // await checkChangelog(dangerJS);
+  await checkChangelog(dangerJS);
 
-  await detectNonApprovedDependencies(dangerJS);
+  //await detectNonApprovedDependencies(dangerJS);
```
ecraig12345: Revert this? Looks like a merge conflict with Jason's PR.

scripts/dangerjs/checkChangelog.ts
```diff
@@ -26,32 +27,42 @@ const getMalformedChangelogEntries = async (danger): Promise<string[]> => {
...
+
+    return [];
   });
 };
 
 export default async ({ danger, fail, warn }: DangerJS) => {
-  // Check for a CHANGELOG entry
-  const hasChangelog = danger.git.modified_files.some(f => f === CHANGELOG_FILE);
+  // Check for a CHANGELOG entry for changes inside /packages/fluentui
+  const changes = [...danger.git.created_files, ...danger.git.deleted_files, ...danger.git.modified_files].filter(
```
ecraig12345: 1. Does danger always use UNIX style paths?
2. Deleting the changelog file should be an error. Also, checking the created files only seems like it would be useful for this specific PR, since after that the file will already exist.

dangerfile.ts
kenotron: Ooops yes

scripts/dangerjs/checkChangelog.ts
kenotron: 1. danger uses git directly which normalizes to UNIX style paths

kenotron: 2a. deleting changelog SHOULD be error - but i'm not really doing much to the logic here
2b. creating files can occur post merge of repo - we'll want to check for these changes

ecraig12345: For 2 I misunderstood what it was doing, sorry

dangerfile.ts
```diff
@@ -13,7 +14,7 @@ export default async () => {
   // TODO: Make sure scripts don't trigger warnings on Fabric-only PRs. Don't confuse users.
 
   // TODO: Consolidate approach with Fabric's change/checkchange functionality. Remove script implementation if unused.
```
JasonGore: This TODO is resolved now and can be removed.

PR #12002, opened 2020-02-19
Prettier script updates

ecraig12345:
#### Pull request checklist

#### Description of changes

Prettier script updates. Mainly renaming a function and adding quotes around paths so the shell doesn't expand `**/*`. Also deleting the .prettierrc.json from fluent so that it doesn't cause incorrect formatting for files in the same folder.

~~Add `trailingComma: all` rule.~~ (Postponing this until another PR that updates all files, to avoid noise in the meantime.)

LATER: actually run prettier for all the files. (putting scripts up for review first)

#### Focus areas to test

Ensure index.html files don't have trailing commas which will break ES5.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12002)


ecraig12345:
I decided to get rid of the trailing comma rule for now to avoid creating noise before the PR that updates all the files is merged.

PR #12003, opened 2020-02-19
allowScrollOnElement: fix memory leak caused by closure

xugao:
#### Pull request checklist

#### Description of changes
Repro of the issue (Panel): https://codesandbox.io/s/inspiring-wu-dwrl9?fontsize=14&hidenavigation=1&theme=dark

Removed the outer scope in `allowScrollOnElement` so that `_element` does not hold on to the DOM element forever

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12003)

### DIFF
--- a/packages/utilities/src/scroll.ts
+++ b/packages/utilities/src/scroll.ts
@@ -19,7 +19,15 @@ const DisabledScrollClassName = mergeStyles({
  */
 export const DATA_IS_SCROLLABLE_ATTRIBUTE = 'data-is-scrollable';
 
-const _makeElementScrollAllower = () => {
+/**
+ * Allows the user to scroll within a element,
+ * while preventing the user from scrolling the body
+ */
+export const allowScrollOnElement = (element: HTMLElement | null, events: EventGroup): void => {
+  if (!element) {
+    return;
+  }
+
   let _previousClientY = 0;
   let _element: Element | null = null;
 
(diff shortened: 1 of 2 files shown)


packages/utilities/src/scroll.ts
```diff
@@ -66,24 +74,12 @@ const _makeElementScrollAllower = () => {
...
   };
 
-  return (element: HTMLElement | null, events: EventGroup): void => {
-    if (!element) {
-      return;
-    }
-
-    events.on(element, 'touchstart', _saveClientY, { passive: false });
-    events.on(element, 'touchmove', _preventOverscrolling, { passive: false });
+  events.on(element, 'touchstart', _saveClientY, { passive: false });
```
aneeshack4: Ooh prime spot to use EventListener haha (we can do that later ofc)

packages/utilities/src/scroll.ts
```diff
@@ -66,24 +74,12 @@ const _makeElementScrollAllower = () => {
...
+  events.on(element, 'touchmove', _preventOverscrolling, { passive: false });
```
KevinTCoughlin: I noticed the callers do not invoke `EventGroup.dispose` on `componentWillUnmount`, should they to prevent additional leaks? cc: @dzearing 

https://github.com/OfficeDev/office-ui-fabric-react/blob/8111845351329a689c13050dd7c881089bbea140/packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx

https://github.com/OfficeDev/office-ui-fabric-react/blob/b37e5730260552b5f4ef6e1a6d6d6e289792830f/packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx

xugao: `BaseComponent` calls dispose on willunmount.

packages/utilities/src/scroll.ts
xugao: actually David just helped point out, here is not a react componnet/place we can use hook. this will require some deeper thoughts

packages/utilities/src/scroll.ts
KevinTCoughlin: You're right. Let's keep in mind this area when `BaseComponent` is removed which my understanding is planned. During review of that refactor we must make sure event handlers are properly disposed. Thanks for the fix!

PR #12008, opened 2020-02-20
add scheduled signal

amshan15:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12008)

### DIFF
--- a/packages/experiments/package.json
+++ b/packages/experiments/package.json
@@ -68,6 +68,7 @@
     "@uifabric/theme-samples": "^7.0.22",
     "@uifabric/utilities": "^7.12.1",
     "@uifabric/variants": "^7.1.12",
+    "core-js": "3",
     "deep-assign": "^2.0.0",
     "office-ui-fabric-react": "^7.93.1",
     "prop-types": "^15.5.10",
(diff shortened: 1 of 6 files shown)


packages/experiments/package.json
```diff
@@ -68,6 +68,7 @@
     "@uifabric/theme-samples": "^7.0.22",
     "@uifabric/utilities": "^7.12.1",
     "@uifabric/variants": "^7.1.12",
+    "core-js": "3",
```
ecraig12345: We shouldn't add a dependency on core-js in this package. If you explain why you needed it, I can probably help find a workable alternative.

amshan15: whoops that should not be there. I'll remove it and update the PR

ecraig12345: Thanks! You'll need to update yarn.lock too.

amshan15: Just updated. Thanks!

PR #12009, opened 2020-02-20
Adding support for running FUI screener tests

kenotron:
#### Pull request checklist

#### Description of changes

This adds screener support for FUI packages. There are some caveats: Screener doesn't have a notion of multiple projects inside a repo. So we have to FAKE it with fake baseBaseline's.

1. preserve running screener for FUI from `gulp screener` (for now)
2. update screener gulp tasks to match the new locations of examples
3. added branch detection logic from vrtest (will be refactored when this is verified to be working in the PR)
4. added yet another job to be run in parallel after build to run the screener tests for FUI

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12009)


scripts/screener/screener.config.js
```diff
@@ -11,9 +35,14 @@ require('tsconfig-paths').register({
   paths: compilerOptions.paths
 });
 
+// this is a fake branch to pretend we have a feature branch as a separate project
+const baseBranch = 'master-fluentui';
```
ecraig12345: Are you actually adding this branch? If so, how can we ensure that it stays up-to-date?

kenotron: This is a fake branch, doesn't exist in life at all. I am hoping that screener merely uses it as a key to store their baselines. They support feature branches now through this.

azure-pipelines.yml
```diff
@@ -188,7 +191,6 @@ jobs:
   #         else
   #           echo "Skipping Danger JS because API token wasn't found"
   #         fi
-
```
JasonGore: I still don't understand how this is running formatting for you when it never does for me. My changes to these files never get formatted on commit. Is it an IDE plugin for you? (This will cause a bunch of unnecessary conflicts too, I'm removing this section of code entirely in another PR.)

kenotron: I don't know why prettier ignore file doesn't have azure pipeline yml files. But that's why - the IDE plugin will format everything not listed in prettier ignore. It should have done so for you as well... if you use that plugin :P

ecraig12345: In my change yesterday, I actually just added yml to the list of file types that will be formatted. I can revert that if it's not desirable, but I don't really see any reason why we shouldn't be formatting them.

scripts/screener/screener.config.js
```diff
@@ -2,6 +2,30 @@
 
 require('@uifabric/build/babel/register');
 
+const cp = require('child_process');
```
ecraig12345: This comment applies to more than just this one file, but I think it would be helpful to have a better way to denote which files only apply for the fluent build. Especially if we end up in this state for a while, it will confuse anyone coming in and trying to understand the build tools if there are two completely separate build systems combined into the same folder. (This is kind of a similar problem to how the scripts specific to docs make the overall fluent build look a lot more complicated than it is.)

kenotron: I'm taking a simple approach to not block us from repo merging. We need to quickly move files around once everyone's on this one repo.

scripts/screener/screener.config.js
kenotron: This hack was an experiment, I think it isn't needed anymore.

azure-pipelines.yml
kenotron: I think we should format them. I think that will promote consistency.

PR #12025, opened 2020-02-21
Panel: use align-items for cross axis alignment.

### LINKED ISSUE #12020: Panel Body CSS Shift IE11
### Environment Information

- **Package version(s)**: at least 7.86.0 and forward
- **Browser and OS versions**: IE11

https://developer.microsoft.com/en-us/fabric#/controls/web/panel

Open any of the first two panels in IE11, you can see they are rendered completely unusable.
image

image

#### Actual behavior:

Inner content shifted outside of the panel
Unclear if this is the root cause, but by switching the css property in the class ms-Panel-header, margin-right: auto to margin-right:0px seems to at least display the content

#### Expected behavior:

All content displayed appropriately

Requested priority: Blocking / High

Products/sites affected: All Panel using sites, including SharePoint/MSIT

aneeshack4:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12020 

#### Description of changes

Panel's header styles are broken in IE11 because IE11 isn't compliant with the W3 spec on margins - instead of changing the marginRight auto setting (which would break the other browsers), I'm fixing the other problem which is that with flexbox, vertical alignment (cross axis) should be set with alignItems, not justifyContent (that's for the main axis). 

#### Focus areas to test
IE 11:
image
Edge:
image
Chrome:
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12025)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
@@ -273,15 +273,7 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
       !isOpen && isAnimating && isOnRightSide && AnimationClassNames.slideRightOut40,
       focusTrapZoneClassName
     ],
-    commands: [
-      classNames.commands,
-      {
-        marginTop: 18
-      },
-      hasCustomNavigation && {
-        marginTop: 'inherit'
-      }
-    ],
+    commands: [classNames.commands],
     navigation: [
       classNames.navigation,
       {
@@ -292,15 +284,7 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
         height: commandBarHeight
       }
     ],
-    closeButton: [
-      classNames.closeButton,
-      {
-        marginRight: 14
-      },
-      hasCustomNavigation && {
-        marginRight: 0
-      }
-    ],
+    closeButton: [classNames.closeButton],
     contentInner: [
       classNames.contentInner,
       {
@@ -314,13 +298,11 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
       classNames.header,
       sharedPaddingStyles,
       {
-        flexShrink: 1,
-        marginRight: 'auto'
+        flexShrink: 1
       },
       hasCustomNavigation && {
         // Ensure that title doesn't shrink if screen is too small
-        flexShrink: 0,
-        marginRight: 0
+        flexShrink: 0
       }
     ],
     headerText: [
@@ -329,7 +311,6 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
       {
         color: semanticColors.bodyText,
         lineHeight: '27px',
-        margin: 0,
         overflowWrap: 'break-word',
         wordWrap: 'break-word',
         wordBreak: 'break-word',
@@ -350,7 +331,6 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
       classNames.content,
       sharedPaddingStyles,
       {
-        marginBottom: 0,
         paddingBottom: 20
       }
     ],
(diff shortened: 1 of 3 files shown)


aneeshack4:
@betrue-final-final could you look over the screener changes? https://screener.io/v2/states/ryQqxfIYZ.OfficeDev-office-ui-fabric-react/panel_header

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
```diff
@@ -286,7 +286,7 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
       classNames.navigation,
       {
         display: 'flex',
-        justifyContent: 'flex-end'
+        alignItems: 'flex-end'
```
phkuo: wait, what's the intention here? why is this needed?
isn't there a margin property that needs to get removed?

phkuo: we likely want justify-content:space-between to force the title and X button to be on the edges

maxwellred: I just tried out space-between successfully

aneeshack4: same tested it on IE 11, chrome, and edge again all work

phkuo:
@aneeshack4 please also clean up any extraneous margins. Flexed children should avoid using margins altogether, as a general guideline.
https://github.com/OfficeDev/office-ui-fabric-react/blob/028ba917c3633ac3ed312836f9d4f8aeb5356c76/packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts#L318

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
```diff
@@ -273,15 +273,7 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
       !isOpen && isAnimating && isOnRightSide && AnimationClassNames.slideRightOut40,
       focusTrapZoneClassName
     ],
-    commands: [
-      classNames.commands,
-      {
-        marginTop: 18
```
nkanaki: I think the marginTop is still needed. That shouldn't be an issue, right @phkuo?

nkanaki: perhaps same with marginLeft as well

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
```diff
@@ -273,34 +273,18 @@ export const getStyles = (props: IPanelStyleProps): IPanelStyles => {
...
+    commands: [classNames.commands],
     navigation: [
       classNames.navigation,
       {
         display: 'flex',
-        justifyContent: 'flex-end'
+        justifyContent: 'space-between'
       },
       hasCustomNavigation && {
         height: commandBarHeight
```
nkanaki: I would double check this, but changing justifyContent: 'space-between' might be breaking customNavigation which actually needs flex-end

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
phkuo: If it's on a flex parent and not a flex child, then yes, margins are fine.

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
phkuo: What does custom nav look like? flex-end is going to cause the title to right-align. You can try putting flex-grow:1 on the title to make sure it stays left aligned, since then it'll fill up all that left-hand side space and the text inside will left-align to that. Just have to make sure the flex parent there is always 100% width, no more and no less.

aneeshack4: Why does customNavigation need flex-end?

nkanaki: @aneeshack4 @phkuo it doesn't! just checked

packages/office-ui-fabric-react/src/components/Panel/Panel.styles.ts
aneeshack4: Cool that should fix the screener regression

phkuo:
https://css-tricks.com/the-peculiar-magic-of-flexbox-and-auto-margins/ interesting read

PR #12027, opened 2020-02-22
Update ComboBox prop documention

EvilKanoa:
Include documentation for the `openOnKeyboardFocus` prop of the ComboBox component.

#### Pull request checklist

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12027)


EvilKanoa:
Oh, my mistake in that case. I spent quite awhile trying to figure out the correct prop and assumed the documentation was simply missing. Definitely would be a good improvement to be able to see all own and inherited props within the documentation easily.

ecraig12345:
Opened #12043 about this.

PR #12028, opened 2020-02-22
DetailsList: fix memory leak caused by incorrect component memoization

### LINKED ISSUE #11955: Memory Leak caused by __stylesheet__  in Fabric
### Environment Information

- **Package version(s)**: (6.211.2)

 https://codesandbox.io/s/sparsha-sahadetailslistleak-shl1n?fontsize=14&hidenavigation=1&theme=dark

#### Actual behavior:

A memory leak caused by __stylesheet__ is noticeable when we render <DetailsList />. 

#### Expected behavior:

There should not be any memory leaks and once we unmount the list it should release the detached window object

No

Blocking

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11955

#### Description of changes

`DetailsRowCheck` incorrect `getCheck`, which mis-used `memoizeFunction` which is causing react component being kept/saved in memory regardless of its life cycle.

Note: I decided removed `getCheck`  since it doesn't fit to as a public interface/component. I don't think we need to deprecate it as I highly doubt anyone outside is using it.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12028)

### DIFF
--- a/change/office-ui-fabric-react-2020-02-21-18-26-16-xgao-dl-mem-leak.json
+++ b/change/office-ui-fabric-react-2020-02-21-18-26-16-xgao-dl-mem-leak.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "DetailsList: fix memory leak caused by incorrect component memoization",
+  "packageName": "office-ui-fabric-react",
+  "email": "xgao@microsoft.com",
+  "commit": "b292e82b235d6e6ee76f60d6b52aa0c8f7d2f418",
+  "dependentChangeType": "patch",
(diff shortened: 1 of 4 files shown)


ecraig12345:
Can you explain more about what the problem was here? Also I think we use a similar approach for Icon.

ecraig12345:
@xugao  Would still appreciate more explanation of what was wrong here.

PR #12044, opened 2020-02-25
Tile: Use more Tile props to inform link behavior

glangstonb:
#### Pull request checklist

#### Description of changes

Updated Tile (in experiments) to pass on some link-specific props ('target', 'download', 'hrefLang', 'media' and 'rel') to its internal link so that users of Tile can customize the link behavior. The Tile link now also defaults to &lt;span&gt; (instead of &lt;button&gt;) when no href or onClick is passed. Updated Tile.Media.Example to include a linking demo showing the new functionality.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12044)

### DIFF
--- a/packages/experiments/src/components/Tile/examples/Tile.Media.Example.tsx
+++ b/packages/experiments/src/components/Tile/examples/Tile.Media.Example.tsx
@@ -96,6 +96,8 @@ interface IImageTileProps {
   originalImageSize: ISize;
   showBackground: boolean;
   nameplateOnlyOnHover: boolean;
+  linkHref?: string;
+  linkTarget?: string;
   item: typeof ITEMS[0];
 }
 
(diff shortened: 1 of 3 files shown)


packages/experiments/src/components/Tile/examples/Tile.Media.Example.tsx
```diff
@@ -146,6 +150,8 @@ const ImageTile: React.FunctionComponent<IImageTileProps> = (props: IImageTilePr
 export interface ITileMediaExampleState {
   imagesLoaded: boolean;
   nameplateOnlyOnHover: boolean;
+  generateLinks: boolean;
+  linkInNewTab: boolean;
```
aneeshack4: might be missing something here but there should be logic to actually open the link in a new tab somewhere right?

glangstonb: Yes, it's on line 173, inside the render() method. Certainly we could move that part into the state if you think that's more elegant, but the way I did it seems consistent with the existing design (the other state parameters were both booleans). The demo works as intended on my machine in both Edge and Chrome.

aneeshack4: This is fine, thanks

PR #12045, opened 2020-02-25
Turning on 'deprecation' tslint rule in office-ui-fabric-react package

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Contributes to fixing #6310

#### Description of changes

This PR turns on the `deprecation` tslint rule in the `office-ui-fabric-react` package and makes the use of every deprecated prop or interface be opt-in by disabling the tslint rule in file. This allows us to better enforce not using deprecated props and interfaces in the future.

As part of this PR I'm fixing the use of various deprecated props that were being used without first defaulting to the non-deprecated ones:

```
// From
prop={deprecated}

// To
prop={new || deprecated}     or     prop={new !== undefined ? new : deprecated}
```

I am also fixing a bunch of scenarios where the deprecated prop had priority over the non-deprecated one:

```
// From
prop={deprecated || new}     or     prop={deprecated !== undefined ? deprecated : new}

// To
prop={new || deprecated}     or     prop={new !== undefined ? new : deprecated}
```

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12045)


ecraig12345:
It's possible that in some cases, the deprecated prop had priority because the new prop provides a default in defaultProps (and therefore would be used unconditionally). You should check carefully for those cases before merging.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
```diff
@@ -87,11 +87,13 @@ export class ColorPickerBase extends React.Component<IColorPickerProps, IColorPi
     const defaultStrings = ColorPickerBase.defaultProps.strings as Required<IColorPickerStrings>;
 
     this._textLabels = {
-      r: props.redLabel || strings.red || defaultStrings.red,
-      g: props.greenLabel || strings.green || defaultStrings.green,
-      b: props.blueLabel || strings.blue || defaultStrings.blue,
-      a: props.alphaLabel || strings.alpha || defaultStrings.alpha,
-      hex: props.hexLabel || strings.hex || defaultStrings.hex
+      // tslint:disable:deprecation
+      r: strings.red || props.redLabel || defaultStrings.red,
```
ecraig12345: This will cause a problem because of the default props thing I mentioned in another comment. Same for the other color stuff.

packages/office-ui-fabric-react/src/components/DetailsList/DetailsList.base.tsx
```diff
@@ -854,7 +854,8 @@ export class DetailsListBase extends React.Component<IDetailsListProps, IDetails
       const minWidth = column.minWidth || MIN_COLUMN_WIDTH;
       const overflowWidth = totalWidth - availableWidth;
 
-      if (column.calculatedWidth! - minWidth >= overflowWidth || !(column.isCollapsable || column.isCollapsible)) {
+      // tslint:disable-next-line:deprecation
+      if (column.calculatedWidth! - minWidth >= overflowWidth || !(column.isCollapsible || column.isCollapsable)) {
```
ecraig12345: I don't think the order will make a difference here since the condition will succeed if any operand is true

packages/office-ui-fabric-react/src/components/Dialog/Dialog.base.tsx
```diff
@@ -123,33 +125,36 @@ export class DialogBase extends React.Component<IDialogProps, {}> {
...
+
     return (
       <Modal
         elementToFocusOnDismiss={elementToFocusOnDismiss}
         firstFocusableSelector={firstFocusableSelector}
         forceFocusInsideTrap={forceFocusInsideTrap}
         ignoreExternalFocusing={ignoreExternalFocusing}
         isClickableOutsideFocusTrap={isClickableOutsideFocusTrap}
-        onDismissed={onDismissed}
+        onDismissed={mergedModalProps.onDismissed !== undefined ? mergedModalProps.onDismissed : onDismissed}
```
ecraig12345: Can you not just include these defaults with the mergedModalProps?

packages/office-ui-fabric-react/tslint.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": ["@uifabric/tslint-rules"],
   "rules": {
-    "deprecation": false,
+    "deprecation": true,
```
ecraig12345: You can just remove this line since the default is true

packages/office-ui-fabric-react/src/components/DetailsList/DetailsList.base.tsx
khmakoto: Yeah, it doesn't really matter, I moved them just to have a consistent style where the non-deprecated prop appears first. Just a stylistic choice since, like you said, the order doesn't make any difference in execution.

packages/office-ui-fabric-react/src/components/ColorPicker/ColorPicker.base.tsx
khmakoto: I just checked and it seems that this is the only file where that was happening. I've changed it so that it reverts to its previous implementation 👍

packages/office-ui-fabric-react/src/components/Dialog/Dialog.base.tsx
khmakoto: Good catch! Done!

packages/office-ui-fabric-react/tslint.json
khmakoto: Done!

khmakoto:
@ecraig12345 I agree that we should do this for other packages but I think that it should be done in separate PRs.

PR #12046, opened 2020-02-25
Change keytips to pull for updates instead of push

chang47:
#### Pull request checklist

#### Description of changes

In the current world of Keytips, we live in a push model where if our component (buttons, comboBox, spinButton, etc) gets updated, we would update KeyTipData which would then update the KeyTipManager, which would then cause the KeytipLayer component to be rendered.

This is bad, because controls that support keytips are paying a performance cost every single time the props are being updated whether or not if keytips are ever used.

To solve this situation, we are going to move to a poll model where the keytip manager will query the new state of the control whenever we have entered Keytip Mode.

In this PR, there are 2 changes:
1) we change from a push to a poll model
2) optimization was done to the datastrucutre holding the keytips from an array to a dictionary allowing us to get results in constant time instead of doing a linear scan.

Test results:
**Tab Switching**
In Excel online, we did a tab switch from the Formula Tab to the Home Tab. 
We are seeing a reduction in calls that the keytipManger (from 40~ to 9)
In the timings tab, the calls to KeytipLayer is gone:
image
Note this is just for what I can measure, but all instances of keytipLayer are now gone from the profiler, unless explicitly called by entering keytip mode.

We're seeing on average a 10% reduction of code execution from 255ms to 225.5ms due to not needing to render keytipLayer again.

**Disabling controls**
Another example is when we type in Excel Online, which would disable all controls in the Excel Ribbon. This would cause keytips to be rendered too.

We see similar changes where the keytipLayer are no longer being called:

image

Along with we see the committing change time being cut in half from an average of 13ms to 7ms

along with the elimination of the keytipLayer code which is another 4.3 ms

However it appears that the code execution saving is negligable changing from 139ms to 130ms a 7%~ reduction

**Entering keytip mode**
Another test that was done was when we enter keytip mode, in the push model, we have to push of the controls to get the value of the control, initial thoughts were that we would impact the time it takes for the keytips to show. To our surprise, it wasn't impacted too badly.

On average it took about 2-3ms to rebuild the tree every single time, which is negliable to the cost of everything else:

From the profiler tool, we calculate a average change from 123ms to 127ms, which is about a 3% increase, which about correlates to the extra code that we're calling.

#### Focus areas to test

Outside of the statistics, I've tested:
1) Basic usage of keytips in Excel/Word/Keytip test page 
2) Using special keyboard behavior where we would execute options in the overflow

Everything works as expected.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12046)


chang47:
@kelseyyoung 
@whykushal93 (lol)

packages/office-ui-fabric-react/src/utilities/keytips/KeytipManager.test.tsx
```diff
@@ -1,132 +1,192 @@
-import { KeytipManager } from './KeytipManager';
```
chang47: Well this is going to be painful to undo...

kelseyyoung:
@aneeshack4 I know we had (very briefly) talked about keytip changes so want you to also be aware of this

packages/office-ui-fabric-react/src/components/KeytipLayer/KeytipLayer.base.tsx
```diff
@@ -297,6 +294,24 @@ export class KeytipLayerBase extends BaseComponent<IKeytipLayerProps, IKeytipLay
     }
   }
 
+  private _buildTree(): void {
+    this._keytipTree = new KeytipTree();
+    // Add regular and persisted keytips to the tree
+    for (const id in this._keytipManager.keytips) {
+      if (id) {
```
kelseyyoung: > if (id) { [](start = 5, length = 10)

it's possible that id is undefined?

packages/office-ui-fabric-react/src/utilities/keytips/KeytipManager.ts
```diff
@@ -40,22 +42,26 @@ export class KeytipManager {
    * @param persisted - T/F if this keytip should be persisted, default is false
    * @returns {string} Unique ID for this keytip
    */
-  public register(keytipProps: IKeytipProps, persisted: boolean = false): string {
+  public register(keytipProps: IKeytipProps, persisted: boolean = false, pollCallback?: () => IKeytipProps): string {
```
kelseyyoung: >pollCallback?: () => IKeytipProps) [](start = 73, length = 34)

can you explain why the pollCallback is required, when it's returning (seemingly) the same this as what "keytipProps" is?

packages/office-ui-fabric-react/src/utilities/keytips/KeytipManager.ts
```diff
@@ -13,8 +14,9 @@ export interface IUniqueKeytip {
 export class KeytipManager {
   private static _instance: KeytipManager = new KeytipManager();
 
-  public keytips: IUniqueKeytip[] = [];
-  public persistedKeytips: IUniqueKeytip[] = [];
+  public keytips: { [Key: string]: IUniqueKeytip } = {};
+  public persistedKeytips: { [Key: string]: IUniqueKeytip } = {};
+  public sequenceMapping: { [Key: string]: IKeytipProps } = {};
```
kelseyyoung: >sequenceMapping [](start = 9, length = 15)

what benefits does sequenceMapping give you?

packages/office-ui-fabric-react/src/utilities/keytips/KeytipManager.ts
```diff
@@ -13,8 +14,9 @@ export interface IUniqueKeytip {
 export class KeytipManager {
   private static _instance: KeytipManager = new KeytipManager();
 
-  public keytips: IUniqueKeytip[] = [];
-  public persistedKeytips: IUniqueKeytip[] = [];
+  public keytips: { [Key: string]: IUniqueKeytip } = {};
+  public persistedKeytips: { [Key: string]: IUniqueKeytip } = {};
```
kelseyyoung: >{ [Key: string]: IUniqueKeytip }  [](start = 27, length = 33)

did you change them to maps to find them easier?

chang47: Correct before whenever we wanted to find a specific keytips, we had to scan through all elements. By using a map, we get instant look-up.

packages/office-ui-fabric-react/src/utilities/keytips/KeytipManager.ts
chang47: Similar reasoning for the refactoring of keytips. This is primarily used when we are looking for parent sequences of a keytips. In the past we would can through all keytips and then do an array comparison to see if we found the correct keytip.

With this change, we can do a instant lookup.

packages/office-ui-fabric-react/src/utilities/keytips/KeytipManager.ts
chang47: Good point, we don't even need this as we already have a reference to the keytipProps

packages/office-ui-fabric-react/src/components/KeytipLayer/KeytipLayer.base.tsx
chang47: It should not be, but the compiler enforced this =/

kelseyyoung: would Object.keys prevent this? or what happens when you use "of"?

---
In reply to: [385317334](https://github.com/OfficeDev/office-ui-fabric-react/pull/12046#discussion_r385317334) [](ancestors = 385317334)

chang47: of is for arrays, but I did change it to Object.keys and solve the problem :)

packages/office-ui-fabric-react/src/components/KeytipData/KeytipData.tsx
```diff
@@ -41,12 +41,12 @@ export class KeytipData extends React.Component<IKeytipDataProps & IRenderCompon
     return children(nativeKeytipProps);
   }
 
-  private _getKtpProps(): IKeytipProps {
+  private _getKtpProps = (): IKeytipProps => {
```
kelseyyoung: > =  [](start = 22, length = 3)

can revert

packages/office-ui-fabric-react/src/components/KeytipData/KeytipData.tsx
```diff
@@ -16,7 +16,7 @@ export class KeytipData extends React.Component<IKeytipDataProps & IRenderCompon
   public componentDidMount() {
     // Register Keytip in KeytipManager
     if (this.props.keytipProps) {
-      this._uniqueId = this._keytipManager.register(this._getKtpProps());
+      this._uniqueId = this._keytipManager.register(this._getKtpProps(), undefined);
```
kelseyyoung: >undefined [](start = 73, length = 9)

can revert

packages/office-ui-fabric-react/src/utilities/keytips/KeytipManager.ts
```diff
@@ -45,17 +46,21 @@ export class KeytipManager {
...
-    const uniqueKeytip: IUniqueKeytip = this._getUniqueKtp(props);
-    // Add to array
-    persisted ? this.persistedKeytips.push(uniqueKeytip) : this.keytips.push(uniqueKeytip);
-
-    const event = persisted ? KeytipEvents.PERSISTED_KEYTIP_ADDED : KeytipEvents.KEYTIP_ADDED;
-    EventGroup.raise(this, event, {
-      keytip: props,
-      uniqueID: uniqueKeytip.uniqueID
-    });
+    const uniqueKeytip: IUniqueKeytip = this._getUniqueKtp(props, undefined);
```
kelseyyoung: >undefined [](start = 66, length = 9)

needed?

chang47:
@micahgodbolt @aneeshack4 We are happy with the change now. Can you look at the PR when you have the chance?

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.test.tsx
```diff
@@ -174,8 +180,9 @@ describe('OverflowSet', () => {
       jest.useRealTimers();
 
       // Clean up the keytip items
-      keytipManager.keytips = [];
-      keytipManager.persistedKeytips = [];
+      keytipManager.keytips = {};
+      keytipManager.persistedKeytips = {};
+      keytipManager.inKeytipMode = false;
```
aneeshack4: I haven't looked at keytip code before this & don't have access to the app test pages so might be missing some context here - but I assume pressing ALT enters into keytip mode and pressing again will exit it is this correct?

chang47: Correct, but the specific reason why this was set to be false was because in certain parts of the unit test I set the value to be true.

chang47:
@aneeshack4 @micahgodbolt Can someone review the code again? Thanks!

dzearing:
@chang47 Correct me if I'm wrong... this is a "pull" model, not a "poll" model. Polling means you're querying for something on a regular interval. Pulling means we ask for the information when needed.

packages/office-ui-fabric-react/src/utilities/keytips/KeytipManager.ts
```diff
@@ -13,8 +13,9 @@ export interface IUniqueKeytip {
 export class KeytipManager {
   private static _instance: KeytipManager = new KeytipManager();
 
-  public keytips: IUniqueKeytip[] = [];
-  public persistedKeytips: IUniqueKeytip[] = [];
+  public keytips: { [Key: string]: IUniqueKeytip } = {};
+  public persistedKeytips: { [Key: string]: IUniqueKeytip } = {};
```
dzearing: nit: `Key` should be lowercased

change/office-ui-fabric-react-2020-02-24-17-30-43-master.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
```
dzearing: Is this a minor? I can't see any API changes or behavior changes.

change/office-ui-fabric-react-2020-02-24-17-30-43-master.json
```diff
@@ -0,0 +1,8 @@
...
+  "comment": "Add polling option for the Keytip Manager",
```
dzearing: The comment here shows up in our release notes. Can you clean this up to be readable to a consumer?

KeytipManager: Adjusting how keytip data is sent to the manager to optimize rendering times for components which use keytips.

dzearing:
One more comment; this is a larger discussion, but the fact that there is ANY keytip code inside of any component is a design problem. I totally appreciate the changes here, and we should get these in, but we need to refactor keytips at some point to do something outside of components. Would improve perf and bundle size of a number of components, especially for partners who do not use keytips.

This would work.

```tsx
<Keytip shortcuts={...} target="#foo" />
<Button id="foo" />
```

change/office-ui-fabric-react-2020-02-24-17-30-43-master.json
chang47: Good point, there were some, but I ended up not needing them anymore. I'll change this to be a minor.

change/office-ui-fabric-react-2020-02-24-17-30-43-master.json
chang47: Will do!

chang47:
> @chang47 Correct me if I'm wrong... this is a "pull" model, not a "poll" model. When I hear "poll", to me it means you're querying for something on a regular interval. (I literally looked for a `setInterval` call. Pulling means we ask for the information when needed.

Awkward, you're correct. Polling would be the worst thing we can do in the 3 options that we have right now.

chang47:
@dzearing @aneeshack4  I've addressed the comment and had to close the PR multiple times, but I finally got it working! Can you look at it again? Thanks!

chang47:
Thanks!

@micahgodbolt Do you think you can approve and merge the PR now? Thank you

micahgodbolt:
@msft-github-bot wait for @dzearing to ok this before merging

chang47:
Wow I'm impressed with the functionality of msft-github-bot!

chang47:
@dzearing Actually please don't merge this change in, yet. After talking to another co-worker we decided that we want to make this behavior optional so we can measure the impact this would bring. I'm currently trying to see what's the best way to accomplish this.

chang47:
@dzearing I've updated the keytipmanager to support a parameter in an init function that would allow us to set a configuration to use this new functionality. 

What do you think? In theory the new model should be better in almost every way, but I think its fair to ask if we can run an experiment with this and then come back and clean it up on fabric if that is indeed the case.

change/office-ui-fabric-react-2020-02-24-17-30-43-master.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
chang47: Minor now that we're updating a API

chang47:
@micahgodbolt everything is good on my end, can we merge the PR? Thanks!

PR #12047, opened 2020-02-25
Fixing prettier call inside the fluent-import script

kenotron:
#### Pull request checklist

#### Description of changes

prettier ignores d.ts files and we need to run it on those as well

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12047)

### DIFF
--- a/scripts/fluent-import/index.js
+++ b/scripts/fluent-import/index.js
@@ -313,6 +313,15 @@ function fixReactDeps(outputPath) {
   const file = path.join(outputPath, 'react/package.json');
   const contents = fs.readJsonSync(file);
   contents.devDependencies['@testing-library/jest-dom'] = '^5.1.1';
+
+  const keys = Object.keys(contents.devDependencies);
+  const sortedKeys = keys.sort();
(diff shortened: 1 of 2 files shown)


scripts/fluent-import/index.js
```diff
@@ -439,9 +449,30 @@ function importChangeLogMD(outputPath) {
   fs.copyFileSync(path.join(tmp, 'CHANGELOG.md'), path.join(outputPath, 'CHANGELOG.md'));
 }
 
-function runPrettier(outputPath, root) {
+function runPrettierOnImportedFiles(outputPath, root) {
   runPrettierForFolder(outputPath);
   runPrettierForFolder(path.join(root, '.github'));
+
+  console.log('running prettier on ' + path.join(outputPath, 'e2e/global.d.ts'));
+  console.log('node', [
```
ecraig12345: Can we update the helper with a way to do this instead of doing it manually?

PR #12049, opened 2020-02-25
chore(Label | AccordionTitle): remove Layout usage

mnajdova:
This PR removed the usages of the `Layout` component inside the `Label` and `AccordionTitle` components.

We don't want to depend on this component internally, so the necessary styles are just moved to the label and accordionTitle styles. In addion to this, we are rendering much less number of elements not.

**BREAKING CHANGES**
The `content` props in both `Label` and `AccordionTitle` are not shorthands for the `Box` component.

The `content` prop in the `AccordionTitleSlotClassNames` was renamed to `contentWrapper`.


packages/fluentui/react/src/themes/teams/components/Accordion/accordionTitleStyles.ts
```diff
@@ -8,6 +8,16 @@ const accordionTitleStyles: ComponentSlotStylesPrepared<AccordionTitleProps> = {
...
     cursor: p.disabled ? 'default' : 'pointer'
   }),
+  content: ({ props: p }) => ({
+    display: 'grid',
+    alignItems: 'center',
+    gridTemplateColumns: [
+      'auto', // indicator
+      !!p.content && '1fr'
+    ]
+      .filter(Boolean)
```
layershifter:
```tsx
gridTemplateColumns: `auto ${p.content ? '1fr' : ''}`
```

nit: Do we really need arrays magic there?

mnajdova: Not really, I just copied the logic from the layoutStyles. It's refactored now

Issue #2225, opened 2017-07-17
withViewport decorator does not respect viewport given as props to a decorated component


petdun2519:
#### Priorities and help requested (not applicable if asking question):

Are you willing to submit a PR to fix? Yes

Requested priority: Normal

Products/sites affected: Testing

#### Describe the issue:

Right now if I am trying to write some component unit tests for a component that renders a details list, it is impossible to get the details list to render, because it is wrapped in a withViewport decorator which checks for viewport size before rendering the child component. In a test environment there is often no actual viewport. So I would expect that for testing purposes, I should be able to override this behavior by supplying my own viewport via props. In reality, the decorator just ignores these props and keeps on reading viewport size from window/another parent. I would expect the withViewport decorator to respect the overriden props, or at least have some option to turn off it's viewport check before rendering child component, so that a unit test can be written for these components

Issue #2224, opened 2017-07-17
People picker always selects the first entry in the results, irrespective of which one you select


arvind1234:
### Bug Report

- __Package version(s)__: 4.15.0
- __Browser and OS versions__: chrome windows

#### Priorities and help requested (not applicable if asking question):

Are you willing to submit a PR to fix? No

Requested priority: Blocking (Hi Pri, blocking our release)

Products/sites affected: staffhub.ms

#### Describe the issue:

Have a normal people picker in your page. Start searching and select an item that is not the first one in the list. The picker always selects the first search result.

#### Actual behavior:

Have a normal people picker in your page. Start searching and select an item that is not the first one in the list. The picker always selects the first search result.

#### Expected behavior:

The picker should select the search result the user selects

### If applicable, please provide a codepen repro:
https://dev.office.com/fabric#/components/peoplepicker 

Video capture: https://1drv.ms/v/s!AgsF-JmM9dtAgyDCxyHfN20b4mDS

Type in a search query and select any entry, you will see the first one being selected

arvind1234:
The selection works fine, if you select something without entering any query

micahgodbolt:
this appears to be solved in the current release and only affecting some older releases.

lrecknagel:
Related to my Bugreport #2187 and was closed with: [Release 4.21.1](https://github.com/OfficeDev/office-ui-fabric-react/releases/tag/office-ui-fabric-react_v4.21.1)

PR #12050, opened 2020-02-25
Fluent : carousel accessibility fixes

kolaps33:
This PR is created based on the from another repository:
https://github.com/microsoft/fluent-ui-react/pull/2278/files

Related carousel: Carousel with Pagination

In this PR are fixes related to the feedback which we got from Rekha + feedback from testing.

1. 
Issue: Index numbers (1 of 4) are read twice as you down arrow through the carousel item.
Fix: add aria-hidden="true" on the text

2.1 Issue: The first and the last item in the carousel are not read when the prev/next button is pressed in browse mode.
2.2 #2224
Fix: moving focusing the next/previous paddle to generic function, which is call always, even when "onclick" is executed...

3.
Issue: Should there be focus on the entire carousel control and the left/right buttons? Seems redundant and adds more keyboard stops
Fix: change the behavior that for carousel without the "tab" navigation there will be tabindex=-1 on itemsContainer

4.
Issue: Carousel itself should have a label property defined. The roledescription is set to “Carousel” so label can be “Portrait collection” for this example
Fix: added prop aria-label to be able label the carousel

5.
Issue: #2225
Fix: was decided that if reader will narrate "carousel" user will know how to use keyboard to rotate through the carousel
For this reason to narrate "carousel" in application mode was added additional props(role, aria-label,aria-roledescription) into "itemsContainer" in "carouselBehavior.ts".
Now there is role "group". After more testing could be change for "region". For me region was not narrating "carousel" in focus mode for NVDA.

6.
Issue: Programmatically Prev/Next buttons are outside carousel container so it will likely confuse the user. Carousel container element should include all elements within carousel.
Fix: apply region only for "Carousel with Pagination", provide there proper label and roledescription. Fix is in "carouselBehavior.ts" for "root" slot.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12050)

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Carousel/BestPractices/CarouselBestPractices.tsx
+++ b/packages/fluentui/docs/src/examples/components/Carousel/BestPractices/CarouselBestPractices.tsx
@@ -14,7 +14,7 @@ const doList = [
   <Box>
     If carousel contains `navigation`:
     <ul>
-      <li> provide label to `navigation` and to navigation item using `aria-label` attribute</li>
+      <li> provide label to `navigation` slot and to navigation item using `aria-label` attribute</li>
       <li> add `aria-controls` attribute to navigation item referencing to `carouselItem` id </li>
     </ul>
   </Box>
--- a/packages/fluentui/docs/src/examples/components/Carousel/BestPractices/CarouselBestPractices.tsx
+++ b/packages/fluentui/docs/src/examples/components/Carousel/BestPractices/CarouselBestPractices.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Text } from '@fluentui/react';
+import { Text, Box } from '@fluentui/react';
 import { link } from '../../../../utils/helpers';
 
 import ComponentBestPractices from 'docs/src/components/ComponentBestPractices';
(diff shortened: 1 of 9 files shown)


JasonGore:
This PR looks like it succumbed to github issues that were occurring last night so I will close and reopen to trigger a new build.

aneeshack4:
@kolaps33 Any updates on this?

packages/fluentui/docs/src/examples/components/Carousel/BestPractices/CarouselBestPractices.tsx
```diff
@@ -8,7 +8,16 @@ const doList = [
...
     'Add textual representation for `CarouselItem`. Use `aria-label` attribute ( refer to{' '}
     {link('reported issue', 'https://bugs.chromium.org/p/chromium/issues/detail?id=1040924')} for details).
-  </Text>
+  </Text>,
+  'Provide  localized string of the "carousel" using `ariaRoleDescription` prop.',
+  'Provide label to the carousel using `ariaLabel` prop.',
+  <Box>
+    If carousel contains `navigation`:
+    <ul>
+      <li> provide label to `navigation` and to navigation item using `aria-label` attribute</li>
```
jurokapsiar: provide label to `navigation` slot?

packages/fluentui/docs/src/examples/components/Carousel/BestPractices/CarouselBestPractices.tsx
```diff
@@ -8,7 +8,16 @@ const doList = [
...
+      <li> add `aria-controls` attribute to navigation item referencing to `carouselItem` id </li>
```
jurokapsiar: do we have an example that uses `aria-controls`?

kolaps33: yes, already first Carousel example on the page 
https://fluentsite.z22.web.core.windows.net/components/carousel/definition
using 'aria-controls'

packages/fluentui/docs/src/examples/components/Carousel/BestPractices/CarouselBestPractices.tsx
kolaps33: change to "navigation slot"

PR #12051, opened 2020-02-25
being more specific with the codeowner file changes

kenotron:
#### Pull request checklist

#### Description of changes

With the influx of new code, we need to make sure the code owner triggers for the right set of files. This change makes it so that the CODEOWNERS are more specific about which package triggers what.

#### Focus areas to test

(optional)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -33,7 +33,7 @@ azure-pipelines.*.yml @ecraig12345 @kenotron @JasonGore
 sizeauditor.json @aftab-hassan
 
 #### Fluent UI
-packages/fluentui/ @JasonGore @kenotron
+packages/fluentui/ @davezuko @kuzhelov @layershifter @levithomason @miroslavstastny @mnajdova @kenotron @jdhuntington @ecraig12345 @dzearing @jasongore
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -33,7 +33,7 @@ azure-pipelines.*.yml @ecraig12345 @kenotron @JasonGore
 sizeauditor.json @aftab-hassan
 
 #### Fluent UI
-packages/fluentui/ @JasonGore @kenotron
+packages/fluentui/ @davezuko @kuzhelov @layershifter @levithomason @miroslavstastny @mnajdova @kenotron @jdhuntington @ecraig12345 @dzearing @jasongore
```
ecraig12345: Do I *have* to be listed here? :P

.github/CODEOWNERS
```diff
@@ -75,140 +75,136 @@ common/_semanticSlots.scss  @phkuo
 common/_themeOverrides.scss @phkuo
 common/_common.scss @phkuo
 
-### Utilities
```
ecraig12345: Wait why did you remove these?

PR #12060, opened 2020-02-25
For all gulp build:docs tasks that have a src->cache->dest, use cwd/base

kenotron:
#### Pull request checklist

#### Description of changes

We need to specify the cwd and base in these tasks because of the move away from a centralized gulp build call (it now can be that the cwd is different than paths.base). So for caching scenarios, we need to make sure the src, cache and dest use the correct base / cwd values.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12060)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -174,12 +174,6 @@ jobs:
       # No need for checkout since we're using build artifacts.
       - checkout: none
 
-      - task: DeleteFiles@1
-        inputs:
-          SourceFolder: /tmp/gulp-cache
-          Contents: '**/*'
-        condition: always()
-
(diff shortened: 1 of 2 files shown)


ecraig12345:
@msft-github-bot merge this pr in 1 minute

JasonGore:
Great! I'm hoping this fixes the regular build errors I get on fluentui packages after git updates locally, too. (I've been having to do full repo cleans to resolve.. even `yarn clean` didn't help.)

PR #12073, opened 2020-02-26
perf: add variables caching to v0

layershifter:
#### Pull request checklist

### WAT? 😱

This PR adds `enableBooleanVariablesCaching` which is disabled by default. It allows to handle styles caching for cases when boolean variables are used:

```tsx
<Provider performance={{ enableBooleanVariablesCaching: true }}>
  <Button variables={{ isFoo: true }} />
</Provider>
```

### Limitations 💔

We use `JSON.stringify()` for the implementation and it means that boolean variables should go in the same order, i.e. `Button`s below will create different cache entries:

```tsx
<>
  <Button variables={{ isFoo: true, isBar: true }} />
  <Button variables={{ isBar: true, isFoo: true }} />
</>
```

To avoid this the order should be always the same, proposed solution for customers: enforce the alphabetic order via ESLint rules.

### DIFF
--- a/packages/fluentui/react-bindings/src/hooks/useStyles.ts
+++ b/packages/fluentui/react-bindings/src/hooks/useStyles.ts
@@ -36,7 +36,8 @@ const defaultContext: StylesContextValue<{ renderRule: RendererRenderRule }> = {
   performance: {
     enableSanitizeCssPlugin: process.env.NODE_ENV !== 'production',
     enableStylesCaching: true,
-    enableVariablesCaching: true
+    enableVariablesCaching: true,
+    enableHardVariablesCaching: false
   },
   renderer: { renderRule: () => '' },
   theme: emptyTheme
--- a/packages/fluentui/react-bindings/src/hooks/useStyles.ts
+++ b/packages/fluentui/react-bindings/src/hooks/useStyles.ts
@@ -37,7 +37,7 @@ const defaultContext: StylesContextValue<{ renderRule: RendererRenderRule }> = {
     enableSanitizeCssPlugin: process.env.NODE_ENV !== 'production',
     enableStylesCaching: true,
     enableVariablesCaching: true,
-    enableHardVariablesCaching: false
+    enableBooleanVariablesCaching: false
   },
   renderer: { renderRule: () => '' },
   theme: emptyTheme
--- a/packages/fluentui/react-bindings/src/styles/resolveStyles.ts
+++ b/packages/fluentui/react-bindings/src/styles/resolveStyles.ts
@@ -43,20 +43,48 @@ const resolveStyles = (
   resolvedVariables: ComponentVariablesObject,
   renderStylesInput?: (styles: ICSSInJSStyle) => string
 ): ResolveStylesResult => {
-  const { className: componentClassName, theme, displayName, props, rtl, disableAnimations, renderer, performance } = options || {};
+  const { className: componentClassName, theme, displayName, props, rtl, disableAnimations, renderer, performance } = options;
 
   const { className, design, styles, variables, ...stylesProps } = props;
-  const noInlineOverrides = !(design || styles || variables);
 
-  const cacheEnabled = performance.enableStylesCaching && noInlineOverrides;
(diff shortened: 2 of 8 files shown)


packages/fluentui/react-bindings/src/styles/resolveStyles.ts
```diff
@@ -43,20 +43,48 @@ const resolveStyles = (
...
+        throw new Error('@fluentui/react: With "enableHardVariablesCaching" only plain objects can be passed to "variables" prop.');
+      }
+
+      const hasOnlyBooleanVariables = Object.keys(variables).every(variableName => {
+        return (
+          variables[variableName] === null || typeof variables[variableName] === 'boolean' || typeof variables[variableName] === 'undefined'
+        );
+      });
+
+      if (!hasOnlyBooleanVariables) {
```
mnajdova: We should not throw here, we should just not cache for this instance. We are not throwing in the other cases of caching, so we shouldn't here too. We cannot guarantee that **all** usages of variables are booleans in a project.

packages/fluentui/react-bindings/src/hooks/useStyles.ts
```diff
@@ -36,7 +36,8 @@ const defaultContext: StylesContextValue<{ renderRule: RendererRenderRule }> = {
   performance: {
     enableSanitizeCssPlugin: process.env.NODE_ENV !== 'production',
     enableStylesCaching: true,
-    enableVariablesCaching: true
+    enableVariablesCaching: true,
+    enableHardVariablesCaching: false
```
mnajdova: Why not `enableBooleanVariablesCaching`?

layershifter: Renamed, it's definitely sounds better

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
layershifter: During offline conversation we discovered issues and I changed the implementation to avoid throwing errors.

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
```diff
@@ -99,10 +124,11 @@ const resolveStyles = (
     }
   }
 
-  const componentCacheKey =
-    cacheEnabled && displayName && stylesProps
-      ? `${displayName}:${JSON.stringify(stylesProps)}${styleParam.rtl}${styleParam.disableAnimations}`
-      : '';
+  const propsCacheKey = cacheEnabled ? JSON.stringify(stylesProps) : '';
+  const variablesCacheKey = cacheEnabled && performance.enableBooleanVariablesCaching ? JSON.stringify(variables) : '';
```
mnajdova: For the props it was not problematic that the props order would change, because we are always mapping them in the same order. On the other hand for the variables we cannot guarantee that, as those are purely users input. I propose to always order them alphabetically and than we can safely use `JSON.stringify`, or clearly state somewhere that those should always be in the same order...

packages/fluentui/react-bindings/test/styles/resolveStyles-test.ts
```diff
@@ -260,4 +261,92 @@ describe('resolveStyles', () => {
...
+
+    test('requires "enableStylesCaching" to be enabled', () => {
+      const options = resolveStylesOptions({
+        performance: { enableStylesCaching: false, enableBooleanVariablesCaching: true }
+      });
+
+      expect(() => resolveStyles(options, resolvedVariables)).toThrowError(/Please check your "performance" settings on "Provider"/);
+    });
+
+    test('when enabled only plain objects can be passed as "variables"', () => {
```
mnajdova: Are these last two tests still valid?

layershifter: Yep, I changed their contents and updated titles

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
layershifter: Added this to **Limitations** section in PR description

PR #12074, opened 2020-02-26
Fluent: Remove bundle all packages task

JasonGore:
…with yarn build.

#### Pull request checklist

#### Description of changes

Remove fluent's `bundle:all-packages` script since it is now redundant with `yarn build` and is occasionally causing build errors in pipelines.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12074)

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -31,4 +31,4 @@ require('./scripts/gulp/tasks/test-circulars');
 require('./scripts/gulp/tasks/test-dependencies');
 
 // global tasks
-task('build', parallel('bundle:all-packages', 'build:docs'));
+task('build', parallel('build:docs'));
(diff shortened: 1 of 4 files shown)


gulpfile.ts
```diff
@@ -31,4 +31,4 @@ require('./scripts/gulp/tasks/test-circulars');
 require('./scripts/gulp/tasks/test-dependencies');
 
 // global tasks
-task('build', parallel('bundle:all-packages', 'build:docs'));
+task('build', parallel('build:docs'));
```
layershifter: Next step will be probably to remove `build` task, too

JasonGore: I can remove it as part of this PR, I just wasn't sure if `build:docs` is still needed separately from `yarn build` or not. 

I also noticed `build:docs:json` was an explicit step in CircleCI pipeline when auditing pipeline steps. However, it doesn't seem to be needed in ADO pipeline, so I'm guessing that's covered by `yarn build`.

layershifter: >I also noticed build:docs:json was an explicit step in CircleCI pipeline when auditing pipeline steps. However, it doesn't seem to be needed in ADO pipeline, so I'm guessing that's covered by yarn build.

For ADO it's not needed indeed

PR #12078, opened 2020-02-26
chore: refactor Flex to be functional component [v0]

layershifter:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `Flex` & `FlexItem` components to be functional. As in all other cases it contains a breaking change with restricted props set that will be passed to styles functions.

## Prop sets

| `Flex`    |
| --------- |
| `column`  |
| `debug`   |
| `fill`    |
| `gap`     |
| `hAlign`  |
| `inline`  |
| `padding` |
| `space`   |
| `vAlign`  |
| `wrap`    |

| `FlexItem`      |
| --------------- |
| `align`         |
| `grow`          |
| `flexDirection` |
| `push`          |
| `shrink`        |
| `size`          |


packages/fluentui/react/src/components/Flex/FlexItem.tsx
```diff
@@ -113,3 +67,81 @@ const applyStyles = (element: React.ReactElement<any>, styles, classes): React.R
...
+FlexItem.propTypes = {
+  ...commonPropTypes.createCommon({
+    children: false,
+    accessibility: false,
+    content: false
+  }),
+  children: PropTypes.oneOfType([PropTypes.element, PropTypes.func]),
+
+  align: PropTypes.oneOf(['auto', 'start', 'end', 'center', 'baseline', 'stretch']),
+  size: PropTypes.oneOfType([PropTypes.oneOf(['size.half', 'size.quarter', 'size.small', 'size.medium', 'size.large']), PropTypes.string]),
```
layershifter: Previously was incorrect 💣

packages/fluentui/react/src/components/Flex/Flex.tsx
```diff
@@ -38,63 +41,93 @@ export interface FlexProps extends UIComponentProps, ChildrenComponentProps {
...
 
-  static displayName = 'Flex';
-  static className = 'ui-flex';
+  const ElementType = getElementType(props);
+  const unhandledProps = getUnhandledProps(Flex.handledProps, props);
 
-  static defaultProps = {
-    as: 'div'
-  };
+  const content = React.Children.map(children, child => {
```
mnajdova: Let's refactor this with adding `FlexContextProvider`

packages/fluentui/react/src/components/Flex/FlexItem.tsx
mnajdova: A bit worried that this can be any css value (px or %) which may cause problems with the caching... :(

packages/fluentui/react/src/components/Flex/Flex.tsx
layershifter: Agree, can we do this thing separately? I want to make a small showcase from it

packages/fluentui/react/src/components/Flex/FlexItem.tsx
layershifter: Let's see the feedback. I agree there that it can be a real issue, but:
- I don't think that it's used enough frequently
- `size` manages `flexBasis`, may be in this case we should manage it via `design` prop...
- we can disable caching for `FlexItem`

PR #12085, opened 2020-02-26
Slider: allow setting id

### LINKED ISSUE #12080: Cannot set id on Slider using buttonProps.id
**Environment Information**

    Browser and OS versions: Edge + Narrator.

**Repro steps**

1. Navigate to slider control using tab key when Narrator is on.(To turn on Narrator use Ctrl + Win + Enter)
2. Verify narrator is reading the label places for slider or not.

**Code pin will be provided in comment section.**

**Actual:**
Narrator is reading only as "slider at 20 min 0 max 20" if the slider is at 20%.

**Expected:** 
Narrator should also read label information.
Ex: if the label placed is threshold then narrator should read as "Threshold slider at 20 min 0 max 20".

xugao:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #12080

#### Description of changes
Allow setting `id` instead of only letting `id` be generated internally.
Added unit test along with cleaning up mounted react wrapper.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12085)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Slider/Slider.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/Slider/Slider.test.tsx
@@ -3,13 +3,22 @@ import * as ReactDOM from 'react-dom';
 import * as renderer from 'react-test-renderer';
 import * as ReactTestUtils from 'react-dom/test-utils';
 
-import { mount } from 'enzyme';
+import { mount, ReactWrapper } from 'enzyme';
 import { Slider } from './Slider';
 import { ISlider } from './Slider.types';
 import { ONKEYDOWN_TIMEOUT_DURATION } from './Slider.base';
 import { KeyCodes } from '../../Utilities';
 
 describe('Slider', () => {
+  let wrapper: ReactWrapper | undefined;
+
+  afterEach(() => {
+    if (wrapper) {
+      wrapper.unmount();
+      wrapper = undefined;
+    }
+  });
+
   it('renders correctly', () => {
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/Slider/Slider.test.tsx
```diff
@@ -3,13 +3,22 @@ import * as ReactDOM from 'react-dom';
...
 
-import { mount } from 'enzyme';
+import { mount, ReactWrapper } from 'enzyme';
 import { Slider } from './Slider';
 import { ISlider } from './Slider.types';
 import { ONKEYDOWN_TIMEOUT_DURATION } from './Slider.base';
 import { KeyCodes } from '../../Utilities';
 
 describe('Slider', () => {
+  let wrapper: ReactWrapper | undefined;
```
aneeshack4: This seems great because we're not creating so many new consts down below & ReactWrapper lets us do a full DOM render in these tests - any other benefits/reasons you chose this (just out of curiosity)?

xugao: reason to have this declared here is so that we can unmount/clean it up after each test. see `afterEach` below. you can see a lot of similar pattern in our code base for other tests

PR #12087, opened 2020-02-26
Fix Fluent UI's snapshot script

Jahnp:
#### Pull request checklist

- [ ] Addresses an existing issue: N/A

#### Description of changes
Running `yarn update-snapshots` was failing on what I assume to be usage of outdated `jest:snapshots`. This PR passes the `-u` option instead, which is what's used elsewhere in the repo.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12087)

### DIFF
--- a/packages/fluentui/react-theming/package.json
+++ b/packages/fluentui/react-theming/package.json
@@ -28,8 +28,7 @@
     "lint:fix": "yarn lint:eslint --fix && yarn lint:old --fix",
     "lint:old": "tslint -t stylish -p .",
     "start-test": "just-scripts test:watch",
-    "test": "just-scripts test",
-    "update-snapshots": "just-scripts jest -u"
+    "test": "just-scripts test"
   },
   "dependencies": {
     "@types/classnames": "^2.2.9",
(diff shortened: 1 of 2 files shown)


packages/fluentui/react-theming/package.json
```diff
@@ -29,7 +29,7 @@
     "lint:old": "tslint -t stylish -p .",
     "start-test": "just-scripts test:watch",
     "test": "just-scripts test",
-    "update-snapshots": "just-scripts jest:snapshots"
```
kenotron: I think these scripts should get gone - we don't need a jest -u in these I think... otherwise, we should make sure these packages work with `just-scripts jest:snapshots`...

Jahnp: @kenotron gotcha, thanks for the heads-up. I hadn't been tracking the repo convergence work so this was just to unblock myself on another PR. Should I just remove line 32 altogether?

kenotron: Yeah that's what I mean. I hope this helps :)

Jahnp: Chatted offline; agreed to remove these lines.

PR #12089, opened 2020-02-26
Fix test-projects cleanup

ecraig12345:
Ensure that fluent test-projects cleans up after itself to help prevent the VMs from running out of space.

Also update test-projects to only pack the packages once (instead of doing it for every type of project test) and exclude fabric packages.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12089)

### DIFF
--- a/scripts/gulp/tasks/test-projects.ts
+++ b/scripts/gulp/tasks/test-projects.ts
@@ -1,12 +1,11 @@
 import express from 'express';
-import fs from 'fs';
+import fs from 'fs-extra';
 import { series, task } from 'gulp';
 import { rollup as lernaAliases } from 'lerna-alias';
 import path from 'path';
 import portfinder from 'portfinder';
 import puppeteer from 'puppeteer';
 import sh from '../sh';
-import del from 'del';
 
 import config from '../../config';
 import tmp from 'tmp';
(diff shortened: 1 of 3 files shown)


scripts/gulp/tasks/test-projects.ts
```diff
@@ -139,37 +159,40 @@ task('test:projects:cra-ts', async () => {
   const logger = log('test:projects:cra-ts');
   const scaffoldPath = paths.base.bind(null, 'scripts/gulp/tasks/test-projects/cra');
 
+  const tmpDirectory = createTempDir('project-cra-');
+  logger(`✔️ Temporary directory was created: ${tmpDirectory}`);
```
kenotron: way to stick to the convention :)

ecraig12345:
@layershifter or @miroslavstastny I went ahead and merged this to reduce immediate issues with boxes running out of space, but if you have any comments I can address them in another PR.

scripts/gulp/tasks/test-projects.ts
```diff
@@ -17,16 +16,25 @@ type PackedPackages = Record<string, string>;
...
 const log = (context: string) => (message: string) => {
   console.log();
   console.log('='.repeat(80));
   console.log(`${context} : ${message}`);
   console.log('='.repeat(80));
 };
 
-export const runIn = targetPath => cmd => sh(cmd, targetPath);
+const createTempDir = (prefix: string): string => {
+  return tmp.dirSync({ prefix, unsafeCleanup: true }).name;
```
ecraig12345: "Unsafe" means clean up even if there are still files in the directory (so perfectly safe in our case). Should probably have added that in a comment.

PR #12093, opened 2020-02-27
Fix CI environment variable checks

ecraig12345:
Switch checks for CircleCI environment variables (such as `process.env.CI`) to the corresponding ADO variables. Among other things, this helps reduce the amount of log output which is not helpful in CI, such as from ProgressPlugin.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12093)


kenotron:
looks like some of this is duped by @ecraig12345 's fix here #12093 FYI

PR #12097, opened 2020-02-27
Eventing proto

layershifter:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12097)


packages/fluentui/react-component-event-listener/src/EventListenerWithComputedTarget.ts
```diff
@@ -0,0 +1,26 @@
...
+type EventListenerWithComputedTargetProps<T extends EventTypes> = Omit<EventListenerOptions<T>, 'target' | 'targetRef'> & {
+  resolveTarget: () => Target;
+};
+
+function EventListenerWithComputedTarget<T extends EventTypes>(props: EventListenerWithComputedTargetProps<T>) {
+  const { resolveTarget, ...rest } = props;
+  const targetRef = React.useRef<Target | null>(null);
+
+  React.useEffect(() => {
+    targetRef.current = resolveTarget();
```
aneeshack4: just to clarify my own understanding, in this effect hook, the return value of resolveTarget passed in from props is set to targetRef.current? if so, what is the [resolveTarget] doing below?

packages/fluentui/react-component-event-listener/src/EventListenerWithComputedTarget.ts
```diff
@@ -0,0 +1,26 @@
...
+  }, [resolveTarget]);
+  useEventListener({ ...rest, target: undefined, targetRef });
```
aneeshack4: how come target is set to undefined here?

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -303,22 +304,11 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
...
-
-    this._events.on(window, 'resize', this._onAsyncResize);
-    if (this._root.current) {
-      this._events.on(this._root.current, 'focus', this._onFocus, true);
-    }
-    if (this._scrollElement) {
-      this._events.on(this._scrollElement, 'scroll', this._onScroll);
-      this._events.on(this._scrollElement, 'scroll', this._onAsyncScroll);
-    }
+    this._scrollElement = this._getScrollableElement();
```
aneeshack4: why this call to the getter instead of the existing find call on 306?

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -400,14 +390,37 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
     }
 
     return (
-      <div ref={this._root} {...divProps} role={pageElements.length > 0 ? role : undefined} className={css('ms-List', className)}>
-        <div ref={this._surface} className={'ms-List-surface'} role="presentation">
-          {pageElements}
+      <>
```
aneeshack4: stray element

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -400,14 +390,37 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
...
+          capture
+          resolveTarget={this._getScrollableElement}
+          type="scroll"
+          listener={this._onAsyncScroll.bind(this)}
+        />
+      </>
     );
   }
 
+  private _getScrollableElement = (): HTMLElement => findScrollableParent(this._root.current) as HTMLElement;
```
aneeshack4: how come this getter is now being to the finder's value after its usage in resolveTarget above?

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -400,14 +390,37 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
...
+          capture
+          resolveTarget={this._getScrollableElement}
+          type="scroll"
+          listener={this._onScroll.bind(this)}
+        />
+        <EventListenerWithComputedTargetProps
+          capture
+          resolveTarget={this._getScrollableElement}
+          type="scroll"
+          listener={this._onAsyncScroll.bind(this)}
```
aneeshack4: I think we need to update the List tests - they are failing but not for the same reasons as before - those I think were fixed with this new resolveTarget approach but I think now sometimes the component is undefined as a result: https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=54933&view=logs&j=92058b04-f16a-5035-546c-cae3ad5e2f5f&t=17e772b0-f0c3-53ab-7aac-ad6df80e30d7&l=558
image

packages/fluentui/react-component-event-listener/src/EventListenerWithComputedTarget.ts
layershifter: Oops, it should not be there

packages/fluentui/react-component-event-listener/src/EventListenerWithComputedTarget.ts
layershifter: We will use `resolveTarget()` as a delayed executor.

```tsx
  React.useEffect(() => {
    targetRef.current = resolveTarget();
  }, [resolveTarget]);
```

You're right, this `useEffect` will call `resolveTarget` and assign it to `targetRef`.

```tsx
useEventListener({ ...rest, targetRef });
```

`useEventListener` will use `targetRef`.

### How it's different?

`targetRef` will be resolved inside `EventListenerWithComputedTarget` instead of `List` and guarantees the proper priority of `ref` resolutions.

PR #12107, opened 2020-02-27
making sure that the resolve.alias do NOT skip over the fluentui packages

kenotron:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12107 

#### Description of changes

This fixes the resolve.alias not being set properly for FUI packages. The issues that blocked us are:

1. FUI packages are currently all `"private": true`
2. the main fields of FUI packages are pointing at `dist/es` or `dist/commonjs` instead of `lib`

So we address both of these concerns with this PR - however, I believe new code should not depend on code deeply inside the packages from same monorepo. That isn't enforced currently, so we will support the deep imports for now.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12107)

### DIFF
--- a/scripts/webpack/getResolveAlias.js
+++ b/scripts/webpack/getResolveAlias.js
@@ -4,26 +4,48 @@ const findRepoDeps = require('../monorepo/findRepoDeps');
 const findGitRoot = require('../monorepo/findGitRoot');
 const { readConfig } = require('../read-config');
 
+function getOutputPath(packageJson) {
+  for (const mainField of ['module', 'main']) {
+    const main = packageJson[mainField];
+    if (main && main.includes('dist/es')) {
+      return 'dist/es';
+    } else {
+      return 'lib';
+    }
+  }
+
+  return 'lib';
+}
+
 function getResolveAlias() {
   const gitRoot = findGitRoot();
   const deps = findRepoDeps();
+
   const alias = {};
-  const excludedPackages = ['@uifabric/api-docs'];
+  const excludedPackages = ['@uifabric/api-docs', '@uifabric/build'];
 
   let cwd = process.cwd();
   const packageJson = readConfig(path.join(cwd, 'package.json'));
 
   deps.forEach(depInfo => {
-    if (!depInfo.packageJson.private && !excludedPackages.includes(depInfo.packageJson.name)) {
+    if (!excludedPackages.includes(depInfo.packageJson.name)) {
       alias[`${depInfo.packageJson.name}$`] = path.join(gitRoot, depInfo.packagePath, 'src');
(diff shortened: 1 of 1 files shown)


scripts/webpack/getResolveAlias.js
```diff
@@ -4,26 +4,48 @@ const findRepoDeps = require('../monorepo/findRepoDeps');
...
 
   alias[`${packageJson.name}$`] = path.join(cwd, 'src');
   alias[`${packageJson.name}/src`] = path.join(cwd, 'src');
-  alias[`${packageJson.name}/lib`] = path.join(cwd, 'src');
+
+  const outputPath = getOutputPath(packageJson);
+
+  alias[`${packageJson.name}/${outputPath}`] = path.join(cwd, 'src');
+
+  console.log(alias);
```
xugao: remove?

scripts/webpack/getResolveAlias.js
```diff
@@ -4,26 +4,48 @@ const findRepoDeps = require('../monorepo/findRepoDeps');
...
+  return 'lib';
+}
+
 function getResolveAlias() {
   const gitRoot = findGitRoot();
   const deps = findRepoDeps();
+
   const alias = {};
-  const excludedPackages = ['@uifabric/api-docs'];
+  const excludedPackages = ['@uifabric/api-docs', '@uifabric/build'];
```
ecraig12345: Can probably also exclude infra stuff like tslint and eslint rules?

scripts/webpack/getResolveAlias.js
```diff
@@ -4,26 +4,48 @@ const findRepoDeps = require('../monorepo/findRepoDeps');
...
 const { readConfig } = require('../read-config');
 
+function getOutputPath(packageJson) {
+  for (const mainField of ['module', 'main']) {
+    const main = packageJson[mainField];
+    if (main && main.includes('dist/es')) {
+      return 'dist/es';
+    } else {
+      return 'lib';
+    }
```
ecraig12345: Nesting is wrong (you're returning `lib` if the field doesn't exist). One possible workaround:
```suggestion
    if (main) {
      return main.includes('dist/es') ? 'dist/es' : 'lib';
    }
```

scripts/webpack/getResolveAlias.js
kenotron: sigh, was hoping nobody would catch me before i updated :)

scripts/webpack/getResolveAlias.js
kenotron: added them

PR #12111, opened 2020-02-27
Faster builds via better gulp-cache

kenotron:
#### Pull request checklist

#### Description of changes

Here we re-enable the pipelines to be all self-hosted and now use gulp-cache to greater effect.

I think this brings it down from 39 to 27minutes... there are more opportunities to drive it down even further.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12111)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -177,7 +177,7 @@ jobs:
           Contents: '**/*'
         condition: always()
 
-  - job: Screener FluentUI
+  - job: ScreenerFluent
     dependsOn: Build
     steps:
       # No need for checkout since we're using build artifacts.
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -173,18 +177,13 @@ jobs:
           Contents: '**/*'
         condition: always()
 
-  - job: Screener
+  - job: ScreenerFluent
```
JasonGore: :tada:

PR #12113, opened 2020-02-27
Text: use semantic color slot as a default.

### LINKED ISSUE #12109: Text and Stack components don't apply theme when calling Customizations.applySettings
### Environment Information

- **Package version(s)**: 7.48.1 (but I believe the issue is valid on the latest version too)

https://codepen.io/up2pixy/pen/QWbvpax?editors=0010

#### Actual behavior:

When calling `Customizations.applySettings`, the Stack and Text don't change their styles like other components such as Link and Button. Seems like they are not wrapped with `styled()` HOC

#### Expected behavior:

When calling `Customizations.applySettings`, the Stack and Text component will listen to the event to update the styles with the updated theme.

Are you willing to submit a PR to fix?  Yes if we want them to be `styled`

Requested priority: Low

Products/sites affected: (if applicable) N/A

aneeshack4:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12109 

#### Description of changes

The `Text` component isn't compatible with applying theme through `Customizations.applySettings` because the `color` in `Text.styles.ts` doesn't directly use Fabric's palette or semantic color slots - instead it goes through the half-baked fonts which don't have color set properly. So this is an attempt to set the default color to be a semantic slot (`bodyText` has the value of `neutralPrimary` which is appropriate and typically used for foregroundColor which is text) so that if the font color is undefined (`variantObject.color`), then it'll fall to a defined theme color. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12113)

### DIFF
--- a/change/office-ui-fabric-react-2020-02-27-14-35-07-text.json
+++ b/change/office-ui-fabric-react-2020-02-27-14-35-07-text.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Text: use semantic color slot as a default",
+  "packageName": "office-ui-fabric-react",
+  "email": "aneeshak@microsoft.com",
+  "commit": "9f2a8f9d54b6eb37934f1b07243bc9bb11546837",
+  "dependentChangeType": "patch",
+  "date": "2020-02-27T22:35:07.408Z"
+}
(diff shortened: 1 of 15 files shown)


aneeshack4:
Closing and reopening to trigger another build because the build error seems random "Archive creation failed for archive file"

aneeshack4:
Ok this does regress a lot on screener - thinking of another semantic slot we can use or another solution

aneeshack4:
This will need a different approach, going to close this for now.

PR #12114, opened 2020-02-27
Added ApplicationInsights to public site to enable usage tracking

joschemd:
#### Pull request checklist

#### Description of changes

Added Application Insights telemetry to public fabric site

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12114)

### DIFF
--- a/apps/fabric-website/index.html
+++ b/apps/fabric-website/index.html
@@ -53,6 +53,80 @@
         outline: none;
       }
     </style>
+    <script type="text/javascript">
+      var appInsights =
+        window.appInsights ||
+        (function(e) {
+          function n(e) {
+            t[e] = function() {
+              var n = arguments;
+              t.queue.push(function() {
+                t[e].apply(t, n);
+              });
+            };
+          }
+          var t = { config: e };
+          t.initialize = !0;
+          var i = document,
+            a = window;
+          setTimeout(function() {
+            var n = i.createElement('script');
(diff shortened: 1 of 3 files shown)


apps/fabric-website/index.html
```diff
@@ -53,6 +53,80 @@
         outline: none;
       }
     </style>
+    <script type="text/javascript">
```
ecraig12345: Adding this here is okay, but there are a couple other locations you'll need to add it to actually make it be used on the public site (this file is only used for the PR deploy of the website and a couple other secondary locations). 

The file that's actually used on the public site is in a private repo which I can point you to, but fabric-website/homepage.htm is a mirror of that file. So typically we check in updates to homepage.htm (and index.html) here first, then port them to the private repo.

ecraig12345:
@aneeshack4 We can ignore the screener failure in this case since there's no way it could be caused by this change. Once homepage.htm is also updated I'll merge it.

apps/fabric-website/index.html
joschemd: Completed .. will now add to the public-internal

micahgodbolt:
@joschemd is working with dev.msft to get the rest of this hooked up

PR #12115, opened 2020-02-28
Allowing overflow set to be reversed

mltejera:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000 // Should I file an issue simply for tracking purposes?

#### Description of changes

Our scenario requires reversed visual behavior (rtl) but we'd still like to maintain proper keyboard navigation (ltr). We can achieve this simply by reversing the DOM order of the elements.

#### Focus areas to test

Make sure elements behave and reflow as you'd expect.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12115)


mltejera:
@brandonthomas FYI

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -9,6 +9,8 @@ import { IOverflowSet, IOverflowSetItemProps, IOverflowSetProps, IOverflowSetSty
 
 const getClassNames = classNamesFunction<IOverflowSetStyleProps, IOverflowSetStyles>();
 
+const defaultIsReversed = false;
```
brandonthomas: I think you can just keep this inline with your destructure

mltejera: I have to destructure twice though, so I'd rather it be tied to a constant.

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.types.ts
```diff
@@ -54,6 +54,13 @@ export interface IOverflowSetProps extends React.ClassAttributes<OverflowSetBase
    */
   vertical?: boolean;
 
+  /**
+   * Controls wether or not the DOM order is reversed.
```
brandonthomas: specify what is reversed. It's not quite the whole thing, just the placement of the overflow button vs the rest.

mltejera: updated to only reverse overflow position

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.types.ts
```diff
@@ -54,6 +54,13 @@ export interface IOverflowSetProps extends React.ClassAttributes<OverflowSetBase
    */
   vertical?: boolean;
 
+  /**
+   * Controls wether or not the overflow button is placed on the right or left of the items.
+   * This gives a reveresed visual behavior but maintains correct keyboard navigation.
+   * @defaultValue false
+   */
+  isOverflowReversed?: boolean;
```
micahgodbolt: now about "overflowSide": "start" | "end"

with default being end

mltejera:
@micahgodbolt I'm getting linter errors from deprecated api's that aren't related to my change.

Are those going to block this getting in?

micahgodbolt:
yeah, linting error is on "toolbar" one of the new controls pulled over from teams. you certainly aren't to blame there. merge master back in and give it another shot.

mltejera:
Looks like the linter errors are specific on over flow set?

micahgodbolt:
Ah that makes sense. FocusZone got moved to it's own package. The old import is deprecated.

PR #12116, opened 2020-02-28
De-duplicate babel dependencies

ecraig12345:
Duplicate versions of babel-related dependencies appear to have been contributing to spurious typing errors. Manually editing `yarn.lock` to combine duplicate entries then re-running `yarn` (to remove entries which aren't needed due to the deduplication) should fix this.

I also updated parseType.ts which is where the problem was surfacing.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12116)

### DIFF
--- a/scripts/gulp/plugins/util/parseType.ts
+++ b/scripts/gulp/plugins/util/parseType.ts
@@ -32,12 +32,12 @@ const normalizeType = (propType: string): string => {
 };
 
 const getTypeFromBabelTree = (componentFile: t.File, componentName: string, propName: string) => {
-  let typeAnnotation: t.TSType;
+  let typeAnnotation: t.TSType | undefined;
 
   const propertyVisitor: Babel.Visitor = {
     TSPropertySignature: path => {
       if (path.get('key').isIdentifier({ name: propName })) {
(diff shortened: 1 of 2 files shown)


scripts/gulp/plugins/util/parseType.ts
```diff
@@ -32,12 +32,12 @@ const normalizeType = (propType: string): string => {
...
 
 const getTypeFromBabelTree = (componentFile: t.File, componentName: string, propName: string) => {
-  let typeAnnotation: t.TSType;
+  let typeAnnotation: t.TSType | undefined;
 
   const propertyVisitor: Babel.Visitor = {
     TSPropertySignature: path => {
       if (path.get('key').isIdentifier({ name: propName })) {
-        const annotationPath: NodePath<t.TSTypeAnnotation> = path.get('typeAnnotation');
+        const annotationPath = path.get('typeAnnotation') as NodePath<t.TSTypeAnnotation>;
```
ecraig12345: Without the cast this is `NodePath<t.TSTypeAnnotation | null>` and I wasn't sure how to properly detect/handle the null case.

PR #12128, opened 2020-02-28
Make screener for FUI even faster by not depending on a build

kenotron:
#### Pull request checklist

#### Description of changes

This will make the Screener FUI job start earlier and rely on its own webpack and build:docs step to generate the bundles. This should cut the build times from 27minutes to 20minutes:

image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12128)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -146,9 +146,6 @@ jobs:
   - job: Deploy
     dependsOn: Build
     steps:
-      # No need for checkout since we're using build artifacts.
-      - checkout: none
-
       - template: azure-pipelines.cache.yml
 
       - template: azure-pipelines.artifacts.yml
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -146,9 +146,6 @@ jobs:
   - job: Deploy
     dependsOn: Build
     steps:
-      # No need for checkout since we're using build artifacts.
-      - checkout: none
-
```
JasonGore: Why was this removed for Deploy?

PR #12130, opened 2020-02-28
DetailsList: updated drag drop example to be more intuitive, allow dragging to the last row

### LINKED ISSUE #11675: DetailsList: unable to drag and drop item to the last row
The drag and drop for rows (an example with 2 is enough) is very janky. Sometimes it works, sometimes it doesn't when dragging a row from bottom up. If you try to drag from top down that never works.

### Environment Information

- **Package version(s)**: latest 7
- **Browser and OS versions**: Windows 10. Firefox 73.0a1

https://codepen.io/dbi3j2934/pen/abzGJOG?editors=0010

#### Actual behavior:

Dragging row 2 to row 1 sometimes works, you have to place it in a very specific position for it to work. Dragging row 1 to row 2 (or slightly below or whatever) just doesn't work.

#### Expected behavior:

Whenever the rows overlap the dragged row should replace the destination row in true drag & drop fashion. 

Are you willing to submit a PR to fix?  No

Products/sites affected: (if applicable) Mission Control

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11675

#### Description of changes
Updated the Drag and Drop DetailsList example:
1. Put the dragged row onto the row it got dropped onto (this is a more intuitive experience)
2. By doing 1, we also enable dragging any row the the last row (it's currently impossible to do so)
3. Removed an odd logic: dragging and dropping onto a same row put the row as first row

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12130)

### DIFF
--- a/change/office-ui-fabric-react-2020-02-28-15-48-39-xgao-drag-drop-dl.json
+++ b/change/office-ui-fabric-react-2020-02-28-15-48-39-xgao-drag-drop-dl.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "DetailsList Example: change drag drop behavior to be more intuitive, allow dragging to the last row",
+  "packageName": "office-ui-fabric-react",
+  "email": "xgao@microsoft.com",
+  "commit": "7061c70910e6d5078c7405c839fccb4c05e2e9b1",
+  "dependentChangeType": "patch",
+  "date": "2020-02-28T23:48:39.855Z"
+}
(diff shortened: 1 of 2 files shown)


danielwnn:
http://clauderic.github.io/react-sortable-hoc/#/basic-configuration/drag-handle?_k=gcvn0j

Is there a way to add the drag handle in the far left to indicate the rows are draggable? Thank you!

PR #12142, opened 2020-03-02
chore(Image): replace `withSafeTypeForAs`

layershifter:
#### Pull request checklist

#### Description of changes

Currently we have `withSafeTypeForAs` utility which partially solves polymorphic typings for `as` prop but it's too complex and creates over-complicated [type definitions](https://cdn.jsdelivr.net/npm/@fluentui/react@0.44.0/dist/es/components/Image/Image.d.ts) in `dist`.

This PR removes usage of `withSafeTypeForAs` in `Image` component with simple typings. Approach is inspired by recent changes in emotion: emotion-js/emotion#1405 & [`react-polymorphic-box`](https://github.com/kripod/react-polymorphic-box). It's also compatible `React.forwardRef`.

#### Focus areas to test

I made a test snippet that allows to test all possible cases with broken props, feel free to add missing things 👍

```tsx
import * as React from "react";
import { Image, Button } from "@fluentui/react";

const ImageExample = () => (
  <>
    {/* Wrong default props */}
    <Image src="public/images/wireframe/square-image.png" to="Foo" />
    <Image src="public/images/wireframe/square-image.png" type="Foo" />

    {/* Wrong elements */}
    <Image as="aa" src="public/images/wireframe/square-image.png" />
    <Image as="blockq" src="public/images/wireframe/square-image.png" />

    {/* Wrong attrs on els */}
    <Image
      as="button"
      src="public/images/wireframe/square-image.png"
      type="Foo"
    />
    <Image as="a" src="public/images/wireframe/square-image.png" href1="Foo" />
    <Image
      as="button"
      src="public/images/wireframe/square-image.png"
      href="Foo"
    />

    {/* Wrong attr on components */}
    <Image
      as={Button}
      src="public/images/wireframe/square-image.png"
      type="Foo"
    />
    <Image
      as={Button}
      src="public/images/wireframe/square-image.png"
      typ="Foo"
    />
  </>
);

export default ImageExample;
```

Fixes https://github.com/microsoft/fluent-ui-react/issues/1527.

***

I also have to bump `@types/react` as `16.8.11` doesn't contain a needed change as cases like this will not work:
```tsx
<Image as="a" src="public/images/wireframe/square-image.png" href1="Foo" />
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12142)

### DIFF
--- a/apps/a11y-tests/package.json
+++ b/apps/a11y-tests/package.json
@@ -31,7 +31,7 @@
     "@types/jest": "~24.9.0",
     "@types/node": "^10.3.2",
     "@types/puppeteer": "1.12.3",
-    "@types/react": "16.8.11",
+    "@types/react": "16.8.25",
     "@types/react-dom": "16.8.4",
     "@types/sarif": "^2.1.1",
     "@uifabric/build": "^7.0.0",
--- a/apps/dom-tests/package.json
+++ b/apps/dom-tests/package.json
@@ -30,7 +30,7 @@
     "@types/es6-promise": "0.0.32",
     "@types/jest": "~24.9.0",
     "@types/puppeteer": "1.12.3",
-    "@types/react": "16.8.11",
+    "@types/react": "16.8.25",
     "@types/react-dom": "16.8.4",
     "@types/webpack-env": "1.15.1",
     "@uifabric/build": "^7.0.0",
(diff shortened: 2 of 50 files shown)


layershifter:
Was implemented in `compose()`.

PR #12145, opened 2020-03-02
feat(MenuItem): remove icon dependency

mnajdova:
#### Pull request checklist

#### Description of changes

This PR removes the internal usage of the `Icon` component inside the `MenuItem` for the submenu indicator. The icon that was previously rendered as submenu indicator, is now added as part of the styles in the theme.

As the colors used inside these icons can be theme dependent, I've created a functions which receives some parameters (colors) and is returning the svg url string which is used as a background image.

The alternative approach that was tried was adding the css filter prop, but as it does not allow direct color definition, so it may be hard to adjust the color in every theme to get to the perfect hex value desired. On top of this, it is not supported in IE11.

### BREAKING CHANGE
The `indicator` prop in the `MenuItem` and `Menu` components is now `ShorthandValue<BoxProps>` instead of `ShorthandValue<IconProps>`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12145)


packages/fluentui/react/src/components/Menu/MenuItem.tsx
```diff
@@ -220,12 +215,16 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
             defaultProps: () => ({ as: 'span', styles: styles.content })
           })}
           {menu &&
-            Icon.create(indicatorWithDefaults, {
-              defaultProps: () => ({
-                name: vertical ? 'icon-menu-arrow-end' : 'icon-menu-arrow-down',
-                styles: styles.indicator
-              })
-            })}
+            Box.create(
```
layershifter: Should be a `div`?

packages/fluentui/react/src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -5,6 +5,8 @@ import { MenuItemProps, MenuItemState } from '../../../../components/Menu/MenuIt
 import { getColorScheme } from '../../colors';
 import getIconFillOrOutlineStyles from '../../getIconFillOrOutlineStyles';
 import { ComponentSlotStylesPrepared, ICSSInJSStyle } from '@fluentui/styles';
+import svgContentHorizontal from './submenuIndicatorUrl';
+import svgContentVertical from './submenuIndicatorVerticalUrl';
```
layershifter: nit: I would rather keep them in the same file

packages/fluentui/react/src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -411,7 +413,17 @@ const menuItemStyles: ComponentSlotStylesPrepared<MenuItemPropsAndState, MenuVar
       top: pxToRem(6),
       right: pxToRem(2),
       left: 'unset'
-    })
+    }),
+
+    backgroundImage: p.vertical ? svgContentVertical(v.color) : svgContentHorizontal(v.color),
```
layershifter:
```suggestion
    backgroundImage: svgIndicator(v.color, p.vertical)
```

```suggestion
    backgroundImage: svgIndicator(v.color, p.vertical ? 'up' : 'right')
```

Not fully sure about it

packages/fluentui/react/src/themes/teams-high-contrast/components/Menu/menuItemStyles.ts
```diff
@@ -15,29 +17,44 @@ const menuItemStyles: ComponentSlotStylesPrepared<MenuItemPropsAndState, MenuVar
...
+            backgroundImage: p.vertical ? svgContentVertical(v.colorActive) : svgContentHorizontal(v.colorActive)
+          }
         }),
 
       ...((iconOnly || vertical) && {
         ...(isFromKeyboard && {
           color: v.colorActive,
-          background: v.backgroundColorFocus
+          background: v.backgroundColorFocus,
+          '& .ui-menu__item__indicator': {
```
layershifter: Should `slotClasses` be used there?

packages/fluentui/react/src/themes/teams/components/Menu/menuItemStyles.ts
mnajdova: I like the first option, we will have them in one function, hence avoiding two files. Would not like the second one, in order to avoid up, right terms which are not rtl safe..

PR #12146, opened 2020-03-02
Chore: Moving TableCell to be functional

pompomon:
#### Pull request checklist

#### Description of changes

This PR converts `TableCell` component to be functional. As in all other cases it contains a breaking change with restricted props set that will be passed to styles functions.

|TableCell|
|-|
|truncateContent|

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12146)

### DIFF
--- a/packages/fluentui/react/src/components/Table/TableCell.tsx
+++ b/packages/fluentui/react/src/components/Table/TableCell.tsx
@@ -1,30 +1,29 @@
 import * as customPropTypes from '@fluentui/react-proptypes';
-import { Accessibility, tableCellBehavior } from '@fluentui/accessibility';
+import { Accessibility, tableCellBehavior, TableCellBehaviorProps } from '@fluentui/accessibility';
 import { Ref } from '@fluentui/react-component-ref';
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
 import * as _ from 'lodash';
 import {
-  UIComponent,
   childrenExist,
-  RenderResultConfig,
-  UIComponentProps,
   ChildrenComponentProps,
   commonPropTypes,
   ContentComponentProps,
-  ShorthandFactory,
   createShorthandFactory,
-  applyAccessibilityKeyHandlers
+  UIComponentProps
 } from '../../utils';
+import { useTelemetry, useStyles, getElementType, getUnhandledProps, useAccessibility } from '@fluentui/react-bindings';
+// @ts-ignore
+import { ThemeContext } from 'react-fela';
 import Box, { BoxProps } from '../Box/Box';
-import { WithAsProp, ShorthandValue, withSafeTypeForAs } from '../../types';
+import { WithAsProp, ShorthandValue, withSafeTypeForAs, FluentComponentStaticProps, ProviderContextPrepared } from '../../types';
 
 export interface TableCellProps extends UIComponentProps, ChildrenComponentProps, ContentComponentProps<ShorthandValue<BoxProps>> {
   /**
    * Accessibility behavior if overridden by the user.
    * @available TableCellBehavior
    * */
-  accessibility?: Accessibility;
+  accessibility?: Accessibility<TableCellBehaviorProps>;
 
   /**
    * Truncate cell's content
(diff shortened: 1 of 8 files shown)


packages/fluentui/react/src/components/Table/TableCell.tsx
```diff
@@ -1,108 +1,135 @@
...
+        {...getA11yProps('root', {
+          className: classes.root,
+          onClick: handleClick,
+          ...unhandledProps
+        })}
+      >
+        {hasChildren
+          ? children
+          : Box.create(content, {
+              defaultProps: () => ({ as: 'div', styles: resolvedStyles.content })
```
mnajdova: No need for `as: 'div'` as it is the default

packages/fluentui/react/src/components/Table/TableCell.tsx
```diff
@@ -1,108 +1,135 @@
...
+      <ElementType
+        {...getA11yProps('root', {
+          className: classes.root,
+          onClick: handleClick,
+          ...unhandledProps
+        })}
+      >
+        {hasChildren
+          ? children
+          : Box.create(content, {
```
mnajdova: We should propagate the getA11yProps's content props to this element

packages/fluentui/react/src/themes/teams/components/Table/tableCellStyles.ts
```diff
@@ -25,7 +28,7 @@ export default {
       height: '100%'
     };
   },
-  content: ({ props: { truncateContent } }: ComponentStyleFunctionParam<TableCellProps, TeamsTableVariables>): ICSSInJSStyle => {
+  content: ({ props: { truncateContent } }: ComponentStyleFunctionParam<TableCellStylesProps, TeamsTableVariables>): ICSSInJSStyle => {
```
mnajdova: It may be better if you just create a const with the typing and avoid adding typings on each slot, something like:
```
const tableCellStyles: ComponentSlotStylesPrepared<TableCellStyleProps, TeamsTableVariables> = {
  // slots here
}
```

packages/fluentui/react/src/themes/teams/types.ts
```diff
@@ -110,6 +113,9 @@ export type TeamsThemeStylesProps = {
   HierarchicalTreeItem?: HierarchicalTreeItemProps;
   HierarchicalTreeTitle?: HierarchicalTreeTitleProps;
   Video?: VideoProps;
+  Table?: TableProps;
```
mnajdova: 👍

packages/fluentui/react/src/components/Table/TableCell.tsx
```diff
@@ -1,108 +1,135 @@
...
+    customPropTypes.disallow(['children']),
+    PropTypes.oneOfType([PropTypes.arrayOf(customPropTypes.nodeContent), customPropTypes.nodeContent])
+  ]),
+  truncateContent: PropTypes.bool
+};
+
+TableCell.handledProps = Object.keys(TableCell.propTypes) as any;
+
+TableCell.defaultProps = {
+  accessibility: tableCellBehavior as Accessibility
```
layershifter:
```suggestion
  accessibility: tableCellBehavior
```

Is not needed anymore

packages/fluentui/react/src/components/Table/TableCell.tsx
```diff
@@ -1,108 +1,135 @@
...
-            ? children
-            : Box.create(content, {
-                defaultProps: () => ({ as: 'div', styles: styles.content })
-              })}
-        </ElementType>
-      </Ref>
-    );
-  }
-}
+  const tableCell = (
```
layershifter:
```suggestion
  const element = (
```

_nit:_ We used `element` everywhere

packages/fluentui/react/src/components/Table/TableCell.tsx
```diff
@@ -1,108 +1,135 @@
...
+      className,
+      design,
+      styles,
+      variables
+    }),
+    rtl: context.rtl
+  });
 
-  handleClick = (e: React.SyntheticEvent) => {
+  const handleClick = (e: React.SyntheticEvent) => {
```
layershifter:
```suggestion
  const handleClick = (e: React.MouseEvent | React.KeyboardEvent) => {
```

Let's be more explicit

packages/fluentui/react/src/components/Table/TableCell.tsx
pompomon: Agreed with @mnajdova and @layershifter to keep current change like that, so we don't break existing behaviour (where we did not spread any props to the Box element), but we will revise it later

packages/fluentui/react/src/themes/teams/components/Table/tableCellStyles.ts
```diff
@@ -25,7 +25,7 @@ export default {
       height: '100%'
     };
   },
-  content: ({ props: { truncateContent } }: ComponentStyleFunctionParam<TableCellProps, TeamsTableVariables>): ICSSInJSStyle => {
+  content: ({ props: { truncateContent } }): ICSSInJSStyle => {
```
layershifter:
```suggestion
  content: ({ props: p }): ICSSInJSStyle => {
```

packages/fluentui/react/src/themes/teams/components/Table/tableCellStyles.ts
```diff
@@ -25,7 +25,7 @@ export default {
...
     return {
       alignSelf: 'center',
       ...(truncateContent && {
```
layershifter:
```suggestion
      ...(p.truncateContent && {
```

PR #12153, opened 2020-03-02
ColorPicker: Add support for selecting transparency instead of alpha

ecraig12345:
#### Pull request checklist

#### Description of changes

Add support for selecting transparency values instead of alpha values in the ColorPicker. This was started by @navkuma in #11836, but after thinking a good bit about what the best approach would be, I decided to make a separate PR implementing what I was thinking of.

The feature is enabled by setting a new prop `alphaType: 'alpha' | 'transparency' | 'none'`. This prop supersedes `alphaSliderHidden` and provides a single place to control if/how alpha/transparency controls are displayed. (Also considered `useTransparency` or `useTransparencySlider`, but `alphaType` has the advantage of putting all control in one place and mirroring the new `IColorSliderProps.type`. Of the other two options, I prefer `useTransparency` because the setting affects a text field, not just the slider.)

This is what it looks like in transparency mode:
image

vs. alpha mode:
image

Transparency values are stored in a new `IColor` property `t` (for clarity to the consumer). I updated the color utility functions ensure this mirrors the alpha value `IColor.a`.

For `ColorSlider`, to handle transparency sliders, I deprecated the `isAlpha` prop and added a new prop `type: 'hue' | 'alpha' | 'transparency'` (and deprecated `minValue` and `maxValue`, which are inferred from `type`). This is cleaner than having multiple boolean props for slider types.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12153)

### DIFF
--- a/change/office-ui-fabric-react-2020-04-06-09-05-28-transparency2.json
+++ b/change/office-ui-fabric-react-2020-04-06-09-05-28-transparency2.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "ColorPicker: Add support for selecting transparency instead of alpha",
+  "packageName": "office-ui-fabric-react",
+  "email": "elcraig@microsoft.com",
+  "commit": "8daab0296711d7edf6060e0f43d2794653db4387",
+  "dependentChangeType": "patch",
+  "date": "2020-04-06T16:05:28.690Z"
+}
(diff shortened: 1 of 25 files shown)


kenotron:
What I don't get is why there is a need to represent in the underlying state with another T value? I could be wrong in what you're trying to do but why not just have this be internally represented as alpha and then transparency is but a derived state?

aneeshack4:
@ecraig12345 any update on this?

PR #12159, opened 2020-03-03
chore: create a separate "lerna.json" for FluentUI v0

layershifter:
#### Pull request checklist

#### Description of changes

This PR creates a separate `lerna.json` to remove existing hacks with the `private` field in `fluentui-publish` task:
- new `lerna.json` is under `packages/fluentui` directory
- new `@fluentui/noop` is required as `@lerna/project` requires to have a `package.json` under the same directory:
https://github.com/lerna/lerna/blob/6c4ee5294002c0e464f557b149516017eccf2a07/core/project/index.js#L101

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12159)


lerna.json
```diff
@@ -2,5 +2,5 @@
   "packages": ["apps/*", "packages/*", "scripts", "packages/fluentui/*"],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "0.45.0"
+  "version": "0.0.0"
```
layershifter: A value there was reverted to the original one as `version` field is required in `lerna.json`

packages/fluentui/lerna.json
```diff
@@ -0,0 +1,5 @@
+{
+  "packages": ["**/*"],
```
layershifter: This glob with match only packages under `packages/fluentui` 👍

packages/fluentui/lerna.json
```diff
@@ -0,0 +1,5 @@
...
+  "npmClient": "yarn",
+  "version": "0.45.0"
```
layershifter: We cannot use `useWorkspaces` there as Lerna has a validation for it: https://github.com/lerna/lerna/blob/6c4ee5294002c0e464f557b149516017eccf2a07/core/project/index.js#L74

However, it should not be an issue with publishing as Yarn Workspaces are not used there.

packages/fluentui/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "@fluentui/noop",
+  "private": true,
+  "version": "0.0.0"
```
layershifter: https://github.com/lerna/lerna/blob/6c4ee5294002c0e464f557b149516017eccf2a07/core/project/index.js#L101

Is required as we should have `package.json` under the same directory

scripts/tasks/fluentui-publish.js
```diff
@@ -1,91 +1,30 @@
...
-module.exports.fluentuiPostpublish = function() {
-  const root = findGitRoot();
-  setPrivateFlagOnFluentuiPackages(root, true);
-  setLernaPackages(root, ['apps/*', 'packages/*', 'scripts', 'packages/fluentui/*']);
-  spawnSync('git', ['add', 'packages/fluentui', 'lerna.json'], { cwd: root });
-  spawnSync('git', ['commit', '-m', 'bumping @fluentui packages'], { cwd: root });
+module.exports.fluentuiPrepublish = function() {};
 
-  gitTag(root);
-};
```
layershifter: `pre` and `post` stages are not required anymore

scripts/tasks/fluentui-publish.js
```diff
@@ -1,91 +1,30 @@
...
-    const root = findGitRoot();
-
+    const fluentRoot = path.resolve(findGitRoot(), 'packages', 'fluentui');
     const lernaPublishArgs = [
       'lerna',
       'publish',
       "--tag-version-prefix='fluentuizero_v'",
       '--no-git-reset',
-      '--no-push',
-      '--no-git-tag-version',
```
layershifter: Lerna can do both things:
- `push` will be done to a proper branch, it's important if we want to release from branches
- Lerna will also tag properly so we don't to run this in `post`

scripts/tasks/fluentui-publish.js
```diff
@@ -1,91 +1,30 @@
...
       argv().registry,
       bumpType
     ];
 
-    logger.info(`Running this command: npx ${lernaPublishArgs.join(' ')}`);
+    logger.info(`Running this command: yarn ${lernaPublishArgs.join(' ')}`);
 
-    const result = spawnSync('npx', lernaPublishArgs, {
-      cwd: root,
+    const result = spawnSync('yarn', lernaPublishArgs, {
```
layershifter: I don't know why `npx` should be used as we use `yarn` 🤔

scripts/tasks/fluentui-publish.js
```diff
@@ -1,91 +1,30 @@
...
-};
+const findGitRoot = require('../monorepo/findGitRoot');
 
-module.exports.fluentuiPostpublish = function() {
-  const root = findGitRoot();
-  setPrivateFlagOnFluentuiPackages(root, true);
-  setLernaPackages(root, ['apps/*', 'packages/*', 'scripts', 'packages/fluentui/*']);
-  spawnSync('git', ['add', 'packages/fluentui', 'lerna.json'], { cwd: root });
-  spawnSync('git', ['commit', '-m', 'bumping @fluentui packages'], { cwd: root });
+module.exports.fluentuiPrepublish = function() {};
```
ecraig12345: Should we just delete these since they're not used now? (should only have ever been used within the repo, so it should be easy to remove any usage)

scripts/tasks/fluentui-publish.js
ecraig12345: Does the yarn version still work if you don't have lerna globally installed? I'm guessing that was the reason for using npx.

scripts/tasks/fluentui-publish.js
layershifter: Yep, I think that we can remove them 👍

scripts/tasks/fluentui-publish.js
layershifter: https://classic.yarnpkg.com/en/docs/cli/#locally-installed-clis

Oops, however I see what you mean. Are there any reasons why `lerna` is not a dependency under `fabricui/scripts`?

ecraig12345: Oh cool, didn't know that. I'm guessing lerna isn't a dep there because it's only used at the root, but @kenotron would know better.

kenotron: I just had been using npx locally that knew it worked. But if yarn works, I'm fine with that change.

kenotron:
Right, so the issue here is beachball check is getting in your way. We need to have a scoped check now to help move you forward!

kenotron:
@xugao and I are discussing ways to do beachball scoped publish as well here: https://github.com/microsoft/beachball/issues/265

We can quickly do a scoped checker to unblock you much faster than having the publish story done.

kenotron:
There's now a PR from @xugao in beachball that would allow for scoped publishes. This will be needed to help address this for us: https://github.com/microsoft/beachball/pull/267

kenotron:
The feature for beachball scoping has been released. I have separate PR #12175 here that upgrades it and convert Fabric's packages to use the scopes so that even this PR flipping things to public in FUI would not affect the beachball publishes.

As a followup after *this* PR, we can _technically_ begin using beachball for an automated workflow for FluentUI packages. However, beachball currently doesn't generate as pretty-looking release notes / changelog.md as FUI's hand maintained one does. It is likely one of the main features we need nail before moving to beachball there.

scripts/tasks/fluentui-publish.js
layershifter: @ecraig12345 Oh, I see. `lerna` is under root `package.json`. It's also fine, `yarn` will work in this case 👍

scripts/tasks/fluentui-publish.js
layershifter: Removed 👍

layershifter:
@ecraig12345 @kenotron I cleaned up the branch. I think that it's fine to go until we will decide something on `beachball`.

PR #12160, opened 2020-03-03
feat(AccordionTitle): remove hard icon dependencies

mnajdova:
#### Pull request checklist

#### Description of changes
This PR removes the internal usage of the Icon component inside the AccordionTitle for the active indicator. The icon that was previously rendered as submenu indicator, is now added as part of the styles in the theme. See more details regarding this effort on https://github.com/OfficeDev/office-ui-fabric-react/pull/12145

### BREAKING CHANGE
The indicator prop in the `AccordionTitle` component is now `ShorthandValue<BoxProps>` instead of `ShorthandValue<IconProps>`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12160)


packages/fluentui/react/src/themes/teams/components/Accordion/accordionTitleStyles.ts
```diff
@@ -1,7 +1,10 @@
 import { ComponentSlotStylesPrepared } from '@fluentui/styles';
 import { AccordionTitleProps } from '../../../../components/Accordion/AccordionTitle';
+import { AccordionVariables } from './accordionVariables';
+import activeIndicatorUrl from './activeIndicatorUrl';
+import { pxToRem } from 'src/utils';
```
layershifter: Oops

packages/fluentui/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -123,13 +120,14 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
           {...accessibility.attributes.content}
           {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.content, unhandledProps)}
         >
-          {Icon.create(indicatorWithDefaults, {
+          {Box.create(indicator === undefined ? {} : indicator, {
```
layershifter:
```suggestion
          {Box.create(indicator, {
```

And then:

```
AccordionTitle.defaultProps = {
  indicator: {},
}
```

***

We will have a `propTypes` warning, so:

```diff
-indicator: customPropTypes.shorthand,
+indicator: customPropTypes.shorthandAllowingChildren,
```

PR #12169, opened 2020-03-03
Don't use gulp-cache in CI for now

ecraig12345:
gulp-cache is sometimes restoring files to the wrong place and breaking CI builds. Turn off caching in CI until we figure out what's wrong.

This will slow down builds, but at least they'll succeed!  👍 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12169)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - template: azure-pipelines.tools.yml
 
-      - template: azure-pipelines.cache.yml
+      # - template: azure-pipelines.cache.yml
 
       - script: |
           yarn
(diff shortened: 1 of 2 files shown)


JasonGore:
We should put this on hold to try adding `base` [here](https://github.com/OfficeDev/office-ui-fabric-react/blob/45f3583dcc35dad6c5a2c8aa339bcc1b904c2809/scripts/gulp/plugins/gulp-react-docgen.ts#L27) as Shift mentioned.

ecraig12345:
@kenotron has probably fixed this properly, so closing this PR.

PR #12170, opened 2020-03-03
fixes the docgen plugin to inject base and cwd for caching purposes

kenotron:
#### Pull request checklist

#### Description of changes

@layershifter pointed out that we're not currently using the fixed cwd / base stuff in the docgen plugin. This will address the fact that some of the gulp-cache files are randomly injected with incorrect base

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12170)

### DIFF
--- a/azure-pipelines.cache.yml
+++ b/azure-pipelines.cache.yml
@@ -2,9 +2,9 @@
 steps:
   - task: Cache@2
     inputs:
-      key: 'gulp-cache | "$(Agent.OS)"'
+      key: 'gulp-cache-v001 | "$(Agent.OS)"'
       restoreKeys: |
-        gulp-cache | "$(Agent.OS)"
-        gulp-cache
+        gulp-cache-v001 | "$(Agent.OS)"
+        gulp-cache-v001
       path: /tmp/gulp-cache
     displayName: Cache gulp-cache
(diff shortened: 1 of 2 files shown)


azure-pipelines.cache.yml
```diff
@@ -2,9 +2,9 @@
 steps:
   - task: Cache@2
     inputs:
-      key: 'gulp-cache | "$(Agent.OS)"'
+      key: 'gulp-cache-v001 | "$(Agent.OS)"'
       restoreKeys: |
-        gulp-cache | "$(Agent.OS)"
-        gulp-cache
+        gulp-cache-v001 | "$(Agent.OS)"
+        gulp-cache-v001
```
JasonGore: Curious why this is needed? Is there now a permanent `gulp-cache` store somewhere?

JasonGore:
It's on my todo list to fix these false regressions from Fluent. There aren't enough iterations in some cases.

azure-pipelines.cache.yml
ecraig12345: The gulp cache in the OS temp directory isn't being cleared between builds.

ecraig12345: And this PR's build was still failing due to what we assume is a previous invalid cache.

JasonGore: Oops, totally missed there were multiple commits.

PR #12174, opened 2020-03-03
Tests: Fix spurious infinite timer issue

JasonGore:
#### Pull request checklist

#### Description of changes

Fix [spurious infinite timer issue on unit tests](https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=55536&view=logs&j=92058b04-f16a-5035-546c-cae3ad5e2f5f&t=17e772b0-f0c3-53ab-7aac-ad6df80e30d7) due to [an issue with jest's `runAllTimers` and debounce implementations](https://github.com/facebook/jest/issues/3465) by adding a mock implementation for `Async.debounce`, similar to [the mock for lodash `debounce`](https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/fluentui/react/test/__mocks__/lodash.ts).

image

I was able to repro the issue by forcibly calling `Async.debounce` in a unit test, like so:

```
+import { Async } from '../../Utilities';
 import { renderToStaticMarkup } from 'react-dom/server';
 
@@ -412,6 +414,13 @@ describe('TextField with error message', () => {
 
     assertErrorMessage(wrapper.getDOMNode(), errorMessage);
 
+    const AsyncInstance = new Async();
+    const DebounceResultFn2 = AsyncInstance.debounce(() => {
+      console.log('debounced');
+    }, 1000);
+
+    DebounceResultFn2();
+
     wrapper.setProps({ value: '' });
     jest.runAllTimers();
```

And confirmed that this mock fixed the issue.

Search terms: Ran 100000 timers, and there are still more! Assuming we've hit an infinite recursion and bailing out...

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12174)

### DIFF
--- a/packages/office-ui-fabric-react/__mocks__/@uifabric/utilities.ts
+++ b/packages/office-ui-fabric-react/__mocks__/@uifabric/utilities.ts
@@ -1,0 +1,33 @@
+export * from '@uifabric/utilities';
+import { Async } from '@uifabric/utilities';
+
+// Known issue with jest's runAllTimers and debounce implementations resulting in
+// "Ran 100000 timers, and there are still more! Assuming we've hit an infinite recursion and bailing out..."
+// https://github.com/facebook/jest/issues/3465
+// Mock impl inspired from issue.
+class MockAsync extends Async {
+  public debounce(callback: Function, timeout: number) {
+    let timeoutId: number | null = null;
+    const debounced = (...args: any[]) => {
+      if (timeoutId) {
+        window.clearTimeout(timeoutId);
+      }
+      // Callback functions throughout repo aren't binding properly, so we have to access
+      // Async's private _parent member and invoke callbacks the same way Async.debounce does.
+      const invokeFunction = () => callback.apply((this as any)._parent, args);
+      timeoutId = window.setTimeout(invokeFunction, timeout);
+    };
+
+    const cancel = () => {
+      if (timeoutId) {
+        window.clearTimeout(timeoutId);
+      }
+    };
+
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/__mocks__/@uifabric/utilities.ts
```diff
@@ -0,0 +1,33 @@
...
+// Known issue with jest's runAllTimers and debounce implementations resulting in
+// "Ran 100000 timers, and there are still more! Assuming we've hit an infinite recursion and bailing out..."
+// https://github.com/facebook/jest/issues/3465
+// Mock impl inspired from issue.
+class MockAsync extends Async {
+  public debounce(callback: Function, timeout: number) {
+    let timeoutId: number | null = null;
+    const debounced = (...args: any[]) => {
+      if (timeoutId) {
+        clearTimeout(timeoutId);
```
ecraig12345: Shouldn't you also do `timeoutId = null` both here and in `cancel`?

JasonGore: Well, it gets set immediately again below, and would have no effect in timeoutId, and it's a mock. But I can add them just to be safe.

ecraig12345: Yeah it's more an issue of general good patterns than an actual problem in this case.

PR #12181, opened 2020-03-04
chore(Toolbar): move more components to functional

layershifter:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `ToolbarDivider`, `ToolbarCustomItem` & `ToolbarItem` components to be functional. As in all other cases it contains a breaking change with restricted props set that will be passed to styles functions.

## Prop sets

| `ToolbarMenu` |
| ------------- |
| no props      |

| `ToolbarMenuDivider` |
| -------------------- |
| no props             |

| `ToolbarMenuRadioGroup` |
| ----------------------- |
| no props                |

| `ToolbarRadioGroupStylesProps` |
| ------------------------------ |
| no props                       |


packages/fluentui/accessibility/src/behaviors/Toolbar/toolbarMenuRadioGroupBehavior.ts
```diff
@@ -6,12 +6,14 @@ import { Accessibility } from '../../types';
  * @specification
  *  Adds role='group'.
  */
-const toolbarRadioGroupBehavior: Accessibility = () => ({
+const toolbarMenuRadioGroupBehavior: Accessibility = () => ({
```
layershifter: Oops, previous name was wrong

packages/fluentui/react/src/components/Toolbar/ToolbarMenu.tsx
```diff
@@ -45,75 +58,73 @@ export interface ToolbarMenuProps extends UIComponentProps, ChildrenComponentPro
...
         ...itemProps,
         menuOpen: !!itemProps.menu
       });
     },
     variables: mergeComponentVariables(variables, predefinedProps.variables)
   });
 
-  handleDividerOverrides = variables => predefinedProps => ({
+  const handleDividerOverrides = predefinedProps => ({
     variables: mergeComponentVariables(variables, predefinedProps.variables)
```
layershifter: A devil is here 😈 Previously we merged resolved variables while now we merge input, the simplified change:

```diff
-renderComponent({ variables }) {
+  const { variables } = this.props
}
```

But as I discussed this with @miroslavstastny, it's valid. As we want to pass down only overrides.

PR #12182, opened 2020-03-04
fix(Tree): use context to supply generated props to items

silviuavram:
#### Pull request checklist

#### Description of changes

- Passing onFocusParent, onTitleClick, onFocusFirstChild and onSiblingsExpand through context from Tree to TreeItem. Removed them from overrideProps.
- Creating contentRef only once.
- Removing itemsForRender and generating the props relevant for accessibility directly in `renderItems`.
- Removed siblings prop and compute it only when needed.

Fixes https://github.com/microsoft/fluent-ui-react/issues/2295.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12182)

### DIFF
--- a/packages/fluentui/react/src/components/Tree/Tree.tsx
+++ b/packages/fluentui/react/src/components/Tree/Tree.tsx
@@ -197,14 +197,6 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
     }
   });
 
-  setActiveItemIds = (e: React.SyntheticEvent, activeItemIds: string[]) => {
-    _.invoke(this.props, 'onActiveItemIdsChange', e, { ...this.props, activeItemIds });
-
-    this.setState({
-      activeItemIds
-    });
-  };
-
   onFocusFirstChild = (itemId: string) => {
     const currentElement = this.itemsRef.get(itemId);
 
@@ -220,6 +212,7 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
 
     elementToBeFocused.focus();
   };
+
   onSiblingsExpand = (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => {
     if (this.props.exclusive) {
       return;
@@ -240,6 +233,7 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
 
     this.setActiveItemIds(e, activeItemIds);
   };
+
   onFocusParent = (parent: string) => {
     const parentRef = this.itemsRef.get(parent);
 
@@ -293,6 +287,14 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
 
     return renderItems(items);
   }
+
+  setActiveItemIds = (e: React.SyntheticEvent, activeItemIds: string[]) => {
+    _.invoke(this.props, 'onActiveItemIdsChange', e, { ...this.props, activeItemIds });
+
+    this.setState({
+      activeItemIds
+    });
+  };
 
   renderComponent({ ElementType, classes, accessibility, unhandledProps }) {
     const { children, renderedItems } = this.props;
--- a/packages/fluentui/react/src/components/Tree/Tree.tsx
+++ b/packages/fluentui/react/src/components/Tree/Tree.tsx
@@ -27,7 +27,7 @@ import {
   ShorthandValue,
   ComponentEventHandler
 } from '../../types';
-import { hasSubtree, removeItemAtIndex } from './utils';
+import { hasSubtree, removeItemAtIndex, TreeItemContext, TreeItemRenderContextValue } from './utils';
 
 export interface TreeSlotClassNames {
   item: string;
@@ -80,7 +80,7 @@ export interface TreeItemForRenderProps {
   id: string;
   index: number;
   level: number;
-  parent: ShorthandValue<TreeItemProps>;
+  parentRef: React.RefObject<HTMLElement>;
   siblings: ShorthandCollection<TreeItemProps>;
 }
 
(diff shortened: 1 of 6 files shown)


packages/fluentui/react/src/components/Tree/Tree.tsx
```diff
@@ -225,101 +186,102 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
-            key: item['id'],
-            contentRef: elementRef,
-            ...restItemForRender
-          }),
-          overrideProps: this.handleTreeItemOverrides
-        });
-
-        return [...renderedItems, renderedItem, ...(isSubtreeExpanded ? renderItems(item['items']) : ([] as any))];
+        const elementRef = React.createRef<HTMLElement>();
+        const contextValue: TreeItemRenderContextValue = {
```
layershifter: This will not work 😨 https://smykhailov.github.io/react-patterns/#/context

So, even if consumers will inject `React.memo()` around `TreeItem` they will still receive updates as context value will always change.

packages/fluentui/react/src/components/Tree/Tree.tsx
```diff
@@ -225,101 +186,102 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
+          level,
+          index: index + 1,
+          size: items.length,
```
layershifter: I don't see any benefits from passing `level`, `index` & `size` via context as they are plain.
***
I proposed to use Context because I want to avoid creation of new functions in `overrideProps`, it's done ✔️ 
Now we need:
- find a way to avoid `elementRef`, `onSiblingsExpand`, `onFocusFirstChild`, `onFocusParent` creation on each render, i.e. remove dependency on exact item
- `TreeItemContext` should `TreeContext`, we obviously don't want to wrap each item with own `Provider`, it will be easy doable after previous item will be solved

packages/fluentui/react/src/components/Tree/Tree.tsx
```diff
@@ -191,135 +159,136 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
+      activeItemIds.push(id);
     }
-  });
+
+    this.setActiveItemIds(e, activeItemIds);
+  };
 
   setActiveItemIds = (e: React.SyntheticEvent, activeItemIds: string[]) => {
-    _.invoke(this.props, 'onActiveItemIds', e, { ...this.props, activeItemIds });
+    _.invoke(this.props, 'onActiveItemIdsChange', e, { ...this.props, activeItemIds });
```
silviuavram: sometimes my coding is simply bad.

packages/fluentui/react/src/components/Tree/Tree.tsx
```diff
@@ -191,135 +159,136 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
             className: Tree.slotClassNames.item,
             expanded: isSubtreeExpanded,
             renderItemTitle,
             key: item['id'],
-            contentRef: elementRef,
-            ...restItemForRender
-          }),
-          overrideProps: this.handleTreeItemOverrides
+            parent,
+            level,
```
silviuavram: I want to refactor level, index, treeSize in a subsequent PR. I think it makes sense to just call them ariaPosInSet, ariaSetSize, ariaLevel since that's their only use. Opinions?

packages/fluentui/react/src/components/Tree/utils/index.ts
```diff
@@ -9,3 +10,42 @@ export const hasSubtree = (item: TreeItemProps | ShorthandValue<TreeItemProps>):
...
+};
+
+export interface TreeItemRenderContextValue {
+  onFocusFirstChild: (itemId: string) => void;
+  onFocusParent: (itemId: string) => void;
+  onSiblingsExpand: (e: React.SyntheticEvent, itemProps: TreeItemProps) => void;
+  onTitleClick: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => void;
+}
+
+export const TreeItemContext = React.createContext<TreeItemRenderContextValue>(null);
```
layershifter:
```suggestion
export const TreeContext = React.createContext<TreeContextValue>(null);
```

I would say that this is `TreeContext`

packages/fluentui/react/src/components/Tree/Tree.tsx
```diff
@@ -191,135 +159,136 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
 
-      if (!currentElement || !currentElement.current) {
-        return;
-      }
+    if (!currentElement || !currentElement.current) {
+      return;
+    }
 
-      const elementToBeFocused = getNextElement(this.treeRef.current, currentElement.current);
+    const elementToBeFocused = getNextElement(this.treeRef.current, currentElement.current);
```
layershifter: We know that `this.treeRef` is a ref object, so it's stable, correct? 🤔 

Can we just pass it via Context and use it inside `TreeItem`?

`TreeItem.tsx`

```diff
-  onFocusFirstChild = (itemId: string) => {
-    const currentElement = this.itemsRef.get(itemId);
-
-    if (!currentElement || !currentElement.current) {
-      return;
-    }
-
-    const elementToBeFocused = getNextElement(this.treeRef.current, currentElement.current);
-
-    if (!elementToBeFocused) {
-      return;
-    }
-
-    elementToBeFocused.focus();
-  };
```

`TreeItem.tsx`

```diff
- const { onFocusParent, onSiblingsExpand, onFocusFirstChild, onTitleClick } = React.useContext(TreeItemContext);
+ const { onFocusParent, onSiblingsExpand, onTitleClick, treeRef } = React.useContext(TreeItemContext);

//
  const handleFocusFirstChild = e => {
    _.invoke(props, 'onFocusFirstChild', e, props);
-    onFocusFirstChild(props.id);
+   const elementToBeFocused = getNextElement(treeRef.current, contentRef.current);
+
+  if (elementToBeFocused) elementToBeFocused.focus()
  };
```

silviuavram: I don't see much improvement here, we are just swapping one callback with a ref. I would rather go with the callback approach since it looks more consistent, we have all 4 callbacks passed as context.

packages/fluentui/react/src/components/Tree/utils/index.ts
silviuavram: Cool!

packages/fluentui/react/src/components/Tree/Tree.tsx
layershifter: `aria-level` in this case. But makes sense only if they are used only for attributes...

packages/fluentui/react/src/components/Tree/utils/index.ts
```diff
@@ -9,3 +10,42 @@ export const hasSubtree = (item: TreeItemProps | ShorthandValue<TreeItemProps>):
...
+  }
+
+  return getSiblingsFn(items);
+};
+
+export interface TreeRenderContextValue {
+  onFocusFirstChild: (itemId: string) => void;
+  onFocusParent: (itemId: string) => void;
+  onSiblingsExpand: (e: React.SyntheticEvent, itemProps: TreeItemProps) => void;
+  onTitleClick: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => void;
```
layershifter:
```suggestion
  onTitleClick: (e: React.SyntheticEvent, itemProps: TreeItemProps) => void;
```

_nit_

PR #12183, opened 2020-03-04
chore(Input | Checkbox): remove hard icon dependencies

mnajdova:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12183)


packages/fluentui/react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -156,10 +156,9 @@ const Checkbox: React.FC<WithAsProp<CheckboxProps>> &
       {Icon.create(icon, {
         defaultProps: () =>
           getA11Props('icon', {
```
pompomon:
```suggestion
          getA11yProps('icon', {
```

packages/fluentui/react/src/components/Input/Input.tsx
```diff
@@ -202,7 +202,7 @@ class Input extends AutoControlledComponent<WithAsProp<InputProps>, InputState>
     const { value } = this.state;
 
     if (clearable && (value as string).length !== 0) {
```
layershifter: Now it should be something like:

```
if (clearable && (value as string).length !== 0) {
  return <Box styles />
}

return Icon.create(icon)
```

packages/fluentui/react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -156,10 +156,9 @@ const Checkbox: React.FC<WithAsProp<CheckboxProps>> &
       {Icon.create(icon, {
```
layershifter: Box?

packages/fluentui/react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -156,10 +156,9 @@ const Checkbox: React.FC<WithAsProp<CheckboxProps>> &
...
         defaultProps: () =>
           getA11Props('icon', {
-            outline: toggle && !state.checked,
             size: toggle ? 'medium' : 'smaller',
             className: Checkbox.slotClassNames.indicator,
-            name: toggle ? 'icon-circle' : 'icon-checkmark',
+            name: '',
```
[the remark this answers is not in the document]
mnajdova: name is required prop

packages/fluentui/react/src/components/Checkbox/Checkbox.tsx
mnajdova: Not sure if Box is valid here, as from client's perspective I think icon is the valid slot

packages/fluentui/react/src/components/Input/Input.tsx
mnajdova: We cannot have children in the Icon component

packages/fluentui/react/src/components/Checkbox/Checkbox.tsx
layershifter: I don't see any reasons to be honest 🤔  You will not use `book` for as checkmark

packages/fluentui/react/src/components/Input/Input.tsx
layershifter: I see what you mean, but it's really weird

packages/fluentui/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
```diff
@@ -80,53 +116,61 @@ const checkboxStyles: ComponentSlotStylesPrepared<CheckboxStylesProps, CheckboxV
...
     ...(p.disabled && {
-      color: v.disabledToggleIndicatorColor,
-      background: v.disabledBackground,
-      borderColor: v.disabledBorderColor
-    }),
-
-    ...(p.disabled &&
-      p.checked && {
-        color: v.disabledCheckedIndicatorColor,
+      ...(!p.checked && {
```
layershifter: I don't understand this... Why `!p.checked` is required there? Should it be just:

```
...(p.disabled && {
  background: v.disabledBackground,
  borderColor: v.disabledBorderColor
  ...(p.checked && {
    background: v.disabledBackgroundChecked,
    borderColor: 'transparent',
  }),
}
```

packages/fluentui/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
```diff
@@ -80,53 +116,61 @@ const checkboxStyles: ComponentSlotStylesPrepared<CheckboxStylesProps, CheckboxV
...
     },
 
-    ...(p.checked && {
-      background: v.checkedBackground,
-      borderColor: v.checkedBorderColor,
-      color: v.checkedIndicatorColor,
-      padding: v.toggleCheckedPadding
-    }),
+    ...(p.checked &&
+      !p.disabled && {
```
layershifter: https://github.com/OfficeDev/office-ui-fabric-react/pull/12183/files#r388786483

Same thing about `!p.disabled`... We have following states:
- default
- `checked`
- `disabled`
- `checked` & `disabled`

Why we need negative conditions for other states here?

packages/fluentui/react/src/themes/teams/components/Input/inputStyles.ts
```diff
@@ -49,18 +51,31 @@ const inputStyles: ComponentSlotStylesPrepared<InputProps, InputVariables> = {
     })
   }),
 
-  icon: ({ props: p, variables: v }): ICSSInJSStyle => ({
-    color: v.iconColor,
-    outline: 0,
-    position: v.iconPosition as PositionProperty,
+  icon: ({ props: p, variables: v }): ICSSInJSStyle => {
+    const hasValue = !!p.value && (p.value as string).length !== 0;
```
layershifter: Can we move thing to `getAutoControlledStateFromProps()` for now? And then we can just reuse it in `useStyles`

PR #12185, opened 2020-03-04
fix(useAccessibility): make callbacks referentially stable

layershifter:
#### Pull request checklist

#### Description of changes

This PR makes returned `onKeyDown` by `getA11yProps` from `useAccessibility()` hook stable between rerenders, so it will stop to break memoized component (`PureComponent`, `React.memo()`) which can be used via `as` prop or as `trigger` in `Popup` & `Tooltip`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12185)


packages/fluentui/react-bindings/src/hooks/useAccessibility.ts
```diff
@@ -19,44 +19,41 @@ type UseAccessibilityResult = (<SlotProps extends Record<string, any>>(
...
+  const definition = getAccessibility(debugName, behavior, mapPropsToBehavior(), rtl, actionHandlers);
 
-    finalProps.onKeyDown = onKeyDown;
-  }
+  const slotHandlers = React.useRef<Record<string, KeyboardEventHandler>>({});
+  const slotProps = React.useRef<Record<string, UserProps>>({});
 
-  return finalProps;
-};
+  const getA11Props: UseAccessibilityResult = (slotName, userProps) => {
```
pompomon:
```suggestion
  const getA11yProps: UseAccessibilityResult = (slotName, userProps) => {
```

packages/fluentui/react-bindings/src/hooks/useAccessibility.ts
```diff
@@ -19,44 +19,41 @@ type UseAccessibilityResult = (<SlotProps extends Record<string, any>>(
...
+      ...definition.attributes[slotName],
+      ...userProps,
+      onKeyDown: slotHandlers.current[slotName]
+    };
+
+    return finalProps;
+  };
 
   // Provides an experimental handling for FocusZone definition in behaviors
   getA11Props.unstable_wrapWithFocusZone = (element: React.ReactElement) => {
```
pompomon:
```suggestion
  getA11yProps.unstable_wrapWithFocusZone = (element: React.ReactElement) => {
```

packages/fluentui/react-bindings/src/hooks/useAccessibility.ts
```diff
@@ -19,44 +19,41 @@ type UseAccessibilityResult = (<SlotProps extends Record<string, any>>(
...
-};
+  const getA11Props: UseAccessibilityResult = (slotName, userProps) => {
+    slotProps.current[slotName] = userProps;
 
-const useAccessibility = <Props>(behavior: Accessibility<Props>, options: UseAccessibilityOptions<Props> = {}) => {
-  const { actionHandlers, debugName = 'Undefined', mapPropsToBehavior = () => ({}), rtl = false } = options;
-  const definition = getAccessibility(debugName, behavior, mapPropsToBehavior(), rtl, actionHandlers);
+    if (!slotHandlers.current[slotName]) {
+      slotHandlers.current[slotName] = (e, ...args) => {
+        const accessibilityHandler = definition.keyHandlers[slotName].onKeyDown;
```
miroslavstastny: `keyHandlers` for a slot can be undefined. That was previously checked on line 38 but you don't have the check in the new code.

Also the previous contract was that the onKeyDown listener was added to an element only if it was handling anything. And that could change based on `props` change. Now the listener is added and never removed.
I think it would be better to be unstable in these edge cases and add/remove the listener when needed.

You can use `ButtonBehavior` to test both issues.
- It does not add the listener by default -> first issue.
- If you rerender with `as='div'`, it is supposed to add the listener, if you remove the`as` later, it should remove the listener -> second issue.

packages/fluentui/react-bindings/src/accessibility/types.ts
```diff
@@ -11,7 +11,7 @@ export interface ReactAccessibilityBehavior extends AccessibilityDefinition {
 }
 
 export type AccessibilityKeyHandlers = {
-  [slotName: string]: AccessibilityHandlerProps;
+  [slotName: string]: AccessibilityHandlerProps | undefined;
```
layershifter: It can be `undefined`, yep

packages/fluentui/react-bindings/src/hooks/useIsomorphicLayoutEffect.ts
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 
 // useLayoutEffect() produces a warning with SSR rendering
 // https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a
-const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect;
+const useIsomorphicLayoutEffect =
+  typeof window !== 'undefined' && process.env.NODE_ENV !== 'test' ? React.useLayoutEffect : React.useEffect;
```
layershifter: Without `process.env.NODE_ENV !== 'test'` it produces warning for tests because we use JSDom which has `window` for sure

PR #12192, opened 2020-03-04
chore(FocusZone): Using the same DOM utilities in v0 that we use in v7

khmakoto:
#### Pull request checklist

#### Description of changes

This PR is part of a series of changes to bring the `FocusZone` in v0 and v7 closer together. This PR specifically deals with replacing v0's DOM utilities used in the `FocusZone` component with those that are used in v7. That means, for the moment, using utilities in the `uifabric/utilities` package. The location of these utilities will change in the future to a more appropriate place after the renaming has taken place, however, it is good to start using the same set of utilities so that things move more smoothly once the utilities breakup work starts.

In order to use `uifabric/utilities` inside v0's `FocusZone` we had to make some changes inside `uifabric/utilities` and `uifabric/merge-styles`. These changes are also included as part of this PR.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12192)


packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -128,7 +129,7 @@ export default class FocusZone extends React.Component<FocusZoneProps> implement
 
     // @ts-ignore
     this.windowElement = getWindow(this._root.current);
-    let parentElement = getParent(this._root.current);
+    let parentElement = getParent(this._root.current, ALLOW_VIRTUAL_ELEMENTS);
```
dzearing: any reason for this?

packages/merge-styles/src/index.ts
```diff
@@ -18,7 +18,7 @@ export { fontFace } from './fontFace';
 
 export { keyframes } from './keyframes';
 
-export { IStyleSheetConfig, ICSPSettings, InjectionMode, Stylesheet } from './Stylesheet';
+export * from './Stylesheet';
```
dzearing: any reason for this?

packages/utilities/src/object.ts
```diff
@@ -93,6 +93,7 @@ export function mapEnumByName<T>(
         // if the property is not just a number (because enums in TypeScript will map both ways)
         return callback(p as string, theEnum[p]);
       }
+      return undefined;
```
dzearing: is this needed? returning undefined is the same as not returning anything.

packages/merge-styles/src/styleToClassName.ts
```diff
@@ -235,6 +235,8 @@ export function styleToRegistration(options: IStyleOptions, ...args: IStyle[]):
 
     return registration as IRegistration;
   }
+
+  return undefined;
```
dzearing: needed?

packages/fluentui/docs/tsconfig.json
```diff
@@ -4,6 +4,9 @@
     "module": "esnext",
     "paths": {
       "@fluentui/*": ["packages/fluentui/*/src/index"],
+      "@uifabric/merge-styles": ["packages/merge-styles/src"],
```
dzearing: is merge-styles needed here?

packages/fluentui/react-bindings/src/FocusZone/focusUtilities.ts
```diff
@@ -412,6 +414,8 @@ export function getWindow(rootElement?: Element | null): Window | undefined {
  * Helper to get the document object.
  *
  * @public
+ *
+ * @deprecated Use getDocument from @uifabric/utilities instead
  */
 export function getDocument(rootElement?: Element | null): Document | undefined {
```
layershifter: If signatures of these functions are compatible I suggest to reexport them instead.

packages/utilities/src/DelayedRender.tsx
```diff
@@ -49,11 +49,13 @@ export class DelayedRender extends React.Component<IDelayedRenderProps, IDelayed
...
     let { delay } = this.props;
-    this._timeoutId = setTimeout(() => {
+    this._timeoutId = (setTimeout(() => {
       this.setState({
         isRendered: true
       });
-    }, delay);
+      // This cast has to be done so that Fluent can take @uifabric/utilities as a dependency given diferent transpilations between TS to
+      // ES5 and babel to ES6-like.
+    }, delay) as unknown) as number;
```
layershifter: I don't see how it's related to Babel as you're fixing types there as I believe you have there this issue:
https://stackoverflow.com/questions/45802988/typescript-use-correct-version-of-settimeout-node-vs-window

Can you please clarify?

packages/utilities/src/object.ts
JasonGore: Returning nothing has `void` return type. Returning nothing will generate a TS error if `void` is not declared as the function's return type. It forces you to explicitly return something.

packages/merge-styles/src/styleToClassName.ts
JasonGore: I put a comment in the wrong spot earlier. I'm surprised this isn't generating a TS as-is (before this PR's changes.) Returning nothing should be the same as `void` and should generate a TS error if the function's return type does not include `void`.

JasonGore: [Here](https://www.typescriptlang.org/play/?noImplicitAny=false#code/PTAEBUAsEsGdQPYDsCmoUCcMI-ARigMYCGArrGtAC6gC2cF8xS6AHgA4A20h1oGKKqQxIAdAFgAUIWSwaxACYKAjKAC8oABSsAXKCSlaBDABpQATz0GjmAJRXDx0AB9QpJApQAzaKgXqAPlAAbylQUGgvLVZ1DQAGWxCw8P5BYRYYgGoLAG5kgF8pfLzJKRAIGHhkNAUEFFgkAHIaTGwMUAIScjRGgDcEaAVG0GJOTgQAd3gkBDYuHj4BIRFYCWlZeSUAFnVohxtTC33je31HTBdQfsHApMlwyOjY0AS7lNTljNBs8xLwwskxSkQA) is a simplified example.

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
khmakoto: v7's and v0's implementation of `getParent` is the same except that v0 doesn't allow virtual elements. v7's `getParent` has the second parameter, which allows virtual elements, set to true by default. In order to preserve v0's functionality as is we need to pass `false` to the second parameter, which is the value that is set to `ALLOW_VIRTUAL_ELEMENTS`.

packages/merge-styles/src/index.ts
khmakoto: When importing v7's utilities into v0, when building `@fluentui/docs` everything was being parsed. v0 uses the babel compiler, while v7 uses the typescript compiler. v0's parsing was complaining because `ICSPSettings` and `IStyleSheetConfig` are interfaces that it was throwing out and not finding when exported here. Exporting as `*` instead of individually fixes this issue.

There were some concerns about doing this and there is a 100+ long thread offline under the `Engineering Systems` teams channel. The concerns were mostly related to webpack incorrectly tree shaking `export *` when it's two levels deep. However, I checked that this wasn't happening by creating a new react app and consuming both the regular version of `office-ui-fabric-react` and a tarball with the changes without much change in the file size at all.

Regular without changes:

```
File sizes after gzip:
  86.85 KB  build\static\js\2.a2312137.chunk.js
  772 B     build\static\js\runtime-main.a43b6eb2.js
  556 B     build\static\css\main.d1b05096.chunk.css
  541 B     build\static\js\main.87aabca2.chunk.js
```

Tarball with changes:

```
  86.87 KB (+15 B)  build\static\js\2.5b88c310.chunk.js
  772 B             build\static\js\runtime-main.a43b6eb2.js
  556 B             build\static\css\main.d1b05096.chunk.css
  541 B             build\static\js\main.87aabca2.chunk.js
```

Given this I think it would be safe to include this change.

packages/utilities/src/object.ts
khmakoto: Similar thing here, v0's parsing was complaining about not all paths returning something, adding `return undefined` here allows us to mitigate this warning while maintaining the logic.

packages/merge-styles/src/styleToClassName.ts
khmakoto: Same as above, v0's parsing was complaining about not all paths returning something, adding return undefined here allows us to mitigate this warning while maintaining the logic.

packages/fluentui/docs/tsconfig.json
khmakoto: `utilities` has a dependency on `merge-styles`. When v0 is parsing `utilities` we have to include the path map to `merge-styles` as well, otherwise it will complain about `merge-styles` module not being found.

packages/utilities/src/DelayedRender.tsx
khmakoto: I am having that problem, I thought it was babel enforcing this difference but I might have been wrong. I'll use the fix in the link you sent me and delete the comment, thanks!

packages/fluentui/react-bindings/src/FocusZone/focusUtilities.ts
khmakoto: They are not entirely the same in all cases, but I think they are always compatible. They are also only used in `FocusZone` as far as I can tell so I __think__ it might be ok to do that.

Is my understanding correct in saying you're suggesting to delete these function implementations and just re-exporting them?

layershifter: > Is my understanding correct in saying you're suggesting to delete these function implementations and just re-exporting them?

Exactly 👍

PR #12203, opened 2020-03-05
chore(Toolbar): use Context to pass variables

layershifter:
#### Pull request checklist

#### Description of changes

# BREAKING CHANGES

If you have `componentVariables` for `Toolbar` component in theme and they are used in `Toolbar` siblings they should be registered in your theme.

### Before

```tsx
const componentVariables = {
  Toolbar: toolbarVariables
};
```

### After

```tsx
const componentVariables = {
  Toolbar: toolbarVariables,
  ToolbarDivider: toolbarVariables,
  ToolbarItem: toolbarVariables
  // etc.
};
```

---

This PR implements React Context to pass `variables` to unblock Children API usages for `Toolbar` and `ToolbarItem`.

I decided to use single `ToolbarVariablesContext` as it works for all components in `Toolbar`'s group. Tests are removed to avoid testing internals, `variables` usages was covered in #12204.


packages/fluentui/docs/src/examples/components/Toolbar/Performance/CustomToolbar.perf.tsx
```diff
@@ -79,40 +112,11 @@ const darkThemeOverrides: ThemeInput = {
...
-      ctItemNotificationSize: pxToRem(8),
-
-      ctItemPrimaryBackground: siteVars.colorScheme.default.background3,
-      ctItemPrimaryBackgroundHover: siteVars.colorScheme.brand.backgroundHover1,
-      ctItemPrimaryColorHover: siteVars.colorScheme.brand.foregroundHover1
-    })
+    Toolbar: toolbarVariables,
+    ToolbarCustomItem: toolbarVariables,
+    ToolbarDivider: toolbarVariables,
+    ToolbarItem: toolbarVariables
```
miroslavstastny: I think this is OK, but I would consider it a breaking change.

miroslavstastny:
> I thought you wanted to replace variables UTs with a Screener test but I don't see the Screener test added.

Never mind, that was added in a separate PR #12204. Is it intentional that it is not visible in docsite?

packages/fluentui/docs/src/examples/components/Toolbar/Performance/CustomToolbar.perf.tsx
layershifter: Makes a sense, how you would describe it? 🤔

layershifter:
> Never mind, that was added in a separate PR #12204. Is it intentional that it is not visible in docsite?

Nope, will make it visible for local doc site

packages/fluentui/docs/src/examples/components/Toolbar/Performance/CustomToolbar.perf.tsx
layershifter: Added an entry and updated PR description, please check

PR #12204, opened 2020-03-05
docs(Toolbar): add visual test for variables nesting

layershifter:
#### Pull request checklist

#### Description of changes

This PR was extracted from #12203 to check that there will be no changes during refactor. Add a visual test for `variables` nesting in `Toolbar`. Also fixes a minor issue in `ToolbarMenuItem` with passing `variables` down.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12204)

### DIFF
--- a/packages/fluentui/react/src/components/Toolbar/ToolbarDivider.tsx
+++ b/packages/fluentui/react/src/components/Toolbar/ToolbarDivider.tsx
@@ -36,7 +36,7 @@ const ToolbarDivider: React.FC<WithAsProp<ToolbarDividerProps>> & FluentComponen
   const ElementType = getElementType(props);
   const unhandledProps = getUnhandledProps(ToolbarDivider.handledProps, props);
 
-  const element = <ElementType {...getA11yProps('root', { ...unhandledProps, className: classes.root, })} />;
+  const element = <ElementType {...getA11yProps('root', { ...unhandledProps, className: classes.root })} />;
   setEnd();
 
   return element;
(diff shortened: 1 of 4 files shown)


packages/fluentui/react/src/components/Toolbar/ToolbarDivider.tsx
```diff
@@ -36,7 +36,7 @@ const ToolbarDivider: React.FC<WithAsProp<ToolbarDividerProps>> & FluentComponen
   const ElementType = getElementType(props);
   const unhandledProps = getUnhandledProps(ToolbarDivider.handledProps, props);
 
-  const element = <ElementType {...getA11yProps('root', { ...unhandledProps, className: classes.root, })} />;
```
layershifter: ¯\_(ツ)_/¯

packages/fluentui/react/src/themes/teams/components/Toolbar/toolbarRadioGroupStyles.ts
```diff
@@ -4,6 +4,7 @@ import { ToolbarVariables } from './toolbarVariables';
 
 const toolbarRadioGroupStyles: ComponentSlotStylesPrepared<ToolbarRadioGroupStylesProps, ToolbarVariables> = {
   root: (): ICSSInJSStyle => ({
+    display: 'flex',
```
layershifter: It's a required otherwise `ToolbarDivider` will be broken inside `ToolbarRadioGroup`

PR #12207, opened 2020-03-05
Fix the legacy items wrapper styles.

lijunle:
#### Pull request checklist

#### Description of changes

The CSS-in-JS styles and legacy styles for base picker is not compatible. There is a small differences and it will cause the picker item exceed the picker area.

image

#### Focus areas to test

Base picker, people picker.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12207)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.scss
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.scss
@@ -41,6 +41,7 @@
 .pickerItems {
   display: flex;
   flex-wrap: wrap;
+  max-width: 100%;
 }
 
 .screenReaderOnly {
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/pickers/BasePicker.scss
```diff
@@ -41,6 +41,7 @@
 .pickerItems {
   display: flex;
   flex-wrap: wrap;
+  max-width: 100%;
```
lijunle: The corresponding CSS-in-JS styles is here:

> https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/pickers/BasePicker.styles.ts#L80

lijunle:
@dzearing Could you please help on taking a look on this one line change? Thanks!

lijunle:
@joschect Thank you for helping me reviews! Could you please merge the PR?

PR #12209, opened 2020-03-05
Pass contextual menu default item renderer to onRenderMenuList

CampbellOwen:
#### Pull request checklist

#### Description of changes

When overriding the ContextualMenu's default onRenderMenuList, you lose access to the default ContextualMenuItem renderer.

This change adds a defaultMenuItemRenderer field to IContextualMenuListProps, allowing consumers to override onRenderMenuList and still use the default ContextualMenuItem renderer.

The motivation for this change was to add virtualization support to ContextualMenu by using a <List> component in onRenderMenuList while still using the default item renderer.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12209)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -352,7 +352,10 @@ export class ContextualMenuBase extends BaseComponent<IContextualMenuProps, ICon
                     items,
                     totalItemCount,
                     hasCheckmarks,
-                    hasIcons
+                    hasIcons,
+                    defaultMenuItemRenderer: (items, index, focusableElementIndex, totalItemCount, hasCheckmarks, hasIcons) => {
+                      return this._renderMenuItem(items, index, focusableElementIndex, totalItemCount, hasCheckmarks, hasIcons);
+                    }
                   },
                   this._onRenderMenuList
                 )}
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -5,7 +5,8 @@ import {
   ContextualMenuItemType,
   IContextualMenuListProps,
   IContextualMenuStyleProps,
-  IContextualMenuStyles
+  IContextualMenuStyles,
+  IInternalContextualMenuItem
 } from './ContextualMenu.types';
 import { DirectionalHint } from '../../common/DirectionalHint';
 import { FocusZone, FocusZoneDirection, IFocusZoneProps, FocusZoneTabbableElements } from '../../FocusZone';
@@ -353,9 +354,7 @@ export class ContextualMenuBase extends BaseComponent<IContextualMenuProps, ICon
                     totalItemCount,
                     hasCheckmarks,
                     hasIcons,
-                    defaultMenuItemRenderer: (_items, _index, _focusableElementIndex, _totalItemCount, _hasCheckmarks, _hasIcons) => {
-                      return this._renderMenuItem(_items, _index, _focusableElementIndex, _totalItemCount, _hasCheckmarks, _hasIcons);
-                    }
+                    defaultMenuItemRenderer: this._renderMenuItem
                   },
                   this._onRenderMenuList
                 )}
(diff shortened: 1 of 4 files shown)


packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -352,7 +352,10 @@ export class ContextualMenuBase extends BaseComponent<IContextualMenuProps, ICon
                     items,
                     totalItemCount,
                     hasCheckmarks,
-                    hasIcons
+                    hasIcons,
+                    defaultMenuItemRenderer: (_items, _index, _focusableElementIndex, _totalItemCount, _hasCheckmarks, _hasIcons) => {
```
joschect: Inline lambda's can add a perf hit. You should be able to just pass in `this._renderMenuItem` directly to the defaultMenuItemRenderer prop.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.types.ts
```diff
@@ -272,6 +272,14 @@ export interface IContextualMenuListProps {
   totalItemCount: number;
   hasCheckmarks: boolean;
   hasIcons: boolean;
+  defaultMenuItemRenderer: (
+    item: IContextualMenuItem,
```
joschect: Looking at all these arguments in here, I think that I'm in favor of moving this into a single prop object. Maybe extending `IContextualMenuItem`? It could look something like

```
interface IInternalContextualMenuItem extends IContextualMenuItem {
    index: number;
    focusableElementIndex: number;
    totalItemCount: number;
    hasCheckmarks: boolean;
    hasIcons: boolean}
```

Or something similar to that. That might help us keep props and arguments more consistent between the various render methods. Porting that to all the functions could happen later though.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.types.ts
```diff
@@ -272,6 +272,14 @@ export interface IContextualMenuListProps {
...
+    index: number,
+    focusableElementIndex: number,
+    totalItemCount: number,
+    hasCheckmarks: boolean,
+    hasIcons: boolean
+  ) => React.ReactNode;
```
joschect: I'm not 100% on this but I think that this return type should be `React.ReactElement | null` since that's the same type that `React.FunctionComponent` returns

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
CampbellOwen: The default item renderer accesses the props through `this` so we have to capture it here. I think this means we need the lambda?

CampbellOwen: Discussed offline - will change _renderMenuItem to be an arrow function in order to capture this.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.types.ts
CampbellOwen: I think it needs to be a ReactNode based on this: https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement 

Since we're returning an array of ReactNodes, only ReactNode can represent this. When we call this function in ContextualMenu's render function, that will end up being a ReactElement since there's only a single parent.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.types.ts
```diff
@@ -264,6 +264,17 @@ export interface IContextualMenuProps extends IBaseProps<IContextualMenu>, IWith
   delayUpdateFocusOnHover?: boolean;
 }
 
+/**
+ * {@docCategory ContextualMenu}
+ */
+export interface IInternalContextualMenuItem extends IContextualMenuItem {
```
dotnetjunky: It's weird to export a type that says "internal". How about calling it `IContextualMenuItemRenderArgs`?

CampbellOwen: My only concern is the fact that it's just an extension of IContextualMenuItem so it still is a menu item. Adding RenderArgs might make it look more like a props object instead.

Regardless, I like it better than IInternalContextualMenuItem so I'll update this

joschect: I think that IConextualMenuItemRender is definitely better, I might say it should be `IContextualMenuItemRenderProps` to keep things consistent with our naming conventions.

PR #12214, opened 2020-03-06
Lint and prettier updates

ecraig12345:
#### Pull request checklist

#### Description of changes

Make `@uifabric/tslint-rules` extend `tslint-config-prettier` to prevent tslint/prettier conflicts, and remove rules which are redundant with [rules provided by that preset](https://unpkg.com/tslint-config-prettier@1.18.0/lib/index.json). (Same with fluentui tslint config, which already extended `tslint-config-prettier`.)

Remove `typedef` rule from `@uifabric/tslint-rules`--we turn it off in almost all packages, so clearly we don't want it. And remove all the overrides of that rule (as well as `no-inferrable-types`, which we already turn off by default).

Also some prettier-related changes before the giant PR to add trailing commas and change line length:
- Add a preview of the new settings (including override which prevents problematic non-ES5 trailing commas in JS within HTML files)
- Revert a hack I'd previously added to two HTML files to prevent a problematic trailing comma
- Un-ignore most .d.ts files, and format them
- Ignore change files (humans don't look at them and they're quickly deleted, so formatting doesn't matter)
- Other minor ignore path updates

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12214)

### DIFF
--- a/prettier.config.js
+++ b/prettier.config.js
@@ -1,6 +1,6 @@
 // https://prettier.io/docs/en/configuration.html
 module.exports = {
-  printWidth: 140,
+  printWidth: 140, // 120
   tabWidth: 2,
   singleQuote: true,
   // trailingComma: 'all',
--- a/prettier.config.js
+++ b/prettier.config.js
@@ -3,6 +3,7 @@ module.exports = {
   printWidth: 140,
   tabWidth: 2,
   singleQuote: true,
+  // trailingComma: 'all',
   overrides: [
     {
       files: 'apps/vr-tests/**/*',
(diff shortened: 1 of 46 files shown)


prettier.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   printWidth: 140,
```
layershifter:
```suggestion
  printWidth: 120,
```

image

ecraig12345: I'm doing that in a separate PR on Monday probably (120 does seem to be the most popular choice so far), but I just added a comment in the config so it's clear that's one of the proposed changes to be enabled.

ecraig12345:
@kenotron This is actually the non-automated parts (so they're readable). 😆The automated parts will be a separate PR since it's 3000+ files and literally impossible to review in github UI.

PR #12215, opened 2020-03-06
Panel: closeButton should stick to the right even if header is removed

### LINKED ISSUE #12195: Panel: 7.98.1 regressed position of close button for panels with no header text
## Environment Information

- **Package version(s)**: 7.98.1 is the first affected version
- **Browser and OS versions**: new Edge stable Version 80.0.361.66 (Official build) (64-bit)

(these are just the results of exporting-to-codepen the "Basic" [panel example](https://developer.microsoft.com/en-us/fabric#/controls/web/panel), removing the headerText prop from the panel, and updating the fabric version)

* [Repro in 7.98.1](https://codepen.io/dbjorge/pen/ExjwgEp) (note the panel's close button appearing at top left instead of top right)
* [No repro in 7.98.0](https://codepen.io/dbjorge/pen/BaNwLxK) (note the panel's close button appearing in top right)

#### Actual behavior:

Panel's close button appears at the top-left, like this:
screenshot of close button at top-left in v7.98.1

#### Expected behavior:

Panel's close button should appear at the top-right, like this:
screenshot of close button at top-right in v7.98.0

Requested priority: High (visual regression in patch release)

Products/sites affected: [Accessibility Insights for Web](https://github.com/microsoft/accessibility-insights-web)'s help content panels:

screenshot of misplaced close button in Accessibility Insights

aneeshack4:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12195 

#### Description of changes

Previous change #12025 regressed this behavior in an attempt to fix an IE11 specific bug - that's been handled by the removing of margins, there's no need to use `justify-content: space-between` for main axis alignment - we can achieve the same affect with `flex-end` as long as we have the header as `flex-grow: 1`.
image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12215)


aneeshack4:
Oooh closing this fell off my radar as shield got busy - this screener change is really strange https://screener.io/v2/states/ryQqxfIYZ.OfficeDev-office-ui-fabric-react/panel_flex/1024x768/Chrome/panel-with-no-navigation-default-0. I don't understand why the "Content goes here" gets cut off... and the explanation is apparently the height changed which seems unrelated...

aneeshack4:
The screener failure is still there because the header's `flex-grow: 1` is conflicting with the `flex-shrink: 1` of `hasCustomNavigation`:
image
I've played around with flexbox a lot here with no luck... removing this `flex-grow: 1` on header breaks all the other scenarios but keeping it breaks the navigation scenarios...

aneeshack4:
Actually I was able to get all the screener scenarios to look right with just `alignSelf: 'flex-start'` - at least they looked correct when I ran Storybook locally - without having to do `justify-content: flex-end` on `navigation` AND `flex-grow: 1` on the `header` let's see if screener fully passes...

aneeshack4:
@phkuo @khmakoto what do y'all think of this? Screener's fully passing but would like your input on this approach/approval.

khmakoto:
That should work, thanks @aneeshack4!

change/office-ui-fabric-react-2020-03-05-22-14-47-panel_flex.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Panel: closeButton should stick to the left even if header is removed",
```
khmakoto:
```suggestion
  "comment": "Panel: closeButton should stick to the right even if header is removed",
```

aneeshack4:
@khmakoto looks like the new snapshot is blank for some reason... 
image

aneeshack4:
I'll add a story to Panel for this new case!

aneeshack4:
The 2 screener regressions are in callout which makes no sense...

aneeshack4:
I guess Github didn't refresh when I last viewed the changes so I thought there wasn't a story added for close button, no header - removed the duplicate now.

aneeshack4:
Ok this should be good now:
image

aneeshack4:
@khmakoto or @phkuo are you able to accept the new screener scenario? https://screener.io/v2/states/Skgr32fUL.microsoft-fluentui/panel_flex/1024x768/Chrome/panel-with-no-header-close-button-default-0 If not, I can ask ben truelove.

khmakoto:
I can't, I'm not an admin of our screener account.

aneeshack4:
@betrue-final-final Would you mind looking over the screener results please?

PR #12216, opened 2020-03-06
withViewport: cleaning up resize observer on unmount.

dzearing:
Seems that resize observer was added but never cleaned up.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12216)


packages/office-ui-fabric-react/src/utilities/decorators/withViewport.tsx
```diff
@@ -116,6 +116,8 @@ export function withViewport<TProps extends { viewport?: IViewport }, TState>(
       if (this._viewportResizeObserver) {
```
KevinTCoughlin: This clause can be removed with your change

PR #12219, opened 2020-03-06
Fixing drag/drop issues with Groups in DetailsList

### LINKED ISSUE #12218: DetailsList Component: dragDropEvents  do not work correctly on Groups
### Environment Information

- **Package version(s)**: Latest

Code Pen: https://codepen.io/sachinma/pen/NWqwKeM

#### Actual behavior:

Groups do not become draggable
OnDragEnter, OnDrag does not work correctly on Groups

#### Expected behavior:
They work correctly

Products/sites affected: (if applicable) MS

sachinma:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12218

#### Description of changes

1. Add the draggble attribute to the Group if the group props return it to be draggable. Also using the prop canDrag while computing if a group is draggable.
2. Fixing the issue with onDragEnter and onDragLeave on the group. Now they are called with right parameters and at right time.
3. Using the css class returned by onDragEnter when rendering a group.
#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12219)


change/office-ui-fabric-react-2020-03-06-14-36-14-master.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Fixing drag/drop issues with Groups in DetailsList",
```
aneeshack4: Please write a more descriptive comment and preface with "DetailsList: " as these go out in release notes to our customers

packages/office-ui-fabric-react/src/components/GroupedList/GroupedListSection.tsx
```diff
@@ -222,9 +223,12 @@ export class GroupedListSection extends React.Component<IGroupedListSectionProps
     const groupShowAllProps: IGroupShowAllProps = { ...showAllProps, ...dividerProps };
     const groupFooterProps: IGroupFooterProps = { ...footerProps, ...dividerProps };
 
+    const isDraggable: boolean = !!this.props.dragDropHelper && this._getGroupDragDropOptions().canDrag!(group);
```
aneeshack4: Could you describe why this new isDraggable const is being introduced? More detailed description on this whole change box in the PR would help.

sachinma:
I have updated the description. Can you review it again?

sachinma:
Ping!

packages/office-ui-fabric-react/src/components/GroupedList/GroupedListSection.tsx
```diff
@@ -430,19 +434,19 @@ export class GroupedListSection extends React.Component<IGroupedListSectionProps
...
 
-    if (!isDropping) {
-      if (dragDropEvents && dragDropEvents.onDragLeave) {
-        dragDropEvents.onDragLeave!(event, undefined);
-      }
-    } else {
-      if (dragDropEvents && dragDropEvents.onDragEnter) {
-        dragDropEvents.onDragEnter(event, undefined);
+    if (isDropping !== newIsDropping) {
+      if (isDropping) {
```
KevinTCoughlin: I assume this logic has always been incorrectly inverted based on your changes?

sachinma: Yes, this might cause issue to someone who has taken a dependency on the old logic - but I feel that might not be the case. For example, dragEnter earlier was called very late (when dragLeave should be called). I am not sure if people would be taking dependency on that

KevinTCoughlin: Your fix seems correct, so let's just note it in case we need to revert / hotfix.

packages/office-ui-fabric-react/src/components/GroupedList/GroupedListSection.tsx
```diff
@@ -222,9 +223,12 @@ export class GroupedListSection extends React.Component<IGroupedListSectionProps
     const groupShowAllProps: IGroupShowAllProps = { ...showAllProps, ...dividerProps };
     const groupFooterProps: IGroupFooterProps = { ...footerProps, ...dividerProps };
 
+    const isDraggable: boolean = !!this.props.dragDropHelper && this._getGroupDragDropOptions().canDrag!(group);
+
     return (
       <div
         ref={this._root}
+        {...(isDraggable && { draggable: true })}
```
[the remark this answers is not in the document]
KevinTCoughlin: Does it change the UX? If so, perhaps update so this is opt-in rather than opt-out.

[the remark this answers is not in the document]
sachinma: Yes, this causes the whole group along with children to show as dragged item. This might look odd in some cases!. I will try to change this logic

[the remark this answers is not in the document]
sachinma: I have repurposed canDragGroups to now opt-in to this behavior. That property was recently introduced with this [commit](https://github.com/OfficeDev/office-ui-fabric-react/commit/2f8e4302d8998f13836e66212eb3fa20933e1748#diff-828536b170bd4765d79e649b3274718b), and i believe was conflicting with canDrag proprety that already existed. I am assuming it is unlikely someone took a dependency on that - and i feel this makes it better for the long run

PR #12221, opened 2020-03-06
chore: refactor `Popup` to be functional component

layershifter:
#### Pull request checklist

#### Description of changes

This PR converts `Popup` to be functional component. As `Popup` doesn't render any styles we don't have there breaking changes. Also fixes typings/makes them more strict in some places.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12221)


packages/fluentui/react/src/components/Popup/createReferenceFromContextClick.ts
```diff
@@ -1,15 +1,20 @@
...
-  function getBoundingClientRect() {
+  function getBoundingClientRect(): ClientRect {
     return {
       left,
       top,
       right,
-      bottom
+      bottom,
+
+      height: 0,
```
pompomon: Please, why do we need to setup `height` and `width` here?

layershifter: It's required by `ClientRect` typings...

packages/fluentui/react/src/components/Popup/Popup.tsx
```diff
@@ -547,17 +456,116 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
+    React.useEffect(() => {
+      if (inline && trapFocus) {
+        console.warn('Using "trapFocus" in inline popup leads to broken behavior for screen reader users.');
+      }
+      if (!inline && autoFocus) {
+        console.warn(
+          'Beware, "autoFocus" prop will just grab focus at the moment of mount and will not trap it. As user is able to TAB out from popup, better use "inline" prop to keep correct tab order.'
+        );
+      }
+    }, [autoFocus, inline, trapFocus]);
```
layershifter: Now we will have warnings only on props changes 👍

packages/fluentui/react/src/components/Popup/Popup.tsx
```diff
@@ -506,38 +411,42 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
+    newOpen ? trySetOpen(true, e) : schedulePopupClose(e);
+  };
 
-    this.closeTimeoutId = setTimeout(() => {
-      this.trySetOpen(false, e);
-    }, mouseLeaveDelay);
+  const schedulePopupClose = (e: React.KeyboardEvent) => {
+    closeTimeoutId.current = setTimeout(() => {
+      trySetOpen(false, e);
+    }, mouseLeaveDelay) as any;
```
layershifter: We have collision between Node and DOM typings

packages/fluentui/react/src/components/Popup/Popup.tsx
```diff
@@ -109,191 +109,135 @@ export interface PopupProps
...
 
-    const isOutsidePopupElement = this.popupDomElement && !isInsideNested;
-    return isOutsidePopupElement;
-  }
+    return isOutsidePopup;
+  };
 
-  getTriggerProps(triggerElement) {
+  const getTriggerProps = triggerElement => {
     const triggerProps: any = {};
```
layershifter: `any` will be fixed in next PR

packages/fluentui/react/src/components/Popup/Popup.tsx
```diff
@@ -506,38 +411,42 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
-    this.trySetOpen(false, e);
+  const dismissOnScroll = (e: TouchEvent | WheelEvent) => {
+    trySetOpen(false, e);
   };
 
-  trySetOpen(newValue: boolean, eventArgs: any) {
-    const isOpenedByRightClick = newValue && isRightClick(eventArgs);
+  const trySetOpen = (
+    newValue: boolean,
+    event: React.KeyboardEvent | React.MouseEvent | KeyboardEvent | MouseEvent | TouchEvent | WheelEvent
```
layershifter: Someday we should enable [`no-implicit-any`](https://www.typescriptlang.org/docs/handbook/compiler-options.html#compiler-options)...

PR #12227, opened 2020-03-06
Export GroupedListSection

CampbellOwen:
#### Pull request checklist

#### Description of changes

This change exports the GroupedListSection control used internally in GroupedList.

I need a version of GroupedList that supports non-contiguous groups i.e. given an array of groups, which each contain an array of items. In order to implement this control, I'd prefer to be able to reuse the GroupedListSection used in GroupedList. This way, my control can be a fairly thin wrapper over a List of GroupedListSections.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12227)

### DIFF
--- a/change/office-ui-fabric-react-2020-03-06-11-38-13-Public-GroupedListSection.json
+++ b/change/office-ui-fabric-react-2020-03-06-11-38-13-Public-GroupedListSection.json
@@ -1,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "Export GroupedListSection",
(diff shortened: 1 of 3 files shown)


KevinTCoughlin:
Seems reasonable. We should consider documenting the methods if we're making this component public.

The screener failure seems unrelated to your changes...

PR #12239, opened 2020-03-09
feat(compose): add compose to Fluent v0

layershifter:
This PR adds composition functionality to Fluent v0 via `compose()` function:

```tsx
const PrimaryButton = compose(Button, {
  displayName: 'PrimaryButton',
})
```

This will produce a single component in React tree:

image

#### Every component that uses hooks can be composed

There is no need to existing modify components except replacement `getUnhandledProps()` with `useUnhandledProps()`. See `Button` component as an example.

#### Unlimited composition

Every component can be composed multiple times:

```tsx
const PrimaryButton = compose(Button, {
  displayName: 'PrimaryButton',
})
const SecondaryButton = compose(PrimaryButton)
const TertiaryButton = compose(SecondaryButton)
```

#### Style aspects can be overridden, replaced & enhanced

```tsx
// will merge styles with Button's
const PrimaryButton = compose(Button, {
  displayName: 'PrimaryButton',
})
// will use only styles for SecondaryButton
const SecondaryButton = compose(Button, {
  displayName: 'SecondaryButton ',
  overrideStyles: true,
})
// will pass `primary` prop to styles functions
const PrimaryButton = compose(Button, {
  mapPropsToStyles: (props) => ({ primary: props.primary }),
})
```

#### Types are not inherited properly 

`compose()` function is type safe and will merge types properly, but current implementation of `as` prop breaks HTML props:

image

Can be solved with #12142.

#### Usage of React internals

`useReactElement()` uses React's internals to fetch `type` of current element and `props`. Props can be fetched obviously, but for `type` it's more hard. To be honest even `type` is not required however we need a way to pass `options` to a proper hook.

## Follow ups ⏭ 

- We need somehow notify user that component is not composable
- Add support for `React.forwardRef()`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12239)


packages/fluentui/docs/src/examples/components/Button/Types/ButtonExample.shorthand.tsx
```diff
@@ -1,6 +1,143 @@
...
+type ComposedButtonProps = {
+  tertiary?: boolean;
+};
+
+type ComposedButtonStylesProps = ComposedButtonProps;
+
+const ComposedButton = compose<ComposedButtonProps, ButtonBehaviorProps, ComposedButtonStylesProps, ButtonProps>(Button, {
+  className: 'ui-composed-button',
+  displayName: 'ComposedButton',
+  mapPropsToStyles: props => ({ tertiary: props.tertiary })
```
layershifter: Should be chain, not last one

packages/fluentui/docs/src/examples/components/Button/Types/ButtonExample.shorthand.tsx
```diff
@@ -1,6 +1,143 @@
...
+
+type OverriddenButtonStylesProps = Required<OverriddenButtonProps>;
+
+const OverriddenButton = compose<OverriddenButtonProps, ButtonBehaviorProps, OverriddenButtonStylesProps, ButtonProps>(Button, {
+  className: 'ui-overridden-button',
+  displayName: 'OverriddenButton',
+  mapPropsToStyles: props => ({
+    fitted: props.fitted
+  }),
+  handledProps: ['fitted'],
```
layershifter: Should be merged between composed components

packages/fluentui/docs/src/examples/components/Button/Types/ButtonExample.shorthand.tsx
```diff
@@ -1,6 +1,143 @@
...
+};
+
+type OverriddenButtonStylesProps = Required<OverriddenButtonProps>;
+
+const OverriddenButton = compose<OverriddenButtonProps, ButtonBehaviorProps, OverriddenButtonStylesProps, ButtonProps>(Button, {
+  className: 'ui-overridden-button',
+  displayName: 'OverriddenButton',
+  mapPropsToStyles: props => ({
+    fitted: props.fitted
+  }),
```
layershifter: Try to use inferring there

packages/fluentui/react-bindings/src/compose/useCompose.ts
```diff
@@ -0,0 +1,50 @@
...
+
+// https://github.com/facebook/react/blob/master/packages/react/src/ReactCurrentOwner.js
+type ReactCurrentOwner = {
+  current: null | ReactFiber;
+};
+
+function useCompose<AllProps, BehaviorProps, StylesProps>(
+  props: AllProps
+): ComposeResolvedOptions<AllProps, BehaviorProps, StylesProps> | null {
+  const { ReactCurrentOwner: currentOwner } = (React as any).__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;
```
mnajdova: Please tell me we won't do this in the final version :D

mnajdova: Or you will be fired I guess 🗡

packages/fluentui/react-bindings/src/compose/compose.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+function compose<OverrideProps, BehaviorProps, StylesProps, ComponentProps = {}>(
+  InputComponent: ComposeInputComponent<ComponentProps>,
+  options: ComposeInputOptions<ComponentProps & OverrideProps, BehaviorProps, StylesProps> = {}
+): ComposePreparedComponent<ComponentProps & OverrideProps> {
+  const ComposedComponent: React.FunctionComponent<ComponentProps & OverrideProps> & {
+    [COMPOSE_PROP]: ComposePreparedOptions;
+  } = InputComponent.bind(null);
+
+  const { handledProps = [], mapPropsToBehavior, mapPropsToStyles } = options;
```
dzearing: if you pass in a composed component, where is the merging happening? I am not seeing it

packages/fluentui/react-bindings/src/compose/useCompose.ts
```diff
@@ -0,0 +1,32 @@
...
+
+      behaviorProps: mapPropsToBehaviorChain.reduce((acc, fn) => ({ ...acc, ...fn(props) }), {}),
+      stylesProps: mapPropsToStylesChain.reduce((acc, fn) => ({ ...acc, ...fn(props) }), {}),
+
+      handledProps,
+      overrideStyles
+    };
+  }
+
+  return null;
```
layershifter:
```suggestion
  return ElementType?.[COMPOSE_PROP] || null;
```

As `behaviorProps` & `stylesProps` should be resolved in hooks with proper state

packages/fluentui/react-bindings/src/compose/compose.tsx
layershifter: Few lines below: https://github.com/OfficeDev/office-ui-fabric-react/pull/12239/files#diff-ad68020a8eafa54ca1569743fb0a40b2R53

packages/fluentui/react-bindings/src/compose/compose.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const inputOptions: ComposePreparedOptions = InputComponent[COMPOSE_PROP] || {
+    className: process.env.NODE_ENV === 'production' ? '' : 'no-classname-🙉',
+    displayNames: [],
+
+    mapPropsToBehaviorChain: [],
+    mapPropsToStylesChain: [],
+
+    handledProps: [],
+    overrideStyles: false
+  };
```
layershifter:
```suggestion
  const inputOptions: ComposePreparedOptions = InputComponent[COMPOSE_PROP] || defaultOptions
```

packages/fluentui/react-bindings/src/compose/compose.tsx
```diff
@@ -0,0 +1,60 @@
+import * as React from 'react';
+import { ComposeInputComponent, ComposeInputOptions, ComposePreparedComponent, ComposePreparedOptions } from './types';
+
+export const COMPOSE_PROP = 'FLUENT_COMPOSE_CONFIG';
+
+function computeDisplayNames(InputComponent: ComposeInputComponent, options: ComposeInputOptions<any, any, any>): string[] {
```
layershifter: This can be simplified

packages/fluentui/react-bindings/src/compose/useCompose.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export default useCompose;
```
layershifter:
```suggestion
export default useComposeOptions;
```

packages/fluentui/react-bindings/src/compose/useReactElement.ts
```diff
@@ -0,0 +1,26 @@
...
+};
+
+const useReactElement = <T extends React.ElementType, P extends Record<string, any>>(): [
+  ReactFiber<T, P>['type'] | undefined,
+  P | undefined,
+] => {
+  const currentOwner: ReactCurrentOwner<T, P> = (React as any).__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
+    .ReactCurrentOwner;
+
+  return [currentOwner.current?.type, currentOwner.current?.pendingProps];
```
mnajdova: Can we return an object here?

packages/fluentui/react-bindings/src/hooks/useStyles.ts
```diff
@@ -63,22 +65,34 @@ const useStyles = <StyleProps extends PrimitiveProps>(
...
   const {
     className = process.env.NODE_ENV === 'production' ? '' : 'no-classname-🙉',
     mapPropsToStyles = () => ({} as StyleProps),
     mapPropsToInlineStyles = () => ({} as InlineStyleProps<StyleProps>),
     rtl = false,
   } = options;
+  const componentStylesProps = mapPropsToStyles();
+
+  // `composeProps` should include all props including stylesProps as they can contain state
+  const composeProps = { ...componentProps, ...componentStylesProps };
```
mnajdova: Just to clarify, we have here all styles props from the prev component + this component's props. We do not have the state from the component, it was not assigned previously to some styles props?

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleCompose.shorthand.tsx
```diff
@@ -0,0 +1,163 @@
...
+// Adds a custom design term
+//
+
+type TertiaryButtonProps = {
+  tertiary?: boolean;
+};
+
+type TertiaryButtonStylesProps = TertiaryButtonProps;
+
+const ComposedButton = compose<TertiaryButtonProps, TertiaryButtonStylesProps, ButtonProps, {}>(Button, {
```
mnajdova:
```suggestion
const TertiaryButton = compose<TertiaryButtonProps, TertiaryButtonStylesProps, ButtonProps, {}>(Button, {
```

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleCompose.shorthand.tsx
```diff
@@ -0,0 +1,163 @@
...
+  className: 'ui-tertiary-button',
+  displayName: 'TertiaryButton',
+  mapPropsToStylesProps: props => ({ tertiary: props.tertiary }),
+  handledProps: ['tertiary'],
+});
+
+// Adds overrides for a design term
+//
+
+const CompactTertiaryButton = compose(ComposedButton, {
```
mnajdova:
```suggestion
const CompactTertiaryButton = compose(TertiaryButton, {
```

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleCompose.shorthand.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+//
+// Example
+//
+
+const ButtonExample = () => (
+  <Provider theme={customTheme}>
+    <Header as="h3" content="A tertiary button" description="Adds a custom design term" />
+    <Flex>
+      <ComposedButton content="Click here" />
```
mnajdova:
```suggestion
      <TertiaryButton content="Click here" />
```

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleCompose.shorthand.tsx
```diff
@@ -0,0 +1,163 @@
...
+      <ComposedButton content="Click here" tertiary />
```
mnajdova:
```suggestion
      <TertiaryButton content="Click here" tertiary />
```

packages/fluentui/react-bindings/src/index.ts
```diff
@@ -13,6 +16,7 @@ export { default as useAccessibility } from './hooks/useAccessibility';
 export { default as useAutoControlled } from './hooks/useAutoControlled';
 export { default as unstable_useDispatchEffect } from './hooks/useDispatchEffect';
 export { default as useIsomorphicLayoutEffect } from './hooks/useIsomorphicLayoutEffect';
+export { default as useUnhandledProps } from './hooks/useUnhandledProps';
```
mnajdova:
```suggestion
```

We have this already on line 22

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleCompose.shorthand.tsx
```diff
@@ -0,0 +1,163 @@
...
+  };
+};
+
+const componentStyles: {
+  // @ts-ignore TODO fix type there
+  [C in keyof ComponentStylesProps]: ComponentSlotStylesInput<ComponentStylesProps[C], ComponentVariables[C]>;
+} = {
+  TertiaryButton: {
+    root: ({ props: p, variables: v }) => ({
+      ...(p.tertiary && {
```
mnajdova: We are basically breaking the disabled and any other state related colors that we have specified before with this. Can we update the styles to apply these colors if it is not disabled? Apart from this I don't have better idea of how to handle this :( This means that we will have always some dependency on the base styles that we are composing, but I guess it is expected. What do you think @layershifter ?

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleCompose.shorthand.tsx
```diff
@@ -0,0 +1,163 @@
...
+  fitted?: boolean;
+};
+
+type OverriddenButtonStylesProps = Required<OverriddenButtonProps>;
+
+const OverriddenButton = compose<OverriddenButtonProps, OverriddenButtonStylesProps, ButtonProps, {}>(Button, {
+  className: 'ui-overridden-button',
+  displayName: 'OverriddenButton',
+  mapPropsToStylesProps: props => ({
+    fitted: props.fitted,
```
mnajdova: It would be good if we can use also some other style prop here from the `Button`, just as a showcase that we have access to them too. I would propose `primary` or `disabled`.

packages/fluentui/react-bindings/src/index.ts
layershifter: Good catch, fixed 👍

packages/fluentui/react-bindings/src/compose/useReactElement.ts
layershifter: Done, I also renamed it to `useCurrentReactElement` 👍

packages/fluentui/react-bindings/src/hooks/useStyles.ts
layershifter: > We do not have the state from the component, it was not assigned previously to some styles props?

Yep, only styles props that in host components were mapped from state

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleCompose.shorthand.tsx
layershifter: image

Agree, added 👍

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleCompose.shorthand.tsx
layershifter: Yep, you have to resolve styles priorities even for CSS. So any combination of design terms should be predefined

packages/fluentui/react-bindings/src/hooks/useStyles.ts
mnajdova: I hope this won't be a problem :\

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleCompose.shorthand.tsx
mnajdova: I understand, but in CSS you can directly write the new class in the right order, here you don't have that option.. I mention this because `disabled` is a great example of a prop that you expect to be defined in the initial styles, and tertiary is not something very common.

mnajdova:
> Types are not inherited properly
> compose() function is type safe and will merge types properly, but current implementation of as prop breaks HTML props:
> 
> Can be solved with #12142.

This scares me a bit. Is there a reason why we didn't move further with #12142? Should we merge that PR first?

layershifter:
> Is there a reason why we didn't move further with #12142?

IDK, let's discuss it today.

> Should we merge that PR first?

It will require updates to many components. Should we block progress with this PR?

mnajdova:
> It will require updates to many components. Should we block progress with this PR?

Not sure about this. I would say type safety is important, depends on how urgent we consider that this is. Let's see what @miroslavstastny thinks

PR #12241, opened 2020-03-09
chore(themes): remove styles overrides for `MenuItem` in High Contrast theme

layershifter:
#### Pull request checklist

#### Description of changes

Removes styles overrides from High Contrast theme and creates new tokens instead for `MenuItem` component. New Screener tests were added in #12294.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12241)


packages/fluentui/react/src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -402,30 +404,41 @@ const menuItemStyles: ComponentSlotStylesPrepared<MenuItemPropsAndState, MenuVar
...
-    display: 'block',
-    overflow: 'hidden',
-    height: pxToRem(16),
-    width: pxToRem(16)
-  })
+      backgroundImage: submenuIndicatorUrl(v.indicatorColor, p.vertical),
+
+      ...(p.active && {
+        backgroundImage: submenuIndicatorUrl(v.activeIndicatorColor, p.vertical),
+        ...(p.primary && { backgroundImage: submenuIndicatorUrl(v.activePrimaryIndicatorColor, p.vertical) })
```
mnajdova: One potential regression could be this: 
image

The same is for the non-vertical primary menu. Not sure if these are valid variants for teams theme, but I noticed them, so thought I should report :)

packages/fluentui/react/src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -150,7 +101,6 @@ const pointingBeak = ({
 const menuItemStyles: ComponentSlotStylesPrepared<MenuItemPropsAndState, MenuVariables> = {
```
mnajdova: Menu's divider and borders are off in dark theme:

image

image

packages/fluentui/react/src/themes/teams-high-contrast/components/Menu/menuVariables.ts
```diff
@@ -18,10 +18,40 @@ export default (siteVars: any): Partial<MenuVariables> => ({
...
+  wrapperColorFocus: siteVars.colors.black,
+  primaryWrapperColorHover: siteVars.colors.black,
+  wrapperColorHover: siteVars.colors.black,
+
+  primaryWrapperColorFocus: siteVars.colors.black,
+  primaryWrapperBackgroundColorFocus: siteVars.accessibleCyan,
+  underlinedColorHover: siteVars.colors.white,
+  underlinedWrapperColorHover: siteVars.colors.white,
+
+  iconOnlyBackgroundColorHover: siteVars.accessibleCyan,
```
mnajdova: I noticed that all hover styles are changed from accessibleYellow to accessibleCyan in a prev PR. I really hope the changes were intentional..

packages/fluentui/react/src/themes/teams/components/Menu/menuItemStyles.ts
mnajdova: Never mind it looks like this on master

packages/fluentui/react/src/themes/teams-high-contrast/components/Menu/menuVariables.ts
layershifter: Yep, we discussed this with @pompomon when we compared with `master` and it was really confusing for us. For reference, this was done in #12112.

packages/fluentui/react/src/themes/teams/components/Menu/menuItemStyles.ts
layershifter: Good catch, I expected that it will be catched by Screener. Will check is there a visual test for it, if no, will add it before merging this 👍

PR #12244, opened 2020-03-10
Added useFocusRects hook (to replace initializeFocusRect)

xugao:
#### Pull request checklist

#### Description of changes
1. Added hook implementation of `initializeFocusRect`. Hook in addition removes the attached event listeners. (`initializeFocusRect` doesn't clean up events currently)
2. Use hook implementation everywhere except in `BaseComponent`
3. Allow disable Focus style for all Fabric components by setting `window. disableFabricFocusRects = true` (related to discussion in email _RE: Perf impact caused by Recalculate Style when hitting 'tab' key_)
4. Removed classes `.is-focusVisible` `.is-focusHidden`. (TODO: confirm this is safe to remove) 

There are some follow up changes after this one:
1. Remove usage of `initializeFocusRect` from `BaseComponent` and use hook instead in individual components. **Caveat**: ideally we only use the hook in `Fabric` component but require user to wrap with  `Fabric` if they want to have focus style. but i don't think we can make such breaking change before next major release.
2. Update `Fabric` to a function component and call hook `useFocusRect` directly.

**Note:**
For long-term, we should use `:focus-visible` (https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) to apply custom focus border style. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12244)


packages/utilities/src/useFocusRects.ts
```diff
@@ -0,0 +1,81 @@
...
+import { isDirectionalKeyCode } from './keyboard';
+import { setFocusVisibility } from './setFocusVisibility';
+
+export { IsFocusVisibleClassName } from './setFocusVisibility';
+
+/**
+ * Counter for mounting component that uses focus rectangle.
+ * We want to cleanup the listners before last component that uses focus rectangle unmounts.
+ */
+let mountCounter = 0;
```
dzearing: won't work on child windows; move to weakmap [key: window]: number;

dzearing:
It is weird that flamegrill isn't chiming in.

JasonGore:
I don't see any builds that have successful build phases. The perf job won't even start unless the build phase is successful. Looks like the PR needs change files.

xugao:
Size increase in `GroupedList` is due to it uses other component like `Link` which extends base component. So the bundle actually contains both `initializedFocusRects` and `useFocusRects`

My next PR will remove usage of `initializedFocusRects` from `BaseComponent`

PR #12246, opened 2020-03-10
chore(themes): remove styles overrides in High Contrast theme for Alert & Attachment

layershifter:
#### Pull request checklist

#### Description of changes

This PR:

- Renames param for `getBorderFocusStyles()` to be `variables` to emphasize that `variables` from component's should be passed instead of `siteVariables`. Next step there will be to omit usage of `siteVariables` in components.
- Removes styles overrides from High Contrast theme and creates new tokens instead.

Scenarios for removed styles:

- Alert: hover on `dismissAction` (covered by `AlertExampleDismissible.shorthand.steps.ts`)
- Alert: focus by keyboard on `dismissAction` (covered by `AlertExampleDismissible.shorthand.steps.ts`)
- Alert: focus by keyboard & hover on `dismissAction` (covered by `AlertExampleDismissible.shorthand.steps.ts`)

- Attachment: focus by keyboard (should affect `Button` & `Icon`)  (covered by `AttachmentActionableExample.shorthand.steps.ts`)
- Attachment: hover (should affect `Button` & `Icon`) (covered by `AttachmentActionableExample.shorthand.steps.ts`)


packages/fluentui/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -129,14 +128,30 @@ const alertStyles: ComponentSlotStylesPrepared<AlertProps, AlertVariables> = {
...
       },
 
       ':focus': borderFocusStyles[':focus'],
       ':focus-visible': {
+        backgroundColor: v.focusBackgroundColor,
         ...iconFilledStyles,
-        ...borderFocusStyles[':focus-visible']
+        ...borderFocusStyles[':focus-visible'],
+
+        ':hover': {
```
mnajdova: This is really confusing, we have first in the file:
```
':hover': {
   ...getBorderStyles({...})[':focus-visible']
}
```
and than:
```
':focusVisible': {
  // ...
  ':hover': { ... }
}
```
and we have the backgroundColor in both of them.. I have no idea which one will win and why...

mnajdova: Should we always define these in some order?

packages/fluentui/react/src/themes/teams/getBorderFocusStyles.ts
```diff
@@ -4,7 +4,16 @@ import * as React from 'react';
 type CSSBorderStyles = Pick<React.CSSProperties, 'borderWidth' | 'borderRadius'>;
 
 type BorderFocusStyles = CSSBorderStyles & {
-  siteVariables?: SiteVariablesPrepared;
+  variables?:
+    | SiteVariablesPrepared
+    | {
```
mnajdova: I hope we will get rid of the `SiteVariablesPrepared` soon.

packages/fluentui/react/src/themes/teams/components/Alert/alertStyles.ts
layershifter: Oops, it had a sense before when we have had different values for `hoverBackgroundColor` & `focusBackgroundColor`. Now it can be removed at 🤔

layershifter: I removed `:hover` as there is no sense in it

packages/fluentui/react/src/themes/teams/components/Attachment/attachmentStyles.ts
```diff
@@ -5,38 +5,65 @@ import { pxToRem } from '../../../../utils';
...
+
+      ...((p.actionable || p.onClick) && {
+        cursor: 'pointer',
+
+        ':focus-visible': {
+          ...borderFocusStyles[':focus-visible'],
+          backgroundColor: v.focusBackgroundColor,
+          color: v.focusColor,
+
+          [`& .${Button.className}`]: {
```
mnajdova: A bit worried for these empty subselectors, but I don't have any better proposal. We should come up soon with a strategy of how to handle these scenarios consistently.

layershifter:
> They are potentially breaking changes for the Teams themes...

We don't have there enough clarity to be honest 😿

We consider `variables` as public API and there are no breaking changes. As styles overrides are merged on top of base themes these changes _should not_ be breaking, but it's really hard to say. It the same time I definitely don't like an idea to mark any changes in styles as breaking...

PR #12250, opened 2020-03-10
feat: add disabled version for Dropdown and Input

### LINKED ISSUE #12316: Dropdown: loading case has <li> contained by another <li>
Opening the `loading` Dropdown example will throw a warning in console: `validateDOMNesting(...): <li> cannot appear as a descendant of <li>.`

Since the `no results` list item is also suffering from the same issue, these two items should be fixed to not use nested `<li>`.

### Environment Information

- **Package version(s)**: Fluent 0.46.0

#### Actual behavior:

Throws an error because of nested `<li>` elements.

#### Expected behavior:

Second `<li>` element should be replaced with a different type.

Requested priority: Normal

Products/sites affected: FluentUI Dropdown, Northstar Dropdown uses.

silviuavram:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #12316 

#### Description of changes

Provides `disabled` behavior for `Dropdown` and also for `Input` (because we are using `Input` for `DropdownSearchInput`).

- only executes event handlers from `predefinedProps` if `disabled` is true.
- add `disabled` prop on native `input` and `button`.
- add `disabled` to Downshift so it does not give us handlers anymore.

Added examples in docsite for both components and also unit tests.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12250)

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Dropdown/State/DropdownExampleDisabled.shorthand.tsx
+++ b/packages/fluentui/docs/src/examples/components/Dropdown/State/DropdownExampleDisabled.shorthand.tsx
@@ -1,0 +1,15 @@
+import { Dropdown, Flex } from '@fluentui/react';
+import * as React from 'react';
+
+const inputItems = ['Bruce Wayne', 'Natasha Romanoff', 'Steven Strange', 'Alfred Pennyworth'];
+
+const DropdownExampleDisabled = () => {
+  return (
+    <Flex column gap="gap.smaller">
+      <Dropdown disabled={true} items={inputItems} placeholder="Select your hero, if you can ..." />
+      <Dropdown disabled={true} search={true} items={inputItems} placeholder="Start typing a name, if you can ..." />
+    </Flex>
+  );
+};
+
+export default DropdownExampleDisabled;
--- a/packages/fluentui/docs/src/examples/components/Dropdown/State/DropdownExampleDisabled.shorthand.tsx
+++ b/packages/fluentui/docs/src/examples/components/Dropdown/State/DropdownExampleDisabled.shorthand.tsx
@@ -6,8 +6,8 @@ const inputItems = ['Bruce Wayne', 'Natasha Romanoff', 'Steven Strange', 'Alfred
 const DropdownExampleDisabled = () => {
   return (
     <Flex column gap="gap.smaller">
-      <Dropdown disabled={true} items={inputItems} placeholder="Select your hero, if you can ..." />
-      <Dropdown disabled={true} search={true} items={inputItems} placeholder="Start typing a name, if you can ..." />
+      <Dropdown disabled items={inputItems} placeholder="Select your hero, if you can ..." />
+      <Dropdown disabled search items={inputItems} placeholder="Start typing a name, if you can ..." />
     </Flex>
   );
 };
--- a/packages/fluentui/react/src/components/Button/Button.tsx
+++ b/packages/fluentui/react/src/components/Button/Button.tsx
@@ -193,10 +193,6 @@ const Button: React.FC<WithAsProp<ButtonProps>> &
   };
 
   const handleFocus = (e: React.SyntheticEvent) => {
-    if (disabled) {
-      return;
-    }
-
     _.invoke(props, 'onFocus', e, props);
   };
 
--- a/packages/fluentui/react/src/components/Button/Button.tsx
+++ b/packages/fluentui/react/src/components/Button/Button.tsx
@@ -193,6 +193,10 @@ const Button: React.FC<WithAsProp<ButtonProps>> &
   };
 
   const handleFocus = (e: React.SyntheticEvent) => {
+    if (disabled) {
+      return;
+    }
+
     _.invoke(props, 'onFocus', e, props);
   };
 
--- a/packages/fluentui/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/fluentui/react/src/components/Dropdown/Dropdown.tsx
@@ -479,12 +479,11 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
                         }),
                         overrideProps: (predefinedProps: BoxProps) => ({
                           onClick: e => {
-                            if (disabled) {
-                              return;
+                            if (!disabled) {
+                              getToggleButtonProps({ disabled }).onClick(e);
                             }
 
                             _.invoke(predefinedProps, 'onClick', e);
-                            getToggleButtonProps({ disabled }).onClick(e);
                           }
                         })
                       })}
@@ -559,19 +558,17 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
               _.invoke(predefinedProps, 'onFocus', e, predefinedProps);
             },
             onBlur: e => {
-              if (disabled) {
-                return;
+              if (!disabled) {
+                onBlur(e);
               }
 
-              onBlur(e);
               _.invoke(predefinedProps, 'onBlur', e, predefinedProps);
             },
             onKeyDown: e => {
-              if (disabled) {
-                return;
+              if (!disabled) {
+                onKeyDown(e);
               }
 
-              onKeyDown(e);
               _.invoke(predefinedProps, 'onKeyDown', e, predefinedProps);
             }
           })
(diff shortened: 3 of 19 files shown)


packages/fluentui/docs/src/examples/components/Dropdown/State/DropdownExampleDisabled.shorthand.tsx
```diff
@@ -0,0 +1,15 @@
+import { Dropdown, Flex } from '@fluentui/react';
+import * as React from 'react';
+
+const inputItems = ['Bruce Wayne', 'Natasha Romanoff', 'Steven Strange', 'Alfred Pennyworth'];
+
+const DropdownExampleDisabled = () => {
+  return (
+    <Flex column gap="gap.smaller">
+      <Dropdown disabled={true} items={inputItems} placeholder="Select your hero, if you can ..." />
+      <Dropdown disabled={true} search={true} items={inputItems} placeholder="Start typing a name, if you can ..." />
```
pompomon:
```suggestion
      <Dropdown disabled search={true} items={inputItems} placeholder="Start typing a name, if you can ..." />
```

iirc @layershifter likes it this way

packages/fluentui/docs/src/examples/components/Dropdown/State/DropdownExampleDisabled.shorthand.tsx
```diff
@@ -0,0 +1,15 @@
+import { Dropdown, Flex } from '@fluentui/react';
+import * as React from 'react';
+
+const inputItems = ['Bruce Wayne', 'Natasha Romanoff', 'Steven Strange', 'Alfred Pennyworth'];
+
+const DropdownExampleDisabled = () => {
+  return (
+    <Flex column gap="gap.smaller">
+      <Dropdown disabled={true} items={inputItems} placeholder="Select your hero, if you can ..." />
```
pompomon:
```suggestion
      <Dropdown items={inputItems} placeholder="Select your hero, if you can ..." />
```
iirc @layershifter  likes it this way

packages/fluentui/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -82,6 +82,9 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
   /** The initial value (or value array if the array has multiple selection). */
   defaultValue?: ShorthandValue<DropdownItemProps> | ShorthandCollection<DropdownItemProps>;
 
+  /** A Dropdown can be disabled. */
+  disabled?: boolean;
```
mnajdova: I would use some description consistent to the Button's disabled prop: 
`/** A button can show that it cannot be interacted with. */`

packages/fluentui/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -483,8 +487,13 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
                         }),
                         overrideProps: (predefinedProps: BoxProps) => ({
                           onClick: e => {
+                            if (disabled) {
```
mnajdova: check whether the color of the indicator should change when disabled

packages/fluentui/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -542,25 +552,46 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
             content,
+            disabled,
             id: triggerButtonId,
             fluid: true,
             styles: styles.triggerButton,
             ...restTriggerButtonProps
           }),
           overrideProps: (predefinedProps: ButtonProps) => ({
             onClick: e => {
+              if (disabled) {
```
mnajdova: The Button already has this in it's own onClick event.. I don't think it is necessary here

packages/fluentui/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -542,25 +552,46 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+                e.preventDefault();
+                return;
+              }
+
               onClick(e);
               _.invoke(predefinedProps, 'onClick', e, predefinedProps);
             },
             onFocus: e => {
+              if (disabled) {
```
mnajdova: I would rather handle this in the Button component

packages/fluentui/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -159,15 +159,17 @@ const dropdownStyles: ComponentSlotStylesPrepared<DropdownPropsAndState, Dropdow
...
     display: 'flex',
     justifyContent: 'center',
 
     backgroundImage: toggleIndicatorUrl(v.color),
     backgroundPosition: 'center',
     backgroundRepeat: 'no-repeat',
-    cursor: 'pointer',
+    ...(!p.disabled && {
+      cursor: 'pointer'
+    }),
```
layershifter:
```suggestion
    cursor: 'pointer',
    ...(p.disabled && {
      cursor: 'default'
    }),
```

Let's prefer CSS approach

packages/fluentui/react/src/components/Input/Input.tsx
```diff
@@ -185,13 +190,23 @@ class Input extends AutoControlledComponent<WithAsProp<InputProps>, InputState>
...
       _.invoke(predefinedProps, 'onClick', e, this.props);
     }
   });
 
   handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    if (this.props.disabled) {
+      e.preventDefault();
+      return;
+    }
+
```
layershifter:
```suggestion
```



Can you please explain why we need this? It's not possible to dispatch a `change` for `disabled` `input` even manually:
https://codesandbox.io/s/blue-architecture-2l2i4

packages/fluentui/react/src/components/Input/Input.tsx
```diff
@@ -185,13 +190,23 @@ class Input extends AutoControlledComponent<WithAsProp<InputProps>, InputState>
...
     onClick: (e: React.SyntheticEvent) => {
+      if (this.props.disabled) {
+        e.preventDefault();
+        return;
+      }
+
       this.handleOnClear(e);
       this.inputRef.current.focus();
       _.invoke(predefinedProps, 'onClick', e, this.props);
     }
```
layershifter:
```suggestion
      if (!this.props.disabled) {
        this.handleOnClear(e);
        this.inputRef.current.focus();
      }
      
      _.invoke(predefinedProps, 'onClick', e, this.props);
    }
```

Can you please explain why we should `preventDefault()` in this case?

packages/fluentui/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -859,12 +891,24 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     getInputProps: (options?: GetInputPropsOptions) => any
   ) => (predefinedProps: DropdownSearchInputProps) => {
     const handleInputBlur = (e: React.SyntheticEvent, searchInputProps: DropdownSearchInputProps) => {
+      if (disabled) {
+        e.preventDefault();
```
layershifter: Also not sure about this as `disabled` input is not focusable:

https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_input_disabled

kolaps33:
I was looking on the PR. Checkout the branch and was looking on it with screen readers, it looks good :)

But it looked for me suspicious that we need to stop propagate events in "onClick" and when I was debugging it, then I figure out is it because of chevron div(image) element. Am I correct? (pl. see screenshot)

image

packages/fluentui/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -1296,4 +1297,61 @@ describe('Dropdown', () => {
...
+      const { clickOnToggleIndicator, toggleIndicatorNode, getItemNodes } = renderDropdown({
+        disabled: true
+      });
+
+      clickOnToggleIndicator();
+
+      expect(getItemNodes()).toHaveLength(0);
+
+      toggleIndicatorNode.focus();
+
```
layershifter: A newline after each action looks a bit strange, is it intent?

packages/fluentui/react/src/themes/teams/components/Input/inputStyles.ts
```diff
@@ -42,6 +42,15 @@ const inputStyles: ComponentSlotStylesPrepared<InputProps & InputState, InputVar
       opacity: 1 // undo Firefox default opacity
     },
 
+    // Overrides for "disabled" inputs
+    ...(p.disabled && {
+      color: v.colorDisabled,
+      boxShadow: 'none',
+      '::placeholder': {
+        color: v.colorDisabled
+      }
```
layershifter:
```suggestion
```

This will overwrite previous rule completely I suggest to change it to:

```
    '::placeholder': {
      color: v.placeholderColor,
      opacity: 1 // undo Firefox default opacity,
      ...p.disabled && { color: v.colorDisabled },
    },
```

packages/fluentui/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -73,18 +73,20 @@ const dropdownStyles: ComponentSlotStylesPrepared<DropdownPropsAndState, Dropdow
     borderRadius: v.containerBorderRadius,
     ...(p.open && p.position === 'above' && { borderRadius: v.openAboveContainerBorderRadius }),
     ...(p.open && p.position === 'below' && { borderRadius: v.openBelowContainerBorderRadius }),
-    ':hover': {
-      backgroundColor: v.backgroundColorHover,
-      borderColor: v.borderColorHover,
+    ...(!p.disabled && {
```
layershifter: Is there any way to avoid negative conditions? I.e.

```tsx
':hover': {
   ...p.disabled && {
      background: v.disabledBackgroundColor,
      borderColor: v.disabledBorderColor,
   }
}
```

packages/fluentui/react/src/components/Button/Button.tsx
```diff
@@ -193,6 +193,10 @@ const Button: React.FC<WithAsProp<ButtonProps>> &
   };
 
   const handleFocus = (e: React.SyntheticEvent) => {
+    if (disabled) {
```
layershifter: ⚠️ Can you please explain why? Does it work in the same way for plain `button`?

packages/fluentui/react/test/specs/components/Dropdown/Dropdown-test.tsx
silviuavram: Not trying to make it strange intentionally, just separating test steps from expects.

packages/fluentui/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
silviuavram: well, it's `!p.disabled` so I don't know how to negate that. any suggestions?

packages/fluentui/react/src/components/Button/Button.tsx
silviuavram: Just copy pasted how it's done for handleClick.

packages/fluentui/react/src/themes/teams/components/Input/inputStyles.ts
silviuavram: changed, thanks!

packages/fluentui/react/src/components/Button/Button.tsx
layershifter: I suggest to remove it as by it's nature `disabled` `button` should not be focusable: https://codesandbox.io/s/focused-feather-2koj0

silviuavram: ok, sure :)

packages/fluentui/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
layershifter: I expanded it below a bit:

```tsx
':hover': {
      backgroundColor: v.backgroundColorHover,
      borderColor: v.borderColorHover,

      ...(p.open && {
        borderColor: v.openBorderColorHover
      }),

      ...p.disabled  && {
         backgroundColor: v.backgroundColor,
         borderColor: v.borderColor, // this will color to the default one, may be a separate variable `disabledBorderColorHover`
      },

      [`& .${Dropdown.slotClassNames.triggerButton}`]: {
        color: v.triggerButtonColorHover,

        ...p.disabled && {
          color: v.color, // may be a separate variable `disabledTriggerColorHover`          
        }
      }
    },
```


Why it's better that `!disabled`?
- it's closer to CSS
- it creates valid variables for the state, if we will need different values for HC theme it will be easy to add them

PR #12252, opened 2020-03-10
chore(themes): remove styles overrides in High Contrast theme for Dialog & Button

layershifter:
#### Pull request checklist

#### Description of changes

Removes styles overrides from High Contrast theme and creates new tokens instead for `Button` & `Dialog` components.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12252)


packages/fluentui/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -121,10 +120,7 @@ const buttonStyles: ComponentSlotStylesPrepared<ButtonStylesProps, ButtonVariabl
           boxShadow: 'none',
           ...borderFocusStyles[':focus']
         },
-
-        ':focus-visible': {
-          ...borderFocusStyles[':focus-visible']
```
layershifter: There is no sense in this spread

packages/fluentui/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -71,13 +71,12 @@ const buttonStyles: ComponentSlotStylesPrepared<ButtonStylesProps, ButtonVariabl
           boxShadow: 'none'
         },
 
-        ':focus': {
```
layershifter: There is no sense in this spread

packages/fluentui/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -145,12 +141,10 @@ const buttonStyles: ComponentSlotStylesPrepared<ButtonStylesProps, ButtonVariabl
             boxShadow: 'none'
           },
 
-          ':focus': {
-            ...borderFocusStyles[':focus']
-          },
-
+          ':focus': borderFocusStyles[':focus'],
```
layershifter: There is no sense in this spread

PR #12254, opened 2020-03-10
Add react-docs package

Shavior:
#### Description of changes

Adds the initial incarnation of the react-docs package to the repo to be consumed for documentation generation.

These changes are to allow for documentation of other component libraries created with fluentui. Previously all documentation was generated with a grunt script that was tied into the build system, this change abstracts that process to a separate package within fluentui.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12254)


packages/fluentui/react-docs/package.json
```diff
@@ -0,0 +1,43 @@
...
+  ],
+  "homepage": "https://github.com/microsoft/fluent-ui-react/tree/master/packages/accessibility",
+  "jsnext:main": "dist/es/index.js",
+  "license": "MIT",
+  "main": "dist/commonjs/index.js",
+  "module": "dist/es/index.js",
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": "microsoft/fluent-ui-react.git",
```
layershifter: Oops, this is wrong

packages/fluentui/react-docs/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "devDependencies": {
+    "@fluentui/internal-tooling": "^0.43.0",
+    "@types/node": "^10.3.2",
+    "lerna-alias": "^3.0.3-0"
+  },
+  "files": [
+    "dist"
+  ],
+  "homepage": "https://github.com/microsoft/fluent-ui-react/tree/master/packages/accessibility",
+  "jsnext:main": "dist/es/index.js",
```
layershifter:
```suggestion
```

We should have there only CommonJS build I think

packages/fluentui/react-docs/package.json
```diff
@@ -0,0 +1,43 @@
...
+  },
+  "devDependencies": {
+    "@fluentui/internal-tooling": "^0.43.0",
+    "@types/node": "^10.3.2",
+    "lerna-alias": "^3.0.3-0"
+  },
+  "files": [
+    "dist"
+  ],
+  "homepage": "https://github.com/microsoft/fluent-ui-react/tree/master/packages/accessibility",
```
layershifter: Oops, should be updated

packages/fluentui/react-docs/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "bugs": "https://github.com/microsoft/fluent-ui-react/issues",
+  "bin": "dist/commonjs/index.js",
+  "dependencies": {
+    "@babel/core": "^7.7.7",
+    "@babel/parser": "^7.6.3",
+    "@babel/preset-typescript": "^7.7.7",
+    "@babel/runtime": "^7.7.6",
+    "@types/yargs": "^15.0.3",
+    "doctrine": "2.0.0",
+    "keyboard-key": "^1.0.1",
```
layershifter:
```suggestion
```

I don't think that we have this dependency

packages/fluentui/react-docs/README.md
```diff
@@ -0,0 +1,84 @@
...
+## Errors
+
+### Error: Could not find a component definition in "YourComponent.tsx".
+
+This error occurs when your component file does not have a recognizable React component exported.
+
+Supported component and export formats:
+
+```jsx
+const Component = React.createClass(<def>);
```
layershifter: Do we want to support deprecated syntax?

packages/fluentui/react-docs/README.md
```diff
@@ -0,0 +1,84 @@
...
+export default Component
+```
+
+**NOT** supported formats
+
+```jsx
+const Component = props => {
+  // ...
+};
+module.exports = Component;
```
layershifter: Is there any difference between previous one?

packages/fluentui/react-docs/src/getComponentInfo.ts
```diff
@@ -0,0 +1,133 @@
...
+  }) as t.File;
+
+  // add exported Component info
+  // this 'require' instruction might break by producing partially initialized types - because of ts-node module's cache used during processing
+  // - in that case we might consider to disable ts-node cache when running this command: https://github.com/ReactiveX/rxjs/commit/2f86b9ddccbf020b2e695dd8fe0b79194efa3f56
+  const Component = require(absPath).default;
+
+  if (!Component) {
+    throw new Error(`Your file with component under "${absPath}" doesn't have a default export`);
+  }
```
layershifter: I would prefer to use there Babel's AST, but it's a separate change

packages/fluentui/react-docs/src/getShorthandInfo.ts
```diff
@@ -0,0 +1,66 @@
+import * as Babel from '@babel/core';
+import { NodePath } from '@babel/traverse';
+import * as t from '@babel/types';
+
+import { ComponentInfo } from '@fluentui/docs/src/types';
```
layershifter: Oops, hidden dependency

packages/fluentui/react-docs/src/index.ts
```diff
@@ -0,0 +1,8 @@
+import * as docgen from './docgen';
+
+export { default as getComponentInfo } from './getComponentInfo';
+export { default as parseDefaultValue } from './parseDefaultValue';
+export { default as parseDocblock } from './parseDocblock';
+export { default as parseType } from './parseType';
```
layershifter: Not sure that this should be exported

packages/fluentui/react-docs/src/parseBuffer.ts
```diff
@@ -0,0 +1,9 @@
+import { parse } from '@babel/parser';
+
+const parseBuffer = buffer =>
```
layershifter: Is this used anywhere?

packages/fluentui/react-docs/src/parseDefaultValue.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    if (_.isPlainObject(defaultValue) || _.isArray(defaultValue)) {
+      return defaultValue;
+    }
+
+    if (_.isNull(defaultValue)) {
+      return null;
+    }
+
+    throw new Error(`Can't parse a value in "${Component.name}.defaultProps.${propDef.name}"`);
```
layershifter: This can also produce error that should be noted on readme

ecraig12345:
@Shavior Could you please update the PR description with more background about what you're doing here? Especially the reason for adding so many duplicated files?

ecraig12345:
Also assuming you're a Microsoft employee, you should link your github account via https://repos.opensource.microsoft.com/people instead of signing the CLA. Linking your account also lets other internal people and tools more easily connect your github account and your real name, for example if we need to contact you.

ecraig12345:
Another thing, any packages that aren't under `packages/fluentui` really should not be using gulp-based build, since the goal is to move away from that. If we're adding some new variation of the fluentui doc site, this would be a good opportunity to standardize its build.

ecraig12345:
Yet another issue, the current component info generation process is **extremely** slow and probably not something we ought to be building new stuff on top of, at least in its current state. I've been investigating ways to improve the performance (such as not parsing each file multiple times; the best would be if we could get rid of `require`-ing the file during the build process) but don't have anything ready quite yet.

packages/fluentui/react-docs/package.json
Shavior: Removed from package json

packages/fluentui/react-docs/package.json
Shavior: Updated repo references.

packages/fluentui/react-docs/package.json
Shavior: Updated repo references.

packages/fluentui/react-docs/README.md
Shavior: Will look into this more. Currently supported because babel is able to properly parse these types of components.

packages/fluentui/react-docs/README.md
Shavior: The previous example uses `( )` while line 71 does not.

packages/fluentui/react-docs/src/getShorthandInfo.ts
Shavior: Removed.

packages/fluentui/react-docs/src/index.ts
Shavior: Slimmed down exports to just required entry points of `getComponentInfo` and `fluentuiSchema`

packages/fluentui/react-docs/src/parseBuffer.ts
Shavior: Removed from package.

packages/fluentui/react-docs/src/parseDefaultValue.ts
Shavior: Error message documented a bit more thoroughly in readme

packages/fluentui/react-docs/src/docgen.ts
```diff
@@ -0,0 +1,730 @@
+import * as fs from 'fs';
+import * as path from 'path';
```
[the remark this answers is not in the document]
Shavior: Removed all references to old docgen file, refactored to use `react-docgen-typescript`

ecraig12345:
Unless @Shavior is actively working on this (in which case we need to talk), this should stay closed--I'm working on an alternate version which will have better perf and address various other concerns.

PR #12257, opened 2020-03-11
Icon roles and naming

### LINKED ISSUE #12256: Icons set aria-label on generic role
### Environment Information

- **Package version(s)**: latest / 7.101

### Describe the issue:

The suggested way to label non-presentational icons is with `aria-label`, but that is currently applied to an `<i>` element without a role. That element maps to a generic role / no role by default, and does not support having an accessible name (https://w3c.github.io/aria/#namefromprohibited).

Image icons with an alt but no aria-label also have an issue -- they're hidden by default even with the `alt`, since `aria-hidden` is applied to its parent.

On the demo page, add `aria-label` to any of the examples: https://fabricweb.z5.web.core.windows.net/oufr/7.101.0/#/examples/icon

#### Actual behavior:

Non-presentational icons do not expose the `img` role, and the name is not consistently announced (or supported by the ARIA spec).

ImageIcons with an `alt` are not exposed

#### Expected behavior:

Meaningful icons are exposed as graphics / `img`'s, with a name defined

Role mapping for `<i>`: https://w3c.github.io/html-aam/#el-i
Roles supporting name from author: https://w3c.github.io/aria/#namefromauthor

My suggested fixes would be:
- adding `role="img"` to non-presentational icons (and dropping `role="presentation"`, since it does nothing when combined with `aria-hidden`)
- for image icons, dropping `aria-hidden` entirely, and relying on the image `alt` attribute

smhigley:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12256

#### Description of changes

- Adds `role="img"` to labelled font icons
- Removes `role="presentation"` from hidden icons (not needed with `aria-hidden`)
- Hides ImageIcons by using an empty `alt` attribute on the `` tag
- Respects the passed `alt` attribute for ImageIcons
- Uses the `alt` attribute for ImageIcons that are given an `aria-label` (more in line with the [first rule of ARIA](https://www.w3.org/TR/using-aria/#rule1))
- Adds some labels to icons in the examples for testing

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12257)


packages/office-ui-fabric-react/src/components/Icon/ImageIcon.tsx
```diff
@@ -14,17 +14,11 @@ export const ImageIcon: React.FunctionComponent<IImageIconProps> = props => {
...
 
   const nativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(props, htmlElementProperties);
-
-  const containerProps = props['aria-label']
-    ? {}
-    : {
-        role: 'presentation',
-        'aria-hidden': imageProps.alt || imageProps['aria-labelledby'] ? false : true
-      };
+  const imageAlt = imageProps['alt'] || props['aria-label'] || '';
```
ecraig12345: 1. When the base Icon is displaying an image, it should follow the same logic for applying `alt` as ImageIcon does (*if* we keep this special logic--see next point; but whatever approach we settle on, it should be the same in both components).
2. I'm not sure that we should automatically transfer `aria-label` to `alt`, since they have some different behaviors (for example, `alt` is used as hover text; possibly other differences too)
3. Can you explain more about the rationale for empty alt text? IIRC this will cause accessibility checkers to flag it.

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.Basic.Example.tsx
```diff
@@ -14,8 +14,8 @@ export const IconBasicExample: React.FunctionComponent = () => {
   // You could also use the standard Icon here.
   return (
     <div>
-      <FontIcon iconName="CompassNW" className={iconClass} />
-      <FontIcon iconName="Dictionary" className={iconClass} />
+      <FontIcon iconName="CompassNW" aria-label="compass" className={iconClass} />
```
ecraig12345: I'm not sure we should add labels to the examples in all cases. Ideally we should find some way of demonstrating that you add an aria-label when the icon has additional meaning which isn't conveyed by adjacent text (and don't necessarily add it if the icon is just presentational).

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.ImageSheet.Example.tsx
```diff
@@ -45,6 +45,7 @@ export const IconImageSheetExample: React.FunctionComponent = () => {
     <div>
       <ImageIcon
         className={classNames.one}
+        aria-label="one note"
```
ecraig12345: If you keep this label, it should be "OneNote"

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
```diff
@@ -48,13 +48,13 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
     const ImageType = (imageLoadError && imageErrorAs) || Image;
 
     // tslint:disable-next-line:deprecation
-    const ariaLabel = this.props['aria-label'] || this.props.ariaLabel;
-    const containerProps = ariaLabel
+    const hasName = this.props['aria-label'] || this.props.ariaLabel || this.props['aria-labelledby'] || imageProps['aria-labelledby'];
```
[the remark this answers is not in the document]
smhigley: @ecraig12345 since `aria-label` only will work/be allowed on `<RootType>` if it also has a `role="img"` (since otherwise it's a generic element, and does not support naming), how do you feel about having the same `isImage` check for it in containerProps?

i.e.:
```
role: isImage ? undefined : 'img',
'aria-label': isImage ? undefined : this.props.ariaLabel
```

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
```diff
@@ -48,13 +48,13 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
...
+    const containerProps = hasName
       ? {
-          'aria-label': ariaLabel
+          role: isImage ? undefined : 'img'
```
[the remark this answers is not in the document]
smhigley: That would be overridden by `{...nativeProps}` coming after {...containerProps}` already, I think. At least that seemed to work when I tried it out :)

packages/office-ui-fabric-react/src/components/Icon/ImageIcon.tsx
smhigley: 1. Done, image alt added to `Icon.base`, thanks for catching that!
2. the `title` attribute will create hover text, but not `alt`; in the case of an `img` tag, the only difference between `alt` and `aria-label` is that an empty `alt` is used to make an image presentational/hidden, and `alt` is more robust. Both are used to calculate the accessible name of an image, and `aria-label` will actually override `alt` if present. Neither have any effect beyond that.
3. An empty `alt` attribute is the canonical way to mark an image as presentational, and is generally preferred to `aria-hidden`, `role="presentation"`, or other methods (see the [first rule of ARIA](https://w3c.github.io/using-aria/#rule1)). Accessibility checkers won't flag it :)

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.Basic.Example.tsx
smhigley: I'm happy to remove it, or open a separate issue/PR about examples and the Do's/Don'ts (I'd love to see a line in there about labeling, if possible 😃).

Is there a way you'd prefer to handle it? Just based on the examples themselves, the icons seem to require a label (they appear neither presentational nor accompanied by text), so that's why I added labels on a couple of them. I'm totally happy to remove/change/open a different PR/do whatever you think best, though.

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
```diff
@@ -54,18 +52,20 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
     const ImageType = (imageLoadError && imageErrorAs) || Image;
 
     // eslint-disable-next-line deprecation/deprecation
-    const ariaLabel = this.props['aria-label'] || this.props.ariaLabel;
-    const containerProps = ariaLabel
+    const accessibleName = imageProps.alt || this.props['aria-label'] || this.props.ariaLabel;
```
ecraig12345, line 55: I think it would be clearer to do this (plus the suggestion a couple lines down, and adding back aria-label to the native props exclude list above)
```suggestion
    const ariaLabel = this.props['aria-label'] || this.props.ariaLabel;
    const accessibleName = imageProps.alt || ariaLabel;
```

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
```diff
@@ -54,18 +52,20 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
...
+    const hasName = !!(accessibleName || this.props['aria-labelledby'] || imageProps['aria-labelledby']);
+    const containerProps = hasName
       ? {
-          'aria-label': ariaLabel,
-        }
+        role: isImage ? undefined : 'img',
+        'aria-label': isImage ? undefined : this.props.ariaLabel
```
ecraig12345, line 60:
```suggestion
        'aria-label': isImage ? undefined : ariaLabel
```

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
```diff
@@ -54,18 +52,20 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
     const ImageType = (imageLoadError && imageErrorAs) || Image;
 
     // eslint-disable-next-line deprecation/deprecation
-    const ariaLabel = this.props['aria-label'] || this.props.ariaLabel;
-    const containerProps = ariaLabel
+    const accessibleName = imageProps.alt || this.props['aria-label'] || this.props.ariaLabel;
+    const hasName = !!(accessibleName || this.props['aria-labelledby'] || imageProps['aria-labelledby']);
```
ecraig12345, line 56: What if `imageProps` has an `aria-label`?

packages/office-ui-fabric-react/src/components/Icon/ImageIcon.tsx
ecraig12345: Even though `alt` and `aria-label` have basically the same effect, I'm not sure we should copy `aria-label` to `alt`. Native props like `aria-label` are typically applied exactly as provided, so moving them around feels like violating consumers' expectations.

Also, I don't think we should apply an empty `alt` (hiding the icon) if it has an `aria-labelledby`?

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.Basic.Example.tsx
ecraig12345: I see what you mean. The examples are a bit unfortunate because they're not the most realistic demonstration of how to use icons. So the ideal would be to switch to still-simple yet more realistic examples (if we can come up with some), but the (probably) easier option would be to leave the labels and add a comment in each example about when you should actually use labels or not.

PR #12258, opened 2020-03-11
Fix unnecessary render of components decorated by withResponsiveMode

KevinTCoughlin:
#### Pull request checklist

#### Description of changes

The default value of `ResponsiveMode` has presented a performance problem for SharePoint Online because it causes an unnecessary render with initial `this.state = { responsiveMode: ResponsiveMode.Large }`. 

image

To mitigate this render, developers have written:

```tsx
setResponsiveMode(ResponsiveMode.unknown);
```

The above removes the unnecessary render pass. However, it introduces another performance problem where the React tree short-circuits on initial commit, introducing artificial delay to what _was_ a synchronous render pass.

**Example from SPO:**

image

This fix exposes an `initializeResponsiveMode` function to be invoked by the host app to set the value for `responsiveMode`, thereby avoiding the unnecessary render (borrowed from #11870).

I've confirmed locally that this fix resolves _both_ performance issues mentioned above.

cc: @johnguy0 @pingj @patmill @ThomasMichon 

#### Focus areas to test

- CI should pass

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12258)

### DIFF
--- a/packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
+++ b/packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
@@ -37,6 +37,19 @@ export function setResponsiveMode(responsiveMode: ResponsiveMode | undefined): v
   _defaultMode = responsiveMode;
 }
 
+/**
+ * Initializes the responsive mode to the current window size. This can be used to avoid
+ * a re-render during first component mount since the window would otherwise not be measured
+ * until after mounting.
+ */
+export function initializeResponsiveMode(element?: HTMLElement): void {
+  if (typeof window !== 'undefined') {
+    const currentWindow = (element && getWindow(element)) || window;
+
+    getResponsiveMode(currentWindow);
+  }
+}
+
 export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveMode }, TState>(
   ComposedComponent: new (props: TProps, ...args: any[]) => React.Component<TProps, TState>
 ): any {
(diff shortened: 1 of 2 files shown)


KevinTCoughlin:
Alternatively, the following PR also fixes this issue: https://github.com/OfficeDev/office-ui-fabric-react/pull/11870.

change/office-ui-fabric-react-2020-03-10-20-59-00-keco-remove-withresponsivemode-default-render.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "Remove default ResponsiveMode causing unnecessary render",
```
aneeshack4: Perhaps saying which utility this is in would clarify this comment in the release notes

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -46,7 +45,7 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
       this._updateComposedComponentRef = this._updateComposedComponentRef.bind(this);
 
       this.state = {
-        responsiveMode: _defaultMode || _lastMode || ResponsiveMode.large
+        responsiveMode: this._getResponsiveMode()
```
ecraig12345: Is there any scenario where any of the values used in getResponsiveMode won't be available yet?

Also, you may want to remove the onResize call in componentDidMount.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -79,8 +78,7 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
 
     private _getResponsiveMode(): ResponsiveMode {
       let responsiveMode = ResponsiveMode.small;
-      const element = findDOMNode(this) as Element;
-      const win = getWindow(element);
+      const win = getWindow();
```
xugao: isn't this a breaking change for multi-window scenario?

KevinTCoughlin: As I called out in my email, yes. I am incorporating changes from #11870 to account for this. Namely, the export of a setter to invoke prior to mount of `withResponsiveMode`.

Do you have alternative recommendations I should consider?

KevinTCoughlin: https://github.com/OfficeDev/office-ui-fabric-react/pull/12258/commits/e6454e02c8a936bbdace936fb4065d2654f52a68 requires the caller to avoid the unnecessary render while still supporting the multi-window scenario.

PR #12266, opened 2020-03-11
fix(Toolbar) :  Toolbar menu item doesn't focus menu trigger element after 'onClick' was executed

kolaps33:
#### Pull request checklist

- [x ] Addresses an existing issue: Fixes https://github.com/microsoft/fluent-ui-react/issues/2249

#### Description of changes
Before we moved focus back on the trigger menu button when we been closing the menu. 
Now we don't handle focus when closing the menu and consumer of toolbar has to handle them self. 

#### Focus areas to test
How to test?:
1. into onClick add logic for any element focus
2. execute click on give item in menu
Expected: focus moved to the element which you defined in "onClick" 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12266)


packages/fluentui/e2e/tests/toolbarMenuOverflow-example.tsx
```diff
@@ -1,45 +1,59 @@
...
   const itemData = _.times(itemsCount, i => ({
     key: `b${i}`,
+    id: `${selectors.itemButtonId}-${i}`,
     content: `${icons[i % icons.length]} #${i}`,
     icon: icons[i % icons.length],
     title: `${icons[i % icons.length]} #${i}`,
-    // first hald of items are unwrapped, rest are wrapped.
-    ...(i >= itemsCount / 2 && { menu: [] })
+    onClick: i + 1 == itemsCount ? () => buttonAfterToolbarRef.current.focus() : undefined,
+    // first hald of items are unwrapped, rest are wrapped, expect of the last item
```
silviuavram: // first half 

That was my comment while working a previous bug, can you fix it? :D

packages/fluentui/e2e/tests/toolbarMenuOverflow-example.tsx
```diff
@@ -1,45 +1,59 @@
...
 
   const itemData = _.times(itemsCount, i => ({
     key: `b${i}`,
+    id: `${selectors.itemButtonId}-${i}`,
     content: `${icons[i % icons.length]} #${i}`,
     icon: icons[i % icons.length],
     title: `${icons[i % icons.length]} #${i}`,
-    // first hald of items are unwrapped, rest are wrapped.
-    ...(i >= itemsCount / 2 && { menu: [] })
+    onClick: i + 1 == itemsCount ? () => buttonAfterToolbarRef.current.focus() : undefined,
```
silviuavram: why not `===` ?

silviuavram: probably `? buttonAfterToolbarRef.current.focus` will also work instead of `() => buttonAfterToolbarRef.current.focus()`

packages/fluentui/e2e/tests/toolbarMenu-example.tsx
```diff
@@ -27,7 +29,13 @@ const ToolbarExampleMenuShorthand = () => {
             id: selectors.triggerButtonId,
             menu: {
               items: [
-                { key: 'play', content: 'Play', icon: 'play' },
+                {
+                  key: 'play',
+                  id: `${selectors.menuItemButtonId}-0`,
+                  content: 'Play',
+                  icon: 'play',
+                  onClick: () => buttonAfterToolbarRef.current.focus()
```
silviuavram: will `onClick: buttonAfterToolbarRef.current.focus` work?

kolaps33: image
when I changed it doesn't look promising...

packages/fluentui/e2e/tests/toolbarMenuOverflow-example.tsx
kolaps33: sure :) done

packages/fluentui/e2e/tests/toolbarMenuOverflow-example.tsx
kolaps33: Thank you for comparing catch, changed for === 
"buttonAfterToolbarRef.current.focus" was the same issue as in previous comment  that editor complains...

packages/office-ui-fabric-react/src/utilities/decorators/withViewport.tsx
```diff
@@ -121,7 +121,7 @@ export function withViewport<TProps extends { viewport?: IViewport }, TState>(
```
kolaps33: I have trouble with this file, somehow it changed and I cannot revert the change...

kolaps33: done changes online, this helps

packages/fluentui/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -188,7 +188,6 @@ const ToolbarItem: React.FC<WithAsProp<ToolbarItemProps>> &
     if (menu) {
       if (doesNodeContainClick(menuRef.current, e.nativeEvent as MouseEvent, context.target)) {
         trySetMenuOpen(false, e);
-        _.invoke(itemRef.current, 'focus');
```
layershifter: Should we remove the same thing in `closeMenuAndFocusTrigger`?

packages/fluentui/e2e/tests/toolbarMenu-example.tsx
jurokapsiar: _type_script :)

packages/fluentui/react/src/components/Toolbar/ToolbarItem.tsx
jurokapsiar: no, if that is the one used for handling ESC key, we need to focus trigger in that case

kolaps33: to @jurokapsiar: yes "closeMenuAndFocusTrigger" is calling when ESC key is executed :)

PR #12268, opened 2020-03-11
Styling: fix sideEffects scope in package.json

xugao:
#### Pull request checklist

#### Description of changes
It's unclear to me why we had `sideEffects: true` for styling package. trying to scope down to reduce bundle size. let's see if any test fails 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12268)

### DIFF
--- a/packages/styling/package.json
+++ b/packages/styling/package.json
@@ -9,7 +9,9 @@
   "license": "MIT",
   "main": "lib-commonjs/index.js",
   "module": "lib/index.js",
-  "sideEffects": true,
+  "sideEffects": [
+    "lib/version.js"
+  ],
   "typings": "lib/index.d.ts",
   "scripts": {
     "build": "just-scripts build",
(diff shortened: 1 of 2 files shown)


packages/styling/package.json
```diff
@@ -9,7 +9,9 @@
   "license": "MIT",
   "main": "lib-commonjs/index.js",
   "module": "lib/index.js",
-  "sideEffects": true,
+  "sideEffects": [
```
ecraig12345: @kenotron do you know more about the background for having sideEffects: true here?

dzearing: I do; the concern was mainly fontFace registration getting dropped. Working with Xu on this.

PR #12270, opened 2020-03-11
FocusZone: Using keyboard-key library in v7 to align it closer to v0

khmakoto:
#### Pull request checklist

#### Description of changes

This PR replaces the references to `ev.which` and `KeyCodes` in v7's version of `FocusZone` in favor of the `keyboard-key` library that is being used already in v0. This allows us to bring the two versions closer together.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12270)


packages/react-focus/package.json
```diff
@@ -51,6 +51,7 @@
     "@uifabric/set-version": "^7.0.6",
     "@uifabric/styling": "^7.10.17",
     "@uifabric/utilities": "^7.13.2",
+    "keyboard-key": "1.0.1",
```
ecraig12345: You should update everyone to 1.1.0 which has types. Trouble is that the types only work properly if you have `esModuleInterop` enabled, which fluent doesn't (Levi or Shift could give details of why). The current typing format is necessary to properly reflect the code's assignment to `module.exports`...though come to think of it, you could just update keyboard-key to a more interop-friendly export format. It's here: https://github.com/levithomason/keyboard-key/blob/master/src/keyboardKey.js

dzearing:
OOF on bundle size...

packages/react-focus/package.json
khmakoto: I agree, we should update everyone. @layershifter and @levithomason why doesn't Fluent have `esModuleInterop`?

ecraig12345: There's more background here (`allowSyntheticDefaultImports` and `esModuleInterop` are closely related): https://github.com/microsoft/fluent-ui-react/pull/1395

layershifter:
I would say that `keyboard-key` should be refactored to ES modules as now it will always import everything...

khmakoto:
Better but still not great:

__Before:__
Before

__After:__
After

khmakoto:
Closing in favor of #12339. I'll copy [Levi's comment](https://github.com/OfficeDev/office-ui-fabric-react/pull/12270#pullrequestreview-377336274) as TODOs in the description there.

PR #12277, opened 2020-03-12
feat(Avatar): added square prop

mnajdova:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12277)

### DIFF
--- a/packages/fluentui/react/src/themes/teams/components/Avatar/avatarStyles.ts
+++ b/packages/fluentui/react/src/themes/teams/components/Avatar/avatarStyles.ts
@@ -3,7 +3,7 @@ import { ComponentSlotStylesPrepared, ICSSInJSStyle } from '@fluentui/styles';
 import { AvatarProps } from '../../../../components/Avatar/Avatar';
 import { AvatarVariables } from './avatarVariables';
 
-export type AvatarStylesProps = Pick<AvatarProps, 'size'>;
+export type AvatarStylesProps = Pick<AvatarProps, 'size' | 'square'>;
 
 const sizeToPxValue = {
   smallest: 24,
@@ -28,7 +28,7 @@ const avatarStyles: ComponentSlotStylesPrepared<AvatarStylesProps, AvatarVariabl
       width: sizeInRem
     };
   },
-  image: ({ variables: v }): ICSSInJSStyle => ({
+  image: ({ props: p, variables: v }): ICSSInJSStyle => ({
     borderColor: v.avatarBorderColor,
     borderStyle: 'solid',
     borderWidth: v.avatarBorderWidth,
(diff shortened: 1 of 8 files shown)


packages/fluentui/react/src/themes/teams/components/Avatar/avatarStyles.ts
```diff
@@ -3,7 +3,7 @@ import { ComponentSlotStylesPrepared, ICSSInJSStyle } from '@fluentui/styles';
 import { AvatarProps } from '../../../../components/Avatar/Avatar';
 import { AvatarVariables } from './avatarVariables';
 
-export type AvatarStylesProps = Pick<AvatarProps, 'size'>;
+export type AvatarStylesProps = Pick<AvatarProps, 'size' | 'square'>;
```
layershifter: Let's move this to the component as we are here 🙂

packages/fluentui/react/src/themes/teams/components/Avatar/avatarStyles.ts
```diff
@@ -28,27 +28,34 @@ const avatarStyles: ComponentSlotStylesPrepared<AvatarStylesProps, AvatarVariabl
...
       height: sizeInRem,
       lineHeight: sizeInRem,
-      fontSize: pxToRem(sizeToPxValue[size] / 2.333),
+      fontSize: pxToRem(sizeToPxValue[p.size] / 2.333),
       verticalAlign: 'top',
       textAlign: 'center',
-      padding: '0px'
+      padding: '0px',
+      ...(p.square && {
+        borderRadius: v.avatarBorderRadius
```
layershifter:
```suggestion
        borderRadius: v.squareBorderRadius
```

Should not contain prop's name?

packages/fluentui/react/src/themes/teams/components/Avatar/avatarVariables.ts
```diff
@@ -1,13 +1,17 @@
...
   avatarBorderWidth: string;
+  squareAvatarBorderRadius: string;
   statusBorderColor: string;
   statusBorderWidth: string;
 }
 
 export default (siteVariables): AvatarVariables => ({
   avatarBorderColor: '',
   avatarBorderWidth: '0',
+  squareAvatarBorderRadius: pxToRem(4),
```
layershifter: Just wondering why `squareAvatarBorderRadius`? As I remember convention is `${PROP}${SLOT}${RULE}` 🤔

mnajdova: This component is awkward as there is a root, and than avatar and status part. The avatar part can be image, label, soon also icon, I guess that's why it was chosen to be like this... I am just following previous convention, didn't want to change all variables names..

PR #12284, opened 2020-03-12
[feat] Moving keyboard-key into repo

dzearing:
The `keyboard-key` library is currently used in v0 code. There is a [pull request](https://github.com/OfficeDev/office-ui-fabric-react/pull/12270) to include it in the `react-focus` package, but the PR is regressing bundle size for all downstream components by 3k.

We'd like to move it into the repo to manage, build es6 output, convert it to typescript, reduce bundle size, and make any improvements inside the repo.

This PR does that:

✅Moved keyboard-key in
✅Converted to typescript with full typings
✅Reduced bundle size; removed static redundant definitions, replaced with looping
✅all tests pass

Once this goes in we can:
- [ ] Consume it by the `react-focus` PR
- [ ] Evaluate bundle improvements
- [ ] Get it published and consumed
- [ ] Remove all `KeyCodes` usage
- [ ] Deprecate `KeyCodes`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12284)


dzearing:
@JasonGore @miroslavstastny I think the perf test might be too touchy. Maybe we need more iterations

JasonGore:
I already started a discussion in the Shield channel. @aneeshack4 is going to fix them. 😄

JasonGore:
Perf iterations bumped up in #12300

packages/keyboard-key/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@fluentui/keyboard-key",
+  "version": "0.1.0",
+  "description": "A simple utility for determining the KeyboardEvent.key property from a keyboard event.",
+  "main": "lib-commonjs/index.js",
+  "module": "lib/index.js",
+  "typings": "lib/index.d.ts",
+  "sideEffects": [
```
ecraig12345: This can be removed unless you add back the `@uifabric/set-version` dep (which I don't totally understand why we have that everywhere?)

khmakoto:
@ecraig12345 @kenotron @dzearing is there any reason to not merge this as is?

packages/keyboard-key/src/index.ts
```diff
@@ -0,0 +1,418 @@
...
+keyboardKey.SingleQuote = keyboardKey["'"];
+keyboardKey.DoubleQuote = keyboardKey['"'];
+keyboardKey.LeftCurlyBrace = keyboardKey['{'];
+keyboardKey.RightCurlyBrace = keyboardKey['}'];
+keyboardKey.LeftParenthesis = keyboardKey['('];
+keyboardKey.RightParenthesis = keyboardKey[')'];
+keyboardKey.LeftAngleBracket = keyboardKey['<'];
+keyboardKey.RightAngleBracket = keyboardKey['>'];
+keyboardKey.LeftSquareBracket = keyboardKey['['];
+keyboardKey.RightSquareBracket = keyboardKey[']'];
```
layershifter: Too late, but this is not treeshakable as it's statics 😭 

But even with `export const RightSquareBracket` will not help much as `getCode()` and `getKey()` functions rely either on `this` or on `codes`.

So `2.4kB` is a price for this module: https://bundlephobia.com/result?p=@fluentui/keyboard-key@0.1.1

ecraig12345: Is there any way we could break up the functionality into smaller pieces so some of it is tree-shakeable? Or are there any parts of this that could optionally rely on types for compile-time-only verification? (runtime versions would also be available, but could be tree-shaken out if a TS project didn't use them)

PR #12289, opened 2020-03-12
perf(fela): use Stylis for prefixing

layershifter:
#### Pull request checklist

# What ❓ 

This PR introduces [Stylis](https://github.com/thysultan/stylis.js) for vendor prefixing job instead of [`fela-plugin-prefixer`](https://github.com/robinweser/fela/tree/master/packages/fela-plugin-prefixer) to get performance benefits.

# BREAKING CHANGES

### [CSS Grid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout) is not prefixed anymore

Our previous prefixing was not ideal, too. The biggest bummer here that not all features can be resolved with prefixing for IE11, especially with `auto` sizing, i.e. each case with CSS Grid should be handled manually even now.

#### Before

```jsx
const styles = {
  display: 'grid',
  gridTemplateColumns: 'auto 1fr',
}
```

#### After

```js
const styles = {
  display: ['grid', '-ms-grid'],
  gridTemplateColumns: 'auto 1fr',
  msGridColumns: 'auto 1fr',
}
```

[Autoprefixer playground](https://autoprefixer.github.io/) can be also useful.

### [`linear-gradient()`](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient#Browser_compatibility) is not prefixed anymore

- works in IE11 by default
- works in from Chrome 26 (2013-03-26) & Firefox 16 (2012-06-05)

To archive previous behavior usage fallback values feature:

```js
const styles = {
  background: [
    'linear-gradient(#e66465, #9198e5)',
    '-webkit-linear-gradient(#e66465, #9198e5)',
    '-moz-linear-gradient(#e66465, #9198e5)'
  ]
}
```

### [ `background-clip`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip#Browser_compatibility) is not prefixed anymore

Previous prefixing was not correct:

```css
background-clip: -webkit-text;	
background-clip: text;
```

Should be:
```css
background-clip: text;
-webkit-background-clip: text;
```

### [ `cross-fade()`](https://developer.mozilla.org/en-US/docs/Web/CSS/cross-fade#Browser_compatibility) is not prefixed anymore

It's not supported in IE11.
Works only with a prefix in Chrome.

# Performance 🐎 

It's faster than by Fela nature as the plugin operates with `renderRule()`: Fela will run plugins on every `renderRule()` call, but, `_emitChange()` will be called only for new CSS rules:

image

It will be easier to understand with CodeSandbox that shows this: https://codesandbox.io/s/vigorous-fire-jhzkq

image

### Before

| Example                  | min    | avg    | median | max    |
| ------------------------ | ------ | ------ | ------ | ------ |
| ChatWithPopover.perf.tsx | 396.11 | 435.93 | 427.42 | 556.65 |
| ChatWithPopover.perf.tsx | 393.02 | 434.84 | 419.76 | 570.68 |
| ChatWithPopover.perf.tsx | 395.83 | 432.41 | 424.12 | 500.64 |

### After 🔽 ~28%

| Example                  | min    | avg    | median | max    |
| ------------------------ | ------ | ------ | ------ | ------ |
| ChatWithPopover.perf.tsx | 299.91 | 311.17 | 310.18 | 329.78 |
| ChatWithPopover.perf.tsx | 301.28 | 316.03 | 311.39 | 438.31 |
| ChatWithPopover.perf.tsx | 301.03 | 311.79 | 311.29 | 340.73 |

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12289)


packages/fluentui/CHANGELOG.md
```diff
@@ -43,6 +43,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Expose `contentWrapper` shorthand from `AccordionTitle` @silviuavram ([#12265](https://github.com/OfficeDev/office-ui-fabric-react/pull/12265))
 - Add `disabled` versions for `Dropdown` and `Input` @silviuavram ([#12250](https://github.com/OfficeDev/office-ui-fabric-react/pull/12250))
 
+### Performance
```
miroslavstastny: Also add an entry to breaking changes

PR #12295, opened 2020-03-13
BaseComponent: remove using BaseComponent from oufr package

xugao:
#### Pull request checklist

#### Description of changes
Removing usage of `BaseComponent` to help with:
1) Bundle size
2) Converting to FunctionComponent

In this PR, I only removed usage from `office-ui-fabric-react` package. I will have a separate change to remove everywhere else (24 files) and completely deprecate `BaseComponent`

In order to remove `extends BaseComponent`, I did the following:
1. removed `initializeDir` logic completely as I don't believe it's needed
2. call `initializeComponentRef` instead in constructors
3. call `warnDeprecations` and `warnMutuallyExclusive` directly
4. create `Async` and `EventGroup` where it's being used and make sure they are disposed in `componentWillUnmount` (related work: replace EventGroup with EventListener. e.g. #11961)
5. remove usage of `_resolveRef` and use `createRef` instead
6. make sure `FocusRects` are added (only if it's needed) in replacement of `initializeFocusRects` 
(related PR: #12244)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12295)


packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
```diff
@@ -751,7 +772,7 @@ export class BaseButton extends BaseComponent<IBaseButtonProps, IBaseButtonState
     if (!ev.defaultPrevented && this._isValidMenuOpenKey(ev)) {
       const { onMenuClick } = this.props;
       if (onMenuClick) {
-        onMenuClick(ev, this);
+        onMenuClick(ev, this.props);
```
dzearing: jinkies... seems like a bad. bug.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -306,10 +316,8 @@ export class ComboBox extends BaseComponent<IComboBoxProps, IComboBoxState> {
   }
 
   public componentWillUnmount(): void {
-    super.componentWillUnmount();
-
-    // remove the eventHanlder that was added in componentDidMount
-    this._events.off(this._comboBoxWrapper.current);
```
khmakoto: Is it safe to remove this?

packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
```diff
@@ -4,7 +4,6 @@ import { Label } from '../../Label';
 import { Icon } from '../../Icon';
 import {
   initializeComponentRef,
-  FocusRects,
```
khmakoto: Why are we removing `FocusRects` here?

packages/office-ui-fabric-react/src/components/pickers/PeoplePicker/PeoplePickerItems/SelectedItemWithMenu.tsx
```diff
@@ -1,70 +0,0 @@
...
-export interface IPeoplePickerItemWithMenuState {
-  contextualMenuVisible: boolean;
-}
-
-/**
- * PeoplePickerItem with an additional contextual menu.
- * @deprecated Do not use. Will be removed in Fabric 7.0
- */
-// tslint:disable-next-line:deprecation
-export class SelectedItemWithMenu extends BaseComponent<IPeoplePickerItemWithMenuProps, IPeoplePickerItemWithMenuState> {
```
khmakoto: Is it ok to just remove this?

xugao: the comment says:  `@deprecated Do not use. Will be removed in Fabric 7.0`. 
so i assumed so

packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
xugao: Because it's needed. the focus style is done on focus. doesn't have to be keyboard focus. 
in general , i don't want to add `FocusRects` if not needed. that's why you don't see it everywhere where i removed `BaseComponent`

packages/office-ui-fabric-react/src/components/pickers/PeoplePicker/PeoplePickerItems/SelectedItemWithMenu.tsx
khmakoto: I mean, we did add that comment, but we didn't remove it in 7 so I wonder if we should wait till 8. We have things that say "to be removed in Fabric 5" that are still there.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
KevinTCoughlin: `EventGroup.dispose` invokes `EventGroup.off` so it should be OK

packages/office-ui-fabric-react/src/components/FloatingPicker/Suggestions/SuggestionsCore.tsx
```diff
@@ -9,14 +9,16 @@ const styles: any = stylesImport;
 /**
  * Class when used with SuggestionsStore, renders a basic suggestions control
  */
-export class SuggestionsCore<T> extends BaseComponent<ISuggestionsCoreProps<T>, {}> {
+export class SuggestionsCore<T> extends React.Component<ISuggestionsCoreProps<T>, {}> {
   public currentIndex: number;
   public currentSuggestion: ISuggestionModel<T> | undefined;
-  protected _selectedElement: HTMLDivElement;
+  protected _selectedElement = React.createRef<HTMLDivElement>();
```
KevinTCoughlin: Thanks for these! They will help with memory management :).

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
xugao: yes calling `dispose` does the same

packages/office-ui-fabric-react/src/components/pickers/PeoplePicker/PeoplePickerItems/SelectedItemWithMenu.tsx
xugao: sure i can revert just to be safe

xugao: `SelectedItemWithMenu` is not a publicly facing component (not exported in our index.js).
and it's not used anywhere internally. i think we are good to remove!

ecraig12345:
@xugao Can you also get rid of `IBaseProps` usage at some point? (Basically means manually adding the componentRef prop various places.) While it doesn't contribute to bundle size, it's odd to still use that when we're not using BaseComponent.

PR #12300, opened 2020-03-13
fix(performance): increase iterations for potential perf regressions

aneeshack4:
#### Pull request checklist

#### Description of changes

The PRs in [this list](https://github.com/OfficeDev/office-ui-fabric-react/pulls?q=is%3Apr+is%3Aopen+%22potential+regressions%22) have potential regressions that are false - only failing because they didn't run for enough iterations. Went through each of the scenarios that were supposedly regressing and if they were already hitting 1000 ticks, I ignored them but if they weren't I increased the iteration count.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12300)

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Dropdown/Performance/DropdownManyItems.perf.tsx
+++ b/packages/fluentui/docs/src/examples/components/Dropdown/Performance/DropdownManyItems.perf.tsx
@@ -10,6 +10,7 @@ const inputItems = _.times(300, (i: number) => ({
 
 const DropdownManyItemsPerf = () => <Dropdown defaultOpen items={inputItems} placeholder="Select your hero" />;
 
+DropdownManyItemsPerf.iterations = 5000;
 DropdownManyItemsPerf.filename = 'DropdownManyItems.perf.tsx';
 
 export default DropdownManyItemsPerf;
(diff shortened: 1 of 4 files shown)


packages/fluentui/docs/src/examples/components/Dropdown/Performance/DropdownManyItems.perf.tsx
```diff
@@ -10,6 +10,7 @@ const inputItems = _.times(300, (i: number) => ({
 
 const DropdownManyItemsPerf = () => <Dropdown defaultOpen items={inputItems} placeholder="Select your hero" />;
 
+DropdownManyItemsPerf.iterations = 1000;
```
JasonGore: I forgot that defaultIterations for Fluent perf tests is 1. This test is already rendering 300 items per iteration (line 5 above), so the most iterations we should have is 3 or 5.

aneeshack4: done in the latest commit, thank you!

PR #12302, opened 2020-03-13
Renaming @fluentui/react to @fluentui/react-northstar

dzearing:
Setting out to address these goals for customers:

* Know which component package to use when (@fluentui/react@7 vs office-ui-fabric-react vs@fluentui/react@0 is confusing. Instead we will have `@fluentui/react`, `@fluentui/react-next`, and `@fluentui/react-northstar`.)

* Ensure that the current production LTS channel can live side by side with "next" and "northstar" code.

To meet these needs, we will be renaming the `@fluentui/react` package to `@fluentui/react-northstar`.

### BREAKING CHANGE

This means that `@fluentui/react` package imports would be changed to `@fluentui/react-northstar`. It is possible however to use new versions of the packages without changing imports via npm aliasing:

```bash
npm install @fluentui/react@npm:@fluentui/react-northstar
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12302)


packages/fluentui/docs/src/pages/Composition.mdx
```diff
@@ -46,10 +45,10 @@ Our components handle only props that are defined in their interfaces, all unhan
...
 ```jsx
-import { Link } from 'react-router-dom'
+import { Link } from 'react-router-dom';
 
 // 💡 `to` prop is not handled in `Button` and will be passed to `Link` component
-;<Button as={Link} to="/home">
+<Button as={Link} to="/home">
   To homepage
-</Button>
+</Button>;
```
miroslavstastny: nit: Not sure about this semicolon

dzearing: good catch!

ecraig12345:
What's the rationale for `react-future` instead of `react-experimental` or other names that were discussed?

dzearing:
@miroslavstastny I believe we are ready to go here; can you help with merge/publish/update teams?

I will start on the oufr aliasing pr next.

PR #12312, opened 2020-03-16
chore(getStyles): allow to handle multiple displayNames

layershifter:
#### Pull request checklist

#### Description of changes

Extracted from #12239.

## `resolveVariables` strategy

This features is implemented for upcoming `compose()` functionality, to implement caching in `resolveVariables` we discussed with @mnajdova two possible options.

`componentVariables.ts` are represented by an object:
```tsx
const componentVariables = {
  Button: buttonVariables,
  PrimaryButton: primaryButtonVariables,
}
```

#### Option 1: cache separately

I initially tried to cache `variables` for each component:

```tsx
const cache = {
  Button: buttonCache,
  PrimaryButton: buttonPrimaryCache,
}
const resolvedVariables = mergeVariables(buttonCache, buttonPrimaryCache)
```

But it has a downside as we will need always use `mergeVariables()` even on cached ones.

#### Option 2: cache all

```tsx
const cache = {
   Button: buttonCache,
   PrimaryButton: mergeVariables(buttonCache, buttonPrimaryCache)
}
```

We decided to go with this option to avoid merging. As downside there will be higher memory usage.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12312)


packages/fluentui/react-bindings/src/styles/getStyles.ts
```diff
@@ -22,18 +22,17 @@ const getStyles = (options: ResolveStylesOptions): GetStylesResult => {
...
     options.performance.enableVariablesCaching
   );
-
   const { classes, resolvedStyles, resolvedStylesDebug } = resolveStyles(options, resolvedVariables);
 
   // conditionally add sources for evaluating debug information to component
   if (process.env.NODE_ENV !== 'production' && isDebugEnabled) {
     options.saveDebug({
-      componentName: options.displayName,
+      componentName: options.displayNames[0],
```
mnajdova: This should probably be composed or all displayNames, or the last one I guess..

packages/fluentui/react-bindings/src/styles/resolveVariables.ts
```diff
@@ -10,33 +10,50 @@ import {
...
+    // const displayNames = { Foo: variables, Bar: undefined, Baz: undefined }
+    // This allows to avoid creating useless cache entries for `Bar` & `Baz` components
+    const handlingDisplayName = effectiveDisplayNames[effectiveDisplayNames.length - 1];
+
+    if (!variablesThemeCache[handlingDisplayName]) {
+      effectiveDisplayNames.map(displayName => theme.componentVariables[displayName]);
+
+      variablesThemeCache[handlingDisplayName] = mergeComponentVariables(
+        ...effectiveDisplayNames.map(displayName => theme.componentVariables[displayName])
+      )(theme.siteVariables);
```
mnajdova: May be a premature optimization, but maybe we can reuse here prev caching if we find for some other displayName..

packages/fluentui/react-bindings/src/styles/resolveVariables.ts
```diff
@@ -10,33 +10,50 @@ import {
...
     }
 
-    componentThemeVariables = variablesThemeCache[displayName];
+    componentThemeVariables = variablesThemeCache[handlingDisplayName];
   } else {
-    componentThemeVariables = callable(theme.componentVariables[displayName])(theme.siteVariables) || {};
+    if (effectiveDisplayNames.length === 1) {
+      componentThemeVariables = callable(theme.componentVariables[effectiveDisplayNames[0]])(theme.siteVariables) || {};
+    } else {
+      componentThemeVariables = mergeComponentVariables(...effectiveDisplayNames.map(displayName => theme.componentVariables[displayName]))(
```
mnajdova: We should be able to just take this from the cache. No need for merging again as far as I can see

packages/fluentui/react/src/components/Animation/Animation.tsx
```diff
@@ -187,7 +187,7 @@ const Animation: React.FC<AnimationProps> & {
 
     return getStyles({
       className: Animation.className,
-      displayName: Animation.displayName,
+      displayNames: [Animation.displayName],
```
mnajdova: Should we have both options: displayName and displayNames? May complicate implementation, but for cases like this make sense... I'd be ok with any outcome..

packages/fluentui/react-bindings/test/styles/resolveVariables-test.ts
```diff
@@ -1,84 +1,120 @@
...
+      const fooVariables = jest.fn().mockReturnValue({ content: 'foo' });
+      const theme = createTheme({ Foo: fooVariables, Bar: undefined });
+
+      expect(resolveVariables(['Foo', 'Bar'], theme, {}, true)).toMatchObject({ content: 'foo' });
+      expect(fooVariables).toHaveBeenCalledTimes(1);
+    });
+
+    test('handles multiple displayNames', () => {
+      const fooVariables = jest.fn().mockReturnValue({ backgroundColor: 'blue' });
+      const barVariables = jest.fn().mockReturnValue({ color: 'red' });
```
mnajdova: Would suggest adding a scenario with ovedrriding variabels too.

packages/fluentui/react/src/components/Animation/Animation.tsx
layershifter: It's our internal API, so I would say that it should be more strict

packages/fluentui/react-bindings/src/styles/resolveVariables.ts
layershifter: This brunch is out of caching:

packages/fluentui/react-bindings/src/styles/resolveVariables.ts
layershifter: Can you please add a code snippet? Not sure that I get your idea 🤔

packages/fluentui/react-bindings/src/styles/getStyles.ts
layershifter: I replaced it with `componentName: options.displayNames.join(':'),`. Will it be enough for now?

packages/fluentui/react-bindings/test/styles/resolveVariables-test.ts
layershifter: I changed this test to handle it:

```tsx
const fooVariables = jest
  .fn()
  .mockReturnValue({ backgroundColor: "blue", borderColor: "black" });
const barVariables = jest
  .fn()
  .mockReturnValue({ backgroundColor: "green", color: "red" });
const theme = createTheme({ Foo: fooVariables, Bar: barVariables });

expect(resolveVariables(["Foo", "Bar"], theme, {}, true)).toMatchObject({
  backgroundColor: "blue",
  borderColor: "black",
  color: "red"
});
```

packages/fluentui/react/src/components/Animation/Animation.tsx
mnajdova: Got it, thought is `useStyles`, not I see that it is `getStyles`. Makes sense!

packages/fluentui/react-bindings/src/styles/resolveVariables.ts
```diff
@@ -10,33 +10,50 @@ import {
...
     const variablesThemeCache = variablesCache.get(theme) || {};
 
-    if (!variablesThemeCache[displayName]) {
-      variablesThemeCache[displayName] = callable(theme.componentVariables[displayName])(theme.siteVariables) || {};
+    // const displayNames = { Foo: variables, Bar: undefined, Baz: undefined }
+    // This allows to avoid creating useless cache entries for `Bar` & `Baz` components
+    const handlingDisplayName = effectiveDisplayNames[effectiveDisplayNames.length - 1];
+
+    if (!variablesThemeCache[handlingDisplayName]) {
+      effectiveDisplayNames.map(displayName => theme.componentVariables[displayName]);
```
mnajdova: Do we need this? We have the same on line 43, not sure this is even working..

packages/fluentui/react-bindings/src/styles/resolveVariables.ts
mnajdova: NVM, we need to merge those, we cannot reuse prev results.

packages/fluentui/docs/src/examples/components/Provider/Performance/ProviderMergeThemes.perf.tsx
```diff
@@ -10,7 +10,6 @@ const providerMergeThemesPerf = () => {
   const resolvedStyles = _.mapValues(merged.componentStyles, (componentStyle, componentName) => {
     const compVariables = _.get(merged.componentVariables, componentName, callable({}))(merged.siteVariables);
     const styleParam: ComponentStyleFunctionParam = {
-      displayName: componentName,
```
mnajdova: ❤️

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
```diff
@@ -105,7 +108,7 @@ const resolveStyles = (
   const felaParam: RendererParam = {
     theme: { direction },
     disableAnimations,
-    displayName, // does not affect styles, only used by useEnhancedRenderer in docs
+    displayName: displayNames[0], // does not affect styles, only used by useEnhancedRenderer in docs
```
mnajdova:
```suggestion
    displayName: displayNames.join(':'), // does not affect styles, only used by useEnhancedRenderer in docs
```

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
```diff
@@ -127,7 +130,7 @@ const resolveStyles = (
   const propsCacheKey = cacheEnabled ? JSON.stringify(stylesProps) : '';
   const variablesCacheKey = cacheEnabled && performance.enableBooleanVariablesCaching ? JSON.stringify(variables) : '';
   const componentCacheKey = cacheEnabled
-    ? `${displayName}:${propsCacheKey}:${variablesCacheKey}:${styleParam.rtl}${styleParam.disableAnimations}`
+    ? `${displayNames.join(':')}:${propsCacheKey}:${variablesCacheKey}:${styleParam.rtl}${styleParam.disableAnimations}`
```
mnajdova: nit: maybe extract this in a utility function, or at least store the separator in a const.

packages/fluentui/react-bindings/test/styles/resolveVariables-test.ts
```diff
@@ -1,84 +1,124 @@
...
       backgroundColor: 'green'
     };
-    const variables = resolveVariables('Test', createTheme('Test'), propsVariables, false);
+    const variables = resolveVariables(['Test'], createTheme(), propsVariables, false);
     expect(variables).toMatchObject({ backgroundColor: 'green' });
   });
 
-  test('caches resolved variables', () => {
-    const componentVariablesMock = jest.fn().mockReturnValue({ backgroundColor: 'blue' });
+  test('avoid merging if variables are undefined', () => {
```
mnajdova:
```suggestion
  test('avoids merging if variables are undefined', () => {
```

packages/fluentui/react-bindings/test/styles/resolveVariables-test.ts
```diff
@@ -1,84 +1,124 @@
...
-  test('considers displayName while caching resolved variables', () => {
-    const componentVariablesMock1 = jest.fn().mockReturnValue({ backgroundColor: 'blue' });
-    const componentVariablesMock2 = jest.fn().mockReturnValue({ color: 'red' });
-    const theme = createTheme();
-    theme.componentVariables['Test1'] = componentVariablesMock1;
-    theme.componentVariables['Test2'] = componentVariablesMock2;
-
-    const variables = resolveVariables('Test1', theme, {}, true);
-    const secondVariables = resolveVariables('Test2', theme, {}, true);
+  test('avoid merging for multiple display names if variables are undefined', () => {
```
mnajdova:
```suggestion
  test('avoids merging for multiple display names if variables are undefined', () => {
```

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
layershifter: I am not sure that it will make code there more readable, so will skip this change for now. Obviously we will back there and can reconsider it 👍

PR #12313, opened 2020-03-16
chore(scripts): fix "jest:watch" command

layershifter:
#### Pull request checklist

#### Description of changes

This PR updates existing Gulp tasks to fix them.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12313)

### DIFF
--- a/packages/fluentui/react/gulpfile.ts
+++ b/packages/fluentui/react/gulpfile.ts
@@ -5,4 +5,4 @@ import '../../../gulpfile';
 
 // Redefine test and test:watch to build info json files beforehand
 task('test', series('build:docs:component-info', 'test:jest'));
-task('test:watch', series('build:docs:component-info', 'test:jest', parallel('test:jest:watch', 'watch:docs:component-info')));
+task('test:watch', series('build:docs:component-info', parallel('test:jest:watch', 'watch:docs:component-info')));
--- a/scripts/gulp/tasks/test-unit.ts
+++ b/scripts/gulp/tasks/test-unit.ts
@@ -9,11 +9,13 @@ const argv = yargs
   .option('runInBand', {})
   .option('maxWorkers', {})
   .option('detectLeaks', {})
+  .option('coverage', { default: true })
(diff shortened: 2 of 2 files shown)


packages/fluentui/react/gulpfile.ts
```diff
@@ -5,4 +5,4 @@ import '../../../gulpfile';
 
 // Redefine test and test:watch to build info json files beforehand
 task('test', series('build:docs:component-info', 'test:jest'));
-task('test:watch', series('build:docs:component-info', 'test:jest', parallel('test:jest:watch', 'watch:docs:component-info')));
```
layershifter: We don't need to run `test:jest` before `test:jest:watch`

scripts/gulp/tasks/test-unit.ts
```diff
@@ -53,5 +54,5 @@ if (config.isRoot) {
   task('test:watch', series('test:jest:setup', parallel('test:jest:watch', 'watch:docs:component-info')));
 } else {
   task('test', series('test:jest'));
-  task('test:watch', series('test:jest', 'test:jest:watch'));
+  task('test:watch', series('test:jest:watch'));
```
layershifter: Same thing here

scripts/gulp/tasks/test-unit.ts
```diff
@@ -9,11 +9,13 @@ const argv = yargs
   .option('runInBand', {})
   .option('maxWorkers', {})
   .option('detectLeaks', {})
+  .option('coverage', { default: true })
```
layershifter: It was really annoying to run tests without coverage metrics

silviuavram: I personally don't care to see by default coverage when I run tests, so I would rather have `yarn test:cover` separately. But that's my preference.

PR #12318, opened 2020-03-16
Presence color props

andrehag:
#### Pull request checklist

#### Description of changes
We have a case in Microsoft Teams for the LivePersonaCard where we need to support high contrast colors for presence, when the OS/Browser is not in high contrast mode.
Since the presence-colors are not themed I added an optional prop on the Persona component that allows you to set the fore- and background colors used for presence. 

image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12318)


packages/office-ui-fabric-react/src/components/Persona/Persona.types.ts
```diff
@@ -88,6 +88,17 @@ export interface IPersonaSharedProps extends React.HTMLAttributes<PersonaBase |
    */
   initialsColor?: PersonaInitialsColor | string;
 
+  /** The colors to be used for the presence-icon and it's background */
+  presenceColors?: {
+    presenceColorAvailable: string;
```
mtennoe: Should we make the naming less redundant?

andrehag: Yes we should.

PR #12319, opened 2020-03-16
feat(icons): add factory for creating svg icons

mnajdova:
This PR introduces `SvgIcon` utility component that can be used for creating SVG icons. The motitivation of this is mostly for reducing the bundle size. Currently all icons that we have (all svgs) are part of the gigantic theme object. With this factory, we are allowing the users to create themable component icons, without including the svgs inside the theme object.

For this to work there are two ways of creating the icons.

1. Using the factory for generating svg icons:

```tsx
const teamsIconClassNames = {
  filled: 'ui-icon__filled',
  outline: 'ui-icon__outline'
};

const callVideoSvg = ({ classes }) => (
  <svg className={classes.svg} viewBox="8 8 16 16" role="presentation" focusable="false">
    <g className={cx(teamsIconClassNames.outline, classes.outlinePart)}>
      <path d="..." />
      <path d="..." />
    </g>
    <g className={cx(teamsIconClassNames.filled, classes.filledPart)}>
      <path d="..." />
      <path d="..." />
    </g>
  </svg>
);

const CallVideoIcon = createSvgIcon({ svg: callVideoSvg, displayName: 'CallVideoIcon'});
```

2. Creating simple component that will use the `SvgIcon` under the hood (this will result with two components inside the React tree)

```tsx
const BookIcon = props => (
  <SvgIcon {...props}>
    {({ classes }) => (
      <svg role="presentation" focusable="false" viewBox="8 8 16 16" className={classes.svg}>
        <path d="..." />
      </svg>
    )}
  </SvgIcon>
);
```


packages/fluentui/react/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,146 @@
...
+    }),
+    mapPropsToInlineStyles: () => ({ className, design, styles, variables }),
+    rtl: context.rtl
+  });
+
+  const ElementType = getElementType(props);
+  const unhandledProps = getUnhandledProps(SvgIcon.handledProps, props);
+
+  const element = (
+    <ElementType {...getA11Props('root', { className: classes.root, ...unhandledProps })}>
```
mnajdova: Some icons may have more props, so unhandledProps may spread those on the root :\

packages/fluentui/react/src/themes/teams/components/Icon/svgIconStyles.ts
```diff
@@ -0,0 +1,110 @@
...
+    case 'before':
+      return { marginLeft: horizontalSpace, marginRight: 0 };
+    case 'after':
+      return { marginLeft: 0, marginRight: horizontalSpace };
+    case 'both':
+      return { marginLeft: horizontalSpace, marginRight: horizontalSpace };
+  }
+};
+
+const svgIconStyles: ComponentSlotStylesPrepared<SvgIconStylesProps, IconVariables> = {
```
mnajdova: Simplified iconStyles that do not consider font icons and colors

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,146 @@
...
+  } = props;
+
+  const getA11Props = useAccessibility(accessibility, {
+    debugName: SvgIcon.displayName,
+    mapPropsToBehavior: () => ({
+      alt,
+      'aria-label': ariaLabel
+    }),
+    rtl: context.rtl
+  });
```
layershifter: I don't think that we need to use there `useAccessibility`, instead let's apply just props from that behavior

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
layershifter: Started to think about [@emotion/is-prop-valid](https://www.npmjs.com/package/@emotion/is-prop-valid), however there we can add `handledProps` to the `.create()` factory

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,146 @@
...
+  circular: PropTypes.bool,
+  disabled: PropTypes.bool,
+  outline: PropTypes.bool,
+  rotate: PropTypes.number,
+  size: customPropTypes.size,
+  xSpacing: PropTypes.oneOf(['none', 'before', 'after', 'both'])
+};
+SvgIcon.handledProps = Object.keys(SvgIcon.propTypes) as any;
+
+SvgIcon.create = (svgIcon: SvgIconChildrenFn) => props => <SvgIcon {...props}>{svgIcon}</SvgIcon>;
```
layershifter: We will need something more smart to avoid two items in React's tree

packages/fluentui/docs/src/examples/components/Icon/Types/IconExample.shorthand.tsx
```diff
@@ -1,12 +1,45 @@
...
+      </svg>
+    )}
+  </SvgIcon>
+);
+
+const callVideoSvg = ({ classes }) => (
```
layershifter: Do we really need there classes? 🤔 Can we avoid these via selectors in styles?

```
const styles = {
   root: () => ({
      // styles
      'svg': { /* styles for `svg` /* },
   })
}
```

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,197 @@
...
+  circular: PropTypes.bool,
+  disabled: PropTypes.bool,
+  outline: PropTypes.bool,
+  rotate: PropTypes.number,
+  size: customPropTypes.size,
+  xSpacing: PropTypes.oneOf(['none', 'before', 'after', 'both'])
+};
+SvgIcon.handledProps = Object.keys(SvgIcon.propTypes) as any;
+
+SvgIcon.create = ({ svg, displayName, handledProps }: SvgIconCreateFnParams) => {
```
mnajdova: Lots of duplicates with the component itself, try to reuse some parts

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,197 @@
...
+        xSpacing
+      }),
+      mapPropsToInlineStyles: () => ({ className, design, styles, variables }),
+      rtl: context.rtl
+    });
+
+    const unhandledProps = getUnhandledProps(handledProps, props);
+
+    return (
+      <span
```
layershifter: 👍

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,197 @@
...
+        role="img"
+        aria-hidden={alt || ariaLabel ? undefined : 'true'}
+        aria-label={ariaLabel}
+        className={classes.root}
+        {...unhandledProps}
+      >
+        {callable(svg)({ classes, rtl: context.rtl, props })}
```
layershifter:
```suggestion
        {svg({ classes, rtl: context.rtl, props })}
```

_nit_
I would say that it should be always a function

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,197 @@
...
+      mapPropsToStyles: () => ({
+        bordered,
+        circular,
+        disabled,
+        outline,
+        rotate,
+        size,
+        xSpacing
+      }),
+      mapPropsToInlineStyles: () => ({ className, design, styles, variables }),
```
layershifter:
```suggestion
      mapPropsToInlineStyles: () => ({ className }),
```

For now, let's keep compatibility with other components, but I think that these props (`styles`, `variables`) should be removed

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,197 @@
...
+      design,
+      outline,
+      rotate = 0,
+      size = 'medium',
+      styles,
+      variables,
+      xSpacing
+    } = props;
+
+    const { classes } = useStyles(SvgIcon.displayName, {
```
layershifter: Once we will agree on API I want to check how we can optimize styles to make it faster

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,197 @@
...
+  disabled: PropTypes.bool,
+  outline: PropTypes.bool,
+  rotate: PropTypes.number,
+  size: customPropTypes.size,
+  xSpacing: PropTypes.oneOf(['none', 'before', 'after', 'both'])
+};
+SvgIcon.handledProps = Object.keys(SvgIcon.propTypes) as any;
+
+SvgIcon.create = ({ svg, displayName, handledProps }: SvgIconCreateFnParams) => {
+  // TODO: add props types for the component
```
layershifter: Just noticing, that it should allow to type custom props properly

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
layershifter: `.create()` is a good name for prototype, but I would to even with `createIcon()` function to avoid confusion between factories in other components

packages/fluentui/docs/src/examples/components/Icon/Types/IconExample.shorthand.tsx
layershifter: This also can have a performance impact, BTW

layershifter:
1. Do we have a migration plan? Codemods? [codemod](https://github.com/codemod-js/codemod) or other?
2. Are we going to create a separate package for icons? Will we move factory to `react-bindings`?
3. What our strategy for components? Something like below?

```tsx
<MenuItem icon={{ content: <BookIcon /> }} />
<MenuItem icon={<BookIcon />} />
```

Will `icon` be a `Box` shorthand?

packages/fluentui/react/src/components/Icon/SvgIcon.tsx
mnajdova: Yep, that's why I left the comment. Added.

packages/fluentui/docs/src/examples/components/Icon/Types/IconExample.shorthand.tsx
mnajdova: At this moment we agreed that it's better to keep it as close as possible to the prev implementation, so we can have a smooth migration

packages/fluentui/react/src/themes/teams/components/Icon/svg/icons/error.tsx
```diff
@@ -10,6 +10,7 @@ export default {
       />
     </svg>
   ),
+  // TODO: styles overrides
```
mnajdova: This was never working in the previous implementation :\ It seems like this is the only one used in production, so I believe it is save to move this to the svgIconStyles, and later if necessary we can allow users to define additional styles to their icons.

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
```diff
@@ -1,17 +1,40 @@
...
-import { default as SvgIcon, SvgIconProps, SvgIconChildrenFn } from './SvgIcon';
+import { StylesContextValue } from '../styles/types';
+import useStyles from '../hooks/useStyles';
+import getUnhandledProps from '../utils/getUnhandledProps';
+import { SvgIconCreateFnParams, SvgIconProps } from './types';
+
+export const SvgIconClassName = 'ui-icon';
+export const SvgIconDisplayName = 'SvgIcon';
+
+export const SvgIconPropTypes = {
```
layershifter: We are going to remove `propTypes` at all. Should we drop them here?

mnajdova: Make sense..

mnajdova:
@layershifter answers to the questions:

> Do we have a migration plan? Codemods? codemod or other?

For migrating icons definition to components, some script with regexp will be sufficient as the API is almost identical as the prev icons definition

> Are we going to create a separate package for icons? Will we move factory to react-bindings?

Yes, we will have a separate package for the icons. I've moved the factory to react-bindings 👍 

> What our strategy for components? Something like below?
> <MenuItem icon={{ content: <BookIcon /> }} />
> <MenuItem icon={<BookIcon />} />
> Will icon be a Box shorthand?

I see it as a plain JSX element property, I've played with the updates icons as well as third party icons, and I believe this can work, but will create a separate RFC dedicated for that.

packages/fluentui/react-bindings/test/icons/createSvgIcon-test.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const TestIcon = createSvgIcon({ svg: testSvg, displayName: 'TestIcon' });
+    expect(TestIcon.displayName).toEqual('TestIcon');
+  });
+
+  it('spreads unhandled props on the root element', () => {
+    const TestIcon = createSvgIcon<SvgIconProps>({ svg: testSvg, displayName: 'TestIcon' });
+
+    const styles = jest.fn();
+    const wrapper = mount(<TestIcon id="test-id" />, {
+      // @ts-ignore typings are outdated
```
layershifter:
```suggestion
```

We have actual in this repo

packages/fluentui/react-bindings/test/icons/createSvgIcon-test.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  it('sets displayName to the created Component', () => {
+    const TestIcon = createSvgIcon({ svg: testSvg, displayName: 'TestIcon' });
+    expect(TestIcon.displayName).toEqual('TestIcon');
+  });
+
+  it('spreads unhandled props on the root element', () => {
+    const TestIcon = createSvgIcon<SvgIconProps>({ svg: testSvg, displayName: 'TestIcon' });
+
+    const styles = jest.fn();
```
layershifter: This is never called

packages/fluentui/react-northstar/src/components/Icon/SvgIcon.tsx
```diff
@@ -0,0 +1,90 @@
...
+SvgIcon.defaultProps = {
+  as: 'span',
+  size: 'medium',
+  rotate: 0,
+};
+
+/**
+ * An SvgIcon displays a pictogram with semantic meaning.
+ */
+export default withSafeTypeForAs<typeof SvgIcon, SvgIconProps, 'span'>(SvgIcon);
```
layershifter: Is this cruft?

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
```diff
@@ -0,0 +1,91 @@
...
+  'variables',
+  'xSpacing',
+];
+
+const createSvgIcon = <TProps extends SvgIconProps>({
+  svg,
+  displayName,
+  handledProps = SvgIconHandledProps,
+}: SvgIconCreateFnParams) => {
+  const Component: React.FC<TProps> & { handledProps: string[] } = (props: Extendable<SvgIconProps>) => {
```
layershifter: If this will be always `span`:

```suggestion
  const Component: React.FC<React.HTMLAttributes<HTMLSpanElement> & TProps> & { handledProps: string[] } = (props) => {
```

packages/fluentui/react-bindings/README.md
```diff
@@ -221,3 +224,26 @@ const { classes } = useStyles(
...
+const circleSvg = ({ classes }) => (
+  <svg role="presentation" focusable="false" viewBox="8 8 16 16" className={classes.svg}>
+    <g>
+      <path d="M16,8c-4.418,0-8,3.582-8,8s3.582,8,8,8s8-3.582,8-8S20.418,8,16,8z M16,22.85c-3.783,0-6.85-3.067-6.85-6.85S12.217,9.15,16,9.15s6.85,3.067,6.85,6.85S19.783,22.85,16,22.85z" />
+      <circle className={classes.filledPart} cx="16" cy="16" r="8" />
+    </g>
+  </svg>
+);
+
+const CircleIcon = createSvgIcon({ svg: circuleSvg, displayName: 'CirculeIcon' });
```
layershifter:
```suggestion
const CircleIcon = createSvgIcon({ svg: circuleSvg, displayName: 'CircleIcon' });
```

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
```diff
@@ -0,0 +1,92 @@
...
+  'rotate',
+  'styles',
+  'variables',
+  'xSpacing',
+];
+
+const createSvgIcon = <TProps extends SvgIconProps>({
+  svg,
+  displayName,
+  handledProps = SvgIconHandledProps,
```
layershifter:
```suggestion
  handledProps = []
```

I suggest this and then:

```
const unhandledProps = getUnhandledProps([ ...SvgIconHandledProps, ...handledProps], props);
```

Because otherwise I will need to merge this in `createSvgIcon` by myself 😭

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+import { StylesContextValue } from '../styles/types';
+import useStyles from '../hooks/useStyles';
+import getUnhandledProps from '../utils/getUnhandledProps';
+import { SvgIconCreateFnParams, SvgIconProps } from './types';
+
+export const SvgIconClassName = 'ui-icon';
+export const SvgIconDisplayName = 'SvgIcon';
+
+export const SvgIconHandledProps: (keyof SvgIconProps)[] = [
```
layershifter:
```suggestion
const svgIconHandledProps: (keyof SvgIconProps)[] = [
```

I don't think that we need to export this

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
```diff
@@ -0,0 +1,92 @@
+import * as React from 'react';
+// @ts-ignore
+import { ThemeContext } from 'react-fela';
+
+import { StylesContextValue } from '../styles/types';
+import useStyles from '../hooks/useStyles';
+import getUnhandledProps from '../utils/getUnhandledProps';
+import { SvgIconCreateFnParams, SvgIconProps } from './types';
+
+export const SvgIconClassName = 'ui-icon';
```
layershifter:
```suggestion
export const svgIconClassName = 'ui-icon';
```

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
```diff
@@ -0,0 +1,92 @@
...
+export const SvgIconDisplayName = 'SvgIcon';
```
layershifter:
```suggestion
export const svgIconDisplayName = 'SvgIcon';
```

packages/fluentui/react-bindings/src/icons/types.ts
```diff
@@ -0,0 +1,68 @@
...
+  styles?: ComponentSlotStyle;
+
+  /** Override for theme site variables to allow modifications of component styling via themes. */
+  variables?: ComponentVariablesInput;
+
+  /** Adds space to the before, after or on both sides of the icon, or removes the default space around the icon ('none' value) */
+  xSpacing?: SvgIconXSpacing;
+}
+
+export type SvgIconFuncArg = {
```
layershifter: This probably should be generic:

```
type BookProps = { foo: boolean }



const BookIcon = createIcon<BookProps>({
  svg: (classes, rtl, props /* Oops, doesn't include BookProps */ ) => ()
})
```

layershifter: I suggest to improve our UTs to test this case

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
mnajdova: The same is used in the `SvgIcon` component, that's why we need to export it

packages/fluentui/react-bindings/test/icons/createSvgIcon-test.tsx
```diff
@@ -41,12 +41,35 @@ describe('createSvgIcon', () => {
...
+    expect(wrapper.find(`.${svgIconClassName}`).props().id).toEqual('test-id');
+  });
+
+  it('provides all props on the svg function', () => {
+    interface BookProps extends SvgIconProps {
+      foo: boolean;
+    }
+    const svg = jest.fn();
+    const BookIcon = createSvgIcon<BookProps>({
+      svg,
```
layershifter:
```suggestion
      svg: (classes, rtl, props) => <svg data-foo={props.foo.toString()} data-outline={props.outline.toString()} />
```

I suggest to test implementation more, this will also ensure type safety

packages/fluentui/react-bindings/test/icons/createSvgIcon-test.tsx
```diff
@@ -41,12 +41,35 @@ describe('createSvgIcon', () => {
...
     const wrapper = mount(<TestIcon id="test-id" />, {
       wrappingComponent: TestProvider,
       wrappingComponentProps: { theme: createTheme() },
     });
-    expect(wrapper.find(`.${SvgIconClassName}`).props().id).toEqual('test-id');
+    expect(wrapper.find(`.${svgIconClassName}`).props().id).toEqual('test-id');
+  });
+
+  it('provides all props on the svg function', () => {
+    interface BookProps extends SvgIconProps {
```
layershifter:
```suggestion
    interface BookProps {
```

_I_ expect that `BookProps` should be enough

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
```diff
@@ -26,13 +26,9 @@ export const svgIconHandledProps: (keyof SvgIconProps)[] = [
...
 ];
 
-const createSvgIcon = <TProps extends SvgIconProps>({
-  svg,
-  displayName,
-  handledProps = [],
-}: SvgIconCreateFnParams<TProps>) => {
-  const Component: React.FC<React.HTMLAttributes<HTMLSpanElement> & TProps> & {
-    handledProps: (keyof TProps)[];
+const createSvgIcon = <TProps,>({ svg, displayName, handledProps = [] }: SvgIconCreateFnParams<TProps>) => {
```
layershifter:
```suggestion
const createSvgIcon = <TProps>({ svg, displayName, handledProps = [] }: SvgIconCreateFnParams<TProps>) => {
```

nit

packages/fluentui/react-bindings/test/icons/createSvgIcon-test.tsx
```diff
@@ -51,25 +51,22 @@ describe('createSvgIcon', () => {
...
       wrappingComponentProps: { theme: createTheme() },
     });
 
-    expect(svg).toBeCalledWith(
-      expect.objectContaining({
-        props: { foo: true, outline: true },
-      }),
-    );
+    expect((wrapper.find('svg').props() as any)['data-foo']).toEqual('true');
+    expect((wrapper.find('svg').props() as any)['data-outline']).toEqual('true');
```
layershifter:
```suggestion
    expect(wrapper.find('svg').prop('data-foo').toEqual('true');
    expect(wrapper.find('svg').prop('data-outline').toEqual('true');
```

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
mnajdova: Hm , I think it was auto formatted. Thanks!

packages/fluentui/react-bindings/src/icons/createSvgIcon.tsx
mnajdova: Fixed

packages/fluentui/react-bindings/src/icons/createSvgIcon.ts
```diff
@@ -0,0 +1,87 @@
...
+      rtl: context.rtl,
+    });
+
+    const unhandledProps = getUnhandledProps([...svgIconHandledProps, ...handledProps], props);
+
+    return React.createElement(
+      'span',
+      {
+        role: 'img',
+        'aria-hidden': alt || ariaLabel ? undefined : 'true',
```
mnajdova: @kolaps33 @jurokapsiar I copied these props from the iconBehavior, but as @silviuavram pointed out alt is not a valid span element and even more we are not applying it anywhere. Should we remove it?

silviuavram: I brought it up after going through https://css-tricks.com/accessible-svgs/ .

There it says that in order to have `alt` like description in `<svg>` then the svg should have a child `<title>` with that text.
Also I see that they suggest adding `role="img"` on the `<svg>` element.

Everything is explained in the `2. Inline SVG` section. Let me know your thoughts.

mnajdova: But as it worked before, we have the role="img" on the span element (based on the iconBeavior not the svg). If the role="img" should be on the svg, than I guess I should remove all attributes from here, and we can just add comments of best practices.

mnajdova: BTW this means that the iconBehavior should probably be changed too.

packages/fluentui/react-bindings/test/icons/createSvgIcon-test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    expect(wrapper.find(`.${svgIconClassName}`).props().id).toEqual('test-id');
+  });
+
+  it('provides all props on the svg function', () => {
+    interface BookProps {
+      foo: boolean;
+    }
+
+    const BookIcon = createSvgIcon<BookProps>({
+      svg: ({ props }) => <svg data-foo={props.foo?.toString()} data-outline={props.outline?.toString()} />,
```
layershifter:
```suggestion
      svg: ({ props }) => <svg data-foo={props.foo.toString()} data-outline={props.outline?.toString()} />,
```

`foo` is not optional in `interface`, this should work

packages/fluentui/react-bindings/src/icons/createSvgIcon.ts
mnajdova: Resolved after offline discussion. We won't do any changes, as this is a working solution. If we decide to change the behavior in the future, we will update the factory too.

PR #12286, opened 2020-03-12
Fabric fluent 1

micahgodbolt:
This is stage 1 of the Fabric to Fluent website rename being merged into `website-update` branch. THIS IS A SUBSET, NOT EVERY INSTANCE OF "FABRIC"

In this PR I focused on the following:

1. change `OfficeDev/office-ui-fabric-react` to `microsoft/fluentui` in website md files
2. change `office-ui-fabric-react` to `@uifabric/react`
3. change `Office UI Fabric React`, `Office UI Fabric`, and `UI Fabric` to `Fluent UI`

Changes to the string "Fabric " (Title case with space after) will be in subsequent PR
Followed by finding all of the remaining 'fabric' matches.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12286)

PR #12321, opened 2020-03-16
Fabric to Fluent PR #2

micahgodbolt:
This is built off of #12286, so don't merge (or review) this until the first is done and merged in.

This PR is focused on replaced Fabric with Fluent UI in text. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12321)


ecraig12345:
I'm not clear on the reasoning for splitting this out?

micahgodbolt:
I wanted to make each PR a bit easier to parse. focus on a set of changes.

I can glob everything into one monster huge PR is that's prefered.

ecraig12345:
I see the point in general, but it's really hard to tell right now what this PR actually does since it also includes everything from the other one. And it makes the other one a little harder to review (at least as the descriptions are right now) since I'm not clear on what you plan to address later.

micahgodbolt:
Yeah, when I created the PR is forgot that the diff would include both steps. I'll close this one down until the first is finished, as I don't need eyes on this one yet.

ecraig12345:
Sounds good. Maybe also update the other one with a list of things that it's intentionally NOT doing. :)

PR #12335, opened 2020-03-17
feat(compose): add compose() & scaffold `react-compose-next`

layershifter:
#### Pull request checklist

#### Description of changes

This PR adds a `compose()` function to scaffolded package `@fluentui/react-compose-next`. It will reused in `@fluentui/react-bindings` (V0, #12239). As `compose` just adds a static field there should be a way how it can be reused in V7.

Changes between implementation there:
- `mapPropsToBehavior` were removed as it's not a case for V7
- `mapPropsToStyles` were renamed to `mapPropsToStylesProps` as was suggested by @levithomason 

#### Focus areas to test

(optional)


packages/react-compose-next/README.md
```diff
@@ -0,0 +1,3 @@
+# @fluentui/react-compose-next
+
+**TBD**
```
layershifter: I will add docs in a follow up PR

packages/react-compose-next/src/compose.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  return [inputDisplayName, options.displayName].filter(Boolean) as string[];
+}
+
+function compose<
+  InputProps extends Record<string, any>,
+  InputStylesProps extends Record<string, any>,
+  ParentProps extends Record<string, any>,
+  ParentStylesProps extends Record<string, any>
```
layershifter: Typings are a bit noisy now, but it's better to have them than have nothing. Typings will be changed based on dogfood

layershifter:
I didn't include any tests here to avoid tests for internal. Tests for V0 will be added in #12239.

packages/react-compose-next/src/compose.ts
```diff
@@ -0,0 +1,62 @@
...
+>(
+  InputComponent: React.FunctionComponent<ParentProps>,
```
layershifter: This will not accept a `forwardRef`, I will test it separately and will make a follow up

packages/react-compose-next/src/compose.ts
dzearing: We can iterate :)

packages/react-compose-next/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@fluentui/react-compose-next",
+  "version": "7.0.0",
```
dzearing: Suggest v0 until we're ready

```suggestion
  "version": "0.1.0",
```

packages/react-compose-next/README.md
levithomason: I think we should fill this out with what we have working.  Let's keep merges as user-ready as possible.

packages/react-compose-next/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@fluentui/react-compose-next",
```
levithomason:
```suggestion
  "name": "@fluentui/react-compose",
```

packages/react-compose-next/README.md
```diff
@@ -0,0 +1,30 @@
+# @fluentui/react-compose
+
+**Composition functionality for in [Fluent UI](https://aka.ms/fluent-ui/)**
```
mnajdova:
```suggestion
**Composition functionality for [Fluent UI](https://aka.ms/fluent-ui/)**
```

packages/react-compose-next/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+This functionality is in the experimental stage and **should never be used in product** as APIs might change before final release.
```
mnajdova:
```suggestion
This functionality is in the experimental stage and **should never be used in production** as APIs might change before final release.
```

packages/react-compose-next/README.md
```diff
@@ -0,0 +1,30 @@
...
+  handledProps: ['compact'],
+  overrideStyles: true
+});
+```
+
+- `className` overrides a `className` that we will be assigned to the `root` slot of a component
+- `displayName` modifies a React's `displayName` for component, it will be also used as a selector for styles overrides in your theme
+- `mapPropsToStylesProps` passes additional design terms to styles functions
+- `handledProps` is an array of props that will not be bypassed to the component's root slot
+- `overrideStyles` indicates should the component use only its styles or will merge with parent's
```
mnajdova:
```suggestion
- `overrideStyles` indicates should the component will use only its own styles or will merge those with parent's
```

packages/react-compose-next/README.md
```diff
@@ -0,0 +1,30 @@
+# @fluentui/react-compose
+
+**Composition functionality for in [Fluent UI](https://aka.ms/fluent-ui/)**
+
+This functionality is in the experimental stage and **should never be used in product** as APIs might change before final release.
+
+### Usage
+
+To import `compose` and use:
```
mnajdova: Maybe we should clarify that this is currently working on functional components only

layershifter: Added 👍

packages/react-compose-next/README.md
layershifter: I added initial docs for `compose()` function. Feel free to update it if I missed something

PR #12337, opened 2020-03-17
Button: add an example of IconButton with a Tooltip.

### LINKED ISSUE #12327: Build an IconButton that is wrapped in an accessible TooltipHost by default
We use IconButtons all over the place in our apps and shared libraries. We're getting hammered by WCAG 2.1 accessibility bugs because the current IconButton doesn't provide a tool tip that explains the intent of the button on keyboard focus. 

We're having to wrap all our icon buttons in a tool tip host, or create a shared component to do this. However it feels like something that we should get from the Fabric component by default.

IconButton

Not yet - would love an opportunity to, and maybe share back the fix.

aneeshack4:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12327 

#### Description of changes

Adding an example to show how `IconButton` can be wrapped with a `Tootlip`.
image

#### Focus areas to test

PR deploy site should have new example added on Buttons page.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12337)


packages/office-ui-fabric-react/src/components/Button/examples/Button.IconWithTooltip.Example.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <div>
+      <TooltipHost
+        content="This is the tooltip content"
+        // This id is used on the tooltip itself, not the host
+        // (so an element with this id only exists when the tooltip is shown)
+        id={tooltipId}
+        calloutProps={calloutProps}
+        styles={hostStyles}
+      >
+        <IconButton iconProps={emojiIcon} title="Emoji" ariaLabel="Emoji" disabled={disabled} checked={checked} />
```
xugao: remove `title="Emoji"` here. instead, set `TooltipHost` `content` prop to be `"Emoji"`

packages/office-ui-fabric-react/src/components/Button/examples/Button.IconWithTooltip.Example.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const ButtonIconWithTooltipExample: React.FunctionComponent<IButtonExampleProps> = props => {
+  // Use useId() to ensure that the ID is unique on the page.
+  // (It's also okay to use a plain string and manually ensure uniqueness.
+  const tooltipId = useId('tooltip');
+
+  const { disabled, checked } = props;
+
+  return (
+    <div>
```
xugao: nit: do we need the `div` here?

packages/office-ui-fabric-react/src/components/Button/Button.doc.tsx
```diff
@@ -65,6 +67,11 @@ export const ButtonPageProps = (props: IButtonDocPageProps): IDocPageProps => ({
       code: ButtonIconExampleCode,
       view: <ButtonIconExample disabled={props.areButtonsDisabled} checked={props.areButtonsChecked} />
     },
+    {
+      title: 'Icon with Tooltip Button',
```
xugao: Icon Button with Tooltip

packages/office-ui-fabric-react/src/components/Button/examples/Button.IconWithTooltip.Example.tsx
ecraig12345: Does that still pass a11y insights?

change/office-ui-fabric-react-2020-03-17-17-21-14-iconbutton.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "minor",
+  "comment": "Button: add an example of IconButton with a Tooltip",
```
khmakoto:
```suggestion
  "comment": "Button: Add an example of an IconButton that has a Tooltip.",
```

packages/office-ui-fabric-react/src/components/Button/examples/Button.IconWithTooltip.Example.tsx
khmakoto: Worst case scenario I think it should be "Emoji" in both the `title` and the `content`.

change/office-ui-fabric-react-2020-03-17-17-21-14-iconbutton.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "minor",
```
ecraig12345: Should be patch since it's not adding a new API.
```suggestion
  "type": "patch",
```

packages/office-ui-fabric-react/src/components/Button/examples/Button.IconWithTooltip.Example.tsx
aneeshack4: Yes bc of the <p> 
image

packages/office-ui-fabric-react/src/components/Button/examples/Button.IconWithTooltip.Example.tsx
aneeshack4: Done.

packages/office-ui-fabric-react/src/components/Button/examples/Button.IconWithTooltip.Example.tsx
xugao: use `<>` `</>` instead

packages/office-ui-fabric-react/src/components/Button/examples/Button.IconWithTooltip.Example.tsx
```diff
@@ -0,0 +1,46 @@
...
+        // (so an element with this id only exists when the tooltip is shown)
+        id={tooltipId}
+        calloutProps={calloutProps}
+        styles={hostStyles}
+      >
+        <IconButton iconProps={emojiIcon} title="Emoji" ariaLabel="Emoji" disabled={disabled} checked={checked} />
+      </TooltipHost>
+      <p>
+        For now, we advise you to take this approach of wrapping IconButton with a Tooltip. We'll address providing this
+        behavior out of the box in next version of this component in Fluent.
```
ecraig12345: Minor wording suggestion
```suggestion
        behavior out of the box in a future major version.
```

aneeshack4:
@khmakoto need codeowner approval

PR #12339, opened 2020-03-17
keyboard-key: Reducing bundle size and integrating into FocusZone 

khmakoto:
#### Pull request checklist

#### Description of changes

This PR is an attempt to reduce the bundle size impact of importing `@fluentui/keyboard-key` package into other packages. Although this increases the size of the package, it actually reduces the impact when importing things because it is splitting them into multiple exports instead of one big export. For example, if we were to introduce the `keyboard-key` package into `FocusZone` its impact before these changes would be ~3.3KB vs ~1.6KB after these changes.

As part of this effort this PR also introduces a custom tslint rule to enforce using the `keyboard-key` library instead of `ev.which` and `ev.keyCode` and turns on this lint rule in the `@fluentui/react-focus` package to vet it and disable the rule as needed.

A comment in #12270 prompted to add a snippet of documentation showing devs how to handle event checks. I feel like this, however, should be made whenever we make the full switch across our packages.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12339)


packages/keyboard-key/src/index.ts
```diff
@@ -303,54 +303,57 @@ const keyboardKeyDefinition = {
    * Mapping from numeric key code to key name. If the value is an array, the first element is the
    * primary key name, and the second element is the key name when shift is pressed.
    */
-  codes: codes,
+  codes: codes
```
ecraig12345: Do  we actually need an object with a `codes` property or can `codes` be exported as a separate const?

khmakoto: Funny for you to write this. It is specifically what I'm trying right now to see how much bundle size I can shave off 😆.

ecraig12345: Wait why? 😆

khmakoto: I meant exporting `codes` as a separate const.

dzearing:
AWESOME @khmakoto !!!

packages/keyboard-key/src/index.ts
ecraig12345: Ahhh I incorrectly parsed your comment yesterday :)

Issue #11952, opened 2020-02-14
Focus wont go inside tooltip link. how to provide focus ? And It should work with Narrator & NVDA.


v-jashr:
Current Behavior:
on hover tooltip appears but we have link inside tooltip and focus won't go inside tooltip link. how to provide focus.?

Expected Behavior:
Focus should go inside tooltip link. and It Narrator & NVDA should work provide info about it.

xugao:
@v-jashr - that is expected. see [ARIA](https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip)
>Tooltip widgets do not receive focus. A hover that contains focusable elements can be made using a non-modal dialog.

Dup w/ #11711

Issue #11711, opened 2020-01-15
Hyperlinks within custom tooltip content are not accessible


nkalpana:
### Environment Information

- **Package version(s)**: Latest version of Fabric used by "Export to codepen" from sampls
- **Browser and OS versions**: Chrome and Windows 10

### Describe the issue:
When TooltipHost's content prop is set to a JSX.Element which has focusable content, like a Hyperlink, despite setting the ariaLabel for the focusable content, user is not able to tab into it.
Only in the rare case of the cursor moves over the tooltip host's child content and is also on the automatically displayed tooltip content, can I tab into the hyperlink and hit enter to see it.
Also the ariaLabel of the hyperlink is not announced by Narrator/NVDA

### Please provide a reproduction of the issue in a codepen:
https://codepen.io/nkalpana/pen/rNaKqpP

#### Actual behavior:
Tooltip content appears when user clicks on the tooltip host's child contorl or moves mouse over it. But, in that state if the user tabs or hits right arrow, the focus does not get to the hyperlink in the tooltip hosts' content

#### Expected behavior:
Tooltip content appears when user clicks on the tooltip host's child contorl or moves mouse over it. User hits tab key or left arrow key, and the focus moves over to the hyperlink.

xugao:
@nkalpana - this is by design. tooltip widget should not be focusable, see [ARIA guidance here](https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip)

sonimishra:
@xugao Here in the doc it is mentioned that focus should not go to the tooltip widget outline but if the widget contains any interactive controls then focus should navigate to that particular control like in this case "click here" link which is interactive.

khmakoto:
@sonimishra where in the ARIA doc does it say that? I explicitly states:

>Tooltip widgets do not receive focus. A hover that contains focusable elements can be made using a non-modal dialog.

Issue #10240, opened 2019-08-22
Tooltip: moving hover from tooltip back to trigger closes tooltip


jlukosky:
### Environment Information

- **Package version(s)**: (fill this out)
- **Browser and OS versions**: (fill this out)

> Note that our screen reader support is limited to Edge + Narrator. Issues that do not reproduce in Edge + Narrator, and aren't caused by obvious invalid aria values, should be filed with the respective screen reading software, not the Fabric repo.

### Describe the issue:
When moving hover from tooltip back to trigger the tooltip closes. The trigger shouldn't close the tooltip, it's behavior should be consistent. As it is the trigger is serving as a show/hide toggle.

#### Actual behavior:

Moving hover from tooltip to trigger closes tooltip.

#### Expected behavior:

Moving hover from tooltip to trigger should not close the tooltip

### Documentation describing expected behavior

The expectation is that the behavior will be consistent: hovering the trigger should show the tooltip. As it is now sometimes hovering the trigger shows the tooltip, sometimes it hides it.

maxwellred:
Hi @jlukosky . Thanks for opening this issue about Tooltip and congratulations on this being your first in UI Fabric React! I have contacted the code owner in regards to this behavior and will be attending to it shortly.

micahgodbolt:
Having tooltips be sticky when hovered isn't a 100% normal HTML experience, so we rely on our designers to inform us on how it should work. The fix shouldn't be too hard to do, but I want to first confirm that this is the correct design choice before doing the work.

@betrue-final-final do you have any context on if tooltip should stay open after hovereing over the tooltip and then back onto the button? @jspurlin do you know what the expectation is on the ribbon with tooltips?

jspurlin:
I know in the desktop apps, the user can hover over the tooltip and then back over the button (but windows and native tooltips in the browser don't do that). This sounds like a feature request that should be vetted with designers.

kenotron:
This needs spec'ing before proceeding, I'm removing myself and Max off of the assignee in this case leaving the backlog review label.

jlukosky:
This should be reopened and resolved. If this isn't addressed it is a violation of 1.4.13: Content on Hover or Focus

PR #12344, opened 2020-03-18
fix(Tooltip): Prevent callout content from gaining mouse focus

### LINKED ISSUE #12285: Tooltip cannot be interacted with when displayed on pressed button
### Environment Information

- **Package version(s)**: 7.83.0
- **Browser and OS versions**: Chrome Version 80.0.3987.132 (Official Build) (64-bit)/MacOS Mojave 10.14.6

1. click on the [Click me first] button
    (due to mouse hover, a tooltip appears)
3. mouse over to the [Mouse over on me second] button in the tooltip
4. primary click the [Mouse over on me second] button

https://codepen.io/joshjcarrier/pen/mdJXGpK

#### Actual behavior:

The mouse click down on [Mouse over on me second?] button does not activate the button, and causes the tooltip to dismiss.

fabric_actual

#### Expected behavior:

The mouse click down on [Mouse over on me second?] button should activate the button, and prevent the tooltip from dismissing.

fabric_expected

Requested priority: Normal

Products/sites affected: n/a

joshjcarrier:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12285

#### Description of changes

We've already concluded that [tooltips don't support focus](https://github.com/OfficeDev/office-ui-fabric-react/issues/10240#issuecomment-525906499) and in some scenarios allowing mouse down on tooltip content attempts to do so. This returns focus to the target element hosting the tooltip, and since the mouse is no longer hovering over it, the tooltip dismisses skipping any component being interacted with (like a link or button).

This adds a mouse down handler to the tooltip callout content to prevent focus from moving while still allowing elements under the cursor to receive the click.

#### Focus areas to test

* on the [Default Tooltip Usage](https://developer.microsoft.com/en-us/fabric#/controls/web/tooltip) example, click the [Hover over me] button and then click the tooltip. Expect that it no longer dismisses unexpectedly.

* on the [Default Tooltip Usage](https://developer.microsoft.com/en-us/fabric#/controls/web/tooltip) example, hover over the [Hover over me] button and double-click the text content `This is the tooltip content`. Expect that it no longer receives focus via text selection.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12344)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.base.tsx
@@ -45,6 +45,10 @@ export class TooltipBase extends React.Component<ITooltipProps, any> {
       maxWidth: maxWidth!
     });
 
+    const onMouseDown = (ev: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+      ev.preventDefault();
+    };
+
     return (
       <Callout
         target={targetElement}
@@ -58,6 +62,7 @@ export class TooltipBase extends React.Component<ITooltipProps, any> {
           className={this._classNames.content}
           id={id}
           role="tooltip"
+          onMouseDown={onMouseDown}
           onMouseEnter={this.props.onMouseEnter}
           onMouseLeave={this.props.onMouseLeave}
         >
--- a/change/office-ui-fabric-react-2020-03-17-19-45-29-tooltipFocus.json
+++ b/change/office-ui-fabric-react-2020-03-17-19-45-29-tooltipFocus.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "fix(Tooltip): Prevent callout content from gaining focus, which may cause the tooltip to close unexpectedly.",
+  "packageName": "office-ui-fabric-react",
+  "email": "joshjcarrier@gmail.com",
+  "commit": "a440ba759cf55005d08a051c4d7125d32a77cdba",
+  "dependentChangeType": "patch",
+  "date": "2020-03-18T02:45:29.706Z"
+}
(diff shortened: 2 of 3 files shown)


packages/office-ui-fabric-react/src/components/Tooltip/Tooltip.base.tsx
```diff
@@ -45,6 +45,10 @@ export class TooltipBase extends React.Component<ITooltipProps, any> {
       maxWidth: maxWidth!
     });
 
+    const onMouseDown = (ev: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+      ev.preventDefault();
```
aneeshack4: Looks like this has regressed a lot of Screener scenarios...
image
but the Screener results link itself only shows 2 that were changed in TeachingBubble this is a bit confusing... and they don't seem related to your change. I'm going to close and reopen to trigger another run.

aneeshack4:
Pinging @micahgodbolt for codeowner approval

micahgodbolt:
waiting for passing build

aneeshack4:
@joshjcarrier Can you fix the build so that this can be merged?

joshjcarrier:
Retriggering build

ecraig12345:
I'm a little confused by this change. Seems like it has the side effect of disabling selection of tooltip text, which while it's not the primary thing you'd want to do with a tooltip, may be desirable in some cases (and I suspect someone will file an issue if we disable it).

joshjcarrier:
Yes, I was comparing the options of:
1. tooltip contains focusable content, or
2. tooltip dismisses through matching interactions (e.g. [mouseIn/mouseOut](https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-21))

But with multiple issues (#11711 #11952 #10240) already citing tooltips that should not receive focus, I propose the inconsistent dismiss behavior (as reproed on the public docs) is the defect and focusable text (and the entailing accessibility concerns) should be handled with Callout/Dialog to be supported.

PR #12360, opened 2020-03-19
feat(ResponsiveTable): Adding responsive table prototype

assuncaocharles:
#### Pull request checklist

#### Description of changes

Adding a prototype of Responsive Table which responds to the viewport changes by hiding and showing columns based on whether they fit on the screen. The order in which they disappear an reappear is determined by each columns priority.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12360)

### DIFF
--- a/packages/fluentui/docs/src/prototypes/table/ResponsiveTableContainer.tsx
+++ b/packages/fluentui/docs/src/prototypes/table/ResponsiveTableContainer.tsx
@@ -1,0 +1,51 @@
+import * as React from 'react';
+import { TableCell, TableProps, Table } from '@fluentui/react-northstar';
+import calculateBreakpoints, { Column, Breakpoint } from './calculateBreakpoints';
+import * as _ from 'lodash';
+
+interface ResponsiveTableProps extends TableProps {
+  columns: Column[];
+  children: React.ReactElement<typeof Table>;
+  breakpoints?: number[];
+  id?: string;
+}
+
+const ResponsiveTableContainer: React.FC<ResponsiveTableProps> = props => {
+  const { columns: config, children } = props;
+
+  const responsiveTableID = React.useMemo(() => props.id || _.uniqueId(`${Table.className}__responsive-wrapper`), [
+    props.id,
+  ]);
+
+  const [breakpoints, setBreakpoints] = React.useState<Breakpoint | undefined>();
+
+  /**
+   * Generate the stylesheet once and for all checking if
+   * the total size of the table ( based in th esum of the minWidth columns ) still fits in the current breakpoint
+   */
+  React.useLayoutEffect(() => {
+    setBreakpoints(calculateBreakpoints(config, props.breakpoints || [1200, 960, 600, 280]));
+  }, []);
+
+  return (
+    <div id={responsiveTableID}>
+      <style>
+        {_.map(
+          breakpoints,
+          (breakpoint, key) => `
+              @media (max-width: ${key}px) {
+                ${breakpoint
+                  .map(
+                    column =>
+                      `#${responsiveTableID} .${TableCell.className}:nth-child(${column.childIndex}) { display: none }`,
+                  )
+                  .join('\n  ')}
+              }`,
+        ).join('\n')}
+      </style>
+      {children}
+    </div>
+  );
+};
+
+export default ResponsiveTableContainer;
--- a/packages/fluentui/docs/src/prototypes/table/index.tsx
+++ b/packages/fluentui/docs/src/prototypes/table/index.tsx
@@ -1,10 +1,22 @@
 import { gridCellMultipleFocusableBehavior, gridCellWithFocusableElementBehavior } from '@fluentui/accessibility';
-import { Avatar, Button, Checkbox, Dropdown, Flex, Icon, Menu, MenuButton, Text } from '@fluentui/react-northstar';
+import {
+  Avatar,
+  Button,
+  Checkbox,
+  Dropdown,
+  Flex,
+  Icon,
+  Menu,
+  MenuButton,
+  Text,
+  Table,
+} from '@fluentui/react-northstar';
 import * as React from 'react';
 import chatProtoStyle from '.././chatPane/chatProtoStyle';
 import { ComponentPrototype, PrototypeSection } from '../Prototypes';
 import AdvancedTable, { stringCellComparator } from './AdvancedTable';
 import InteractiveTable from './InteractiveTable';
+import ResponsiveTableContainer from './ResponsiveTableContainer';
 
 function handleRowClick(index) {
   alert(`OnClick on the row ${index} executed.`);
@@ -157,6 +169,30 @@ const rowsChannels = [
   },
 ];
 
+const columnsPerson = {
+  items: ['id', 'Name', 'Age', 'Picture'],
+};
+
+const rowsPerson = [
+  ['1', 'Roman van', '30 years', 'None'],
+  ['2', 'Alex', '1 year', 'None'],
+  ['3', 'Ali', '30000000000000 years', 'None'],
+];
+
+const responsiveColumnsConfig = [
+  { priority: 4, minWidth: 200 },
+  { priority: 3, minWidth: 360 },
+  { priority: 2, minWidth: 300 },
+  { priority: 1, minWidth: 200 },
+];
+
+const responsiveColumnsConfigPriorityOrder = [
+  { priority: 3, minWidth: 200 },
+  { priority: 2, minWidth: 360 },
+  { priority: 1, minWidth: 300 },
+  { priority: 4, minWidth: 200 },
+];
+
 export default () => (
   <PrototypeSection title="Advanced table">
     <ComponentPrototype title="Table example 1" description="Table with sorting, tags and dropdown menu in a cell">
@@ -168,6 +204,22 @@ export default () => (
     </ComponentPrototype>
     <ComponentPrototype title="Table example 3" description="Table with popover and context menu ">
       <InteractiveTable />
+    </ComponentPrototype>
+    <ComponentPrototype
+      title="Responsive Table"
+      description="Responsive table hiding columns based in the priority passed to the Resposive Container as columns configurarion. The container can also receive a Breakpoint input with an array of number representing the breakpoints"
+    >
+      <ResponsiveTableContainer columns={responsiveColumnsConfig}>
+        <Table rows={rowsPerson} header={columnsPerson} arial-label="Persons" />
+      </ResponsiveTableContainer>
+    </ComponentPrototype>
+    <ComponentPrototype
+      title="Responsive Table"
+      description="Responsive table hiding middle columns keeping the first and the last"
+    >
+      <ResponsiveTableContainer columns={responsiveColumnsConfigPriorityOrder}>
+        <Table rows={rowsPerson} header={columnsPerson} arial-label="Persons" />
+      </ResponsiveTableContainer>
     </ComponentPrototype>
   </PrototypeSection>
 );
--- a/packages/fluentui/docs/src/prototypes/table/calculateBreakpoints.ts
+++ b/packages/fluentui/docs/src/prototypes/table/calculateBreakpoints.ts
@@ -1,0 +1,28 @@
+export type Breakpoints = Record<number, number> & { default?: number };
+
+export type Column = {
+  priority: number;
+  minWidth: number;
+  childIndex?: number;
+};
+
+export type Breakpoint = Record<string, Column[]>;
+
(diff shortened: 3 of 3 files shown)


packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+const ResponsiveTable: React.FC<ResponsiveTableProps> = props => {
+  const {
+    rows,
+    label,
+    header: { items },
+  } = props;
+
+  const responsiveTableID = props.id || _.uniqueId(`${Table.className}__responsive-wrapper`);
```
jurokapsiar: this way, id will regenerate on each render, can you save it in state so that it remains constant once generated?

assuncaocharles: Done

assuncaocharles:
@layershifter Generator removed :)

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  const [responsiveTableID] = React.useState(props.id || _.uniqueId(`${Table.className}__responsive-wrapper`));
+
+  const columns = ChildrenTable.props.header.items;
+
+  const [breakpoints, setBreakpoints] = React.useState<Breakpoint[]>();
+
+  // Generate the stylesheet once and for all
+  React.useLayoutEffect(() => {
+    setBreakpoints(calculateBreakpoints(config));
```
mnajdova: Don't we need to have a prop for the `breakpoints` param here? If not, I would suggest removing that param from the `calculateBreakpoints`

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }, []);
+
+  return (
+    <div id={responsiveTableID}>
+      <style>
+        {breakpoints &&
+          breakpoints.map(
+            breakpoint => `
+            /* Viewport ${breakpoint.viewport} | Table ${breakpoint.table} | ${breakpoint.columns
```
mnajdova: Can we replace this with a useful comment or remove it?

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+interface ResponsiveTableProps extends TableProps {
+  config: Column[];
+  children: React.ReactElement<typeof Table>;
+  id?: string;
+}
+
+const ResponsiveTableContainer: React.FC<ResponsiveTableProps> = props => {
+  const { config, children: ChildrenTable } = props;
+
+  const [responsiveTableID] = React.useState(props.id || _.uniqueId(`${Table.className}__responsive-wrapper`));
```
mnajdova: ❤️

packages/fluentui/docs/src/prototypes/table/index.tsx
```diff
@@ -157,6 +169,23 @@ const rowsChannels = [
...
+  items: ['id', 'Name', 'Age', 'Picture'],
+};
+
+const rowsPerson = [
+  ['1', 'Roman van von der Longername', '30 years', 'None'],
+  ['2', 'Alex', '1 year', 'None'],
+  ['3', 'Ali', '30000000000000 years', 'None'],
+];
+
+const responsiveColumnsConfig = [
```
mnajdova: Can we add also example where the priority is in ascending order, just for illustration that functionality too

assuncaocharles: Done, I added one where we keep the first and last columns ( Imagining the scenario where the last column would have actions )

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
assuncaocharles: Removed

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
assuncaocharles: Done

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
```diff
@@ -0,0 +1,50 @@
...
+  // Generate the stylesheet once and for all
+  React.useLayoutEffect(() => {
+    setBreakpoints(calculateBreakpoints(config, props.breakpoints || [1200, 960, 600, 280]));
+  }, []);
+
+  return (
+    <div id={responsiveTableID}>
+      <style>
+        {breakpoints &&
+          Object.entries(breakpoints)
```
layershifter: Is not supported in IE11: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

We can use `_.map()` from Lodash here, in this case w will not check for `undefined`

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const ResponsiveTableContainer: React.FC<ResponsiveTableProps> = props => {
+  const { config, children: ChildrenTable } = props;
+
+  const responsiveTableID = React.useMemo(() => props.id || _.uniqueId(`${Table.className}__responsive-wrapper`), [
+    props.id,
+  ]);
+
+  const [breakpoints, setBreakpoints] = React.useState<{ [key: string]: Column[] }>();
```
layershifter:
```suggestion
  const [breakpoints, setBreakpoints] = React.useState<Record<string, Column[]> | undefined>();
```

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+interface ResponsiveTableProps extends TableProps {
+  config: Column[];
+  children: React.ReactElement<typeof Table>;
+  breakpoints?: number[];
+  id?: string;
+}
+
+const ResponsiveTableContainer: React.FC<ResponsiveTableProps> = props => {
+  const { config, children: ChildrenTable } = props;
```
layershifter:
```suggestion
  const { config, children } = props;
```
```suggestion
  const { config, children: childrenTable } = props;
```

Not sure about naming there. Pascalcase is used for element types, not for elements.

packages/fluentui/docs/src/utils/calculateBreakpoints.ts
```diff
@@ -0,0 +1,33 @@
...
+  default?: number;
+};
+
+export type Column = {
+  priority: number;
+  minWidth: number;
+  childIndex?: number;
+};
+
+export type Breakpoint = {
```
layershifter: `Record<string, Column[]>`

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
layershifter: And how the type here is a different from `Breakpoint`? :thinking:

packages/fluentui/docs/src/utils/calculateBreakpoints.ts
```diff
@@ -0,0 +1,33 @@
+export type Breakpoints = {
```
layershifter: `Record<number, number> & { default?: number }`

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
assuncaocharles: `_.toPairs(object)` would make more sense?

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
assuncaocharles: Done

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
assuncaocharles: Done

packages/fluentui/docs/src/utils/calculateBreakpoints.ts
assuncaocharles: Done

packages/fluentui/docs/src/utils/calculateBreakpoints.ts
assuncaocharles: Done

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
assuncaocharles: Done with `_.toPairs(object)`

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react';
+import { TableCell, TableProps, Table } from '@fluentui/react-northstar';
+import calculateBreakpoints, { Column, Breakpoint } from '../../utils/calculateBreakpoints';
+import * as _ from 'lodash';
+
+interface ResponsiveTableProps extends TableProps {
+  config: Column[];
```
jurokapsiar: `columns` or `cols` would be more clear I think

packages/fluentui/docs/src/utils/calculateBreakpoints.ts
```diff
@@ -0,0 +1,28 @@
+export type Breakpoints = Record<number, number> & { default?: number };
```
jurokapsiar: can you move it to the same folder as the prototype? I don't think `docs/*` is exported and this is very specific to the prototype

packages/fluentui/docs/src/prototypes/table/ResponsiveTable.tsx
assuncaocharles: Done

packages/fluentui/docs/src/utils/calculateBreakpoints.ts
assuncaocharles: Done

packages/fluentui/docs/src/prototypes/table/ResponsiveTableContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+  React.useLayoutEffect(() => {
+    setBreakpoints(calculateBreakpoints(config, props.breakpoints || [1200, 960, 600, 280]));
+  }, []);
+
+  return (
+    <div id={responsiveTableID}>
+      <style>
+        {_.toPairs(breakpoints)
+          .map(
+            ([key, breakpoint]: [string, Column[]], i) => `
```
layershifter:
```suggestion
        {_.map(breakpoints, (breakpoint: Column:[], key: string) => `
```

[_.map](https://lodash.com/docs/4.17.15#map) works with objects

packages/fluentui/docs/src/prototypes/table/ResponsiveTableContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const ResponsiveTableContainer: React.FC<ResponsiveTableProps> = props => {
+  const { columns: config, children } = props;
+
+  const responsiveTableID = React.useMemo(() => props.id || _.uniqueId(`${Table.className}__responsive-wrapper`), [
+    props.id,
+  ]);
+
+  const [breakpoints, setBreakpoints] = React.useState<Breakpoint>();
```
layershifter:
```suggestion
  const [breakpoints, setBreakpoints] = React.useState<Breakpoint | undefined>();
```

As you don't have default value for `useState()`

packages/fluentui/docs/src/prototypes/table/ResponsiveTableContainer.tsx
assuncaocharles: Done

packages/fluentui/docs/src/prototypes/table/ResponsiveTableContainer.tsx
assuncaocharles: Done

PR #12361, opened 2020-03-19
fix(Dropdown): refactor event handlers for component and Downshift

silviuavram:
#### Pull request checklist

- [x] Addresses an existing issue: 
Fixes https://github.com/OfficeDev/office-ui-fabric-react/issues/12225
Fixes https://github.com/OfficeDev/office-ui-fabric-react/issues/12314
Fixes https://github.com/OfficeDev/office-ui-fabric-react/issues/12304

#### Description of changes

- stop propagation of change event from reaching Dropdown, since onChange for Dropdown means something different than onChange for Input (value vs query).
- if we get empty search query and we had value set, we also set that to null, since we will consider that if user removed all the contents from the searchQuery, he intends the value to be removed.
- refactor the Downshift handlers. Merged all the changes from our end in `handleOnStateChange` and used types to make decisions. Removed the Downshift handlers of selectedItem and inputValue and only used the state handler.
- improve the unit tests.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12361)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
+++ b/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
@@ -860,10 +860,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
         newState.searchQuery = this.getSelectedItemAsString(changes.selectedItem);
         newState.value = multiple ? [...value, changes.selectedItem] : [changes.selectedItem];
         newState.open = false;
-
-        if (shouldAddHighlightedIndex) {
-          newState.highlightedIndex = items.indexOf(changes.selectedItem);
-        }
+        newState.highlightedIndex = shouldAddHighlightedIndex ? items.indexOf(changes.selectedItem) : null;
 
         if (getA11ySelectionMessage && getA11ySelectionMessage.onAdd) {
           this.setA11ySelectionMessage(getA11ySelectionMessage.onAdd(changes.selectedItem));
@@ -959,6 +956,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
       newState.searchQuery !== undefined && 'onSearchQueryChange',
     ].filter(Boolean) as (keyof DropdownProps)[];
 
+    console.log(type, newState);
     this.setStateAndInvokeHandler(handlers, null, newState);
   };
 
--- a/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
+++ b/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
@@ -441,10 +441,8 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
       <ElementType className={classes.root} onChange={this.handleChange} {...unhandledProps}>
         <Downshift
           isOpen={open}
-          // onChange={this.handleSelectedChange}
-          // onInputValueChange={this.handleSearchQueryChange}
           inputValue={search ? searchQuery : null}
-          stateReducer={this.handleDownshiftStateChanges}
+          stateReducer={this.downshiftStateReducer}
           itemToString={itemToString}
           // downshift does not work with arrays as selectedItem.
           selectedItem={multiple || !value.length ? null : value[0]}
@@ -808,7 +806,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     );
   }
 
-  handleDownshiftStateChanges = (
+  downshiftStateReducer = (
     state: DownshiftState<ShorthandValue<DropdownItemProps>>,
     changes: StateChangeOptions<ShorthandValue<DropdownItemProps>>,
   ) => {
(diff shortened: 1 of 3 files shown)


silviuavram:
Will take a look at the e2e tests and see what's broken.

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -826,31 +817,154 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+      handlers.push('onChange');
+    }
+    if (newState.highlightedIndex !== undefined) {
+      handlers.push('onHighlightedIndexChange');
+    }
+    if (newState.open !== undefined) {
+      handlers.push('onOpenChange');
+    }
+
+    this.setStateAndInvokeHandler(handlers, null, newState);
```
layershifter:
```suggestion
    const handlers: keyof DropdownProps = [
      newState.value !== undefined && 'onChange',
      newState.highlightedIndex !== undefined && 'onHighlightedIndexChange',
      newState.open !== undefined && 'onOpenChange',
    ].filter(Boolean)

    this.setStateAndInvokeHandler(handlers, null, newState);
```

nit

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -826,31 +817,154 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+
+        if (getA11ySelectionMessage && getA11ySelectionMessage.onAdd) {
+          this.setA11ySelectionMessage(getA11ySelectionMessage.onAdd(newValue));
+        }
+
+        if (multiple) {
+          setTimeout(() => (this.selectedItemsRef.current.scrollTop = this.selectedItemsRef.current.scrollHeight), 0);
+        }
+
+        this.tryFocusTriggerButton();
```
layershifter: Can be this and `setTimeout()` be moved to `componentDidUpdate()`?

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -826,31 +817,154 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+      case Downshift.stateChangeTypes.clickButton:
+      case Downshift.stateChangeTypes.keyDownSpaceButton:
+        newState.open = newOpen;
+
+        if (newOpen) {
+          const highlightedIndexOnArrowKeyOpen = this.getHighlightedIndexOnArrowKeyOpen(changes);
+
+          if (_.isNumber(highlightedIndexOnArrowKeyOpen)) {
+            newState.highlightedIndex = highlightedIndexOnArrowKeyOpen;
+          }
```
layershifter: As I `getHighlightedIndexOnArrowKeyOpen()` is used twice in current code, I suggest to modify to provide proper results:

```diff
-  getHighlightedIndexOnArrowKeyOpen = (changes: StateChangeOptions<ShorthandValue<DropdownItemProps>>): number => {
+  getHighlightedIndexOnArrowKeyOpen = (changes: StateChangeOptions<ShorthandValue<DropdownItemProps>>): number | undefined => {
    const { filteredItems, highlightedIndex, value } = this.state;
    const { highlightFirstItemOnOpen, items, multiple, search } = this.props;
    const isArrowUp = changes.type === Downshift.stateChangeTypes.keyDownArrowUp;
    const isArrowDown = changes.type === Downshift.stateChangeTypes.keyDownArrowDown;
    const itemsLength = filteredItems.length;

    if (highlightedIndex) {
      return highlightedIndex;
    }

    if (highlightFirstItemOnOpen) {
      // otherwise, if highlightFirstItemOnOpen prop is provied, highlight first item.
      return 0;
    }

    if (!multiple && !search && value.length > 0) {
      // in single selection, if there is a selected item, highlight it.
      const offset = isArrowUp ? -1 : isArrowDown ? 1 : 0;
      const newHighlightedIndex = items.indexOf(value[0]) + offset;
      if (newHighlightedIndex >= itemsLength) {
        return 0;
      }
      if (newHighlightedIndex < 0) {
        return itemsLength - 1;
      }
      return newHighlightedIndex;
    }

    if (isArrowDown) {
      return 0;
    }
    if (isArrowUp) {
      return itemsLength - 1;
    }

-    return null;
+    return undefined;
  };
```

```suggestion
          newState.highlightedIndex = this.getHighlightedIndexOnArrowKeyOpen(changes);
```

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -826,31 +817,154 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+      case Downshift.stateChangeTypes.keyDownArrowUp:
+        if (newOpen !== undefined) {
+          newState.open = newOpen;
+          newState.highlightedIndex = newHighlightedIndex;
+
+          if (newOpen) {
+            const highlightedIndexOnArrowKeyOpen = this.getHighlightedIndexOnArrowKeyOpen(changes);
+
+            if (_.isNumber(highlightedIndexOnArrowKeyOpen)) {
+              newState.highlightedIndex = highlightedIndexOnArrowKeyOpen;
```
layershifter: With [these changes](https://github.com/microsoft/fluentui/pull/12361/files#r397094666) we can simplify:

```suggestion
            newState.highlightedIndex = this.getHighlightedIndexOnArrowKeyOpen(changes);
```

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
layershifter: Another option is to integrate the check for `newOpen` to `getHighlightedIndexOnArrowKeyOpen` as it already accepts `changes`

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -826,31 +817,154 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+          newState.open = true;
         }
-      } else {
-        newState.highlightedIndex = null;
+
+        break;
       }
+      case Downshift.stateChangeTypes.keyDownEnter:
+      case Downshift.stateChangeTypes.clickItem:
+        const shouldAddHighlightedIndex = !multiple && !!items;
```
layershifter:
```suggestion
        const shouldAddHighlightedIndex = !multiple && _.isNil(items);
```

```suggestion
        const shouldAddHighlightedIndex = !multiple && (items && items.length === 0);
```

This check can be buggy as:

```js
!![] // true
!!undefined // false
```

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -826,31 +817,154 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+      selectedItem: newValue,
+      isOpen: newOpen,
+      highlightedIndex: newHighlightedIndex,
+    } = changes;
+    const newState = {} as DropdownState;
+    const handlers = [];
+
+    switch (type) {
+      case Downshift.stateChangeTypes.changeInput: {
+        const shouldValueChange = newSearchQuery === '' && !multiple && !!value.length;
```
layershifter:
```suggestion
        const shouldValueChange = newSearchQuery === '' && !multiple && value.length > 0;
```

nit

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -826,31 +817,154 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
-          newState.highlightedIndex = highlightedIndexOnArrowKeyOpen;
+    const { search, multiple, highlightFirstItemOnOpen, items, getA11ySelectionMessage } = this.props;
+    const { value, open } = this.state;
+    const {
+      type,
+      inputValue: newSearchQuery,
+      selectedItem: newValue,
+      isOpen: newOpen,
+      highlightedIndex: newHighlightedIndex,
+    } = changes;
```
layershifter: nit: I don't think that we should destructure there as it hard to follow that comes from `props`/`state` & `downshift.changes`

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
silviuavram: `const shouldAddHighlightedIndex = !multiple && (items && items.length);`

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
silviuavram: Since it's not straightforward to do and not included in scope, I'd rather leave it like this for now.

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
silviuavram: it will set state to undefined for highlightedIndex, not what we want, but most probably this code can be improved. For this PR I just tried to bring all the handling for state change of Downshift into one place, so it can be read better. Maybe we will improve the code in the future, for now I think it's good enough.

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
silviuavram: reverted this and used changes.variable. thanks!

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
layershifter: Agree, let's leave it here for now, but at least consider this for future in other components, too

Issue #11534, opened 2019-12-21
ChoiceGroupOption is exported in two places and that is causing errors


sushruth:
### Environment Information

- **Package version(s)**: 7.76.0
- **Browser and OS versions**: Any browser / Win 10

### Please provide a reproduction of the bug in a codepen: N/A

Related to #10017 and #9661 - 

Looks like ChoiceGroupOptions is exported twice in this hierarchy - 

```
office-ui-fabric-react/src/index.ts

    export * from './ChoiceGroup.ts';
    export * from './ChoiceGroupOption.ts';

office-ui-fabric-react/src/ChoiceGroupOption.ts

    export * from './components/ChoiceGroup/ChoiceGroupOption/index.tsx'; // here

office-ui-fabric-react/src/ChoiceGroup.ts

    export * from './components/ChoiceGroup/index.tsx';

office-ui-fabric-react/src/components/ChoiceGroup/index.ts

    export * from './ChoiceGroup.tsx';
    export * from './ChoiceGroup.base.tsx';
    export * from './ChoiceGroup.types.tsx';
    export * from './ChoiceGroupOption/index.tsx'; // here

office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/index.ts

    export * from './ChoiceGroupOption.tsx';
    export * from './ChoiceGroupOption.types.tsx';
```

This is leading to similar error as #10017 and #9661 - 

```
Uncaught TypeError: Cannot redefine property: ChoiceGroupOption
```

#### Actual behavior:

Error should not be present.

#### Expected behavior:

compile process shows errors.

### Priorities and help requested:

Are you willing to submit a PR to fix? Yes

Requested priority: High

Products/sites affected: `jest` tests and `parcel` bundled projects.

ecraig12345:
It's odd that we have a top-level file for ChoiceGroupOption. It's been there long enough (was added over a year ago in #6492) that I'm hesitant to remove it, but I can get rid of the export from it in the main index file.

sushruth:
@ecraig12345 Thank you! There are actually multiple such exports within the whole codebase. Doing a pass for all such instances could be great for `jest` and `parcel` users and also IMO feels a tiny bit saner.

ecraig12345:
Do you have a list of the duplicate exports or know how to find them? (This isn't something we've run into with our toolchain internally.)

sushruth:
[...]

PR #12364, opened 2020-03-19
Creating `@fluentui/react` package

dzearing:
This PR introduces the `@fluentui/react` v7 package. This stages the new name for our current bits as we transition our code to Fluent UI.

This code currently exports bits from `office-ui-fabric-react` so if you're using Fabric today, you may change your imports with no breaking changes going forward.

Before:

```tsx
import { PrimaryButton } from 'office-ui-fabric-react';
import { PrimaryButton } from 'office-ui-fabric-react/lib/Button';
```

After:
```tsx
import { PrimaryButton } from '@fluentui/react';
import { PrimaryButton } from '@fluentui/react/lib/Button';
```

NOTE: this should not affect any code dependent on `office-ui-fabric-react` today. It will simply stage the ability for partners to refer to our LTS releases using the `@fluentui/react` package name.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12364)


packages/react/src/ChoiceGroupOption.ts
```diff
@@ -0,0 +1 @@
+export * from 'office-ui-fabric-react/lib/ChoiceGroupOption';
```
ecraig12345: Let's get rid of this file in the new version. ChoiceGroupOption is also exported from the ChoiceGroup file, and having it in two places causes issues like #11534.

packages/react/src/Dialog.ts
```diff
@@ -0,0 +1,2 @@
+export * from 'office-ui-fabric-react/lib/Dialog';
+export { Dialog as default } from 'office-ui-fabric-react/lib/Dialog';
```
ecraig12345: What is this doing here?

packages/react/src/Modal.ts
```diff
@@ -0,0 +1,2 @@
+export * from 'office-ui-fabric-react/lib/Modal';
+export { Modal as default } from 'office-ui-fabric-react/lib/Modal';
```
ecraig12345: Same here

packages/react/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@fluentui/react",
+  "version": "7.105.1",
```
ecraig12345: Be sure to check if this gets updated before you merge

packages/react/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "forceConsistentCasingInFileNames": true,
+    "strictNullChecks": true,
+    "noImplicitAny": true,
+    "moduleResolution": "node",
+    "preserveConstEnums": true,
+    "lib": ["es5", "dom"],
+    "types": ["jest", "webpack-env"],
+    "paths": {
+      "@uifabric/react/lib/*": ["./src/*"],
+      "@uifabric/react": ["./src"]
```
ecraig12345:
```suggestion
      "@fluentui/react/lib/*": ["./src/*"],
      "@fluentui/react": ["./src"]
```

packages/react/webpack.config.js
```diff
@@ -0,0 +1,65 @@
+let path = require('path');
+const resources = require('../../scripts/webpack/webpack-resources');
```
ecraig12345:
```suggestion
const resources = require('@uifabric/build/webpack/webpack-resources');
```

packages/react/webpack.config.js
```diff
@@ -0,0 +1,65 @@
...
+const ManifestServicePlugin = require('@uifabric/webpack-utils/lib/ManifestServicePlugin');
+
+const BUNDLE_NAME = 'fluent-ui-react';
```
ecraig12345: Should this be `fluentui-react` to more closely match the package name?

packages/react/webpack.serve.config.js
```diff
@@ -0,0 +1,25 @@
+const path = require('path');
+const resources = require('@uifabric/build/webpack/webpack-resources');
+
+module.exports = resources.createServeConfig({
```
ecraig12345: Delete this file probably, since it won't work

packages/react/webpack.config.js
dzearing: interesting suggestion.

packages/react/README.md
```diff
@@ -0,0 +1,11 @@
+# [Fluent UI React](https://dev.microsoft.com/fabric)
+
+**The React-based front-end framework for building experiences for M365.**
```
ecraig12345: Probably do this? M365 is a bit jargon-y. :) But this new version is much better overall!
```suggestion
**The React-based front-end framework for building experiences for Microsoft 365.**
```

packages/react/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+**Fluent UI 7** Summary, breaking changes, and more details available in [the wiki](https://github.com/OfficeDev/office-ui-fabric-react/wiki/Fabric-7).
+
+Fluent UI React is a collection of robust React-based components designed to make it simple for you to create consistent web experiences using the Office Design Language.
```
ecraig12345:
```suggestion
Fluent UI React is a collection of robust React-based components designed to make it simple for you to create consistent web experiences using the Fluent Design Language.
```

ecraig12345: Bonus: add a link to the fluent design site

packages/react/src/version.ts
```diff
@@ -0,0 +1,4 @@
+// @uifabric/react@0.1.0
```
ecraig12345: Guessing this file will be overwritten/corrected when the real version is generated, but wouldn't hurt to update it for initial checkin as well.

dzearing: maybe i should also change the scope name. :)

packages/react/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "description": "Reusable React components for building experiences for M365.",
```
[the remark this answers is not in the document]
dzearing: Resolved with 
  "description": "Reusable React components for building web experiences.",

packages/react/src/ChoiceGroupOption.ts
dzearing: That would be a breaking change.

packages/react/src/Dialog.ts
dzearing: Honestly not sure why we have a default export here.

PR #12370, opened 2020-03-20
Minor script fixes

ecraig12345:
For all:
- Eliminate usage of `--parallel` in main package.json scripts because it tries to run absolutely everything in parallel and pegs the CPU, which is not very useful

For v0 scripts:
- When cleaning, invoke `del` with `{ force: true }`, which means delete things even if it's outside the current working directory. (This comes up when running commands within individual packages, but deleting something like component info which lives in a different package.)
- Fix the test script in the accessibility package to not run test before test:watch.
- Disable webpack.ProgressPlugin output during multi-project build since it spams logs

For v7 scripts:
- Fix types in webpack-resources
- Improve wording in update release notes script for when there are no changes recorded in changelog: instead of "Changes not tracked" (which is confusing), say "No change notes provided"

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12370)


scripts/webpack/webpack.config.ts
```diff
@@ -103,7 +103,10 @@ const webpackConfig: webpack.Configuration = {
       contextRegExp: /moment$/,
     }),
     __DEV__ &&
```
xugao: just checking: you mention in the comment "in CI" but here checks for dev environment.

PR #12371, opened 2020-03-20
feat(bindings): add useUnhandledProps hook

layershifter:
#### Pull request checklist

#### Description of changes

This PR adds `useUnhandledProps()` hook to replace `getUnhandledProps()` function in functional components. Functionality of `useUnhandledProps()` will be changed in `compose()` PR.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12371)

### DIFF
--- a/packages/fluentui/react-bindings/src/hooks/useUnhandledProps.ts
+++ b/packages/fluentui/react-bindings/src/hooks/useUnhandledProps.ts
@@ -1,0 +1,19 @@
+import getUnhandledProps from '../utils/getUnhandledProps';
+
+/**
+ * Returns an object consisting of props beyond the scope of the Component.
+ * Useful for getting and spreading unknown props from the user.
+ *
+ * @param handledProps - An array with names of props
+ * @param props - A ReactElement props object
+ * @returns A shallow copy of the prop object
+ */
+function useUnhandledProps<P extends Record<string, any>>(handledProps: (keyof P)[], props: P): Partial<P> {
+  if (process.env.NODE_ENV === 'test') {
+    return getUnhandledProps(handledProps, { ...props, 'data-uses-unhanded-props': true });
+  }
+
+  return getUnhandledProps(handledProps, props);
+}
+
+export default useUnhandledProps;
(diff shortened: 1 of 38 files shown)


packages/fluentui/react-bindings/src/hooks/useUnhandledProps.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Returns an object consisting of props beyond the scope of the Component.
+ * Useful for getting and spreading unknown props from the user.
+ *
+ * @param handledProps - An array with names of props
+ * @param props - A ReactElement props object
+ * @returns A shallow copy of the prop object
+ */
+function useUnhandledProps<P extends Record<string, any>>(handledProps: (keyof P)[], props: P): Partial<P> {
```
mnajdova: Can we add example in the react-bindings README? Or do you consider this as an internal hook?

packages/fluentui/react-bindings/src/index.ts
```diff
@@ -15,6 +15,7 @@ export { default as unstable_useDispatchEffect } from './hooks/useDispatchEffect
 export { default as useIsomorphicLayoutEffect } from './hooks/useIsomorphicLayoutEffect';
 export { default as useStateManager } from './hooks/useStateManager';
 export { default as useStyles } from './hooks/useStyles';
+export { default as useUnhandledProps } from './hooks/useUnhandledProps';
```
mnajdova: As we are exporting it, I believe it deserves changelog entry

packages/fluentui/react-northstar/src/components/Animation/Animation.tsx
```diff
@@ -231,7 +231,7 @@ const Animation: React.FC<AnimationProps> & {
   const { animationDuration, animationDelay } = animationStyles.root;
   const timeoutResult = timeout || calculateAnimationTimeout(animationDuration, animationDelay) || 0;
 
-  const unhandledProps = getUnhandledProps(Animation.handledProps, props);
+  const unhandledProps = useUnhandledProps(Animation.handledProps, props);
```
mnajdova: Should we use the generic typings?

layershifter: Nope, they will be inferred as it was previously for `getUnhandledProps`

packages/fluentui/react-bindings/src/index.ts
layershifter: Agree, added

packages/fluentui/react-bindings/src/hooks/useUnhandledProps.ts
layershifter: Added 👍

PR #12374, opened 2020-03-20
Measure bundle size of individual react-northstar components

miroslavstastny:
#### Pull request checklist

### TODO
- [x] Update `yarn stats:build` pipeline
- [x] Add support for `bsize` examples
- [ ] Add `bsize` examples for all components
- [x] Add charts for `bsize` examples
 
### Description of changes

Measure bundle size of individual `react-northstar` components.
It does not make much sense to measure the bundle size of the component alone as it is always used with a theme.
This PR adds explicit bundle size examples to docsite and measures bundle size of these examples.
The example renders single component wrapper by a `Provider` with `teams` light theme:

```jsx
const ButtonDefaultBsize = () => (
  <Provider theme={themes.teams}>
    <Button content="Click here" />
  </Provider>
);
```

#### What is in the bundle which is measured?
image

#### Why?
This should give information not only about the component size but mainly about treeshakability of the library. The absolute number is meaningful. It is the trend which is important

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12374)


packages/fluentui/docs/src/components/ComponentDoc/ComponentBundleSizeExample/ComponentBundleSizeExample.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+import { Accordion, Segment } from 'src/index';
```
layershifter:
```suggestion
import { Accordion, Segment } from '@fluentui/react-northstar';
```

packages/fluentui/docs/src/components/ComponentDoc/ComponentBundleSizeExample/ComponentBundleSizeExample.tsx
```diff
@@ -0,0 +1,46 @@
...
+import ComponentExampleTitle from 'docs/src/components/ComponentDoc/ComponentExample/ComponentExampleTitle';
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample';
```
layershifter: Let's usage absolute paths

packages/fluentui/docs/src/components/ComponentDoc/ComponentBundleSizeExample/ComponentBundleSizeExample.tsx
```diff
@@ -0,0 +1,46 @@
...
+                content: 'Show example',
+                styles: ({ theme }) => {
+                  return {
+                    fontSize: theme.siteVariables.fontSizes.smaller,
+                  };
+                },
+              },
+              content: {
+                key: 'c',
+                content: <ComponentExample {..._.omit(props, 'title', 'description')} resetTheme />, // resetTheme to make sure the bundle size example contains the theme
```
layershifter: As I see `ComponentBundleSizeExampleProps` is not extendable, so:

```suggestion
                content: <ComponentExample examplePath={examplePath} resetTheme />, // resetTheme to make sure the bundle size example contains the theme
```

scripts/webpack/webpack.config.stats.ts
```diff
@@ -55,13 +58,21 @@ const makeConfig = (srcPath: string, name: string): webpack.Configuration => ({
     react: 'react',
     'react-dom': 'reactDOM',
   },
+  optimization: {
+    concatenateModules: false,
```
layershifter: How much this affects result bundle size? I think that it can produce false-negative measures

packages/fluentui/docs/src/components/ComponentDoc/ComponentBundleSizeExample/ComponentBundleSizeExample.tsx
layershifter: This will reset only React's context, BTW

scripts/webpack/webpack.config.stats.ts
```diff
@@ -55,13 +58,21 @@ const makeConfig = (srcPath: string, name: string): webpack.Configuration => ({
...
+  },
   plugins: [
     new CleanWebpackPlugin([paths.base('stats')], {
       root: paths.base(),
       verbose: false, // do not log
     }),
     new IgnoreNotFoundExportPlugin(),
     new webpack.DefinePlugin(config.compiler_globals),
+    // new BundleAnalyzerPlugin({
```
layershifter: It will be really useful to store reports as artifacts

scripts/webpack/webpack.config.stats.ts
miroslavstastny: WIth concatenated modules, the bundle analyzer output is not much useful.
The numbers are not the same, but seem to be pretty close and still representative:
image

packages/fluentui/docs/src/components/ComponentDoc/ComponentBundleSizeExample/ComponentBundleSizeExample.tsx
miroslavstastny: I think it is good enough for this case - If there will be no `Provider` in the example, it will be visually broken.

scripts/webpack/webpack.config.stats.ts
layershifter: In this case: go 🏎

layershifter:
@miroslavstastny Should bundle size examples use Screener? I don't think so, but I interested in our opinion.

scripts/webpack/webpack.config.stats.ts
miroslavstastny: Yup, I plan to do that in PR gate work

miroslavstastny:
> @miroslavstastny Should bundle size examples use Screener? I don't think so, but I interested in our opinion.

I thought about it when adding perf examples some time ago. The reason why these examples are in Screener is that if you broke the example, you would start measuring non-sense. And screener regression is the safety net for that.
It will add some additional time to Screener but we will have tens of bsize examples vs more than a thousand of regular screener examples = 1% time increase - I think it is worth it.

packages/fluentui/docs/src/components/ComponentDoc/PerfChart/BundleSizeChart.tsx
```diff
@@ -0,0 +1,143 @@
+import * as React from 'react';
+import * as _ from 'lodash';
+import {
+  DecorativeAxis,
+  FlexibleXYPlot,
+  HorizontalGridLines,
+  LabelSeries,
+  LineSeries,
+  XAxis,
+  // YAxis,
```
layershifter: cruft?

packages/fluentui/docs/src/components/ComponentDoc/PerfChart/BundleSizeChart.tsx
```diff
@@ -0,0 +1,143 @@
...
+      />
+    ));
+
+  return (
+    <FlexibleXYPlot
+      onMouseLeave={() => {
+        setNearestX(undefined);
+      }}
+    >
+      {/* <YAxis title="B" /> */}
```
layershifter: Cruft?

packages/fluentui/docs/src/components/ComponentDoc/PerfChart/BundleSizeChart.tsx
```diff
@@ -0,0 +1,143 @@
...
+ */
+const BundleSizeChart: React.FC<BundleSizeChartProps> = ({ perfData }) => {
+  const availableCharts: string[] = perfData
+    .reduce((acc, next) => {
+      return Array.from(new Set([...acc, ...Object.keys(next.bundleSize)]));
+    }, [] as string[])
+    .sort();
+
+  // @ts-ignore
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
layershifter:
```suggestion
```

I don't think that it's needed

packages/fluentui/docs/src/examples/components/Box/Performance/BoxDefault.bsize.tsx
```diff
@@ -0,0 +1,19 @@
...
+        color: 'blue',
+        textAlign: 'center',
+        width: '50px',
+        ':hover': { color: 'red' },
+      }}
+    />{' '}
+  </Provider>
+);
+
+export default BoxShorthandExample;
```
layershifter:
```suggestion
export default BoxExampleBSize;
```

packages/fluentui/docs/src/examples/components/Box/Performance/BoxDefault.bsize.tsx
```diff
@@ -0,0 +1,19 @@
...
+  <Provider theme={themes.teams}>
+    <Box
+      content="Box"
+      styles={{
+        border: '1px dashed #ccc',
+        color: 'blue',
+        textAlign: 'center',
+        width: '50px',
+        ':hover': { color: 'red' },
+      }}
```
layershifter:
```suggestion
```

packages/fluentui/docs/src/components/ComponentDoc/PerfChart/useBundleSizeData.ts
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
+import * as _ from 'lodash';
+import PerfDataContext from './PerfDataContext';
+
+const useBundleSizeData = (filter?: string) => {
```
layershifter: I suggest to switch later to [`react-query`](https://github.com/tannerlinsley/react-query) and slice data

image
image

Currently it takes 1 min to get an answer from backend and around 1 min to fetch data 💣

PR #12378, opened 2020-03-20
Replace OfficeDev/office-ui-fabric-react with microsoft/fluentui

ecraig12345:
Replace OfficeDev/office-ui-fabric-react with microsoft/fluentui across all files.

Remove OfficeDev/uifabric-calendar team from codeowners.

PR #12383, opened 2020-03-21
Group publishing for @fluentui/react and office-ui-fabric-react packages

xugao:
#### Pull request checklist

#### Description of changes
- Use beachball [version groups](https://microsoft.github.io/beachball/version-groups) feature to bump `@fluentui/react` and `office-ui-fabric-react` packages with exact same pace
- Bumped beachball to latest which contains couple version groups fixes
- Fixed a change file which has incorrect `dependentChangeType`
- Changed to only disallow major change type for `@fluentui/react` and `office-ui-fabric-react` packages

**Note**: currently `@fluentui/react` and `office-ui-fabric-react` are published to different latest versions. they should have same version when next time one of the package does a `minor` bump

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12383)

### DIFF
--- a/change/@fluentui-react-2020-03-20-12-24-02-remainder-pr-feedback.json
+++ b/change/@fluentui-react-2020-03-20-12-24-02-remainder-pr-feedback.json
@@ -4,6 +4,6 @@
   "packageName": "@fluentui/react",
   "email": "dzearing@microsoft.com",
   "commit": "b00a4fb9a29289adea7cb85463d5fd0b7923c226",
-  "dependentChangeType": "patch",
+  "dependentChangeType": "none",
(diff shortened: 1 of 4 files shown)


ecraig12345:
@xugao This was failing because it needed the changes from #12378. I just merged that, so it should mostly succeed now.

Screener will continue to not report status due to an outstanding issue, so for now we'll have to manually look at reports from the link in the build output. Though for your PR you can ignore that issue since screener isn't relevant.

xugao:
@ecraig12345 thanks!!

PR #12389, opened 2020-03-23
RFC(icon): update icon slots

mnajdova:
# Icon properties

## Problem description
Currently for `@fluentui/react-northstar` we use shorthands for `Icon` component to inject icons. 

The `Icon` component uses the `name` prop to lookup for SVGs in themes and requires to have all icons registered in a giant single object. As objects are not treeshakable and growing amount of icons this lead to bigger bundle size.

## Proposal
We introduced createSvgIcon factory (https://github.com/microsoft/fluentui/pull/12319) to move out SVG icons from a theme object to a dedicated component per icon. This opens a question how to handle existing Icon shorthands. 

It's impossible to use existing approach as there is no single component that handles icons, so there are two alternatives.

### Proposal 1 (icon as `Box` shorthand)
```jsx
<Button icon={{content: <SoundIcon/>}} content="Sound on"/>
```
Edited per @layershifter 's comment for clarifying confusion. The following code will result in the same as the above code
```jsx
<Button icon={<SoundIcon />} content="Sound on" />
```
This proposal defined for each icon slot a `Box` shorthand, where the users can specify the icon they want to use inside the content of the `Box`. The `Box` component is used here for simplification, we may even expend this by introducing `ButtonIcon` component where the slot may even support some additional props that make sense of the context where it is used, the `Button` in the example above.

If we go with this proposal, inside the button styles we will have dedicated slot for the icon's box, where we can add some additional styles for the icon. For example:

```jsx
const buttonStyles = {
  root: () => ({...}),
  icon: ({props: p}) => ({
    // styles for the icon box
    lineHeight: pxToRem(12),
    height: pxToRem(16),
    
    // we may target some icon's specific parts
    '& svg': {
      height: pxToRem(16),
    },
    
    // we may add some styles based on props
    ...(p.hasContent && {
      margin: `0 ${pxToRem(10)} 0 0`,
    })
  }),
}

```

Pros:
- keeping it consistent with the othar API that we have so far regarding slots
- reusing styles with potentional dedicated component (See `ButtonIcon proposal below`)

Cons:
- additional element in the DOM

This pattern is used in:
- Material UI - https://material-ui.com/components/buttons/#IconLabelButtons.js (the API looks like the Proposal 2, but the implementation is with element wrapping the icon)

### Proposal 2 (icon as JSX element)
```jsx
<Button icon={<SoundIcon/>} content="Sound on"/>
```
With this proposal, we will render the element specified in the icon slot, with adding additional custom className, for being able to target it consistently. 

Pros:
- no additional element in the DOM
- will work with whatever icons library we choose, as we will have one selector for the icons

Cons:
- if we absolutely need to style the icon, we will have to use selector (if we avoid style overrides for this - which is doable, this may not even be a problem)

This pattern is used in:
- Ant design - https://ant.design/components/button/
- Evergreen - https://evergreen.segment.com/components/button/
- Semantic UI react - https://react.semantic-ui.com/elements/button/#types-icon

### Othar notes
In addition to this any of the proposals above, we may even consider adding new components for the icons shorthands, like `ButtonIcon` component, which will allow the users to use children API too. 

## Final comment
This PR will not be merged as is, and the changes provided are just for showcase od the different proposals.

### DIFF
--- a/packages/fluentui/docs/package.json
+++ b/packages/fluentui/docs/package.json
@@ -8,6 +8,7 @@
     "copy-to-clipboard": "^3.2.0",
     "faker": "^4.1.0",
     "flat": "^4.1.0",
+    "grommet-icons": "^4.4.0",
     "react-ace": "^5.1.2",
     "react-codesandboxer": "^3.1.3",
     "react-custom-scrollbars": "^4.2.1",
@@ -18,7 +19,8 @@
     "react-router-dom": "^5.0.1",
     "react-source-render": "4.0.0-1",
     "react-virtualized": "^9.21.1",
-    "react-vis": "^1.11.6"
+    "react-vis": "^1.11.6",
+    "styled-components": "^5.0.1"
   },
   "devDependencies": {
     "@types/color": "^3.0.0",
--- a/packages/fluentui/docs/package.json
+++ b/packages/fluentui/docs/package.json
@@ -15,6 +15,7 @@
     "react-document-title": "^2.0.3",
     "react-element-to-jsx-string": "^14.0.2",
     "react-hot-loader": "^4.12.15",
+    "react-icons": "^3.9.0",
     "react-markdown": "^4.0.8",
     "react-router-dom": "^5.0.1",
     "react-source-render": "4.0.0-1",
--- a/packages/fluentui/docs/src/components/Playground/renderConfig.ts
+++ b/packages/fluentui/docs/src/components/Playground/renderConfig.ts
@@ -7,6 +7,7 @@ import * as _ from 'lodash';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import * as Classnames from 'classnames';
+import * as GrommetIcons from 'grommet-icons';
 
 const accessibilityPackageJson = require('@fluentui/accessibility/package.json');
 const docsComponentsPackageJson = require('@fluentui/docs-components/package.json');
@@ -63,6 +64,10 @@ export const imports: Record<string, { version: string; module: any }> = {
   prettier: {
     version: docsComponentsPackageJson.peerDependencies['prettier'],
     module: null, // no need to use it in our examples
+  },
+  'grommet-icons': {
+    version: docsComponentsPackageJson.dependencies['grommet-icons'],
+    module: GrommetIcons,
   },
 };
 
--- a/packages/fluentui/docs/src/components/Playground/renderConfig.ts
+++ b/packages/fluentui/docs/src/components/Playground/renderConfig.ts
@@ -8,6 +8,7 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import * as Classnames from 'classnames';
 import * as GrommetIcons from 'grommet-icons';
+import * as ReactIconsFa from 'react-icons/fa';
 
 const accessibilityPackageJson = require('@fluentui/accessibility/package.json');
 const docsComponentsPackageJson = require('@fluentui/docs-components/package.json');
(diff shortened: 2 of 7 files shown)


layershifter:
I think that it will be the same story, but can we also test [`react-icons`](https://www.npmjs.com/package/react-icons) with both approaches?

To `Box` approach: is it possible to use `as` prop? I.e. `<Button icon={{ as: BookIcon }} />`.
If so, it also opens third option: `<Button icon={BookIcon} />`

mnajdova:
> I think that it will be the same story, but can we also test react-icons with both approaches?

Added, didn't need to change any of the styles, works out of the box. 👍 

> To Box approach: is it possible to use as prop? I.e. <Button icon={{ as: BookIcon }} />.
If so, it also opens third option: <Button icon={BookIcon} />

Will try this tomorrow, good point! There may open some typings issues, but will check.

mnajdova:
> To Box approach: is it possible to use as prop? I.e. <Button icon={{ as: BookIcon }} />.
> If so, it also opens third option: <Button icon={BookIcon} />

@layershifter I tried this approach, but there are some issues, with typings/itellisense... Basically we don't have intellisense to any of the icon component's props. Plus It seems a bit awkward to just pass the component :\ I haven't see an API like this anywhere, so I would rather avoid it, as it is in the end possible with the Box shorthand anyway. What do you think?

dzearing:
#### `<Foo icon={ <FooIcon /> }`

:+1: Love this. Feels pretty ergonomic.

Here are some things to consider:

Right now, the icon `className` owned by `Foo` is injected onto `FooIcon`. You are using `React.cloneElement` to clone the element with the extra class name processing.

* :warning: Potential perf concern (extra work done in bulk will add up.) Every time a custom slot is rendered, you must `cloneElement` just to pass through the class name. You could avoid this by always having a wrapper that applies the class rather than slot content (which is how web component slots work) but yeah, extra wrapper. :-1:
* :warning: Can be problematic where you're wrapping content with some utility (or partial) that might not forward the `className` prop:

`<Foo icon={{ <Tooltip><CoolIcon /></Tooltip> }}`

The workaround would be to use the `children` workaround for these edge cases:

`<Foo icon={{ children: (C, p) => <Route><CoolIcon className={p.className} /></Route> }}`

It would be a bit more ergonomic to just use a function:

```tsx
<Foo icon={ ({ props }) => (
  <Tooltip>
    <CoolIcon className={props.className} />
  </Tooltip>
} /> 
```
(It would also be nice to be able to access parent state and slot component through the function.)

layershifter:
I prefer proposal 1 and go with `Box`.

---

```jsx
<>
  <Button icon={<SoundIcon />} content="Sound on" />
  <Button icon={{ content: <SoundIcon /> }} content="Sound on" />
</>
```

Currently will produce the same result (shorthands handle JSX elements as `{ content: element }`:

```html
<button class="ui-button">
  <span class="ui-button__icon">
    <span class="ui-icon">Actual icon element</span>
  </span>
  <span class="ui-button__content">Sound on</span>
</button>
```

### Cons

👎 Will have an additional DOM-element in tree.

### Pros

👍 The same API, no different behaviors
👍 No `cloneElement()`, no issues
👍 `<Button icon={<Tooltip><SoundIcon /></Tooltip>} />` is easy as classes are applied on `Box`

miroslavstastny:
It would be interesting to see the code for having outline icon which fills on button hover :-)

For me the question - is it less worse to have wrapping `Box` or deal with `cloneElement` limitations.
I see the wrapping `Box` as a smaller problem, you can get rid of it by render children prop.

mnajdova:
We decided to go with the `ShorthandValue<BoxProps>`. See https://github.com/microsoft/fluentui/pull/12489 for implementation details. Closing this as it was resolved.

PR #12395, opened 2020-03-23
Dialog/Panel: Change heading from 2 to 1

joschect:
#### Pull request checklist

#### Description of changes

Both panels and dialogs are entirely new contexts and should reset the heading level.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12395)


aneeshack4:
Looks like the snapshots just need to be updated!

joschect:
@aneeshack4 yeah! I've been trying to update it but keep getting the following error:
image

I've tried debugging it and it seems that the full path isn't resolving correctly at this line 
https://github.com/microsoft/fluentui/blob/master/scripts/jest/jest-resources.js#L7 . I see that for whatever reason, `@uifabric/jest-serializer-merge-styles` seems to get to turned into `@uifabric` at some point.

joschect:
I've also tried it from a fresh clone of the repo and `yarn update-snapshots` fails even then.

aneeshack4:
@joschect Looks like you fixed it! Size auditor is failing for everyone now so that's unrelated

PR #12405, opened 2020-03-24
feat: add a shorthand to specify Popper props

layershifter:
#### Pull request checklist

#### Description of changes

- adds unit tests to cover created scenarios
- exposes `positionFixed` prop on `Popup` component
- exposes `positionFixed` and `unstable_pinned` props on `Tooltip` component
- modifies shorthands on components listed below to add `popper` into them that allows to tweak props of `Popper` component

| Component         | Shorthand    |
| ----------------- | ------------ |
| `ChatMessage`     | `actionMenu` |
| `Dropdown`        | `list`       |
| `MenuItem`        | `menu`       |
| `ToolbarItem`     | `menu`       |
| `ToolbarMenuItem` | `menu`       |


packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
```diff
@@ -145,6 +145,7 @@ const Popup: React.FC<PopupProps> &
     offset,
     pointing,
     position,
+    positionFixed,
```
jurokapsiar: do we need to do the same for `MenuButton`?

layershifter: Good catch, added 👍

packages/fluentui/react-northstar/test/specs/commonTests/implementsShorthandProp.tsx
```diff
@@ -66,5 +69,27 @@ export default ((Component: React.ComponentType) => {
         });
       });
     });
+
+    if (options.implementsPopper) {
+      describe('implements all positioning props for Popper', () => {
+        Object.keys(positioningProps).forEach((positioningProp: keyof PositioningProps) => {
```
silviuavram: nit: `Object.entries` since you also use value in a couple of places, but it's not tragic :D

silviuavram: And there was another place similar to this one.

PR #12407, opened 2020-03-24
feat(Dropdown): support disabled items

silviuavram:
#### Pull request checklist

#### Description of changes

Downshift already supports handling `disabled` as a key in the item object. We need to pass that from our `Dropdown` to `getItemProps` in order to make it work.

Fully functional disabled item navigation should be achievable by updating to Downshift https://github.com/downshift-js/downshift/releases/tag/v4.0.1 or above. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12407)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
+++ b/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
@@ -743,7 +743,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
                 key: (item as any).header,
               }),
           }),
-          overrideProps: this.handleItemOverrides(item, index, getItemProps, selected),
+          overrideProps: this.handleItemOverrides(item, index, getItemProps, selected, !!item['disabled']),
           render: renderItem,
         });
       },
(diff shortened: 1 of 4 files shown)


packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -991,7 +992,11 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     toggleMenu: () => void,
   ): void => {
     if (this.state.open) {
-      if (!_.isNil(highlightedIndex) && this.state.filteredItems.length) {
+      if (
+        !_.isNil(highlightedIndex) &&
+        this.state.filteredItems.length &&
+        !this.props.items[highlightedIndex]['disabled']
```
layershifter: It will not work with old render callback, but it's already deprecated:

```tsx
const items = [render => render("foo")];
```

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -867,6 +867,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
       ...getItemProps({
         item,
         index,
+        disabled: !!item['disabled'],
```
layershifter:
```suggestion
        disabled: item['disabled'],
```

Is it really required?

silviuavram: double checked in downshift and it's not. will remove, thanks!

PR #12412, opened 2020-03-25
[WIP] Enable faster rule caching for fela renderer

davezuko:
Opening this PR up early so I can get feedback from the PR bots on performance impacts.

Updates fela's `renderRule` method to compute classNames on a more granular level: by property rather than by style object.

Rather than generating a full className for an entire style object at once, it now walks the style object and generates classes for each individual property.  This allows us to avoid reprocessing styles that should already be cached.

NOTE: this approach does not work for monolithic classNames, since that treats an entire style object as a single entity. So, do not use this if you are using that mode.

#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12412)


davezuko:
Closing in favor of https://github.com/microsoft/fluentui/pull/12445.

PR #12418, opened 2020-03-25
fix(Dropdown): fix hover styles for color

silviuavram:
#### Pull request checklist

#### Description of changes

Override values for hovered disabled color on triggerButton for Dropdown. Right now if you hover over disabled dropdown the text color changes. It should not.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12418)


packages/fluentui/react-northstar/src/themes/teams/components/Dropdown/dropdownVariables.ts
```diff
@@ -102,7 +102,7 @@ export default (siteVars): DropdownVariables => ({
   overlayZIndex: siteVars.zIndexes.overlay,
   // disabled state
   disabledBorderColorHover: 'transparent',
-  disabledTriggerColorHover: siteVars.bodyColor,
+  disabledTriggerColorHover: siteVars.colorScheme.brand.foregroundDisabled,
```
layershifter: Do we have a screener test for this?

silviuavram: now we do!👍

PR #12432, opened 2020-03-26
feat(SplitButton): add SplitButtonToggle component

mnajdova:
This PR replaces the `Button` used as `toggle` inside the `SplitButton` with new `SplitButtonToggle` component. With this, we are able to restrict the props inside the toggle to only those which make sense for this usage. We have default styles for the chevron as part of the styles too. As a bonus we are gaining around 23% performance gain:

image

# BREAKING CHANGES
`SplitButton.slotClassNames` was removed. You should replace `SplitButton.slotClassNames.toggleButton` with `SplitButtonToggle.className`.

The shorthand for the `toggleButton` is now of type `SplitButtonToggleProps`, which is restricted set of the `ButtonProps`

### DIFF
--- a/packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
+++ b/packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
@@ -16,6 +16,7 @@ import {
   ShorthandFactory,
   SizeValue,
 } from '../../utils';
+import SplitButtonToggle, { SplitButtonToggleProps } from './SplitButtonToggle';
 import Button, { ButtonProps } from '../Button/Button';
 import MenuButton, { MenuButtonProps } from '../MenuButton/MenuButton';
 import { MenuProps } from '../Menu/Menu';
@@ -81,7 +82,7 @@ export interface SplitButtonProps extends UIComponentProps, ChildrenComponentPro
   size?: SizeValue;
 
   /** Shorthand for the toggle button. */
-  toggleButton?: ShorthandValue<ButtonProps>;
+  toggleButton?: ShorthandValue<SplitButtonToggleProps>;
 }
 
 export interface SplitButtonState {
@@ -195,13 +196,12 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>, 
             overrideProps: this.handleMenuButtonOverrides,
           },
         )}
-        {Button.create(toggleButton, {
+        {SplitButtonToggle.create(toggleButton, {
           defaultProps: () => ({
             styles: styles.toggleButton,
             className: SplitButton.slotClassNames.toggleButton,
             disabled,
-            icon: 'icon-menu-arrow-down',
-            iconOnly: true,
+            icon: 'icon-menu-arrow-down', // TODO: fix me
             primary,
             secondary,
             ...accessibility.attributes.toggleButton,
@@ -222,6 +222,8 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>, 
     );
   }
 }
+
+SplitButton.Toggle = SplitButtonToggle;
 
 /**
  * A SplitButton enables users to take one of several related actions, one being dominant and rest being displayed in a menu.
--- a/packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
+++ b/packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
@@ -162,7 +162,7 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>, 
     styles,
     unhandledProps,
   }: RenderResultConfig<MenuButtonProps>): React.ReactNode {
-    const { button, disabled, menu, primary, secondary, toggleButton } = this.props;
+    const { button, disabled, menu, primary, secondary, toggleButton, size } = this.props;
     const trigger = Button.create(button, {
       defaultProps: () => ({
         styles: styles.menuButton,
@@ -191,10 +191,10 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>, 
         )}
         {SplitButtonToggle.create(toggleButton, {
           defaultProps: () => ({
-            styles: styles.toggleButton,
             disabled,
             primary,
             secondary,
+            size,
             ...accessibility.attributes.toggleButton,
           }),
           overrideProps: (predefinedProps: ButtonProps) => ({
--- a/packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
+++ b/packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
@@ -95,6 +95,8 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>, 
 
   static displayName = 'SplitButton';
 
+  static Toggle = SplitButtonToggle;
+
   static className = 'ui-splitbutton';
 
   static slotClassNames: SplitButtonSlotClassNames = {
@@ -222,8 +224,6 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>, 
     );
   }
 }
-
-SplitButton.Toggle = SplitButtonToggle;
 
 /**
  * A SplitButton enables users to take one of several related actions, one being dominant and rest being displayed in a menu.
--- a/packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
+++ b/packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
@@ -23,11 +23,6 @@ import { MenuProps } from '../Menu/Menu';
 import { MenuItemProps } from '../Menu/MenuItem';
 import { PopupProps } from '../Popup/Popup';
 
-export interface SplitButtonSlotClassNames {
-  toggleButton: string;
-  size?: SizeValue;
-}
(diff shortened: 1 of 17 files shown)


packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
```diff
@@ -16,17 +16,13 @@ import {
...
   SizeValue,
 } from '../../utils';
+import SplitButtonToggle, { SplitButtonToggleProps } from './SplitButtonToggle';
 import Button, { ButtonProps } from '../Button/Button';
 import MenuButton, { MenuButtonProps } from '../MenuButton/MenuButton';
 import { MenuProps } from '../Menu/Menu';
 import { MenuItemProps } from '../Menu/MenuItem';
 import { PopupProps } from '../Popup/Popup';
 
-export interface SplitButtonSlotClassNames {
```
layershifter: Breaking change?

packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
```diff
@@ -195,13 +189,10 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>,
             overrideProps: this.handleMenuButtonOverrides,
           },
         )}
-        {Button.create(toggleButton, {
+        {SplitButtonToggle.create(toggleButton, {
           defaultProps: () => ({
             styles: styles.toggleButton,
```
layershifter:
```suggestion
```

Do we really want this? I don't think so...

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
```diff
@@ -0,0 +1,174 @@
...
+  FluentComponentStaticProps,
+  ProviderContextPrepared,
+} from '../../types';
+import { getElementType, useAccessibility, useStyles, useTelemetry, useUnhandledProps } from '@fluentui/react-bindings';
+// @ts-ignore
+import { ThemeContext } from 'react-fela';
+
+export interface SplitButtonToggleProps extends UIComponentProps, ContentComponentProps, ChildrenComponentProps {
+  /** Accessibility behavior if overridden by the user. */
+  accessibility?: Accessibility;
```
layershifter:
```suggestion
  accessibility?: Accessibility<never>;
```

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
```diff
@@ -0,0 +1,174 @@
...
+  onFocus?: ComponentEventHandler<SplitButtonToggleProps>;
+
+  /** A button can emphasize that it represents the primary action. */
+  primary?: boolean;
+
+  /** A button can emphasize that it represents an alternative action. */
+  secondary?: boolean;
+}
+
+export type SplitButtonToggleStylesProps = Pick<SplitButtonToggleProps, 'primary' | 'disabled'>;
```
layershifter: Do we want to include `disabled`? Do we have a screener test for it?

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
```diff
@@ -0,0 +1,174 @@
...
+    design,
+  } = props;
+
+  const hasChildren = childrenExist(children);
+
+  const getA11Props = useAccessibility(accessibility, {
+    debugName: SplitButtonToggle.displayName,
+    mapPropsToBehavior: () => ({
+      as,
+      active,
```
layershifter:
```suggestion
      active: undefined,
```

And remove usage of `active` in `props`

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
```diff
@@ -0,0 +1,174 @@
...
+  disabled: PropTypes.bool,
+  onClick: PropTypes.func,
+  onFocus: PropTypes.func,
+  primary: customPropTypes.every([customPropTypes.disallow(['secondary']), PropTypes.bool]),
+  secondary: customPropTypes.every([customPropTypes.disallow(['primary']), PropTypes.bool]),
+};
+
+SplitButtonToggle.handledProps = Object.keys(SplitButtonToggle.propTypes) as any;
+
+SplitButtonToggle.create = createShorthandFactory({ Component: SplitButtonToggle, mappedProp: 'children' });
```
layershifter:
```suggestion
SplitButtonToggle.create = createShorthandFactory({ Component: SplitButtonToggle, mappedProp: 'content' });
```

I think that it's `content`

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const handleClick = (e: React.SyntheticEvent) => {
+    if (disabled) {
+      e.preventDefault();
+      return;
+    }
+
+    _.invoke(props, 'onClick', e, props);
+  };
+
+  const handleFocus = (e: React.SyntheticEvent) => {
```
layershifter: If we don't need to handle there anything, I would rather remove it at all

packages/fluentui/react-northstar/src/themes/teams/components/SplitButton/splitButtonToggleStyles.ts
```diff
@@ -0,0 +1,122 @@
...
+  };
+};
+
+const splitButtonToggleStyles: ComponentSlotStylesPrepared<SplitButtonToggleStylesProps, SplitButtonVariables> = {
+  root: ({ props: p, variables: v, theme }): ICSSInJSStyle => {
+    const { siteVariables } = theme;
+    const { borderWidth } = siteVariables;
+
+    const borderFocusStyles = getBorderFocusStyles({
+      variables: siteVariables,
```
layershifter: Let's use `vairables` instead

packages/fluentui/react-northstar/src/themes/teams/components/SplitButton/splitButtonToggleStyles.ts
```diff
@@ -0,0 +1,122 @@
...
+        ':active': {
+          ...createAnimationStyles('scaleDownSoft', theme),
+          backgroundColor: v.toggleButtonPrimaryBackgroundColorActive,
+          boxShadow: 'none',
+        },
+
+        ':focus': borderFocusStyles[':focus'],
+        ':focus-visible': {
+          backgroundColor: v.toggleButtonPrimaryBackgroundColorFocus,
+        },
```
layershifter: Should we move these to `root`? I.e.

```jsx
':hover': {
   //
   ...p.primary && {}
}
```

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
```diff
@@ -0,0 +1,174 @@
...
+} from '../../types';
+import { getElementType, useAccessibility, useStyles, useTelemetry, useUnhandledProps } from '@fluentui/react-bindings';
+// @ts-ignore
+import { ThemeContext } from 'react-fela';
+
+export interface SplitButtonToggleProps extends UIComponentProps, ContentComponentProps, ChildrenComponentProps {
+  /** Accessibility behavior if overridden by the user. */
+  accessibility?: Accessibility;
+
+  /** A button can show that it cannot be interacted with. */
```
silviuavram: probably rename all button to split button toggle or something.

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
mnajdova: Actually I am using here `buttonBehavior`, so it probably should be `ButtonBehaviorProps`. Should Is pecify this or create new behavior that will invoke the `buttonBehavior`?

packages/fluentui/react-northstar/src/themes/teams/components/SplitButton/splitButtonToggleStyles.ts
mnajdova: Will move this on root and remove the styles prop sent from the parent.

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
mnajdova: It's not even used anymore in `buttonbehavior`. Will remove it.

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
mnajdova: Yep, we have screener test for it

packages/fluentui/react-northstar/src/components/SplitButton/SplitButton.tsx
mnajdova: Yep, and the classname is changed. Will add breaking change entry in changelog.

packages/fluentui/react-northstar/src/components/SplitButton/SplitButtonToggle.tsx
layershifter: `ButtonBehaviorProps` is fine

PR #12436, opened 2020-03-26
chore: fix lerna-aliases for Northstar packages

layershifter:
#### Pull request checklist

#### Description of changes

1. When we run Jest we don't have `dist` (i.e. compiled JS) for Northstar packages
1. We use [`lerna-aliases`](https://github.com/Andarist/lerna-alias) package to point to exact directory with TS instead (i.e. `@fluentui/react-compose/src/index.ts`)
1. To solve publishing issues I have created `lerna.json` (#12159) in `packages/fluentui` to not include Fabric's packages to our publishing pipeline
1. Now `lerna-aliases` uses that config instead of root's one (which contains all packages)
1. There is no `react-compose` and it's under these packages in `packages/fluentui/lerna.json`
1. 💣 

This PR adds proper `directory` setting to use root `lerna.json`.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12436)

### DIFF
--- a/packages/fluentui/accessibility/jest.config.js
+++ b/packages/fluentui/accessibility/jest.config.js
@@ -1,5 +1,7 @@
 module.exports = {
   ...require('@uifabric/build/jest'),
   name: 'accessibility',
-  moduleNameMapper: require('lerna-alias').jest(),
+  moduleNameMapper: require('lerna-alias').jest({
+    directory: require('@uifabric/build/monorepo/findGitRoot')(),
+  }),
 };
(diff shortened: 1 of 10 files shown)


ecraig12345:
@layershifter Can you either merge this if it's still needed or close it if not? The change makes sense to me.

layershifter:
@ecraig12345 merged, thanks for the follow-up 👍

PR #12451, opened 2020-03-27
RFC: do not use Box component

layershifter:
# RFC: do not use `Box` component

This RFC stars conversation to remove usages of `Box.create()` in our components?

## Why blame `Box`es?

Answer is simple: **Performance**.

image

- `Box` is not cheap by itself (hooks, component itself)
- `Box` with `styles` doesn't cache styles (see opposite options)

## Still performance

To fix issue with `Box` & style caching we can also create additional components. I took `ButtonContent` as the simplest example:

_`ButtonMinimal.perf.tsx` was updated to render 100 `Button`s_

Conclusion: `div` are better, on bigger examples it will be more visible

#### `ButtonContent`

| Example                | min   | avg   | median | max   |
| ---------------------- | ----- | ----- | ------ | ----- |
| ButtonMinimal.perf.tsx | 15.85 | 23.67 | 24.15  | 50.49 |
| ButtonMinimal.perf.tsx | 15.62 | 22.46 | 23.74  | 40.98 |

#### `div`

| Example                | min   | avg   | median | max   |
| ---------------------- | ----- | ----- | ------ | ----- |
| ButtonMinimal.perf.tsx | 12.85 | 18.38 | 20.27  | 36.68 |
| ButtonMinimal.perf.tsx | 12.72 | 18.06 | 16.05  | 35.12 |

#### Styles/behavior structure

It becomes different as `content` was before just a slot on `Button`: now styles are moved to `ButtonContent`.

- Should we create a behavior for `ButtonContent`?
  - Yes = oops
  - No = Why? Now it's a separate unit

#### Bundle size

Still affects it. Especially in the case of `ListItem`. 5 components vs 5 `div`s.

## Proposal

Use plain elements instead of `Box`es:

```tsx
// Before
Box.create(media);

// After
createPrimitiveShorthand("div", media);
```

### What we will keep?

- support for `children` render props
- `as` prop will continue to work

### What we will loose?

- `design`, `styles` and `variables` will not longer work on shorthands

### DIFF
--- a/packages/fluentui/react-northstar/src/components/List/ListItem.tsx
+++ b/packages/fluentui/react-northstar/src/components/List/ListItem.tsx
@@ -8,7 +8,6 @@ import * as React from 'react';
 // @ts-ignore
 import { ThemeContext } from 'react-fela';
 
-import Box, { BoxProps } from '../Box/Box';
 import {
   ShorthandValue,
   WithAsProp,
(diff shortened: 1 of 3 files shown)


layershifter:
Closing.

***

For now go with `Box`es:
- `Box` are important as a primitive as they handling styling
- performance impact between cached `Box`es/plain `div`s is not big
- allows to avoid questions when `Box` should be used or not

PR #12453, opened 2020-03-27
feat(icons): add react-icons-northstar package

mnajdova:
This PR adds `react-icons-northstar` package, that contains the `createSvgIcon` factory together with all icon components, which were converted from the teams theme icons svgs definitions.

Some considerations:
Currently the icons are exported by simply the name that was previously used in the theme, but camel-cased.
We may consider addion the Icon suffix on the name, in order to avoid potential collision with other component names from the react-package. This will also may it obvious when users see the component, that it is an icon component.

# What other libraries do:
- `react-icons` - they have prefix in from of the icon, that indicates from which set the icon comes from. For example: `<FaAndroid />` comes from the fontawesome set, `<GoAlert />` comes from the github octions set etc.

- `grommet-icons` - they do not provide any prefix/suffix

- `materialui` - they provide the `Icon` suffix after the icon name. (I like this the most, and it is the closest to our deifnition, as it is a components library, not just for icons.

### Edit
All icons now have the Icon suffix. The conversino of the name prop of the icon is simple kebab-camelCase + the Icon suffix. There are few exception of this (required changes or improvements).

```
'icon-circle' -> CircleIcon
'chevron-right-medium' -> ChevronEndMediumIcon
'triangle-right' -> TriangleEndIcon
onenote -> OneNoteIcon
'onenote-color' -> OneNoteColorIcon
powerpoint -> PowerPointIcon
'powerpoint-color' -> PowerPointColorIcon
onedrive -> OneDriveIcon
// previously required icons mappings
'icon-checkmark' -> AcceptIcon
'icon-circle' -> CircleIcon
'icon-close' -> CloseIcon
'icon-arrow-up' -> TriangleUpIcon
'icon-arrow-down' -> TriangleDownIcon
'icon-arrow-end' -> TriangleEndIcon
'icon-menu-arrow-down' -> ChevronDownMediumIcon
'icon-menu-arrow-end' -> ChevronEndMediumIcon
'icon-pause' -> PauseIcon
'icon-play' -> PlayIcon
```


packages/fluentui/react-icons-northstar/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@fluentui/react-icons-northstar",
+  "description": "A set svg icons and utilities for creating custom ones.",
+  "version": "0.47.0",
+  "author": "Marija Najdova <mnajdova@gmail.com>",
```
layershifter:
```suggestion
```

As I understand packages in Fluent doesn't have `author` field. @kenotron ?

packages/fluentui/react-icons-northstar/jest.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  ...require('@uifabric/build/jest'),
+  name: 'react-utils-northstar',
```
layershifter:
```suggestion
  name: 'react-icons-northstar',
```

packages/fluentui/react-northstar/src/components/Icon/SvgIcon.tsx
```diff
@@ -1,14 +1,11 @@
...
-  getUnhandledProps,
-  useStyles,
-  useTelemetry,
   SvgIconProps,
   svgIconClassName,
   svgIconDisplayName,
   svgIconHandledProps,
   SvgIconChildrenFn,
-} from '@fluentui/react-bindings';
+} from '@fluentui/react-icons-northstar';
```
layershifter: Why we need `SvgIcon` component at all? 🤔

layershifter:
## Tests

I would like to have some tests here:

```tsx
import * as icons from 'react-icons-northstart'

const iconsDir = 'SOME_DIR'

fs.readdirSync(testFolder).forEach(file => {
 test(`${file}` has default export and valid component`, () => {})
 test(`${file}` matches displayName`, () => {})
 test(`${file}` is exported in package`, () => {
    expect(icons[file]).toBeAFunction()
  })
});
```

layershifter:
What is the purpose of moving `createSvgIcon` from `react-bindings`?

layershifter:
What will be a temporary guidance to add icons?

packages/fluentui/react-icons-northstar/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "noImplicitReturns": true,
+    "noImplicitThis": true,
+    "noImplicitAny": true,
+    "noUnusedParameters": true,
+    "strictNullChecks": true
+  },
+  "include": ["src", "test"],
+  "references": [
+    {
+      "icons": "../icons"
```
layershifter: This is wrong 💣 

It should contains only `paths` to referenced local packages, https://www.typescriptlang.org/docs/handbook/project-references.html#what-is-a-project-reference

If you will keep there `createSvgIcon`:

```json
  "references": [
    {
      "path": "../accessibility"
    },
    {
      "path": "../react-bindings"
    },
    {
      "path": "../styles"
    },
  ]
```

Please also add `@fluentui/accessibility` in this case to dependencies:

image

packages/fluentui/react-northstar/src/components/Icon/SvgIcon.tsx
mnajdova: The only reason why I left the component is, if there are use cases where the svgs are dynamic, so that users can render them inside the utility component, without us requiring from them to create new components.

mnajdova:
@layershifter added tests for the components, was able to catch some inconsistencies 👍 

> What is the purpose of moving createSvgIcon from react-bindings?

Now that we have icons package, I believe it makes more sense to be inside that package.

> What will be a temporary guidance to add icons?

In the migration period, I believe that the new icons will have to be added both to the theme and the icons package, in order for us to be able to make the switch with one version and remove the theme icons.

Let me know if this makes sense...

packages/fluentui/react-icons-northstar/test/components/components-test.tsx
```diff
@@ -0,0 +1,24 @@
+import * as icons from '@fluentui/react-icons-northstar';
+import * as fs from 'fs';
+import * as React from 'react';
+
+const iconsDir = 'src/components';
```
layershifter:
```suggestion
const iconsDir = path.resolve(__dirname, 'src/components');
```

packages/fluentui/react-icons-northstar/test/components/components-test.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+describe('Icon components', () => {
+  fs.readdirSync(iconsDir).forEach(file => {
+    const componentName = file.split('.')[0];
```
layershifter:
```suggestion
    const componentName = path.basename(file, '.tsx');
```

packages/fluentui/react-icons-northstar/test/components/components-test.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const IconComponent = (icons as any)[componentName];
+
+    test(`${file} is exported in package`, () => {
+      expect(IconComponent).toBeDefined();
+    });
+
+    test(`${file} is a valid component`, () => {
+      expect(React.isValidElement(<IconComponent />)).toEqual(true);
```
layershifter:
```suggestion
      expect(ReactIs.isValidElementType(IconComponent)).toEqual(true);
```

https://www.npmjs.com/package/react-is

packages/fluentui/react-icons-northstar/package.json
dzearing: Yes let's please remove authors.

ecraig12345: A bunch of the other imported packages also have author fields (usually crediting Levi), which in general I think we should consider removing--at least to me it seems weird to give credit to one person when all this is actually a team effort.

packages/fluentui/react-icons-northstar/test/components/components-test.tsx
mnajdova: But the `src/components` is not in the current `test` folder. And would like to avoid navigating up the parent to the folder.. Let me know if you have some other idea..

packages/fluentui/react-icons-northstar/package.json
mnajdova: ok, will remove it. I just copied this from other package and changed it..

packages/fluentui/react-icons-northstar/test/components/components-test.tsx
```diff
@@ -10,6 +10,10 @@ describe('Icon components', () => {
     const componentName = path.basename(file, '.tsx');
     const IconComponent = (icons as any)[componentName];
 
+    test(`${componentName} has 'Icon' suffix in the name`, () => {
+      expect(componentName.substr(-4, 4)).toEqual('Icon');
```
layershifter:
```suggestion
      expect(componentName.endsWith('Icon').toBeTruthy();
```

_nit_

packages/fluentui/react-icons-northstar/test/components/components-test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const iconsDir = 'src/components';
+
+describe('Icon components', () => {
+  fs.readdirSync(iconsDir).forEach(file => {
+    const componentName = path.basename(file, '.tsx');
+    const IconComponent = (icons as any)[componentName];
+
+    test(`${componentName} has 'Icon' suffix in the name`, () => {
+      expect(componentName.endsWith('Icon').toBeTruthy();
```
mnajdova:
```suggestion
      expect(componentName.endsWith('Icon')).toBeTruthy();
```

PR #12454, opened 2020-03-27
chore: drop `render` callback

layershifter:
# BREAKING CHANGES

We are removing `render` callback, please check [microsoft/fluent-ui-react#1951](https://github.com/microsoft/fluent-ui-react/pull/1951) for the migration path.

`ShorthandRenderProp` was removed as duplicate, please use `ShorthandRenderFunction` instead.

---

We deprecated our previous API with `render` callback 5 months ago:

```tsx
<Menu
  items={[
    render =>
      render(
        { key: "editorials", content: "Editorials" },
        (Component, props) => (
          <Popup content="Foo" trigger={<Component {...props} />} />
        )
      )
  ]}
/>
```

[microsoft/fluent-ui-react#1951](https://github.com/microsoft/fluent-ui-react/pull/1951) also provided a replacement via `children` render function. Time to remove it ⏰


packages/fluentui/react-northstar/src/types.ts
```diff
@@ -57,23 +57,14 @@ export type PropsOf<T> = T extends React.Component<infer TProps>
 // Shorthand Factories
 // ========================================================
 
-export type ShorthandRenderFunction<P> = (Component: React.ReactType, props: P) => React.ReactElement<any>;
```
layershifter: image

packages/fluentui/docs/src/prototypes/hexagonalAvatar/CustomAvatar.tsx
```diff
@@ -26,13 +27,12 @@ const CustomAvatar = (props: AvatarProps & { hexagonal?: boolean }) => {
                   clipPath="url(#bot-hexagon-clip-path)"
                   width="34px"
                   height="34px"
-                  xlinkHref={src}
```
mnajdova, line 29: Ups why was this removed?

layershifter: Oops, fixed

PR #12467, opened 2020-03-30
chore(jest): add typeahead for watch mode

silviuaavram:
#### Pull request checklist

#### Description of changes

Adds https://github.com/jest-community/jest-watch-typeahead to our jest config.
You now have autocomplete/typeahead functionality when you use filtering by file name or by test name in your `yarn test:watch`.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12467)

### DIFF
--- a/packages/fluentui/react-northstar/jest.config.js
+++ b/packages/fluentui/react-northstar/jest.config.js
@@ -12,5 +12,4 @@ module.exports = {
     'test/(.*)$': `<rootDir>/test/$1`,
   },
   setupFilesAfterEnv: [...commonConfig.setupFilesAfterEnv, './test/setup.ts'],
-  watchPlugins: ['jest-watch-typeahead/filename', 'jest-watch-typeahead/testname'],
 };
--- a/packages/fluentui/react-northstar/jest.config.js
+++ b/packages/fluentui/react-northstar/jest.config.js
@@ -12,4 +12,5 @@ module.exports = {
     'test/(.*)$': `<rootDir>/test/$1`,
   },
   setupFilesAfterEnv: [...commonConfig.setupFilesAfterEnv, './test/setup.ts'],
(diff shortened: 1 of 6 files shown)


packages/fluentui/react-northstar/jest.config.js
```diff
@@ -12,4 +12,5 @@ module.exports = {
     'test/(.*)$': `<rootDir>/test/$1`,
   },
   setupFilesAfterEnv: [...commonConfig.setupFilesAfterEnv, './test/setup.ts'],
+  watchPlugins: ['jest-watch-typeahead/filename', 'jest-watch-typeahead/testname'],
```
layershifter: Please move this to shared config

packages/fluentui/react-northstar/package.json
```diff
@@ -38,6 +38,7 @@
     "@types/simulant": "^0.2.0",
     "@uifabric/build": "^7.0.0",
     "faker": "^4.1.0",
+    "jest-watch-typeahead": "^0.4.2",
```
dzearing: @kenotron  or @ecraig12345 should this just go in the scripts/package.json file?

ecraig12345: I think so

dzearing:
It'd be great if we could add to `jest-resources.js` as well, and the `scripts/package.json` devDependencies. 

I'd love for there to be one standard config really, but that might be outside the scope here.

ecraig12345:
Yes, standardizing on one config is definitely out of scope for this PR.

silviuavram:
@ecraig12345 @dzearing I pushed latest change, moved the devDependency to the scripts/package.json.

PR #12478, opened 2020-03-30
feat(FocusZone): Adding implementation for tryInvokeClickForFocusable function

khmakoto:
#### Pull request checklist

#### Description of changes

The `tryInvokeClickForFocusable` function in `FocusZone` existed in v7 but was just a stub in v0. This PR brings the implementation from v7 to v0 to have parity between the two versions and introduces a new `shouldRaiseClicks` prop in both to allow for opt-in/opt-out scenarios for this behavior.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12478)


packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -676,9 +681,36 @@ export default class FocusZone extends React.Component<FocusZoneProps> implement
 
   /**
    * Walk up the dom try to find a focusable element.
-   * TODO
    */
-  tryInvokeClickForFocusable(): boolean {
+  tryInvokeClickForFocusable(targetElement: HTMLElement): boolean {
```
jurokapsiar, line 685: in v0, we are handling enter or space in the behaviors, and they are in some cases different based on the component type (for example in selectable tree, space select the item, enter executes click on it). If we activate this, we would need to do changes across components to prevent double clicking I think

jurokapsiar: but that would be confusing - since the components are not always inside of focus zone, we would need to handle this differently in different scenarios. How does it work in v7?

dzearing: The FocusZone can technically be wrapped around any focusable element, including `div` elements with the `data-is-focusable` attribute. When this attribute is present, the FocusZone makes the element focusable. We've tried to follow a principle of ensuring elements we make focusable are interactive (except for special cases, like "parking" scenarios.) Otherwise it feels like an accessibility concern without the implementer ensuring they do the right thing (I wouldn't argue that they *should* ensure their components are accessible, but we can't control that.)

I'm believe we do not hit the double click event scenario, because when we raiseClick, we return true, stopping propagation and canceling the normal behavior (which may have been simulate the click.)

I'll also say, it seems weird that the FocusZone would effectively "patch" the element's keypress behavior. It seems like the interaction model should be up to the component itself, and whether its inside or outside of the FocusZone shouldn't really matter.

If we remove this behavior I believe some user cases would be affected. If we keep this behavior, I believe v0 scenarios are ok. I'd recommend we keep it as an opt in defaulted to false. In v7 code, we could default it to true to avoid dispruption, and in v8 default it to false with a clear release note on the change.

jurokapsiar: agree - we have similar case, where we have different default for v0/v7 already: https://github.com/microsoft/fluentui/blob/d8c9366bfd9ea973eaf71ae9bcb7a4fe6e43b945/packages/fluentui/react-bindings/src/accessibility/getAccessibility.ts#L36

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -79,6 +86,7 @@ export default class FocusZone extends React.Component<FocusZoneProps> implement
     direction: FocusZoneDirection.bidirectional,
     as: 'div',
     preventDefaultWhenHandled: true,
+    shouldRaiseClicks: false,
```
jurokapsiar, line 89: can you please add it to the test as well? https://github.com/microsoft/fluentui/blob/b66a3351f9b643bc4845fe012e9a3afc42b1654f/packages/fluentui/react-bindings/test/hooks/useAccessibility-test.tsx#L332

PR #12484, opened 2020-03-31
FocusZone: Using v0 prop names for props where names differed and deprecating old ones

khmakoto:
#### Pull request checklist

#### Description of changes

This PR adds to v7 the prop names that are used in v0 for props that have different names across versions, then deprecates the old names in v7 and updates all references to these props while making changes to maintain backwards compatibility for partners using the old props.

The changes made were:
- Using `shouldReceiveFocus` instead of `onBeforeFocus`
- Using `shouldEnterInnerZone` instead of `isInnerZoneKeyStroke`
- Using `stopFocusPropagation` instead of `doNotAllowFocusEventToPropagate`
- Using `onFocus` instead of `onFocusNotification`

The PR also makes some small non-affecting changes to v0 in regard to some const names used in tests and some comments to bring the two versions closer together.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12484)


change/office-ui-fabric-react-2020-03-30-17-38-28-focusZoneDeprecatingProps.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "patch",
```
dzearing, line 2: minor

packages/react-focus/src/components/FocusZone/FocusZone.tsx
```diff
@@ -217,11 +217,10 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
 
   public render(): React.ReactNode {
     // tslint:disable-next-line:deprecation
-    const { rootProps, ariaDescribedBy, ariaLabelledBy, className } = this.props;
+    const { as: tag, elementType, rootProps, ariaDescribedBy, ariaLabelledBy, className } = this.props;
     const divProps = getNativeProps(this.props, htmlElementProperties);
 
-    // tslint:disable-next-line:deprecation
-    const Tag = this.props.as || this.props.elementType || 'div';
+    const Tag = tag || elementType || 'div';
```
joschect, line 223: With composition coming down the pipes, should we hold off on this part and have it get handled by slots?

joschect: @dzearing  and @levithomason to chime in

khmakoto: @joschect This is doing the same thing it was doing before, I'm just using the aliases in line 220 instead to keep the `tslint:disable-next-line:deprecation` comments to a minimum.

PR #12502, opened 2020-04-01
fix(Avatar): Updating sizes of avatars and status

notandrew:
before:
image

after
image


notandrew:
> This looks good. Are you planning on adding examples in this PR or future PR?

Not yet, that might be in a later PR.

packages/fluentui/react-northstar/src/themes/teams/components/Status/statusStyles.ts
```diff
@@ -36,13 +36,13 @@ const getTextColor = (state: string, variables: StatusVariables) => {
...
+  smallest: 6,
+  smaller: 10,
+  small: 10,
   medium: 10,
-  large: 12,
-  larger: 14,
-  largest: 16,
+  large: 10,
+  larger: 16,
+  largest: 0,
```
codepretty, line 45: Does `largest: 0` mean we don't support presence on our largest Avatars?

notandrew: correct, we just turn it off for the largest

PR #12507, opened 2020-04-01
Delete foundation-scenarios; other build script updates

ecraig12345:
Delete the foundation-scenarios package because it wasn't doing anything. (Jason confirmed this.)

Update tslint-rules to have explicit dependencies on the rule sets it extends from.

Remove random scripts/jest.config.js which nobody was using.

This also ended up including the beginning of some Fabric=>Fluent wording updates (the parts that affected the modified packages).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12507)


packages/tslint-rules/package.json
```diff
@@ -1,16 +1,17 @@
 {
   "name": "@uifabric/tslint-rules",
   "version": "7.1.2",
-  "description": "Shared TSLint rules for UI Fabric projects",
+  "description": "Shared TSLint rules for Fluent UI React projects",
```
dzearing, line 4: is this specific to React?

packages/tslint-rules/README.md
```diff
@@ -1,22 +1,10 @@
 # TSLint standard rules
 
-This project contains the baseline standard TSLint rules for [Office UI Fabric React](https://dev.microsoft.com/fabric) projects.
+This project contains the baseline standard TSLint rules for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui) ([formerly Office UI Fabric React](https://developer.microsoft.com/en-us/office/blogs/ui-fabric-is-evolving-into-fluent-ui/)) projects.
```
dzearing, line 3: Maybe we should remove the React part.
```suggestion
This project contains the baseline standard TSLint rules for [Fluent UI](https://developer.microsoft.com/en-us/fluentui) ([formerly Office UI Fabric React](https://developer.microsoft.com/en-us/office/blogs/ui-fabric-is-evolving-into-fluent-ui/)) projects.
```

packages/tslint-rules/package.json
ecraig12345: I think at the moment it is, because other existing variants don't use TS. If we add a TS-based variant which doesn't use React later, we can do a pass on updating all instances where appropriate.

PR #12508, opened 2020-04-01
More fabric=>fluent wording updates

ecraig12345:
#### Pull request checklist

#### Description of changes

Even more Fabric=>Fluent updates (outside the website):
- "Office UI Fabric React" and variants to "Fluent UI React" almost everywhere, with a note in each readme "(formerly Office UI Fabric React)" linking to the blog post
- "Office" and "Office 365" to "Microsoft 365"
- `developer.microsoft.com/fabric` (and variants) to `developer.microsoft.com/en-us/fluentui`
- Update example code in readmes to import from `@fluentui/react`

Extra updates in certain packages:
- `office-ui-fabric-react`: add note about the rename, link to blog post, and instructions for migrating
- `example-app-base`: add note that the components are deprecated and we recommend Storybook instead
- `experiments`, `react-cards`: remove details about experimental component graduation since the plan for that may change
- `file-type-icons`: remove section about how to use the fluent file type icons which is presumably irrelevant now

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12508)

### DIFF
--- a/packages/example-app-base/README.md
+++ b/packages/example-app-base/README.md
@@ -1,8 +1,8 @@
 # @uifabric/example-app-base
 
-**Deprecated** components and utilities used to build internal documentation sites and inner loops for various [Fluent UI React](https://developer.microsoft.com/en-us/fluentui) ([formerly Office UI Fabric React](https://developer.microsoft.com/en-us/office/blogs/ui-fabric-is-evolving-into-fluent-ui/)) packages.
+Components and utilities used to build internal documentation sites and inner loops for various [Fluent UI React](https://developer.microsoft.com/en-us/fluentui) ([formerly Office UI Fabric React](https://developer.microsoft.com/en-us/office/blogs/ui-fabric-is-evolving-into-fluent-ui/)) packages.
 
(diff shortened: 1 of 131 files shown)


packages/example-app-base/README.md
```diff
@@ -1,12 +1,12 @@
 # @uifabric/example-app-base
 
-Components and utilities used to build documentation sites for various [Office UI Fabric React](https://dev.microsoft.com/fabric) packages.
+**Deprecated** components and utilities used to build internal documentation sites and inner loops for various [Fluent UI React](https://developer.microsoft.com/en-us/fluentui) ([formerly Office UI Fabric React](https://developer.microsoft.com/en-us/office/blogs/ui-fabric-is-evolving-into-fluent-ui/)) packages.
```
dzearing, line 3: Why is this marked as Deprecated?

ecraig12345: Because we really really don't want people using this going forward. They should use storybook for their demo apps.

PR #12509, opened 2020-04-02
ChoiceGroup: fix bad cache key for getClassNames

xugao:
#### Pull request checklist

#### Description of changes
When `imageSize` is not defined, we create a new default prop object for it on every render.
That leads to we re-calculate styles on every render

#### Focus areas to test
Added a perf test for `ChoiceGroup`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12509)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
@@ -12,10 +12,16 @@ import { composeRenderFunction } from '@uifabric/utilities';
 
 const getClassNames = classNamesFunction<IChoiceGroupOptionStyleProps, IChoiceGroupOptionStyles>();
 
+const LARGE_IMAGE_SIZE = 71;
+
 /**
  * {@docCategory ChoiceGroup}
  */
-export class ChoiceGroupOptionBase extends React.Component<IChoiceGroupOptionProps, any> {
+export class ChoiceGroupOptionBase extends React.Component<IChoiceGroupOptionProps, {}> {
+  public static defaultProps: Partial<IChoiceGroupOptionProps> = {
+    imageSize: { width: 32, height: 32 },
+  };
+
   private _classNames: IProcessedStyleSet<IChoiceGroupOptionStyles>;
 
   constructor(props: IChoiceGroupOptionProps) {
@@ -31,7 +37,7 @@ export class ChoiceGroupOptionBase extends React.Component<IChoiceGroupOptionPro
       theme,
       iconProps,
       imageSrc,
-      imageSize = { width: 32, height: 32 },
+      imageSize,
       disabled,
       // tslint:disable-next-line:deprecation
       checked,
@@ -48,7 +54,7 @@ export class ChoiceGroupOptionBase extends React.Component<IChoiceGroupOptionPro
       hasImage: !!imageSrc,
       checked,
       disabled,
-      imageIsLarge: !!imageSrc && (imageSize.width > 71 || imageSize.height > 71),
+      imageIsLarge: !!imageSrc && (imageSize!.width > LARGE_IMAGE_SIZE || imageSize!.height > LARGE_IMAGE_SIZE),
       imageSize,
       focused,
     });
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
```diff
@@ -12,10 +12,16 @@ import { composeRenderFunction } from '@uifabric/utilities';
...
 
+const LARGE_IMAGE_SIZE = 71;
+
 /**
  * {@docCategory ChoiceGroup}
  */
-export class ChoiceGroupOptionBase extends React.Component<IChoiceGroupOptionProps, any> {
+export class ChoiceGroupOptionBase extends React.Component<IChoiceGroupOptionProps, {}> {
+  public static defaultProps: Partial<IChoiceGroupOptionProps> = {
+    imageSize: { width: 32, height: 32 },
```
ecraig12345, line 22: Can you add a comment explaining why it's important for this to be in defaultProps?

I wonder how many other places we have similar issues...

xugao: yes it's on my to-do list to check every where :) but outside of scope of this change

packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
```diff
@@ -48,8 +54,8 @@ export class ChoiceGroupOptionBase extends React.Component<IChoiceGroupOptionPro
       hasImage: !!imageSrc,
       checked,
       disabled,
-      imageIsLarge: !!imageSrc && (imageSize.width > 71 || imageSize.height > 71),
-      imageSize,
+      imageIsLarge: !!imageSrc && (imageSize!.width > LARGE_IMAGE_SIZE || imageSize!.height > LARGE_IMAGE_SIZE),
+      imageSize, // avoid unnecessary mutation for better style caching
```
ecraig12345, line 58: Might be better to have this comment up by defaultProps

xugao: i like it here better bc if people see perf issue with component/style, they will come to this method, and then this comment

ecraig12345: But it doesn't explain specifically why the default value needs to be in defaultProps instead of provided inline.

xugao: it doesn't have to be using `defaultProps`, as long as we don't mutate it on every render (which is what we do before).

ecraig12345: Sorry to keep being nitpicky about this, but the comment as it is now doesn't explain *how* we're ensuring unnecessary mutation is avoided (or under what circumstances it was happening before).

JasonGore: To take that one step further, is there an explicit description of behavior that `defaultProps` will never mutate? It may mutate less, but unless there's an explicit React specification for behavior, I wonder if it wouldn't be better to make this completely explicit, defined a module const, and assign it to `defaultProps`.

ecraig12345: That's also why I wanted something like that included with defaultProps, so someone doesn't change how the default is provided (without happening to see this comment later) and reintroduce the issue.

JasonGore: (I'm not sure how much using `static` in a class really guarantees only one instance is ever created in JS. Maybe this isn't an issue.)

xugao: @ecraig12345 - ah ok i get your concern now. will update the comment
@JasonGore - when it's defined as static, there is only one instance.

Issue #12237, opened 2020-03-09
Convert FluentUI components to Functional


silviuaavram:
### Convert from Class to Functional

This task means to keep track of which FluentUI components have migrated to Functional and which ones are needed to move as well. Should also be used for synchronisation and planning.

#### Functional Components so far 

- [x] Accordion
- [x] AccordionContent
- [x] AccordionTitle
- [x] Animation
- [x] Alert
- [x] Attachment
- [x] Avatar
- [x] Box
- [x] Button
- [x] ButtonGroup
- [x] Carousel
- [x] CarouselItem
- [x] CarouselNavigation
- [x] CarouselNavigationItem
- [x] Chat
- [x] ChatItem
- [x] ChatMessage
- [x] Checkbox
- [x] Dialog
- [x] DialogFooter
- [x] Divider
- [ ] Dropdown
- [x] DropdownItem
- [x] DropdownSearchInput
- [x] DropdownSelectedItem
- [x] Embed
- [x] Flex
- [x] FlexItem
- [x] Form
- [x] FormField
- [x] Header
- [x] HeaderDescription
- [x] HierarchicalTree
- [x] HierarchicalTreeItem
- [x] HierarchicalTreeTitle
- [x] Icon
- [x] Image
- [x] Input
- [x] ItemLayout
- [x] Label
- [x] Layout
- [x] List
- [x] ListItem
- [x] Loader
- [x] Menu
- [x] MenuDivider
- [x] MenuItem
- [x] MenuButton
- [x] Popup
- [x] PopupContent
- [x] Portal
- [x] PortalInner
- [x] Provider
- [x] RadioGroup
- [x] RadioGroupItem
- [x] Reaction
- [x] ReactionGroup
- [x] Segment
- [x] Slider
- [x] SplitButton
- [x] Status
- [x] Table
- [x] TableCell
- [x] TableRow
- [x] Text
- [x] TextArea
- [x] Toolbar
- [x] ToolbarCustomItem
- [x] ToolbarDivider
- [x] ToolbarItem
- [x] ToolbarMenu
- [x] ToolbarMenuDivider
- [x] ToolbarMenuItem
- [x] ToolbarMenuRadioGroup
- [x] ToolbarRadioGroup
- [x] Tooltip
- [x] TooltipContent
- [x] Tree
- [x] TreeItem
- [x] TreeTitle
- [x] Video

PR #12513, opened 2020-04-02
chore(Popup): refactor PopupContent to be functional component

layershifter:
# BREAKING CHANGES

As in other components converted to hooks `PopupContent` now passes only these props to styles: `placement`, `pointing`.

This PR refactors `PopupContent` to be functional component, related to #12237. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12513)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Popup/PopupContent.tsx
+++ b/packages/fluentui/react-northstar/src/components/Popup/PopupContent.tsx
@@ -1,25 +1,41 @@
-import { FocusTrapZone, FocusTrapZoneProps, AutoFocusZoneProps, AutoFocusZone } from '@fluentui/react-bindings';
-import cx from 'classnames';
-import * as React from 'react';
-import * as PropTypes from 'prop-types';
-import * as _ from 'lodash';
+import { Accessibility } from '@fluentui/accessibility';
+import {
+  AutoFocusZone,
+  AutoFocusZoneProps,
+  FocusTrapZone,
+  FocusTrapZoneProps,
+  getElementType,
+  useAccessibility,
+  useStyles,
+  useTelemetry,
+  useUnhandledProps,
+} from '@fluentui/react-bindings';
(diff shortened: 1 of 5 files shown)


packages/fluentui/react-northstar/test/specs/components/Popup/PopupContent-test.ts
```diff
@@ -2,5 +2,5 @@ import { isConformant } from 'test/specs/commonTests';
 import PopupContent from 'src/components/Popup/PopupContent';
 
 describe('PopupContent', () => {
-  isConformant(PopupContent, { rendersPortal: true });
```
layershifter, line 5: It never rendered `Portal` 😯

packages/fluentui/react-northstar/src/components/Popup/PopupContent.tsx
```diff
@@ -59,82 +75,123 @@ export interface PopupContentProps extends UIComponentProps, ChildrenComponentPr
...
+  } = props;
+
+  const getA11yProps = useAccessibility(accessibility, {
+    debugName: PopupContent.displayName,
+    rtl: context.rtl,
+  });
+  const { classes } = useStyles<PopupContentStylesProps>(PopupContent.displayName, {
+    className: PopupContent.className,
+    mapPropsToStyles: () => ({
+      pointing,
```
pompomon, line 107:
```suggestion
      pointing,
      placement,
```

Shouldn't we pass placement as well, since it's used in packages/fluentui/react-northstar/src/themes/teams/components/Popup/popupContentStyles.ts ?

layershifter: Good catch, thank you

PR #12519, opened 2020-04-02
Updating Searchbox and Divider examples to use function components

czearing:
#### Pull request checklist

#### Description of changes

Updated `Searchbox` and `Divider` examples to functional components.

#### Focus areas to test

The example pages for the search box and the `Divider`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12519)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Divider/examples/VerticalDivider.Basic.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/Divider/examples/VerticalDivider.Basic.Example.tsx
@@ -31,15 +31,14 @@ const getExampleClassNames = memoizeFunction(
   },
 );
 
-export class VerticalDividerBasicExample extends React.Component<any, any> {
-  public render(): JSX.Element {
-    const classNames = getExampleClassNames();
-    return (
-      <div className={classNames.wrapper}>
-        <p className={classNames.text}> Some text before the divider. </p>
-        <VerticalDivider />
-        <p className={classNames.text}>Some text after the divider. </p>
-      </div>
-    );
-  }
-}
+export const VerticalDividerBasicExample = () => {
+  const classNames = getExampleClassNames();
+
+  return (
+    <div className={classNames.wrapper}>
+      <p className={classNames.text}>Some text before the divider. </p>
+      <VerticalDivider />
+      <p className={classNames.text}>Some text after the divider. </p>
+    </div>
+  );
+};
--- a/packages/office-ui-fabric-react/src/components/Divider/examples/VerticalDivider.Custom.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/Divider/examples/VerticalDivider.Custom.Example.tsx
@@ -31,27 +31,26 @@ const getExampleClassNames = memoizeFunction(
   },
 );
 
-export class VerticalDividerCustomExample extends React.Component<any, any> {
-  public render(): JSX.Element {
-    const exampleClassNames = getExampleClassNames();
-    return (
-      <div className={exampleClassNames.wrapper}>
-        <p className={exampleClassNames.text}> Some text before the divider. </p>
-        <VerticalDivider
-          styles={{
-            wrapper: {
-              height: 40,
-              backgroundColor: '#F4F4F4',
-              padding: 0,
-            },
-            divider: {
-              height: 28,
-              backgroundColor: 'pink',
-            },
-          }}
-        />
-        <p className={exampleClassNames.text}>Some text after the divider. </p>
-      </div>
-    );
-  }
-}
+export const VerticalDividerCustomExample = () => {
+  const exampleClassNames = getExampleClassNames();
+
+  return (
+    <div className={exampleClassNames.wrapper}>
+      <p className={exampleClassNames.text}> Some text before the divider. </p>
+      <VerticalDivider
+        styles={{
+          wrapper: {
+            height: 40,
+            backgroundColor: '#F4F4F4',
+            padding: 0,
+          },
+          divider: {
+            height: 28,
+            backgroundColor: 'pink',
+          },
+        }}
+      />
+      <p className={exampleClassNames.text}>Some text after the divider. </p>
+    </div>
+  );
+};
--- a/packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.CustomIcon.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.CustomIcon.Example.tsx
@@ -2,15 +2,13 @@ import * as React from 'react';
 import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
 
 // tslint:disable:jsx-no-lambda
-export class SearchBoxCustomIconExample extends React.Component<any, any> {
-  public render(): JSX.Element {
-    return (
-      <SearchBox
-        placeholder="Filter"
-        onFocus={() => console.log('onFocus called')}
-        onBlur={() => console.log('onBlur called')}
-        iconProps={{ iconName: 'Filter' }}
-      />
-    );
-  }
-}
+export const SearchBoxCustomIconExample = () => {
+  return (
+    <SearchBox
+      placeholder="Filter"
+      onFocus={() => console.log('onFocus called')}
+      onBlur={() => console.log('onBlur called')}
+      iconProps={{ iconName: 'Filter' }}
+    />
+  );
+};
--- a/packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.CustomIcon.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.CustomIcon.Example.tsx
@@ -2,13 +2,11 @@ import * as React from 'react';
 import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
 
 // tslint:disable:jsx-no-lambda
-export const SearchBoxCustomIconExample = () => {
-  return (
-    <SearchBox
-      placeholder="Filter"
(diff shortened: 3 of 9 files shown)


packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.CustomIcon.Example.tsx
```diff
@@ -2,15 +2,13 @@ import * as React from 'react';
...
-      <SearchBox
-        placeholder="Filter"
-        onFocus={() => console.log('onFocus called')}
-        onBlur={() => console.log('onBlur called')}
-        iconProps={{ iconName: 'Filter' }}
-      />
-    );
-  }
-}
+export const SearchBoxCustomIconExample = () => {
```
dzearing, line 5: This can be simplified, since there is no logic in the example:

```tsx
export const SearchBoxCustomIconExample = () => (
  <SearchBox ... />
);
```

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.Disabled.Example.tsx
```diff
@@ -3,25 +3,23 @@ import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
...
-    return (
-      <Stack tokens={{ childrenGap: 20 }}>
-        <SearchBox
-          placeholder="Search"
-          onFocus={() => console.log('onFocus called')}
-          onBlur={() => console.log('onBlur called')}
-          disabled
-        />
+export const SearchBoxDisabledExample = () => {
+  return (
```
dzearing, line 7: this too

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.FullSize.Example.tsx
```diff
@@ -3,26 +3,48 @@ import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
...
+        onSearch={newValue => console.log('value is ' + newValue)}
+        onFocus={() => console.log('onFocus called')}
+        onBlur={() => console.log('onBlur called')}
+        onChange={() => console.log('onChange called')}
+        disableAnimation
+      />
+    </Stack>
+  );
+};
+// export class SearchBoxFullSizeExample extends React.Component<any, any> {
```
dzearing, line 28: delete commented out code.

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.FullSize.Example.tsx
```diff
@@ -3,26 +3,25 @@ import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
...
-          disableAnimation
-        />
-      </Stack>
-    );
-  }
-}
+
+export const SearchBoxFullSizeExample = () => {
+  return (
+    <Stack tokens={{ childrenGap: 20 }}>
```
ecraig12345, line 9: Same here with saving this in a constant.

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.FullSize.Example.tsx
```diff
@@ -3,26 +3,25 @@ import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
...
+      <SearchBox
+        placeholder="Search"
+        onSearch={newValue => console.log('value is ' + newValue)}
```
ecraig12345, line 12: For the same reason, all these callbacks should also go in constants near the top or bottom of the file, outside the component (and you can remove the tslint:disable).
```ts
const onSearch = (newValue: any) => console.log(`searched for "${newValue}"`);
const onFocus = (ev: React.FocusEvent<HTMLInputElement>) => console.log('onFocus called');
const onBlur = (ev: React.FocusEvent<HTMLInputElement>) => console.log('onBlur called');
const onChange = (ev: React.ChangeEvent<HTMLInputElement>) => console.log(`onChange called. Value: "${ev.target.value}"`);
```

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.Small.Example.tsx
```diff
@@ -2,23 +2,19 @@ import * as React from 'react';
...
-        onBlur={() => console.log('onBlur called')}
-      />
-    );
-  }
-}
+export const SearchBoxSmallExample = () => (
+  <SearchBox
+    styles={{ root: { width: 200 } }}
+    placeholder="Search"
+    onEscape={ev => {
```
ecraig12345, line 9: Same as the previous file with these callbacks

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.Small.Example.tsx
```diff
@@ -2,23 +2,19 @@ import * as React from 'react';
...
-        onSearch={newValue => console.log('SearchBox onSearch fired: ' + newValue)}
-        onFocus={() => console.log('onFocus called')}
-        onBlur={() => console.log('onBlur called')}
-      />
-    );
-  }
-}
+export const SearchBoxSmallExample = () => (
+  <SearchBox
+    styles={{ root: { width: 200 } }}
```
ecraig12345, line 7: This should also go in a constant at the top of the file.

packages/office-ui-fabric-react/src/components/SearchBox/examples/SearchBox.Underlined.Example.tsx
```diff
@@ -2,15 +2,12 @@ import * as React from 'react';
...
-        underlined={true}
-      />
-    );
-  }
-}
+
+export const SearchBoxUnderlinedExample = () => (
+  <SearchBox
+    placeholder="Search"
+    onFocus={() => console.log('onFocus called')}
```
ecraig12345, line 9: Remove the callbacks and the tslint:disable

packages/office-ui-fabric-react/src/components/Divider/examples/VerticalDivider.Basic.Example.tsx
```diff
@@ -31,15 +31,14 @@ const getExampleClassNames = memoizeFunction(
...
-      <div className={classNames.wrapper}>
-        <p className={classNames.text}> Some text before the divider. </p>
-        <VerticalDivider />
-        <p className={classNames.text}>Some text after the divider. </p>
-      </div>
-    );
-  }
-}
+export const VerticalDividerBasicExample = () => {
+  const classNames = getExampleClassNames();
```
ecraig12345, line 35: You can get rid of the `getExampleClassNames` function and just save the value in a const.

packages/office-ui-fabric-react/src/components/Divider/examples/VerticalDivider.Custom.Example.tsx
```diff
@@ -31,27 +31,26 @@ const getExampleClassNames = memoizeFunction(
...
-            },
-          }}
-        />
-        <p className={exampleClassNames.text}>Some text after the divider. </p>
-      </div>
-    );
-  }
-}
+export const VerticalDividerCustomExample = () => {
+  const exampleClassNames = getExampleClassNames();
```
ecraig12345, line 35: Same here about getting rid of the function

packages/office-ui-fabric-react/src/components/Divider/examples/VerticalDivider.Custom.Example.tsx
```diff
@@ -31,27 +31,26 @@ const getExampleClassNames = memoizeFunction(
...
+
+  return (
+    <div className={exampleClassNames.wrapper}>
+      <p className={exampleClassNames.text}> Some text before the divider. </p>
+      <VerticalDivider
+        styles={{
```
ecraig12345, line 41: These should go in a const at the top of the file so they're

packages/office-ui-fabric-react/src/components/Divider/examples/VerticalDivider.Basic.Example.tsx
```diff
@@ -31,15 +31,14 @@ const getExampleClassNames = memoizeFunction(
...
-    return (
-      <div className={classNames.wrapper}>
-        <p className={classNames.text}> Some text before the divider. </p>
-        <VerticalDivider />
-        <p className={classNames.text}>Some text after the divider. </p>
-      </div>
-    );
-  }
-}
+export const VerticalDividerBasicExample = () => {
```
ecraig12345, line 34: For all the examples, it's good to explicitly add a type to the component. In this case:
```suggestion
export const VerticalDividerBasicExample: React.FunctionComponent = () => {
```

packages/office-ui-fabric-react/src/components/Divider/examples/VerticalDivider.Custom.Example.tsx
dzearing: Probably fine to leave, as calling getTheme would get theme values at render, rather than at bundle parsing time.

ecraig12345: We really should change this--running an unnecessary calculation every render isn't a good pattern. 

Regarding calling `getTheme` at parse time, lots of other examples do this currently. Maybe not ideal, but if we want to change it, that should probably be approached all up.

If the styles must stay inside a function call, at least wrap the call in `useConst`:
`const exampleClassNames = useConst(() => getExampleClassNames());`

PR #12053, opened 2020-02-25
TilesList: Add role "row" around rows

erichdev:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [x] Include a change request file using `$ yarn change`

#### Description of changes

Add a wrapper around TilesList rows to give them role="row". This is a portion of an accessibility fix to support aria-selected in TilesList.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12053)

PR #12521, opened 2020-04-02
Support ability to add Grid accessibility semantics to TilesList

ThomasMichon:
#### Description of changes

The `TilesList` component renders using a flexbox layout, and declares its role as `list`. This works because the layout is fluid, and is the most flexible for various forms of content. However, for interactive content and selectable items, this is insufficient because `listitem` elements cannot support `aria-selected`.

This change provides a way for host consumers of `TilesList` to override the necessary pieces to use the `grid` role instead, with `row` elements inserted as-needed and `aria-colindex` supplied to the rendered content cells. Now, `TilesList` computes the row/column index of each cell, and supplies the total number of rows and maximum number of columns to an override for `onRenderRoot`. This, in turn overrides a new `onRenderRoot` hook of `List` itself, which is necessary because only `List` has the final set of 'page' specifications as it renders.

#### Focus areas to test

The existing `TilesList` examples have been updated to showcase the overrides to add 'grid' semantics. The 'generic' demo is left unchanged, showing that it preserves the 'list' semantic.

The best way to test is to use Narrator on Edge Chromium to listen for announcements of the 'selected' state of items in the "Document" and "Media" examples.

This replaces #12053 as it also fully-implements the 'grid' semantics.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12521)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/List/List.types.ts
+++ b/packages/office-ui-fabric-react/src/components/List/List.types.ts
@@ -30,6 +30,57 @@ export const ScrollToMode = {
 export type ScrollToMode = typeof ScrollToMode[keyof typeof ScrollToMode];
 
 /**
+ * Props passed to the render override for the list root.
+ * {@docCategory List}
+ */
+export interface IListOnRenderRootProps<T> {
+  /**
+   * The ref to be applied to the list root.
+   * The `List` uses this element to track scroll position and sizing.
+   */
+  rootRef: React.Ref<HTMLDivElement>;
+  /**
+   * Props to apply to the list root element.
+   */
+  divProps: React.HTMLAttributes<HTMLDivElement>;
+  /**
+   * The active pages to be rendered into the list.
+   * These will have been rendered using `onRenderPage`.
+   */
+  pages: IPage<T>[];
+  /**
+   * The content to be rendered as the list surface element.
+   * This will have been rendered using `onRenderSurface`.
+   */
+  surfaceElement: JSX.Element | null;
+}
+
+/**
+ * Props passed to the render override for the list surface.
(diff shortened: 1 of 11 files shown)


packages/office-ui-fabric-react/src/components/List/List.types.ts
```diff
@@ -180,7 +231,20 @@ export interface IListProps<T = any> extends React.HTMLAttributes<List<T> | HTML
...
+   * Use this to apply some final attributes or structure to the content
+   * each time the list is updated with new active pages or items.
+   */
+  onRenderRoot?: IRenderFunction<IListOnRenderRootProps<T>>;
+
+  /**
+   * Render override for the element representing the surface of the `List`.
+   * Use this to alter the structure of the rendered content if necessary on each update.
+   */
+  onRenderSurface?: IRenderFunction<IListOnRenderSurfaceProps<T>>;
```
aneeshack4, line 247: > The expected type comes from property 'onRenderSurface' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<List<any>> & Pick<Readonly<IListProps<any>> & Readonly<...>, "style" | ... 269 more ... | "ignoreScrollingState"> & Partial<...> & Partial<...>'


There are some more types errors like this: https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=62858&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=cb33ee24-f492-54d2-727a-ae0a7bb84b2c&l=1234

PR #12530, opened 2020-04-03
chore: update Popper to v2

layershifter:
#### Pull request checklist

# BREAKING CHANGES

_All changes are related to all components that consume `Popper`._

## `offset` can't be specified as `string` anymore

Previous implementation based on strings was complicated as for RTL support it required to parse strings. V2 changed this behavior and accepts a tuple or a function.

```jsx
<>
  <Popup offset={[20, 20]} />
  <Popup
    offset={({ placement, reference, popper }) => {
      if (placement === "bottom") {
        return [0, popper.height / 2];
      } else {
        return [];
      }
    }}
  />
</>
```

The same options can be used for consumers:

```jsx
<MenuItem menu={{ items: [1, 2, 3], offset: [20, 20] }} />
```

#### Before

```tsx
<>
  <Popup offset="50px" />
  <Popup offset="-100%p" />
</>
```

#### After

```tsx
<>
  <Popup offset={[50, 0]} />
  <Popup offset={({ popper }) => [-popper.width, 0]} />
</>
```

## Browser support

Additional polyfills are required for IE11:
- `Array.prototype.find`
- `Promise`
- `Object.assign`

More details: https://popper.js.org/docs/v2/browser-support/

---

This PR upgrades Popper's version to V2, [detailed announcement](https://dev.to/fezvrasta/smarter-tooltips-and-popovers-with-popper-2-44bh). Key improvements:
- better performance
- a lot of fixed bugs
- improved support for modifiers, for example: https://twitter.com/atomiksdev/status/1222735500030566403?s=19

#### `padding`

As V2 [doesn't allow to use `margin`s anymore ](https://popper.js.org/docs/v2/migration-guide/#4-remove-all-css-margins), we had to use `padding` (thanks to @pompomon) to set an offset when `Popup`/`Tooltip` has a pointing beak. However, this means that our handing of clicks inside `Popup`/`Tooltip` content should be moved to `content` slot, that's why I introduced `pointerEvents` properties there.

image

## Follow ups 💅 

- do not recreate Popper's instance as [`update()`](https://popper.js.org/docs/v2/lifecycle/#set-new-options) is support in V2
- try to avoid `useDeepMemo()` hook


packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
```diff
@@ -377,6 +377,10 @@ const Popup: React.FC<PopupProps> &
   };
 
   const shouldBlurClose = e => {
+    console.log('shouldBlurClose', e);
```
pompomon, line 380: Please could you remove console.log before merging?

packages/fluentui/react-northstar/test/specs/utils/positioner/positioningHelper-test.ts
```diff
@@ -69,43 +69,4 @@ describe('positioningHelper', () => {
       expectedPlacement: 'left-end',
     });
   });
-
-  describe('positioningHelper offset argument transformed correctly in RTL', () => {
```
pompomon, line 73: Please why have you removed all the tests for applyRtlToOffset?

layershifter: Oops, should restored

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
layershifter: Cruft, good catch

packages/fluentui/react-northstar/test/specs/utils/positioner/positioningHelper-test.ts
layershifter: Tests were updated

packages/fluentui/react-northstar/test/__mocks__/popper.js.ts
```diff
@@ -1,14 +0,0 @@
-import * as PopperJs from 'popper.js';
```
miroslavstastny, line 1: Consumers might be using similar mock. Don't you want to describe it in breaking changes?

packages/fluentui/react-northstar/src/utils/positioner/types.ts
```diff
@@ -1,22 +1,89 @@
+import * as PopperJs from '@popperjs/core';
 import * as React from 'react';
-import PopperJS from 'popper.js';
+
+// Temporary typings for modifiers
```
miroslavstastny, line 4: How temporary?

packages/fluentui/react-northstar/src/utils/positioner/getBoundary.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export default function getBoundary(element: HTMLElement, boundary: Boundary): PopperJs.Boundary {
+  if (boundary === 'window') {
+    return element.ownerDocument.documentElement;
+  }
+
+  if (boundary === 'scrollParent') {
+    let boundariesNode = getScrollParent(element);
+
+    if (boundariesNode.nodeName === 'BODY') {
```
miroslavstastny, line 17: Why do you need to do this?
Isn't `getScrollParent()` doing the same?

packages/fluentui/react-northstar/src/components/MenuButton/MenuButton.tsx
```diff
@@ -117,19 +117,24 @@ export default class MenuButton extends AutoControlledComponent<MenuButtonProps,
...
       PropTypes.arrayOf(PropTypes.oneOf(['click', 'focus', 'context'])),
       PropTypes.arrayOf(PropTypes.oneOf(['hover', 'focus', 'context'])),
     ]),
     flipBoundary: PropTypes.oneOfType([
-      PropTypes.object as PropTypes.Requireable<Element>,
-      PropTypes.oneOf<'scrollParent' | 'window' | 'viewport'>(['scrollParent', 'window', 'viewport']),
+      PropTypes.object as PropTypes.Requireable<HTMLElement>,
+      PropTypes.arrayOf(PropTypes.object) as PropTypes.Requireable<HTMLElement[]>,
+      PropTypes.oneOf<'clippingParents' | 'window' | 'scrollParent'>(['clippingParents', 'window', 'scrollParent']),
     ]),
```
miroslavstastny, line 133: nit: create a separate propType? or move it to `createCommon`?

PR #12552, opened 2020-04-06
feat(Card): Add border-radius variables

pompomon:
#### Pull request checklist

#### Description of changes

Add borderRadius variable to card compoenent

image

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12552)

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Card/Variations/CardExampleHorizontal.tsx
+++ b/packages/fluentui/docs/src/examples/components/Card/Variations/CardExampleHorizontal.tsx
@@ -5,7 +5,10 @@ import { DownloadIcon, MoreIcon } from '@fluentui/react-icons-northstar';
 const CardExampleHorizontal = () => (
   <Card compact horizontal>
     <Card.Preview horizontal>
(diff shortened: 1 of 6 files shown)


packages/fluentui/docs/src/examples/components/Card/Variations/CardExampleHorizontal.tsx
```diff
@@ -5,7 +5,10 @@ import { DownloadIcon, MoreIcon } from '@fluentui/react-icons-northstar';
 const CardExampleHorizontal = () => (
   <Card compact horizontal>
     <Card.Preview horizontal>
-      <Image style={{ height: '115px', width: '115px' }} src="public/images/wireframe/square-image.png" />
+      <Image
+        style={{ height: '115px', width: '115px', borderRadius: '4px 0 0 4px' }}
```
layershifter, line 9: Should we use `overflow: hidden` instead on `Card`/`Card.Preview`?

pompomon: Oh, good point!

PR #12555, opened 2020-04-06
fix(Tooltip): don't add aria-labelled by if aria-label is passed to Tooltip or trigger

silviuaavram:
#### Pull request checklist

#### Description of changes

Checks for `aria-labelled` in either the `root` props or `trigger` props. If found in any, will not add `aria-labelledby` on `trigger`.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12555)


packages/fluentui/accessibility/src/behaviors/Tooltip/tooltipAsLabelBehavior.ts
```diff
@@ -10,30 +14,21 @@ import tooltipAsDescriptionBehavior, { TooltipBehaviorProps } from './tooltipAsD
  */
 const tooltipAsLabelBehavior: Accessibility<TooltipBehaviorProps> = props => {
   const behaviorData = tooltipAsDescriptionBehavior(props);
-  const defaultAriaLabeledBy = getDefaultAriaLabelledBy(props);
+  const triggerAriaLabel = props.triggerAriaLabel || props['aria-label'];
```
kolaps33, line 17: I am not sure here if should we take into the "triggerAriaLabel" the "aria-label" prop which comes from `Tooltip` ? 

When I tried to simulate, then "aria-label" which was set on the `Tooltip` is set on the trigger button. (of course different question is why would someone set aria-label on the  `Tooltip`)
image

packages/fluentui/react-northstar/src/components/Tooltip/Tooltip.tsx
```diff
@@ -228,7 +230,12 @@ const Tooltip: React.FC<TooltipProps> &
   const element = (
     <>
       {triggerElement && (
-        <Ref innerRef={triggerRef}>{React.cloneElement(triggerElement, getA11Props('trigger', triggerProps))}</Ref>
+        <Ref innerRef={triggerRef}>
+          {React.cloneElement(triggerElement, {
+            className: Tooltip.slotClassNames.trigger,
```
layershifter, line 235: Let's avoid doing this as it will silently override existing `className` on a `trigger`

silviuaavram: So what do you suggest? Doing a `cx` composition of classes for that? I need to get that triggerElement in tests so I need the slot class.

packages/fluentui/accessibility/src/behaviors/Tooltip/tooltipAsLabelBehavior.ts
jurokapsiar: if there is a aria-label on the tooltip, it should just stay on the tooltip and should not be applied on the trigger I think

packages/fluentui/react-northstar/src/components/Tooltip/Tooltip.tsx
layershifter: `cx` can work, but in this case it should be done also in `Popup`. However, I don't like this option as `trigger` is not a real slot.

If it's only for tests I don't see any blocker to use `Button.className` or use a specific classname on it.

packages/fluentui/accessibility/src/behaviors/Tooltip/tooltipAsLabelBehavior.ts
silviuaavram: I am now considering aria-label from trigger when I am conditionally applying aria-labelledby.

As for that classname, I fixed it by removing the classname and using another selector in the test. should be fine now

packages/fluentui/react-northstar/src/components/Tooltip/Tooltip.tsx
silviuaavram: removed, using button class now.

PR #12558, opened 2020-04-06
Add custom changelog formatting; convert updateReleaseNotes to TS

ecraig12345:
Add PR links and compare links to changelogs using new beachball custom renderer options.

Convert updateReleaseNotes script to TS, simplify some things, and share some of the utilities with the beachball custom rendering.

Example of the new changelog entry format:

```
## [7.23.0](https://github.com/microsoft/fluentui/tree/@uifabric/experiments_v7.23.0)

Wed, 08 Apr 2020 23:35:25 GMT
[Compare changes](https://github.com/microsoft/fluentui/compare/@uifabric/experiments_v7.22.16..@uifabric/experiments_v7.23.0)

### Minor changes

- Support TilesList overrides for altering list semantics for grid accessibility ([PR #12521](https://github.com/microsoft/fluentui/pull/12521) by tmichon@microsoft.com)
```
which looks like this when rendered (note that **the links are not valid yet** because this example is from a local run, where the version hasn't actually been published):

> ## [7.23.0](https://github.com/microsoft/fluentui/tree/@uifabric/experiments_v7.23.0)
> 
> Wed, 08 Apr 2020 23:35:25 GMT 
> [Compare changes](https://github.com/microsoft/fluentui/compare/@uifabric/experiments_v7.22.16..@uifabric/experiments_v7.23.0)
> 
> ### Minor changes
> 
> - Support TilesList overrides for altering list semantics for grid accessibility ([PR #12521](https://github.com/microsoft/fluentui/pull/12521) by tmichon@microsoft.com)

Compared to the old format:
```
## 7.22.16
Wed, 08 Apr 2020 12:28:27 GMT

### Patches

- fix: fix typings for asAsync & internal usages of React.Children.map (olfedias@microsoft.com)
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12558)


scripts/beachball.config.ts
```diff
@@ -0,0 +1,60 @@
...
+export const config: BeachballConfig = {
+  groups: [
+    {
+      name: 'Fluent UI React',
+      include: ['packages/office-ui-fabric-react', 'packages/react'],
+      disallowedChangeTypes: ['major'],
+    },
+  ],
+  changelog: {
+    customRenderers: {
```
xugao, line 22: let's move this to a different file and keep config.ts short and simple

scripts/beachball.config.ts
```diff
@@ -0,0 +1,60 @@
...
+          : `[commit](${repoUrl}/commit/${entry.commit})`;
+        return `- ${entry.comment} (${commitLink} by ${entry.author})`;
+      },
+    },
+  },
+};
+
+async function _getPrNumber(entry: ChangelogEntry): Promise<number | undefined> {
+  // Look for (presumably) the PR number at the end of the first line of the commit
+  const prMatch = entry.comment.split(/\r?\n/)[0].match(/\(#(\d+)\)$/m);
```
xugao, line 52: i don't think this is gonna help much. `comment` is user input when prompt for change file. we almost never include PR number at that stage. instead we should try parse the commit message from `entry.commit`

scripts/updateReleaseNotes/init.ts
```diff
@@ -41,12 +42,10 @@ if (argv.token === 'your token here') {
...
 }
 
-const repoDetails = {
+export const repoDetails = {
   owner: argv.owner,
   repo: argv.repo,
 };
 
 // Authenticate with github and set up logging if debug arg is provided
-const github = new GitHubApi({ ...(argv.debug ? { log: console } : {}), auth: 'token ' + argv.token });
```
xugao, line 50: pass `token` instead of `auth` similar, keep it the same as the other place

ecraig12345: Not sure what you mean?

ecraig12345: This script has always taken an auth token as a command line argument. But I couldn't use that pattern in the beachball config.

xugao: oh i meant if we can change here to 
```
const github = new GitHubApi({ ...(argv.debug ? { log: console } : {}), token:  argv.token })
```

pass `token` instead of `auth` (assuming they are the same?)

ecraig12345: Hmm it looks like `token` isn't actually valid. I'll change the new one to also use `auth`.

scripts/beachball/customRenderers.ts
```diff
@@ -0,0 +1,66 @@
...
+import { getPullRequestForCommit, fluentRepoDetails } from '../github';
+
+const githubPAT = process.env.GITHUB_PAT;
+if (!githubPAT) {
+  console.warn('\nGITHUB_PAT environment variable not found. GitHub requests may be rate-limited.\n');
+}
+
+const github = new GitHubApi({
+  ...fluentRepoDetails,
+  ...(githubPAT && { auth: 'token ' + githubPAT }),
```
xugao, line 13: check and handle null. `...false` or `...undefined` will throw

scripts/beachball/customRenderers.ts
```diff
@@ -0,0 +1,66 @@
...
+    const logResult = spawnSync('git', ['log', '--pretty=format:%s', '-n', '1', entry.commit]);
+    if (logResult.status === 0) {
+      const message = logResult.stdout.toString().trim();
+      const prMatch = message.split(/\r?\n/)[0].match(/\(#(\d+)\)$/m);
+      if (prMatch) {
+        return Number(prMatch[1]);
+      }
+    }
+  } catch (ex) {
+    // ignore
```
xugao, line 56: add a simple log here

scripts/github/pullRequests.ts
```diff
@@ -0,0 +1,68 @@
...
+      console.warn(`Multiple PRs found for ${commit}:`);
+      console.warn(prs.map(pr => `  ${pr.url}`).join('\n'));
+    }
+
+    if (prs[0]) {
+      verbose && console.log(`Found matching PR #${prs[0].number}.\n`);
+      return processPullRequestApiResponse(prs[0], authorEmail);
+    }
+  } catch (ex) {
+    console.warn(`Error finding PR for ${commit}: ${ex}`);
```
xugao, line 44: i think `ex` will be printed as (unreadable) Object here.

PR #12561, opened 2020-04-06
fix(doc): do not fetch Perf data in maximized examples

miroslavstastny:
#### Pull request checklist

#### Description of changes

Do not fetch perf data in maximized example view.

react-northstar docsite fetches perf data from backend. Before this fix all maximized examples fetched the data as well. This caused unnecessary high load of backend especially when running screener.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12561)

### DIFF
--- a/packages/fluentui/docs/src/routes.tsx
+++ b/packages/fluentui/docs/src/routes.tsx
@@ -51,74 +51,77 @@ import EditorToolbarPrototype from './prototypes/EditorToolbar';
 import HexagonalAvatarPrototype from './prototypes/hexagonalAvatar';
 import TablePrototype from './prototypes/table';
 import VirtualizedTablePrototype from './prototypes/VirtualizedTable';
+import { PerfDataProvider } from './components/ComponentDoc/PerfChart';
 
 const Routes = () => (
   <BrowserRouter basename={__BASENAME__}>
     <Switch>
       <Route exact path="/maximize/:exampleName/:rtl?" component={ExternalExampleLayout} />
       <DocsLayout>
-        <Switch>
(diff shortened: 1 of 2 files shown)


packages/fluentui/docs/src/routes.tsx
```diff
@@ -51,74 +51,77 @@ import EditorToolbarPrototype from './prototypes/EditorToolbar';
...
-            <MarkdownPage page={Composition} />
-          </Route>
-          <Route exact path="/colors" component={Colors} />
-          <Route exact path="/color-palette" component={ColorPalette} />
-          <Route exact path="/color-palette-category" component={CategoryColorPalette} />
-          <Route exact path="/color-schemes" component={ColorSchemes} />
-          <Route exact path="/color-schemes-category" component={CategoryColorSchemes} />
-          <Route exact path="/*" component={PageNotFound} />
-        </Switch>
+        <PerfDataProvider>
```
layershifter, line 61: Is there any sence to fetch this data for prototypes and behaviors? Can we fetch it only for components pages?

miroslavstastny: I don't see any way how to do that.
Once the docsite renders (`DocsLayout` is rendered) I want the data to be fetched and held for the whole lifetime of the docsite.

I agree it makes no sense to fetch the data if you go directly to a prototype (so I could theoretically start the fetch only after a chart is rendered) but I don't see this as the main problem. The problem this PR is solving is to avoid screener to kill the backend.

PR #12574, opened 2020-04-07
fix(Dialog): Fix body scroll when open dialog

### LINKED ISSUE #12197: <Dialog /> overlay doesn't disable the scroll at the top level when open.
### Environment Information
- Library: Fluent-ui-react
- **Package version(s)**:  ^0.44.0

Image Attached:
image

Link to repro with an example:
https://codesandbox.io/s/fluent-ui-example-qq1qh

#### Actual behavior:

`< Dialog />`  doesn't disable the scroll at the root when open. Causes an awkward scrolling of the element in backdrop when ⬇️ ⬆️ keys are pressed.

#### Expected behavior:

Scroll should be disabled at the top element when dialog is open.

Requested priority: Normal

assuncaocharles:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12197

#### Description of changes

Prevents scroll on the body element toggling the overflow css prop

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12574)


packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -26,6 +26,7 @@ import Header, { HeaderProps } from '../Header/Header';
 import Portal, { TriggerAccessibility } from '../Portal/Portal';
 import Flex from '../Flex/Flex';
 import DialogFooter, { DialogFooterProps } from './DialogFooter';
+import { documentRef } from '../../../../react-component-event-listener/src/index';
```
mnajdova, line 29:
```suggestion
import { documentRef } from '@fluentui/react-component-event-listener';
```

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -163,6 +166,7 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
   getInitialAutoControlledState(): DialogState {
     return {
       open: false,
+      documentRef: documentRef.current as HTMLDocument,
```
layershifter, line 169: I don't see from where this is captured

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -176,6 +179,7 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
   handleDialogCancel = (e: Event | React.SyntheticEvent) => {
     _.invoke(this.props, 'onCancel', e, { ...this.props, open: false });
     this.setState({ open: false });
+    this.state.documentRef.body.style.overflow = this.state.prevOverflow;
```
mnajdova, line 182: As @layershifter pointed out, there is no need for this to be in the state. Plus the `documentRef` may be undefined, so add a check for it

layershifter:
There is also a prototype for nested dialogs `packages/fluentui/docs/src/prototypes/NestedPopupsAndDialogs` that should work with these changes.

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
assuncaocharles: Done

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
assuncaocharles: I am using it from `import { documentRef } from '@fluentui/react-component-event-listener'`

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
assuncaocharles: I put it on the state to avoid using `(documentRef.current as HTMLDocument)` all the time... But, I am reming it then

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
layershifter: It should be taken from `this.context` (`this.content.target` is an intance of `document`) otherwise it will not for multiple windows

assuncaocharles: Thanks... Would it be `props.content.content`? 🤔

assuncaocharles: Done

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
assuncaocharles: Done

assuncaocharles:
> I don't think that this will be enough. When we discussed this issue with @jurokapsiar we tried https://www.npmjs.com/package/body-scroll-lock which obviously does more things
> 
> Another question: should be `Dialog` scrollable without `backdrop`?
> 
> image

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -227,6 +229,23 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
     }
   };
 
+  /**
+   * It's needed to control after state changes because self controlled dialogs
+   */
+  componentDidUpdate() {
```
layershifter, line 235: Similar should be for `didMount` as `Dialog` can be opened by default

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -227,6 +229,23 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
...
+    const openedDialogs = +this.context.target.body.getAttribute(Dialog.dialogsCounterAttribute) || 0;
+    if (this.state.open) {
```
layershifter, line 237: This will occur for every dialog rerender

assuncaocharles: Yes, on the dialog is open/closed based in the changes of state. So where would you suggest to place this logic if not in the `componentDidUpdate`?

packages/fluentui/react-northstar/package.json
```diff
@@ -15,6 +15,7 @@
     "@fluentui/react-icons-northstar": "^0.47.0",
     "@fluentui/react-proptypes": "^0.47.0",
     "@fluentui/styles": "^0.47.0",
+    "body-scroll-lock": "^3.0.1",
```
pompomon, line 18: @jurokapsiar @layershifter , please do we need any approvals for third-party library usage?

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
pompomon, line 235: Please could you add tests for this logic?

assuncaocharles: Done

assuncaocharles: e2e tests added

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
layershifter: `Dialog` can be rerendered without changes in the state, for example its parent can produce a render 

In current logic:
```
render ({ state: open }) // disableBodyScroll
render ({ state: open }) // disableBodyScroll
render ({ state: open }) // disableBodyScroll
// still will executed
```

```suggestion
    if (!prevState.open && this.state.open) { 
```

In this case it will executed what it will really become opened. `prevState` is a second param in [`didUpdate`](https://reactjs.org/docs/react-component.html#componentdidupdate)

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -227,6 +229,38 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
     }
   };
 
+  handleBodyScroll(isFirstRender = false) {
+    if (isFirstRender && !this.props.defaultOpen) {
```
pompomon, line 233:
```suggestion
    if (isFirstRender && !(this.props.defaultOpen || this.props.open)) {
```
@layershifter , do I understand correctly that dialog can be open by default by using either `defaultOpen` or `open={true}` interchangeably?

layershifter: `this.state.open` is SSOT for `AutoControlled` components, we covered this with @assuncaocharles already

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -227,6 +229,49 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
...
+    }
+  }
+
+  unlockBodyScroll() {
+    const openDialogs = (+this.context.target.body.getAttribute(Dialog.dialogsCounterAttribute) || 0) - 1;
+    this.context.target.body.setAttribute(Dialog.dialogsCounterAttribute, `${openDialogs}`);
+
+    // Only enables scroll if all dialogs are closed
+    if (openDialogs === 0) {
+      enableBodyScroll(this.context.target.scrollingElement);
```
layershifter, line 248: https://developer.mozilla.org/en-US/docs/Web/API/Document/scrollingElement

It's not supported in IE 😭

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -109,6 +110,7 @@ export interface DialogState {
 class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogState> {
   static displayName = 'Dialog';
   static className = 'ui-dialog';
+  static dialogsCounterAttribute = 'fluent-dialogs-count';
```
layershifter, line 113: Let's move this to the root scope

```
const countAttr = 'fluent-dialogs-count';

class Dialog {. ..
```

packages/fluentui/e2e/tests/dialogPreventScroll-example.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
+import { Button, Dialog } from '@fluentui/react-northstar';
+
+export const selectors = {
+  outerClose: 'outer-close',
```
layershifter, line 5: Do we need all these selectors for this test?

assuncaocharles: Removed

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
assuncaocharles: Done

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
assuncaocharles: Replaced by `body`

packages/fluentui/e2e/tests/dialogPreventScroll-test.ts
```diff
@@ -0,0 +1,60 @@
...
+  it('should decrement the dialogs counter in the body attribute', async () => {
+    await e2e.clickOn(outerTrigger);
+    await e2e.clickOn(innerTrigger);
+    const bodyStyles = await e2e.getAttributeValue('body', 'style');
+    const bodyCounter = await e2e.getAttributeValue('body', Dialog.dialogsCounterAttribute);
+    expect(bodyStyles).toContain(hiddenOverflow);
+    expect(bodyCounter).toBe('2');
+    await e2e.clickOn(innerClose);
+    const bodyCounterAfterCloseInner = await e2e.getAttributeValue('body', Dialog.dialogsCounterAttribute);
+    expect(bodyCounterAfterCloseInner).toBe('1');
```
layershifter, line 42: As this static do not exist anymore, it can be removed

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -227,6 +229,49 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
...
+    }
+  }
+
+  unlockBodyScroll() {
+    const openDialogs = (+this.context.target.body.getAttribute(dialogsCounterAttribute) || 0) - 1;
+    this.context.target.body.setAttribute(dialogsCounterAttribute, `${openDialogs}`);
+
+    // Only enables scroll if all dialogs are closed
+    if (openDialogs === 0) {
+      enableBodyScroll(this.context.target.body);
```
layershifter, line 248:
```suggestion
    if (openDialogs === 0) {
      enableBodyScroll(this.context.target.body);
      this.context.target.body.removeAttribute(dialogsCounterAttribute);
```

assuncaocharles: Done

packages/fluentui/react-northstar/package.json
assuncaocharles: Removed

packages/fluentui/e2e/tests/dialogPreventScroll-test.ts
```diff
@@ -0,0 +1,58 @@
+import { selectors } from './dialogPreventScroll-example';
+
+const outerClose = `#${selectors.outerClose}`;
+const outerTrigger = `#${selectors.outerTrigger}`;
+const innerClose = `#${selectors.innerClose}`;
+const innerTrigger = `#${selectors.innerTrigger}`;
+
+const hiddenOverflow = 'overflow: hidden';
+const unsetOverflow = 'overflow: unset;';
+const dialogsCounterAttribute = 'fluent-dialogs-count';
```
layershifter, line 10: Let's avoid this as it actually tests internals: we want only to ensure that proper styles will be present.

packages/fluentui/e2e/tests/dialogPreventScroll-test.ts
```diff
@@ -0,0 +1,58 @@
+import { selectors } from './dialogPreventScroll-example';
+
+const outerClose = `#${selectors.outerClose}`;
+const outerTrigger = `#${selectors.outerTrigger}`;
+const innerClose = `#${selectors.innerClose}`;
+const innerTrigger = `#${selectors.innerTrigger}`;
+
+const hiddenOverflow = 'overflow: hidden';
+const unsetOverflow = 'overflow: unset;';
```
layershifter, line 9:
```suggestion
const unsetOverflow = 'overflow: unset';
```

?

packages/fluentui/react-northstar/src/components/Dialog/utils/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+  if (target) {
+    target.style.overflow = 'unset';
+    target.removeEventListener('touchmove', _disableIosBodyScroll);
+  }
+}
+
+export function disableBodyScroll(target: HTMLElement): void {
+  if (target) {
+    target.style.overflow = 'hidden';
+    target.addEventListener('touchmove', _disableIosBodyScroll, { passive: false, capture: false });
```
layershifter, line 15: FYI: IE11 :( However, it will not throw, so we are fine

https://caniuse.com/#feat=passive-event-listener
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Browser_compatibility

yarn.lock
```diff
@@ -5868,6 +5868,7 @@ body-parser@1.19.0:
     raw-body "2.4.0"
     type-is "~1.6.17"
 
+
```
layershifter, line 5871:
```suggestion
```

packages/fluentui/e2e/tests/dialogPreventScroll-test.ts
assuncaocharles: Done

packages/fluentui/e2e/tests/dialogPreventScroll-test.ts
assuncaocharles: Done

packages/fluentui/react-northstar/src/components/Dialog/utils/index.tsx
assuncaocharles: Alright

packages/fluentui/CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ Removed types: `TeamsSvgIconSpec`, `ThemeIconSpec`, `SvgIconSpec`, `SvgIconSpecW
 - Give higher `zIndex` to `Tooltip` to make it appear on top of `Dialog` @silviuavram ([#12572](https://github.com/microsoft/fluentui/pull/12572))
 - Fix classes cache miss on empty string result @mnajdova ([#12584](https://github.com/microsoft/fluentui/pull/12584))
 - An update to the `RaiseHandColoredIcon` @TanelVari ([#12618](https://github.com/microsoft/fluentui/pull/12618))
+- Fix body scroll when open `Dialog` @assuncaocharles ([#12197](https://github.com/microsoft/fluentui/issues/12197))
```
layershifter, line 49:
```suggestion
- Fix body scroll when open `Dialog` @assuncaocharles ([#12574](https://github.com/microsoft/fluentui/pull/12574))
```

assuncaocharles: done

PR #12576, opened 2020-04-07
chore: refactor Attachment to be functional component

layershifter:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `Attachment` component to be functional. As in all other cases it contains a breaking change with restricted props set that will be passed to styles functions.

## Prop sets

| `Attachment`    |
| --------- |
| `actionable` |
| `disabled` |

***

## Notable changes

- Adds a test for telemetry capturing
- `progress` was moved out from CSS styles as in `Slider` component

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12576)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
+++ b/packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
@@ -1,24 +1,28 @@
-import { Accessibility, attachmentBehavior } from '@fluentui/accessibility';
+import { Accessibility, attachmentBehavior, AttachmentBehaviorProps } from '@fluentui/accessibility';
+import { getElementType, useAccessibility, useStyles, useTelemetry, useUnhandledProps } from '@fluentui/react-bindings';
 import * as customPropTypes from '@fluentui/react-proptypes';
+import * as _ from 'lodash';
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
-import * as _ from 'lodash';
-import { WithAsProp, ShorthandValue, ComponentEventHandler, withSafeTypeForAs } from '../../types';
+// @ts-ignore
+import { ThemeContext } from 'react-fela';
+
 import {
-  UIComponent,
-  createShorthandFactory,
-  commonPropTypes,
-  applyAccessibilityKeyHandlers,
-  ShorthandFactory,
-} from '../../utils';
+  WithAsProp,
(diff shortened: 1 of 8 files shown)


packages/fluentui/react-northstar/test/specs/commonTests/isConformant.tsx
```diff
@@ -554,32 +555,25 @@ export default function isConformant(
     });
   });
 
-  const validListenerNames = _.reduce(syntheticEvent.types, (result, { listeners }) => [...result, ...listeners], []);
-
   // ---------------------------------------
-  // Opt-in tests
```
layershifter, line 560: This was never used, IDK what it does and why 🤔

packages/fluentui/react-northstar/src/themes/teams/components/Attachment/attachmentStyles.ts
```diff
@@ -130,7 +131,6 @@ const attachmentStyles: ComponentSlotStylesPrepared<AttachmentProps, AttachmentV
     display: 'block',
     bottom: 0,
     left: 0,
-    width: `${p.progress}%`,
```
layershifter, line 133: I moved this to `style` prop on `progress` as in `Slider` component as obviously we don't want to have one hundled entries in cache to different values of `progress`

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
```diff
@@ -49,103 +53,130 @@ export interface AttachmentProps extends UIComponentProps, ChildrenComponentProp
...
+          {Text.create(header, {
+            defaultProps: () => ({ styles: resolvedStyles.header }),
+          })}
+
+          {Text.create(description, {
+            defaultProps: () => ({ styles: resolvedStyles.description }),
+          })}
+        </div>
+      )}
+      {Button.create(action, {
```
pompomon, line 139: Please why don't you have check that `action` is defined, compared with the previous code?

layershifter: Because it will be done by `.create()` factory 😺

PR #12590, opened 2020-04-07
fix(Checkbox): align checkbox with label

silviuaavram:
#### Pull request checklist

#### Description of changes

Checkbox and label should be vertically aligned on medium size for label. Replaced the margin top for the checkbox with a value suggested by @notandrew .

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12590)


layershifter:
Does it work in IE11?

silviuaavram:
> Does it work in IE11?

I don't know, did not try it. But I am just changing the values for existing CSS props, not actually adding anything new.

PR #12595, opened 2020-04-07
Updating Separator and TeachingBubble examples to use function components

czearing:
#### Pull request checklist

#### Description of changes

Separator and TeachingBubble examples are now written as function components.

#### Focus areas to test

Separator and TeachingBubble examples.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12595)


packages/office-ui-fabric-react/src/components/TeachingBubble/examples/TeachingBubble.Basic.Example.tsx
```diff
@@ -1,74 +1,39 @@
...
-          ref={menuButton => (this._menuButtonElement = menuButton!)}
+import { useBoolean } from '@uifabric/react-hooks';
+
+const examplePrimaryButton: IButtonProps = {
+  children: 'Try it out',
+};
+
+export const TeachingBubbleBasicExample: React.FunctionComponent = () => {
+  const [teachingBubbleVisible, { toggle: toggleTeachingBubbleVisible }] = useBoolean(false);
+  const exampleSecondaryButtonProps: IButtonProps = {
```
ecraig12345, line 12: This is a good place for [useMemo](https://reactjs.org/docs/hooks-reference.html#usememo) (same in the other examples that use this pattern)
```suggestion
  const exampleSecondaryButtonProps: IButtonProps = React.useMemo(() => ({
    children: 'Maybe later',
    onClick: toggleTeachingBubbleVisible,
  }, [toggleTeachingBubbleVisible]);
```

packages/office-ui-fabric-react/src/components/Separator/examples/Separator.Theming.Example.tsx
```diff
@@ -14,16 +14,11 @@ const theme: ITheme = createTheme({
...
-        <Text>Horizontal center aligned with custom theme</Text>
-        <Separator theme={theme}>{content}</Separator>
-      </Stack>
-    );
-  }
-}
+export const SeparatorThemingExample: React.FC = () => (
+  <Stack tokens={stackTokens}>
+    <Text>Horizontal center aligned with custom theme</Text>
+    <Separator theme={theme}>{content}</Separator>
```
ecraig12345, line 22: In this case you can just move the content inline since it's only used once.
```suggestion
    <Separator theme={theme}>Today</Separator>
```

packages/office-ui-fabric-react/src/components/TeachingBubble/examples/TeachingBubble.Basic.Example.tsx
```diff
@@ -1,74 +1,39 @@
...
+    children: 'Maybe later',
+    onClick: toggleTeachingBubbleVisible,
+  };
+
+  return (
+    <div className="ms-TeachingBubbleExample">
```
ecraig12345, line 18: This className isn't doing anything, so it should be removed (same in the other TeachingBubble examples)

packages/office-ui-fabric-react/src/components/TeachingBubble/examples/TeachingBubble.Basic.Example.tsx
```diff
@@ -1,74 +1,39 @@
...
-    };
-
-    return (
-      <div className="ms-TeachingBubbleExample">
-        <span
-          className="ms-TeachingBubbleBasicExample-buttonArea"
-          ref={menuButton => (this._menuButtonElement = menuButton!)}
+import { useBoolean } from '@uifabric/react-hooks';
+
+const examplePrimaryButton: IButtonProps = {
```
ecraig12345, line 6: This should probably be `examplePrimaryButtonProps` (same in other examples). Or just `primaryButtonProps`.

packages/office-ui-fabric-react/src/components/TeachingBubble/examples/TeachingBubble.WideIllustration.Example.tsx
```diff
@@ -1,86 +1,49 @@
...
+      <DefaultButton
+        id="targetButton"
+        onClick={toggleTeachingBubbleVisible}
+        text={teachingBubbleVisible ? 'Hide TeachingBubble' : 'Show TeachingBubble'}
+      />
+
+      {teachingBubbleVisible && (
+        <TeachingBubble
+          illustrationImage={exampleImageProps}
+          calloutProps={{ directionalHint: DirectionalHint.bottomCenter }}
```
ecraig12345, line 32: This should go in a const too

packages/office-ui-fabric-react/src/components/TeachingBubble/examples/TeachingBubble.Basic.Example.tsx
```diff
@@ -1,74 +1,42 @@
...
-
-  public render(): JSX.Element {
-    const { isTeachingBubbleVisible } = this.state;
-    const examplePrimaryButton: IButtonProps = {
-      children: 'Try it out',
-    };
-    const exampleSecondaryButtonProps: IButtonProps = {
+export const TeachingBubbleBasicExample: React.FunctionComponent = () => {
+  const [teachingBubbleVisible, { toggle: toggleTeachingBubbleVisible }] = useBoolean(false);
+  const exampleSecondaryButtonProps: IButtonProps = React.useMemo(
```
micahgodbolt, line 12: useMemo should be used for expensive functions. This is a very cheap calculation and using useMemo would probably be confusing for new developer looking at the example code.

micahgodbolt: nice work getting useMemo working though...i had to look it up to remember how it worked!

packages/office-ui-fabric-react/src/components/TeachingBubble/examples/TeachingBubble.MultiStep.Example.tsx
```diff
@@ -1,75 +1,43 @@
...
 
-    this.state = {
-      isTeachingBubbleVisible: false,
-    };
-  }
-
-  public render(): JSX.Element {
-    const { isTeachingBubbleVisible } = this.state;
-    const exampleSecondaryButtonProps: IButtonProps = {
+  const exampleSecondaryButtonProps: IButtonProps = React.useMemo(
```
micahgodbolt, line 12: Same as before. useMemo should be used for expensive functions as suggested here https://usehooks.com/useMemo/

packages/office-ui-fabric-react/src/components/TeachingBubble/examples/TeachingBubble.Basic.Example.tsx
ecraig12345: Caleb actually added this in response to my earlier comment--we discussed this awhile back (unfortunately don't remember who/where/when...) and decided that even though useMemo is not strictly necessary here, we want to demonstrate it (and similar patterns) in our examples to hopefully encourage their use when it actually matters.

ecraig12345: Also, useMemo is important here to reduce the frequency that the button props object mutates. For that reason it should definitely stay in this case.

dzearing: We'll stick with them for now.

PR #12596, opened 2020-04-07
Perf: avoid constant styles mutation in customizable which leads to constant style re-calculation on each render

xugao:
#### Pull request checklist

#### Description of changes
Currently there is a big flaw in `customizable`, that is:
if `concatStyles` is set to `true` and either global styles or component level styles are present, we always mutate the final `styles` that gets passed down to wrapped component. Every time `styles` prop changes, we will re-calculate all styles for the component.

This change simply adds a layer of caching similar to logic inside `styled`. The cache will invalidate as long as either global or component styles mutates by shallow comparison. 

The gain of this fix is quite significant:
All our button variations use `customizable`, and all of them hits this problem because `styles` prop is being passed. therefore with this fix, as long as `styles` props that passed to button doesn't mutate, we avoid a ton of style recalculation on each render. This also improves perf for components that use buttons (e.g. [`CommandBar`](https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.base.tsx#L168), [`Nav`](https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx#L116))
(Gain can also be observed from flamegrill results once ready)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12596)

### DIFF
--- a/apps/perf-test/src/scenarios/BaseButton.tsx
+++ b/apps/perf-test/src/scenarios/BaseButton.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { BaseButton } from 'office-ui-fabric-react';
 
-const scenario = <BaseButton text="I am a button" />;
+const scenario = <BaseButton text="I am a button" styles={{ root: { background: 'red' } }} />;
 
 export default scenario;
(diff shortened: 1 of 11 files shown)


ThomasMichon:
Hooray!

PR #12608, opened 2020-04-08
feat(icons): add @fluentui/react-icons package

mnajdova:
This PR is adding new `@fluentui/react-icons` package that will contain utilities for creating icons, as well as set of svg icons.

There are currently two proposals on the svg factory function:
1. `createSvgIcon` that can be used in the following manner:
```jsx
// Users may provide additional props
const OneDriveIcon = createSvgIcon<{/* addition custom props if any */}>({
  // full signiture ({ classes, props }) - we may add rtl here too
  svg: ({ classes }) => ( // classes are processed classes for the different slots inside the icon (for example svg)
    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0,0,2048,2048" className={classes.svg}>
      <g fill="#1B559B">
        <path d="M 1860 ...." />
      </g>
    </svg>
  ),
  displayName: 'OneDriveIcon', // the name for the icon that will appear in the React tree
});
```

The root `span` element with all processed props is created automatically here inside the `createSvgIcon` factory.

One we decide on the factory used, I will migrate all svg icons.

### NEEDS SOME HELP/SUGGESTIONS
Things that we need to decide on:
- witch factory we want to use or do we want to keep both? - keep `createSvgIcon`
- do we want to use rtl from the theme, or provide separate context for it? - will be added as next step
- do we want to align on a same API for the `SvgIconProps` on v0 and v7, or should we keep the API simple for the v7? - This will help when we want to merge the icons set - will be aligned as next step


packages/react-icons/src/components/BorderBlindsIcon.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+import createSvgIconWithRoot from '../utils/createSvgIconWithRoot';
+import { mergeStyles } from 'office-ui-fabric-react/src/Styling';
+
+const BorderBlindsIcon = createSvgIconWithRoot<{ color1?: string; color2?: string; color3?: string }>({
+  // TODO: convert classNames.svg to selector to selector
+
+  children: ({ classNames, props, processedRootProps }) => {
+    const { className, ...restRootPRops } = processedRootProps;
```
dzearing, line 9:
```suggestion
    const { className, ...restRootProps } = processedRootProps;
```

packages/react-icons/src/components/BorderBlindsIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+          fill: color2,
+        },
+        '.borderblinds-part3': {
+          fill: color3,
+        },
+      },
+    });
+
+    return (
+      <span {...restRootPRops} className={classesOverrides}>
```
dzearing, line 29:
```suggestion
      <span {...restRootProps} className={classesOverrides}>
```

packages/react-icons/src/utils/createSvgIconWithRoot.ts
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+// @ts-ignore
+import { css, getNativeProps, htmlElementProperties } from '@uifabric/utilities/src';
```
dzearing, line 3: remove `/src` here

packages/react-icons/src/utils/createSvgIconWithRoot.ts
```diff
@@ -0,0 +1,56 @@
...
+import { classNames, MS_ICON } from 'office-ui-fabric-react/src/components/Icon/SvgIcon.styles';
```
dzearing, line 4: I think this would result in a circular dependency.

packages/react-icons/src/utils/createSvgIconWithRoot.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ISvgIconProps } from 'office-ui-fabric-react/src/components/Icon/SvgIcon.types';
+
+export type SvgIconFuncArg<TProps = ISvgIconProps> = {
+  classNames: { [iconSlot: string]: string }; // renamed from classes
```
dzearing, line 8: I'd prefer `classes` as its shorter, but also I had read that `class` was going to make its way back to replace `className` at some point (https://github.com/facebook/react/issues/13525).

packages/react-icons/src/utils/createSvgIconWithRoot.ts
```diff
@@ -0,0 +1,56 @@
...
+  // rtl: boolean; // how do we support this?
```
dzearing, line 9: useRTL hook possibly with an RTL specific context

packages/react-icons/src/components/YammerIcon.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import createSvgIcon from '../utils/createSvgIcon';
+
+const YammerIcon = createSvgIcon({
+  svg: ({ classNames }) => (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0,0,2048,2048">
+      <g className="thing">
```
dzearing, line 7: className="thing", also classNames doesn't seem to be used.

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
```diff
@@ -0,0 +1,30 @@
+// tslint:disable:jsx-wrap-multiline max-line-length
+import * as React from 'react';
+import { OneDriveIcon, YammerIcon, BorderBlindsIcon } from '@fluentui/react-icons';
+import { mergeStyles } from 'office-ui-fabric-react/lib/Styling';
+
+const iconClass = mergeStyles({
```
dzearing, line 6: This is fine temporarily, but I'd like to move to raw styling. We could stage this change, so this is a fine short term implementation.

dzearing: Calling mergeStyles here is unnecessary if you're also calling it in the render function. I would suggest removing it from the render function if possible.

packages/react-icons/src/components/OneDriveIcon.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import createSvgIcon from '../utils/createSvgIcon';
+
+const OneDriveIcon = createSvgIcon({
+  svg: ({ classNames }) => (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0,0,2048,2048">
```
dzearing, line 6: Should `classNames` get injected on the root?

packages/react-icons/README.md
```diff
@@ -0,0 +1,11 @@
+# @fluentui/react-icons
+
+**ReactIcons components for [Office UI Fabric React](https://dev.microsoft.com/fabric)**
```
ecraig12345, line 3: Can you update the content of this file? This line should be as follows but you should also look over the rest of the content.
```suggestion
**Icon components for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
```

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
```diff
@@ -0,0 +1,30 @@
+// tslint:disable:jsx-wrap-multiline max-line-length
```
ecraig12345, line 1: Remove?

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
```diff
@@ -0,0 +1,30 @@
...
+  width: 50,
+  margin: '0 25px',
+});
+
+export const IconSvgFactoryExample: React.FunctionComponent = () => {
+  return (
+    <div>
+      <OneDriveIcon className={iconClass} />
+      <YammerIcon
+        className={mergeStyles(iconClass, {
```
ecraig12345, line 18: You should probably get rid of the inline mergeStyles call--this isn't a good pattern to demonstrate perf-wise. Since this class doesn't depend on anything within the component, you could just move the call to the top of the file.

packages/react-icons/src/components/YammerIcon.tsx
ecraig12345: I know this className is being used in an example, but this doesn't seem like a good pattern to demonstrate in general. Maybe find another way to re-style the element for the example?

packages/react-icons/src/components/BorderBlindsIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+            <path className="borderblinds-part3" d="M768 224l256-192v768l-256 192z" />
+          </g>
+        </svg>
+      </span>
+    );
+  },
+  displayName: 'BorderBlindsIcon',
+});
+
+export default BorderBlindsIcon;
```
ecraig12345, line 43: Have we discussed whether we want to use default or named exports in new code?

packages/react-icons/src/utils/createSvgIcon.ts
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react';
+// @ts-ignore
+import { css, getNativeProps, htmlElementProperties } from '@uifabric/utilities/src';
```
ecraig12345, line 3: Please don't use ts-ignore except as a last resort. Is there some reason this didn't work?
```suggestion
import { css, getNativeProps, htmlElementProperties } from '@uifabric/utilities';
```

packages/react-icons/src/utils/createSvgIcon.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export type SvgIconCreateFnParams<TProps> = {
+  svg: SvgIconChildrenFn<TProps & ISvgIconProps>;
+  displayName: string;
+  // handledProps?: (keyof TProps)[];
+};
+
+// TODO: SvgIconProps are different...
+const createSvgIcon = <TProps = {}>({ svg, displayName }: SvgIconCreateFnParams<TProps>) => {
+  // @ts-ignore
```
ecraig12345, line 23: Can you do something else besides ts-ignore here too? Like a cast to any and/or the appropriate type?

packages/react-icons/src/utils/createSvgIcon.ts
```diff
@@ -0,0 +1,55 @@
...
+          role: 'presentation',
+          ['aria-hidden']: true,
+        };
+
+    return React.createElement(
+      'span',
+      {
+        ...containerProps,
+        ...nativeProps,
+        className: css(MS_ICON, classNames.root, className),
```
ecraig12345, line 42: What's the reasoning for using MS_ICON on the new icons?

packages/react-icons/src/utils/createSvgIconWithRoot.ts
mnajdova: Let's keep classes then, we will have a closer API with v0

packages/react-icons/src/components/YammerIcon.tsx
mnajdova: Ah, I just copied this examples from the `Icon.Svg.Example.tsx`. Will create new example with showcase of how the things inside the svgs can be styled

packages/react-icons/src/components/OneDriveIcon.tsx
mnajdova: `createSvgIcon` is doind that. I have two alternatives of the factory once that is for injecting only the svg and another one that allows user to have custom props and define the root element too. Let's see if we will need the both or choose one

packages/react-icons/src/components/BorderBlindsIcon.tsx
mnajdova: Not really, I just used whatever I was used to :S Let me know what makes more sense and will update it..

packages/react-icons/src/utils/createSvgIcon.ts
mnajdova: Sry, this is still work in progress and in early prototype phase, just wanted to get more feedback on general stuff like how the factory's API, how we style things inside the svgs etc.. Will be fixed in the final stage.

packages/react-icons/src/utils/createSvgIcon.ts
mnajdova: Sry very early prototype, not a prod ready version, just was looking for an early review on API and general stuff..

packages/react-icons/src/utils/createSvgIcon.ts
mnajdova: Removed

packages/react-icons/src/utils/createSvgIcon.ts
mnajdova: I thought it's nice to have some common classname for all icons. Not sure if this is the right name for it, but we can change it..

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
mnajdova: Got rid of classnames selectors and inline overrides.

packages/react-icons/src/components/YammerIcon.tsx
mnajdova: Refactored and improved examples. Please take a look.

mnajdova:
@ecraig12345 & @dzearing thanks for the early review. Please take another look on the example/usages and the PR description and let's decide on the unknowns so we can faster move the svgs inside the package.

packages/react-icons/src/components/BorderBlindsIcon.tsx
dzearing: I don't really have a problem with it.

Original thinking with not exporting default was:

A downstream index.ts file might do something like:

```tsx
export * from './BorderBlindsIcon';
```

Which would not include the default export, but only named ones. That said, we have seen concerns multiple times about `export *`. And there is NO friction in `exporting { default as Foo }` explicitly; explicit exports are better anyway in that we don't accidentally leak something unintentionally.

So my 2c is "it's fine".

mnajdova:
See also PR: https://github.com/microsoft/fluentui/pull/12641 which is adding all svg icons

packages/react-icons/src/utils/createSvgIcon.ts
dzearing: If possible, I'd let the author bring them for the new icons and keep them as vanilla as possible. We made the mistake of using ms-Icon for the existing Icon component and have seen a number of issues where developers do things like:

```css
.ms-Icon { font-size: 22px !important; }
```

We can certainly add them later, but let's try without for now.

mnajdova: Will remove it!

dzearing:
> Which factory we want to use or do we want to keep both?

I would prefer to avoid "withRoot" variant. Is there any reason that the svg can't be considered the root? Is there a benefit of wrapping it in a `span`?

> Do we want to use rtl from the theme, or provide separate context for it?

I don't think you need to resolve that yet but I *think* it belongs in a separate context that's easy to reuse without betting on the entire theme. I keep wanting to gravitate to a `useRtl` hook, so that I can make decisions in code. I think this would belong in the standardized theming context package once we make one :)

> Do we want to align on a same API for the SvgIconProps on v0 and v7, or should we keep the API simple for the v7? - This will help when we want to merge the icons set

Alignment would be awesome, but what specifically? (I don't think this blocks this checkin, but curious specifically which props you want to converge.)

packages/react-icons/src/components/BorderBlindsIcon.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import createSvgIconWithRoot from '../utils/createSvgIconWithRoot';
+import { mergeStyles } from 'office-ui-fabric-react/src/Styling';
+
+const BorderBlindsIcon = createSvgIconWithRoot<{ color1?: string; color2?: string; color3?: string }>({
+  children: ({ classes, props, processedRootProps }) => {
+    const { color1 = 'red', color2 = 'green', color3 = 'blue' } = props;
+
+    // custom styles for the border blinds
+    const borderBlindsPart1 = mergeStyles({
```
dzearing, line 10: I would not use class names to do things which could constantly mutate. Could we use inline styles or css variables?

dzearing: (don't let this block checkin, but I think it's important to get the right pattern to replicate.)

packages/react-icons/src/components/OneDriveIcon.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import createSvgIcon from '../utils/createSvgIcon';
+
+const OneDriveIcon = createSvgIcon({
+  svg: ({ classes }) => (
```
dzearing, line 5: Minor suggestion/thought:

will there ever be more than 1 class name here?  Maybe it should just be className.

packages/react-icons/src/utils/createSvgIcon.ts
```diff
@@ -0,0 +1,36 @@
...
+import { css, getNativeProps, htmlElementProperties } from '@uifabric/utilities';
+import { classes, MS_ICON } from './SvgIcon.styles';
+import { ISvgIconProps } from './SvgIcon.types';
+import { SvgIconCreateFnParams } from './types';
+
+const createSvgIcon = <TProps = {}>({ svg, displayName }: SvgIconCreateFnParams<TProps>) => {
+  const Component: React.FC<React.HTMLAttributes<HTMLSpanElement> & TProps & ISvgIconProps> = props => {
+    const { className, style = {} } = props;
+
+    const nativeProps = getNativeProps<React.HTMLAttributes<HTMLElement>>(props, htmlElementProperties);
```
dzearing, line 11: Icons are a hot spot in perf of an app, as they can be rendered tons of times. I wonder if it makes sense to just avoid any unnecessaries altogether:

```tsx
// get all the things and have ...rest
const { color1, color2, color3, className, ...rest } = props;

return (
  <span 
    role="presentation"
    aria-hidden="true" 
    {...rest} 
    className={cx(classes, className)} 
  />);
```

mnajdova: Yes but that would mean that we may end up with wrong props on the root, which is not the case with any other component in both libraries... Would rather merge it as is now and we can measure perf and decide in the next phase...

packages/react-icons/src/components/OneDriveIcon.tsx
mnajdova: In our experience with react-icons-northstar there are cases like this, so would rather leave it as is at this moment as it is easier for extending this way and won't introduce necessity for changes in all icons in the future..

packages/react-icons/src/components/BorderBlindsIcon.tsx
mnajdova: Added inline styles for now. Anyway in the next PR this icon will be remomved, so wouldn't like to spend much time on it..

mnajdova:
@ecraig12345 can you double check the dependencies version. I used `yarn create-package`, but it seems like meanwhile there were some updates on the dependencies.. I check those manually, but would be nice to have ea second par of eyes on them. Thank you!

PR #12613, opened 2020-04-08
DetailsList: Add missing aria label to example

### LINKED ISSUE #12569: Screen reader(Narrator) is reading incorrect count of columns in detailList
Screen reader(Narrator) is reading incorrect count of columns in detailList

### Environment Information
OS: Win 10
Browser:  Edge

### Describe the issue:

Screen reader(Narrator) is reading incorrect count of columns in detailList, Please see below attached snap, You can see this snap contains 4 columns but narrator is reading as 6 columns(including 2 other columns i.e. radio and image)
So aria-colindex would be 4 rather than 6.
DetailList

You can see same behavior on below link.
https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist/customitemcolumns

#### Actual behavior:

Screen reader(Narrator) is reading incorrect count of columns in detailList, Please see below attached snap, You can see this snap contains 4 columns but narrator is reading as 6 columns(including 2 other columns i.e. radio and image)
DetailList
#### Expected behavior:

So aria-colindex would be 4 rather than 6. so that narrator should read column count as 4 rather than 6, It means aria-colindex should be user defined.

joschect:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #12569

#### Description of changes

DetailsList: Add missing aria label to example

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12613)

### DIFF
--- a/change/office-ui-fabric-react-2020-04-08-13-49-26-add-arialabel-custom-column.json
+++ b/change/office-ui-fabric-react-2020-04-08-13-49-26-add-arialabel-custom-column.json
@@ -1,0 +1,9 @@
+{
+  "type": "patch",
+  "comment": "DetailsList: add missing aria-label to example",
+  "packageName": "office-ui-fabric-react",
+  "email": "joschect@microsoft.com",
+  "commit": "5473caaa3faf91b72813ed96438df8234ee102ca",
+  "dependentChangeType": "patch",
+  "date": "2020-04-08T20:49:26.685Z"
+}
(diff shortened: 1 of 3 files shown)


khmakoto:
It seems that you need to update snapshots.

PR #12614, opened 2020-04-08
Migrate Checkbox to functional component

MLoughry:
#### Pull request checklist

#### Description of changes

This change migrates the Checkbox control to a function component in the `react-next` package.

Ping: @dzearing @ecraig12345 @micahgodbolt 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12614)


MLoughry:
@dzearing @ecraig12345 No that `react-next` has been bootstrapped, this change has been moved to that package and is ready for review.

Moving forward with other changes, there's a couple conventions I want to get clarity on:

1. How should we define/wrap the function components with `React.forwardRef()` and `React.memo()`? I see a few options:
```tsx
// - Requires TSLint disable for "no-function-expression" and "no-shadowed-variable"
// + Keeps component name for React errors
export const CheckboxBase = React.memo(
  // tslint:disable-next-line:no-function-expression no-shadowed-variable
  React.forwardRef(function CheckboxBase(props: ICheckboxProps, ref: React.Ref<HTMLDivElement>) {}));
```

```tsx
// + No tslint disables
// - Lose component name for React errors
export const CheckboxBase = React.memo(
  React.forwardRef((props: ICheckboxProps, ref: React.Ref<HTMLDivElement>) => {}));
```

```tsx
// - More verbose
// - Separates wrappers from component
// + Component is not anonymous in errors
// - Component is named differently
function CheckboxBaseComponent(props: ICheckboxProps, ref: React.Ref<HTMLDivElement>) {}
export const CheckboxBase = React.memo(React.forwardRef(CheckboxBaseComponent));
```

```tsx
// + Keep diagnostic name
// - Diagnostic name is defined separately from the component
export const CheckboxBase = React.memo(
  React.forwardRef((props: ICheckboxProps, ref: React.Ref<HTMLDivElement>) => {}));
CheckboxBase.displayName = "CheckboxBase";
```

2. What's the convention around variables starting with an `_`? I know it was meant for private class methods, and in some modules for non-exported modules. Should we use that convention in function components (aside from private hook functions, which by convention must always begin with `use`), and how so?

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -0,0 +1,167 @@
+import * as React from 'react';
+import { classNamesFunction, mergeAriaAttributeValues, warnMutuallyExclusive, FocusRects } from '../../Utilities';
+import { Icon } from '../../Icon';
+import { ICheckboxProps, ICheckboxStyleProps, ICheckboxStyles } from './Checkbox.types';
+import { KeytipData } from '../../KeytipData';
+import { useId, useControllableValue } from '@uifabric/react-hooks';
+
+const getClassNames = classNamesFunction<ICheckboxStyleProps, ICheckboxStyles>();
+
+export const CheckboxBase = React.memo(
```
xugao, line 10: why are you using `React.memo` here? whether to memoize should be determined upon using `checkbox` (mostly in app code) but not at `checkbox` level (component library)

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -0,0 +1,167 @@
...
+    const onRenderLabel = (): JSX.Element | null => {
+      return label ? (
+        <span aria-hidden="true" className={classNames.text} title={title}>
+          {label}
+        </span>
+      ) : null;
+    };
+
+    //#region Input callbacks
+    const _onFocus = React.useCallback(
```
xugao, line 59: inline function like this is fine. no need to memoize (memoize should be done only if it's solving a real perf issue). checkout this article: https://kentcdodds.com/blog/usememo-and-usecallback

xugao: same everywhere

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -0,0 +1,167 @@
...
+      (ev: React.FocusEvent<HTMLElement>): void => {
+        if (inputProps?.onFocus) {
+          inputProps.onFocus(ev);
+        }
+      },
+      [inputProps?.onFocus],
+    );
+
+    const _onBlur = React.useCallback(
```
xugao, line 68: actually why do we need this? should be able to directly use `inputProps?.onBlur`?

packages/react-next/src/components/Checkbox/examples/Checkbox.Basic.Example.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
```
xugao, line 1: I believe examples should go to `packages/examples` package. @ecraig12345  can you confirm?

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -0,0 +1,167 @@
...
+      },
+      [isChecked, isIndeterminate],
+    );
+    //#endregion
+
+    return (
+      <KeytipData keytipProps={keytipProps} disabled={disabled}>
+        {(keytipAttributes: any): JSX.Element => (
+          <div className={classNames.root} title={title} ref={ref}>
+            <FocusRects />
```
xugao, line 97: use `useFocusRects` instead. `<FocusRects />` is just a wrapper component which is meant for be used only in class components where hook cannot be called directly

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
MLoughry: I can't say I agree. Why shouldn't we prevent rerenders if the props don't change?

packages/react-next/src/components/Checkbox/examples/Checkbox.Basic.Example.tsx
MLoughry: This is copied directly from the OUFR package

MLoughry:
Rather than adding new (cloned) one-off perf tests for each component, wouldn't it make sense to figure out how to run the existing perf tests against react-next as well?

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
xugao: https://twitter.com/dan_abramov/status/1083897065263034368
memoization comes at a cost. we should memoize only if it's solving a real perf issue. and that decision should be made when Checkbox is being used, not as default behavior we enforce

MLoughry: As [this reply to that tweet] points out, there's no memory increase (React stores the old props anyway), and we're comparing a small set of props.

xugao: as Dan's reply: the idea of small-but-accumulating cost when you do it everywhere is similar.
I still think we shouldn't make such decision for users. what if user want to use other ways of comparison instead of default (shallow) comparison

packages/react-next/src/components/Checkbox/Checkbox.styles.ts
```diff
@@ -0,0 +1,302 @@
+import { ICheckboxStyleProps, ICheckboxStyles } from './Checkbox.types';
```
xugao, line 1: i'm not sure we want to fork this file now. I would prefer we start with importing Checkbox.styles from OUFR package

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
MLoughry: I fundamentally disagree. 
* Apps don't/shouldn't know how Checkbox is implemented, so they'd have no reason to implement their own comparison operator.
* As the argument goes, JS is fast. Running a shallow comparison against props that were already being held in memory anyway is minor. This comparison isn't run for the first render, so the question is which is more likely?
  * The accumulated cost of these shallow compares during a rerender across a sufficiently large number of components causes noticeable perf degradation, or
  * The shallow comparison prevents enough (more expensive) rerenders to prevent a perf degredation.
  * It makes no noticeable impact on perf at all.

My money is on one of the latter two.

Lastly, as an app author, I don't want to have to wrap imported components in `React.memo()`, especially if I would have to do it across multiple components. Maintaining that would be a massive headache.

packages/react-next/src/components/Checkbox/Checkbox.styles.ts
MLoughry: Is there some clear guidance on what should be forked, and what imported from OUFR? This whole business of staging a separate package rather than a separate branch is.. novel. 

Is there a plan to clean up all the cross references later on?

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
xugao: Let's take a step back and ask ourselves when we should memo a (any) component:
1. Renders often
2. Re-renders with the same props (by shallow or deep or other custom comparison)
3. Component render is expensive

Both 1 & 2 are controlled by App (user's) code

If it's 3, then I think we (component library) should do one of the following:
1. why is render expensive? make it better
2. consider apply memo by default if we have bad perf (in this case, Nav/DetailsList can be such candidate)

For this particular case - Checkbox - we don't see bad perf on re-renders (https://codesandbox.io/s/festive-architecture-nfe33) with same props, so why do we need to apply default memo on it?

packages/react-next/src/components/Checkbox/Checkbox.styles.ts
xugao: We are figuring out these as we go :). I doubt next package will ever have no dependency on current

ecraig12345: We did a branch for 7 and it didn't go so well, hence trying side-by-side this time. That said they both have their issues...if we start on side-by-side and early on don't feel like it's viable we may still be able to switch back.

MLoughry: I'm not familiar with the issues your team had with v7, and I'll work within whatever framework you guys implement as long as there's clear guidance. But I fear that this method makes it harder to see what changed in a PR from baseline, and easier to lose changes made to v7 during the v8 development.

dzearing: My intuition is to copy the whole component content and when shipping, copy it back and diff, rather than one-offing specific files.

I recognize there may be concerns.

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -0,0 +1,145 @@
...
+          </label>
+        </div>
+      )}
+    </KeytipData>
+  );
+});
+
+function useDebugWarning(props: ICheckboxProps) {
+  React.useEffect(() => {
+    if (process.env.NODE_ENV !== 'production') {
```
dzearing, line 113: You can probably move this line up 1 to remove `React.useEffect` usage

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -0,0 +1,149 @@
...
+  );
+});
+
+CheckboxBase.displayName = 'CheckboxBase';
+
+function useDebugWarning(props: ICheckboxProps) {
+  if (process.env.NODE_ENV !== 'production') {
+    // This is a build-time conditional that will be constant at runtime
+    // tslint:disable-next-line:react-hooks-nesting
+    React.useEffect(() => {
```
xugao, line 117: you need to call hook at top level: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level
@ecraig12345  we really need to get react hook lint rules enabled. do you know what's still blocking?

xugao: oh sorry i just saw the comments

MLoughry: This change was made in response to @dzearing's c[omment](https://github.com/microsoft/fluentui/pull/12614#discussion_r416918566). As stated in the comment, it's constant conditional at run-time, so it doesn't run afoul of the Rules of Hooks.

https://github.com/Gelio/tslint-react-hooks has the `react-hooks-nesting` rule I'm pre-emptively disabling with this comment, which we have enabled in OWA. Unfortunately, there's currently no equivalent of the `react-hooks/exhaustive-deps` ESlint rule implemented for TSLint (that I'm aware of)

xugao:
@MLoughry - anything blocking you from merging?

MLoughry:
@xugao I have no reason not to merge this. However,
1. I'm not sure I can merge myself, since I don't have write permissions.
2. Earlier, Size Auditor was failing, due to it seeing all new `react-next` bundles
3. Currently, it looks like there's some issue in the Fluent build that I believe is unrelated to my changes: https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=70518&view=logs&j=1387381c-70b4-581c-de41-e64c81c7b28d&t=1e6c1588-661c-5653-a617-b498ff3c1493

xugao:
@MLoughry - got it. i will merge it once ready

ecraig12345:
@MLoughry Looks like you have write permissions but not admin. (we may have gotten a bit too much in the habit of relying on admin perms to merge in some cases lately)

PR #12615, opened 2020-04-08
FocusZone: Changing restoreFocusFromRoot to preventFocusRestoration in v0 and bringing functionality to v7

khmakoto:
#### Pull request checklist

#### Description of changes

This PR brings changes the `restoreFocusFromRoot` prop in v0 to `preventFocusRestoration` so that the default value is false and brings it to the v7 version. This prop, if true, prevents the FocusZone from attempting to restore the focus to the inner element when the focus is on the root element after componentDidUpdate.`

This PR also changes the comment documentation for the prop to make it more easily understandable.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12615)


packages/react-focus/src/components/FocusZone/FocusZone.tsx
```diff
@@ -178,7 +178,9 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
     if (
       doc &&
       this._lastIndexPath &&
-      (doc.activeElement === doc.body || doc.activeElement === root || doc.activeElement === null)
+      (doc.activeElement === doc.body ||
+        doc.activeElement === null ||
+        (this.props.restoreFocusFromRoot && doc.activeElement === root))
```
jurokapsiar, line 183: looks good to me, but pls note that this changes the behavior for v7 slightly as the prop would be false by default

dzearing: @jurokapsiar I was discussing with @khmakoto , really glad you called this out.

Today if focus is on a button in the zone and a render causes the button to go away, focus will be moved to the next button. With this change it would just go to the body.

We talked about whether this should be opt-in or opt-out behavior. I think in most cases you'd want focus to be restored. There are some cases where the consume is managing focus and doesn't want the behavior, but i think that would be minority scenarios.

I'd propose we change this to `preventFocusRestoration` to opt out if it causes problems. This would enable us to ship this change now without a fundamental break, but would mean that for scenarios where that causes issues, there's a way to opt out.

What do you think?

jurokapsiar: sorry for late response. I agree. If I remember correctly, this was causing some issues when focus zone was in a focus trap zone, but opting out gives us a way to handle this case.

Issue #12588, opened 2020-04-07
RFC: standardize onChange arguments


dzearing:
## Overview

We should provide guidance on how `onChange` events are exposed across components. 

Today we have a discrepancy between v0 and v7 components:

```tsx
// v7: newValue is the new value of the component's main data, depending on the component type
onChange: (ev: React.FormEvent, newValue:  ___ )
```

```tsx
// v0: data is props, with the new value mixed in on top
onChange: (ev: React.FormEvent, data: TProps)
```

## Use cases for `onChange`

1. The primary data has changed, and I need the new value to support controlled component cases or to forward the data to a state management system.

2. When the data changes I not only need the new value, but I need the original props of the component, so that I can differentiate which instance fired the change event to a general handler.

There are a few types of components which might fire an `onChange` event:

* Single literal value components: `Input` and `Toggle`
* Single-object value components: `RadioGroup`, `Select`
* Multi-object selection components: multiselect `Dropdown`, `DetailsList`, `Tree`

## Prior considerations

Semantic UI's thread for `onChange` which provided some background on thinking for Option B below:
https://github.com/Semantic-Org/Semantic-UI-React/issues/623#issuecomment-261018287

Material UI follows a similar approach to v7 code: the second argument represents the new value being changed. See [Select API](https://material-ui.com/api/select/) as an example of this. However the components aren't consistent - See [Input API](https://material-ui.com/api/input/) or [Switch API](https://material-ui.com/api/switch/) where neither have a second argument.

### Option A (recommended): 

* Standardize on a `data` object where we provide the following minimum props:

|Prop|Description|
|-|-|
|`value` | new value for the data of the component |
| `props` | the original props given to the parent component |

Example:

```tsx
interface ChangeData<TValue, TProps> {
  value: TValue;
  props: TProps;
}

const onChange = (ev: FormEvent, data: ChangeData<string, InputProps>) => { 
  // I can access the new value.
  console.log(`The new value is ${data.value}`);

  // I can still access the props of the parent component.
  console.log(`The input (#${data.props.id}) user's passed in props are ${JSON.stringify(data.props)`);

 // I can even access additional metadata specific to the change if needed.
 const { id } = rest;
};
```

#### Pros

👍  value is always predictable in the `value` prop
👍  Parent component's current props are accessible and don't overlap other things in the data object 
👍  Future proof; data object is extendable without ever accidentally overriding a potentially needed prop value.
👍 Type safety is simple.

#### Cons

[...]

PR #12629, opened 2020-04-09
Add new hooks used in migrating Fabric components to functional components

MLoughry:
#### Pull request checklist

#### Description of changes

The hook changes/additions are plucked from my branches migrating a couple components to functional components (https://github.com/MLoughry/office-ui-fabric-react/tree/functional/fabric, https://github.com/MLoughry/office-ui-fabric-react/tree/functional/checkbox). While those changes may have to wait for a new major version, these changes are non-breaking and will allow me to use these hooks across further migrations while we figure out how/when to merge the migrations.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12629)


MLoughry:
Not sure why the Screener baseline is missing an image

packages/react-hooks/src/useControllableValue.ts
```diff
@@ -0,0 +1,37 @@
...
+  controlledValue: TValue | undefined,
+  defaultUncontrolledValue: TValue | undefined,
+  onChange?: (newValue: TValue, ...args: TOnChangeArgs) => void,
+) {
+  const [value, setValue] = React.useState<TValue | undefined>(
+    controlledValue !== undefined ? controlledValue : defaultUncontrolledValue,
+  );
+
+  if (controlledValue !== undefined && controlledValue !== value) {
+    setValue(controlledValue);
```
dzearing, line 21: is this line needed? If `controlledValue` is provided, that's the state, period. Setting the value here would potentially cause a re-render unnecessarily i'd think.

packages/react-hooks/src/useControllableValue.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  const setValueOrCallOnChange = React.useCallback(
+    (newValue: TValue, ...args: TOnChangeArgs) => {
+      if (onChange) {
+        onChange(newValue, ...args);
```
dzearing, line 27: I'm a bit unsure how this would work. Isn't the first param of `onChange` always an event arg? Why hardcode it to `TValue`?

packages/react-hooks/src/useControllableValue.ts
MLoughry: @dzearing It's copied from the implementation in `Checkbox.Base.tsx`. It also ensures that if the controlled value "goes missing" on a future render, the state remains consistent.

packages/react-hooks/src/useControllableValue.ts
MLoughry: @dzearing Is that true for every `onChange` handler across components? This seems more extensible, and allows for a consistent signature for the case where there is no `onChange` handler.

See #12614 for this hook used in practice.

packages/react-hooks/src/useControllableValue.ts
dzearing: I believe this ends up penalizing every controlled render which wouldn't be shippable. I'd recommend removing it, and maybe even adding a test to validate that a component's updating of a controlled value equates to exactly 1 render.

If the value ever goes missing, I would suggest a debug-only warning if you want to protect against this scenario.

packages/react-hooks/src/useControllableValue.ts
dzearing: See RFC here: https://github.com/microsoft/fluentui/issues/12588

For sure the intent of `onChange` is to always keep it as predictable as possible. In current code we are using `(ev: FormEvent, newValue: TValue)`, though in the proposed change the second param would be a data object with a predictable way to access `value`, `props` passed to the originating component, and any additional metadata required.

MLoughry:
Don't merge just yet; in integrating with my other draft PRs, found an issue in the typings for the case when no `onChange` is passed.

MLoughry:
Ping. @dzearing, is this ok to merge?

packages/react-hooks/src/useControllableValue.ts
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+
+type ChangeCallback<TElement extends HTMLElement, TValue> = (
```
ecraig12345, line 3: Shouldn't this be exported since it's used in an exported type?

packages/react-hooks/src/useControllableValue.ts
```diff
@@ -0,0 +1,51 @@
...
+        onChange(ev!, newValue);
+      }
+      if (controlledValue === undefined) {
+        setValue(newValue);
+      }
+    },
+    [onChange, controlledValue === undefined],
+  );
+
+  return [controlledValue !== undefined ? controlledValue : value, setValueOrCallOnChange] as const;
```
ecraig12345, line 50: If `controlledValue` was previously provided but becomes `undefined` later, and a `defaultUncontrolledValue` is also provided (which is incorrect but will likely happen), we need to ensure `undefined` is returned. One possible fix is something like `const isControlled = useConst(controlledValue !== undefined)` and always return the controlled value in the future in that case.

packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+
+/**
+ * React hook to merge multile React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
```
ecraig12345, line 4: typo
```suggestion
 * React hook to merge multiple React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
```

packages/react-hooks/src/index.ts
```diff
@@ -3,3 +3,5 @@ export * from './useBoolean';
 export * from './useConst';
 export * from './useConstCallback';
 export * from './useId';
+export * from './useMergedRefs';
```
ecraig12345, line 6: Can you add the new hooks in the package readme too?

packages/react-hooks/src/useControllableValue.ts
```diff
@@ -0,0 +1,51 @@
...
+  controlledValue: TValue | undefined,
+  defaultUncontrolledValue: TValue | undefined,
+  onChange: TCallback,
+): Readonly<[TValue | undefined, (newValue: TValue | undefined, ev: React.FormEvent<TElement>) => void]>;
+export function useControllableValue<
+  TValue,
+  TElement extends HTMLElement,
+  TCallback extends ChangeCallback<TElement, TValue> | undefined
+>(controlledValue: TValue | undefined, defaultUncontrolledValue: TValue | undefined, onChange?: TCallback) {
+  const [value, setValue] = React.useState<TValue | undefined>(
```
ecraig12345, line 34: Now that there's separate logic to return the controlled value, I don't think we need to store the initial controlled value in state. Might also be good to rename for clarity:
`const [uncontrolledValue, setUncontrolledValue] = React.useState<TValue | undefined>(defaultUncontrolledValue)`

(sorry for all the incremental comments--still thinking through this)

packages/react-hooks/src/useControllableValue.ts
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+
+export type ChangeCallback<TElement extends HTMLElement, TValue> = (
+  ev: React.FormEvent<TElement>,
```
ecraig12345, line 4: If I remember right there may be some cases where the event is actually undefined (like if a change callback is used for manual updates). So the typings should probably allow for that.

packages/react-hooks/README.md
```diff
@@ -109,3 +109,34 @@ const MyComponent = () => {
...
+
+## useControllableValue
+
+`function useControllableValue<TValue, TElement extends HTMLElement>( controlledValue: TValue | undefined, defaultUncontrolledValue: TValue | undefined, ): Readonly<[TValue | undefined, (newValue: TValue | undefined) => void]>`
+
+`function useControllableValue< TValue, TElement extends HTMLElement, TCallback extends ChangeCallback<TElement, TValue> | undefined \>( controlledValue: TValue | undefined, defaultUncontrolledValue: TValue | undefined, onChange: TCallback, ): Readonly<[TValue | undefined, (newValue: TValue | undefined, ev: React.FormEvent<TElement>) => void]>`
+
+Hook to manage the current value for a component that could be either controlled or uncontrolled, such as a checkbox or input field.
+
+It's two required parameters are the `controlledValue` (the current calue of the control in the controlled state), and the `defaultUncontrolledValue` (for the uncontrolled state). Optionally, you may pass a third `onChange` callback to be notified of any changes triggered by the control.
```
ecraig12345, line 121:
```suggestion
Its two required parameters are the `controlledValue` (the current value of the control in the controlled state), and the `defaultUncontrolledValue` (for the uncontrolled state). Optionally, you may pass a third `onChange` callback to be notified of any changes triggered by the control.
```

packages/react-hooks/README.md
```diff
@@ -109,3 +109,34 @@ const MyComponent = () => {
...
+
+The return value will be a setter function that will set the internal state in the unctrolled state, and invoke the onChange callback if present.
```
ecraig12345, line 123:
```suggestion
The return value will be a setter function that will set the internal state in the uncontrolled state, and invoke the `onChange` callback if present.
```

packages/react-hooks/src/useControllableValue.ts
```diff
@@ -0,0 +1,51 @@
...
+  defaultUncontrolledValue: TValue | undefined,
+  onChange: TCallback,
+): Readonly<[TValue | undefined, (newValue: TValue | undefined, ev?: React.FormEvent<TElement>) => void]>;
+export function useControllableValue<
+  TValue,
+  TElement extends HTMLElement,
+  TCallback extends ChangeCallback<TElement, TValue> | undefined
+>(controlledValue: TValue | undefined, defaultUncontrolledValue: TValue | undefined, onChange?: TCallback) {
+  const [value, setValue] = React.useState<TValue | undefined>(defaultUncontrolledValue);
+  const isControlled = useConst<boolean>(!!controlledValue);
```
ecraig12345, line 36: This won't have the desired result for empty strings or 0. Should do either `controlledValue !== undefined` or `controlledValue != null` (depending on how we want to handle null).

MLoughry: Yeah, the tests caught me on this one in my last run. Already fixed. 👍

packages/react-hooks/src/useControllableValue.ts
```diff
@@ -0,0 +1,51 @@
...
+  TElement extends HTMLElement,
+  TCallback extends ChangeCallback<TElement, TValue> | undefined
+>(controlledValue: TValue | undefined, defaultUncontrolledValue: TValue | undefined, onChange?: TCallback) {
+  const [value, setValue] = React.useState<TValue | undefined>(defaultUncontrolledValue);
+  const isControlled = useConst<boolean>(controlledValue !== undefined);
+
+  const setValueOrCallOnChange = React.useCallback(
+    (newValue: TValue | undefined, ev?: React.FormEvent<TElement>) => {
+      if (onChange) {
+        onChange(ev!, newValue);
```
dzearing, line 41:
```
if (ev.defaultPrevented) {
  return;
}
```

MLoughry:
Per our Teams discussion, preventing the call to `onChange` if the value doesn't change would require adding the current state value to the dependencies on the `setValueOrCallOnChange` callback, which would very likely lead to more rerenders than would be saved in this corner case.

PR #12630, opened 2020-04-09
Panel: passing close button styles down using subComponentStyles

### LINKED ISSUE #5689: Refactor Panel styles to use subcomponent styles when Button is refactored and adds IButtonStyleProps
Refactor Dropdown styles to use subcomponent styles when Panel is converted to JS Styling. (At present it uses selectors and fragile global classnames).

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #5689

#### Description of changes

First and foremost, we are re-creating `styles` prop on every render (main reason why got me to make this PR).

This change fixes that by moving the close button styles to `subComponentStyles`. 

Also fixing typing bc i run into errors when `IStyleSet` contains more than one levels of `subComponentStyles` (e.g. Dropdown styles defined subComponentStyles.panel and panel defined subComponentStyles). 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12630)

### DIFF
--- a/packages/merge-styles/src/IStyleSet.ts
+++ b/packages/merge-styles/src/IStyleSet.ts
@@ -25,7 +25,7 @@ export type __MapToFunctionType<T> = Extract<T, Function> extends never
 export type IStyleSet<TStyleSet extends IStyleSet<TStyleSet>> = {
   [P in keyof Omit<TStyleSet, 'subComponentStyles'>]: IStyle;
 } & {
-  subComponentStyles?: { [P in keyof TStyleSet['subComponentStyles']]: IStyleFunctionOrObject<any, IStyleSet<any>> };
+  subComponentStyles?: { [P in keyof TStyleSet['subComponentStyles']]: IStyleFunctionOrObject<any, any> };
 };
 
 /**
@@ -34,7 +34,7 @@ export type IStyleSet<TStyleSet extends IStyleSet<TStyleSet>> = {
 export type IConcatenatedStyleSet<TStyleSet extends IStyleSet<TStyleSet>> = {
   [P in keyof Omit<TStyleSet, 'subComponentStyles'>]: IStyle;
 } & {
-  subComponentStyles?: { [P in keyof TStyleSet['subComponentStyles']]: IStyleFunction<any, IStyleSet<any>> };
+  subComponentStyles?: { [P in keyof TStyleSet['subComponentStyles']]: IStyleFunction<any, any> };
 };
 
 /**
--- a/packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
@@ -332,7 +332,9 @@ export class PanelBase extends React.Component<IPanelProps, IPanelState> impleme
   private _onRenderNavigationContent = (props: IPanelProps): JSX.Element | null => {
     const { closeButtonAriaLabel, hasCloseButton, onRenderHeader = this._onRenderHeader } = props;
     if (hasCloseButton) {
-      const iconButtonStyles = this._classNames.subComponentStyles.closeButton();
+      const iconButtonStyles = this._classNames.subComponentStyles
+        ? this._classNames.subComponentStyles.closeButton()
+        : undefined;
 
       return (
         <>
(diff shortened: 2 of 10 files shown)


packages/merge-styles/src/IStyleSet.ts
```diff
@@ -25,7 +25,7 @@ export type __MapToFunctionType<T> = Extract<T, Function> extends never
 export type IStyleSet<TStyleSet extends IStyleSet<TStyleSet>> = {
   [P in keyof Omit<TStyleSet, 'subComponentStyles'>]: IStyle;
 } & {
-  subComponentStyles?: { [P in keyof TStyleSet['subComponentStyles']]: IStyleFunctionOrObject<any, IStyleSet<any>> };
+  subComponentStyles?: { [P in keyof TStyleSet['subComponentStyles']]: IStyleFunctionOrObject<any, any> };
```
ecraig12345, line 28: Why was this wrong?

packages/office-ui-fabric-react/src/components/Panel/Panel.base.tsx
```diff
@@ -331,42 +331,25 @@ export class PanelBase extends React.Component<IPanelProps, IPanelState> impleme
...
     const { closeButtonAriaLabel, hasCloseButton, onRenderHeader = this._onRenderHeader } = props;
-    const theme = getTheme();
     if (hasCloseButton) {
-      // TODO -Issue #5689: Comment in once Button is converted to mergeStyles
-      // const iconButtonStyles = this._classNames.subComponentStyles
-      // ? (this._classNames.subComponentStyles.iconButton as IStyleFunctionOrObject<IButtonStyleProps, IButtonStyles>)
-      // : undefined;
+      const iconButtonStyles = this._classNames.subComponentStyles
+        ? this._classNames.subComponentStyles.closeButton()
+        : undefined;
```
ecraig12345, line 337: Not that important but you could use the new `?.` syntax here
```suggestion
      const iconButtonStyles = this._classNames.subComponentStyles?.closeButton();
```

packages/merge-styles/src/IStyleSet.ts
xugao: After adding `subComponentStyles: IPanelSubComponentStyles;` in `IPanelStyle`, i got this error on `IDropdownStyles`:
```
Type 'IDropdownStyles' does not satisfy the constraint 'IStyleSet<IDropdownStyles>'.
  Type 'IDropdownStyles' is not assignable to type '{ subComponentStyles?: { panel: IStyleFunctionOrObject<any, IStyleSet<any>>; label: IStyleFunctionOrObject<any, IStyleSet<any>>; multiSelectItem: IStyleFunctionOrObject<...>; } | undefined; }'.
    The types of 'subComponentStyles.panel' are incompatible between these types.
      Type 'IStyleFunctionOrObject<IPanelStyleProps, IPanelStyles>' is not assignable to type 'IStyleFunctionOrObject<any, IStyleSet<any>>'.
        Type 'IStyleFunction<IPanelStyleProps, IPanelStyles>' is not assignable to type 'IStyleFunctionOrObject<any, IStyleSet<any>>'.
          Type 'IStyleFunction<IPanelStyleProps, IPanelStyles>' is not assignable to type 'IStyleFunction<any, IStyleSet<any>>'.
            Call signature return types 'Partial<IPanelStyles>' and 'Partial<IStyleSet<any>>' are incompatible.
              The types of 'subComponentStyles' are incompatible between these types.
                Type 'IPanelSubComponentStyles | undefined' is not assignable to type '{ [x: string]: IStyleFunctionOrObject<any, IStyleSet<any>>; } | undefined'.
                  Type 'IPanelSubComponentStyles' is not assignable to type '{ [x: string]: IStyleFunctionOrObject<any, IStyleSet<any>>; }'.
                    Index signature is missing in type 'IPanelSubComponentStyles'.
```

Honestly, I don't understand exactly why this error. would be great if you know here. 
but note having `IStyleFunctionOrObject<any, any>` is still type safe bc:
```
export declare type IStyleFunctionOrObject<TStylesProps, TStyleSet extends IStyleSet<TStyleSet>> 
```
here generic type `TStyleSet` extends `IStyleSet`

JasonGore: It's really weird that the use of `IStyleFunctionOrObject` would impose an interface requirement on a type arg. That should be left up to `IStyleFunctionOrObject`'s definition, which it seems to already do in this case. Like Xu said, `IStyleSet<any>` appears redundant since `IStyleFunctionOrObject` is requiring it on the second type argument.

ecraig12345: In that case I'm okay with it too

PR #12636, opened 2020-04-09
FocusZone: Adding preventDefaultWhenHandled to v7 version

khmakoto:
#### Pull request checklist

#### Description of changes

This PR brings the `preventDefaultWhenHandled` prop from the v0 version of `FocusZone` to the v7 version. This prop, if true, prevents the default behavior of focus events.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12636)


packages/react-focus/src/components/FocusZone/FocusZone.types.ts
```diff
@@ -163,6 +163,12 @@ export interface IFocusZoneProps extends React.HTMLAttributes<HTMLElement | Focu
    * Callback to notify creators that focus has been set on the FocusZone
    */
   onFocusNotification?: () => void;
+
+  /**
+   * If true, FocusZone prevents default behavior.
```
dzearing, line 168: It's hard as a customer to know when to use this. Is there a reason for a customer to do this or not to do this?

khmakoto: I just copied the comment from their side, but I'll update it to be more descriptive.

dzearing: Once you've updated the PR to be more clear, let me know :)

PR #12647, opened 2020-04-10
Explicit Website LTR and Fabric LTR

micahgodbolt:
Somehow in the past our website explicitly set LTR on the <HTML>. Then we moved to <Fabric> always returning LTR and that HTML one disappeared (still can't see why).

So when we decided to remove the redundant LTR in Fabric, our website broke. 

So I'm adding an explicit LTR since we don't have direct access to the <HTML> element on the hosted page.  

Semi related. The Fabric control wasn't returning the 'dir' value unless it didn't match theme/page dir value. This could cause a bug where Fabric was the only element setting 'LTR'. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12647)

### DIFF
--- a/apps/fabric-website/src/root.tsx
+++ b/apps/fabric-website/src/root.tsx
@@ -1,4 +1,4 @@
-import { registerIcons, on, KeyCodes } from 'office-ui-fabric-react';
+import { registerIcons, on, KeyCodes, setRTL } from 'office-ui-fabric-react';
 import { initializeFileTypeIcons } from '@uifabric/file-type-icons';
 import { createSite } from './utilities/createSite';
 import * as platformPickerStyles from '@uifabric/example-app-base/lib/components/PlatformPicker/PlatformPicker.module.scss';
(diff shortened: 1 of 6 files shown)


apps/fabric-website/src/root.tsx
```diff
@@ -11,6 +11,8 @@ import { AndroidLogo, AppleLogo, WebLogo } from './utilities/index';
 
 initializeFileTypeIcons('https://static2.sharepointonline.com/files/fabric/assets/item-types-fluent/');
 
+setRTL(false);
```
dzearing, line 14: This really shouldn't be required. Did you debug the old version to find what codepath was setting RTL, and why it's no longer done?

micahgodbolt: i dug around for a while but couldn't find anything. we aren't use setRTL anywhere in the website or example-app-base

PR #12649, opened 2020-04-10
Foundation: Fixing contextual RTL scenarios for Foundation components

### LINKED ISSUE #12644: Wrong gaps between Stack items in RTL mode
### Environment Information

- **Package version(s)**: "@fluentui/react": "7.106.0"

https://codepen.io/kamil7x/pen/rNOaNEe

#### Actual behavior:

When in RTL mode, margins set by `childrenGap` does not mirror, so first item does not have any margin, but is rendered as right-most item and stick to second item.

#### Expected behavior:

Stack should not only reverse display order, but also mirror values for `childrenGap` property

Requested priority: High

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12644

#### Description of changes

#12644 pointed out how `Stack`, when in RTL mode, didn't mirror the margins set by `childrenGap` for the first element, which made it so that the item had no margin. After further investigation it was found that this was an issue happening on context level RTL for all Foundation components because Foundation was never updated when contextual RTL was introduced into the codebase.

This PR fixes this issue by making the code that handles styling in Foundation components aware of contextual RTL.

__Before (ltr on top, rtl on bottom):__
image

__After (ltr on top, rtl on bottom):__
image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12649)


packages/foundation/src/slots.tsx
```diff
@@ -200,7 +209,9 @@ function _constructFinalProps<TProps extends IProcessedSlotProps>(
     assign(finalProps, props);
   }
 
-  finalProps.className = mergeStyles(defaultStyles, classNames);
+  const rtl = theme && theme.rtl !== undefined ? theme.rtl : getRTL();
```
dzearing, line 212: I believe getRTL can take in the `theme` object here

dzearing: can probably condense this down below `{ rtl: getRTL(theme) }`

dzearing:
Can you add an rtl snapshot?

Pooja-G28:
Facing this same issue Wrong gaps between Stack items in RTL mode  in  @fluentui/react": "^8.67.4"
Any advise?

PR #12653, opened 2020-04-11
chore(knob) add option to set min, max and step on number knob

assuncaocharles:
#### Pull request checklist

#### Description of changes

 Add option to set min, max and step on number knob

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12653)

### DIFF
--- a/packages/fluentui/docs-components/src/knobs/defaultComponents.tsx
+++ b/packages/fluentui/docs-components/src/knobs/defaultComponents.tsx
@@ -1,5 +1,11 @@
 import * as React from 'react';
-import { KnobComponentProps, KnobComponents, KnobRangeKnobComponentProps, LogInspectorProps } from './types';
+import {
+  KnobComponentProps,
+  KnobComponents,
+  KnobRangeKnobComponentProps,
+  LogInspectorProps,
+  KnobNumberKnobComponentProps,
+} from './types';
 import parseValue from './utils/parseRangeValue';
 
 const KnobField: React.FunctionComponent<KnobComponentProps> = props => (
@@ -35,11 +41,16 @@ const KnobBoolean: React.FunctionComponent<KnobComponentProps> = props => (
   />
 );
 
-const KnobNumber: React.FunctionComponent<KnobComponentProps> = props => (
+const KnobNumber: React.FunctionComponent<KnobNumberKnobComponentProps> = props => (
   <input
     onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
-      props.setValue(parseInt(e.target.value, 10));
+      let newValue = parseInt(e.target.value, 10);
+      newValue = newValue < props.min ? props.min : newValue > props.max ? props.max : newValue;
+      props.setValue(newValue);
     }}
+    min={props.min}
+    max={props.max}
+    step={props.step}
     type="number"
     value={props.value}
   />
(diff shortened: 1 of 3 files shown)


packages/fluentui/docs-components/src/knobs/defaultComponents.tsx
```diff
@@ -35,11 +41,16 @@ const KnobBoolean: React.FunctionComponent<KnobComponentProps> = props => (
...
 );
 
-const KnobNumber: React.FunctionComponent<KnobComponentProps> = props => (
+const KnobNumber: React.FunctionComponent<KnobNumberKnobComponentProps> = props => (
   <input
     onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
-      props.setValue(parseInt(e.target.value, 10));
+      let newValue = parseInt(e.target.value, 10);
+      newValue = newValue < props.min ? props.min : newValue > props.max ? props.max : newValue;
+      props.setValue(newValue);
```
layershifter, line 49:
```suggestion
      props.setValue(parseInt(e.target.value, 10));
```

This looks redundant as `input` with `type=number` already handles this: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_number

packages/fluentui/docs-components/src/knobs/types.ts
```diff
@@ -40,6 +40,12 @@ export type KnobRangeKnobComponentProps = KnobComponentProps & {
   unit: string;
 };
 
+export type KnobNumberKnobComponentProps = KnobComponentProps & {
+  min: number;
+  max: number;
+  step: number;
```
layershifter, line 46:
```suggestion
  min: string;
  max: string;
  step: string;
```

Just to have the same as in `KnobRangeKnobComponentProps`

packages/fluentui/docs-components/src/knobs/useNumberKnob.ts
```diff
@@ -1,11 +1,26 @@
...
+type UseNumberKnobSpecificOptions = {
+  min?: number;
+  max?: number;
+  step?: number;
+};
+
+type UseNumberKnobOptions<T> = UseKnobOptions<T> & UseNumberKnobSpecificOptions;
+
+const useStringKnob = (options: UseNumberKnobOptions<number>) => {
+  const { initialValue = 0, min = 0, max = parseValue(initialValue), step = 1, ...rest } = options;
```
layershifter, line 14:
```suggestion
  const { initialValue = 0, min = 0, max = undefined, step = 1, ...rest } = options;
```

Should it be just `undefined` to keep existing behavior?

packages/fluentui/docs-components/src/knobs/defaultComponents.tsx
assuncaocharles: It doesn't if the user actually type a bigger number than the max...

packages/fluentui/docs-components/src/knobs/types.ts
assuncaocharles: Done

packages/fluentui/docs-components/src/knobs/useNumberKnob.ts
assuncaocharles: Done

PR #12656, opened 2020-04-11
Adding storybook-addon-performance plugin to fabric package

czearing:
#### Pull request checklist

#### Description of changes

Added storybook-addon-performance plugin into office-ui-fabric-react's storybook.

Here is a picture of the plugin:

image

#### Focus areas to test

Run storybook

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12656)


packages/office-ui-fabric-react/.storybook/main.js
```diff
@@ -5,4 +5,5 @@ module.exports = {
     return custom({ config });
   },
   addons: ['@storybook/addon-a11y/register'],
+  addons: ['storybook-addon-performance/register'],
```
ecraig12345, line 8: These should be combined. Otherwise you're overwriting the first entry with the second one.
```suggestion
  addons: ['@storybook/addon-a11y/register', 'storybook-addon-performance/register'],
```

ecraig12345: (I just fixed it by applying the suggestion)

czearing:
@xugao after @dzearing's React de-duping PR the invalid hook error has been resolved.

PR #12660, opened 2020-04-13
feat(Avatar): add icon prop

mnajdova:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12660)


packages/fluentui/docs/src/examples/components/Avatar/Variations/AvatarExampleIcon.shorthand.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
```
layershifter, line 1: Should be this example in `Slots` section?

packages/fluentui/react-northstar/src/components/Avatar/Avatar.tsx
```diff
@@ -98,6 +103,15 @@ const Avatar: React.FC<WithAsProp<AvatarProps>> & FluentComponentStaticProps<Ava
           }),
       })}
       {!image &&
```
layershifter, line 105: In this case there should be a same condition for `image`, i.e. `!icon && Image.create`.

I suggest: 

```
const imageElement = Image.create()
const iconElement = Icon.create()

const hasGlyph = !!image || !!icon

return (
  <JSX>
    {hasGlyph && (imageElement || iconElement)} 
    {!hasGlyph && Label.create()}
  <JSX>
)
```

packages/fluentui/react-northstar/src/components/Avatar/Avatar.tsx
```diff
@@ -98,6 +103,15 @@ const Avatar: React.FC<WithAsProp<AvatarProps>> & FluentComponentStaticProps<Ava
...
+        Box.create(icon, {
+          defaultProps: () =>
+            getA11Props('icon', {
+              title: name,
+              styles: resolvedStyles.icon,
+            }),
+        })}
+      {!image &&
+        !icon &&
         Label.create(label || {}, {
```
layershifter, line 115: Not related to your changes, but this is not cool as if I will pass there `null` the slot will be still rendered

packages/fluentui/react-northstar/src/themes/teams/components/Avatar/avatarStyles.ts
```diff
@@ -13,6 +13,16 @@ const sizeToPxValue = {
   largest: 96,
 };
 
+const iconSizeToPxValue = {
```
layershifter, line 16:
```suggestion
const iconSizeToPxValue: Record<SizeValue, number> = {
```

packages/fluentui/react-northstar/src/components/Avatar/Avatar.tsx
```diff
@@ -24,6 +25,9 @@ export interface AvatarProps extends UIComponentProps {
    */
   accessibility?: Accessibility<never>;
 
+  /** Avatar can contain icon. */
```
pompomon, line 28: Please should we add that icon is only rendered if `image` is not present here or in the docs example?

packages/fluentui/docs/src/examples/components/Avatar/Variations/AvatarExampleIcon.shorthand.tsx
mnajdova: Probably, but then we should restructure all examples, as we don't even have slots examples here. I will just leave it now after the image example, and we can restructure the examples in a separate PR.

PR #12667, opened 2020-04-13
Card: Removing unwanted outline when Cards are clicked

### LINKED ISSUE #12442: Card component has outline focus state
### Environment Information

Can repro in the docs: https://developer.microsoft.com/en-us/fabric#/controls/web/card

#### Actual behavior:

When the Card component itself is clicked, it gets a focus outline, even if there are not input/interactive elements inside the card.

#### Expected behavior:

The Card component shouldn't have any focus state itself. Rather, focus should be on elements within the card that semantically require focus.

Requested priority: Normal

Products/sites affected: all using Cards

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12442

#### Description of changes

The `Card` component was receiving an outline when clicked even when it wasn't interactable. This PR removes this outline in all scenarios given that even for clickable `Cards` we have different styling to identify that they are clickable.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12667)

### DIFF
--- a/change/@uifabric-react-cards-2020-04-13-15-09-26-cardOutline.json
+++ b/change/@uifabric-react-cards-2020-04-13-15-09-26-cardOutline.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Card: Removing unwanted outline when Cards are clicked.",
+  "packageName": "@uifabric/react-cards",
+  "email": "humbertomakotomorimoto@gmail.com",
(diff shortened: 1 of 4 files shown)


khmakoto:
@msft-github-bot wait 1 minute.

PR #12670, opened 2020-04-14
 Updating Announced, Callout, and Calendar examples to function components

czearing:
#### Pull request checklist

#### Description of changes

Changed AnnouncedSearchResultsExample, Callout,  Calendar examples to be written as function components.

#### Focus areas to test

Announced, Callout, and Calendar examples 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12670)


packages/office-ui-fabric-react/src/components/Callout/examples/Callout.Status.Example.tsx
```diff
@@ -31,56 +26,32 @@ const styles = mergeStyleSets({
...
-    });
-  };
-
-  private _onCalloutDismiss = (): void => {
-    this.setState({
-      isCalloutVisible: false,
-    });
-  };
-}
+export const CalloutStatusExample: React.FunctionComponent = () => {
```
joschect, line 29: I think maybe this should be 
```suggestion
export const StatusCalloutExample: React.FunctionComponent = () => {
```
From what I can tell this might be why the build is failing. @czearing Thanks a bunch for submitting this change!

czearing: Oops that seems like it could be the issue. Thanks for reviewing @joschect!

PR #12674, opened 2020-04-14
chore: Attachment use compose()

layershifter:
#### Pull request checklist

# BREAKING CHANGES

### `slotClassNames` removal

`slotClassNames` were removed from `Attachment`, please use classnames on matching components instead.

```tsx
// Before
Attachment.slotClassNames.action
// After
AttachmentAction.className
```

### `styles` were moved to components from slots

- styles for `action`, `description`, `icon` & `header` slots were moved to separate components
  - as these components are separate they have own props for styles
- styles for `content` slots were moved to separate `AttachmentBody`
- styles for `progress` slot are not separated anymore and should be used on `root` in `attachmentStyles`

#### Before

```ts
const attachmentStyles: ComponentSlotStylesPrepared<AttachmentStylesProps, AttachmentVariables> = {
  content: () => ({ display: 'flex' }),
  progress: () => ({ display: 'flex' }),

  action: () => ({ display: 'flex' }),
  description: () => ({ display: 'flex' }),
  icon: () => ({ display: 'flex' }),
  header: () => ({ display: 'flex' }),
}
```

#### After

```tsx
const attachmentStyles: ComponentSlotStylesPrepared<AttachmentStylesProps, AttachmentVariables> = {
  root: () => ({
    '& .ui-attachment__progress': {
      display: 'flex'
   },
};

const attachmentBodyStyles: ComponentSlotStylesPrepared<AttachmentBodyStylesProps, AttachmentVariables> = {
  root: ({ variables: v }): ICSSInJSStyle => ({
    display: 'flex'
  }),
};

const attachmentIconStyles: ComponentSlotStylesPrepared<AttachmentIconStylesProps, AttachmentVariables> = {
  root: ({ variables: v }): ICSSInJSStyle => ({
    display: 'flex'
  }),
};

const attachmentHeaderStyles: ComponentSlotStylesPrepared<AttachmentHeaderStylesProps, AttachmentVariables> = {
  root: ({ variables: v }): ICSSInJSStyle => ({
    display: 'flex'
  }),
};
```

#### Description of changes

This PR updates `Attachment` to use instead `Box`es composed components.
Adds `body` slot to `Attachment`.

#### Performance

Due enabled caching performance for rerenders was improved improved almost twice 🚀 

image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12674)


packages/fluentui/react-bindings/src/telemetry/useTelemetry.ts
```diff
@@ -34,8 +34,9 @@ export const getTelemetry = (displayName: string, telemetry: Telemetry | undefin
 
 const useTelemetry = (displayName: string, telemetry: Telemetry | undefined): UseTelemetryResult => {
   const composeOptions = useComposeOptions();
+  const telemetryName = composeOptions?.displayNames[composeOptions?.displayNames.length - 1] || displayName;
```
layershifter, line 37: Oops, the actual name is the last one

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
```diff
@@ -122,29 +124,17 @@ const Attachment: React.FC<WithAsProp<AttachmentProps>> &
...
       {(header || description) && (
-        <div className={classes.content}>
-          {Text.create(header, {
-            defaultProps: () => ({ styles: resolvedStyles.header }),
-          })}
-
-          {Text.create(description, {
-            defaultProps: () => ({ styles: resolvedStyles.description }),
-          })}
+        <div className="ui-attachment__content">
```
mnajdova, line 130: Just curious, why we don't have `AttachmentBody` component that will have the header and description props and we would create those inside that component?

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
```diff
@@ -168,13 +155,18 @@ Attachment.propTypes = {
   actionable: PropTypes.bool,
   description: customPropTypes.itemShorthand,
   header: customPropTypes.itemShorthand,
-  icon: customPropTypes.shorthandAllowingChildren,
+  icon: customPropTypes.itemShorthand,
```
mnajdova, line 158: Why was this changed? Does this means that we are changing the API here? I would expect that even if we have different component, to keep the same API..

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
```diff
@@ -122,29 +124,17 @@ const Attachment: React.FC<WithAsProp<AttachmentProps>> &
...
-          iconOnly: true,
-          text: true,
-          styles: resolvedStyles.action,
-          className: Attachment.slotClassNames.action,
-        }),
-      })}
-      {!_.isNil(progress) && <div className={classes.progress} style={{ width: `${progress}%` }} />}
+
+      {AttachmentAction.create(action)}
+      {!_.isNil(progress) && <div className="ui-attachment__progress" style={{ width: `${progress}%` }} />}
```
mnajdova, line 137: Why we don't have `AttachmentProgress` component? I understand that you are just mapping shorthands to components, so I'm just interested whether there is a solid reason for it not being one

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
```diff
@@ -122,29 +124,17 @@ const Attachment: React.FC<WithAsProp<AttachmentProps>> &
...
-        defaultProps: () => ({
-          iconOnly: true,
-          text: true,
-          styles: resolvedStyles.action,
-          className: Attachment.slotClassNames.action,
-        }),
-      })}
-      {!_.isNil(progress) && <div className={classes.progress} style={{ width: `${progress}%` }} />}
+
+      {AttachmentAction.create(action)}
```
mnajdova, line 136: It's interesting that none of the sub-components need any of the parent's props :)

packages/fluentui/react-northstar/test/specs/commonTests/isConformant.tsx
```diff
@@ -65,6 +66,8 @@ export default function isConformant(
   const { throwError } = helpers('isConformant', Component);
 
   const componentType = typeof Component;
+  // composed components store `handledProps` under config
+  const handledProps = Component.handledProps || (Component as ComposedComponent).fluentComposeConfig?.handledProps;
```
mnajdova, line 70: Shuoldn't this be:
```
const handledProps = (Component as ComposedComponent).fluentComposeConfig?.handledProps || Component.handledProps;
```

packages/fluentui/react-northstar/src/themes/teams/components/Attachment/attachmentStyles.ts
```diff
@@ -33,6 +33,21 @@ const attachmentStyles: ComponentSlotStylesPrepared<AttachmentStylesProps, Attac
 
       ...borderFocusStyles,
 
+      '& .ui-attachment__content': {
+        flex: 1,
+      },
+
+      '& .ui-attachment__progress': {
```
mnajdova, line 40: It's not obvious why this is implemented with selector. I would vote for a separate component, to keep it consistent with other slots..

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
layershifter: Merge issue, should be reverted

packages/fluentui/react-northstar/test/specs/commonTests/isConformant.tsx
layershifter: Hm... Good point. Actually these is no difference as there is no `handledProps` in static props

PR #12675, opened 2020-04-14
feat(Icons): update icon svg factory example & added components tests

mnajdova:
This PR is continuing the work from https://github.com/microsoft/fluentui/pull/12608

Things done in this PR:
- added example showing all icons
- added conformant test for the icon components

image


packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
-import { OneDriveIcon, YammerIcon, BorderBlindsIcon } from '@fluentui/react-icons';
+import * as ReactIcons from '@fluentui/react-icons';
+import { PrimaryButton } from 'office-ui-fabric-react/lib/Button';
 import { mergeStyles } from 'office-ui-fabric-react/lib/Styling';
 
 const rootClass = mergeStyles({
```
dzearing, line 6: minor suggestion, you can use `mergeStyleSets` here.

```tsx
mergeStyleSets({
  cell: {...},
  code: {...},
  navigationText: {...}
});
```

packages/react-icons/src/components.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+const iconsDir = 'src/components';
+
+describe('Icon components', () => {
+  fs.readdirSync(iconsDir).forEach(file => {
+    const componentName = path.basename(file, '.tsx');
+    // tslint:disable-next-line:no-any
+    const IconComponent = (icons as any)[componentName];
+
+    test(`${componentName} has 'Icon' suffix in the name`, () => {
+      expect(componentName.endsWith('Icon')).toBeTruthy();
```
silviuaavram, line 15: If this returns a `boolean` just check if it's `true`. Probably not hugely important, but it's more robust.

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
```diff
@@ -1,24 +1,74 @@
...
+const icons = Object.keys(ReactIcons).reduce((acc: React.FC[], exportName) => {
+  if (!!(ReactIcons as any)[exportName].displayName) {
+    acc.push((ReactIcons as any)[exportName] as React.FC);
+  }
+
+  return acc;
+}, []);
+
+const numOfIcons = icons.length;
+const numOfPages = parseInt((numOfIcons / 100).toString(), 10) + 1;
```
silviuaavram, line 43: what if there are 100 icons? are we getting 2 pages?

mnajdova: Fixed

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
```diff
@@ -1,24 +1,74 @@
...
+    borderRadius: '4px',
+    padding: '4px',
+  },
+  navigationText: {
+    width: 100,
+    margin: '0 5px',
+  },
 });
 
+const icons = Object.keys(ReactIcons).reduce((acc: React.FC[], exportName) => {
```
dzearing, line 34: We may want to not load all 2000+ icons here but rather a few key ones, or have a separate page as an icon viewing page.

dzearing: Note that this page ends up showing in both the public doc site and the snapshot tests.

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
```diff
@@ -1,24 +1,74 @@
...
+  if (!!(ReactIcons as any)[exportName].displayName) {
+    acc.push((ReactIcons as any)[exportName] as React.FC);
```
dzearing, line 36:
```suggestion
    acc.push((ReactIcons as any)[exportName] as React.FunctionComponent);
```

packages/react-icons/src/components.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+    test(`${file} is exported in package`, () => {
+      expect(IconComponent).toBeDefined();
+    });
+
+    test(`${file} is a valid component`, () => {
+      expect(ReactIs.isValidElementType(IconComponent)).toEqual(true);
+    });
+
+    test(`${file} matches displayName`, () => {
```
dzearing, line 26: May be good to add a few more conformance tests like it can take a `className` prop, what about passing along `data-` attributes, do we support `as` (I don't think so but can't remember.)

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
mnajdova: Actually I am showing 100 icons at a time, as I have pagination. I didn’t know what’s the best way of testing all icons with screenshots, so let’s sync ok that.

packages/react-icons/src/components.test.tsx
mnajdova: Will add the `data-` attributes and `className` tests, and you are right we do not support `as` prop.

mnajdova:
> Maybe slim down how many icons we show here. This could be merged first and then that could be addressed in the other uber PR.

I have pagination, I show only 100 icons on page. 

> Would be good to make sure we have a few more test cases covered though. :)

Will add the recommended tests.

> I will approve, let's get this one merged first. I also checked in a change to publish this package, as the fabric package was depending on a private unpublished package which broke some cases. My bad for not thinking about that. But we will look at beachball to prevent these scenarios in the future.

This is scary, how this even happend? Aren't we automatically releasing the packages created? :\ Would be good if you can reference some guide for this if there is.

PR #12677, opened 2020-04-14
fix(compose): add className as static and fix handling of handledProps

layershifter:
#### Pull request checklist

#### Description of changes

This PR:
- exposes `className` as statics (was missing in the initial implementation)
- `handledProps` in config include props from parent component, too

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12677)

### DIFF
--- a/packages/react-compose/src/compose.ts
+++ b/packages/react-compose/src/compose.ts
@@ -26,7 +26,10 @@ function computeDisplayNames<InputProps, InputStylesProps, ParentProps, ParentSt
 }
 
 function compose<InputProps, InputStylesProps, ParentProps, ParentStylesProps>(
-  InputComponent: React.FunctionComponent<ParentProps> & { fluentComposeConfig?: ComposePreparedOptions },
+  InputComponent: React.FunctionComponent<ParentProps> & {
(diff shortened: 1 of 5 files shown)


packages/react-compose/src/compose.ts
```diff
@@ -26,7 +26,10 @@ function computeDisplayNames<InputProps, InputStylesProps, ParentProps, ParentSt
 }
 
 function compose<InputProps, InputStylesProps, ParentProps, ParentStylesProps>(
-  InputComponent: React.FunctionComponent<ParentProps> & { fluentComposeConfig?: ComposePreparedOptions },
+  InputComponent: React.FunctionComponent<ParentProps> & {
+    fluentComposeConfig?: ComposePreparedOptions;
+    handledProps?: (keyof ParentProps)[];
```
mnajdova, line 31: Let's try to get rid of these in the future...

layershifter:
@dzearing FYI: this change a temporary we are going to get rid of `className` statics in all V0 components. The same fate will be for `handledProps`.

PR #12683, opened 2020-04-14
chore(Dropdown): add e2e test for closing on outside click

silviuaavram:
#### Pull request checklist

#### Description of changes

Follow-up from https://github.com/microsoft/fluentui/pull/12625 ,creating an e2e test for the outside close scenario.
#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12683)

### DIFF
--- a/packages/fluentui/e2e/tests/dropdownMoveFocusOnTab-test.ts
+++ b/packages/fluentui/e2e/tests/dropdownMoveFocusOnTab-test.ts
@@ -1,8 +1,9 @@
-import { selectors } from './dropdownMoveFocusOnTab-example';
+import { selectors, inputItems } from './dropdownMoveFocusOnTab-example';
 
 const triggerButton = `.${selectors.triggerButtonClass}`;
 const nextFocusableSibling = `#${selectors.nextFocusableSibling}`;
 const previousFocusableSibling = `#${selectors.previousFocusableSibling}`;
+const listItem = `.${selectors.listItem}`;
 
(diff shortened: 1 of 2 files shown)


packages/fluentui/e2e/tests/dropdownMoveFocusOnTab-test.ts
```diff
@@ -29,5 +30,15 @@ describe('Dropdown', () => {
       expect(await e2e.isFocused(triggerButton)).toBe(false);
       expect(await e2e.isFocused(previousFocusableSibling)).toBe(true);
     });
+
+    it('closes dropdown on outside click', async () => {
+      await e2e.clickOn(triggerButton);
+
+      expect(await e2e.count(listItem)).toBe(inputItems.length);
```
layershifter, line 37:
```suggestion
      expect(await e2e.count(listItem)).toBeGreaterThan(0);
```

In this case we don't need to have `inputItems` exported

silviuaavram: Interesting 😀although I am tempted to leave it like this, just in case the example will have some loading / no results indicators added at some point (those count as dropdown items) and we may get a false positive. I think it's more robust like this.

PR #12684, opened 2020-04-14
fix(Dropdown): make clearIndicator tabbable in non-search

silviuaavram:
#### Pull request checklist

#### Description of changes

To give users the ability to clear the selection by keyboard in a `select` dropdown we are making the indicator tabbable.
We are keeping it only clickable in a `combobox` context since clearing can be achievable by Escape.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12684)


packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -506,6 +506,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
                           className: Dropdown.slotClassNames.clearIndicator,
                           styles: styles.clearIndicator,
                           accessibility: indicatorBehavior,
+                          ...(!search && { tabIndex: 0 }),
```
pompomon, line 509: Should we move this logic into `indicatorBehavior` to be more consistent with other components? Right now we work with tabbIndex mostly in Behaviors only (apart from Dropdown and RadioGroup components)

pompomon, line 509: I think we should add some user-defined label, since the clearIndicator would be tabbable, but it won't have any aria-label to indicate it's purpose to screen-reader users

pompomon: I'd suggest to create a separate behavior for clearIndicator

silviuaavram: I think we should leave indicatorBehavior as it is. Based on this comment and the other one maybe we should create a new behavior, clearIndicatorBehavior or something, that will be tabbable, not have any aria-hidden and instead have aria-label.

@jurokapsiar @kolaps33 what do you think?

kolaps33: If I understand it correctly with this change there will be element which will have:
role: img,
aria-hidden: true
tabIndex: 0

What doesn't looks reasonable from my point of view.

I think would be better if clear indicator icon element has:
role: button 
tabIndex: 0

And in this case it would make sense to put in into new behavior.

kolaps33: or even wouldn't be there more clear way to use icon only button? 
    <Button icon="icon-close" text iconOnly title="Close" />
just asking as I don't know the code well if it is possible or not... :)

packages/fluentui/accessibility/src/behaviors/Box/dropdownClearIndicatorBehavior.ts
```diff
@@ -0,0 +1,24 @@
+import { Accessibility } from '../../types';
+
+/**
+ * @description
+ * Indicator is usually only visual representation and therefore is hidden from screen readers, unless 'aria-label' property is provided.
+ * Adds attribute 'tabIndex=-1' to 'root' slot if 'search' property is false. Does not set the attribute otherwise.
```
pompomon, line 6:
```suggestion
 * Adds attribute 'tabIndex=0' to 'root' slot if 'search' property is false. Does not set the attribute otherwise.
```

pompomon: And shouldn't it be in the `@specification` part?

silviuaavram: Thank you!

packages/fluentui/accessibility/src/behaviors/Box/dropdownClearIndicatorBehavior.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @specification
+ * Adds role='button'.
+ */
+const dropdownClearIndicatorBehavior: Accessibility<DropdownClearIndicatorBehaviorProps> = props => ({
+  attributes: {
+    root: {
+      role: 'button',
```
kolaps33, line 14: what do you think about the adding condition(!props.search) for the role as well?
I would propose it, because even element has no tabIndex, but if it is button for example JAWS can navigate there with Virtual cursor
I did testing with this example:
https://codesandbox.io/s/fluent-ui-example-2plsz?file=/example.js

PR #12686, opened 2020-04-14
chore(dependencies): update downshift to v5.0.5

silviuaavram:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12686)


packages/fluentui/docs/src/components/ComponentDoc/ComponentDoc.tsx
```diff
@@ -112,7 +112,9 @@ class ComponentDoc extends React.Component<ComponentDocProps, ComponentDocState>
       onRemove: item => `${item} has been removed.`,
     };
 
-    const getA11yStatusMessage = ({ isOpen, itemToString, previousResultCount, resultCount, selectedItem }) => {
+    // ToDo: remove default parameter from previousResultCount when issue is fixed
+    // https://github.com/downshift-js/downshift/issues/999
+    const getA11yStatusMessage = ({ isOpen, itemToString, previousResultCount = -1, resultCount, selectedItem }) => {
```
layershifter, line 117: This looks like a breaking change

silviuaavram: unfortunately yes, I messed that up, will create a PR with the fix for the issue and will get the typing back to normal.

silviuaavram:
@layershifter I added the version.

packages/fluentui/docs/src/components/ComponentDoc/ComponentDoc.tsx
layershifter: So, can we provide a temporary fix on our side to avoid breaking changes? I mean to have a wrapper around this function in `Dropdown`

silviuaavram: Or we can update all the way to 5.2.0 where this is fixed

PR #12690, opened 2020-04-14
Website: check for window before testing location

micahgodbolt:
These pages won't build statically if they are assuming window is present

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12690)

### DIFF
--- a/apps/fabric-website/src/utilities/location.ts
+++ b/apps/fabric-website/src/utilities/location.ts
@@ -3,7 +3,7 @@
  * has the Universal Header and Footer (UHF).
  */
 export const hasUHF: boolean =
-  typeof window === undefined
+  typeof window === 'undefined'
     ? false
     : window.location.hostname === 'developer.microsoft.com' ||
       window.location.hostname === 'developer.microsoft-tst.com';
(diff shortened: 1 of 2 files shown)


apps/fabric-website/src/utilities/location.ts
```diff
@@ -3,10 +3,15 @@
  * has the Universal Header and Footer (UHF).
  */
 export const hasUHF: boolean =
-  window.location.hostname === 'developer.microsoft.com' || window.location.hostname === 'developer.microsoft-tst.com';
+  window === undefined
```
ecraig12345, line 6: You need to do `typeof window === 'undefined'`; otherwise it will throw.

apps/fabric-website/src/utilities/location.ts
```diff
@@ -3,10 +3,15 @@
  * has the Universal Header and Footer (UHF).
  */
 export const hasUHF: boolean =
-  window.location.hostname === 'developer.microsoft.com' || window.location.hostname === 'developer.microsoft-tst.com';
+  typeof window === undefined
```
dzearing, line 6: I don't think this works.

`typeof window` would return a string, so this would always resolve as false.

ecraig12345: Yes, it needs to be `typeof window === 'undefined'` (with the quotes around undefined)

PR #12711, opened 2020-04-15
Remove tslint from v0 and clean up rules

ecraig12345:
For tslint, v0 is effectively only using `tslint-config-airbnb`, but it's also using the airbnb config for eslint. This means we can remove tslint from v0 with no loss of coverage (and build faster)! :tada:

Also remove some unnecessary overrides from other tslint configs: in particular, since the `@uifabric/tslint-rules` config no longer extends from a stricter Microsoft config, we can remove some settings that were intended to override that config.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12711)


packages/fluentui/docs/src/examples/components/Tree/Performance/TreeWith60ListItems.perf.tsx
```diff
@@ -55,6 +55,7 @@ const items = [
 // TreeTitle with just text, while the level 2 titles are rendered as ListItems.
 const titleRenderer = (Component, { content, header, headerMedia, media, ...restProps }) => {
   // as providing all props to List.Item was showing console errors, therefore reducing props
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
xugao, line 58: pls fix the rule instead: https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings

PR #12712, opened 2020-04-15
Enable classnames caching for Label and Icon

xugao:
#### Pull request checklist

#### Description of changes
1. Re-enable caching for Label & Icon and increased cache size for both.
I don't see a reason to disable them. David told me they were previously disabled bc classname passed to them kept mutating and causing cache miss. But I went through all use cases in our repo and don't find a case for them to keep on mutating (for the caching not being helpful). I increased their cache size from 50 to 100 bc they are 2 components that are commonly used by other components so a larger set of cache key variations can be considered expected. Note, the new cache size is enough for rendering all our examples (we know because `ComponentExamples.test` is passing). I think that is a good threshold to set given it's very unlikely a real app scenario can match the same usages of our components.

2. Added ability to show cache full warning from `classNamesFunction` by setting a flag in `FabricConfig`
    - Users can use this for debugging perf issue. 
    - We can enable the warning which can fail `ComponentExamples.test` if cache is unexpectedly full. This can help us catch bad cache busting code. 

3. Found and fixed a few constantly mutating styles prop in our code

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12712)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
@@ -37,8 +37,9 @@ const ANIMATIONS: { [key: number]: string | undefined } = {
 };
 
 const getClassNames = classNamesFunction<ICalloutContentStyleProps, ICalloutContentStyles>({
-  disableCaching: true,
+  disableCaching: true, // disabling caching becasue stylesProp.position mutates often
 });
+
 const BEAK_ORIGIN_POSITION = { top: 0, left: 0 };
 // Microsoft Edge will overwrite inline styles if there is an animation pertaining to that style.
 // To help ensure that edge will respect the offscreen style opacity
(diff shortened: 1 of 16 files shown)


packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -37,8 +37,9 @@ const ANIMATIONS: { [key: number]: string | undefined } = {
 };
 
 const getClassNames = classNamesFunction<ICalloutContentStyleProps, ICalloutContentStyles>({
-  disableCaching: true,
+  disableCaching: true, // disabling caching becasue stylesProp.position mutates often
```
khmakoto, line 40:
```suggestion
  disableCaching: true, // disabling caching because stylesProp.position mutates often
```

PR #12717, opened 2020-04-15
Remove EventGroup from ColorRectangle, ColorSlider, and Slider

behowell:
#### Pull request checklist

#### Description of changes

Replace uses of EventGroup with the 'on' function and _disposables pattern in ColorRectangle, ColorSlider, and Slider. Also update the 'on' function to accept React events.

#### Focus areas to test

Test mouse and touch handling in ColorPicker and Slider. This affected how those events are registered.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12717)


dzearing:
Woohoo!

image

packages/utilities/src/dom/on.ts
```diff
@@ -1,7 +1,9 @@
+import * as React from 'react';
+
 export function on(
   element: Element | Window,
   eventName: string,
-  callback: (ev: Event) => void,
+  callback: (ev: Event | React.SyntheticEvent<Element, Event>) => void,
```
ecraig12345, line 6: I don't think this is the right change--since the listener is added using `addEventListener`, it will always be a real event, not a React event. So if `callback` (as authored by `on`'s caller) was previously written to only expect React events, its types need to be updated to also handle real events.

behowell: Very good point, I just submitted https://github.com/microsoft/fluentui/pull/12753 to revert this change and update ColorRectangle and ColorSlider to handle native events.

PR #12731, opened 2020-04-16
chore: export consts for className and slotClassNames and replace static usages part 2

mnajdova:
Added const for className and slotClassNames inside `Box`, `Button`* and `Card`* components.

# BREAKING CHANGE
Removed `Card.slotClassNames` as there are already dedicated components for each different type of `Card`* component. Replace `Card.slotNames.componentName` usage with the accordin const `componentNameClassName`. For example

```tsx
// Before
import  { Card } from '@fluentui/react-northstar';
console.log(Card.slotClassNames.body)

// After
import  { cardBodyClassName } from '@fluentui/react-northstar';
console.log(cardBodyClassName);
```

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Card/Card.tsx
+++ b/packages/fluentui/react-northstar/src/components/Card/Card.tsx
@@ -55,14 +55,7 @@ export interface CardProps extends UIComponentProps {
 }
 
 export type CardStylesProps = Pick<CardProps, 'compact' | 'horizontal' | 'centered' | 'size' | 'fluid'>;
-
-export interface CardSlotClassNames {
-  header: string;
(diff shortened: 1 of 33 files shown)


packages/fluentui/react-northstar/src/components/Card/Card.tsx
```diff
@@ -129,15 +121,7 @@ const Card: React.FC<WithAsProp<CardProps>> &
 };
 
 Card.displayName = 'Card';
-Card.deprecated_className = 'ui-card';
-
-Card.slotClassNames = {
```
mnajdova, line 134: No need for this, as there are already dedicated Card components for each slot.

PR #12732, opened 2020-04-16
chore(SplitButton): Replace useRangeKnob by useNumberKnob on split bu…

assuncaocharles:
…tton example

#### Pull request checklist

#### Description of changes

We were using range knob before because number knob didn't allowed min and max props. Now it's fixed so replacing it back.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/docs-components/src/knobs/defaultComponents.tsx
+++ b/packages/fluentui/docs-components/src/knobs/defaultComponents.tsx
@@ -47,7 +47,7 @@ const KnobNumber: React.FunctionComponent<KnobNumberKnobComponentProps> = props 
       const newValue = parseInt(e.target.value, 10);
       const min = parseInt(props.min, 10);
       const max = parseInt(props.max, 10);
-      props.setValue(newValue < min ? props.min : newValue > max ? props.max : newValue);
+      props.setValue(newValue < min ? props.min : newValue > max ? props.max : newValue || props.min);
     }}
     min={props.min}
     max={props.max}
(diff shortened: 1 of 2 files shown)


packages/fluentui/docs-components/src/knobs/defaultComponents.tsx
```diff
@@ -47,7 +47,7 @@ const KnobNumber: React.FunctionComponent<KnobNumberKnobComponentProps> = props
       const newValue = parseInt(e.target.value, 10);
       const min = parseInt(props.min, 10);
       const max = parseInt(props.max, 10);
-      props.setValue(newValue < min ? props.min : newValue > max ? props.max : newValue);
+      props.setValue(newValue < min ? props.min : newValue > max ? props.max : newValue || props.min);
```
pompomon, line 50: _nit_
Please could you do that in a bit more readable way, and not ternary inside ternary and or to finish it all? :)

assuncaocharles: The alternative would be something like:

```javascript=
let newValue = parseInt(e.target.value, 10);
if(newValue < min){
    newValue = props.min;
} else if(newValue > max){
    new value = props.max;
}

props.setValue(newValue || props.min)
```

Would that be better?

pompomon: LGTM!
@layershifter , are you OK with that change?

assuncaocharles: Done

PR #12743, opened 2020-04-16
feat(Card): Add disabled prop

pompomon:
#### Pull request checklist

#### Description of changes

Add `disabled` prop and styles to Card

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12743)

### DIFF
--- a/packages/fluentui/react-northstar/src/themes/teams/components/Card/cardStyles.ts
+++ b/packages/fluentui/react-northstar/src/themes/teams/components/Card/cardStyles.ts
@@ -36,16 +36,15 @@ const cardStyles: ComponentSlotStylesPrepared<CardStylesProps, CardVariables> = 
       ...(p.compact && { padding: v.compactPadding }),
       ...(p.centered && { alignItems: 'center' }),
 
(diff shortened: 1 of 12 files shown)


packages/fluentui/react-northstar/src/themes/teams/components/Card/cardStyles.ts
```diff
@@ -36,13 +36,24 @@ const cardStyles: ComponentSlotStylesPrepared<CardStylesProps, CardVariables> =
...
 
-      ...(p.actionable && {
-        cursor: 'pointer',
-        ':focus-visible': {
-          boxShadow: v.boxShadowFocus,
-        },
-        ':active': {
-          boxShadow: v.boxShadowPressed,
+      ...(p.actionable &&
+        !p.disabled && {
```
mnajdova, line 40: Can we please not use negation in the styles condition. Just apply default styles and then override whatever needs to be overridden for disabled.

PR #12748, opened 2020-04-16
Pivot: componentRef fixed, extra DOM element removed.

dzearing:
#### Description of changes

In looking into `Pivot`, we found that `componentRef` was not resolving correctly resolving due to the move to `React.Component` from `BaseComponent`. Added a focus test to validate.

In the process, we saw that FocusZone was adding an unnecessary extra `div` to the hierarchy, not referenced in class names, and was causing the new test to fail due to thinking the extra element was not visible.

To address the test passing, I've simplified the DOM to remove the extra element. While we typically avoid DOM changes, I believe this one is safe to remove.

#### Focus areas to test

Pivot still works, passes all tests. Snapshots updated.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12748)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Pivot/Pivot.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/Pivot/Pivot.test.tsx
@@ -1,9 +1,9 @@
 import * as React from 'react';
 import * as renderer from 'react-test-renderer';
-
+import { mount } from 'enzyme';
 import { resetIds } from '../../Utilities';
 
-import { Pivot, PivotItem, PivotLinkFormat, PivotLinkSize } from './index';
+import { Pivot, PivotItem, PivotLinkFormat, PivotLinkSize, IPivot } from './index';
 
 describe('Pivot', () => {
   beforeEach(() => {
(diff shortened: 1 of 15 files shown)


packages/office-ui-fabric-react/src/components/Pivot/Pivot.test.tsx
```diff
@@ -22,6 +22,23 @@ describe('Pivot', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('can be focused', () => {
+    const pivotRef = React.createRef<IPivot>();
+
+    mount(
+      <Pivot componentRef={pivotRef}>
+        <PivotItem headerText="Link 1" headerButtonProps={{ 'data-is-visible': true }} />
```
joschect, line 30: Why does 'data-is-visible' need to be set here?

dzearing: When we say `focus()`, the `FocusZone.focus` method is called, which internally uses visibility detection to determine which item should be focused. Setting it here lets FocusZone know that this thing is visible, so that it doesn't skip over thinking it can't be focused.

PR #12750, opened 2020-04-16
Don't create a new item object whenever we update contextual menu

### LINKED ISSUE #12705: MenuItems Trigger Updates on any ContextualMenu Update
### Describe the issue:
OnRenderMenuList and RenderSectionItem in ContextualMenu.base.tsx make calls to _renderMenuItem and modify the contextualMenuItem they are using to render ContextualMenuButtons. This causes all ContextualMenuButtons to rerender as they detect a diff in the item prop they are passed in. This is causing perf regressions in persisted contextual menu subsequent showings. 

Proposed change is to alter the signature of _renderMenuItem to take in the fields we modify as parameters instead. The change would be scoped to the calls to _renderMenuItem made within OnRenderMenuList and RenderSectionItem. Will not change behavior as we are still wiring the same values to the menu item components. We just won't be passing them through the item object.

#### Actual behavior:
Passing in the same list of items to contextual menu is causing Contextual Menu Buttons to update

#### Expected behavior:
Passing in the same list of items should not trigger updates for Contextual Menu Buttons

Requested priority: High

Products/sites affected: Office online

tabrumle:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #12705

#### Description of changes

Contextual Menu does not take advantage of prop memoization as it will create a new object for the cotextualmenuitem we pass to menu controls. To fix this keep the item the same but extend the fields of the object so that we don't change the object reference and trigger unneeded updates.

#### Focus areas to test

Contextual Menu

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12750)

### DIFF
--- a/change/office-ui-fabric-react-2020-04-16-14-07-39-master.json
+++ b/change/office-ui-fabric-react-2020-04-16-14-07-39-master.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Change file",
+  "comment": "Change contextual menu renderMenuItem function to prevent prop mutation and unneeded rerenders",
   "packageName": "office-ui-fabric-react",
   "email": "tabrumle@microsoft.com",
   "dependentChangeType": "patch",
--- a/change/office-ui-fabric-react-2020-04-16-14-07-39-master.json
+++ b/change/office-ui-fabric-react-2020-04-16-14-07-39-master.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Change file",
+  "packageName": "office-ui-fabric-react",
+  "email": "tabrumle@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-04-16T21:07:39.237Z"
+}
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -490,8 +490,10 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     );
   };
 
-  // While we have access to the underlying menu item, we should not mutate or expand it into a new object as it will
-  // cause the menu items to always rerender even if the same props object is passed in.
+  /**
+   * !!!IMPORTANT!!! Avoid mutating `item: IContextualMenuItem` argument. It will
+   * cause the menu items to always re-render because the component update is based on shallow comparison.
+   */
   private _renderMenuItem = (
     item: IContextualMenuItem,
     index: number,
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -474,7 +474,6 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
         {menuListProps.items.map((item, index) => {
           let contextualMenuItemRenderProps = item as IContextualMenuItemRenderProps;
           contextualMenuItemRenderProps.index = index;
-          f;
           contextualMenuItemRenderProps.focusableElementIndex = indexCorrection;
           contextualMenuItemRenderProps.totalItemCount = menuListProps.totalItemCount;
           contextualMenuItemRenderProps.hasCheckmarks = menuListProps.hasCheckmarks;
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -493,7 +493,7 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     return (
       <ul className={this._classNames.list} onKeyDown={this._onKeyDown} onKeyUp={this._onKeyUp} role="menu">
         {menuListProps.items.map((item, index) => {
-          const contextualMenuItemRenderProps: IContextualMenuItemRenderProps = this._convertMenuItemToMenuItemRenderProps(
+          const renderProps: IContextualMenuItemRenderProps = this._convertMenuItemToMenuItemRenderProps(
             item,
             index,
             indexCorrection,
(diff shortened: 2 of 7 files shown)


packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -472,14 +472,13 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
...
         {menuListProps.items.map((item, index) => {
-          const menuItem = this._renderMenuItem({
-            ...item,
-            index,
-            focusableElementIndex: indexCorrection,
-            totalItemCount: menuListProps.totalItemCount,
-            hasCheckmarks: menuListProps.hasCheckmarks,
-            hasIcons: menuListProps.hasIcons,
-          });
+          let contextualMenuItemRenderProps = item as IContextualMenuItemRenderProps;
```
joschect, line 475: I believe that this winds up mutating the items list itself. Part of the reason why we create a new object each time is to ensure that we don't change information that shouldn't be changed. 

Since it seems like this was a breaking perf change, I think we should consider changing the type of the `IContextualMenuItemRenderProps`. It could be that we go to something like `IContextualMenuItemRenderProps {
iitem: IContextualMenuItem}` or that get's removed entirely and the type of the render function changes to`defaultMenuItemRenderer: (item: IcontextualmenuItem, otherProps: {the other props}`

@dzearing since this caused such a big perf change is it alright to consider making a breaking change?

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -129,6 +130,26 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
   // tslint:disable-next-line:deprecation
   private _classNames: IProcessedStyleSet<IContextualMenuStyles> | IContextualMenuClassNames;
 
+  private _convertMenuItemToMenuItemRenderProps = memoizeFunction(
```
tabrumle, line 133: There are some benefits and draw backs to adding this function.

Pro: 
We take advantage of shouldComponentUpdate for menu items and reduce menu item rerenders from ms time to microsecond time. For example this will reduce subsequent render times for Office Online apps by 10-15% based on test profiling I was running with and without the change.

con: 
We will use up more memory as we have to build the cache for the props and since we are using an object in the memoization this can grow the cache to its limit for each menu if the menu item changes on each opening or there isnt proper memoization happening on the implementation side.
We can limit this issue by reducing the cache limit for this function or by making this an external function so we will only hit the cache limit once.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
tabrumle: @joschect @dzearing friendly ping on this. I was able to revisit this in a way that will allow us to still be backwards compatible. This regression is noticeable for perf and fixing this will increase subsequent persistent menu renders for web apps by ~10-15% going off of profiles I've been running with the change.

joschect:
@tabrumle Firstly thanks for all your hard work on this, it's really awesome.

@dzearing I think that this fix is good, I have some concerns about the cache.

Overall, I think that we can probably revert the change that was made and think about another way to implement the same or similar signature given the large regression cost this would seem to be the quickest and safest way. @dzearing  do you have thoughts on that?

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
xugao: i think in case like this, what we need to do is to make sure we don't mutate the item props user passed instead doing memoization. if user passes mutating props (which can happen very easily :(), we don't end up filling up the cache

see my comment here: https://github.com/microsoft/fluentui/pull/12831#pullrequestreview-398729883

tabrumle: Thank you for the detailed response! I've made the changes proposed and have verified it addresses the issue.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -592,6 +594,11 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     return renderedItems;
   };
 
+  private _defaultMenuItemRenderer = (item: IContextualMenuItemRenderProps): React.ReactNode => {
+    const { index, focusableElementIndex, totalItemCount, hasCheckmarks, hasIcons, ...rest } = item;
+    return this._renderMenuItem(rest, index, focusableElementIndex, totalItemCount, hasCheckmarks, hasIcons);
```
xugao, line 599: actually, i think if we do the following we can get good perf for default renderer case too:
```
    const { index, focusableElementIndex, totalItemCount, hasCheckmarks, hasIcons } = item;
    return this._renderMenuItem(item, index, focusableElementIndex, totalItemCount, hasCheckmarks, hasIcons);
```

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -490,17 +490,19 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     );
   };
 
-  private _renderMenuItem = (item: IContextualMenuItemRenderProps): React.ReactNode => {
+  private _renderMenuItem = (
```
xugao, line 493: pls add a comment to comment on avoiding mutating `item` and explain briefly why

change/office-ui-fabric-react-2020-04-16-14-07-39-master.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Change file",
```
xugao, line 3: give meaningful comment to describe the change you are making

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -490,17 +490,21 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     );
   };
 
-  private _renderMenuItem = (item: IContextualMenuItemRenderProps): React.ReactNode => {
+  // While we have access to the underlying menu item, we should not mutate or expand it into a new object as it will
+  // cause the menu items to always rerender even if the same props object is passed in.
```
xugao, line 494:
```suggestion
  /**
   * !!!IMPORTANT!!! Avoid mutating `item: IContextualMenuItem` argument. It will
   * cause the menu items to always re-render because the component update is based on shallow comparison.
   */
```

joschect:
@tabrumle and @xugao Thanks for all your work! this looks like it's ready to go in and I'll merge once ready.

joschect:
Closing and reopening to trigger build

tabrumle:
@joschect Could you merge this in when you get a chance? We're blocked from moving to additional work items until we can rev our dependency to this fix

PR #12752, opened 2020-04-16
ContextualMenu: Improve perf by doing less merge styles

xugao:
#### Pull request checklist

#### Description of changes
1. Improve contextualmenu perf by doing less styles calculation. (see flamegrill result)
2. Found and fixed a bug in `classNamesFunction` that it doesn't traverse the `styleFunctionOrObject` unless it has `__cachedInputs__` (only applies when it's used with `styled`)
3. enable caching for `getClassNames` and `getContextualMenuItemClassNames` (i don't see reason to disable)

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12752)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -35,6 +35,7 @@ import {
   isIOS,
   isMac,
   initializeComponentRef,
+  memoizeFunction,
 } from '../../Utilities';
 import { hasSubmenu, getIsChecked, isItemDisabled } from '../../utilities/contextualMenu/index';
 import { withResponsiveMode, ResponsiveMode } from '../../utilities/decorators/withResponsiveMode';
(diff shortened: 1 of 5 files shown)


packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -543,19 +549,11 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
...
-        : undefined;
-
       // We need to generate default styles then override if styles are provided
       // since the ContextualMenu currently handles item classNames.
-      itemClassNames = mergeStyleSets(
-        getContextualMenuItemClassNames(getItemStyles, itemStyleProps),
-        getContextualMenuItemClassNames(menuItemStyles, itemStyleProps),
-        getContextualMenuItemClassNames(styles, itemStyleProps),
+      itemClassNames = getContextualMenuItemClassNames(
+        _getMenuItemStylesFunction(this._classNames.subComponentStyles?.menuItem, styles),
```
dzearing, line 555: hmm, it would be easy to frequently mutate this no? same with `styles`. Will this save computations or just fill up the memoization cache?

xugao: so we know we are not mutating `this._classNames.subComponentStyles?.menuItem` and `styles` is passed by user. 

so in case where user passed a mutating styles prop, it's going to get filled up and get bad perf (regardless of my change)
But in case user did the right thing, they going to get much better perf than before. 

my change should be only making things better but not worse :)

PR #12755, opened 2020-04-17
Staging @fluentui/react-next package for v8 major release changes

dzearing:
This change creates the `@fluentui/react-next` pre-release package for staging breaking changes which plan to be included in the next major release.

The theme for v8 is to bring customers along with all improvements. We will meticulously track breaking changes in the epic and provide an upgrade path for everything - ideally automated upgrade paths, but manual instructions when we can't automate.

All v8 changes will be tracked in issues linked by the v8 epic here: 

https://github.com/microsoft/fluentui/issues/12770

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12755)


abdullahsalem:
I think leaving the following sentence [(at this line)](https://github.com/microsoft/fluentui/pull/12755/files#diff-44a4f4e5c243f4e9bfdef5a7c5b80e37R3) could lead to confusion 
> The React-based front-end framework for building experiences for Microsoft 365

in particular, this part

> .. **for Microsoft 365**

I think mentioning a specific MS product/service here and linking it to Fluent UI would put the framework in a product/service context, as it was previously within Office's context, even if it is an umbrella ( e.g., Microsoft 365).

So I think it would be better if Fluent UI always be mentioned and linked to either Fluent UI vison, Fluent design language, or design system, as it was announced [here](https://developer.microsoft.com/en-us/office/blogs/ui-fabric-is-evolving-into-fluent-ui):
> On March 20,  we renamed our UI Fabric GitHub repo to reflect our **Fluent UI vision**

I think keeping it out of that product/service context, would give a clear image:
- Fluent UI belongs to Microsoft Fluent
- Microsoft uses Fluent UI to provide that design experience in its products/services
- A developer can use Fluent UI to provide that design experience too (_even if the target product/service dose NOT directedly belong to Microsoft 365_)

abdullahsalem:
Additionally, there are many good points were mentioned in that [announcement](https://developer.microsoft.com/en-us/office/blogs/ui-fabric-is-evolving-into-fluent-ui/), I think it would be very helpful if they could be added to README.md, somehow. 
Here are some:
> Fluent is Microsoft’s design-to-code system

>Our goal is simple—a shared foundation with open source engineering processes to create experience coherence, efficiency, and effectiveness. It enables us to do more with less.

Strengths of Fluent as a design language are implicitly strengths of Fluent IU framework, so I think mentioning a brief of strengths of Fluent design language in README.md is an important

packages/react-next/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@fluentui/react-next",
+  "version": "8.0.0-beta",
```
xugao, line 3: should be alpha?

packages/react-next/README.md
```diff
@@ -0,0 +1,16 @@
+# [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)
+
+**The React-based front-end framework for building experiences for Microsoft 365.**
+
+**Fluent UI 8**
```
ecraig12345, line 5: Shouldn't this be "Fluent UI React 8" (and "Fluent UI React 7" later)?

packages/react-next/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+This package represents a staging ground for breaking changes in v8. We will be
+staging potentially breaking updates here with the intent that
```
ecraig12345, line 8: unfinished sentence (also please combine the lines rather than manual line breaks)

packages/react-next/etc/react.api.md
```diff
@@ -0,0 +1,12 @@
+## API Report File for "@fluentui/react"
```
ecraig12345, line 1: Why is this here?

packages/react-next/README.md
ecraig12345: And add a note about API stability or lack thereof

packages/react-next/etc/react.api.md
dzearing: This is what API extractor generates. It fails without this. This will fill up with things once we've added code to this package.

ecraig12345: What I meant is that there are API review files for *both* `react` and `react-next`, which seems weird.

dzearing: oh weird! Thanks!

dzearing:
@abdullahsalem Thank you for the feedback! I've made a few updates to the README but will keep it short for this PR. It will definitely contain more information before shipping. For now I've linked to the tracking issue.

ecraig12345:
Build failed (and is now failing other PRs) because deps were out of sync in the new package. I fixed it in 77f67ba70feab37f85a00fdce892417b4dff2b13 but please watch for that type of issue when adding packages in the future.

PR #12745, opened 2020-04-16
feature: improve `compose()` signature

layershifter:
This PR updates the compose signature to allow us to use it for creating as well as composing compoennts.

Examples:

```jsx
const Button = compose<'button', ButtonProps, ButtonStylesProps, {}, {}>(
  (props, ref, composeOptions) => {
    // component's code
  },
  {
    className: buttonClassName,
    displayName: 'Button',

    handledProps: [
      'accessibility',
      'as',
      'children',
      'circular',
      'className',
      'content',
      'styles',
      'variables',
    ],
  },
)

```

```jsx

const AttachmentAction = compose<
  'button',
  AttachmentActionOwnProps,
  AttachmentActionStylesProps,
  ButtonProps,
  ButtonStylesProps
>(Button, {
  className: attachmentActionClassName,
  displayName: 'AttachmentAction',
});
```

PR #12757, opened 2020-04-17
chore: use proper tsconfig.json in docgen

layershifter:
#### Pull request checklist

#### Description of changes

During work on #12745 I have notices that my changes fail `gulp-plugin-docgen` without any reasons. During our investigation via @mnajdova we found out that default `tsconfig.json` that will be used for generating info files doesn't have required aliases (for `@fluentui/react-compose` as an example).

image

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12757)

### DIFF
--- a/scripts/gulp/plugins/util/docgen.ts
+++ b/scripts/gulp/plugins/util/docgen.ts
@@ -87,7 +87,15 @@ const defaultOptions: ts.CompilerOptions = {
   allowUnreachableCode: true,
 };
 
-const reactComponentSymbolNames = ['StatelessComponent', 'Stateless', 'StyledComponentClass', 'FunctionComponent'];
+const reactComponentSymbolNames = [
+  'StatelessComponent',
+  'Stateless',
+  'StyledComponentClass',
+  'FunctionComponent',
+
+  // magic for ComponentWithAs
+  '__type',
+];
 
 type MaybeIntersectType = ts.Type & { types?: ts.Type[] };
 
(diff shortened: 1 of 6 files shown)


scripts/gulp/plugins/util/docgen.ts
```diff
@@ -87,7 +87,15 @@ const defaultOptions: ts.CompilerOptions = {
...
 
-const reactComponentSymbolNames = ['StatelessComponent', 'Stateless', 'StyledComponentClass', 'FunctionComponent'];
+const reactComponentSymbolNames = [
+  'StatelessComponent',
+  'Stateless',
+  'StyledComponentClass',
+  'FunctionComponent',
+
+  // magic for ComponentWithAs
+  '__type',
```
ecraig12345, line 97: What was this meant to do? This addition of custom logic makes it harder to move back to the official `react-docgen-typescript`. :( @layershifter or maybe @mnajdova ?

layershifter: `react-typescript-docgen` uses TS AST, for some reason it returns `__type` from `.getSymbolName()`. I tested it with the latest `react-docgen-typescript` and in this case it works 🤔 But there is another blocking issue: https://github.com/styleguidist/react-docgen-typescript/issues/246

ecraig12345: Got it, in that case we can keep the fork for now.

PR #12763, opened 2020-04-17
chore: add isConformant test for exporting className consts

mnajdova:
Changed the `isConformant` test that checks for static className prop on the components to check for an `componentNameClassName` export in `@fluentui/react-northstar`

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
+++ b/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
@@ -84,14 +84,14 @@ export interface TreeState {
   activeItemIds: string[];
 }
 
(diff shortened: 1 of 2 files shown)


packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -84,14 +84,14 @@ export interface TreeState {
   activeItemIds: string[];
 }
 
-export const treeClassNames = 'ui-tree';
+export const treeClassName = 'ui-tree';
```
mnajdova, line 87: Catched one wrong name :)

pompomon: Nice!

PR #12765, opened 2020-04-17
chore: remove deprecated_className and slotClassNames 

mnajdova:
This PR removed the static `className` (`deprecated_className`) and `slotClassNames` props from the fluent components, in order to eliminate the component dependencies inside the theme definition.

**BREAKING CHANGES**

Replace all component's  `className` (`deprecated_className`) and `slotClassNames` props with the dedicated consts: `componentNameClassName` and `componentNameSlotClassNames`.

```jsx
// before
import { Button, MenuItem } from '@fluentui/react-northstar';

console.log(Button.className);
console.log(MenuItem.slotClassNames.indicator);

// after
import { buttonClassName, menuItemSlotClassNames } from '@fluentui/react-northstar';

console.log(buttonClassName);
console.log(menuItemSlotClassNames.indicator);
```


packages/fluentui/react-northstar/src/components/Accordion/Accordion.tsx
```diff
@@ -102,8 +102,6 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, Acco
 
   static deprecated_className = accordionClassName;
```
pompomon, line 103: You said in the description "Replace all component's className (deprecated_className)", but I see them still in place, is that OK?

pompomon: Or are we keeping `deprecated_className` for class components?

mnajdova: We still need them internally for the components that extending the `UIComponent` and `AucontrolledComponents`, but we don't want users to rely on those.

packages/fluentui/react-northstar/src/components/Flex/FlexItem.tsx
```diff
@@ -58,7 +58,7 @@ const applyStyles = (
   // if element is DOM element
   if (typeof element.type === 'string') {
     return React.cloneElement(element, {
-      className: cx(element.props.deprecated_className, classes.root),
+      className: cx(element.props.className, classes.root),
```
pompomon, line 61: Please how would that work with class components, which still have `deprecated_className` ?

mnajdova: This was unintended change where we are using the prop className, not the static field, but I accidentally replaced it with `deprecated_className` in one of the prev PRs.

PR #12768, opened 2020-04-17
chore(ComponentsExamples): Fix development console errors

assuncaocharles:
#### Description of changes

Some components examples are currently throwing errors in development, mostly missing unique keys. This PR solves some of this problems.

- add proper keys on `Accordion` examples
- add keys on `Alert` examples
- add keys on `ComponentPerfExample`
- Removed `body` and `dismissAction` from `Alert.defaultProps` to stop the following error:
    - ` Failed prop type: Prop 'body' in 'Alert' conflicts with props: 'children'.`
- add `content` for `Button` example so doesn't complain about lack of label
- remove checkedChanged from `RadioGroup` example since it's not a prop from `RadioGroupItem`
- remove dup files on from `RadioGroup` examples
- change duplicate id on Input examples

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12768)


packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
```diff
@@ -4,12 +4,12 @@ import { Accordion } from '@fluentui/react-northstar';
 const panels = [
   {
     key: 'first',
-    title: 'مرحبا العالم',
+    title: 'مرحبا العالم مرحبًا',
```
layershifter, line 7:
```suggestion
    title: { key: 'title-1', content: 'مرحبا العالم' },
```

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
```diff
@@ -4,12 +4,12 @@ import { Accordion } from '@fluentui/react-northstar';
 const panels = [
   {
     key: 'first',
```
layershifter, line 6:
```suggestion
```

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
```diff
@@ -4,12 +4,12 @@ import { Accordion } from '@fluentui/react-northstar';
...
-    title: 'مرحبا العالم',
+    title: 'مرحبا العالم مرحبًا',
     content: 'مرحبا العالم',
```
layershifter, line 8:
```suggestion
    content: { key: 'content-1', content: 'مرحبا العالم' },
```

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
```diff
@@ -4,12 +4,12 @@ import { Accordion } from '@fluentui/react-northstar';
...
   },
   {
     key: 'second',
-    title: 'مرحبا المريخ',
+    title: 'مرحبا المريخ مرحبًا',
     content: 'مرحبا المريخ',
```
layershifter, line 13:
```suggestion
    content: { key: 'content-2', content: 'مرحبا المريخ' },
```

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
```diff
@@ -4,12 +4,12 @@ import { Accordion } from '@fluentui/react-northstar';
...
   {
     key: 'first',
-    title: 'مرحبا العالم',
+    title: 'مرحبا العالم مرحبًا',
     content: 'مرحبا العالم',
   },
   {
     key: 'second',
-    title: 'مرحبا المريخ',
+    title: 'مرحبا المريخ مرحبًا',
```
layershifter, line 12:
```suggestion
    title: { key: 'content-2', content: 'مرحبا المريخ مرحبًا' },
```

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
```diff
@@ -4,12 +4,12 @@ import { Accordion } from '@fluentui/react-northstar';
 const panels = [
   {
     key: 'first',
-    title: 'مرحبا العالم',
+    title: 'مرحبا العالم مرحبًا',
     content: 'مرحبا العالم',
   },
   {
     key: 'second',
```
layershifter, line 11:
```suggestion
```

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
assuncaocharles: done

packages/fluentui/react-northstar/src/components/Alert/Alert.tsx
```diff
@@ -170,11 +168,11 @@ class Alert extends AutoControlledComponent<WithAsProp<AlertProps>, AlertState>
     const {
       actions,
       dismissible,
-      dismissAction,
+      dismissAction = {},
       content,
```
layershifter, line 172: I would rather suggest to fix `propTypes` and use for these `shorthandAllowingChildren`

assuncaocharles: Great, thanks... Didn't know about `shorthandAllowingChildren`

assuncaocharles: Done

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
layershifter: Please check one more time the original comment. It's wrong to have a key on an item, keys should exist on `title` and `content`.

https://github.com/microsoft/fluentui/blob/bb553acd81f8a6987ba49370a02e04fe2aaf3755/packages/fluentui/react-northstar/src/components/Accordion/Accordion.tsx#L258

```tsx
<Accordion>
  // No panel 🙁
  <Accordion.Title/>
<Accordion>
```

It's extremely confusing, in Semantic UI React is was solved via `AccordionPanel` component that handles keys.

https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Accordion/AccordionPanel.js

packages/fluentui/accessibility/src/behaviors/Button/buttonBehavior.ts
```diff
@@ -16,7 +16,7 @@ const buttonBehavior: Accessibility<ButtonBehaviorProps> = props => {
         role: props.as === 'button' ? undefined : 'button',
         tabIndex: props.as === 'button' ? undefined : 0,
         disabled: props.disabled && !props.loading ? (props.as === 'button' ? true : undefined) : undefined,
-        'aria-disabled': props.disabled || props.loading,
```
assuncaocharles, line 19: @layershifter @jurokapsiar @pompomon @mnajdova  I removed it, for 2 reasons: 

1. Although it was adding `aria-disabled="true"`when loading, the button wasn't disabled. Click could still be performed on it. 
2. It was throwing the following accessibility error:
`Inconsistent value of parameter 'disabled': 'false' != 'true' in class 'Button', $('[data-aa-error-id="1"]')`

Removing it and leaving it for the user explicit set the condition to disable makes sense for me.... If it's not the desired behavior we would have to proper disable the button when loading ( what is not currently happening ). 

So let me know your opinion.

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
assuncaocharles: I see, confusing indeed... Anyway, fixed!

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
```diff
@@ -585,7 +585,7 @@ Popup.propTypes = {
   positionFixed: PropTypes.bool,
   renderContent: PropTypes.func,
   target: PropTypes.any,
-  trigger: customPropTypes.every([customPropTypes.disallow(['children']), PropTypes.any]),
+  trigger: customPropTypes.shorthandAllowingChildren,
```
assuncaocharles, line 588: Changed so the example from `PopupCustomTargetExample` doesn't complain about conflict of props

layershifter: Please revert this and update an example as in: https://github.com/microsoft/fluentui/pull/12759

packages/fluentui/docs/src/examples/components/Button/Performance/ButtonMinimal.perf.tsx
```diff
@@ -1,7 +1,7 @@
 import { Button } from '@fluentui/react-northstar';
 import * as React from 'react';
 
-const ButtonMinimalPerf = () => <Button />;
+const ButtonMinimalPerf = () => <Button content="click" />;
```
layershifter, line 4: I don't that it's expected in this perf example as `content` will an additional slot

packages/fluentui/docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
```diff
@@ -38,15 +38,13 @@ class RadioGroupVerticalExample extends React.Component {
           </Text>
         ),
         value: 'custom',
-        checkedChanged: this.handleCustomCheckedChange,
+
```
layershifter, line 41:
```suggestion
```

packages/fluentui/docs/src/examples/components/Popup/Types/PopupExampleInline.tsx
```diff
@@ -1,11 +0,0 @@
...
-import { Button, Popup } from '@fluentui/react-northstar';
-import { MoreIcon } from '@fluentui/react-icons-northstar';
-
-const PopupExampleInline = () => (
-  <Popup content="This popup is rendered next to the trigger." inline>
-    <Button icon={<MoreIcon />} title="Show popup" />
-  </Popup>
-);
-
-export default PopupExampleInline;
```
layershifter, line 11: Just wondering, why these examples are removed?

packages/fluentui/docs/src/examples/components/Button/Performance/ButtonMinimal.perf.tsx
layershifter: image

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
assuncaocharles: Done

packages/fluentui/docs/src/examples/components/Button/Performance/ButtonMinimal.perf.tsx
assuncaocharles: Yeah, I understand there's a change on the screen, basically this:

Screen Shot 2020-04-18 at 10 07 05 AM
But otherwise it throws errors on console complaining about lack of label...

So I will change `label` for `aria-label="Minimal button"` so there's no visual changes and the error is gone... Not sure if it's proper so let me know if I should remove.

packages/fluentui/docs/src/examples/components/Popup/Types/PopupExampleInline.tsx
assuncaocharles: Because for what I understood non of these files were being used, all the example still there since there was the same files but with shorthand in the end `PopupExampleInline .shorthand` and exactly same code... 

If I am wrong just let me know and I will revert.

packages/fluentui/accessibility/src/behaviors/Button/buttonBehavior.ts
silviuaavram: I would actually expect the whole thing to be more like:
```js
  const isButton = props.as === 'button';
  const definition: AccessibilityDefinition = {
    attributes: {
      root: {
        role: isButton ? undefined : 'button',
        tabIndex: isButton ? undefined : 0,
        disabled: isButton && (props.disabled || props.loading),
        'aria-disabled': !isButton && (props.disabled || props.loading),
      },
    },
},
```

I would add `disabled` only if the element is a button, because `disabled` is supported only by form elements (buttons, inputs etc.) and on the condition that it is either `disabled` or `loading` via prop.
Consequently, I would add `aria-disabled` only if the element is not a button, just in case users create buttons without using native element. And the condition is the same as above.

I would expect screen readers to work with native `disabled` without the need of `aria-disabled`, in the case of elements that natively support it.

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
```diff
@@ -3,13 +3,11 @@ import { Accordion } from '@fluentui/react-northstar';
...
   {
-    key: 'first',
-    title: 'مرحبا العالم',
+    title: { key: 'content-1', content: 'مرحبا العالم' },
     content: 'مرحبا العالم',
   },
   {
-    key: 'second',
-    title: 'مرحبا المريخ',
+    title: { key: 'content-2', content: 'مرحبا المريخ' },
```
silviuaavram, line 10: cool! can you also make all the keys unique in the Accordion example? Esclusive and Exclusive and Expanded have the same keys (oups my fault) but since you created this PR ... :D

silviuaavram: also probably it is better to use `rtl-content-2` just to be super sure we will not use the same keys on another example. Same as below for alert

packages/fluentui/docs/src/examples/components/Alert/Usage/AlertExampleDismissActions.shorthand.tsx
```diff
@@ -8,8 +8,8 @@ const AlertExampleDismissActions = () => {
   return (
     <Alert
       actions={[
-        { content: 'Privacy policy', primary: true },
-        { content: 'Dismiss', onClick: () => setVisible(false) },
+        { content: 'Privacy policy', primary: true, key: 'privacy' },
```
silviuaavram, line 11: look here :D `privacy` is the same as in the example above. It might cause some strange behavior.

https://egghead.io/lessons/react-use-the-key-prop-when-rendering-a-list-with-react-12564a86

packages/fluentui/docs/src/examples/components/Input/Variations/InputExampleTargeting.shorthand.tsx
```diff
@@ -20,16 +20,22 @@ const propsTargettingInput = {
 const InputExampleTargeting = () => (
   <Grid columns="1fr 1fr" styles={{ justifyItems: 'start', alignItems: 'center', gap: '10px' }}>
     <Text content="Input with props that will be applied to either the input or the wrapper:" />
-    <Input {...propsForInput} {...propsForWrapper} />
+    <Input id="inputID-1" {...propsForInput} {...propsForWrapper} />
```
silviuaavram, line 23: at least use kebab case here. `id="input-targeting-1"` or something

packages/fluentui/docs/src/examples/components/Input/Variations/InputExampleTargeting.shorthand.tsx
```diff
@@ -20,16 +20,22 @@ const propsTargettingInput = {
...
+    <Input id="inputID-2" {...propsForInput} {...propsForWrapper} input={propsTargettingInput} />
 
     <Text content="Input with wrapper slot props that have to be applied to the wrapper element:" />
-    <Input {...propsForInput} {...propsForWrapper} wrapper={propsTargettingWrapper} />
+    <Input id="inputID-3" {...propsForInput} {...propsForWrapper} wrapper={propsTargettingWrapper} />
 
     <Text content="Input with input and wrapper slot props that have to be applied to the input and wrapper elements, respectively:" />
-    <Input {...propsForInput} {...propsForWrapper} input={propsTargettingInput} wrapper={propsTargettingWrapper} />
+    <Input
+      id="inputID-4"
```
silviuaavram, line 33: I think this will get overriden by `propsTargettingWrapper.id`.

packages/fluentui/docs/src/examples/components/List/Variations/ListExampleTruncate.shorthand.tsx
```diff
@@ -11,12 +11,14 @@ const actions = (
         iconOnly: true,
         text: true,
         title: 'Check',
+        key: 'check',
```
silviuaavram, line 14: as with other cases, maybe we should try to make these unique, maybe include the name of the example in their value, like 'example-truncate-check` or something.

packages/fluentui/docs/src/examples/components/Menu/Usage/MenuExampleWithTooltip.shorthand.tsx
```diff
@@ -5,7 +5,7 @@ const itemRenderer = (MenuItem, props) => {
   const { tooltip = '', ...rest } = props;
```
silviuaavram, line 5: we can destructure `key` from here and avoid passing it to `MenuItem` since it's on the `Tooltip` we actually need it.

packages/fluentui/docs/src/examples/components/Accordion/Rtl/AccordionExample.rtl.tsx
assuncaocharles: Done

packages/fluentui/docs/src/examples/components/Alert/Usage/AlertExampleDismissActions.shorthand.tsx
assuncaocharles: It makes no diff since are separated list. Does it?

packages/fluentui/docs/src/examples/components/Input/Variations/InputExampleTargeting.shorthand.tsx
assuncaocharles: Changed

packages/fluentui/docs/src/examples/components/Input/Variations/InputExampleTargeting.shorthand.tsx
assuncaocharles: Nop, it's not

packages/fluentui/docs/src/examples/components/List/Variations/ListExampleTruncate.shorthand.tsx
assuncaocharles: It's unique in this list.

packages/fluentui/docs/src/examples/components/Menu/Usage/MenuExampleWithTooltip.shorthand.tsx
assuncaocharles: Done

packages/fluentui/accessibility/src/behaviors/Button/buttonBehavior.ts
assuncaocharles: This still throws errors:

Screen Shot 2020-04-20 at 10 31 56 AM

Also, the button still clickable, so it didn't proper disable it.

pompomon: @kolaps33 please could you have a look at this change as well?

kolaps33: This is tricky, because when button get common 'disabled' attribute, then button lost focus to the body. Then reader start narrate the body of the page. What is not good. 
Aria-disabled ensure that for reader button is disabled but focus is not removed from the button. :)

Marat was also trying to solve this issue I am not sure what was his conclusion?

This is background why we didn't implemented 'disabled'.

packages/fluentui/docs/src/examples/components/Popup/Types/PopupExampleInline.tsx
assuncaocharles: Reverted. Thanks @pompomon for the explanations :D

packages/fluentui/accessibility/src/behaviors/Button/buttonBehavior.ts
assuncaocharles: But other than it should have or not the `disable` or `aria-disabled` my question would be why to make it dependent on the `loading` prop?

assuncaocharles: Talked with Marat about it. He will check the schema that it's throwing the error and I reverted these chances for now.

packages/fluentui/docs/src/examples/components/RadioGroup/Item/RadioGroupItemExample.shorthand.tsx
```diff
@@ -1,15 +1,11 @@
...
-const handleChange = () => {
-  alert('The radio checked value was changed!');
-};
-
 const RadioGroupItemExample = () => (
   <RadioGroup
     items={[
-      { key: '1', label: 'Make your choice', value: '1', checkedChanged: handleChange },
-      { key: '2', label: 'Another option', value: '2', checkedChanged: handleChange },
+      { key: '1', label: 'Make your choice', value: '1' },
```
pompomon, line 7: Looks like those could be just renamed see https://github.com/microsoft/fluent-ui-react/commit/d3e6c152549171ec31944bbddfaf0f232f14445f
```suggestion
      { key: '1', label: 'Make your choice', value: '1', onCheckedValueChange: handleChange},
```

Issue #12754, opened 2020-04-17
Out of memory- DetailsList examples


abdollahkahne:
### Environment Information

- **Package version(s)**: (fill this out)
- **Browser and OS versions**: Linux Ubuntu-Mozilla Firefox

### Describe the issue:

When I open the page related to this component to study documentations and examples, After pressing View Code and some scrolling, Browser Crashes! I try this with two different browsers (Mozilla and Chrome) and unfortunately both of them crashed.  I think there is some performance issue related to the example over there! I will appreciate if someone test it and share the result here.
Thanks in advance

vist0r:
Yes, I have same issue, first open https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist/innernavigation then ciick 'show code', the browser crashed.

3YH:
Same issue on List https://developer.microsoft.com/en-us/fluentui#/controls/web/list

donschaefer:
Same issue on https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist/customgroupheaders

dzearing:
Is this on Windows or Mac? Or either?

dzearing:
Reproing!

dzearing:
useEditor=0 qs param disables the live code editor, which seems to be causing the looping issue. Looking for a fix, but if we can't find one, I'll temporarily disable this feature.

joschect:
This is a dupe of #12643

dzearing:
We've disabled the code editing, so you should be able to view the examples now. We'll investigate what's going on more deeply and re-enable the editing feature once we have the right fix. Sorry for the trouble.

ecraig12345:
The crash should be unrelated to the foreign module error in the console (that's been around for awhile and has a different cause we haven't had time to address, and doesn't prevent the editor from working). Something is causing an infinite rendering loop, possibly related to measuring/resizing. The `EditorErrorHandler` error boundary component may also be handling things incorrectly.

abdollahkahne:
Unfortunately I exprienced something similar in the following Code Which It removed after I edited my code as the second version:
The Buggy Version
```
export default class Tasks extends React.Component<ITasksProps, ITasksState> {
  protected SPClient: SPClient;
  constructor(props) {
    super(props);
    this.state = { items: [] };
    this.SPClient = new SPClient(this.props.webPartContext);
    // this._getItems();
  }
  private _getItems() {
[...]

PR #12776, opened 2020-04-18
Disabling live code editing until the OOM error can be pinpointed.

dzearing:
Addresses #12754.

After a couple hours, I haven't been able to repro the OOM with a local deployment. I have an idea of where it's coming from but will take some refactoring to simplify the error handling of the live code editing.

This change disables live code editing until we have a real fix. I did not disable specific examples, as I don't know what the root cause is and can't guarantee there aren't other examples hitting this.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12776)

### DIFF
--- a/change/@uifabric-example-app-base-2020-04-17-17-53-58-fix-disable-editor.json
+++ b/change/@uifabric-example-app-base-2020-04-17-17-53-58-fix-disable-editor.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Doc site: disable live code editing until it does not cause infinite loops resulting in browser crashes.",
+  "packageName": "@uifabric/example-app-base",
+  "email": "dzearing@microsoft.com",
+  "dependentChangeType": "patch",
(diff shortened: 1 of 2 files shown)


dzearing:
The feature is not critical. Seeing the example and not crashing the browser is critical.

dzearing:
Also this is a temporary fix until we cab understand the root cause.

dzearing:
Merging this to buy some time to resolve properly. It's trivial to back out and trying to isolate the disabling to specific examples (which disabling in itself is a short term workaround) would be much more involved and there aren't any guarantees this won't happen again with other examples until we know why it's happening.

JustSlone:
Given the severity of the issue (browser tab crash/hang), let's take this simple fix to mitigate the issue and buy us some time to determine the right fix. We should find the right fix soon and not leave live code editing disabled.

PR #12779, opened 2020-04-19
Add `renderActiveOnly?` prop to Pivot component to control rendering of PivotItems 

### LINKED ISSUE #11965: Controlling Tab re-renders

Controlling pivotItem re-renders

- renderActiveOnly  Only the active pivotItem is rendered. Switching pivotItem unmounts the current pivotItem and mounts the new pivotItem.
- renderActiveOnly={false} All pivotItems are rendered on Pivot mount. Switching pivotItem hides the current pivotItem and shows the new pivotItem, without unmounting pivotItems.

this will prevent filled forms to re-render on pivotitems swith ...

Pivot

hecerinc:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11965

#### Description of changes

This PR:
- Adds a `renderActiveOnly` optional prop to the Pivot component to enable controlling of rendering of the PivotItems.
- Adds a story (example) to test the behaviour.

Expensive components (components that are costly to mount / render) have to be unmounted/ remounted with the current implementation every time a new PivotItem is shown. This PR adds a toggle (`false` by default, to maintain current behaviour) that instead renders everything on the first Pivot render and only hides the inactive PivotItems, thereby removing the extra mounts/unmounts.

Example of expensive components:

`<iframes>` have to load/reload everything every time a pivot changes. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12779)

### DIFF
--- a/change/office-ui-fabric-react-2020-04-19-16-24-23-herincon-pivotRendering.json
+++ b/change/office-ui-fabric-react-2020-04-19-16-24-23-herincon-pivotRendering.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Add pivot renderActive prop to controll re-rendering of pivot items",
+  "comment": "Pivot: Adding `alwaysRender` prop to `PivotItem` to allow for persistent `PivotItems` to control the amount of re-renders that happen.",
   "packageName": "office-ui-fabric-react",
   "email": "herincon@microsoft.com",
   "dependentChangeType": "patch",
--- a/change/office-ui-fabric-react-2020-04-19-16-24-23-herincon-pivotRendering.json
+++ b/change/office-ui-fabric-react-2020-04-19-16-24-23-herincon-pivotRendering.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Add pivot renderActive prop to controll re-rendering of pivot items",
+  "packageName": "office-ui-fabric-react",
+  "email": "herincon@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-04-19T23:24:23.500Z"
+}
--- a/packages/office-ui-fabric-react/src/components/Pivot/Pivot.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Pivot/Pivot.base.tsx
@@ -100,11 +100,16 @@ export class PivotBase extends React.Component<IPivotProps, IPivotState> {
     const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
 
     this._classNames = this._getClassNames(this.props);
+    const { renderActiveOnly } = this.props;
 
     return (
       <div role="toolbar" {...divProps}>
         {this._renderPivotLinks(linkCollection, selectedKey)}
-        {selectedKey && this._renderPivotItem(linkCollection, selectedKey)}
+        {renderActiveOnly !== false
+          ? selectedKey && this._renderPivotItem(linkCollection, selectedKey)
+          : linkCollection.links.map(link =>
+              this._renderPivotItem(linkCollection, link.itemKey, selectedKey === link.itemKey),
+            )}
       </div>
     );
   }
@@ -208,7 +213,11 @@ export class PivotBase extends React.Component<IPivotProps, IPivotState> {
   /**
    * Renders the current Pivot Item
    */
-  private _renderPivotItem(linkCollection: PivotLinkCollection, itemKey: string | undefined): JSX.Element | null {
+  private _renderPivotItem(
+    linkCollection: PivotLinkCollection,
+    itemKey: string | undefined,
+    isActive = true,
+  ): JSX.Element | null {
     if (this.props.headersOnly || !itemKey) {
       return null;
     }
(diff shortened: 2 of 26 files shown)


hecerinc:
@jdhuntington would you be able to help me out with a review on this PR, please?

packages/office-ui-fabric-react/src/components/Pivot/Pivot.base.tsx
```diff
@@ -208,7 +213,11 @@ export class PivotBase extends React.Component<IPivotProps, IPivotState> {
   /**
    * Renders the current Pivot Item
```
katiearriagam, line 214: nit: no necessarily the current Pivot Item - could you update this comment?

packages/office-ui-fabric-react/src/components/Pivot/Pivot.base.tsx
```diff
@@ -100,11 +100,16 @@ export class PivotBase extends React.Component<IPivotProps, IPivotState> {
...
+    const { renderActiveOnly } = this.props;
 
     return (
       <div role="toolbar" {...divProps}>
         {this._renderPivotLinks(linkCollection, selectedKey)}
-        {selectedKey && this._renderPivotItem(linkCollection, selectedKey)}
+        {renderActiveOnly !== false
+          ? selectedKey && this._renderPivotItem(linkCollection, selectedKey)
+          : linkCollection.links.map(link =>
+              this._renderPivotItem(linkCollection, link.itemKey, selectedKey === link.itemKey),
```
joschect, line 111: Another option for this, instead of passing in a renderActiveOnly, you could delegate this to _renderPivotItem. So instead of having `renderActiveOnly` prop, each item might take a `renderPivotItem` prop which would have a signature like `renderPivotItem: (props: {index, isActive, className}) => JSX.Element`. That way it can handle a bunch more scenarios. It would allow more selective rendering as well, in the instance where only 1 should persist, or only 1 shouldn't persist.

hecerinc: I agree, it definitely would make it more flexible, but looking at the feedback from the issue (and our own personal use case in Mission Control) it seems a bit overkill (YAGNI?) for what (IMO) should be provided out-of-the-box, + a lot more implementation from the consumer.

hecerinc:
@joschect  Can you help me out with a review here?

jdhuntington:
Sorry, I realize my comments are late on this PR. Thanks for submitting this PR - use-case makes complete sense.

One concern I have is with the number of props on the top level `Pivot` control. Could we instead push this logic down to each `PivotItem`? For instance, if it was called `alwaysRender`, it would be easy spot when looking at a component. 

```
    <Pivot>
      <PivotItem headerText="foo">
        foo
      </PivotItem>
      <PivotItem headerText="Expensive control" alwaysRender>
        iframe
      </PivotItem>
    </Pivot>
```

This will additionally give application authors the ability to have finer grained control over when a pivot is rendered.

Would also prefer that props don't default to true as `renderActiveOnly` does; the code should work the same whether `undefined` or `false` is passed.

hecerinc:
Reactivate

hecerinc:
@jdhuntington  I've addressed your feedback, can you help me with a review here?

packages/office-ui-fabric-react/src/components/Pivot/Pivot.base.tsx
```diff
@@ -206,9 +211,13 @@ export class PivotBase extends React.Component<IPivotProps, IPivotState> {
...
 
   /**
-   * Renders the current Pivot Item
+   * Renders a Pivot Item
    */
-  private _renderPivotItem(linkCollection: PivotLinkCollection, itemKey: string | undefined): JSX.Element | null {
+  private _renderPivotItem(
+    linkCollection: PivotLinkCollection,
+    itemKey: string | undefined,
+    isActive = true,
```
khmakoto, line 219: Since it seems like you'll always pass this value to the function, do you have to give it a default?

change/office-ui-fabric-react-2020-04-19-16-24-23-herincon-pivotRendering.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Add pivot renderActive prop to controll re-rendering of pivot items",
```
khmakoto, line 3:
```suggestion
  "comment": "Pivot: Adding `alwaysRender` prop to `PivotItem` to allow for persistent `PivotItems` to control the amount of re-renders that happen.",
```

packages/office-ui-fabric-react/src/components/Pivot/PivotItem.types.ts
```diff
@@ -63,4 +63,12 @@ export interface IPivotItemProps extends React.HTMLAttributes<HTMLDivElement> {
    * Optional keytip for this PivotItem
    */
   keytipProps?: IKeytipProps;
+
+  /**
+   * Whether to always render the pivot item (regardless of whether it is selected or not)
```
khmakoto, line 68:
```suggestion
   * Defines whether to always render the pivot item (regardless of whether it is selected or not).
```

packages/office-ui-fabric-react/src/components/Pivot/examples/Pivot.AlwaysRender.Example.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }, []);
+  return <div>{mounted ? 'Rendered' : 'Mounting...'}</div>;
+};
+
+export const PivotRenderActiveOnlyExample: React.FC = () => {
+  const [alwaysRender, setAlwaysRender] = React.useState(false);
+
+  const toggleActiveOnly = () => {
+    setAlwaysRender(!alwaysRender);
+  };
```
khmakoto, line 20: You can use the `useBoolean` hook from `@uifabric/react-hooks` to manage this for you.

hecerinc:
@khmakoto I've addressed your comments now :)

khmakoto:
@hecerinc I have approved the PR! Although it seems like you need to update snapshots in `react-next`.

hecerinc:
@jdhuntington could you help me out with a review here, please?

PR #12783, opened 2020-04-20
feat(factories): changed createShorthand signature, to use the optional shorthandConfig static on the component

mnajdova:
This PR is adding updated `createShorthand` utility, that based on the static `shorthandConfig` on the component, allows user to create react elements from the shorthand. As a showcase it is used on this PR to replace the `Button.create`, `AccordionTitle.create` and `AccordionContent.create` usages (this covers both functional and class components). This will allow us to have more flexible shorthand story, for allowing users to define their own components for the shorthands (this will be required later on on the compose factory).

**BREAKING CHANGES**
The `mappedProp`, `allowsJSX` and `mappedArrayProp` are expected to be defined as a static shorthandConfig on the component. If it is not defined, this are the defaults:
```jsx
mappedProp = 'children'
allowsJSX = true
mappedArrayProp = undefined
```

Example:
```jsx

// before
createShorthand({
  Component: Button,
  mappedProp: 'content',
  value: buttonShorthandValue,
  options = {},
});

// after
Button.shorthandConfig = {
   mappedProp: 'content',
};

createShorthand(Button, buttonShorthandValue, {});

```


packages/fluentui/react-northstar/src/utils/factories.ts
```diff
@@ -98,6 +104,7 @@ export function createShorthand<P>({
   value?: ShorthandValue<P>;
   options?: CreateShorthandOptions<P>;
 }) {
+  // console.log("mappedProp ", mappedProp);
```
mnajdova, line 107: TODO: remove this

packages/fluentui/react-northstar/src/utils/factories.ts
```diff
@@ -234,3 +241,20 @@ export function createShorthand<P>({
 
   return null;
 }
+
+export const createShorthand = <P>(
+  Component: React.ElementType<P> & { shorthandConfig?: ShorthandConfig },
+  value?: ShorthandValue<P>,
+  options?: CreateShorthandOptions<P>,
+) => {
+  const { mappedProp = 'children', allowsJSX = true, mappedArrayProp = undefined } = Component.shorthandConfig || {};
```
silviuaavram, line 250: Why do we need `mappedArrayProp = undefined` ? Isn't it going to be `undefined` anyway in that case?

packages/fluentui/react-northstar/src/components/Button/ButtonGroup.tsx
```diff
@@ -68,7 +69,7 @@ class ButtonGroup extends UIComponent<WithAsProp<ButtonGroupProps>, any> {
     return (
       <ElementType {...unhandledProps} className={classes.root}>
         {_.map(buttons, (button, idx) =>
-          Button.create(button, {
+          createShorthand<WithAsProp<ButtonProps>>(Button, button, {
```
layershifter, line 72: I expect that this will be inferred from the first param

packages/fluentui/react-northstar/test/specs/utils/factories-test.tsx
```diff
@@ -2,15 +2,15 @@ import { callable } from '@fluentui/styles';
...
+import { createShorthandInternal as createShorthand, createShorthandFactory } from 'src/utils';
 import { Props, ShorthandValue, ObjectOf, ShorthandRenderFunction } from 'src/types';
 import { consoleUtil } from 'test/utils';
 
 // ----------------------------------------
 // Utils
 // ----------------------------------------
 
-type ShorthandConfig = {
+type GetShorthandConfig = {
```
layershifter, line 13: Is this change intentional?

mnajdova: Yes, it conflicts with the actual ShorthandConfig, this however is used only in the test

packages/fluentui/react-northstar/src/components/Button/ButtonGroup.tsx
mnajdova: Make sense, I changed it

PR #12784, opened 2020-04-20
fix(Dropdown): don't clear value on same item selection

silviuaavram:
#### Pull request checklist

#### Description of changes

Checks on selection if item coming from Downshift is not undefined. This is the case when the same item is selected. If it's the same item, will not update `value` or trigger a selection a11y message.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12784)


packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -912,14 +912,18 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
       case Downshift.stateChangeTypes.keyDownEnter:
       case Downshift.stateChangeTypes.clickItem:
         const shouldAddHighlightedIndex = !multiple && items && items.length > 0;
+        const isSameItemSelected = changes.selectedItem === undefined;
```
pompomon, line 915:
```
const isSameItemSelected = changes.selectedItem === undefined;
const newValue = isSameItemSelected ? value[0] : changes.selectedItem;

newState.open = false;
newState.searchQuery = this.getSelectedItemAsString(newValue);
if (!isSameItemSelected) {
    newState.value = multiple ? [...value, changes.selectedItem] : [changes.selectedItem];
    newState.highlightedIndex = shouldAddHighlightedIndex ? items.indexOf(newValue) : null;

    if (getA11ySelectionMessage && getA11ySelectionMessage.onAdd) {
      this.setA11ySelectionMessage(getA11ySelectionMessage.onAdd(changes.selectedItem));        
    }
}
```
_nit_
Would having `newState.value`, `newState.highlightedIndex` and `this.setA11ySelectionMessage` inside `! isSameItemSelected` condition work the same?

silviuaavram: cool! it might work, let me try!

PR #12785, opened 2020-04-20
Callout: Updating positioning logic to account for dual screen scenarios

khmakoto:
#### Pull request checklist

#### Description of changes

This PR updates the positioning logic in the `Callout` component to make it dual-screen aware using the new `windowSegments` API. Components using `Callout` underneath such as `ContextualMenu` are also updated.

__Before:__
Before

__After:__
After

#### Focus areas to test

If anyone has any idea how to add tests for this type of scenarios please let me know.

### DIFF
--- a/change/office-ui-fabric-react-2020-04-20-13-28-44-dualScreenPositioning.json
+++ b/change/office-ui-fabric-react-2020-04-20-13-28-44-dualScreenPositioning.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Callout: Updating positioning logic to account for dual screen scenarios.",
+  "packageName": "office-ui-fabric-react",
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/utilities/positioning/positioning.ts
```diff
@@ -889,3 +889,73 @@ export function getMaxHeight(
 export function getOppositeEdge(edge: RectangleEdge): RectangleEdge {
   return edge * -1;
 }
+
+/**
+ * Window with typings for experimental features regarding Dual Screen devices.
+ */
+export interface IWindowWithSegments extends Window {
```
joschect, line 896: Should this be in types? And maybe we should keep this as internal for now until segments get added in more places?

PR #12786, opened 2020-04-21
Improve editor debugging

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Related to #12754

#### Description of changes

Always build the dev bundle of the website (possibly in addition to production bundle) to facilitate debugging.

Add support for `?useEditor=1` to force the editor on (for debugging the problem examples).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12786)

### DIFF
--- a/packages/example-app-base/src/components/ExampleCard/ExampleCard.tsx
+++ b/packages/example-app-base/src/components/ExampleCard/ExampleCard.tsx
@@ -168,7 +168,8 @@ export class ExampleCardBase extends React.Component<IExampleCardProps, IExample
               </div>
               {isCodeVisible ? (
                 <EditorWrapper
-                  useEditor={false}
+                  // Disable editor for list examples until crash can be figured out
+                  useEditor={!/^export (const|class) (List|DetailsList|GroupedList)\w+Example/m.test(latestCode)}
                   code={latestCode}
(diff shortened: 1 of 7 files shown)


packages/example-app-base/src/components/ExampleCard/ExampleCard.tsx
```diff
@@ -168,7 +168,11 @@ export class ExampleCardBase extends React.Component<IExampleCardProps, IExample
               </div>
               {isCodeVisible ? (
                 <EditorWrapper
-                  useEditor={false}
+                  // Disable editor for list examples until crash can be figured out
+                  useEditor={
```
dzearing, line 172: Could we just default to disabled unless useEditor=1 is set in the QS param? At least until we know why it's failing.

ecraig12345: I still disagree about this but have reverted it

dzearing: Thanks. We'll get it back online fully once we know why it is crashing the browser.

PR #12789, opened 2020-04-21
Use ts-node for v0 scripts and remove docs alias

ecraig12345:
Switch v0 scripts from using `@babel/register` to `ts-node`. This is a minor step in converging scripts.

Remove remaining `docs` aliases from v0 code.

I had to downgrade `ts-node` to version 7 because version 8 [disables caching](https://github.com/TypeStrong/ts-node/issues/672), and the perf without caching is abysmal (~30s/project for `clean`!). I also turned off type checking for perf reasons. While downgrading to the version with caching does involve some risk of cache-related issues, I've definitely seen similar issues with @babel/register especially while changing build scripts (unfortunately never pinned down repro steps), so switching to ts-node probably preserves a similar level of risk.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12789)


scripts/ts-node-register.js
```diff
@@ -1,3 +1,6 @@
-// Register ts-node so that it uses the scripts directory's tsconfig
+// Register ts-node so that it uses the scripts directory's tsconfig and only transpiles
 const tsNode = require('ts-node');
-tsNode.register({ dir: __dirname });
+tsNode.register({
+  dir: __dirname,
+  transpileOnly: true,
```
layershifter, line 5: So we will not have any benefits from type checks? Am I correct?

ecraig12345: I can turn on type checks if you want, I was just concerned about regressing build perf.

ecraig12345: In general I like type checks aside from the perf concerns :)

ecraig12345: Turned off type checking again due to perf concerns...nearly 30s per project for `clean`!! 😬

ecraig12345:
Note in case anyone else wants to try it, I've been testing perf with this absurdly long command intended to target only the projects still using gulp:
`lerna run clean --scope '@fluentui/!(react|examples|digest|perf-test|playground|react-theming|react-compose|react-focus|react-icons|keyboard-key)'`

ecraig12345:
There may also be issues with how ts-node handles output, at least the way that gulp does it. With gulp-based tasks, it shows all the output at once rather than as it comes. I haven't noticed this problem in the just-based tasks which also use ts-node, so not sure what's going on there.

PR #12791, opened 2020-04-21
fix(Carousel): handle the case where getItemPositionText is not passed

silviuaavram:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes https://github.com/microsoft/fluent-ui-react/issues/2337

#### Description of changes

Add a check for the `getItemPositionText` before rendering the pagination text.
Add Best Practices entry for users to provide the localised string for this prop.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12791)

### DIFF
--- a/packages/fluentui/react-northstar/test/specs/components/Carousel/Carousel-test.tsx
+++ b/packages/fluentui/react-northstar/test/specs/components/Carousel/Carousel-test.tsx
@@ -266,5 +266,12 @@ describe('Carousel', () => {
       expect(wrapper.state('activeIndex')).toEqual(1);
       expect(document.activeElement.firstElementChild.innerHTML).toEqual('item2');
     });
+
+    it('should show no pagination if getItemPositionText is not passed', () => {
+      const wrapper = renderCarousel({ getItemPositionText: undefined });
+      const paginationWrapper = getPaginationWrapper(wrapper);
+
+      expect(paginationWrapper.exists()).toBe(false);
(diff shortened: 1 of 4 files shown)


packages/fluentui/react-northstar/test/specs/components/Carousel/Carousel-test.tsx
```diff
@@ -266,5 +266,12 @@ describe('Carousel', () => {
       expect(wrapper.state('activeIndex')).toEqual(1);
       expect(document.activeElement.firstElementChild.innerHTML).toEqual('item2');
     });
+
+    it('should show no pagination if getItemPositionText is not passed', () => {
```
mnajdova, line 270: Don't we have screener tests for this? Just wondering why you decided with unit test, when we are testing similar things with screener? Also do we have a test for the opposite scenario?

silviuaavram: 1. Yes we have, although that depends on whether we are keeping the example with pagination in the docsite. I would rather have a unit test for this as well, we are making sure this one stays here and tests this case, and it gives faster feedback comparing it to the screener test.
2. Yep, it's `should show pagination if navigation prop is not passed` .

PR #12800, opened 2020-04-21
chore(Card): Prototype navigation in grid

pompomon:
#### Pull request checklist

#### Description of changes

Two examples of card grids - with clickable cards and with cards containing actionable children

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12800)


kolaps33:
Roman please ensure as well that for these new added Card***Behaviors.ts you will cover it by tests. 
I mean for example if you have
`[IS_FOCUSABLE_ATTRIBUTE]: true,`
then under
@specification 
will be line:

`* Adds attribute 'data-is-focusable=true' to 'root' slot.`

packages/fluentui/accessibility/src/behaviors/Grid/gridWithInnerZoneBehavior.ts
```diff
@@ -0,0 +1,23 @@
+import { Accessibility } from '../../types';
```
jurokapsiar, line 1: Please use more specific name - something like cardsContainerBehavior. There might be details that would be specific for the use case, let's discuss them offline.

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
```diff
@@ -0,0 +1,62 @@
...
+} from '@fluentui/react-northstar';
+import * as React from 'react';
+
+type SampleCardProps = {
+  index?: number;
+};
+
+const SampleCard: React.FC<SampleCardProps> = ({ index }) => {
+  return (
+    <Card accessibility={cardChildrenFocusableBehavior} aria-roledescription="card with action buttons">
```
jurokapsiar, line 21: let's use more real role description, something like 'user card'

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableGrid.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const ClickableCard: React.FC<ClickableCardProps> = ({ index }) => {
+  const [clickCount, setClickCount] = React.useState(0);
+  const updateClickCount = () => {
+    setClickCount(count => count + 1);
+  };
+
+  return (
+    <Card accessibility={cardFocusableBehavior} onClick={updateClickCount} aria-roledescription="clickable card">
```
jurokapsiar, line 15: let's use more real role description, something like 'user card'

packages/fluentui/accessibility/src/behaviors/Card/cardFocusableBehavior.ts
```diff
@@ -16,6 +17,7 @@ const cardFocusableBehavior: Accessibility<CardBehaviorProps> = props => ({
     root: {
       role: 'group',
       tabIndex: 0,
```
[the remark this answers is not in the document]
pompomon: It's there to showcase single focusable card example

packages/fluentui/react-northstar/src/utils/index.ts
```diff
@@ -25,6 +25,7 @@ export { getKindProp } from './getKindProp';
 export * from './whatInput';
 
 export * from './commonPropInterfaces';
+export { screenReaderContainerStyles } from './accessibility/Styles/accessibilityStyles';
```
pompomon, line 28: @mnajdova @layershifter , is that OK to export this style to be able to use it in prototypes and docs?

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
```diff
@@ -0,0 +1,100 @@
...
+  );
+};
+
+const CardExampleFocusableGrid = () => {
+  const navigationMessageId = 'navigationMessage';
+  const defaultMessage = 'Press arrow keys to navigate between cards.';
+  const [navigationMessage, setNavigationMessage] = React.useState(defaultMessage);
+  const timeout = React.useRef<ReturnType<typeof setTimeout>>();
+  const handleGridFocus = () => {
+    console.log(navigationMessage);
```
silviuaavram, line 46: don't forget about this one

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
```diff
@@ -0,0 +1,100 @@
...
+  // our consuming component gets unmounted
+  React.useEffect(() => cleanup, []);
+
+  return (
+    <>
+      <div id={navigationMessageId} style={screenReaderContainerStyles}>
+        {navigationMessage}
+      </div>
+      <Grid accessibility={cardsContainerBehavior} columns="3" onFocus={handleGridFocus} onBlur={handleGridBlur}>
+        <SampleCard index={1} aria-label="1 of 18" aria-describedby={navigationMessageId} />
```
silviuaavram, line 76: wouldn't it be simpler to use `_.times`?

silviuaavram: also in the case of the aria labels, does the user have context to know that he is on cards? so we can be sure that 1 of 18 means anything at this point.

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const cleanup = () => {
+    if (timeout.current) {
+      clearTimeout(timeout.current);
+    }
+  };
+
+  // make sure our timeout gets cleared if
+  // our consuming component gets unmounted
+  React.useEffect(() => cleanup, []);
```
silviuaavram, line 68: is this working? it looks like it should be called once, at mount, and that's it.

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableGrid.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <Image src="public/images/wireframe/square-image.png" />
+        </Flex>
+      </Card.Body>
+    </Card>
+  );
+};
+
+const CardExampleFocusableGrid = () => (
+  <Grid accessibility={cardsContainerBehavior} columns="3">
+    <ClickableCard index={1} />
```
silviuaavram, line 39: same comment with _.times.

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
pompomon: Yep, because cards have `aria-roledescription` attributes

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
assuncaocharles: I guess the `cleanup` will be triggered at `unmount`?

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
```diff
@@ -0,0 +1,92 @@
...
+  return (
+    <>
+      <div id={navigationMessageId} style={screenReaderContainerStyles}>
+        {navigationMessage}
+      </div>
+      <Grid accessibility={cardsContainerBehavior} columns="3" onFocus={handleGridFocus} onBlur={handleGridBlur}>
+        {_.times(cardsNumber, i => {
+          const nonZeroIndex = i + 1;
+          return (
+            <SampleCard
```
kolaps33, line 78: I tried just now with NVDA
image
and the card is missing label. I mean label in the way it narrate only "aria-roledescription" and "position". 

Please extend aria-label itself the easiest way probably. Please put it before the position in the string. 

Or you can use aria-labelledby="(userID) (cardID)" > in this structure it would narrate user name which is displayed in the card and then the aria-label + aria-roledescription + aria+describedby  from the card itself.
But you would need to add ID for user element and ID for card itself.

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const navigationMessageId = 'navigationMessage';
+  const defaultMessage = 'Press arrow keys to navigate between cards.';
+  const [navigationMessage, setNavigationMessage] = React.useState(defaultMessage);
+  const timeout = React.useRef<ReturnType<typeof setTimeout>>();
+  const cardsNumber = 18;
+  const handleGridFocus = () => {
+    if (navigationMessage !== null) {
+      timeout.current = setTimeout(() => {
+        setNavigationMessage(null);
+      }, 3000);
```
kolaps33, line 51: is 3 second enough for us, because when I tested it long time ago for VoiceOver it was not enough and was not narrating the instruction message. 
The NVDA and JAWS works fine with the 3 sec, tested before while in this branch :)

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
pompomon: Yep, it should be called on unmout to clean thing up
https://codesandbox.io/s/competent-grass-9ce2m?file=/src/App.tsx

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
pompomon: I've tested it in VoiceOver right now and with default setup 3 seconds was enough for it

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
pompomon: Updated `aria-labelledby` with ids, tested in VoiceOver, NVDA and JAWS, thanks @kolaps33

packages/fluentui/react-northstar/src/utils/index.ts
mnajdova: Should be ok as we don't depend on any site variables there.. Not sure if this is the right place though, but let's export it for now

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleFocusableChildrenGrid.tsx
```diff
@@ -0,0 +1,144 @@
...
```
mnajdova, line 102: Can you add here comment on how we decided on the `3000` :)

PR #12806, opened 2020-04-22
Popup: add onRestoreFocus method to repalce shouldRestoreFocus

joschect:
#### Pull request checklist

#### Description of changes

This fixes an issue where contextualmenu would not correctly return focus to the correct element when it is destroyed. This delegates the focus returning down to the popup determining when focus should be returned so that method is only called once, and gives a callback so that contextualmenu can determine what to do at that point.

I intentionally didn't make a helper type since we are still figuring out our focus story.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12806)


packages/office-ui-fabric-react/src/components/Popup/Popup.tsx
```diff
@@ -65,20 +66,12 @@ export class Popup extends React.Component<IPopupProps, IPopupState> {
...
-      // This slight delay is required so that we can unwind the stack, let react try to mess with focus, and then
-      // apply the correct focus. Without the setTimeout, we end up focusing the correct thing, and then React wants
-      // to reset the focus back to the thing it thinks should have been focused.
-      if (this._originalFocusedElement) {
-        this._originalFocusedElement.focus();
-      }
-    }
 
+    // tslint:disable-next-line: deprecation
+    if (this.props.shouldRestoreFocus) {
```
dzearing, line 71: any way to write a unit test to validate this works?

packages/office-ui-fabric-react/src/components/Callout/Callout.types.ts
```diff
@@ -243,12 +243,20 @@ export interface ICalloutProps extends React.HTMLAttributes<HTMLDivElement> {
...
 
   /**
-   * If specified, determines whether the underlying "Popup" component should try to restore
-   * focus when it is dismissed.  When set to false, the Popup won't try to restore focus to
-   * the last focused element.
-   * @defaultvalue true;
+   * If true, when this component is unmounted, focus will be restored to the element that had focus when the component
+   * first mounted.
+   * @defaultvalue true
+   * @deprecated use restoreFocus callback instead
```
dzearing, line 249:
```suggestion
   * @deprecated Use `onRestoreFocus` callback instead.
```

packages/office-ui-fabric-react/src/components/Callout/Callout.types.ts
```diff
@@ -243,12 +243,22 @@ export interface ICalloutProps extends React.HTMLAttributes<HTMLDivElement> {
...
   shouldRestoreFocus?: boolean;
+
+  /**
+   * Called when the component is unmounting, and focus needs to be restored.
+   * Argument passed down contains two variables, the element that the underlying
+   * popup believes focus should go to
+   * and whether or not the popup currently contains focus.
+   * If this is provided, focus will not be restored automatically, you'll need to
+   * restore it by calling originalElement.focus()
+   */
```
dzearing, line 260:
```suggestion
  /**
   * Called when the component is unmounting, and focus needs to be restored.
   * The callback will receive `options` argument containing the `originalElement` that
   * was focused on mount, and `containsFocus` which indicates if the popup still contains
   * focus during the unmount.
   * If this callback is provided, focus will not be restored automatically. You'll need to
   * restore it by calling `originalElement.focus()`.
   */
```

packages/office-ui-fabric-react/src/components/Popup/Popup.tsx
dzearing: Also a test to validate that focus is not restored if you provide an empty function.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -1,9634 +1,9644 @@
...
-    const // (undocumented)
-    size72 = "72px";
-    const // (undocumented)
-    size100 = "100px";
-    const // (undocumented)
-    size120 = "120px";
-}
-
-// @public
-export const Pivot: React.FunctionComponent<IPivotPro
```
khmakoto, line 1: Why is this file changing like this?

packages/office-ui-fabric-react/src/components/Callout/Callout.types.ts
```diff
@@ -243,12 +243,22 @@ export interface ICalloutProps extends React.HTMLAttributes<HTMLDivElement> {
...
+   * @defaultvalue true
+   * @deprecated use onRestoreFocus callback instead
    */
   shouldRestoreFocus?: boolean;
+
+  /**
+   * Called when the component is unmounting, and focus needs to be restored.
+   * Argument passed down contains two variables, the element that the underlying
+   * popup believes focus should go to
+   * and whether or not the popup currently contains focus.
```
khmakoto, line 257: nit: I feel like this separation is weird and should be more in line with the length of the lines above it.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -437,17 +434,11 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
...
-      // This slight delay is required so that we can unwind the stack, const react try to mess with focus, and then
-      // apply the correct focus. Without the setTimeout, we end up focusing the correct thing, and then React wants
-      // to reset the focus back to the thing it thinks should have been focused.
-      // Note: Cannot be replaced by this._async.setTimout because those will be removed by the time this is called.
-      setTimeout(() => {
-        this._previousActiveElement && this._previousActiveElement!.focus();
-      }, 0);
+  private _tryFocusPreviousActiveElement = (options?: any) => {
+    if (options && options.containsFocus && this._previousActiveElement) {
+      this._previousActiveElement && this._previousActiveElement!.focus();
```
khmakoto, line 439: You shouldn't need `!` here.

```suggestion
      this._previousActiveElement && this._previousActiveElement.focus();
```

packages/office-ui-fabric-react/src/components/Popup/Popup.tsx
```diff
@@ -65,20 +66,12 @@ export class Popup extends React.Component<IPopupProps, IPopupState> {
...
-    ) {
-      // This slight delay is required so that we can unwind the stack, let react try to mess with focus, and then
-      // apply the correct focus. Without the setTimeout, we end up focusing the correct thing, and then React wants
-      // to reset the focus back to the thing it thinks should have been focused.
-      if (this._originalFocusedElement) {
-        this._originalFocusedElement.focus();
-      }
-    }
 
+    // tslint:disable-next-line: deprecation
```
khmakoto, line 70: nit:

```suggestion
    // tslint:disable-next-line:deprecation
```

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
joschect: No idea. It's done that after a full build and merge with master. I'll try doing it again once I get tests going.

packages/office-ui-fabric-react/src/components/Callout/Callout.test.tsx
```diff
@@ -141,4 +141,72 @@ describe('Callout', () => {
...
+    // to be rendered into the real dom rather than the testutil simulated dom.
+    const root = document.createElement('div');
+    document.body.appendChild(root);
+    try {
+      ReactDOM.render<HTMLDivElement>(
+        <div>
+          <button id="focustarget"> button </button>
+          <button id="target" style={{ top: '10px', left: '10px', height: '0', width: '0px' }}>
+            {' '}
+            target{' '}
```
dzearing, line 172:
```suggestion
target
```

packages/office-ui-fabric-react/src/components/Callout/Callout.test.tsx
```diff
@@ -141,4 +141,72 @@ describe('Callout', () => {
...
+      visible = false;
+    };
+
+    // In order to have eventlisteners that have been added to the window to be called the JSX needs
+    // to be rendered into the real dom rather than the testutil simulated dom.
+    const root = document.createElement('div');
+    document.body.appendChild(root);
+    try {
+      ReactDOM.render<HTMLDivElement>(
+        <div>
```
dzearing, line 168: why not use enzyme `mount`? Does it create a problem?

dzearing: The problem here is that if this test were to fail, say on an assert, it leaves stuff in the dom. That means if another test ran looking for `#target` or `#focustarget` selectors, they would get the wrong element. It's important to try and leave the test environment clean when you're done and not leaking stuff to avoid side effects. You could just for example have a try/finally, or an `afterTest` callback, which unmounts and removes the root div if it exists. But I think enzyme mount tries to scope the render in its own little area. I think... i might be wrong :)

joschect: I'll add an after that cleans this.

PR #12815, opened 2020-04-22
Focus zone import path changed in charting packages

jameelakowsar:
#### Pull request checklist

#### Description of changes

Focus zone import path changed from '@fluentui/react-focus' to 'office-ui-fabric-react'

#### Focus areas to test

Grouped vertical and stacked bar chart focus events

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12815)


khmakoto:
I agree with @ecraig12345 here, this is effectively changing one thing for the same. Would you mind telling us the reasons why you want to do this?

jameelakowsar:
> I agree with @ecraig12345 here, this is effectively changing one thing for the same. Would you mind telling us the reasons why you want to do this?

Hi @ecraig12345 , @khmakoto ,

Yeah I am agree with you, Importing of 'Focus zone' same works for both OUFR and 'react focus'. In charting package dependencies, we have OUFR and not react-focus. If I add Focus zone from 'react-focus' I  need to add 'react-focus' dependency also.  Insted of that If we import from OUFR then no need to add one more dependency. That's why I have changed it.

PR #12819, opened 2020-04-22
Make dark theme inverted

chrismohr:
#### Pull request checklist

- [x] Include a change request file using `$ yarn change`

#### Description of changes

Setting the `isInverted` flag true on the dark sample theme will result in components rendering as expected on the website.  For example, on the [page for the MessageBar component](https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar), selecting dark mode does not show the expected colors.

(Note: after this change the colors for message bar in dark mode still won't be ideal, and I'll have a separate PR coming to address that)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12819)

PR #12267, opened 2020-03-11
MessageBar: update light and dark background colors

marygans:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [ ] Include a change request file using `$ yarn change`

#### Description of changes
 
Updates to background colors to reflect the current design point of view in light and dark theme.
[Resources](https://developer.microsoft.com/en-us/fabric#/resources) for reference. 

**Before (Light):**
Screen Shot 2020-03-10 at 11 04 31 AM

**Before (Dark):**
Screen Shot 2020-03-10 at 11 05 09 AM

**After(Light):**
Screen Shot 2020-03-10 at 11 04 31 AM

**After(Dark):**
Screen Shot 2020-03-10 at 11 49 32 AM

High Contrast Mode: 

message-bar-high-contrast

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/OfficeDev/office-ui-fabric-react/pull/12267)

PR #12821, opened 2020-04-22
Update message bar dark theme styles

chrismohr:
#### Pull request checklist

#### Description of changes

Before:
darkBefore

After:
darkAfter

- converted partially transparent colors to non-transparent values (eg. `rgba(255, 200, 10, .2)` => `#FFF4CE`) since we likely want to guarantee adequate contrast ratios regardless of the underlying container background color 

- updated dark mode background colors and icon colors to match designs

- updated light mode severe warning icon color to match designs

- minor refactoring to achieve a bit more code consistency than before; after my changes:
  - the icon and background color are now all independent from the current theme palette colors
  - added semantic slots for the icon colors
  - added semantic slot for severe warning background
  - deprecated unused slots

Happy to do some follow-up work with further guidance.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12821)

NOTE: There is also a [separate issue I've fixed](https://github.com/microsoft/fluentui/pull/12819) that is preventing the MessageBar from being shown accurately in dark mode on the website.


chrismohr:
The one screener change is expected.  It's the update to the severe warning icon color.

chrismohr:
@ecraig12345 can you review please?

ecraig12345:
Despite being listed as the codeowner, I'm really not the right one to review this. I know @betrue-final-final @phkuo @marygans and @micahgodbolt have worked on or commented on the (many) previous iterations of this change--can you take a look?

packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
```diff
@@ -39,19 +39,20 @@ const getRootBackground = (
...
   semanticColors: ISemanticColors,
+  isInverted: boolean,
 ): string => {
   switch (messageBarType) {
     case MessageBarType.error:
     case MessageBarType.blocked:
       return semanticColors.errorBackground;
     case MessageBarType.severeWarning:
-      return semanticColors.blockingBackground;
+      return isInverted ? '#4F2A0F' : '#FED9CC';
```
phkuo, line 49: don't hardcode colors in components

chrismohr: Hi @phkuo, thanks for taking a look. 

There's already a pretty strong precedent for hardcoding the colors for MessageBar. However, if you'd like me to move these colors elsewhere, or use another approach, please provide some more detailed guidance.

marygans: Hey @chrismohr ! I have a similar [PR(stale)](https://github.com/microsoft/fluentui/pull/12267/files) open for this. You can make new slots in Semantic Slots, for example 'severeWarningBackground' to avoid hard coded values in MessageBar styles directly

chrismohr: Thanks @marygans, I can add the slots.  Just curious if there's a reason you didn't move forward with your PR?  or are you still planning to?  If so, I can abandon this one.  (That said, I was hoping to get the change in this week.  We have a release next week that's depending on it, and rather than overriding the styles in our app, I thought it would be better to tackle it here.)

packages/styling/src/interfaces/ISemanticColors.ts
```diff
@@ -146,6 +146,10 @@ export interface ISemanticColors extends ISemanticTextColors {
   defaultStateBackground: string;
 
   //// Invariants - slots that rarely change color theme-to-theme because the color has meaning
+  /**
+   * Background for info messages.
```
phkuo, line 150: informational

packages/styling/src/styles/theme.ts
```diff
@@ -277,16 +277,26 @@ function _makeSemanticColorsFromPalette(p: IPalette, isInverted: boolean, depCom
...
+    infoBackground: !isInverted ? '#F3F2F1' : '#323130',
+    errorBackground: !isInverted ? '#FDE7E9' : '#442726',
+    blockingBackground: !isInverted ? '#FDE7E9' : '#442726',
+    warningBackground: !isInverted ? '#FFF4CE' : '#433519',
+    severeWarningBackground: !isInverted ? '#FED9CC' : '#4F2A0F',
+    successBackground: !isInverted ? '#DFF6DD' : '#393D1B',
 
     // Deprecated slots, second pass by _fixDeprecatedSlots() later for self-referential slots
+    warningHighlight: '',
+    warningText: '',
```
phkuo, line 298: please still assign these to their original values in _fixDeprecated

phkuo: like listTextColor

packages/styling/src/interfaces/ISemanticTextColors.ts
```diff
@@ -158,4 +147,14 @@ export interface ISemanticTextColors {
   /** @deprecated
    * This slot was incorrectly named. Use listText instead. */
   listTextColor: string;
+
+  /** @deprecated
```
phkuo, line 151: what replaced these if these are deprecated?

chrismohr: Maybe you can help me here...

Before my changes the only thing `warningText` was used for was the warning icon. I suspect that it one point it was also styling the text.  Before my change the text for all messageBars was `neutralPrimary`, however I thought `bodyText` would be a better fit. 

Now I'm second guessing that decision, and wondering if this should be another new slot instead.  Since all the MessageBar backgrounds are "invarient" colors, the text color should probably also be "invarient" as well.

What would you think of adding `messageText?`

chrismohr: @betrue-final-final maybe this is something that you can help clarify for me.  Which elements of the MessageBar should vary with the current theme, if any?  Should any of the colors be derived from the palette?

chrismohr: @phkuo Ok, I've added a few more new slots for this.  Please let me know what you think.

phkuo: Yeah, you shouldn't put neutralPrimary or bodyText on top of message bar background colors. Both of those colors (which should be the same) are designed to go on bodyBackground, not arbitrary backgrounds. Foreground and background sets should "match".

I would prefer we leave warningText in place as the default color for text on message bar backgrounds (instead of introducing a new messageText that does the same thing). Part of the odd name is that we are trying to keep the names agnostic of message bar itself, so it can be used elsewhere, like in forms to highlight issues, or in lists to highlight entire rows.

phkuo:
@betrue-final-final you'll want to verify that list row highlighting still meets contrast requirements in light and dark themes once this change goes in, since I'm not sure what colors it uses for foreground...

packages/styling/src/interfaces/ISemanticTextColors.ts
chrismohr: I'd like to understand more.  

Are you saying that I should use `warningText` as the text color for all message bars?  

I was avoiding that, since I thought it would potentially confuse consumers.  The challenge is that in the designs  each message type has its own icon color, but they all have the same text color. So if we have `errorIcon, warningIcon, severeWarningIcon etc`, if you see `errorText, warningText`, you would also expect there to be a `severeWarningText`.  I think I would be a little surprised if the text color inside a "success message" was `warningText`.  That's why I was looking for a more general name to capture all of them.  I chose `message` which I think is still agnostic from the component `messageBar`.

phkuo: I'd prefer we stick with the existing `warningText` since it already exists so theoretically it is already in use by customers. We're trying to NOT add semantic slots when we can.

I agree with your reasoning though. There are a few things I'd love to rename if given a second chance, but it's too late now :(
The best we can do is have thorough documentation and move on.

chrismohr: > There are a few things I'd love to rename if given a second chance, but it's too late now :(
> The best we can do is have thorough documentation and move on.

Really??  I understand the desire to not have an explosion of semantic colors, but feeling like the semantic colors can never change seems a bit unhealthy.  Is there someone we can loop in for another opinion on this?

phkuo: Well the purpose of semantic slots is that they have meaning. If they constantly cycle in and out then there's not much point to using them if they're just going to go away the next version. If you're making a button, you need to be sure that `buttonBackground` is going to stick around and not turn into `fluentButtonBackground` while `buttonBackground` gets retired and doesn't get updated anymore.
You can ping me on Teams if you want to discuss this in detail.

chrismohr: I understand the desire for stability, however there must be some middle ground between "constantly cycle" and "it's too late now".  That's why there is a deprecation protocol, right?

Are there any guidelines written up about how to manage changes to the theme's semantic color slots?

> the purpose of semantic slots is that they have meaning
Totally agreed, and I'm concerned about moving forward with a slot name thats meaning is inaccurate. Not just by a little, but by a lot.  Instead of `warningText` we could also pick `successText` and from my perspective it would be equally inaccurate.  That said, I acknowledge that different people will pick up different bits of meaning from different words, there's room for interpretation, and that's why I thought it would be helpful to get another opinion in here.

@dzearing @micahgodbolt what do you all think about this?

phkuo: Okay, fine, you've convinced me :)
There are no written guidelines beyond the promise of backwards-compatibility.
My only requirement then is that warningText is set to the value of messageText in the deprecated colors fixup code.

micahgodbolt:
@phkuo @betrue-final-final I don't see any difference/issues with row highlighting contrast colors. Can you be more specific? Otherwise, can we get this PR merged soon? 

image

packages/styling/src/interfaces/ISemanticTextColors.ts
chrismohr: > My only requirement then is that warningText is set to the value of messageText in the deprecated colors fixup code.

I can do this, but I'd like to understand the reasoning first, so that I can make sure I'm achieving the intended result. 

`warningText` is not being consumed within the FluentUI codebase, so the way I understand it, our goal would be to minimize disruption for those consumers that are reading the value for their own purposes.  I imagine their expectations are:

1. If I provide a `warningText` color to `createTheme`, the resulting theme should use the color provided. 
2. If I do not provide a value, the value will be the same as before this change. 

With the current code, both of these are true.  With your requirement, I think we could surprise users on both points.

What am I missing?

phkuo: Yes, that is the intention.
1) It still will. Customized semantic slots (and customized _any_ slots) get merged in after _makeSemantic(), on TOP of the result of that call. Therefore any customizations overwrite the default assignments. (theme.ts:L126)
2) But if you provide a value to messageText, you'd want to change the message bar text for consumers of warningText that are trying to achieve the same semantic scenario. When we deprecate, that doesn't mean we want to lock in its default value forever. We still want to support it as much as reasonably possible.

phkuo:
> @phkuo @betrue-final-final I don't see any difference/issues with row highlighting contrast colors. Can you be more specific? Otherwise, can we get this PR merged soon?

It'll be different on SP's default inverted black and inverted blue themes.

packages/styling/src/interfaces/ISemanticTextColors.ts
chrismohr: Thanks for pointing that out, I missed that detail. I've pushed up the change.

chrismohr:
@phkuo I've addressed your feedback, please take another look.

packages/styling/src/styles/theme.ts
```diff
@@ -277,16 +277,30 @@ function _makeSemanticColorsFromPalette(p: IPalette, isInverted: boolean, depCom
...
+    infoBackground: !isInverted ? '#f3f2f1' : '#323130',
+    errorBackground: !isInverted ? '#FDE7E9' : '#442726',
+    blockingBackground: !isInverted ? '#FDE7E9' : '#442726',
+    warningBackground: !isInverted ? '#FFF4CE' : '#433519',
+    severeWarningBackground: !isInverted ? '#FED9CC' : '#4F2A0F',
+    successBackground: !isInverted ? '#DFF6DD' : '#393D1B',
 
     // Deprecated slots, second pass by _fixDeprecatedSlots() later for self-referential slots
+    warningHighlight: !isInverted ? '#ffb900' : '#fff100',
+    warningText: !isInverted ? '#333333' : '#ffffff',
```
phkuo, line 302: set this to empty string since it gets filled out later (like listTextColor)

packages/styling/src/interfaces/ISemanticTextColors.ts
phkuo: Thanks for all your work!

packages/styling/src/styles/theme.ts
chrismohr: Oops, of course.  Done.  Thanks for catching that.

chrismohr:
@phkuo, now that you've approved, what needs to happen for this PR to move forward?

micahgodbolt:
@chrismohr just `yarn update-api` to get the new api snaps and we can push this through

chrismohr:
@micahgodbolt ok, I've resolved the conflicts, and all checks have passed.  Good to merge?

micahgodbolt:
:shipit:

ecraig12345:
@chrismohr I just noticed while investigating #13019 that you removed the style overrides for Links inside MessageBars--was there a specific reason for doing that? (Sorry for not noticing while reviewing before.)

PR #12824, opened 2020-04-22
Downgrade ts-node, script type fixes

ecraig12345:
Downgrade ts-node to version 7, which has caching (see #12789 discussion). 

Fix some type issues in script files.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12824)

### DIFF
--- a/scripts/gulp/plugins/util/getComponentInfo.ts
+++ b/scripts/gulp/plugins/util/getComponentInfo.ts
@@ -11,7 +11,7 @@ import parseDocblock from './parseDocblock';
 import parseType from './parseType';
 import getShorthandInfo from './getShorthandInfo';
 
-const getAvailableBehaviors = (accessibilityProp: ComponentProp): BehaviorInfo[] => {
+const getAvailableBehaviors = (accessibilityProp: ComponentProp | undefined): BehaviorInfo[] | undefined => {
   const docTags = accessibilityProp && accessibilityProp.tags;
   const availableTag = _.find(docTags, { title: 'available' });
   const availableBehaviorNames = _.get(availableTag, 'description', '');
@@ -87,7 +87,7 @@ const getComponentInfo = (tsConfigPath: string, filepath: string, ignoredParentI
   const isChild = !isParent;
   const parentDisplayName = isParent ? null : dirname;
   // "Field" for "FormField" since it is accessed as "Form.Field" in the API
-  const subcomponentName = isParent ? null : info.displayName.replace(parentDisplayName, '');
+  const subcomponentName = isParent ? null : info.displayName.replace(parentDisplayName!, '');
 
   // "ListItem.js" is a subcomponent is the "List" directory
   const subcomponentRegExp = new RegExp(`^${dirname}\\w+\\.tsx$`);
--- a/packages/fluentui/docs/src/types.ts
+++ b/packages/fluentui/docs/src/types.ts
@@ -38,7 +38,7 @@ export type ComponentInfo = {
   props: ComponentProp[];
   repoPath: string;
   subcomponentName: null | string;
-  subcomponents: string[];
+  subcomponents: string[] | null;
   type: 'component';
 };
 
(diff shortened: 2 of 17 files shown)


scripts/gulp/plugins/util/getComponentInfo.ts
```diff
@@ -11,7 +11,7 @@ import parseDocblock from './parseDocblock';
 import parseType from './parseType';
 import getShorthandInfo from './getShorthandInfo';
 
-const getAvailableBehaviors = (accessibilityProp: ComponentProp): BehaviorInfo[] => {
+const getAvailableBehaviors = (accessibilityProp: ComponentProp | undefined): BehaviorInfo[] | undefined => {
```
assuncaocharles, line 14:
```suggestion
const getAvailableBehaviors = (accessibilityProp?: ComponentProp): BehaviorInfo[] | undefined => {
```

scripts/gulp/plugins/util/getComponentInfo.ts
```diff
@@ -11,7 +11,7 @@ import parseDocblock from './parseDocblock';
...
   const docTags = accessibilityProp && accessibilityProp.tags;
```
assuncaocharles, line 15:
```suggestion
  const docTags = accessibilityProp?.tags;
```

scripts/gulp/plugins/util/getComponentInfo.ts
```diff
@@ -87,7 +87,7 @@ const getComponentInfo = (tsConfigPath: string, filepath: string, ignoredParentI
   const isChild = !isParent;
   const parentDisplayName = isParent ? null : dirname;
   // "Field" for "FormField" since it is accessed as "Form.Field" in the API
-  const subcomponentName = isParent ? null : info.displayName.replace(parentDisplayName, '');
+  const subcomponentName = isParent ? null : info.displayName.replace(parentDisplayName!, '');
```
assuncaocharles, line 90: Didn't got why the `!` added

scripts/gulp/plugins/util/getComponentInfo.ts
layershifter: Nope.

```ts
function a(a: string | undefined) { };
function b(a?: string) { };

console.log(b('a'))
// really optional
console.log(b())

console.log(a('a'))
// requires to pass a param
console.log(a())
```

https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=11&pc=17#code/GYVwdgxgLglg9mABAQwBTIFyIM5QE4xgDmiAPouACYCmwh1lAlIgN6IC+A3ALABQokWAkQAjdAH4suAsWZsufRbwgJscADbUAdOrhFUYgOTJDjRnwD0FxHmrJ16gJ6I4AByFh7fFWDWadegaoZko+ftq6+mjGpua8VjbUAI4gMLbYiFBwiK7I2BnIOch4yAC23qoaEYFoZkA

assuncaocharles: Alright, makes sense but in this case makes really difference?

ecraig12345: In this case the param is always passed, so we'll leave it as `ComponentProp | undefined`.

scripts/gulp/plugins/util/getComponentInfo.ts
ecraig12345: It's a non-null assertion operator (needed here when `strictNullChecks` is enabled). `parentDisplayName` will never actually be null in this case, but TS is unable to infer that for some reason.

assuncaocharles: Alright, makes sense. Thanks :)

scripts/gulp/plugins/util/getComponentInfo.ts
assuncaocharles: Alright :)

PR #12829, opened 2020-04-22
Updating List, Keytips, and Facepile examples to use function components

czearing:
#### Pull request checklist

#### Description of changes

Updated List, Keytips, and Facepile examples to use function components.

#### Focus areas to test

List, Keytips, and Facepile examples.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12829)


packages/office-ui-fabric-react/src/components/Keytip/examples/Keytips.CommandBar.Example.tsx
```diff
@@ -3,146 +3,103 @@ import { CommandBar } from 'office-ui-fabric-react/lib/CommandBar';
...
+];
 
-  private _hideMessageBar = (): void => {
-    this.setState({ showMessageBar: false });
-  };
-}
+export const KeytipsCommandBarExample: React.FunctionComponent = () => {
+  const [showModal, { toggle: toggleShowModal }] = useBoolean(false);
+  const [showMessageBar, { toggle: toggleShowMessageBar }] = useBoolean(false);
+  const commandBarItemProps = [
```
ecraig12345, line 76: This should probably be called `commandBarItems`, and in this case it would be worth doing `useMemo` because changing the identity of the command bar item array is likely to cause recalculations. (Once #12912 is merged you can do `useConst` instead. `useMemo` is needed right now because the `toggle` callback identity changes.)

packages/office-ui-fabric-react/src/components/Keytip/examples/Keytips.Dynamic.Example.tsx
```diff
@@ -1,64 +1,34 @@
...
-          text="Button 2"
-          onClick={this.setCurrButton('Button 2')}
-          keytipProps={{
-            content: 'GG2',
-            keySequences: ['gg2'],
-            onExecute: this.setCurrButton('Button 2'),
-            hasDynamicChildren: true,
-          }}
+          text={'Button 3, active button is: ' + currButton}
+          keytipProps={{ content: 'GG3', keySequences: [startSequence, 'gg3'] }}
```
ecraig12345, line 29: Should potentially do `useMemo` here too.

packages/office-ui-fabric-react/src/components/List/examples/List.Basic.Example.tsx
```diff
@@ -72,53 +54,32 @@ const classNames: IListBasicExampleClassObject = mergeStyleSets({
...
-
-  private _onFilterChanged = (_: any, text: string): void => {
-    this.setState({
-      filterText: text,
-      items: text
-        ? this._originalItems.filter(item => item.name.toLowerCase().indexOf(text.toLowerCase()) >= 0)
-        : this._originalItems,
-    });
+export const ListBasicExample: React.FunctionComponent = () => {
+  const [originalItems] = React.useState(() => createListItems(5000));
```
ecraig12345, line 72: You should do `useConst` here instead, passing in a callback (this is a perfect example of why I made it):
```suggestion
  const [originalItems] = useConst(() => createListItems(5000));
```

Same in all other list examples. 

You should also check to see how long the page takes to render now--as much as we generally want to avoid passing props to examples, not doing so in this case may cause a perf problem. If it is a perf problem, we could consider adding a memoized version of `createListItems` to `example-data`.

packages/office-ui-fabric-react/src/components/List/examples/List.Grid.Example.tsx
```diff
@@ -75,65 +66,49 @@ const classNames: IListGridExampleClassObject = mergeStyleSets({
...
-  private _columnCount: number;
-  private _columnWidth: number;
-  private _rowHeight: number;
-  private _items: IExampleItem[];
-
-  constructor(props: IListGridExampleProps) {
-    super(props);
-    this._items = props.items || createListItems(5000);
-  }
+const onRenderCell = (item: any, index: number | undefined): JSX.Element => (
```
ecraig12345, line 69: Same with other `item: any` in list examples
```suggestion
const onRenderCell = (item: IExampleItem, index: number | undefined): JSX.Element => (
```

packages/office-ui-fabric-react/src/components/List/examples/List.Scrolling.Example.tsx
```diff
@@ -2,18 +2,27 @@ import * as React from 'react';
...
 import { TextField } from 'office-ui-fabric-react/lib/TextField';
-import { Checkbox } from 'office-ui-fabric-react/lib/Checkbox';
 import { createListItems, IExampleItem } from '@uifabric/example-data';
 import { mergeStyleSets, getTheme, normalize } from 'office-ui-fabric-react/lib/Styling';
 
+const evenItemHeight = 25;
+const oddItemHeight = 50;
+const numberOfItemsOnPage = 10;
 const theme = getTheme();
+const dropdownOption = [
```
ecraig12345, line 14: dropdownOptions

packages/office-ui-fabric-react/src/components/Facepile/examples/Facepile.AddFace.Example.tsx
```diff
@@ -1,41 +1,32 @@
...
-      },
-      ariaDescription: 'To move through the items use left and right arrow keys.',
-    };
-
-    return <Facepile {...facepileProps} />;
-  }
-}
+  return (
+    <Facepile
+      personas={facepilePersonas.slice(0, numberOfFaces)}
```
ecraig12345, line 23: This and `overflowPersonas` should potentially be memoized because otherwise it's creating a new array each time (same in other examples). Not sure if that causes any significant recalculations in the facepile case specifically, but it can definitely be problematic for some components.

packages/office-ui-fabric-react/src/components/List/examples/List.Ghosting.Example.tsx
```diff
@@ -70,38 +55,32 @@ const classNames: IListGhostingExampleClassObject = mergeStyleSets({
...
-    return (
-      <FocusZone direction={FocusZoneDirection.vertical}>
-        <div className={classNames.container} data-is-scrollable={true}>
-          <List items={this._items} onRenderCell={this._onRenderCell} />
-        </div>
-      </FocusZone>
-    );
-  }
+export const ListGhostingExample: React.FunctionComponent = () => {
+  const [items] = React.useState(() => createListItems(5000));
```
ecraig12345, line 77: `useConst(() => createListItems(5000));`

Same in other list examples that currently use the `useState` pattern for item creation.

packages/office-ui-fabric-react/src/components/Facepile/examples/Facepile.AddFace.Example.tsx
```diff
@@ -1,41 +1,35 @@
 import * as React from 'react';
-import { Facepile, IFacepileProps, OverflowButtonType } from 'office-ui-fabric-react/lib/Facepile';
+import { Facepile, OverflowButtonType } from 'office-ui-fabric-react/lib/Facepile';
 import { facepilePersonas } from '@uifabric/example-data';
+// import { IFacepilePersona } from '../Facepile.types';
```
ecraig12345, line 4:
```suggestion
```

packages/office-ui-fabric-react/src/components/Facepile/examples/Facepile.Basic.Example.tsx
```diff
@@ -32,115 +27,87 @@ const styles = mergeStyleSets({
...
+  { key: PersonaSize.size40, text: PersonaSize[PersonaSize.size40] },
+  { key: PersonaSize.size48, text: PersonaSize[PersonaSize.size48] },
+];
+
 const checkboxStyles: Partial<ICheckboxStyles> = { root: { margin: '10px 0' } };
 
-export class FacepileBasicExample extends React.Component<{}, IFacepileBasicExampleState> {
-  constructor(props: {}) {
-    super(props);
+const personaPresence = (personaName: string): PersonaPresence => {
```
ecraig12345, line 40: This should probably be called something like `getPersonaPresence`.

packages/office-ui-fabric-react/src/components/Keytip/examples/Keytips.Button.Example.tsx
```diff
@@ -3,127 +3,90 @@ import { DefaultButton, CompoundButton, IButtonStyles } from 'office-ui-fabric-r
...
-
-export interface IKeytipsButtonExampleState {
-  btnDisabled: boolean;
-}
+import { useBoolean } from '@uifabric/react-hooks';
 
 const stackTokens: IStackTokens = { childrenGap: 20 };
 const marginBottom = { root: { marginBottom: 28 } };
 const splitButtonStyles: Partial<IButtonStyles> = { splitButtonContainer: { height: 32 } };
+const splitButtonProps = {
```
ecraig12345, line 11: Probably good to explicitly type this and buttonProps

packages/office-ui-fabric-react/src/components/List/examples/List.Grid.Example.tsx
```diff
@@ -4,24 +4,16 @@ import { List } from 'office-ui-fabric-react/lib/List';
...
-interface IListGridExampleClassObject {
-  listGridExample: string;
-  listGridExampleTile: string;
-  listGridExampleSizer: string;
-  listGridExamplePadder: string;
-  listGridExampleLabel: string;
-  listGridExampleImage: string;
-}
-
+let columnCount: number;
```
ecraig12345, line 9: These need to go inside the example, and you'll have to find some way (either useCallback/useMemo or refs) to preserve the updates across renders.

packages/office-ui-fabric-react/src/components/Keytip/examples/Keytips.Dynamic.Example.tsx
```diff
@@ -1,64 +1,39 @@
 import * as React from 'react';
 import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
 
-interface IKeytipsTestState {
-  currButton: string;
-}
+const keytipProps = {
```
ecraig12345, line 4: I'm not sure this is correct if you want to preserve the previous behavior of having buttons 1 and 2 have different keytips.

packages/office-ui-fabric-react/src/components/List/examples/List.Grid.Example.tsx
```diff
@@ -75,65 +64,60 @@ const classNames: IListGridExampleClassObject = mergeStyleSets({
...
-  private _columnWidth: number;
-  private _rowHeight: number;
-  private _items: IExampleItem[];
-
-  constructor(props: IListGridExampleProps) {
-    super(props);
-    this._items = props.items || createListItems(5000);
-  }
+export const ListGridExample: React.FunctionComponent = () => {
+  let columnCount: number = 0;
```
ecraig12345, line 68: I'm still not sure this is right. It does appear to work, but it feels wrong and I'm not convinced the values would always be persisted as expected. Storing the values in `React.useRef()` would ensure persistence (and is the closest analogue to the old code) but still feels wrong. @dzearing do you have any thoughts on the actual right/idiomatic thing to do here?

(Also, you can get rid of columnWidth and just set rowHeight directly instead.)

dzearing: yeah move columnwidth to a local variable within the function it's used in; it's not read anywhere.

rowheight is read, and well, we could store the value in something like a state object. We need to cache it for the lifetime of the component, but setting it should not trigger re-renders. Something like useRef(0), and we cache the value in `current`.

PR #12838, opened 2020-04-23
docs(Dropdown): add loading message in getA11yStatusMEssage

silviuaavram:
#### Pull request checklist

#### Description of changes

If a dropdown is loading, we can also let the screen reader user know by returning an appropriate message as the dropdown status.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12838)

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Dropdown/State/DropdownExampleLoading.shorthand.tsx
+++ b/packages/fluentui/docs/src/examples/components/Dropdown/State/DropdownExampleLoading.shorthand.tsx
@@ -15,6 +15,21 @@ const DropdownExampleLoading = () => {
       search
       items={inputItems}
       placeholder="Start typing a name"
+      getA11yStatusMessage={({ resultCount, previousResultCount }) => {
+        if (loading) {
+          return 'loading results';
+        }
+
+        if (!resultCount) {
+          return 'No results are available.';
+        }
+        if (resultCount !== previousResultCount) {
+          return `${resultCount} result${
+            resultCount === 1 ? ' is' : 's are'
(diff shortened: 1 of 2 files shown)


packages/fluentui/docs/src/examples/components/Dropdown/State/DropdownExampleLoading.shorthand.tsx
```diff
@@ -15,6 +15,21 @@ const DropdownExampleLoading = () => {
       search
       items={inputItems}
       placeholder="Start typing a name"
+      getA11yStatusMessage={({ resultCount, previousResultCount }) => {
+        if (loading) {
```
kolaps33, line 19: do we want to have the props in this example?
if yes should be better to add to this example as well following ?:
` const [loading] = useBooleanKnob({ name: 'loading', initialValue: true });`

silviuaavram: Are you sure? I am hitting that line every time when loading is true.

Screenshot 2020-04-28 at 9 12 36 AM

kolaps33: sorry I see I commented wrong file, I wanted to comment the file below
DropdownExampleHeaderMessage.shorthand.tsx 

yes in this file it was working fine :)

PR #12840, opened 2020-04-23
fix(Accordion): Render title as div by default

pompomon:
#### Pull request checklist

#### Description of changes

Fix accordion title to be rendered as `div` instead of `span`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12840)


packages/fluentui/react-northstar/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -157,7 +157,7 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
                 })}
                 {Box.create(content, {
                   defaultProps: () => ({
-                    as: 'span',
+                    as: 'div',
```
mnajdova, line 160:
```suggestion
```
'div' is the default :)

PR #12853, opened 2020-04-24
FocusZone: Converging defaultTabbableElement/defaultActiveElement and bringing shouldFocusInnerElementWhenReceivedFocus to v7

khmakoto:
#### Pull request checklist

#### Description of changes

This PR converges `defaultActiveElement` from v7 and `defaultTabbableElement` from v0 into a singular property `defaultTabbableElement` that extends its typing and implementation to account for the functionality in both cases. Another thing this PR does is bring `shouldFocusInnerElementWhenReceivedFocus` from v0 to v7.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12853)


packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -154,8 +154,11 @@ export default class FocusZone extends React.Component<FocusZoneProps> implement
     // Assign initial tab indexes so that we can set initial focus as appropriate.
     this.updateTabIndexes();
 
-    if (this.props.shouldFocusOnMount) {
-      this.focus();
+    if (this.props.defaultTabbableElement && typeof this.props.defaultTabbableElement === 'string') {
+      this._activeElement = this.getDocument().querySelector(this.props.defaultTabbableElement) as HTMLElement;
```
joschect, line 158: For both changes, is there a reason why defaultTabbableElement isn't called as a function here too?

jurokapsiar: it might be beacuse the code paths for string and function are actually different, but I would also like to understand it more

khmakoto: That's a valid concern here! The reason why this is not being called here is because it is called inside the `updateTabIndexes` function that appears just before this in line 155.

PR #12854, opened 2020-04-24
Provide full Nav group state for render overrides

ThomasMichon:
Reworked the `Nav` component's `onRenderHeader` override so that it receives the final `isExpanded` state along with an `onHeaderClick` that actually handles expand/collapse logic. This way, overrides can focus on altering the appearance and do not need to re-implement expand/collapse.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12854)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
@@ -4,7 +4,15 @@ import { buttonStyles } from './Nav.styles';
 import { classNamesFunction, divProperties, getNativeProps, getWindow, initializeComponentRef } from '../../Utilities';
 import { FocusZone, FocusZoneDirection, IFocusZone } from '../../FocusZone';
 import { Icon } from '../../Icon';
-import { INav, INavLink, INavLinkGroup, INavProps, INavStyleProps, INavStyles } from './Nav.types';
+import {
+  INav,
+  INavLink,
+  INavLinkGroup,
+  INavProps,
+  INavStyleProps,
+  INavStyles,
+  IRenderGroupHeaderProps,
+} from './Nav.types';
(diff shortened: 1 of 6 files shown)


packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
```diff
@@ -205,56 +213,79 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
...
     });
 
+    const finalOnHeaderClick = (
+      ev?: React.MouseEvent<HTMLElement, MouseEvent> | undefined,
+      isCollapsing?: boolean | undefined,
+    ) => {
+      this._onGroupHeaderClicked(group, ev);
+    };
+
+    const groupProps: IRenderGroupHeaderProps = {
```
joschect, line 233: As I recently learned with https://github.com/microsoft/fluentui/pull/12750 doing this can cause unwanted rerenders as each group is no longer considered the same object. OTOH we don't want to mutate state either. @xugao do you have any ideas?

xugao: in our default render for group header, we don't rely on any memoized props. so in that sense, this code is fine. But let's say user has a very expensive customized group header, the header IS receiving an always-mutating props. However i don't see how we can avoid the mutation here

ecraig12345:
Looks like you need a tslint disable
```
office-ui-fabric-react: [7:29:09 PM] x /mnt/work/3/s/packages/office-ui-fabric-react/src/components/Nav/Nav.test.tsx:44:14
office-ui-fabric-react: ERROR: 44:14  function-name  Function name does not match /^_?[a-z][a-zA-Z\d]+$/: LinkAs
```

PR #12856, opened 2020-04-24
 proto(dropdown at mentions): improvement to highlight first item when open

kolaps33:
#### Description of changes
During working on bug
https://github.com/microsoft/fluent-ui-react/issues/1193

I recognized I can do prototype improvement. :)

#### Focus areas to test
type "@" to at mention proto
Expected: first item in dropdown is highlighted 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12856)


packages/fluentui/docs/src/prototypes/mentions/MentionsDropdown.tsx
```diff
@@ -19,6 +19,7 @@ const MentionsDropdown: React.FunctionComponent<MentionsContainerProps> = props
     <Provider.Consumer
       render={({ siteVariables: siteVars }) => (
         <Dropdown
+          highlightFirstItemOnOpen={true}
```
silviuaavram, line 22: you can just pass it as `highlightFirstItemOnOpen` without any parameter.

kolaps33: thank you :)

PR #12858, opened 2020-04-24
feat(compose): add slots and mapPropsToSlotProps

mnajdova:
This PR introduced the `slots` and `mapSlotsToPropSlots` concepts in compose.

Those can be used in the following manner:

1. When defining a component:

```jsx
const Button = compose<'button', ButtonProps, ButtonStylesProps, {}, {}>(
  (props, ref, composeOptions) => {
    // component's logic
    const slotProps = composeOptions.resolveSlotProps(props);
    // some other code...
    const contentElement = createShorthand(composeOptions.slots.content, props.content, slotProps.content);
    // return..
  },
  {
    displayName: 'Button',
    // other compose configs
    slots: {
      content: ButtonContent,
      icon: Box,
      loader: Loader,
    },
    mapPropsToSlotProps: props => ({
      content: {
        size: props.size,
        as: 'span',
      },
      loader: {
        role: undefined,
      },
    }),
  }
);
```

Or when composing existing component

```jsx

const TertiaryButton = compose<'button', TertiaryButtonProps, TertiaryButtonStylesProps, ButtonProps, {}>(Button, {
  className: 'ui-tertiary-button',
  displayName: 'TertiaryButton',
  mapPropsToStylesProps: props => ({ tertiary: props.tertiary }),
  handledProps: ['tertiary'],
  slots: { content: TertiaryButtonContent },
  mapPropsToSlotProps: props => ({
    content: { tertiary: props.tertiary },
  }),
});
```

For now I did not include the root in the slots, but we may easily add it :)


packages/react-compose/src/types.ts
```diff
@@ -9,17 +14,31 @@ export type PropsOfElement<
...
+//
+//   defaultProps?: Partial<P & { as: E }>;
+//   propTypes?: React.WeakValidationMap<P> & {
+//     // tslint:disable-next-line:no-any
+//     as: React.Requireable<string | ((props: any, context?: any) => any) | (new (props: any, context?: any) => any)>;
+//   };
+// };
+
+// tslint:disable-next-line:interface-name
+export interface ComponentWithAs<E extends React.ElementType = 'div', P = {}> extends React.FunctionComponent {
```
mnajdova, line 30: @layershifter what do you think about this?

packages/fluentui/react-northstar/src/utils/factories.ts
```diff
@@ -260,6 +260,11 @@ export function createShorthand<E extends React.ElementType, P>(
   value?: ShorthandValue<P>,
   options?: CreateShorthandOptions<P>,
 ): React.ReactElement;
+export function createShorthand<TElementType extends React.ElementType>(
```
mnajdova, line 263: This is becoming over complicated :(

packages/fluentui/react-bindings/src/hooks/useSlotProps.ts
```diff
@@ -0,0 +1,20 @@
+type UseSlotPropsResult = (slot: string) => object;
+
+const useSlotProps = <P = {}>(
+  slotPropsChain: Record<string, (props: P) => Record<string, any> | Record<string, any>>[],
+  props: P,
+): UseSlotPropsResult => {
+  const getSlotProps: UseSlotPropsResult = slot => {
```
assuncaocharles, line 7: I wonder if we would have any benefits here by using `useCallback`:

```
const getSlotProps: UseSlotPropsResult = React.useCallback(
    slot => {
      return slotPropsChain.reduce(
        (acc, slotProps) => ({
          ...acc,
          ...(typeof slotProps[slot] === 'function' ? slotProps[slot](props) : slotProps[slot] || {}),
        }),
        {},
      );
    },
    [slotPropsChain, props],
  );
```

packages/react-compose/src/compose.ts
```diff
@@ -23,7 +23,7 @@ function compose<T extends React.ElementType, InputProps, InputStylesProps, Pare
     Component.defaultProps = (input as InputComposeComponent).defaultProps;
   }
 
-  (Component as ComposedComponent).fluentComposeConfig = composeOptions;
+  ((Component as unknown) as ComposedComponent).fluentComposeConfig = composeOptions;
```
assuncaocharles, line 26: Just out of curiosity, why the `unknown`?

mnajdova: Let me see if I can get rid of it. At some point ts was complaining that ComposedComponent has nothing in common with the initial Components typing, so I had to cast to `unknow`.

assuncaocharles: I tried to remove locally didn't got any errors, maybe it's fixed by now

packages/react-compose/src/types.ts
layershifter: If it works on all scenarios it will be great. But as I remember `interface` breaks type safety

packages/react-compose/src/types.ts
```diff
@@ -51,6 +70,9 @@ export type ComposeOptions<InputProps = {}, InputStylesProps = {}, ParentStylesP
 
   handledProps?: (keyof InputProps | 'as')[];
   overrideStyles?: boolean;
+
+  slots?: Record<string, React.ElementType>;
+  slotProps?: Record<string, (props: Extendable<InputProps>) => object>;
```
layershifter, line 75: This is hidden any, let's avoid it until it's really necessary

packages/fluentui/react-bindings/src/hooks/useSlotProps.ts
```diff
@@ -0,0 +1,20 @@
...
+    return slotPropsChain.reduce(
+      (acc, slotProps) => ({
+        ...acc,
+        ...(typeof slotProps[slot] === 'function' ? slotProps[slot](props) : slotProps[slot] || {}),
```
layershifter, line 11: Why this is not a part of `compose`?

layershifter: I mean the whole function

packages/fluentui/react-northstar/src/components/Button/Button.tsx
```diff
@@ -254,6 +263,22 @@ const Button = compose<'button', ButtonProps, ButtonStylesProps, {}, {}>(
...
+    slots: {
+      content: ButtonContent,
+      icon: Box,
+      loader: Loader,
+    },
+
+    slotProps: {
+      content: props => ({
+        size: props.size,
+        as: 'span',
```
layershifter, line 275: This is weird as this should be defined in `ButtonContent`

packages/fluentui/react-bindings/src/hooks/useSlotProps.ts
mnajdova: Do you mean we should have as part of `composeOption` the `getSlotProps` function, so we can just use `composeOptions.getSlotProps('content')` for example?

packages/fluentui/react-northstar/src/components/Button/Button.tsx
```diff
@@ -184,28 +186,37 @@ const Button = compose<'button', ButtonProps, ButtonStylesProps, {}, {}>(
...
-            role: undefined,
             styles: resolvedStyles.loader,
+            ...getSlotProps('loader'),
           }),
       });
     };
 
+    const renderContent = () => {
+      return createShorthand(composeOptions.slots.content, content, {
+        defaultProps: () => getA11yProps('content', getSlotProps('content')),
```
layershifter, line 216: This is extremely confusing 🥺

packages/fluentui/react-northstar/src/components/Button/Button.tsx
mnajdova: Agree, I will move it there...

packages/react-compose/src/types.ts
mnajdova: `slotProps?: Record<string, (props: InputProps) => object>;`
Like this you mean?

mnajdova: Or you are talking about `Record`?

layershifter: Yep

packages/fluentui/react-northstar/src/components/Button/Button.tsx
mnajdova: Yeah.. To be honest I never linked that we have to pass all props to the `getA11yProps` function :(

packages/fluentui/react-bindings/src/hooks/useSlotProps.ts
layershifter: Yep. Or even `composeOptions.slotProps.content` as I don't understand why we need `getSlotProps()` as function: to component we pass resolved options already

layershifter:
Why have you decided to go with `slotProps` instead of `mapPropsToSlots`?

```
mapPropsToSlots: (props) => ({ content: {} })
```

_just asking_

mnajdova:
> Why have you decided to go with slotProps instead of mapPropsToSlots?
> 
> ```
> mapPropsToSlots: (props) => ({ content: {} })
> ```
> just asking

@layershifter  actually this looks much closer to the other mappings in compose already.. 👍

packages/fluentui/react-bindings/src/hooks/useSlotProps.ts
mnajdova: We do not have the props inside compose, that is the reason why I initially made it a separate hook. If we return some function from compose, I think we will need to still invoke it with the properties inside the component. Am I missing something?

mnajdova: If we have `mapPropsToSlots` as a compose thingy, we will definitely need a hook in order to resolve them, again because we need the props, similarly as `useStyles` uses for now `unstable_props`

layershifter: But we have them all there 🤔

https://github.com/microsoft/fluentui/blob/a4a37e45bee51f66b5c8cabd87590df6e43c6e87/packages/react-compose/src/compose.ts#L17

mnajdova: So after some thoughts.. What I have is following identically what we do with `mapPropsToStylesProps` - compose is creating chain, and than the hook is invoking everything with the props. We don't have access to the props inside compose.

layershifter: It's a different as in `useStyles` we are using `mapPropsToStyles` in `useStyles` options. But here as functions in chain are available at `composeOptions`

mnajdova: Ok let me try to see if everything will work as expected then... I am going to update `slotProps` to be `mapPropsToSlots: (props) => ({ content: {} })` and try to remove the `useSlotProps` hook

mnajdova: Just to conclude, after offline discussions we decided to change API of compose to `mapPropsToSlots: (props) => ({ content: {} })`, but still use a hook for getting the slot props, as for some components we will need both props + state for resolving the slot props.

dzearing: My preference would be in `compose`: keep `slots` and `slotProps` as names. Either of these could be and static object map, or a function taking in `state` and returning the object map. It keeps the consumer experience super clean:

```tsx
compose(Button, {
  slots: {
     icon: MyContainer
  },
  slotProps: {
     icon: { data-is-foo: 'bar'
  }
}

We should consider moving `mapPropsToStyles` to something maybe higher level to be layered on demand: it is a css-in-js ism specifically for caching so as we move to static

I think a function like `mergeProps` would be responsible for deriving the final `slots` and `slotProps` to be distributed. It would need to be called after state has been reconciled. It could manage resolving things like slots/slotProps and classes functions which all should take in state to resolve their stuff.

This means that compose simply glues together the options and passes them along, while the mergeProps helper could finalize the results for consumption in the template.

File: packages/fluentui/react-northstar/src/components/Button/Button.tsx (hunk 8 above)
Comment (mnajdova) [reply packages/fluentui/react-northstar/src/components/Button/Button.tsx]: Looks better now :)

File: packages/fluentui/react-bindings/src/hooks/useSlotProps.ts (hunk 6 above)
Comment (mnajdova) [reply packages/fluentui/react-bindings/src/hooks/useSlotProps.ts]: We already have `mapPropsToStyleProps`, that's why I decided in the end to go with `mapPropsToSlotPros`. If we decide to change the API of the compose we can rename those easily to just stylesProps and slotProps, but for now I would rather keep the names close...

File: packages/react-compose/src/utils.ts [hunk 9]
```diff
@@ -24,13 +24,39 @@ export const defaultComposeOptions: ComposePreparedOptions = {
...
   inputOptions: ComposeOptions,
   parentOptions: ComposePreparedOptions = defaultComposeOptions,
 ): ComposePreparedOptions {
+  const mapPropsToSlotPropsChain = inputOptions.mapPropsToSlotProps
+    ? [...parentOptions.mapPropsToSlotPropsChain, inputOptions.mapPropsToSlotProps]
+    : parentOptions.mapPropsToSlotPropsChain;
+
+  const resolveSlotProps = <P = {}>(props: P) =>
+    mapPropsToSlotPropsChain.reduce<Record<string, object>>((acc, definition) => {
+      const nextProps = definition(props) || {};
```
assuncaocharles, line 43: nit:

```suggestion
      const nextProps =  {...definition(props)};
```

if `definition(props)` returns `undefined` the result of would be just `{}`. Same could be used for `slots` on line 73.

Does it make sense?

packages/react-compose/src/compose.ts
mnajdova: Actually it is ts error, it takes some time for it to process the changes.. I was still hitting this, so I left it as is..

packages/fluentui/react-northstar/src/components/Button/ButtonContent.tsx
```diff
@@ -32,9 +32,8 @@ ButtonContent.propTypes = {
   ...commonPropTypes.createCommon(),
   size: customPropTypes.size,
 };
-ButtonContent.create = createShorthandFactory({
-  Component: ButtonContent,
+ButtonContent.shorthandConfig = {
```
dzearing, line 35: any reason to not make this just a static `mappedProp`? is there anything else in `shorthandConfig`?

packages/react-compose/src/types.ts
dzearing: Would love to know more about when the right time to use type vs interface is.

PR #12865, opened 2020-04-24
[filetype icons] Add to filemap 'pbiapp' (re-applied)

bigbadcapers:
#### Pull request checklist

- [x] Addresses an existing issue: The entry 'pbiapp' (for non-file PowerBI items) was present in Fabric 6 but missing in Fabric 7. Causes no effect in existing apps.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12865)

### DIFF
--- a/packages/file-type-icons/src/FileTypeIconMap.ts
+++ b/packages/file-type-icons/src/FileTypeIconMap.ts
@@ -330,6 +330,7 @@ export const FileTypeIconMap: { [key: string]: { extensions?: string[] } } = {
   onetoc: {
     extensions: ['ms-one-stub', 'onetoc', 'onetoc2', 'onepkg'], // this icon represents a complete, logical notebook.
   },
+  pbiapp: {},
   pdf: {
     extensions: ['pdf'],
   },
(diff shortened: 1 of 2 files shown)


packages/file-type-icons/src/FileTypeIconMap.ts
```diff
@@ -330,6 +330,7 @@ export const FileTypeIconMap: { [key: string]: { extensions?: string[] } } = {
   onetoc: {
     extensions: ['ms-one-stub', 'onetoc', 'onetoc2', 'onepkg'], // this icon represents a complete, logical notebook.
   },
+  pbiapp: {},
```
dzearing, line 333: did you mean to not have any extensions match here?

bigbadcapers: This is a "non-classic" filetype, meaning it is not invoked by passing the file extension because it has none. Clients of the library would invoke it by passing this id as "ItemType" instead. Same situation as for the sharedfolder icon, Sway stories, Stream videos, and so on

bigbadcapers: The PowerBI-related icon that _does_ have extensions (because it represents the content types from PowerBI that are modelled as classic files) is `powerbi`

PR #12866, opened 2020-04-24
chore(ButtonGroup): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `ButtonGroup` component to be functional. Restricting props set that will be passed to styles functions.

Related to #12237

## Prop sets

| `ButtonGroup`    |
| --------- |
| `circular` |

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12866)


packages/fluentui/react-northstar/src/themes/teams/types.ts
```diff
@@ -149,7 +149,7 @@ export type TeamsThemeStylesProps = {
   ToolbarMenuRadioGroup: ToolbarMenuRadioGroupStylesProps;
   TooltipContent: TooltipContentStylesProps;
   Text: TextStylesProps;
-  TextAre: TextAreStylesProps;
+  TextArea: TextAreaStylesProps;
```
pompomon, line 152: Shouldn't we mention that as breaking change as well, since we are changing the exported name?

packages/fluentui/accessibility/src/behaviors/Button/buttonGroupBehavior.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * @specification
+ * Adds role 'group' to 'root' slot.
+ */
+
+const buttonGroupBehavior: Accessibility<ButtonGroupBehaviorProps> = props => ({
+  attributes: {
+    root: {
+      role: 'group',
```
pompomon, line 11: Please could you check with @kolaps33 if that role is needed for the button group?

packages/fluentui/react-northstar/src/themes/teams/types.ts
assuncaocharles: Not sure, we have been doing this in all conversions and I think we haven't mentioned? If we should I can just add it

PR #12869, opened 2020-04-24
Add react-northstar docs to PR deploy site

ecraig12345:
Deploy the react-northstar doc site as part of the PR deploy site.

Delete some images files from the docs package which appear to be unused. The images actually are not the main contributor to the size of the site (the JS bundle is ~23MB unminified!!), but removing unused stuff shouldn't hurt.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12869)


ecraig12345:
Marking as do not merge until I confirm it's actually working (which has been a process...)

layershifter:
@ecraig12345 why did you disable HMR? I think that it will be easier to deploy production build and as we did it previously.

ecraig12345:
Generally we do dev builds for PR deploy for easier debugging if needed, but I could do a prod build in this case if you prefer. (Any issues related to `__BASENAME__` differences in PR deploy would have to be fixed regardless.) 

Even with current code, the deployed version keeps defaulting to the 404 route, and I'm not sure why... (Without all the code to disable hot reload, I was getting errors in the console about that.)
http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/12869/merge/react-northstar/index.html

ecraig12345:
Feel free to mess with it if you have time today. There may just be something I don't know about react-router which is tripping me up, since the routing appears to be the only remaining issue.

miroslavstastny:
> There may just be something I don't know about react-router which is tripping me up, since the routing appears to be the only remaining issue.

I think the necessity to know the deploy path during build time is a wrong architecture. I have a WIP to switch to `HashRouter`: https://github.com/miroslavstastny/office-ui-fabric-react/commit/a081aeba324a9bb5d0299925c06995bab33df7d8
That is IMO cleaner, and the same thing Fluent7 site does.
@levithomason agrees with that (after some discussion).

There are several options:
1. if you are OK with that, I can finish that PR first and that you can continue on this PR
2. I can add that to this PR,
3. you can add it to this PR (keep in mind that that work is incomplete),
4. you can ignore my comment :-)

miroslavstastny:
> Generally we do dev builds for PR deploy for easier debugging if needed, but I could do a prod build in this case if you prefer.

For our existing use cases (e.g. sharing the link with external partners to do testing) we need a prod build.

apps/pr-deploy-site/index.html
```diff
@@ -118,7 +118,12 @@ <h1>Deployed sites for PR <a href="#" class="prLink"></a></h1>
       </li>
       <li class="Tile">
         <a href="./fabric-website/dist/index.html?devhost" class="Tile-link"
-          ><i class="ms-Icon ms-Icon--Website"></i>Web Site</a
+          ><i class="ms-Icon ms-Icon--Website"></i>Website</a
+        >
+      </li>
+      <li class="Tile">
+        <a href="./react-northstar/index.html" class="Tile-link"
```
layershifter, line 125:
```suggestion
        <a href="./react-northstar" class="Tile-link"
```

As we use exact matching:
```
<Route exact path="/" component={Introduction} />
```

layershifter:
Until we will be able to change 404 page (http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/12869/merge/react-northstar/components/alert/definition) I would say that we should go with `HashRouter`. But, I _personally_ don't like hash URLs 😭

ecraig12345:
> I think the necessity to know the deploy path during build time is a wrong architecture. I have a WIP to switch to `HashRouter`: [miroslavstastny@a081aeb](https://github.com/miroslavstastny/office-ui-fabric-react/commit/a081aeba324a9bb5d0299925c06995bab33df7d8)
> That is IMO cleaner, and the same thing Fluent7 site does.
> @levithomason agrees with that (after some discussion).

We actually would really like to move the v7 site *away* from hash routing (if we could find time...). It causes various problems like making scrolling to anchors not work natively and seems to mess up SEO. Also opinion-wise I think it looks messier.

> For our existing use cases (e.g. sharing the link with external partners to do testing) we need a prod build.

Interesting, this is definitely not the same use case we have. Do they test scenarios on the docs site itself, or somehow link their code to the bundle?

ecraig12345:
Reverted the hot loader changes and switched to a prod build, so we'll see what happens now.

Re: @layershifter's comment I think the custom 404 page not working is not a big deal since this is just a demo site, not anything truly user-facing.

miroslavstastny:
> But, I _personally_ don't like hash URLs 😭

URLs are not anything you should like. But if you and @ecraig12345 think it is better not to use `HashRouter` then go without that.

> Do they test scenarios on the docs site itself, or somehow link their code to the bundle?

Using the doc site directly.

layershifter:
> Re: @layershifter's comment I think the custom 404 page not working is not a big deal since this is just a demo site, not anything truly user-facing.

_It's a big deal_ 😨

Without custom 404 page we cannot redirect users from non-existing URLs and it means that I can't share a link to exact page, for example: http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/12869/merge/react-northstar/components/alert/definition

For Github and current docsite we use this trick: https://github.com/microsoft/fluentui/blob/master/packages/fluentui/docs/src/404.html

ecraig12345:
But I still don't understand why not having 404 is a big deal if it **only affects PR deploy**, which is only used for testing? I also don't understand how lack of 404 could be breaking deep links within the site. (Also FYI the site is broken in the latest PR deployed version...fixing it now.)

ecraig12345:
Basically I think having *some* PR deploy experience now is better than waiting until we can set up the perfect PR deploy experience.

ecraig12345:
Reverted the basename in paths change. Talked to @layershifter and he said to go ahead and merge as-is and we can get deep links working later.

ecraig12345:
Actually was able to add a 404 with redirect logic to the root of the storage account, so now it fully works!

ecraig12345:
If anyone is curious, this is what the custom 404 does:
```html
<!DOCTYPE html>
<html lang="en">
  <body>
    <script>
      var northstarMatch = location.href.match(/^(.*\/pr-deploy-site\/.*?\/react-northstar)\//);
      if (northstarMatch) {
        // For URLs within a react-northstar PR deploy site, store the actual URL
        sessionStorage.redirect = location.href;
        // and redirect to the root of the site (which will then look at sessionStorage.redirect)
        var meta = document.createElement('meta');
        meta.setAttribute('http-equiv', 'refresh');
        meta.setAttribute('content', "0;URL='" + northstarMatch[1] + "'");
        document.head.appendChild(meta);
      } else {
        document.body.innerHTML = '<h1>The requested content does not exist.</h1>';
      }
    </script>
  </body>
</html>
```

PR #12870, opened 2020-04-24
Factor out an inner version of DetailsList which uses hooks

### LINKED ISSUE #14292: DetailsList: strict mode compliance (UNSAFE_, hard)
DetailsList has a complex `componentWillReceiveProps` implementation which needs to be refactored for strict mode compliance. In this case a function component conversion is likely out of the question due to complexity and risk.

There's also a significant chance that the code contains side effects during render lifecycles, but determining whether this is the case (and whether any of the side effects would actually cause problems due to being non-idempotent) would require manually going through the code. Whoever takes this work item could potentially try to do that, or we can just leave it and fix issues if/when we discover them later.

ThomasMichon:
Started the conversion process of `DetailsList` to use React Hooks, which will provide a better platform for optimizing `DetailsList` and its excessive re-render issues.

For the moment, a `DetailsListInner` component consumes the Props and State from `DetailsList`, along with any internal callbacks and overridden props uses to manage the component instance.
This preserves the complete behavior and caller compatibility of `DetailsList`, while providing initial optimization using `React.useCallback` and `React.useMemo` for event handlers and other render props.

Fixes #14292

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12870)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/DetailsList/DetailsList.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/DetailsList.base.tsx
@@ -21,7 +21,6 @@ import {
   IDetailsListProps,
   IDetailsListStyles,
   IDetailsListStyleProps,
-  IDetailsGroupRenderProps,
   ColumnDragEndLocation,
 } from '../DetailsList/DetailsList.types';
 import { DetailsHeader } from '../DetailsList/DetailsHeader';
@@ -48,6 +47,7 @@ import { CHECK_CELL_WIDTH as CHECKBOX_WIDTH } from './DetailsRowCheck.styles';
 // For every group level there is a GroupSpacer added. Importing this const to have the source value in one place.
 import { SPACER_WIDTH as GROUP_EXPAND_WIDTH } from '../GroupedList/GroupSpacer';
 import { composeRenderFunction } from '@uifabric/utilities';
+import { useConst } from '@uifabric/react-hooks';
 
 const getClassNames = classNamesFunction<IDetailsListStyleProps, IDetailsListStyles>();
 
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/DetailsList/DetailsList.base.tsx
```diff
@@ -71,6 +71,568 @@ const MIN_COLUMN_WIDTH = 100; // this is the global min width
...
+    getRowAriaLabel,
+    getRowAriaDescribedBy,
+    checkButtonAriaLabel,
+    checkboxCellClassName,
+    useReducedRowRenderer,
+    enableUpdateAnimations,
+    enterModalSelectionOnTouch,
+    onRenderRow: defaultOnRenderRow,
+    selectionZoneRef,
+  } = props;
```
dzearing, line 178: yikes, that's a lot of variables.

JustSlone:
@dzearing and @KevinTCoughlin can you take a look at this PR?

ecraig12345:
Very belatedly circling back on this--one of our major goals for version 8 is strict mode compliance, and the UNSAFE_ methods in DetailsList are one of the blockers (see #14292). We were leaning towards getting rid of the unsafe methods without conversion to a function component, but David reminded me that this PR existed.

@ThomasMichon Would you have time to finish this in the next 6 weeks or so, or will someone else (likely @MLoughry or I) need to help? I also wonder if it should be re-targeted to react-next to reduce regression risk within v7.

joschect:
Triggering new build

joschect:
Closing and reopening a final time before push

PR #12872, opened 2020-04-25
Added 13 icons. Modified 2 icons. Renamed 1 icon

TanelVari:
Added icons: TranscriptIcon, FilesGenericColoredIcon, FilesHtmlColoredIcon, FilesPdfColoredIcon, FilesPictureColoredIcon, FilesTextColoredIcon, PopupIcon, ShareGenericIcon, ComposeIcon, TenantWorkIcon, TenantPersonalIcon, WorkOrSchoolIcon, EmailWithDotIcon

Modified: ShareLocationIcon, SpotlightIcon

Renamed: Survey20pxIcon to ApprovalsAppbarIcon

Repo from where the file type icons come from:
https://onedrive.visualstudio.com/DefaultCollection/Design/_git/fabric-cdn?path=%2Fapp-min%2Fassets%2Fitem-types%2F32&version=GBproduction&_a=contents

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12872)


packages/fluentui/react-icons-northstar/src/index.ts
```diff
@@ -86,6 +87,11 @@ export { default as FilesUploadIcon } from './components/FilesUploadIcon';
 export { default as FilesVideoIcon } from './components/FilesVideoIcon';
 export { default as FilesZipIcon } from './components/FilesZipIcon';
 export { default as FlagIcon } from './components/FlagIcon';
+export { default as FluentFileGenericIcon } from './components/FluentFileGenericIcon';
```
mnajdova, line 90: Why did we start creating icons with the fluent prefix?

packages/fluentui/react-icons-northstar/src/index.ts
```diff
@@ -162,11 +169,14 @@ export { default as SendIcon } from './components/SendIcon';
 export { default as SettingsIcon } from './components/SettingsIcon';
 export { default as SettingsAudioIcon } from './components/SettingsAudioIcon';
 export { default as ShareAltIcon } from './components/ShareAltIcon';
+export { default as ShareGenericIcon } from './components/ShareGenericIcon';
 export { default as ShareLocationIcon } from './components/ShareLocationIcon';
+export { default as ShareLocation12pxIcon } from './components/ShareLocation12pxIcon';
```
mnajdova, line 174: Can we change the name a bit and avoid using `12px`?

packages/fluentui/react-icons-northstar/src/index.ts
TanelVari: Because they come from Fluent repo. It was requested that they be added to Webclient, so this means adding here too. Kept the prefix to disticguish them. IS there other way?

packages/fluentui/react-icons-northstar/src/index.ts
TanelVari: Could we please not? I don't think using "small", "smaller", "verysmall" etc names is better

mnajdova: We have `size` prop for that. Why are we not rendering this as a dedicated svg for some size then?

packages/fluentui/react-icons-northstar/src/index.ts
mnajdova: Just worried that if we start like this, we will end up with having half the icons with Fluent prefix.. If there is a good enough reason and we are tracking where the icons comes from then ok..

packages/fluentui/react-icons-northstar/src/index.ts
TanelVari: Because it is designed explicitly to be used at 12px.

packages/fluentui/react-icons-northstar/src/index.ts
TanelVari: I'll ask around a little.

packages/fluentui/react-icons-northstar/src/index.ts
mnajdova: My point was that `<ShareLocationIcon size="small" />` for example can render that svg. I don't like `12px` in the name, because `<ShareLocation12pxIcon size="large" />` will definitelly not be 12px..

TanelVari: But it's not meant to be used as "large". We already have icons that have this in the name, how else a dev or anyone else should know that this is very-very specific icon. Besides - I was asked to modify existing icon which already has this kind of name. Included in this PR

packages/fluentui/react-icons-northstar/src/index.ts
codepretty: I agree with @mnajdova. The icons might come from Fluent, but I don't see a reason to make that distinction here. Here we just have icons that Teams uses, unless there is some other distinction I'm missing...

packages/fluentui/react-icons-northstar/src/components/Survey20pxIcon.tsx
```diff
@@ -7,14 +7,14 @@ import { iconClassNames } from '../utils/iconClassNames';
 const Survey20pxIcon = createSvgIcon({
```
mnajdova, line 7: Please remove this icon, as it is not exported. If it should be consider changing the name again not to have `px` in the name

packages/fluentui/react-icons-northstar/src/index.ts
```diff
@@ -54,11 +54,13 @@ export { default as CloseIcon } from './components/CloseIcon';
 export { default as CodeSnippetIcon } from './components/CodeSnippetIcon';
 export { default as CollapseIcon } from './components/CollapseIcon';
 export { default as CompanionIcon } from './components/CompanionIcon';
+export { default as ComposeIcon } from './components/ComposeIcon';
```
mnajdova, line 57: Some of the icons are not exported, please check again

packages/fluentui/CHANGELOG.md
```diff
@@ -105,6 +106,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `slots` and `mapPropsToSlotProps` in compose and added example usage in the `Button` component @mnajdova ([#12858](https://github.com/microsoft/fluentui/pull/12858))
 - Add slots support in `Attachment` @mnajdova ([#12913](https://github.com/microsoft/fluentui/pull/12913))
 - Use `compose()` in `ToolbarDivider` @mnajdova ([#13045](https://github.com/microsoft/fluentui/pull/13045))
+- Added icons: `TranscriptIcon`, `FilesGenericColoredIcon`, `FilesHtmlColoredIcon`, `FilesPdfColoredIcon`, `FilesPictureColoredIcon`, `FilesTextColoredIcon`, `PopupIcon`, `ShareGenericIcon`, `ComposeIcon`, `SpotlightStopIcon`, `TenantWorkIcon`, `TenantPersonalIcon`, `WorkOrSchoolIcon`, `EmailWithDotIcon`. Modified: `ShareLocationIcon`, `SpotlightIcon`. @TanelVari ([#12872](https://github.com/microsoft/fluentui/pull/12872))
```
mnajdova, line 109: Oops move this to the unreleased section.

PR #12874, opened 2020-04-25
chore(AccordionTitle): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `AccordionTitle` component to be functional. Restricting props set that will be passed to styles functions.

Related to #12237

## Prop sets

| `AccordionTitle`    |
| --------- |
| `disabled` |
| `content` |
| `active` |

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12874)

### DIFF
--- a/packages/fluentui/accessibility/src/behaviors/Accordion/accordionTitleBehavior.ts
+++ b/packages/fluentui/accessibility/src/behaviors/Accordion/accordionTitleBehavior.ts
@@ -43,7 +43,7 @@ const accordionTitleBehavior: Accessibility<AccordionTitleBehaviorProps> = props
 
 export default accordionTitleBehavior;
 
-type AccordionTitleBehaviorProps = {
+export type AccordionTitleBehaviorProps = {
   /** Element type. */
   as?: string;
   /** Whether or not the title is in the open state. */
--- a/packages/fluentui/react-northstar/src/components/Accordion/AccordionTitle.tsx
+++ b/packages/fluentui/react-northstar/src/components/Accordion/AccordionTitle.tsx
@@ -1,4 +1,4 @@
-import { accordionTitleBehavior, indicatorBehavior } from '@fluentui/accessibility';
+import { accordionTitleBehavior, Accessibility, AccordionTitleBehaviorProps } from '@fluentui/accessibility';
 import { Ref } from '@fluentui/react-component-ref';
 import * as customPropTypes from '@fluentui/react-proptypes';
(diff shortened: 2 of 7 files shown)


packages/fluentui/react-northstar/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -74,113 +85,166 @@ export const accordionTitleSlotClassNames: AccordionTitleSlotClassNames = {
...
+      accordionContentId,
+    }),
+    rtl: context.rtl,
+  });
+
+  const { classes, styles: resolvedStyles } = useStyles<AccordionTitleStylesProps>(AccordionTitle.displayName, {
+    className: accordionTitleClassName,
+    mapPropsToStyles: () => ({
+      disabled,
+      content: !!content,
```
layershifter, line 141:
```suggestion
      hasContent: !!content,
```

packages/fluentui/react-northstar/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -74,113 +85,166 @@ export const accordionTitleSlotClassNames: AccordionTitleSlotClassNames = {
...
+    actionHandlers: {
+      performClick: e => {
+        e.preventDefault();
+        e.stopPropagation();
+        handleClick(e);
+      },
     },
-  };
+    mapPropsToBehavior: () => ({
+      content: true,
```
layershifter, line 127:
```suggestion
      hasContent: !!content,
```

??

packages/fluentui/accessibility/src/behaviors/Accordion/accordionTitleBehavior.ts
```diff
@@ -43,7 +43,7 @@ const accordionTitleBehavior: Accessibility<AccordionTitleBehaviorProps> = props
 
 export default accordionTitleBehavior;
 
-type AccordionTitleBehaviorProps = {
+export type AccordionTitleBehaviorProps = {
   /** Element type. */
   as?: string;
```
pompomon, line 48: Please could you consider not sending `as` prop to the behavior, but instead have `isHeading` prop, which would have the logic on the component?

PR #12878, opened 2020-04-27
chore(Accessibility): Prevents already defined aa-class to be overwritten

assuncaocharles:
#### Pull request checklist

#### Description of changes

Currently on `Button` component page we have the following accessibility error:

```
Inconsistent value of parameter 'disabled': 'false' != 'true' in class 'Button', $('[data-aa-error-id="1"]')
```
It happens because `data-aa-class` get overwritten. This PR prevents that to happen

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12878)

### DIFF
--- a/packages/fluentui/react-bindings/src/accessibility/getAccessibility.ts
+++ b/packages/fluentui/react-bindings/src/accessibility/getAccessibility.ts
@@ -36,7 +36,9 @@ const getAccessibility = <Props extends Record<string, any>>(
     // schema is located in @stardust-ui/ability-attributes package.
     if (definition.attributes) {
       Object.keys(definition.attributes).forEach(slotName => {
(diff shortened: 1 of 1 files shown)


packages/fluentui/react-bindings/src/accessibility/getAccessibility.ts
```diff
@@ -36,7 +36,9 @@ const getAccessibility = <Props extends Record<string, any>>(
     // schema is located in @stardust-ui/ability-attributes package.
```
hoshinetop, line 36: As we're touching the place anyway, makes sense to rename that `@stardust-ui/ability-attributes` leftover to `@fluentui/ability-attributes`.

PR #12884, opened 2020-04-27
docs(Toolbar): accessibility polish - testing preparation

kolaps33:
#### Description of changes
In order to give Toolbar component for testing, we should be consistent with ARIA.
https://www.w3.org/TR/wai-aria-practices-1.2/#toolbar
Apply following:
- Use toolbar as a grouping element only if the group contains 3 or more controls.

and from role "Toolbar" itself 
https://www.w3.org/TR/wai-aria-1.2/#toolbar
Apply following:
- Authors MUST supply a label on each toolbar when the application contains more than one toolbar.

In order to fulfill these requirements above:
- adding label for each Toolbar on the page
- adding more items to the Toolbar at least Toolbar contains 3 items 
#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12884)


packages/fluentui/docs/src/examples/components/Toolbar/Content/ToolbarExampleMenuItemToggle.shorthand.tsx
```diff
@@ -15,10 +15,21 @@ const ToolbarExampleMenuItemToggle = () => {
     }
   };
 
+  const toolbarItem = (iconName, icon) => {
```
jurokapsiar, line 18: nit: I would call it 'itemName' instead of `iconName`

packages/fluentui/docs/src/examples/components/Toolbar/Content/ToolbarExampleMenu.shorthand.tsx
```diff
@@ -14,10 +14,21 @@ const ToolbarExampleMenuShorthand = () => {
     createCallbackLogFormatter(['menuOpen']),
   );
 
+  const toolbarItem = (itemName: string, itemIcon: ShorthandValue<BoxProps>) => {
```
silviuaavram, line 17: not: => ({key: itemName ... })

packages/fluentui/docs/src/examples/components/Toolbar/Types/index.tsx
```diff
@@ -6,6 +6,7 @@ const Types = () => (
   <ExampleSection title="Types">
     <ComponentExample
       title="Default"
+      toolbarAriaLabel="Example default"
```
silviuaavram, line 9: we can probably do a little bit better with this aria label and the ones before. `Default` ... does not give much information. `default toolbar example` is probably better. I would go through all the `aria-labels` and make some improvements.

packages/fluentui/docs/src/examples/components/Toolbar/Usage/ToolbarExampleActionPopupInMenu.shorthand.tsx
```diff
@@ -27,12 +27,23 @@ const CustomPopup = ({ onConfirm }) => {
   );
 };
 
+const toolbarItem = (itemName: string, itemIcon: ShorthandValue<BoxProps>) => {
```
silviuaavram, line 30: I've seen this function about 4 times, we can probably create a commonUtils file in `Toolbar` examples folder and put it there, and re-use.

kolaps33: I fully agree and this was my point even when I did changes some time ago, but if I remember correctly someone(not sure if Shift or Marija) was commenting that I should not do it in the reusable way because when we pop-up the example then it will be not available there.
@mnajdova or @miroslavstastny please what do you think can we somehow extract this code to reusable file? please for your recommendation? thank you

packages/fluentui/docs/src/examples/components/Toolbar/Types/index.tsx
kolaps33: We have already on the Toolbar component 31 toolbars and the idea recently was following:
image

Each toolbar contains the buttons ("Try it, Theme it, RTL, Popout") related to the example will start to with the example wording and following the "example name." To be able distinguish if you cannot see page to which toolbar is "example toolbar" related. 

It is not ideal to have in the toolbar name the wording "Toolbar" because then the narration is following:
"{toolbar aria label} toolbar " and if we already keep word 'Toolbar' in aria-label then it narrate 'Toolbar' twice. 

But unfortunately in some cases if the name of the example is "Toolbar can contain a menu" I rather let it there because the when people will do testing I hope they will find it easier the toolbar because title of toolbar example is as well the heading. 

But I am open any naming improvement here :) just only we need to consider that we have 
- toolbar contains the buttons ("Try it, Theme it, RTL, Popout") 
- toolbar which is as example itself 
And this toolbars are following in the DOM one after another...

packages/fluentui/docs/src/examples/components/Toolbar/Usage/ToolbarExampleActionPopupInMenu.shorthand.tsx
mnajdova: The examples should serve as an example for the clients on how to do things, we don't want to have there partial code and some utilities that they cannot access

kolaps33: Thank you @mnajdova for clarification.

packages/fluentui/docs/src/examples/components/Toolbar/Content/ToolbarExampleMenu.shorthand.tsx
kolaps33: good proposal :)

packages/fluentui/docs/src/components/ComponentDoc/ComponentControls/ComponentControls.tsx
```diff
@@ -33,7 +33,7 @@ type ComponentControlsProps = {
   showRtl: boolean;
   showVariables: boolean;
   showTransparent: boolean;
-  toolbarAriaLabel?: string;
+  title?: React.ReactNode;
```
jurokapsiar, line 36: do we somewhere use something else than `string` for `title`?

packages/fluentui/docs/src/components/ComponentDoc/ComponentControls/ComponentControls.tsx
```diff
@@ -78,7 +78,7 @@ const ComponentControls: React.FC<ComponentControlsProps> = props => {
             {...rest}
             iconOnly
             accessibility={menuAsToolbarBehavior}
-            aria-label={toolbarAriaLabel || null}
+            aria-label={`Actions for ${title ? title : 'untitled'} example`}
```
jurokapsiar, line 81: Actions for Object [Object] example :(

kolaps33: fixed :) changing title to string :)

packages/fluentui/docs/src/examples/components/Toolbar/Content/ToolbarExampleMenuRadioGroup.shorthand.tsx
```diff
@@ -1,15 +1,24 @@
...
+    content: itemName,
+    icon: itemIcon,
+    title: itemName,
+  });
+
   return (
     <Toolbar
       aria-label="Toolbar can contain a radio group in a menu"
       items={[
+        toolbarItem('underline', <UnderlineIcon />),
```
mnajdova, line 20: nit: To be honest, I would just spread objects here. From client's perspective it's much easier to just go trough the items object, instead of seeing what the function that is invoked is doing...

mnajdova: Would be much more readable :)

packages/fluentui/docs/src/examples/components/Toolbar/Content/ToolbarExamplePopup.shorthand.tsx
```diff
@@ -65,6 +65,11 @@ const ToolbarExamplePopupShorthand = () => {
             },
           },
         },
+        {
+          icon: <ItalicIcon {...{ outline: true }} />,
```
mnajdova, line 69:
```suggestion
          icon: <ItalicIcon outline />,
```

packages/fluentui/docs/src/examples/components/Toolbar/Content/index.tsx
```diff
@@ -19,36 +18,32 @@ const Content = () => (
...
     />
     <ComponentExample
       title="Toolbar can contain a radio group in a menu"
-      toolbarAriaLabel="Example Toolbar can contain a radio group in a menu"
       description="Toolbar item can open a menu which can contain radio groups."
       examplePath="components/Toolbar/Content/ToolbarExampleMenuRadioGroup"
     />
     <ComponentExample
       title="Toolbar can contain a submenu in a menu"
+      toolbarAriaLabel="Example Toolbar can contain a submenu in a menu"
```
mnajdova, line 41: Shuoldn't this be removed?

packages/fluentui/docs/src/examples/components/Toolbar/Performance/index.tsx
```diff
@@ -7,6 +7,7 @@ const Performance = () => (
   <ExampleSection title="Performance">
     <ComponentPerfExample
       title="Custom styled"
+      toolbarAriaLabel="Example performance"
```
mnajdova, line 10: Do we still need this?

packages/fluentui/docs/src/examples/components/Toolbar/Content/index.tsx
kolaps33: good catch thank you :)

packages/fluentui/docs/src/examples/components/Toolbar/Content/ToolbarExampleMenuRadioGroup.shorthand.tsx
kolaps33: changed :)

packages/fluentui/docs/src/components/ComponentDoc/ComponentControls/ComponentControls.tsx
kolaps33: changed to string :)

packages/fluentui/docs/src/examples/components/Toolbar/Performance/index.tsx
kolaps33: good catch thank you :)

PR #12887, opened 2020-04-27
Adding @fluentui/react-stylesheets

dzearing:
This package exports a `StylesheetProvider` component and `useStylesheet` hook as a mechanism for registering styles for a component when used.

We need these for a few scenarios:

1. Styles can be registered on render rather than on import. (Note that load order is importance, hence queuing up stylesheets and rendering them on useLayoutEffect.)
2. Registration of styles can use an alternative target such as a child window.
3. Styles can also be registered with an alternative implementation (SSR.)

Example usage:

1. Write a component which uses `useStylesheet` (note that styles will be imported):

```tsx
const styles = `
  .root { background: red; }
`;

const Button = (props) => {
  useStylesheet(styles);

  return <button  {...props} className={cx(props.className, "root")} />;
};
```

2. Use the component:

Without a `StylesheetProvider`, the default behavior would register each style when encountered:

```tsx
<Button>Hello world!</Button> // registers styles on first render
```

With a `StylesheetProvider`, all styles are enqueued and rendered to the `head` in order of leaf to root in a single `style` element:

```tsx
<StylesheetProvider>
  <Button>Hello world!</Button>
</StylesheetProvider>
```

3. Target a new child window with the styles:

```tsx
<StylesheetProvider target={childDocument}>
  <Button>Hello world!</Button>
</StylesheetProvider>
```
4. Collect the styles yourself for SSR:

```tsx
const styles = [];
<StylesheetProvider renderStyles={ sheets => styles.push(...sheets) }>
  ...
</StylesheetProvider>
```

Notes:

Right now this uses a custom `renderStyles` function. I'm likely going to replace this with `load-themed-styles` API soon, but this will happen separately.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12887)


packages/react-stylesheets/src/StylesheetProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+ */
+export const StylesheetProvider = (props: React.PropsWithChildren<StylesheetProviderProps>) => {
+  const context = React.useContext(StylesheetContext);
+  const mergedContext = React.useMemo<StylesheetContextType>(
+    () => ({
+      // tslint:disable-next-line:no-any
+      ...(context as any),
+      ...props,
+    }),
+    [context],
```
xugao, line 18: need to check `props` too
i think here should be: `[...context, ...props]`

packages/react-stylesheets/src/StylesheetContext.tsx
```diff
@@ -0,0 +1,66 @@
...
+  // If there is no target, call renderSheets immediately.
+  if (!target) {
+    renderSheets(context.enqueuedSheets, context);
+  }
+};
+
+const renderSheets = (sheets: string[], context: StylesheetContextType) => {
+  const { target, enqueuedSheets } = context;
+
+  if (enqueuedSheets.length && target) {
```
xugao, line 50: should check `sheets.length` instead

packages/react-stylesheets/src/StylesheetContext.tsx
```diff
@@ -0,0 +1,66 @@
...
+    if (!sheets.has(stylesheet)) {
+      context.enqueuedSheets.unshift(stylesheet);
+      sheets.set(stylesheet, true);
+    }
+  }
+
+  // If there is no target, call renderSheets immediately.
+  if (!target) {
+    renderSheets(context.enqueuedSheets, context);
+  }
```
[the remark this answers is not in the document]
dzearing: With server rendering, `renderSheets` will never get called if the primary entrypoint is through `useLayoutEffect`.

How would you expect the syntax to work for SSR scenarios? Here's an alternative `server` boolean we could apply, rather than a null `target` value.

```tsx
<StylesheetProvider server renderStyles={...}>...</StylesheetProvider>
  ...

File: packages/react-stylesheets/README.md [hunk 4]
```diff
@@ -0,0 +1,105 @@
...
+  `;
+}
+```
+
+## Edge cases covered
+
+### Order of stylesheet registration
+
+Typical webpack `style-loader` style registration approach would dictate that styles would be
+registered in module import order. This means leaf node end up registering styles first, then root elements.
```
khmakoto, line 99:
```suggestion
registered in module import order. This means leaf nodes end up registering styles first, then root elements.
```

packages/react-stylesheets/etc/react-focus.api.md
```diff
@@ -0,0 +1,91 @@
+## API Report File for "@fluentui/react-focus"
```
khmakoto, line 1: I think there might have been a bad copy of this file because I don't think this should live here.

packages/react-stylesheets/src/StylesheetContext.tsx
```diff
@@ -0,0 +1,66 @@
...
+  renderSheets: (stylesheets: string[], context: StylesheetContextType) => void;
+  // hasRegistered: (stylehseet: string, context: StylesheetContextType) => boolean;
+}
+
+// Default registration function for stylesheets.
+export const register = (stylesheets: string[], context: StylesheetContextType) => {
+  const { styleCache, target, dir } = context;
+
+  // Grab the style cache for the target document.
+  let targetDirections = styleCache.get(target!);
```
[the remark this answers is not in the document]
khmakoto: A couple of naming options: `targetStylesCache`, `targetStyles`, `targetCache`.

packages/react-stylesheets/src/StylesheetContext.tsx
khmakoto: Actually, I think `enqueuedSheets` can be removed altogether from this function.

packages/react-stylesheets/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "strictNullChecks": true,
+    "noImplicitAny": true,
+    "moduleResolution": "node",
+    "preserveConstEnums": true,
+    "lib": ["es2015", "dom", "es2015.promise"],
+    "typeRoots": ["../../node_modules/@types", "../../typings"],
+    "types": ["jest", "webpack-env", "custom-global"],
+    "paths": {
+      "@fluentui/react-focus/lib/*": ["./src/*"],
+      "@fluentui/react-focus": ["./src"]
```
khmakoto, line 23:
```suggestion
      "@fluentui/react-stylesheets/lib/*": ["./src/*"],
      "@fluentui/react-stylesheets": ["./src"]
```

packages/react-stylesheets/webpack.config.js
```diff
@@ -0,0 +1,29 @@
...
+const IS_PRODUCTION = process.argv.indexOf('--production') > -1;
+
+module.exports = resources.createConfig(BUNDLE_NAME, IS_PRODUCTION, {
+  entry: {
+    [BUNDLE_NAME]: './lib/index.js',
+  },
+
+  output: {
+    libraryTarget: 'var',
+    library: 'FluentUIReactFocus',
```
khmakoto, line 14:
```suggestion
    library: 'FluentUIReactStylesheets',
```

packages/react-stylesheets/webpack.config.js
```diff
@@ -0,0 +1,29 @@
+const path = require('path');
+const resources = require('@uifabric/build/webpack/webpack-resources');
+
+const BUNDLE_NAME = 'react-focus';
```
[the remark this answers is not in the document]
khmakoto:
```suggestion
const BUNDLE_NAME = 'react-stylesheets';
```

packages/react-stylesheets/webpack.config.js
```diff
@@ -0,0 +1,29 @@
...
+  externals: {
+    react: 'React',
+    'react-dom': 'ReactDOM',
+  },
+
+  resolve: {
+    alias: {
+      '@fluentui/react-focus/src': path.join(__dirname, 'src'),
+      '@fluentui/react-focus/lib': path.join(__dirname, 'lib'),
+      '@fluentui/react-focus': path.join(__dirname, 'lib'),
```
khmakoto, line 26:
```suggestion
      '@fluentui/react-stylesheets/src': path.join(__dirname, 'src'),
      '@fluentui/react-stylesheets/lib': path.join(__dirname, 'lib'),
      '@fluentui/react-stylesheets': path.join(__dirname, 'lib'),
```

packages/react-stylesheets/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "react": "16.8.6",
+    "react-app-polyfill": "~1.0.1",
+    "react-dom": "16.8.6",
+    "react-test-renderer": "^16.3.0"
+  },
+  "dependencies": {
+    "@uifabric/merge-styles": "^7.11.1",
+    "@uifabric/set-version": "^7.0.11",
+    "@uifabric/styling": "^7.11.8",
+    "@uifabric/utilities": "^7.15.11",
```
khmakoto, line 53: I don't think you need `merge-styles`, `styling` or `utilities` dependencies here, but please correct me if I'm wrong.

packages/react-stylesheets/src/StylesheetContext.tsx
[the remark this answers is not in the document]
dzearing: `targetStylesheets`. I will get rid of the `dir` caching here because I don't think it's necessary just yet.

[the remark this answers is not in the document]
dzearing: Done.

packages/react-stylesheets/src/StylesheetContext.tsx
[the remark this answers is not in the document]
xugao: hm ic. maybe something like:
```
  const isSSR = !target;
  if (isSSR) {
    renderSheets(context.enqueuedSheets, context);
  }
```
instead of
```
  // If there is no target, call renderSheets immediately.
  if (!target) {
    renderSheets(context.enqueuedSheets, context);
  }
```

packages/react-stylesheets/src/StylesheetContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+  let targetStylesheets = styleCache.get(target!);
+
+  if (!targetStylesheets) {
+    targetStylesheets = new Map();
+    styleCache.set(target!, targetStylesheets);
+  }
+
+  for (const stylesheet of stylesheets) {
+    if (!targetStylesheets.has(stylesheet)) {
+      context.enqueuedSheets.unshift(stylesheet);
```
xugao, line 28: reminder on Levi's point earlier today - we can't guarantee the correct order based on rendering order?

packages/react-stylesheets/src/StylesheetProvider.tsx
dzearing: I want the context object to be the key only; you can't do `...props` in an array without first converting to an array using Object.values, which is way less efficient than it should be. I'd rather assume props won't change for the perf benefit of predictably avoiding mutations.

dzearing: i added a comment to clarify though.

packages/react-stylesheets/src/StylesheetContext.tsx
dzearing: done!

packages/react-stylesheets/package.json
dzearing: already cleaned up.

packages/react-stylesheets/src/StylesheetContext.tsx
dzearing: No we can't guarantee, but the same applies to the typical style injection on module import. All it takes is an async import to change the order.

What we can do is avoid flipping order completely in the normal case. Typically an async import is not happening on a sub component but rather on an experience boundary.

dzearing: I am going to think about if this is really necessary. Right now I think it is. But maybe we just need to attack the problem at the styling rule layer.

packages/react-stylesheets/src/StylesheetContext.tsx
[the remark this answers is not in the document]
dzearing: makes sense!

packages/react-stylesheets/README.md
```diff
@@ -0,0 +1,105 @@
...
+### Order of stylesheet registration
+
+Typical webpack `style-loader` style registration approach would dictate that styles would be
+registered in module import order. This means leaf nodes end up registering styles first, then root elements.
+
+Because React renders components root to child, stylesheet ordering gets reversed.
+
+### nonce
+
+TBD
```
khmakoto, line 105: What is this about?

packages/react-stylesheets/src/version.ts
```diff
@@ -0,0 +1,4 @@
+// @fluentui/react-focus@0.1.0
+// Do not modify this file, the file is generated as part of publish. The checked in version is a placeholder only.
+import { setVersion } from '@uifabric/set-version';
+setVersion('@fluentui/react-focus', '0.1.0');
```
xugao, line 4:
```suggestion
setVersion('@fluentui/react-stylesheets', '0.1.0');
```

packages/react-stylesheets/src/version.ts
```diff
@@ -0,0 +1,4 @@
+// @fluentui/react-focus@0.1.0
```
xugao, line 1:
```suggestion
// @fluentui/react-stylesheets@0.1.0
```

packages/react-stylesheets/README.md
dzearing: Once I merge load-themed-styles integration here, it's a security thing. You basically need to add a nonce string to dynamically injected style attributes. I'll remove this title but add it back later once it's supported.

See this:

 https://www.npmjs.com/package/@microsoft/load-themed-styles#security-considerations

PR #12890, opened 2020-04-27
Updating examples for ProgressIndicator, FloatingPeoplePicker and ResizeGroupOverflowSet to use FC

czearing:
#### Pull request checklist

#### Description of changes

Updating examples for `ProgressIndicator`, `FloatingPeoplePicker` and `ResizeGroupOverflowSet` to use function components.

#### Focus areas to test
ProgressIndicator, FloatingPeoplePicker and ResizeGroupOverflowSet

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12890)


packages/office-ui-fabric-react/src/components/FloatingPicker/PeoplePicker/examples/FloatingPeoplePicker.Basic.Example.tsx
```diff
@@ -10,128 +9,95 @@ import {
...
-          onRemoveSuggestion={this._onRemoveSuggestion}
-          onValidateInput={this._validateInput}
-          componentRef={this._picker}
-          onChange={this._onPickerChange}
-          inputElement={this._inputElement}
-          resolveDelay={300}
-        />
-      </div>
-    );
+let inputElement: HTMLInputElement;
```
ecraig12345, line 12: This should go inside the component body, not here (and move other things inside the component body as needed). Otherwise different instances of the component would interfere with each other and the element won't get garbage collected when the component unmounts.

packages/office-ui-fabric-react/src/components/FloatingPicker/PeoplePicker/examples/FloatingPeoplePicker.Basic.Example.tsx
```diff
@@ -10,128 +9,95 @@ import {
...
+const suggestionsStore = new SuggestionsStore<IPersonaProps>();
```
ecraig12345, line 13: Empty lines between some of these groups of lines would be good too (certainly after functions).

Same in other examples.

packages/office-ui-fabric-react/src/components/ProgressIndicator/examples/ProgressIndicator.Basic.Example.tsx
```diff
@@ -1,65 +1,22 @@
...
-  constructor(props: {}) {
-    super(props);
-
-    this._async = new Async(this);
-
-    this.state = {
-      percentComplete: 0,
+  React.useEffect(() => {
+    const id = setInterval(() => {
+      setPercentComplete((intervalIncrement + percentComplete) % 1);
```
ecraig12345, line 12: You should add back logic to stop the interval after it reaches 100% :)

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.OverflowSet.Example.tsx
```diff
@@ -41,152 +48,104 @@ function generateData(count: number, cachingEnabled: boolean, checked: boolean):
...
-            />
-          </div>
-        </div>
-      </div>
-    );
-  }
-
-  private _onReduceData = (currentData: any): any => {
+};
+const short = false;
```
ecraig12345, line 66: Remove this and just use false directly wherever needed?

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.OverflowSet.Example.tsx
```diff
@@ -41,152 +48,104 @@ function generateData(count: number, cachingEnabled: boolean, checked: boolean):
...
+);
+
+export const ResizeGroupOverflowSetExample: React.FunctionComponent = () => {
+  const [numberOfItems, setNumberOfItems] = React.useState(20);
+  const [buttonsChecked, { toggle: toggleButtonsChecked }] = useBoolean(false);
+  const [cachingEnabled, { toggle: toggleCachingEnabled }] = useBoolean(false);
+  const [onGrowDataEnabled, { toggle: toggleOnGrowDataEnabled }] = useBoolean(false);
+  const dataToRender = generateData(numberOfItems, cachingEnabled, buttonsChecked);
+
+  const onReduceData = (currentData: any): any => {
```
ecraig12345, line 84: Add a type here?

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.VerticalOverflowSet.Example.tsx
```diff
@@ -16,35 +16,28 @@ export interface IOverflowData {
...
 }
 
-function generateData(count: number, cachingEnabled: boolean, checked: boolean): IOverflowData {
+const generateData = (count: number, cachingEnabled: boolean, checked: boolean): IOverflowData => {
   const icons = ['Add', 'Share', 'Upload'];
-  const dataItems = [];
   let cacheKey = '';
-  for (let index = 0; index < count; index++) {
-    const item = {
+  const dataItems = Array.from({ length: count }).map((item, index) => {
```
ecraig12345, line 22: `Array.from` doesn't work in IE 11, so you need to go back to the previous approach.

packages/office-ui-fabric-react/src/components/FloatingPicker/PeoplePicker/examples/FloatingPeoplePicker.Basic.Example.tsx
dzearing: I suggested to him to useRef here.

packages/office-ui-fabric-react/src/components/ProgressIndicator/examples/ProgressIndicator.Basic.Example.tsx
dzearing: I worked through this with him and had suggested that simplification. Do you think it's very important to keep?

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.OverflowSet.Example.tsx
dzearing:
```tsx
private _onReduceData = (currentData: any): any => {
```

Not arguing against the suggestion, it's a good one, but the whole ResizeGroup typings are full of anys, and the code here is just a port of what was there.

```tsx

  /**
   * Initial data to be passed to the `onRenderData` function. When there is no `onGrowData` provided, this data should
   * represent what should be passed to the render function when the parent container of the ResizeGroup is at its
   * maximum supported width. A `cacheKey` property may optionally be included as part of the data. Two data objects
   * with the same `cacheKey` will be assumed to take up the same width and will prevent measurements.
   * The type of `cacheKey` is a string.
   */
  data: any;

  /**
   * Function to render the data. Called when rendering the contents to the screen and when
   * rendering in a hidden div to measure the size of the contents.
   */
  onRenderData: (data: any) => JSX.Element;

  /**
   * Function to be performed on the data in order to reduce its width and make it fit into the given space.
   * If there are no more scaling steps to apply, it should return undefined to prevent
   * an infinite render loop.
   */
  onReduceData: (prevData: any) => any;
```

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.VerticalOverflowSet.Example.tsx
dzearing: suggested to use `createArray` which is safer and a bit more simpler anyway; he'll find/replace other references too because he mentioned he had added that in some other places.

packages/office-ui-fabric-react/src/components/ProgressIndicator/examples/ProgressIndicator.Basic.Example.tsx
ecraig12345: Progress over 100% might be ignored, but I still think it's cleaner to detect and stop when it's finished.

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.VerticalOverflowSet.Example.tsx
ecraig12345: Huh I forgot that existed

packages/office-ui-fabric-react/src/components/FloatingPicker/PeoplePicker/examples/FloatingPeoplePicker.Basic.Example.tsx
```diff
@@ -10,128 +9,96 @@ import {
...
     if (itemIndex >= 0) {
-      this.setState({
-        peopleList: peopleList.slice(0, itemIndex).concat(peopleList.slice(itemIndex + 1)),
-      });
+      setPeopleList(peopleList.slice(0, itemIndex).concat(peopleList.slice(itemIndex + 1)));
     }
   };
 
-  private _onFilterChanged = (filterText: string, currentPersonas?: IPersonaProps[]): IPersonaProps[] => {
+  const suggestionProps: IBaseFloatingPickerSuggestionProps = {
```
ecraig12345, line 61: This should probably go inside a `useConst` to avoid creating a new object every time (it can be a const because the `picker` ref object itself won't change, even if the thing inside does)

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.OverflowSet.Example.tsx
```diff
@@ -41,152 +50,111 @@ function generateData(count: number, cachingEnabled: boolean, checked: boolean):
...
+    setNumberOfItems(parseInt(option.text, 10));
   };
 
-  private _onNumberOfItemsChanged = (event: React.FormEvent<HTMLDivElement>, option: IDropdownOption): void => {
-    this.setState({ numberOfItems: parseInt(option.text, 10) });
-  };
-}
+  return (
+    // tslint:disable-next-line:no-constant-condition
+    <div className={false ? styles.resizeIsShort : 'notResized'}>
```
ecraig12345, line 136: Probably best to entirely remove the className here (also remove `styles.resizeIsShort` if it's unused elsewhere)

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.OverflowSet.Example.tsx
```diff
@@ -41,152 +50,111 @@ function generateData(count: number, cachingEnabled: boolean, checked: boolean):
...
+        role="tabpanel"
+        aria-label="Resize Group with an Overflow Set"
+        data={dataToRender}
+        onReduceData={onReduceData}
+        onGrowData={onGrowDataEnabled ? onGrowData : undefined}
+        onRenderData={onRenderData}
+      />
+      <div className={styles.settingsGroup}>
+        <Checkbox label="Enable caching" onChange={toggleCachingEnabled} checked={cachingEnabled} />
+        <Checkbox label="Set onGrowData" onChange={toggleOnGrowDataEnabled} checked={onGrowDataEnabled} />
```
ecraig12345, line 147:
```suggestion
        <Checkbox label="Enable onGrowData" onChange={toggleOnGrowDataEnabled} checked={onGrowDataEnabled} />
```

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.OverflowSet.Example.tsx
```diff
@@ -41,152 +50,111 @@ function generateData(count: number, cachingEnabled: boolean, checked: boolean):
...
+      <ResizeGroup
+        role="tabpanel"
+        aria-label="Resize Group with an Overflow Set"
+        data={dataToRender}
+        onReduceData={onReduceData}
+        onGrowData={onGrowDataEnabled ? onGrowData : undefined}
+        onRenderData={onRenderData}
+      />
+      <div className={styles.settingsGroup}>
+        <Checkbox label="Enable caching" onChange={toggleCachingEnabled} checked={cachingEnabled} />
```
ecraig12345, line 146: These should probably all be toggles not checkboxes--toggles are more appropriate when you have on/off settings which take effect instantly. Something to watch for in other examples too.

change/@fluentui-keyboard-key-2020-05-08-15-52-01-fix-example13-update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
```
ecraig12345, line 2: You can delete this now (the changed file got removed after merge)

packages/office-ui-fabric-react/src/components/FloatingPicker/PeoplePicker/examples/FloatingPeoplePicker.Basic.Example.tsx
```diff
@@ -9,129 +8,100 @@ import {
 } from 'office-ui-fabric-react/lib/FloatingPicker';
 import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox';
 import { people } from '@uifabric/example-data';
+import { useConst } from '@uifabric/react-hooks';
+
+const suggestionsStore = new SuggestionsStore<IPersonaProps>();
```
ecraig12345, line 13: This needs to be in a useConst inside the component, so you're not sharing state between instances.
`const suggestionsStore = useConst(() => new SuggestionsStore<IPersonaProps>());`

packages/office-ui-fabric-react/src/components/FloatingPicker/PeoplePicker/examples/FloatingPeoplePicker.Basic.Example.tsx
```diff
@@ -9,129 +8,100 @@ import {
...
+const searchBoxWrapperStyling = { width: 208 };
+
+const getTextFromItem = (persona: IPersonaProps): string => {
+  return persona.text || '';
+};
+const listContainsPersona = (persona: IPersonaProps, personas?: IPersonaProps[]): boolean => {
```
ecraig12345, line 19:
```suggestion

const listContainsPersona = (persona: IPersonaProps, personas?: IPersonaProps[]): boolean => {
```

packages/office-ui-fabric-react/src/components/FloatingPicker/PeoplePicker/examples/FloatingPeoplePicker.Basic.Example.tsx
```diff
@@ -9,129 +8,100 @@ import {
...
+  return !!personas && personas.some((item: IPersonaProps) => item.text === persona.text);
+};
+const validateInput = (input: string): boolean => {
```
ecraig12345, line 22:
```suggestion

const validateInput = (input: string): boolean => {
```

packages/office-ui-fabric-react/src/components/FloatingPicker/PeoplePicker/examples/FloatingPeoplePicker.Basic.Example.tsx
```diff
@@ -9,129 +8,100 @@ import {
...
+  return input.indexOf('@') !== -1;
+};
+const startsWith = (text: string, filterText: string): boolean => {
```
ecraig12345, line 25:
```suggestion

const startsWith = (text: string, filterText: string): boolean => {
```

packages/office-ui-fabric-react/src/components/ResizeGroup/examples/ResizeGroup.OverflowSet.Example.tsx
```diff
@@ -41,152 +47,111 @@ function generateData(count: number, cachingEnabled: boolean, checked: boolean):
...
+  const onNumberOfItemsChanged = (event: React.FormEvent<HTMLDivElement>, option: IDropdownOption): void => {
+    setNumberOfItems(parseInt(option.text, 10));
   };
 
-  private _onNumberOfItemsChanged = (event: React.FormEvent<HTMLDivElement>, option: IDropdownOption): void => {
-    this.setState({ numberOfItems: parseInt(option.text, 10) });
-  };
-}
+  return (
+    // tslint:disable-next-line:no-constant-condition
```
ecraig12345, line 132:
```suggestion
```

PR #12897, opened 2020-04-28
Add react-docgen package

ecraig12345:
Move the v0 component info generation logic into a new package, `@fluentui/react-docgen`. This somewhat builds off Hudson/@Shavior's work in #12254 (thank you!) and works toward several goals, *not all implemented yet in this PR*:
- Have a shareable package so others can use the component info generation logic. (alpha)
- **Dramatically** speed up component info generation using what I learned in https://github.com/microsoft/fluentui/pull/12254#pullrequestreview-377221088 (soon)
- Move build scripts to just (soon)
- Remove conformance tests' dep on the component info build step (soon)

**In this PR, I just added the new package and mostly didn't use it yet or remove all the duplicate code.** The only exception is I deleted or modified a few files from `scripts/gulp/plugins` to allow seeing file diffs (where it was possible).

### Implementation notes

I followed Hudson's general implementation of separating out the "core" component info schema generation from the parts that are more specific to `@fluentui/react-northstar`, and providing a "schema resolver" function which fills in those library-specific properties. All the files and interfaces that are specific to `@fluentui/react-northstar` should have "fluent" in the name.

#### `getComponentInfo({ filePath, schemaResolver, program, tsconfigPath })`

Main function for a single file. You can pass in a pre-created `ts.Program` containing all files to be processed (this is the speed optimization mentioned earlier), a custom tsconfig path, or use the default config. `schemaResolver` is the callback for adding custom properties (none are added by default).

#### `getAllComponentInfo({ filePathsByPackage, schemaResolver })`

Recommended entry point for processing multiple files. It handles creating a shared `ts.Program` for each package for improved perf. The paths are passed in pre-grouped because it's hard to infer the grouping without knowledge of repo structure.

#### `fluentSchemaResolver({...})`

Adds properties specific to `@fluentui/react-northstar` when passed to `getComponentInfo` or `getAllComponentInfo`.

#### `getFluentConformanceInfo({ displayName, absPath })`

Get only the pieces of info which are currently needed for `isConformant`, **without** a dependency on our `react-docgen-typescript` fork for even faster perf. (Details of how this is used in another PR.) This shares helper functions with `getComponentInfo` and `fluentSchemaResolver`.

#### `getAllFluentBehaviorInfo(files: string[])`

Gets behavior info (formerly called behavior menu items) from behavior files. This is less generic/reusable but I had originally put this in the shared package so it can be easily used by the `@fluentui/accessibility` tests--not so sure about that now, and it could potentially just move back into `scripts`.

#### Removed properties and naming changes

I removed a couple redundant properties from the component info:

- `constructorName`: redundant with `displayName`
- `implementsCreateShorthand`: easily inferred from `mappedShorthandProp`; never directly referenced

I also changed `BehaviorMenu` to `BehaviorInfo` since the items are not actually used in a menu, and having "menu" in the name just makes it more confusing to those without history.

### What's next

Coming soon:
- Use the new package in tests
- Fully use the new package in build/doc site
- Probably add back the unit tests from Hudson's PR (or some variant)--omitted here due to being unfinished and adding even more files to the PR

The new package will also be ready to be polished for use by others, but **I do not plan to do this** since my main priority right now is working towards build convergence. (Whoever takes this on, please feel free to modify the APIs, naming, and structure in any way you see appropriate.)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12897)


packages/react-docgen/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "clean": "just-scripts clean",
+    "code-style": "just-scripts code-style",
+    "just": "just-scripts",
+    "lint": "just-scripts lint"
+  },
+  "devDependencies": {
+    "@types/doctrine": "^0.0.3",
+    "@types/node": "^10.3.2",
+    "@types/react": "16.8.25",
+    "@uifabric/tslint-rules": "^7.1.3",
```
layershifter, line 23: ???

packages/react-docgen/src/fluentSchemaResolver.ts
```diff
@@ -0,0 +1,95 @@
...
+      name: 'as',
+      required: false,
+    });
+  }
+  props = _.sortBy(props, 'name');
+
+  // tslint:disable:no-any
+  const componentClassName =
+    (Component as any).className ||
+    (Component as any).deprecated_className ||
```
layershifter, line 81:
```suggestion
```

https://github.com/microsoft/fluentui/pull/12765

packages/react-docgen/src/getAllFluentBehaviorInfo.ts
```diff
@@ -0,0 +1,53 @@
+import * as path from 'path';
+import * as _ from 'lodash';
+
+import { FluentBehaviorInfo, FluentBehaviorVariation, DocTag } from './types';
+import { getDocBlock } from './utils/index';
```
layershifter, line 5:
```suggestion
import { getDocBlock } from './utils';
```

packages/react-docgen/src/getFluentConformanceInfo.ts
```diff
@@ -0,0 +1,27 @@
+import { ComponentFileInfo, FluentApiPathInfo, FluentComponentInfo } from './types';
+import { getComponentFileInfo, getFluentApiPathInfo, getFluentComponentClassName, getDocBlock } from './utils/index';
```
layershifter, line 2:
```suggestion
import { getComponentFileInfo, getFluentApiPathInfo, getFluentComponentClassName, getDocBlock } from './utils';
```

packages/react-docgen/src/utils/getMappedShorthandProp.ts
```diff
@@ -57,10 +50,5 @@ const getShorthandInfo = (componentFile: t.File, componentName: string): Shortha
...
 
-  return {
-    implementsCreateShorthand,
-    mappedShorthandProp: mappedShorthandProp || '',
-  };
-};
-
-export default getShorthandInfo;
+  return mappedShorthandProp;
+}
```
layershifter, line 54: Please pick this update: https://github.com/microsoft/fluentui/pull/12985

ecraig12345:

PR #12898, opened 2020-04-28
Use react-docgen in build/docs

ecraig12345:
Use new package from #12897 in the v0 build steps and doc site (and fully delete the old versions of files). This should overall make the build faster--taking `build:docs:component-info` down **from about ~3.5min to < 20 seconds!** :tada:

### Notes on build changes

Obviously the primary change here is using the new package and deleting old code. The other big change is redefining `build:docs:component-info` and `build:docs:behavior-info` (and `build:docs:component-menu`) to *not* use a streaming approach--basically just glob the files and do the operations directly rather than processing single files as they come through the stream. Operating on all files was a requirement for the [perf optimization](https://github.com/microsoft/fluentui/pull/12254#pullrequestreview-377221088), plus helps unblock moving to just.

Moving away from the streaming approach means **no more fighting vinyl and `gulp-cache`** :tada: but unfortunately also means we lose the caching benefits (assuming it worked right) on subsequent builds. However, this is a lot less of a problem when the step is down to ~20 seconds not ~3.5 minutes, and we can look into alternative caching approaches (or single-file rebuilds) later if needed. (The component info build step will also be required less often once I remove that dependency from conformance tests soon.)

### Notes on doc site changes

In addition to the type name/location updates:

- Renamed `behavior menu` to `behavior info` since the items are not actually used in a menu now, and having "menu" in the name just makes it more confusing to those without history.
- Removed some logic left over from when behaviors were in the sidebar
- Fixed some types in various doc site files.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12898)


ecraig12345:

PR #12899, opened 2020-04-28
Use react-docgen in tests

ecraig12345:
Use new package from #12897 in conformance tests and behavior tests. 

### Conformance tests

The `fluentui/react-northstar` tests will start up much faster now because `getFluentConformanceInfo` only gets a stripped-down version of component info (doesn't depend on our `react-docgen-typescript` fork) and doesn't require running `build:docs:component-info` first.

The main immediate downside is that `getFluentConformanceInfo` needs the path to the component file being tested, and that's not easily available within Jest/Node any way that I could find. The only way I could find to get it is by *forcing the component to throw an exception while rendering* and looking at the stack. You can see the really horrible hacks to do this in `getComponentPath.tsx`. Unfortunately I also had to add something to `UIComponent` 
(Open to better suggestions here, but the benefits of much faster tests seem worth the hack to me.)

### Behavior tests

The behavior test setup file in `@fluentui/accessibility` now directly calls `getAllFluentBehaviorInfo` instead of depending on `behaviorMenu.json`. Perf characteristics should be similar, but it makes the code much cleaner by avoiding circular/implicit deps on the docs package.

I also noticed that almost all the behaviors get the generic tests, so I think it makes more sense to just find all the behavior filenames (which we have to do for `getAllFluentBehaviorInfo` anyway) and run the generic tests except on an exclude list--rather than manually importing each behavior and adding it to the test. (Can revert this part if desired but it makes sense to me?)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12899)


packages/fluentui/accessibility/test/behaviors/behavior-test.tsx
```diff
@@ -1,180 +1,21 @@
...
-testHelper.addBehavior('cardFocusableBehavior', cardFocusableBehavior);
-testHelper.addBehavior('cardChildrenFocusableBehavior', cardChildrenFocusableBehavior);
-testHelper.addBehavior('cardsContainerBehavior', cardsContainerBehavior);
-testHelper.addBehavior('videoBehavior', videoBehavior);
+// Add generic tests for all behaviors except ones on the exclude list
+const excludedBehaviors = ['alertBehavior', 'gridRowBehavior', 'listBehavior', 'listItemBehavior'];
+behaviorFiles
+  .map(filePath => path.basename(filePath).replace('.ts', ''))
+  .filter(behavior => !excludedBehaviors.includes(behavior))
+  .forEach(behavior => testHelper.addBehavior(behavior, accessibility[behavior]));
```
layershifter, line 19: https://github.com/microsoft/fluentui/blob/master/packages/fluentui/test-a-feature.md#behavior-tests
Please perform updates here

packages/fluentui/react-northstar/src/utils/UIComponent.tsx
```diff
@@ -37,7 +37,14 @@ class UIComponent<P, S = {}> extends React.Component<P, S> {
 
   constructor(props, context) {
     super(props, context);
+
     if (process.env.NODE_ENV !== 'production') {
+      // Hack used by conformance tests to get the path of the tested component
+      // eslint-disable-next-line
+      if (typeof window !== 'undefined' && (window as any).gettingComponentPath) {
+        throw new Error();
+      }
```
layershifter, line 46: How this will work with functional components?

packages/fluentui/react-northstar/test/specs/commonTests/utils/getComponentPath.tsx
```diff
@@ -0,0 +1,68 @@
...
+  try {
+    // Attempt to cause an exception to be thrown from within the component code, so we can get
+    // the path from the stack.
+    if (Component.prototype) {
+      // Class component. Try a couple different approaches.
+      new (Component as React.ComponentClass)(null); // eslint-disable-line
+      mount(<Component />);
+    } else {
+      // Function component.
+      // Calling as a function (not rendering) will likely result in an invalid hook error.
```
layershifter, line 38: And if not?

packages/fluentui/react-northstar/test/specs/commonTests/utils/getComponentPath.tsx
```diff
@@ -0,0 +1,68 @@
...
+    (window as any).gettingComponentPath = false;
+  }
+
+  if (!result) {
+    // eslint-disable-next-line no-console
+    console.error(error || `getComponentPath algorithm could not find a path for ${constructorName}!`);
+  }
+
+  return result;
+}
```
layershifter, line 68: Is there any other way to do it?

packages/fluentui/react-northstar/src/utils/UIComponent.tsx
ecraig12345: The hack was only needed for things extending UIComponent. I found other ways to get the name for things which don't extend UIComponent.

packages/fluentui/react-northstar/test/specs/commonTests/utils/getComponentPath.tsx
ecraig12345: In that case the test will fail and we'll find another way to make it work with the newly tested component.

packages/fluentui/react-northstar/test/specs/commonTests/utils/getComponentPath.tsx
ecraig12345: I haven't found one... :(

ecraig12345: To be clear I'd very much prefer something less terrible than this, I just couldn't think of anything that didn't involve updating all tests to manually pass the component's path. (But we could do that too.)

layershifter: What about Babel plugin? For tests we can do a special Babel plugin that does this.

The simplified thing that adds `export const __FILEPATH = 'REAL_FILEPATH'`:
```js
const filepathPlugin = ({ types: t }) => ({
  visitor: {
    Program: (path, state) => {
      const declaration = t.exportNamedDeclaration(
        t.variableDeclaration('const', [
          t.variableDeclarator(t.identifier('__FILEPATH__'), t.stringLiteral(state.file.opts.filename)),
        ]),
      );

      path.node.body.push(declaration);
    },
  },
});

module.exports = filepathPlugin;
```

In the similar way we can add static field to React components

ecraig12345: Interesting idea but I'm hesitant to add a required dep on a babel plugin to conformance tests, since we'd like to make these tests or similar usable for v7 and vnext as well, and at least in the immediate future those won't use babel. (I think also something that makes me nervous about highly customized babel usage in general is that it feels like adding "magic" to the build process. "Magic" features are nice until something breaks and you have to debug, and at that point you'd have to be intimately familiar with the repo-specific build process to understand what's going on.)

Though thinking about it more, @levithomason pointed out in a conversation last week that really most of these tests ought to be converted to lint rules, and then getting the file path wouldn't be an issue. So it might not matter so much which method we choose since it will probably only be temporary.

layershifter: Okay, let's go with the current appoach

ecraig12345:

PR #12910, opened 2020-04-29
Fixing stylis import.

dzearing:
I am getting errors from my local project using northstar code complaining about the `stylis` import:

image

Changing the import to:

```tsx
import * as Stylis from 'stylis';
```

Seems to resolve my project. It seems to break northstar build though.

@layershifter any ideas? 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12910)


dzearing:
Maybe this isn't the right fix. Things are busted here, but they're also busted without it in my feat/avatar branch.

layershifter:
@dzearing Ugly fix was added 🙈 Stylis V4 is ES thing, so after the update we will be able to remove it

dzearing:
Makes sense, thanks @layershifter !

PR #12913, opened 2020-04-29
RFC: how to handle tokens for existing CSS-in-JS approach in v0

mnajdova:
# Tokens structure

### Problem description

Variables (tokens) are a way of customizing styles in the Fluent components. For example:

```jsx
<Attachment
  header="Photo.jpg"
  action={{
    icon: <CloseIcon />,
    onClick: () => alert("'X' is clicked!"),
    title: 'Close',
  }}
  variables={{backgroundColor: 'lightblue'}}
/>
```

will update the attachment to have blue background. 

For customizing the root component itself we don't have any problems, but for the slots things are starting to get more interesting.

Let's see the next example

```jsx
<Attachment
  header="Photo.jpg"
  variables={{backgroundColor: 'lightblue', actionBackgroundColor: 'lightgreen'}}
  action={{
    icon: <CloseIcon />,
    title: 'Close',
  }}
/>
```

We would expect that the following will happen:

But that's not true unfortunately :(

However this code, will produce the prev look:

```jsx
<Attachment
  header="Photo.jpg"
  variables={{backgroundColor: 'lightblue'}}
  action={{
    style: { backgroundColor:'lightgreen' },
    icon: <CloseIcon />,
    title: 'Close',
  }}
/>
```

Let's see why this is happening

The `AttachmentAction` which is used for the action slot, is composing the `Button` component. The `Button` component has it's own styles and variables. If we just use the `Attachment`'s variables for the `AttachmentAction`, we will have a problem, because we will have variables name collision - both the `Button` and the `Attachment` have `backgroundColor` variable, that's why the last code snippet updated the background color for the action.

How can we solve this?

### Proposed solution

When composing components for our Fluent slots, we should always **override the styles** for the base component. That means that we will always need to write the styles from scratch, but it will allow us to use the parent's components variables and pick the correct variables which are intended for the slot (for example set the background color to be the `actionBackgroundColor` variable in the `AttachmentAction`'s styles). This will also save us from unpredictable regressions, when we are updating the styles for one component, but that is doing changes all over the codebases, where the styles of this component may have been used/composed.

In addition to this, it will mean that we won't need to have variables for the slots components. This will be possible because all variables for a component will be defined on the root level, where we can structure the variables names to always start with the slot name, or we can have object structure for the slots. 

#### Option 1

```js
const attachmentVariables = siteVars => ({
    // variables for the root slot
    rootBackgroundColor: siteVars.brand,
    rootColor: siteVars.neutral,
    
    // variables for the action slot
    actionBackgroundColor: siteVars.neutral,
    actionColor: siteVars.brand,
    
    // variables for the icon slot
    iconSize: '2rem',
});

```

#### Option 2

```js
const attachmentVariables = siteVars => ({
    // variables for the root slot
    root: {
        backgroundColor: siteVars.brand,
        color: siteVars.neutral,
    },
    
    // variables for the action slot
    action: {
        backgroundColor: siteVars.neutral,
        color: siteVars.brand,
    },
    
    // variables for the icon slot
    icon: { 
        size: '2rem'
    },
});

```

Apart from this, we have to share the variables between the parent and children components via context, so that the variables usage in the parent component can trigger updates in the styles of the children components inside.

I am leaning towards #Option1.1 as it is closed to what we have, plus it will preserve the flat variables structure, which is closed to the CSS variables

**BREAKING CHANGE**

Removed the following properties from the `AttachmentAction` component:
-   text
-   iconOnly
-   circular
-   size
-   fluid
-   inverted

If any of these is needed we can add it after it is requested


assuncaocharles:
IMO the 2nd option seems more organised and natural to think, I would guess also easy to maintain since the parts are well separated slot by slot?

layershifter:
Q1: How to pass variables down?

I see two options: via React Context or via props (breaks caching)?

Q2: Theme definition?

Currently we have `attachmentBodyVariables.ts` that contain only reexports are we going to solve it somehow?

```tsx
export { default } from './attachmentVariables';
```

Q3: Enforcing naming rules

It looks that all will fit pattern `SLOT_(PROP_VALUE)_CSS_VALUE`. Should we enforce it somehow? Via ESLint rule?

Q4: Collection components

Can you please add an example for collection components? How names will look there? `menuItemContentOpenBackgroundColor`? What will be `root` in this case?

dzearing:
I also intuitively went with option 1 in my prototyping.

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
```diff
@@ -61,104 +70,161 @@ export interface AttachmentProps extends UIComponentProps, ChildrenComponentProp
...
+        return;
+      }
+
+      _.invoke(props, 'onClick', e, props);
+    };
+
+    const element = (
+      <ElementType {...getA11Props('root', { className: classes.root, onClick: handleClick, ref, ...unhandledProps })}>
+        {createShorthand(composeOptions.slots.icon, icon, {
+          defaultProps: () => slotProps.icon || {},
```
layershifter, line 146:
```suggestion
          defaultProps: () => slotProps.icon,
```

?

layershifter: This looks extremely weird

packages/fluentui/react-northstar/src/components/Attachment/AttachmentAction.tsx
```diff
@@ -1,13 +1,29 @@
...
+import * as PropTypes from 'prop-types';
 import { compose, ComponentWithAs } from '@fluentui/react-bindings';
 
-import { createShorthandFactory, ShorthandFactory } from '../../utils';
+import { commonPropTypes, ShorthandConfig } from '../../utils';
 import Button, { ButtonProps, ButtonStylesProps } from '../Button/Button';
 
 interface AttachmentActionOwnProps {}
-export interface AttachmentActionProps extends AttachmentActionOwnProps, ButtonProps {}
+export interface AttachmentActionProps extends AttachmentActionOwnProps, ButtonProps {
```
layershifter, line 10: Should we use `Pick` instead of `never`?

packages/fluentui/react-northstar/src/components/Attachment/AttachmentAction.tsx
```diff
@@ -22,16 +38,30 @@ const AttachmentAction = compose<
...
+  }),
+  disabled: PropTypes.bool,
+  icon: customPropTypes.shorthandAllowingChildren,
+  iconPosition: PropTypes.oneOf(['before', 'after']),
+  loader: customPropTypes.itemShorthandWithoutJSX,
+  loading: PropTypes.bool,
+  onClick: PropTypes.func,
+  onFocus: PropTypes.func,
+  primary: customPropTypes.every([customPropTypes.disallow(['secondary']), PropTypes.bool]),
+  secondary: customPropTypes.every([customPropTypes.disallow(['primary']), PropTypes.bool]),
```
layershifter, line 60: Do we really support `loader`, `loading`, `primary`, `secondary`, etc.?

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
```diff
@@ -61,104 +70,161 @@ export interface AttachmentProps extends UIComponentProps, ChildrenComponentProp
...
+        {(header || description) &&
+          createShorthand(composeOptions.slots.body, body, {
+            defaultProps: () => slotProps.body || {},
+            overrideProps: predefinedProps => ({
+              content: (
+                <>
+                  {createShorthand(composeOptions.slots.header, header, {
+                    defaultProps: () => slotProps.header || {},
+                    overrideProps: predefinedProps => ({
+                      variables: mergeComponentVariables(variables, predefinedProps.variables),
```
layershifter, line 161: Do we have a short circuit in `mergeComponentVariables` (`if (!a && !b) return undefined`)? As I remember it will always create a function and break caching

packages/fluentui/react-northstar/src/components/Attachment/AttachmentBody.tsx
```diff
@@ -18,10 +18,12 @@ const AttachmentBody = compose<'div', AttachmentBodyOwnProps, AttachmentBodyStyl
     className: attachmentBodyClassName,
     displayName: 'AttachmentBody',
   },
-) as ComponentWithAs<'div', AttachmentBodyProps> & { create?: ShorthandFactory<AttachmentBodyProps> };
+) as ComponentWithAs<'div', AttachmentBodyProps> & { shorthandConfig: ShorthandConfig<AttachmentBodyProps> };
 
 AttachmentBody.propTypes = commonPropTypes.createCommon();
 
-AttachmentBody.create = createShorthandFactory({ Component: AttachmentBody, mappedProp: 'content' });
+AttachmentBody.shorthandConfig = {
```
layershifter, line 25: Technically this is a breaking change, but as this was never released we are good

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
mnajdova: Well we wanted this not to he function, so now we have to do this. If nothing is defined here it is undefined, or I guess we can go ahead and change the `createShorthand` factory again.. Maybe we should change it, what do you think?

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
mnajdova: Let me check that, good call

packages/fluentui/react-northstar/src/components/Attachment/AttachmentAction.tsx
mnajdova: I left the styles necessary for them.. I thought we should leave these, as loading and laoder are generic enough, and we have primary and secondary everywhere, so it would be awkward not to have them here..

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
mnajdova: Yeah, it was breaking the caching, should be fixed now

packages/fluentui/react-northstar/src/components/Attachment/AttachmentAction.tsx
mnajdova: In my opinion was easier to just turn off the ones I don't want to use instead of pick everything else, as I am pretty sure it's easier to forget at least some of the common props...

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
mnajdova: Fixed, with change in the factories

packages/fluentui/react-northstar/src/components/Attachment/Attachment.tsx
```diff
@@ -140,12 +140,16 @@ const Attachment = compose<'div', AttachmentProps, AttachmentStylesProps, {}, {}
       _.invoke(props, 'onClick', e, props);
     };
 
+    const mergeShorthandVariables = (variables, shorthandVariables) => {
+      return (variables || shorthandVariables) && mergeComponentVariables(variables, shorthandVariables);
+    };
```
layershifter, line 145: At least let's move this out of render scope and add a comment to do something with as obviously this about is not a key to the bright future

mnajdova: Moved and added comment. I agree that we should wither fix mergeComponentVariables or provide similar utility

packages/fluentui/CHANGELOG.md
```diff
@@ -79,7 +79,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Converted Box, Button* and Attachment* components to use compose @layershifter @mnajdova ([#12745](https://github.com/microsoft/fluentui/pull/12745))
 - Respect `childBehaviors` for slots in `useAccessibility` hook @pompomon ([#12835](https://github.com/microsoft/fluentui/pull/12835))
 - Add `cardsContainerBehavior` for `Card`s in `Grid` navigation @mpompomon ([#12800](https://github.com/microsoft/fluentui/pull/12800))
-- Added `slots` and `mapPropsToSlotProps` in compose and added example usage in the `Button` component @mnajdova ([#12858](https://github.com/microsoft/fluentui/pull/12858))
+- Add `slots` and `mapPropsToSlotProps` in compose and added example usage in the `Button` component @mnajdova ([#12858](https://github.com/microsoft/fluentui/pull/12858))
+- Change `AttachmentAction` to override `Button` styles @mnajdova ([#12913](https://github.com/microsoft/fluentui/pull/12913))
```
layershifter, line 83: Is this breaking change?

packages/fluentui/react-northstar/src/utils/factories.ts
```diff
@@ -150,7 +150,7 @@ export function createShorthandInternal<P>({
   // ----------------------------------------
   // Build up props
   // ----------------------------------------
-  const defaultProps = options.defaultProps ? options.defaultProps() : ({} as Props<P>);
+  const defaultProps = options.defaultProps ? options.defaultProps() || ({} as Props<P>) : ({} as Props<P>);
```
layershifter, line 153: Should we add UT for this case?

mnajdova: Make sense, will add it

packages/fluentui/CHANGELOG.md
mnajdova: Yeah actually, it is a breaking change as I am reducing the num of props too. Will move it and make list of the removed props in the PR description.

PR #12917, opened 2020-04-29
Charting: Change FocusZone imports to @fluentui/react-focus from office-ui-fabric-react/lib/FocusZone

RajeshGoriga:
#### Pull request checklist

#### Description of changes
Change FocusZone imports to @fluentui/react-focus from office-ui-fabric-react/lib/FocusZone in charts

#### Focus areas to test

Chart keyboard accessibility 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12917)


packages/charting/package.json
```diff
@@ -53,6 +53,7 @@
     "sinon": "^4.1.3"
   },
   "dependencies": {
+    "@fluentui/react-focus": "7.7.0",
```
joschect, line 56: Looks like you're getting a failure because this doesn't match what's in production and it's hardcoded. I believe that there is some work done to ensure that all packages within the fluent repo stay in sync. @ecraig12345 might know more.

```suggestion
    "@fluentui/react-focus": "^7.7.0",
```

joschect:
@RajeshGoriga Thanks for submitting this change! It looks like you need to generate a change file by running `yarn change` at the root.

packages/charting/package.json
ecraig12345: It just needs to match the version elsewhere in the repo. react-focus was published again since this change was made, so you need to do `^7.7.1` now.

RajeshGoriga: I have updated to ^7.7.1

PR #12918, opened 2020-04-29
chore(Embed): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `Embed` component to be functional. Restricting props set that will be passed to styles functions.

Related to #12237

## Prop sets

| `Embed`    |
| --------- |
| `active` |
| `iframeLoaded` |

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Embed/Embed.tsx
+++ b/packages/fluentui/react-northstar/src/components/Embed/Embed.tsx
@@ -1,22 +1,32 @@
-import { Accessibility, embedBehavior } from '@fluentui/accessibility';
+import { Accessibility, embedBehavior, EmbedBehaviorProps } from '@fluentui/accessibility';
 import * as _ from 'lodash';
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
 import * as customPropTypes from '@fluentui/react-proptypes';
 
-import {
-  createShorthandFactory,
-  UIComponentProps,
-  applyAccessibilityKeyHandlers,
-  commonPropTypes,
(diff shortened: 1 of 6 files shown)


packages/fluentui/react-northstar/src/components/Embed/Embed.tsx
```diff
@@ -26,6 +36,12 @@ export interface EmbedProps extends UIComponentProps {
   /** Accessibility behavior if overridden by the user. */
   accessibility?: Accessibility;
```
pompomon, line 37:
```suggestion
  accessibility?: Accessibility<EmbedBehaviorProps>;
```
?

packages/fluentui/react-northstar/src/components/Embed/Embed.tsx
```diff
@@ -60,143 +76,173 @@ export interface EmbedProps extends UIComponentProps {
...
+            </Ref>
+          )}
+        </>
+      )}
+
+      {placeholderVisible && placeholderElement}
+      {controlVisible &&
+        Box.create(control, {
+          defaultProps: () =>
+            getA11yProps('control', {
```
pompomon, line 209: Please why do you call `getA11yProps` for control slot, but not for iframe slot on line 193?

packages/fluentui/react-northstar/src/components/Embed/Embed.tsx
assuncaocharles: There's no behavior for Embed

PR #12921, opened 2020-04-29
feat(Card): Selectable card

pompomon:
#### Pull request checklist

#### Description of changes

Adding props, styles and examples for selectable cards

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12921)


packages/fluentui/accessibility/src/behaviors/Card/cardBehavior.ts
```diff
@@ -5,12 +5,14 @@ import { Accessibility } from '../../types';
...
  * Adds role='group'.
  * Adds attribute 'aria-disabled=true' based on the property 'disabled'.
+ * Adds attribute 'aria-selected=true' based on the property 'selected'.
  */
 const cardBehavior: Accessibility<CardBehaviorProps> = props => ({
   attributes: {
     root: {
       role: 'group',
       'aria-disabled': props.disabled,
+      'aria-selected': props.selected,
```
jurokapsiar, line 15: `aria-selected` is not valid for `group`: https://www.w3.org/WAI/PF/aria/roles#group - if we tested that screen readers accept it, we need to document it.

kolaps33:
NVDA narrate blank, when navigate in browse mode (after navigation from user card):
Please see logs below:

user card  clickable    not selected
User #1
blank
out of user card  user card  clickable    not selected
User #2
blank
out of user card  user card  clickable    not selected

kolaps33:
image

Now reader doesn't narrate position of the card. 
I see you keep position of card in aria-label on the element which has role=group
What I think is fine :) only you are missing to reference to the element itself in aria-labelledby.

kolaps33:
Now we can do selection on the card by "Space" or "Enter" key. General question is, if we want to do it be "Enter" key as well? Just raising the question because in some components only space works for selection. Tree for example. But then in selectable list both Enter and Space work.

I think we will not find any recommendation from ARIA about selectable cards.

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,138 @@
...
+      throw new Error(`Action ${action.type} is not supported`);
+  }
+};
+
+const CardExampleSelectableGrid = () => {
+  // Component setup
+  const cardsNumber = 12;
+  const cards = Array(cardsNumber)
+    .fill(undefined)
+    .map((item, index) => ({ key: index + 1, title: `Card ${index + 1}` }));
```
silviuaavram, line 95: Why not `index` instead of `key`? Easier to understand.

packages/fluentui/react-northstar/src/components/Card/Card.tsx
```diff
@@ -91,19 +99,25 @@ const Card: React.FC<WithAsProp<CardProps>> &
...
+    selected,
   } = props;
   const ElementType = getElementType(props);
   const unhandledProps = useUnhandledProps(Card.handledProps, props);
   const getA11yProps = useAccessibility(props.accessibility, {
     debugName: Card.displayName,
     actionHandlers: {
       performClick: e => {
+        // prevent Spacebar from scrolling
+        e.preventDefault();
```
silviuaavram, line 111: Can this be executed in a context of a submission on Enter? Like submitting a form POST? If so, this will prevent it.

pompomon: I don't think cards would have submission elements inside, @jurokapsiar wdyt?

jurokapsiar: I would only expect space to select/unselect, not enter - similar to checkbox or selectable tree items.

pompomon: Fixed, now selectable card has a separate behavior that triggers preformClick only on Spacebar

ecraig12345:
@pompomon Looks like you need to merge with master. That should also fix the build.

pompomon:
Done :)

ecraig12345:
Now it looks like someone needs to [accept new screener states](https://screener.io/v2/dashboard/H17FwCf88.microsoft-fluentui-fluentui/feat%2Fcard-select). Not doing that myself since I don't actually work on this. :)

pompomon:
Yep, but I'm still waiting for code owners :)

packages/fluentui/accessibility/src/behaviors/Card/cardSelectableBehavior.ts
```diff
@@ -0,0 +1,34 @@
+import { Accessibility } from '../../types';
+import { CardBehaviorProps } from './cardBehavior';
+import * as keyboardKey from 'keyboard-key';
+import { IS_FOCUSABLE_ATTRIBUTE } from '../../attributes';
+
+/**
+ * @description
+ * Behavior for a focusable card component
```
layershifter, line 8:
```suggestion
 * Behavior for a selectable card component
```

?

packages/fluentui/accessibility/src/behaviors/Card/cardsContainerBehavior.ts
```diff
@@ -11,7 +11,7 @@ import * as keyboardKey from 'keyboard-key';
  * Provides arrow key navigation in bidirectional direction.
  * Focus can be moved inside a child component with embeded inner FocusZone by pressing a specified key.
  */
-const cardsContainerBehavior: Accessibility = () => ({
+const cardsContainerBehavior: Accessibility<never> = () => ({
```
layershifter, line 14: I suggest to introduce `CardsContainerBehaviorProps` instead

packages/fluentui/docs/src/examples/components/Card/States/CardExampleSelected.tsx
```diff
@@ -0,0 +1,19 @@
+import { Flex, Image, Text, Card } from '@fluentui/react-northstar';
+import * as React from 'react';
+
+const CardExampleSelected = () => {
+  return (
+    <Card selected aria-roledescription="selected card">
+      <Card.Header>
+        <Text content={`Selected card`} weight="bold" />
```
layershifter, line 8:
```suggestion
        <Text content="Selected card" weight="bold" />
```

packages/fluentui/docs/src/examples/components/Card/States/CardExampleSelected.tsx
```diff
@@ -0,0 +1,19 @@
...
+      </Card.Header>
+      <Card.Body>
+        <Flex column gap="gap.small">
```
layershifter, line 11: Do we need `Flex` there?

packages/fluentui/react-northstar/src/components/Card/Card.tsx
```diff
@@ -56,9 +56,17 @@ export interface CardProps extends UIComponentProps {
 
   /** A card can show that it cannot be interacted with. */
   disabled?: boolean;
+
+  /**
+   * A card can show that it is currently selected or not.
+   */
```
layershifter, line 62:
```suggestion
  /** A card can show that it is currently selected or not. */
```

packages/fluentui/react-northstar/src/themes/teams/components/Card/cardStyles.ts
```diff
@@ -52,6 +52,11 @@ const cardStyles: ComponentSlotStylesPrepared<CardStylesProps, CardVariables> =
           boxShadow: v.boxShadowPressed,
         },
       }),
+      ...(p.selected && {
+        backgroundColor: v.backgroundColorSelected,
+        borderColor: v.borderColorSelected,
+        boxShadow: v.boxShadowSelected,
```
layershifter, line 58:
```suggestion
        backgroundColor: v.selectedBackgroundColor,
        borderColor: v.selectedBorderColor,
        boxShadow: v.selectedBoxShadow,
```

Should prop be first there?

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,141 @@
...
+    .fill(undefined)
+    .map((item, index) => ({ index: index + 1, title: `User ${index + 1}` }));
+  const initialState: SelectableCardsGridState = {
+    cards: cards.reduce((cards, card) => {
+      cards[card.index] = false;
+      return cards;
+    }, {}),
+  };
+  const [state, dispatch] = React.useReducer(selectableCardsGridStateReducer, initialState);
+  const handleClick = (isSelected, index) => {
```
layershifter, line 105: As `handleClick` is not memoized let's inline it

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,141 @@
...
+      </Card.Body>
+    </Card>
+  );
+};
+
+interface SelectableCardsGridAction {
+  type: 'TOGGLE_ITEM' | 'TOGGLE_ALL';
+  selected: boolean;
+  itemKey?: string;
+}
```
layershifter, line 70: `type` will work there better

```ts
type SelectableCardsGridActions = { type: 'TOGGLE_ITEM', selected: boolean; itemKey: string } | { type: 'TOGGLE_ALL'; selected: boolean; }
```

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,141 @@
...
+      />
+      <Grid accessibility={cardsContainerBehavior} columns="3">
+        {cards.map(card => {
+          return (
+            <SelectableCard
+              key={card.index}
+              index={card.index}
+              title={card.title}
+              aria-label={`${card.title} ${card.index} of ${cardsNumber}`}
+              handleClick={handleClick}
```
layershifter, line 131:
```suggestion
              onClick={handleClick}
```

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,141 @@
...
+  cards: Record<string, boolean>;
+}
+
+const selectableCardsGridStateReducer: React.Reducer<SelectableCardsGridState, SelectableCardsGridAction> = (
+  state,
+  action,
+) => {
+  switch (action.type) {
+    case 'TOGGLE_ITEM': {
+      return { cards: { ...state.cards, [action.itemKey]: action.selected } };
```
layershifter, line 82:
```suggestion
      return { cards: { ...state.cards, [action.index]: action.selected } };
```

Why it's `itemKey`? Should be just `index`?

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,141 @@
...
+  );
+};
+
+interface SelectableCardsGridAction {
+  type: 'TOGGLE_ITEM' | 'TOGGLE_ALL';
+  selected: boolean;
+  itemKey?: string;
+}
+
+interface SelectableCardsGridState {
```
layershifter, line 72: If there is nothing except cards let's simplify to 
```suggestion
type SelectableCardsGridState = Record<string, boolean>
```

packages/fluentui/react-northstar/src/components/Card/Card.tsx
```diff
@@ -143,6 +158,8 @@ const Card: React.FC<WithAsProp<CardProps>> &
           {...getA11yProps('root', {
             className: classes.root,
             onClick: handleClick,
+            selected,
```
layershifter, line 161: I don't think that `selected` and `disabled` are applicable to `div`s

packages/fluentui/docs/src/examples/components/Card/States/CardExampleSelected.tsx
```diff
@@ -0,0 +1,17 @@
...
+    <Card selected aria-roledescription="selected card">
+      <Card.Header>
+        <Text content="Selected card" weight="bold" />
+      </Card.Header>
+      <Card.Body>
+        <Image src="public/images/wireframe/square-image.png" />
+      </Card.Body>
+    </Card>
+  );
+};
```
layershifter, line 15:
```suggestion
const CardExampleSelected = () => (
  <Card selected aria-roledescription="selected card">
    <Card.Header>
      <Text content="Selected card" weight="bold" />
    </Card.Header>
    <Card.Body>
      <Image src="public/images/wireframe/square-image.png" />
    </Card.Body>
  </Card>
);
```

nit

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,134 @@
...
+        {cards.map(card => {
+          return (
+            <SelectableCard
+              key={card.index}
+              index={card.index}
+              title={card.title}
+              aria-label={`${card.title} ${card.index} of ${cardsNumber}`}
+              onClick={(isSelected, index) => {
+                dispatch({ type: 'TOGGLE_ITEM', selected: isSelected, index: index });
+              }}
```
layershifter, line 124:
```suggestion
              onClick={(isSelected, index) => {
                dispatch({ type: 'TOGGLE_ITEM', selected: isSelected, index: card.index });
              }}
```

It looks that we can get `index` from the current scope and `SelectableCard` will not need special signature for `onClick`

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,134 @@
...
+  // Component setup
+  const cardsNumber = 12;
+  const cards = Array(cardsNumber)
+    .fill(undefined)
+    .map((item, index) => ({ index: index + 1, title: `User ${index + 1}` }));
+  const initialState: SelectableCardsGridState = {
+    ...cards.reduce((cards, card) => {
+      cards[card.index] = false;
+      return cards;
+    }, {}),
```
layershifter, line 96:
```suggestion
  const initialState: SelectableCardsGridState = cards.reduce((cards, card) => {
    cards[card.index] = false;
    return cards;
  }, {}),
```

nit

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,134 @@
...
+import * as _ from 'lodash';
+
+type SelectableCardProps = {
+  index?: number;
+  title?: string;
+  selected?: boolean;
+  onClick?: Function;
+};
+
+const SelectableCard: React.FC<SelectableCardProps> = ({ title, index, selected, onClick, ...unhadledProps }) => {
```
layershifter, line 24:
```suggestion
const SelectableCard: React.FC<SelectableCardProps> = ({ title, index, selected, onClick }) => {
```

I would avoid having there `unhadledProps`, let's have:
- all properties specified
- call it `rest`

layershifter: Reason: `unhadledProps` are coming from `useUnhandledProps` and are used in our components. However, in this case it's just rest/spread

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleSelectableGrid.tsx
```diff
@@ -0,0 +1,134 @@
...
+  hiddenComponentBehavior,
+} from '@fluentui/react-northstar';
+import * as React from 'react';
+import * as _ from 'lodash';
+
+type SelectableCardProps = {
+  index?: number;
+  title?: string;
+  selected?: boolean;
+  onClick?: Function;
```
layershifter, line 21:
```suggestion
  onClick?: (e: React.SyntheticEvent) => void;
```

layershifter: https://github.com/microsoft/fluentui/pull/12921/files#r427108402

Issue #12770, opened 2020-04-17
Fluent UI v8


dzearing:
**UPDATE Feb. 9: Target release date is February 22, or end of February at latest. See [this comment](https://github.com/microsoft/fluentui/issues/12770#issuecomment-776389207) detailing a few significant changes from the original beta version.**

For current details of the release, please see the [draft release notes and migration guide](https://github.com/microsoft/fluentui/wiki/Version-8-release-notes).

Remaining work is tracked [here](https://github.com/microsoft/fluentui/projects/36). 

<details>
<summary>Expand for original content, preserved for reference but may be out of date.</summary>

## Summary
The next release of Fluent UI React is going to be a little different than previous releases of this project and we wanted to help set expectations, detail some of its nuances, and set the stage for subsequent releases.

We view this next release as an incremental release that will set up customers, partners, and contributors for the team's collective vision of improving the baseline components used in all Microsoft 365 experiences. Going forward we will be making smaller, easy to consume, incremental releases of Fluent UI.

The Version 8 (v8) release will focus on:

1. **This is an incremental release**
    * No major visual changes
    * Limited API surface changes (mainly Button)
    * [Package restructuring](#user-content-packaging-restructuring) paves the way for smoother updates in the future
    * Intended to be an easy upgrade
1. **[Adds React Strict Mode Support](#user-content-react-strict-mode-support)**
    * Unblocks the use of [React Strict Mode](https://reactjs.org/docs/strict-mode.html) while developing applications
    * Useful for testing out [React Concurrent Mode](https://reactjs.org/docs/concurrent-mode-intro.html)
    * Converted several components to Function Components and removed deprecated lifecycle methods on the rest

All content in this issue is a work in progress and will not be final until we approach release.
*Note: "v8" here refers to "version 8," not the v8 JavaScript engine.*

~~New functionality available outside of Version 8. ~~
~~The Improved Button was part of the Version 8 release, but we found that it complicated the upgrade story, especially since the old button was still used in some other components. The same button is still available from `@fluentui/react-button` but has no longer replaced the default button. ~~
1. ~~**Improved Button** ~~
    * ~~Smaller bundle size, better render perf~~
    * ~~Introduction of design tokens and slots - looking for feedback!~~
    * ~~Please see the [@fluentui/react-button package README](https://github.com/microsoft/fluentui/tree/master/packages/react-button) for details about improvements.~~

## Timeline
*Note: These dates are our best estimations.*

| Milestone | Date |
[...]

PR #12923, opened 2020-04-29
Migrate Fabric component to function component

MLoughry:
#### Pull request checklist

#### Description of changes

Migrate the `Fabric` component to a function component

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12923)


packages/react-next/src/components/Fabric/Fabric.base.tsx
```diff
@@ -0,0 +1,94 @@
...
+    // Or If contextDir !== pageDir
+    // Then we need to set dir of the Fabric root
+    rootDir: componentDir !== contextDir || componentDir !== pageDir ? componentDir : dir,
+    // If dir !== contextDir || pageDir
+    // then set contextual theme around content
+    needsTheme: componentDir !== contextDir,
+  };
+};
+
+// tslint:disable-next-line:function-name no-function-expression
```
xugao, line 36: should be anonymous function + displayName ?

packages/react-next/src/components/Fabric/Fabric.base.tsx
```diff
@@ -0,0 +1,94 @@
...
+    if (applyThemeToBody) {
+      const currentDoc = getDocument(rootElement.current);
+      if (currentDoc) {
+        currentDoc.body.classList.add(bodyThemed);
+        return () => {
+          currentDoc.body.classList.remove(bodyThemed);
+        };
+      }
+    }
+  }, [bodyThemed]);
```
xugao, line 91: i think we need `rootElement.current` here too

MLoughry: A change to the ref value doesn't take effect until the render is pushed to the DOM, won't trigger a rerender. If we're concerned about it, we'd have to use a state value instead.

xugao: makes sense

PR #12926, opened 2020-04-29
Migrate Popup to function component

MLoughry:
#### Pull request checklist

- [x] Addresses an existing issue:  #12770

#### Description of changes

Migrates `Popup` to a function component.

Additionally adds helper hooks `useAsync()` and `useOnEvent()` to use `Async` and `on()` utilities and handle cleanup.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12926)


packages/react-hooks/README.md
```diff
@@ -109,34 +109,3 @@ const MyComponent = () => {
   // ... code that shows a dialog when a button is clicked ...
 };
 ```
-
```
ecraig12345, line 112: Bad merge?

MLoughry: Yeah, sorry about that. Fixed now

packages/react-hooks/src/useOnEvent.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Hook to attach an event handler on mount and handle cleanup
+ * @param element- Element (or ref to an element) to attach the event handler to
+ * @param eventName- The event to attach a handler for
+ * @param callback- The handler for the event
+ * @param useCapture- Whether or not to attach the handler for the capture phase
+ */
+export function useOnEvent<TElement extends Element>(
+  element: React.RefObject<TElement | undefined | null> | TElement | Window | undefined | null,
+  eventName: string,
+  callback: (ev: Event) => void,
```
ecraig12345, line 14: Do we want a generic for the event type (TEvent extends Event)?

packages/react-hooks/src/useOnEvent.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (element && 'current' in element) {
+      element = element.current;
+    }
+    if (!element) {
+      return;
+    }
+
+    const dispose = on(element, eventName, callback, useCapture);
+    return dispose;
+  }, [element, eventName, callback, useCapture]);
```
ecraig12345, line 27: If someone passes in a callback that mutates every render, this is going to result in a lot of extra attaching/detaching events. Not sure if that's a problem practically, and I don't think we should be artificially preventing it inside the hook (in case the user does want to change the callback), but maybe at least add a note in the doc comment?

MLoughry: I changed the code to stash the callback in a ref, to obviate the need to cleanup/reattach in such cases.

ecraig12345:
Seeing if this works...
@msft-github-bot require @joschect to approve

Issue #12350, opened 2020-03-18
Detect and prevent duplicate exports


ecraig12345:
We need an automated way to detect (and error) if anything is being exported multiple times from the index file, because this breaks consumers using certain setups. 

This has come in as an issue multiple times for different components (there may be more):  #11534, #10017, #9661

aneeshack4:
Tagging build area owner @kenotron for visibility

PR #12928, opened 2020-04-29
Migrate Calendar components to function components

MLoughry:
#### Pull request checklist

#### Description of changes

Migrate all Calendar components to function components.

Also added a utility method for function components to apply default properties, as a replacement for `static defaultProps`

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12928)


MLoughry:
@dzearing @ecraig12345 Not sure of the advice concerning [this build error](https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=71364&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=cb33ee24-f492-54d2-727a-ae0a7bb84b2c&l=1096):
```
@fluentui/react-next: [8:02:30 PM] x src/index.ts:18:1 - error TS2308: Module './Calendar' has already exported a member named 'DateRangeType'. Consider explicitly re-exporting to resolve the ambiguity.
@fluentui/react-next: 18 export * from './DatePicker';
```

Basically, since certain types are being re-exported from multiple components, and those types are now technically different when they're from different source files (OUFR vs react-next), we get this error.

Should we just copy OUFR over to react-next wholesale?

ecraig12345:
@MLoughry Would you mind splitting out the default props utility into a different PR? It will be easier to get that specific part reviewed and merged.

@lorejoh12 Could you take a look at the other parts of this?

MLoughry:
> @MLoughry Would you mind splitting out the default props utility into a different PR? It will be easier to get that specific part reviewed and merged.

I created #13152 to split that off.

ecraig12345:
> Basically, since certain types are being re-exported from multiple components, and those types are now technically different when they're from different source files (OUFR vs react-next), we get this error.
> 
> Should we just copy OUFR over to react-next wholesale?

The issue with exporting stuff from multiple places is probably something we should fix as part of version 8--apparently it's already causing problems for people in some cases (see #12350). 

In some cases it will be a bit tricky to determine the single "right" place things should be exported from, but it seems pretty obvious that DatePicker shouldn't be exporting Calendar stuff. So maybe for this specific case, you could change react-next/src/DatePicker to manually export only DatePicker stuff?

MLoughry:
@ecraig12345 @joschect Any further concerns blocking sign-off?

dzearing:
I've started a conversation with the code owners here; not sure about this one.

dzearing:
Ok after talking with @lorejoh12 , let's do this:

1. Let's NOT move the old Calendar into next.
2. Let's convert the components in date-time and export those.

There are numerous issues which have been resolved in date-time, but not in the existing Calendar due to breaking changes. It would be wasteful to convert it and then replace it later.

PR #12932, opened 2020-04-29
update MessageBarButton border color

### LINKED ISSUE #9184: MessageBar - Fluent Updates
New fluent messagebar styles use the default button instead of the custom gray button we currently use

image

Awaiting responsive design revision

chrismohr:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #9184

#### Description of changes

Updates default MessageBarButton border to match designs.

Previously MessageBarButton overrode the borderColor to be `theme.palette.neutralTertiaryAlt`.  I've removed this override to allow it to take on the default button border styles. 

Alternatively we could consider adding a new slot, `MessageButtonBorder`, if we want to ensure more consistency.

|Before | After
|-|-|
|beforeLight|afterLight|
|beforeDark|afterDark|

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12932)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Button/MessageBarButton/MessageBarButton.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Button/MessageBarButton/MessageBarButton.styles.ts
@@ -1,28 +1,25 @@
 import { IButtonStyles } from '../Button.types';
 import { ITheme, concatStyleSets, getFocusStyle } from '../../../Styling';
 import { memoizeFunction } from '../../../Utilities';
-import { getStyles as getBaseButtonStyles } from '../BaseButton.styles';
 
 export const getStyles = memoizeFunction(
-  (theme: ITheme, customStyles?: IButtonStyles, focusInset?: string, focusColor?: string): IButtonStyles => {
-    const baseButtonStyles: IButtonStyles = getBaseButtonStyles(theme);
-    const messageBarButtonStyles: IButtonStyles = {
-      root: [
-        getFocusStyle(theme, {
-          inset: 1,
-          highContrastStyle: {
-            outlineOffset: '-4px',
-            outlineColor: 'ActiveBorder',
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/Button/MessageBarButton/MessageBarButton.styles.ts
```diff
@@ -1,28 +1,25 @@
...
-          borderColor: 'transparent',
-        }),
-        {
-          height: 24,
-          borderColor: theme.palette.neutralTertiaryAlt,
-        },
-      ],
-    };
-
-    return concatStyleSets(baseButtonStyles, messageBarButtonStyles, customStyles)!;
```
chrismohr, line 26: Unless I'm missing something, there is no benefit to concatting `baseButtonStyles`, since this will also be done within the BaseButton component.

chrismohr:
FYI @phkuo, @micahgodbolt heres a quick additional follow up to the MessageBar changes.

ecraig12345:
@betrue-final-final Can you review the screener changes?

packages/office-ui-fabric-react/src/components/Button/MessageBarButton/MessageBarButton.styles.ts
ecraig12345: It looks like all the button variants are doing a manual concat with the baseButtonStyles, but I'm not sure why... @dzearing do you know the history here?

dzearing: lgtm

dzearing: I don't fully remember the history.

PR #12944, opened 2020-04-30
fix(Accordion): Control itemsRef

assuncaocharles:
#### Pull request checklist

#### Description of changes

Prevents creating a new ref all renders

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12944)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Accordion/Accordion.tsx
+++ b/packages/fluentui/react-northstar/src/components/Accordion/Accordion.tsx
@@ -191,7 +191,7 @@ export const Accordion: React.FC<WithAsProp<AccordionProps>> &
     true,
   );
 
-  let [itemRefs] = React.useState(Array(panels.length).fill(React.useRef<HTMLElement>()));
+  const itemRefs = React.useMemo(() => Array.from({ length: panels.length }, () => React.createRef()), [panels.length]);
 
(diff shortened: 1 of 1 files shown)


assuncaocharles:
> PR in the current state will produce double rendering initially and on each `panels.length` change. This is visible in perf tests and this is no way.
> 
> I suggest to use React's Context for this and explore the approach taken in Reach UI to properly address issue: https://github.com/reach/reach-ui/tree/master/packages/descendants
> 
> https://github.com/reach/reach-ui/blob/25840c3d15278e316cc74eaecf096fa01377f361/packages/accordion/src/index.tsx#L363
> https://github.com/reach/reach-ui/blob/25840c3d15278e316cc74eaecf096fa01377f361/packages/descendants/src/index.tsx#L108

Not sure if we need context here. I think we are trying to solve a smaller problem that the one linked there? Just trying to avoid to create new refs every render. Going more towards what they have in this discussion:

https://github.com/facebook/react/issues/14072

layershifter:
Please also add a changelog entry as for me it looks like a fix 🔧

PR #12946, opened 2020-04-30
Add an imperitive focus call to handleTouch

### LINKED ISSUE #12486: iOS13, Split Buttons when tapped twice return focus back to the last focused input
### Environment Information

- **Package version(s)**: 7.103.0
- **Browser and OS versions**: iOS 13 Safari. Does not reproduce in iOS 12 or emulated browser scenarios.

https://codepen.io/saulnier/pen/eYNxorZ

1) Focus into the CED and type some text. Also reproduces with a regular `<input>`
2) Tap on a split button.
3) Tap on the same split button again.

#### Actual behavior:

The split button's menu dismisses, and focus remains outside the input.

#### Expected behavior:

The split button's menu dismisses, but *focus returns to the input, showing the keyboard*

Are you willing to submit a PR to fix? Yes, with guidance.

Requested priority: High

Products/sites affected: Outlook on the Web is impacted in our touch scenarios.

Lego6245:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12486 

#### Description of changes

Split buttons in touch scenarios would return to their last focused input when tapped twice. This made interactions with them more difficult. Adding an imperative `focus` call as part of handling the touch events, as suggested by @khmakoto, worked well to resolve this issue, and made the touch handling match closer to the click handling (which does this behavior as part of the browser handling of click events)

#### Focus areas to test

Button interactions in touch scenarios.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12946)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
+++ b/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
@@ -861,7 +861,7 @@ export class BaseButton extends React.Component<IBaseButtonProps, IBaseButtonSta
   }
 
   private _handleTouchAndPointerEvent() {
-    // Touch and pointer events don't focus the button naturally.
+    // Touch and pointer events don't focus the button naturally, so adding an imperative focus call to guarantee this behavior.
     this.focus();
 
     // If we already have an existing timeeout from a previous touch and pointer event
--- a/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
+++ b/packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
@@ -136,31 +136,31 @@ export class BaseButton extends React.Component<IBaseButtonProps, IBaseButtonSta
 
     this._classNames = getClassNames
       ? getClassNames(
-        theme!,
-        className!,
-        variantClassName!,
-        iconProps && iconProps.className,
-        menuIconProps && menuIconProps.className,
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
```diff
@@ -861,6 +861,9 @@ export class BaseButton extends React.Component<IBaseButtonProps, IBaseButtonSta
   }
 
   private _handleTouchAndPointerEvent() {
+    // Touch and pointer events don't focus the button naturally.
```
khmakoto, line 864:
```suggestion
    // Touch and pointer events don't focus the button naturally, so adding an imperative focus call to guarantee this behavior.
```

ecraig12345:
I [reverted](https://github.com/microsoft/fluentui/commit/d93ac86ba129ec4809c0ca4f6b8d128c85c39666) this change for now because it's causing the Button tests to fail due to an infinite loop, which blocks all other builds. I'm not sure why the status checks here show the tests passing, because [this most recent PR build](https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=70978&view=results) fails. 

@Lego6245 and/or @khmakoto you'll need to investigate the failure and resubmit the change with a fix.

PR #12952, opened 2020-04-30
ContextualMenu: add aria-selected to checked menu items

FrancisMengx:
#### Pull request checklist

#### Description of changes
In contextmenubutton drop down, aria-checked is not a valid aria property for announciator to announce the checked state. Change aria-checked to aria-selected

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12952)

### DIFF
--- a/change/office-ui-fabric-react-2020-04-30-14-38-07-fix-accessibility-issues.json
+++ b/change/office-ui-fabric-react-2020-04-30-14-38-07-fix-accessibility-issues.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "fix accessibility issue",
+  "packageName": "office-ui-fabric-react",
+  "email": "xiameng@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-04-30T21:38:07.855Z"
+}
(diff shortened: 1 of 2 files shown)


dzearing:
@joschect and @jurokapsiar 

Here's a codepen to illustrate a checked item in a menu:

https://codepen.io/dzearing/pen/QWjqOQp

* The role is `menuitemcheckbox`
* The `aria-checked` value is set to `true`
* There is no `aria-selected`

In reading this spec:

https://www.w3.org/TR/wai-aria-1.1/#menuitemcheckbox

This looks correct. The spec does not specify setting `aria-selected` at all.

FrancisMengx:
I think the issue is that some times the role can be ```menuitem``` .
Just made a change when role is ```menuitemcheckbox``` then use ```aria-checked``` and when role is ```menuitem``` use ```aria-selected```.

dzearing:
Seems like you may not have merged the change you're mentioning.

FrancisMengx:
@dzearing  Here it is.

FrancisMengx:
@joschect Gentle ping. Do you have any concerns on this PR?

ecraig12345:
@joschect Can you review this?

PR #12954, opened 2020-04-30
@fluentui/react-stylesheet fixes

dzearing:
This change updates the registration of stylesheets to be more synchronous.

Some less than ideal things about this:

* load order is dependent on react, rather than imports
* styles for each component create 1 style element each rather than batching.

Batching can be controlled at the composed component layer, where we will know slots and their stylesheets at compose time and can batch groups of things in advance.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12954)


packages/react-stylesheets/src/StylesheetContext.tsx
```diff
@@ -3,38 +3,44 @@ export type HTMLDirection = 'rtl' | 'ltr' | 'auto';
 
 export interface StylesheetContextType {
   target: Document | undefined;
-  registerStyles: (stylesheets: string[], context: StylesheetContextType) => void;
+  registerStyles: (stylesheets: undefined | string | string[], context: StylesheetContextType) => void;
```
khmakoto, line 6: Should we maybe switch the positions of the `stylesheets` and the `context` so that we can pass context without passing undefined?

```suggestion
  registerStyles: (context: StylesheetContextType, stylesheets?: string | string[]) => void;
```

dzearing: I think it's good right now. `registerStyles` purpose is to register styles idempotently - that is, you can call it over and over with the same input and only have it render styles once. The reasoning for adding `undefined` to the typings was that in some cases, the caller may conditionally have a stylesheet based on user input (e.g. compose). Rather than requiring each usage to check for undefined, I just allow it.

But there is no scenario where calling registerStyles makes sense to pass in without stylesheets argument.

KevinTCoughlin:
@dzearing 

>Batching can be controlled at the composed component layer, where we will know slots and their stylesheets at compose time and can batch groups of things in advance.

Will this be the "default"? I have not profiled your change, but am concerned that non-batched stylesheet append calls will cause more costly style recalculations. SPO currently has a performance problem with this.

PR #12973, opened 2020-05-02
Scripts and dependencies cleanup

ecraig12345:
- Add more complete dependency lists to fluentui packages. This is particularly important when the dependencies are on other packages within the repo to ensure that they build in the right order (this was an issue with docs because it was missing dependencies).
  - Enable [import/no-extraneous-dependencies](https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-extraneous-dependencies.md) for future
- Add types for monorepo utilities
- Fix types and spelling in dangerjs checks

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12973)


layershifter:
Will these rules help to catch this in automated way?

[import/no-extraneous-dependencies](https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-extraneous-dependencies.md)
[import/no-internal-modules](https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-internal-modules.md)

ecraig12345:
Enabling no-extraneous-dependencies now...that would have made this change a lot easier... 🙄 

no-internal-modules seems a bit weird to me in that it considers any folder to be a module (even if that folder doesn't have an index file), so unless it's changed in the future to be more granular it's not so useful.

PR #12974, opened 2020-05-02
Updating Toggle to a function component inside React-Next

czearing:
#### Pull request checklist

#### Description of changes

Converted `Toggle` to a function component inside React-Next

#### Focus areas to test
`Toggle`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12974)


packages/react-next/src/components/Checkbox/__snapshots__/Checkbox.test.tsx.snap
```diff
@@ -217,10 +215,8 @@ exports[`Checkbox renders indeterminate correctly 1`] = `
         @media screen and (-ms-high-contrast: active){.ms-Fabric--isFocusVisible &:focus + label::before {
           outline: 1px solid ActiveBorder;
         }
-    id="checkbox-2"
-    onBlur={[Function]}
```
xugao, line 221: how come Checkbox snapshot has changes?

dzearing: beforeEach resetIds. We should not allow ids to make their way into snapshots, creates fragility.

xugao: but why is `onFocus` and `onBlur` removed?

packages/react-next/src/components/Checkbox/__snapshots__/Checkbox.test.tsx.snap
```diff
@@ -63,10 +63,8 @@ exports[`Checkbox renders checked correctly 1`] = `
         @media screen and (-ms-high-contrast: active){.ms-Fabric--isFocusVisible &:focus + label::before {
           outline: 1px solid ActiveBorder;
         }
-    id="checkbox-1"
-    onBlur={[Function]}
+    id="checkbox-2"
```
dzearing, line 66: Checkbox.tests might need beforeEach resetIds as well. Ids like this should not be random depending on load order

packages/react-next/src/components/Toggle/examples/Toggle.Basic.Example.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Toggle } from 'office-ui-fabric-react/lib/Toggle';
+import { Toggle } from '../Toggle';
```
dzearing, line 2: This should probably be '@fluentui/react-next'

packages/react-next/src/components/Toggle/examples/Toggle.CustomLabel.Example.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Toggle } from 'office-ui-fabric-react/lib/Toggle';
+import { Toggle } from '../Toggle';
```
dzearing, line 2:
```suggestion
import { Toggle } from '@fluentui/react-next';
```

packages/react-next/src/components/Checkbox/__snapshots__/Checkbox.test.tsx.snap
czearing: The old Checkbox would always define onFocus and onBlur.

In the function component refactor, these 2 are only set if the user passes in `inputProps`. So the change is expected.

We don't know how this passed in Michael's refactor though, so maybe a bag merge was involved.

packages/react-next/src/components/Checkbox/__snapshots__/Checkbox.test.tsx.snap
czearing: The old Checkbox would always define onFocus and onBlur.

In the function component refactor, these 2 are only set if the user passes in inputProps. So the change is expected.

We don't know how this passed in Michael's refactor though, so maybe a bag merge was involved.

packages/react-next/src/components/Toggle/Toggle.base.tsx
```diff
@@ -134,67 +118,30 @@ export class ToggleBase extends React.Component<IToggleProps, IToggleState> impl
...
-    }
-  };
-
-  private _noop(): void {
-    /* no-op */
-  }
-}
+  },
+);
+ToggleBase.displayName = 'ToggleBase';
```
xugao, line 147:
```suggestion
ToggleBase.displayName = COMPONENT_NAME;
```

packages/react-next/src/components/Toggle/Toggle.base.tsx
```diff
@@ -111,17 +82,30 @@ export class ToggleBase extends React.Component<IToggleProps, IToggleState> impl
         labelledById = stateTextId;
       }
     }
-
-    const ariaRole = this.props.role ? this.props.role : 'switch';
+    const ariaRole = props.role ? props.role : 'switch';
+    useFocusRects(toggleButton);
+    useComponentRef(props, checked, toggleButton);
+    warnDeprecations(COMPONENT_NAME, props, {
```
xugao, line 88: missing `warnMutuallyExclusive` logic

PR #12979, opened 2020-05-02
chore(DialogFooter): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

Converting `DialogFooter` from class component to functional. Any props are not longer passed to styles function.

Related to #12237

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12979)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Dialog/DialogFooter.tsx
+++ b/packages/fluentui/react-northstar/src/components/Dialog/DialogFooter.tsx
@@ -11,7 +11,7 @@ import {
 } from '../../utils';
 
 import { WithAsProp, withSafeTypeForAs, FluentComponentStaticProps, ProviderContextPrepared } from '../../types';
-import { useTelemetry, getElementType, useUnhandledProps, useStyles } from '@fluentui/react-bindings';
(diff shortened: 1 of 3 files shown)


packages/fluentui/react-northstar/src/components/Dialog/DialogFooter.tsx
```diff
@@ -1,41 +1,67 @@
...
-  static propTypes = {
-    ...commonPropTypes.createCommon(),
-  };
+DialogFooter.displayName = 'DialogFooter';
 
-  renderComponent({ accessibility, ElementType, classes, unhandledProps }): React.ReactNode {
-    const { children, content } = this.props;
+DialogFooter.propTypes = {
+  ...commonPropTypes.createCommon(),
+};
```
layershifter, line 62:
```suggestion
DialogFooter.propTypes = commonPropTypes.createCommon();
```

_nit_

assuncaocharles: I will let it as it is, since makes adding new props easier if needed :)

PR #12984, opened 2020-05-04
fix(Tree): Generate id for string items

assuncaocharles:
#### Pull request checklist

#### Description of changes

We rely on "id" attribute and as string has no "id", then there is the error in the console. This PR generate `id` for string items and pass it down to `TreeItem` since it's a required prop. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12984)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
+++ b/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
@@ -398,7 +398,7 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
 
     const renderItems = (items: TreeItemProps[], level = 1, parent?: string): React.ReactElement[] => {
       return items.reduce((renderedItems: React.ReactElement[], item: TreeItemProps, index: number) => {
-        const { id = _.uniqueId() } = item;
+        const id = typeof item === 'string' ? item : item.id;
         const isSubtree = hasSubtree(item);
         const isSubtreeExpanded = isSubtree && this.isActiveItem(id);
         const isSelectedItem = this.isSelectedItem(item);
--- a/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
+++ b/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
@@ -398,7 +398,7 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
 
     const renderItems = (items: TreeItemProps[], level = 1, parent?: string): React.ReactElement[] => {
       return items.reduce((renderedItems: React.ReactElement[], item: TreeItemProps, index: number) => {
-        const { id } = item;
+        const { id = _.uniqueId() } = item;
         const isSubtree = hasSubtree(item);
         const isSubtreeExpanded = isSubtree && this.isActiveItem(id);
         const isSelectedItem = this.isSelectedItem(item);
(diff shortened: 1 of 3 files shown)


packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -398,7 +398,7 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
 
     const renderItems = (items: TreeItemProps[], level = 1, parent?: string): React.ReactElement[] => {
       return items.reduce((renderedItems: React.ReactElement[], item: TreeItemProps, index: number) => {
-        const { id } = item;
+        const { id = _.uniqueId() } = item;
```
layershifter, line 401: @silviuaavram is it a correct fix? As I remember `id`s should be stable as `key`s in React.

assuncaocharles: In this case id is just internal in the component, the alternative here would be to use `item` as `id` when `typeof item === 'string'`. But I am assuming will make no difference since this id is not being used like key to keep track.

assuncaocharles:
```suggestion
        const id = typeof item === 'string' ? item : item.id;
```

This would be the alternative

layershifter: `activeItemIds` rely on `id`s. If they will change on next render some items will accidentally close

assuncaocharles:
@layershifter Anything else here?

layershifter:
@assuncaocharles is it enough? what about these usages?

https://github.com/microsoft/fluentui/blob/f38654caef2c5f4825dd1f50dc2abe6a35a822e6/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx#L149
https://github.com/microsoft/fluentui/blob/f38654caef2c5f4825dd1f50dc2abe6a35a822e6/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx#L238
https://github.com/microsoft/fluentui/blob/f38654caef2c5f4825dd1f50dc2abe6a35a822e6/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx#L243

assuncaocharles:
> @assuncaocharles is it enough? what about these usages?
> 
> https://github.com/microsoft/fluentui/blob/f38654caef2c5f4825dd1f50dc2abe6a35a822e6/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx#L149
> 
> 
> https://github.com/microsoft/fluentui/blob/f38654caef2c5f4825dd1f50dc2abe6a35a822e6/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx#L238
> 
> 
> https://github.com/microsoft/fluentui/blob/f38654caef2c5f4825dd1f50dc2abe6a35a822e6/packages/fluentui/react-northstar/src/components/Tree/Tree.tsx#L243

These cases are only to selectable items, what will never happen with strings.

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -398,7 +398,7 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
 
     const renderItems = (items: TreeItemProps[], level = 1, parent?: string): React.ReactElement[] => {
       return items.reduce((renderedItems: React.ReactElement[], item: TreeItemProps, index: number) => {
-        const { id } = item;
+        const id = typeof item === 'string' ? _.uniqueId(item) : item.id;
```
layershifter, line 401: But this will generate a unique ID each render. Is it expected?

PR #12985, opened 2020-05-04
chore(getComponentInfo): update getShorthandInfo.ts

layershifter:
#### Pull request checklist

#### Description of changes

In https://github.com/microsoft/fluentui/pull/12783 was introduced `shorthandConfig`, after `.create()` static will be removed we will not be longer able to gather info about `mappedProp`. This PR fixes this.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12985)

### DIFF
--- a/scripts/gulp/plugins/util/getShorthandInfo.ts
+++ b/scripts/gulp/plugins/util/getShorthandInfo.ts
@@ -8,6 +8,24 @@ type ShorthandInfo = Required<Pick<ComponentInfo, 'implementsCreateShorthand' | 
 
 /**
  * Checks that an expression matches signature:
+ * [componentName].shorthandConfig = {[config]}
+ */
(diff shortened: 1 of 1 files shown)


scripts/gulp/plugins/util/getShorthandInfo.ts
```diff
@@ -31,35 +49,43 @@ const isShorthandExpression = (componentName: string, path: NodePath<t.Assignmen
 
 const getShorthandInfo = (componentFile: t.File, componentName: string): ShorthandInfo => {
   let implementsCreateShorthand = false;
-  let mappedShorthandProp: string | undefined;
```
layershifter, line 34: We always fallback to `children`, so we can do this in generated info, too.

PR #12994, opened 2020-05-05
Styling package: fixing tsconfig, removing unneeded array spread in ColorPicker

dzearing:
In going through trying to pull in v7 dependencies into the v0 doc site, we ran into some missing tsconfig cleanup items:

* enable tsconfig rules for no implicit anys and no unused locals
* removing unneeded spread in ColorPicker, which was causing issues with babel

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12994)

### DIFF
--- a/change/@uifabric-styling-2020-05-04-17-29-29-fix-cleanup.json
+++ b/change/@uifabric-styling-2020-05-04-17-29-29-fix-cleanup.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Styling tsconfing: adding noImplicitAnies and noUnusedLocals.",
+  "packageName": "@uifabric/styling",
+  "email": "dzearing@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-05-05T00:29:29.477Z"
+}
--- a/packages/styling/src/styles/DefaultFontStyles.ts
+++ b/packages/styling/src/styles/DefaultFontStyles.ts
@@ -84,7 +84,7 @@ function _getFontBaseUrl(): string {
   let win = getWindow();
 
   // tslint:disable-next-line:no-string-literal no-any
-  let fabricConfig: IFabricConfig = win ? win['FabricConfig'] : undefined;
+  let fabricConfig: IFabricConfig = win ? (win as any).FabricConfig : undefined;
 
   return fabricConfig && fabricConfig.fontBaseUrl !== undefined ? fabricConfig.fontBaseUrl : DefaultBaseUrl;
 }
--- a/packages/styling/src/styles/DefaultFontStyles.ts
+++ b/packages/styling/src/styles/DefaultFontStyles.ts
@@ -84,7 +84,7 @@ function _getFontBaseUrl(): string {
   let win = getWindow();
 
   // tslint:disable-next-line:no-any
-  const fabricConfig: IFabricConfig | undefined = win?.FabricConfig;
+  const fabricConfig: IFabricConfig | undefined = (win as any)?.FabricConfig;
 
   return fabricConfig && fabricConfig.fontBaseUrl !== undefined ? fabricConfig.fontBaseUrl : DefaultBaseUrl;
 }
--- a/packages/styling/src/styles/getGlobalClassNames.test.ts
+++ b/packages/styling/src/styles/getGlobalClassNames.test.ts
@@ -1,6 +1,7 @@
 import { getGlobalClassNames } from './getGlobalClassNames';
 import { createTheme } from './theme';
 import { Stylesheet } from '@uifabric/merge-styles';
+import { ITheme } from '../interfaces/ITheme';
 
 const styleSheet = Stylesheet.getInstance();
 
@@ -19,10 +20,9 @@ describe('getGlobalClassNames', () => {
   });
 
   describe('calls are memoized', () => {
-    // tslint:disable-next-line:no-any
-    let theme: any;
-    // tslint:disable-next-line:no-any
-    let globalClassnames: any;
+    let theme: ITheme;
+    let globalClassnames: { [key: string]: string };
+
     beforeAll(() => {
       theme = createTheme({ disableGlobalClassNames: true });
       globalClassnames = { root: 'ms-Memoized' };
--- a/packages/styling/src/styles/getGlobalClassNames.test.ts
+++ b/packages/styling/src/styles/getGlobalClassNames.test.ts
@@ -19,7 +19,10 @@ describe('getGlobalClassNames', () => {
   });
 
   describe('calls are memoized', () => {
-    let theme, globalClassnames;
+    // tslint:disable-next-line:no-any
+    let theme: any;
+    // tslint:disable-next-line:no-any
+    let globalClassnames: any;
     beforeAll(() => {
       theme = createTheme({ disableGlobalClassNames: true });
       globalClassnames = { root: 'ms-Memoized' };
--- a/packages/styling/src/styles/getGlobalClassNames.ts
+++ b/packages/styling/src/styles/getGlobalClassNames.ts
@@ -15,7 +15,8 @@ const _getGlobalClassNames = memoizeFunction(
     if (disableGlobalClassNames) {
       // disable global classnames
       return Object.keys(classNames).reduce((acc: {}, className: string) => {
-        acc[className] = styleSheet.getClassName(classNames[className]);
+        // tslint:disable-next-line:no-any
+        (acc as any)[className] = styleSheet.getClassName((classNames as any)[className]);
         return acc;
       }, {});
     }
--- a/packages/styling/src/styles/theme.ts
+++ b/packages/styling/src/styles/theme.ts
@@ -97,11 +97,11 @@ function _loadFonts(theme: ITheme): { [name: string]: string } {
   const lines = {};
 
   for (const fontName of Object.keys(theme.fonts)) {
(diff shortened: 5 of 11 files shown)


packages/styling/src/styles/DefaultFontStyles.ts
```diff
@@ -84,7 +84,7 @@ function _getFontBaseUrl(): string {
   let win = getWindow();
 
   // tslint:disable-next-line:no-string-literal no-any
```
ecraig12345, line 86:
```suggestion
  // tslint:disable-next-line:no-any
```

packages/styling/src/styles/theme.ts
```diff
@@ -133,13 +135,21 @@ export function createTheme(theme: IPartialTheme, depComments: boolean = false):
...
-      defaultFontStyles[fontStyle] = merge({}, defaultFontStyles[fontStyle], theme.defaultFontStyle);
+      // tslint:disable-next-line:no-any
+      (defaultFontStyles as any)[fontStyle] = merge({}, (defaultFontStyles as any)[fontStyle], theme.defaultFontStyle);
     }
   }
 
   if (theme.fonts) {
     for (const fontStyle of Object.keys(theme.fonts)) {
-      defaultFontStyles[fontStyle] = merge({}, defaultFontStyles[fontStyle], theme.fonts[fontStyle]);
+      // tslint:disable-next-line:no-any
```
ecraig12345, line 145: Not super important but if there's a series of lines with any I tend to do `// tslint:disable:no-any` at the beginning of the second and `// tslint:enable:no-any` after.

packages/styling/tsconfig.json
```diff
@@ -1,25 +1,23 @@
 {
   "compilerOptions": {
-    "baseUrl": ".",
-    "outDir": "lib",
     "target": "es5",
+    "outDir": "lib",
     "module": "commonjs",
+    "lib": ["es5", "es2015.promise", "dom"],
```
ecraig12345, line 6: Change this back? 
```suggestion
    "lib": ["es5", "dom"],
```

packages/styling/tsconfig.json
```diff
@@ -1,25 +1,23 @@
 {
   "compilerOptions": {
```
ecraig12345, line 2: Why did you get rid of baseUrl? IIRC it may still be needed for certain things.

packages/styling/tsconfig.json
```diff
@@ -1,25 +1,23 @@
...
+    "moduleResolution": "node",
     "preserveConstEnums": true,
-    "lib": ["es5", "dom"],
+    "skipLibCheck": true,
     "typeRoots": ["../../node_modules/@types", "../../typings"],
-    "types": ["jest", "webpack-env", "custom-global"],
-    "paths": {
-      "@uifabric/styling": ["./src"]
-    }
+    "types": ["jest", "custom-global", "node"]
```
ecraig12345, line 20: Revert this?

packages/styling/tsconfig.json
```diff
@@ -1,25 +1,23 @@
...
-    "moduleResolution": "node",
-    "noImplicitAny": false,
+    "noImplicitAny": true,
+    "noUnusedLocals": true,
     "strictNullChecks": true,
     "forceConsistentCasingInFileNames": true,
+    "moduleResolution": "node",
     "preserveConstEnums": true,
-    "lib": ["es5", "dom"],
+    "skipLibCheck": true,
```
ecraig12345, line 18: Revert this if it's not currently needed for this package--it's a bit of a sledgehammer which is probably preferable to avoid

packages/styling/src/styles/DefaultFontStyles.ts
dzearing: there's no reason to use a string literal here. it would still need the `any` type cast.

ecraig12345: Right so you should remove the lint rule override relating to string literals.

packages/styling/src/styles/theme.ts
dzearing: my logic would be: lean towards single line disables unless there's a solid reason. it's easy to forget disabling, it's a little more work for a reader as it causes them to scan for the disable to make sure it wasn't forgotten. Plus writing it redundantly seems like a reasonable penalty as we should be avoiding them anyways.

packages/styling/tsconfig.json
dzearing: We don't want "node" typings.

packages/styling/tsconfig.json
dzearing: added back!

change/@uifabric-styling-2020-05-04-17-29-29-fix-cleanup.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Styling tsconfing: adding noImplicitAnies and noUnusedLocals.",
```
ecraig12345, line 3: Not very important but it was bothering me 😆 
```suggestion
  "comment": "Styling tsconfig: adding noImplicitAny and noUnusedLocals.",
```

packages/styling/src/styles/DefaultFontStyles.ts
```diff
@@ -84,7 +84,7 @@ function _getFontBaseUrl(): string {
...
-  let fabricConfig: IFabricConfig = win ? win['FabricConfig'] : undefined;
+  let fabricConfig: IFabricConfig = win ? (win as any).FabricConfig : undefined;
```
ecraig12345, line 87: Also not super important but this can use `?.` now
```suggestion
  let fabricConfig: IFabricConfig = win?.FabricConfig;
```

packages/styling/src/styles/getGlobalClassNames.test.ts
```diff
@@ -19,7 +19,10 @@ describe('getGlobalClassNames', () => {
   });
 
   describe('calls are memoized', () => {
-    let theme, globalClassnames;
+    // tslint:disable-next-line:no-any
+    let theme: any;
```
ecraig12345, line 23: Add the correct types here instead?

packages/styling/src/styles/theme.ts
```diff
@@ -97,15 +97,17 @@ function _loadFonts(theme: ITheme): { [name: string]: string } {
   const lines = {};
 
   for (const fontName of Object.keys(theme.fonts)) {
-    const font = theme.fonts[fontName];
+    // tslint:disable-next-line:no-any
+    const font = (theme.fonts as any)[fontName];
```
ecraig12345, line 101: Really this should be:
```suggestion
    const font = theme.fonts[fontName as keyof IFontStyles];
```

packages/styling/src/styles/theme.ts
```diff
@@ -97,15 +97,17 @@ function _loadFonts(theme: ITheme): { [name: string]: string } {
...
     for (const propName of Object.keys(font)) {
       const name = fontName + propName.charAt(0).toUpperCase() + propName.slice(1);
       let value = font[propName];
       if (propName === 'fontSize' && typeof value === 'number') {
         // if it's a number, convert it to px by default like our theming system does
         value = value + 'px';
       }
-      lines[name] = value;
+      // tslint:disable-next-line:no-any
+      (lines as any)[name] = value;
```
ecraig12345, line 110: Fix the type of lines rather than adding a cast to any. Should be `const lines: { [name: string]: string } = {}`

packages/styling/src/styles/getGlobalClassNames.test.ts
dzearing: resolved

packages/styling/src/styles/theme.ts
dzearing: fixed.

packages/styling/src/styles/DefaultFontStyles.ts
dzearing: oh doh! I misread :)

packages/styling/src/styles/theme.ts
dzearing: fixed

packages/styling/src/styles/getGlobalClassNames.ts
```diff
@@ -15,7 +15,8 @@ const _getGlobalClassNames = memoizeFunction(
     if (disableGlobalClassNames) {
       // disable global classnames
       return Object.keys(classNames).reduce((acc: {}, className: string) => {
-        acc[className] = styleSheet.getClassName(classNames[className]);
+        // tslint:disable-next-line:no-any
+        (acc as any)[className] = styleSheet.getClassName((classNames as any)[className]);
         return acc;
       }, {});
```
ecraig12345, line 21: Ideally this should be `{} as { [key: string]: string }` plus up above,

ecraig12345, line 21: Ideally this should be:
```
      return (Object.keys(classNames) as (keyof T)[]).reduce((acc, className) => {
        acc[className] = styleSheet.getClassName(classNames[className]);
        return acc;
      }, {} as Partial<GlobalClassNames<T>>);
```

ecraig12345: never mind I fixed this

PR #12996, opened 2020-05-05
Adding universal ThemeProvider package, initial checkin.

dzearing:
This PR introduces a universal generalized ThemeProvider package. It currently does 2 things:

* Provides css variables
* Registers stylesheets defined in a theme.

See the `README.md` for usage.

Includes:

* unit tests
* basic storybook stories
* api extractor
* snapshots
* documentation

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12996)


packages/react-theme-provider/README.md
```diff
@@ -0,0 +1,66 @@
+# @fluentui/react-theme-provider
+
+**React theming component and hook for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
+
+# Installation
```
ecraig12345, line 5: nit, these should be h2 (only one h1 per document)

packages/react-theme-provider/README.md
```diff
@@ -0,0 +1,66 @@
...
+yarn add @fluentui/react-theme-provider
+```
+
+# Example usage
+
+First, ensure you use an existing theme, or create your own. Example:
+
+```js
+/**
+ * Note: Themes are partial by default. They will become fully qualified
```
ecraig12345, line 17: Unfinished comment?

packages/react-theme-provider/etc/react-theme-provider.api.md
```diff
@@ -0,0 +1,43 @@
...
+        site: {
+            body: ThemePlateSet;
+            accent: ThemePlateSet;
+            neutral: ThemePlateSet;
+        };
+        [key: string]: TokenSetType;
+    };
+}
+
+// Warning: (ae-forgotten-export) The symbol "ThemeProviderProps" needs to be exported by the entry point index.d.ts
```
ecraig12345, line 30: Should probably fix this and the similar warnings by exporting all appropriate things

packages/react-theme-provider/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "@types/jest": "~24.9.0",
+    "@types/react": "16.8.25",
+    "@types/react-dom": "16.8.4",
+    "@types/react-test-renderer": "^16.0.0",
+    "@types/webpack-env": "1.15.1",
+    "@uifabric/build": "^7.0.0",
+    "@uifabric/tslint-rules": "^7.2.0",
+    "enzyme": "~3.10.0",
+    "enzyme-adapter-react-16": "^1.15.0",
+    "es6-weak-map": "^2.0.2",
```
ecraig12345, line 43: Same comment as with es6-promise types

packages/react-theme-provider/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "start-test": "just-scripts jest-watch",
+    "test": "just-scripts test",
+    "update-api": "just-scripts update-api",
+    "update-snapshots": "just-scripts jest -u"
+  },
+  "devDependencies": {
+    "@types/classnames": "2.2.10",
+    "@types/enzyme": "3.10.3",
+    "@types/enzyme-adapter-react-16": "1.0.3",
+    "@types/es6-promise": "0.0.32",
```
ecraig12345, line 33: Not sure we actually need this (should probably remove it from other packages too unless there's some good reason for it being there)

packages/react-theme-provider/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "@uifabric/build": "^7.0.0",
+    "@uifabric/tslint-rules": "^7.2.0",
+    "enzyme": "~3.10.0",
+    "enzyme-adapter-react-16": "^1.15.0",
+    "es6-weak-map": "^2.0.2",
+    "react": "16.8.6",
+    "react-app-polyfill": "~1.0.1",
+    "react-dom": "16.8.6",
+    "react-test-renderer": "^16.3.0",
+    "typescript-plugin-css-modules": "^2.2.0"
```
ecraig12345, line 48: Remove

packages/react-theme-provider/src/ThemeContext.tsx
```diff
@@ -0,0 +1,3 @@
+import * as React from 'react';
+
+export const ThemeContext = React.createContext({});
```
ecraig12345, line 3: Add a type?

packages/react-theme-provider/tslint.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": ["@uifabric/tslint-rules"],
+  "rules": {
+    "interface-name": false,
```
ecraig12345, line 4:
```suggestion
    "interface-name": [true, "never-prefix"]
```

packages/react-theme-provider/tslint.json
```diff
@@ -0,0 +1,8 @@
...
+    "jsx-ban-props": false,
+    "jsx-no-lambda": false
```
ecraig12345, line 6: Why disable this?

typings/custom-global/index.d.ts
```diff
@@ -1,5 +1,13 @@
 /// <reference types="webpack-env" />
 
+/**
+ * Generic typings for sass files.
+ */
+declare module '*.scss' {
```
ecraig12345, line 6: Downside of this is that it removes type-safety for styles that have already been built...

packages/react-theme-provider/src/ThemeProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+    // Merge the theme only when parent theme or props theme mutates.
+    const fullTheme = React.useMemo<ThemePrepared>(() => createTheme(parentTheme, theme), [parentTheme, theme]);
+
+    // Generate the inline style object only when merged theme mutates.
+    const inlineStyle = React.useMemo<React.CSSProperties>(
+      () => tokensToStyleObject(fullTheme.tokens, undefined, { ...style }),
+      [fullTheme, style],
+    );
+
+    console.log(theme, fullTheme.tokens, inlineStyle);
```
xugao, line 57: remove

packages/react-theme-provider/src/tokensToStyleObject.ts
```diff
@@ -0,0 +1,26 @@
...
+export const tokensToStyleObject = (
+  tokens: {
+    [key: string]: TokenSetType;
+  },
+  prefix?: string,
+  style: {
+    [key: string]: string | number | undefined;
+  } = {},
+) => {
+  console.log('tokens', tokens, prefix, style);
```
xugao, line 12: remove

packages/react-theme-provider/src/tokensToStyleObject.ts
```diff
@@ -0,0 +1,26 @@
+import { TokenSetType } from './ThemeProvider';
+// tslint:disable-next-line:no-any
```
xugao, line 2: can remove

packages/react-theme-provider/src/tokensToStyleObject.ts
```diff
@@ -0,0 +1,26 @@
...
+  } = {},
+) => {
+  console.log('tokens', tokens, prefix, style);
+  for (const name in tokens) {
+    if (tokens.hasOwnProperty(name)) {
+      const varName = prefix ? `${prefix}${name === 'default' ? '' : '-' + name}` : `--${name}`;
+      const varValue = tokens[name];
+      if (varValue && typeof varValue === 'object') {
+        // tslint:disable-next-line:no-any
+        tokensToStyleObject(varValue as any, varName, style);
```
xugao, line 19: why is casting `any` needed here?

packages/react-theme-provider/src/createTheme.tsx
```diff
@@ -0,0 +1,23 @@
+import { Theme, ThemePrepared } from './types';
```
xugao, line 1: file type should be `ts`, not `tsx`

change/@fluentui-react-2020-05-05-09-14-40-feat-react-theme-provider.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "react-theme-provider: Adding a universal ThemeProvider package which will provide universal css variable based themeing to converged components.",
```
xugao, line 3: should be type none and pls fix comment to describe change correctly

packages/react-theme-provider/README.md
```diff
@@ -0,0 +1,66 @@
...
+import { theme } from './theme';
+
+export const App = () => (
+  <ThemeProvider theme={theme}>
+    <>...app</>
+  </ThemeProvider>
+);
+```
+
+## Applying the theme to the body element
```
xugao, line 52: 1. this is not supported in your change
2. why is this feature needed?

packages/react-theme-provider/src/ThemeProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+    // Pull contextual theme.
+    const parentTheme = useTheme();
+
+    // Merge the theme only when parent theme or props theme mutates.
+    const fullTheme = React.useMemo<ThemePrepared>(() => createTheme(parentTheme, theme), [parentTheme, theme]);
+
+    // Generate the inline style object only when merged theme mutates.
+    const inlineStyle = React.useMemo<React.CSSProperties>(
+      () => tokensToStyleObject(fullTheme.tokens, undefined, { ...style }),
+      [fullTheme, style],
```
xugao, line 54:
```suggestion
      [fullTheme.tokens, style],
```

packages/react-theme-provider/src/createTheme.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { createTheme } from './createTheme';
```
khmakoto, line 1: File type can be ts instead of tsx.

packages/react-theme-provider/src/types.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * A set of states for each color plate to use.
+ */
+export type ThemeStateSet =
+  | Partial<{
+      default: string;
+      hover: string;
+      active: string;
```
khmakoto, line 8:
```suggestion
      hovered: string;
      pressed: string;
```

packages/react-theme-provider/src/types.ts
```diff
@@ -0,0 +1,53 @@
...
+export type RecursivePartial<T> = {
+  [P in keyof T]?: T[P] extends (infer U)[]
+    ? RecursivePartial<U>[]
+    : T[P] extends object
+    ? RecursivePartial<T[P]>
+    : T[P];
+};
+
+/**
+ * A basic prepared (fully qualitied) theme object.
```
khmakoto, line 39:
```suggestion
 * A basic prepared (fully qualified) theme object.
```

change/@fluentui-react-2020-05-05-09-14-40-feat-react-theme-provider.json
dzearing: If I do none, it will not publish an initial publish.

packages/react-theme-provider/README.md
dzearing: I've removed the comment. However I believe it will be needed. Otherwise you won't have a way to apply the theme to the body.

packages/react-theme-provider/src/ThemeProvider.tsx
dzearing: I don't think we need the extra bytes here. Comparing the object which I'm creating a few lines up should be enough, no?

packages/react-theme-provider/README.md
dzearing: done!

packages/react-theme-provider/README.md
dzearing: Woops. :)

packages/react-theme-provider/etc/react-theme-provider.api.md
dzearing: Fixed up some of the typings and made sure they're all exported.

packages/react-theme-provider/package.json
dzearing: removed, checked through others as well.

packages/react-theme-provider/src/ThemeContext.tsx
dzearing: Fixed!

packages/react-theme-provider/tslint.json
dzearing: Removed all of these.

dzearing: (except the interface one which i fixed :)

typings/custom-global/index.d.ts
dzearing: I think we need a better way to get type safety than to rely on the build to produce the typings. Right now you get red squiggles without a build.

packages/react-theme-provider/src/ThemeProvider.tsx
dzearing: snuck in there. good catches.

change/@fluentui-react-2020-05-05-09-14-40-feat-react-theme-provider.json
xugao: this is changefile for `fluent/react` package. in fact you marked the themeprovider package to be `private`. i thought that was intentional?

dzearing: huh... that's odd. good catch; digging.

dzearing: Straightened out. :) Thanks for calling that out! Great catch. I didn't intend to privatize the package, and would rather just start publishing it off the bat. Enables codesandbox experiments.

packages/react-theme-provider/README.md
```diff
@@ -0,0 +1,66 @@
...
+<ThemeProvider theme={theme} applyToBody>
+  ...
+</ThemeProvider>
+```
+
+## `ThemeProvider` api
+
+| Prop name | Description       |
+| --------- | ----------------- |
+| theme     | The partial theme |
```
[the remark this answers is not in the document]
dzearing: Removed.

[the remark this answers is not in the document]
dzearing: (for now!)

packages/react-theme-provider/src/types.ts
dzearing: I've updated them.

packages/react-theme-provider/src/ThemeProvider.scss
```diff
@@ -0,0 +1,17 @@
...
+ */
+.root {
+  background: var(--body-fill);
+  color: var(--body-text);
+  font-family: var(--body-fontFamily);
+  font-weight: var(--body-fontWeight);
+  font-size: var(--body-fontSize);
+  line-height: var(--body-fontLineHeight);
+
+  > * {
```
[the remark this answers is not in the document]
dzearing: They *are* to direct children (The `>` character implies direct children.)

packages/react-theme-provider/src/createTheme.test.tsx
dzearing: yup got em all

packages/react-theme-provider/src/types.ts
dzearing: updated!

packages/react-theme-provider/src/types.ts
dzearing: fixed

yarn.lock
```diff
@@ -21472,6 +21625,22 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=
 
+typescript-plugin-css-modules@^2.2.0:
```
xugao, line 21628: are you planning to use this for the scss typing? it doesn't look like it's used anywhere yet in this PR

packages/react-theme-provider/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "just": "just-scripts",
+    "lint": "just-scripts lint",
+    "start": "just-scripts dev:storybook",
+    "start-test": "just-scripts jest-watch",
+    "test": "just-scripts test",
+    "update-api": "just-scripts update-api",
+    "update-snapshots": "just-scripts jest -u"
+  },
+  "devDependencies": {
+    "@types/classnames": "2.2.10",
```
xugao, line 29:
```suggestion
    "@types/classnames": "^2.2.9",
```

yarn.lock
dzearing: hmm it should be removed.

packages/react-theme-provider/package.json
dzearing: nice!

packages/react-theme-provider/src/tokensToStyleObject.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  it('can convert nested tokens', () => {
+    expect(
+      tokensToStyleObject({
+        body: {
+          fill: 'red',
+          text: 'blue',
+        },
+      }),
+    ).toEqual({
+      '--body-fill': 'red',
```
xugao, line 23: I am a bit lean towards name the css variable camelCase (`--bodyFill` instead of `--body-fill`) bc these variable names cannot be minified.

dzearing: Let's chat about this;

packages/react-theme-provider/src/ThemeProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { mergeThemes } from './mergeThemes';
+import { useTheme } from './useTheme';
+import * as classes from './ThemeProvider.scss';
+
+/**
+ * Props for the ThemeProvider component.
+ */
+export interface ThemeProviderProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * The theme provided by the user.
```
khmakoto, line 16: nit: Given our own guidelines for documenting props shouldn't this be something like:

```suggestion
   * Defines the theme provided by the user.
```

packages/react-theme-provider/src/types.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Bad: "hover", Good: "hovered"
+ *
+ * Additional considerations:
+ *
+ * The term "active" in css means that the keyboard or mouse button
+ * which activates the component is pressed down. It is however ambiguous
+ * with a focused state, as the HTML object model refers to the focused
+ * element as the "activeElement". To resolve ambiguity and to be more
+ * compatible with other platforms reusing token names, we'll will snap
```
khmakoto, line 18:
```suggestion
 * compatible with other platforms reusing token names, we have decided to snap
```

packages/react-theme-provider/src/ThemeProvider.scss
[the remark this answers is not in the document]
khmakoto: Sorry, I don't know why I got confused.

packages/react-theme-provider/src/ThemeProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const fullTheme = React.useMemo<ThemePrepared>(() => mergeThemes(parentTheme, theme), [parentTheme, theme]);
+
+    // Generate the inline style object only when merged theme mutates.
+    const inlineStyle = React.useMemo<React.CSSProperties>(
+      () => tokensToStyleObject(fullTheme.tokens, undefined, { ...style }),
+      [fullTheme, style],
+    );
+
+    // Register stylesheets as needed.
+    useStylesheet(fullTheme.stylesheets);
```
KevinTCoughlin, line 42: Is this batched and performed in an rAF to avoid forced reflow?

packages/react-theme-provider/src/ThemeProvider.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+  tokens: {
+    body: {
+      fill: 'black',
+      text: 'white',
+    },
+  },
+};
+
+const themeWithStylesheets: Theme = {
+  stylesheets: ['.foo { font-family: var(--body-customFont); }'],
```
KevinTCoughlin, line 27: Is any validation performed on this API? It seems easy to provide an incorrect selector rule set.

dzearing: Certainly not at runtime, which would be a perf hit. Expectation is at build time. This is similar to `loadStyles`, with the exception that styles would be deferred to be registered until needed rather than on module load. This means

* less css rendering at module load time :+1:
* ability to project rules to child window :+1:
* load order will not be determined by module import order, but rather on component render order :-1:(but selectors should not depend on load order to resolve specificity; that's fragile)

```tsx
// Up to the css compiler build step to fail if you have bad css.
import { stylesheet: fileStyles } from './file.scss';

const theme: Theme = {
  stylesheets: [ fileStyles ]
};
```

packages/react-theme-provider/src/ThemeProvider.tsx
dzearing: Currently, no. I tried doing that with `useLayoutEvent` and saw unstyled markup for a second which seems very very bad.

Would RAF avoid this? I don't want users to see unstyled DOM for a millisecond.

packages/react-theme-provider/src/ThemeProvider.stories.tsx
KevinTCoughlin: What happens if a developer defines `stylesheets: ['foo { bad-selector` ]`? Is the stylesheet appended to HEAD? They may not be aware of the recommended pattern you mention above.

Appending styles at run-time on-demand will cause the browser to perform style recalculations. I asked a question related in a separate comment re: rAF and batching.

packages/react-theme-provider/src/ThemeProvider.tsx
KevinTCoughlin: Have you profiled or benchmarked to see the cost of style recalculation with the proposed approach? Style recalculations can be costly for complex apps with complex selectors.

xugao: Note that some browsers are still firing rAF after layout (IE is one of them) yet: https://nolanlawson.com/2018/09/25/accurately-measuring-layout-on-the-web/

dzearing: This PR doesn't really register styles. It uses `useStylesheet`, which does the work in the same way load-themed-styles does, with the exception it does it on initial component render, rather than module load.

And even then, stylesheets designed this way can be externalized into a css file to preload them all in one swipe if needed. There are zero runtime class calculations here.

packages/react-theme-provider/src/ThemeProvider.stories.tsx
dzearing: The same thing which can happen with `load-themed-styles` or webpack's css loader. The intent of the stylesheets in theme is because there is a need for them to be packaged with the theme. There are a million things which can go wrong with css in general but it's not the duty of this utility to compile and validate css.

packages/react-theme-provider/src/ThemeProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const ThemeProvider = React.forwardRef<HTMLDivElement, ThemeProviderProps>(
+  (
+    { theme, style, className, ...rest }: React.PropsWithChildren<ThemeProviderProps>,
+    ref: React.Ref<HTMLDivElement>,
+  ) => {
+    // Pull contextual theme.
+    const parentTheme = useTheme();
+
+    // Merge the theme only when parent theme or props theme mutates.
+    const fullTheme = React.useMemo<ThemePrepared>(() => mergeThemes(parentTheme, theme), [parentTheme, theme]);
```
layershifter, line 33: As `theme` is a prop it can be inline object: `<ThemeProvider theme={{ foo: 'bar' }} />`. It means that all consumers will receive updates on each render as `fullTheme` is passed to React context.

packages/react-theme-provider/src/ThemeProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+  theme?: Theme;
+}
+
+/**
+ * ThemeProvider, used for providing css variables and registering stylesheets.
+ */
+export const ThemeProvider = React.forwardRef<HTMLDivElement, ThemeProviderProps>(
+  (
+    { theme, style, className, ...rest }: React.PropsWithChildren<ThemeProviderProps>,
+    ref: React.Ref<HTMLDivElement>,
```
layershifter, line 27: There is no need to specify these types obviously

PR #12997, opened 2020-05-05
fix(FocusZone): default behavior of FocusZone in v0

jurokapsiar:
#### Description of changes
- fix `pagingSupportDisabled` functionality in v0 and v7
- use `pagingSupportDisabled` in Chat and ChatMessage behaviors
- use `preventFocusRestoration` by default for v0

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12997)

### DIFF
--- a/packages/fluentui/react-bindings/src/hooks/useAccessibility.ts
+++ b/packages/fluentui/react-bindings/src/hooks/useAccessibility.ts
@@ -81,8 +81,6 @@ const useAccessibility = <Props>(behavior: Accessibility<Props>, options: UseAcc
       }
 
       return React.createElement(FocusZone, {
-        // maintain behavior of focus zone in v7 behaviors
-        preventFocusRestoration: true,
         ...definition.focusZone.props,
         ...child.props,
         as: child.type,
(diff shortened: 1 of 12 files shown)


packages/fluentui/react-bindings/test/hooks/useAccessibility-test.tsx
```diff
@@ -322,6 +322,18 @@ describe('useAccessibility', () => {
       );
     });
 
+    it('applies default props for focuszonr', () => {
```
layershifter, line 325:
```suggestion
    it('applies default props for FocusZone', () => {
```

packages/fluentui/react-bindings/src/hooks/useAccessibility.ts
```diff
@@ -81,6 +81,8 @@ const useAccessibility = <Props>(behavior: Accessibility<Props>, options: UseAcc
       }
 
       return React.createElement(FocusZone, {
+        // maintain behavior of focus zone in v7 behaviors
+        preventFocusRestoration: true,
```
layershifter, line 85: Should be this done in `getAccessibility()`?

```tsx
  if (definition.focusZone) {
    definition.focusZone.props = {
      // maintain behavior of focus zone in v7 behaviors
      preventFocusRestoration: true,
      ...definition.focusZone.props,
    };
  }
```

PR #13000, opened 2020-05-05
docs(Card): Add expandable card example

pompomon:
#### Pull request checklist

#### Description of changes
Expandable card example with `compose` and custom styled elements

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13000)

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleExpandable.tsx
+++ b/packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleExpandable.tsx
@@ -13,6 +13,7 @@ import {
   Text,
   ThemeInput,
 } from '@fluentui/react-northstar';
+import { useBooleanKnob } from '@fluentui/docs-components';
 import * as React from 'react';
 
 // Compose custom styled Box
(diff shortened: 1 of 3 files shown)


packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleExpandable.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const expandCardHandler = () => {
+    setExpandCard(true);
+  };
+
+  const shrinkCardHandler = () => {
+    setExpandCard(false);
+  };
+
+  return (
+    <>
```
layershifter, line 87: Useless fragment?

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleExpandable.tsx
```diff
@@ -0,0 +1,126 @@
...
+  componentVariables,
+};
+
+const CardExampleExpandable = () => {
+  const [expandCard, setExpandCard] = React.useState(false);
+  const expandCardHandler = () => {
+    setExpandCard(true);
+  };
+
+  const shrinkCardHandler = () => {
```
layershifter, line 82: I don't think that there is any values in these callbacks. Let's remove them and inline, we can also use knobs, BTW

packages/fluentui/docs/src/examples/components/Card/Usage/CardExampleExpandable.tsx
```diff
@@ -0,0 +1,126 @@
...
+};
+
+type ExpandableBoxStylesProps = ExpandableBoxProps;
+
+const ExpandableBox = compose<'button', ExpandableBoxProps, ExpandableBoxStylesProps, BoxProps, {}>(Box, {
+  className: 'ui-expandable-box',
+  displayName: 'ExpandableBox',
+  mapPropsToStylesProps: props => ({ expand: props.expand }),
+  handledProps: ['expand'],
+});
```
layershifter, line 30:
```suggestion
});

ExpandableBox.defualtProps = {
  as: 'button',
}
```

If it's really button

PR #13001, opened 2020-05-05
fix(Provider): attach provider element during render

### LINKED ISSUE #13004: Provider: in Fluent v0
It's related to changes in: microsoft/fluent-ui-react#2192.

We use `React.useLayoutEffect` to insert a DOM node in `useProviderBox` where `PortalInner` children will be mounted (used in `Dialog`, `Popup`, etc.) and this effect occurs later than children effects as React executes inner effects first.

A great example is the check in `FocusTrapZone`:

https://github.com/microsoft/fluentui/blob/9da44e35aff722df21d9047cae53ec115c4c56e4/packages/fluentui/react-bindings/src/FocusZone/FocusTrapZone.tsx#L417-L423

```
// @fluentui/react-bindings/FocusZone/FocusTrapZone.tsx
FocusTrapZone:componentDidMount (check occurs here and throws warning)
// @fluentui/react-northstar/components/Provider/usePortalBox.ts
usePortalBox:appendElement (element will be a document.body only here)
```

### Environment Information

- **Package version(s)**: `@fluentui/react-northstar`

#### Actual behavior:

The usage of `FocusTrapZone` produces a warning in tests:

```
Body element does not contain trap zone element. Please, ensure the trap zone element is placed inside body, so it will work properly.
```

#### Expected behavior:

No warnings.

Requested priority: Normal

layershifter:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13004

#### Description of changes

This PR fixes a bug related to changes in It's related to changes in: https://github.com/microsoft/fluent-ui-react/pull/2192.

We use `React.useLayoutEffect` to insert a DOM node in `useProviderBox` where `PortalInner` children will be mounted (used in `Dialog`, `Popup`, etc.) and this effect occurs later than children effects as React executes inner effects first.

A great example is the check in `FocusTrapZone`:

https://github.com/microsoft/fluentui/blob/9da44e35aff722df21d9047cae53ec115c4c56e4/packages/fluentui/react-bindings/src/FocusZone/FocusTrapZone.tsx#L417-L423

```
// @fluentui/react-bindings/FocusZone/FocusTrapZone.tsx
FocusTrapZone:componentDidMount (check occurs here and throws warning)
// @fluentui/react-northstar/components/Provider/usePortalBox.ts
usePortalBox:appendElement (element will be a document.body only here)
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13001)


packages/fluentui/react-northstar/test/specs/components/Provider/Provider-test.tsx
```diff
@@ -287,4 +289,42 @@ describe('Provider', () => {
...
+  describe('document.body', () => {
+    it('adds an element to document.body', () => {
+      const className = faker.lorem.word();
+      const wrapper = mount(
+        <Provider className={className}>
+          <div />
+        </Provider>,
+      );
+
+      expect(document.querySelector(`.${className}`)).toBeInstanceOf(Element);
```
silviuaavram, line 302: nit: if you are checking if an element is / is not in the document, using `(.not).toBeInTheDocument()` might make the assertion easier to understand.

layershifter: Yeah, this looks much cleaner. Thanks for the suggestion

PR #13008, opened 2020-05-05
refresh tiletype icon url salt to bust thru the can cache.

bigbadcapers:
#### Pull request checklist

- [x] Addresses an existing issue: Fluid icons looking stale.
- [ ] Can u believe I'm talking the lingo fam

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13008)

### DIFF
--- a/packages/file-type-icons/src/initializeFileTypeIcons.tsx
+++ b/packages/file-type-icons/src/initializeFileTypeIcons.tsx
@@ -12,7 +12,7 @@ const ICON_SIZES: number[] = [16, 20, 24, 32, 40, 48, 64, 96];
 // (though most users will see them within a week). To force immediate refresh, append a
 // unique string to the end of the URL. The CDN uses the URL as the cache key, so passing a
 // new URL will cause the CDN to create a new cache key.
-const REFRESH_STRING = '?v6';
+const REFRESH_STRING = '?v1';
 
 export function initializeFileTypeIcons(baseUrl: string = DEFAULT_BASE_URL, options?: Partial<IIconOptions>): void {
   ICON_SIZES.forEach((size: number) => {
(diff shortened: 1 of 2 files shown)


packages/file-type-icons/src/initializeFileTypeIcons.tsx
```diff
@@ -12,7 +12,7 @@ const ICON_SIZES: number[] = [16, 20, 24, 32, 40, 48, 64, 96];
 // (though most users will see them within a week). To force immediate refresh, append a
 // unique string to the end of the URL. The CDN uses the URL as the cache key, so passing a
 // new URL will cause the CDN to create a new cache key.
-const REFRESH_STRING = '?v6';
+const REFRESH_STRING = '?v1';
```
dzearing, line 15: why not v7?

bigbadcapers:
Because I don’t want to get to 10 and add one more character over the wire for every icon rendered on-screen 😃

Also, we never used v1. These salts look reusable because it just needs to not have been used frequently. I hand-tested URLs suffixed with v1 and all the right icons came back.

packages/file-type-icons/src/initializeFileTypeIcons.tsx
bigbadcapers: Because I don’t want to get to 10 and add one more character over the wire for every icon rendered on-screen 😃

Also, we never used v1. These salts look reusable because it just needs to not have been used frequently. I hand-tested URLs suffixed with v1 and all the right icons came back.

Jahnp: I'll trust your judgment on this @bigbadcapers , though agreed it's kinda unergonomic. If the value needs to increment again, and if someone besides you tried to bump it, I could see them reaching for v2. But since versioned deployments are a thing now this shouldn't be needed anymore right?

bigbadcapers: thanks, yeah, i probably set the wrong expectation by using the letter "v" there. in fact. just having ?1 would work just as well. it really just needs to be a random trailing string that hasnt been used recently since the cache does seem to flush over longer periods of time.  We had only ever used v5 before I believe.

PR #13011, opened 2020-05-05
GroupedList: Allowing for expand icon override

bcoard:
#### Pull request checklist

#### Description of changes

This change will allow developers to specify a custom expand icon for Group Headers used in Grouped List and Grouped Details List. If no icon is specified the default, current, icon will be used.

#### Focus areas to test

1. The default icon shouldn't change if no icon override is set.
2. If an icon override is set it should be visible to the end user.
3. If an icon override is set it should rotate as expected, even in RTL.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13011)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.types.ts
+++ b/packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.types.ts
@@ -16,7 +16,7 @@ export interface IGroupHeaderProps extends IGroupDividerProps {
   /** Native props for the GroupHeader expand and collapse button */
   expandButtonProps?: React.HTMLAttributes<HTMLButtonElement>;
 
-  /** Defines the expand icon for group headers. If not set, the default icon will be used */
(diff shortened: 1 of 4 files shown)


packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.types.ts
```diff
@@ -16,6 +16,9 @@ export interface IGroupHeaderProps extends IGroupDividerProps {
   /** Native props for the GroupHeader expand and collapse button */
   expandButtonProps?: React.HTMLAttributes<HTMLButtonElement>;
 
+  /** Defines the expand icon for group headers. If not set, the default icon will be used */
```
khmakoto, line 19:
```suggestion
  /** Defines the name of a custom icon to be used for group headers. If not set, the default icon will be used */
```

PR #13013, opened 2020-05-05
ResizeGroup: Migrate ResizeGroup to function component

MLoughry:
#### Pull request checklist

#### Description of changes

Migrate ResizeGroup to a function component inside `react-next`

#### Focus areas to test

Performance. I had to make a few unorthodox tweaks (stashing state in a ref object to have the most recent values for RAF calls) in order to get the perf similar (in my perception) to the original. Not sure if there are hard numbers to be had around resizing perf.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13013)


MLoughry:
@ecraig12345 @micahgodbolt As best as I can tell from debugging that repro, the issue is that `this._overflowSet` in `CommandBarBase.focus()` is `null`. So, it seems to be an issue along the `CommandBar`/`OverflowSet` boundary, and not due to `ResizeGroup`.

ecraig12345:
No, I'm pretty sure it's due to ResizeGroup's measurement logic making an extra copy (sharing the same ref) for off-screen measurement and disposing it in the wrong order, causing the ref to be corrupted.

MLoughry:
This is from debugging the repro in Storybook with my ResizeGroup changes:
image
`CommandBarBase._overflowSet` is passed as `componentRef` to `OverflowSet`.

ecraig12345:
Right but unless `_overflowSet` itself (not its `current` property) is being unset, which is not what I saw when debugging, the question is why `current` is getting unset--and we suspect that's caused by ResizeGroup not CommandBar.

MLoughry:
It's a ref to the `OverflowSet` component instance. I'm not sure I follow why `ResizeGroup`, a child component, would cause the `OverflowSet` component instance to unset.


-Michael

________________________________

packages/react-next/src/components/ResizeGroup/ResizeGroup.base.tsx
```diff
@@ -308,142 +308,194 @@ const hiddenDivStyles: React.CSSProperties = { position: 'fixed', visibility: 'h
...
-      this.setState({ measureContainer: true });
+  const [stateRef, updateResizeState, remeasure] = useResizeState(props, nextResizeGroupStateProvider, rootRef);
+
+  React.useEffect(() => {
+    if (stateRef.current.renderedData) {
+      hasRenderedContent.current = true;
+      props.dataDidRender?.(stateRef.current.renderedData);
     }
-  }
+  });
```
dzearing, line 395: isn't this dependent on dataDidRender? I tjhink you need a deps list.

packages/react-next/src/components/ResizeGroup/ResizeGroup.base.tsx
```diff
@@ -308,142 +308,194 @@ const hiddenDivStyles: React.CSSProperties = { position: 'fixed', visibility: 'h
...
         containerDimension,
       );
 
-      if (nextState) {
-        this.setState(nextState);
-      }
+      updateResizeState(nextState);
     });
-  }
+  });
```
dzearing, line 422: Again no deps, and also I see no symmetrical dispose here.

packages/react-next/src/components/ResizeGroup/ResizeGroup.base.tsx
MLoughry: For `useEffect`, the deps list controls how frequently the effect is run, not memoization as in `useCallback`. I’m not sure we want to prevent this from running every render.

MLoughry: For `useEffect`, the deps list controls how frequently the effect is run, not memoization as in `useCallback`. I’m not sure we want to prevent this from running every render.

MLoughry: For `useEffect`, the deps list controls how frequently the effect is run, not memoization as in `useCallback`. I’m not sure we want to prevent this from running every render.

packages/react-next/src/components/ResizeGroup/ResizeGroup.base.tsx
MLoughry: What would there be to dispose? The `async` instance is disposed in `useAsync`.

MLoughry:
@micahgodbolt Can you review this as code owner?

micahgodbolt:
This is a great read, seeing how hooks can be leveraged to do some pretty complex work.

PR #13014, opened 2020-05-05
Checkbox: Convert base styles to scss

xugao:
#### Pull request checklist

#### Description of changes
**This change is meant to not introduce any breaking changes. If you spot any scenarios that I might be breaking or worths verifying, please point it out.**

Convert VNext Checkbox base styles from css-in-js (`getStyles`) to scss (`getStaticStyles`).
For the styles which previous vary based on props, I added modifier classes (e.g. `.disabled`, `.required`) onto each slot (e.g. `root`, `checkbox`, `label`)

In order to let styles overrides (e.g. styles prop and styles passed from `Customizer` `scopedSettings`) to take higher priority than base component styles, I give these styles higher specificity during `merge-styles` by repeating selectors 5 times (this is subject to tweak. I do think some components need larger `specificityMultiplier` than others.). For example, here is how the css like when `styles` override is passed to `Checkbox.label`:
image
`.label-50` here is generated by `merge-styles` and selector is repeated to bump its specificity.
**Note**: I don't love the solution at all but I can't think of a better way to solve this.
 
**What's pending:**
- `_themeCssVariable` is temporary to unblock developing the styles in scss. Need to connect this change with `ThemeProvider` David wrote.  
- RTL is not processed by current scss pipeline. 

#### Focus areas to test
1. Added screener test for new Checkbox by copying tests for current Checkbox, the only styles regressed is RTL related tests
2. Added perf test for new Checkbox
3. Tested styles override still working using styles prop and `Customizer`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13014)


packages/react-next/src/components/Checkbox/Checkbox.scss
```diff
@@ -0,0 +1,253 @@
+@import '../../common/themeCssVariables';
```
dzearing, line 1: I think this pattern would write the variables to every individual component, leading to bulky downloads.

A better approach would be to define sass variables for the names themselves and import that.

_ThemeVariables.scss:

```sass
@semanticColorsInputBorderHovered: --semanticColorsInputBorderHovered;
@semanticColorsInputBorderHovered: --semanticColorsInputBorderHovered;
```

Checkbox.scss:

```sass
background:  var(@semanticColorsInputBorderHovered);

// Variable is not declared, should break build.
background:  var(@semanticColorsInputBorderHoveredTypo);
```

packages/react-next/src/components/Checkbox/Checkbox.scss
```diff
@@ -0,0 +1,253 @@
...
+  }
+}
+
+.input {
+  position: absolute;
+  background: none;
+  opacity: 0;
+}
+
+:global(.ms-Fabric--isFocusVisible) {
```
dzearing, line 101: We need to do something better here. 

We actually have this defined as a sass function already. I will dig around for it.

packages/react-next/src/components/Checkbox/Checkbox.scss
```diff
@@ -0,0 +1,253 @@
...
+
+.label {
+  @include fonts-medium;
+
+  display: flex;
+  align-items: flex-start;
+  cursor: pointer;
+  position: relative;
+  user-select: none;
+  text-align: left;
```
dzearing, line 120: is `text-align: left` necessary?

packages/react-next/src/components/Checkbox/Checkbox.scss
xugao: yeah like you said , this is temp to just unblock developing the styles. 
Good call on assign the css var to as sass var. i didn't think of that idea!

packages/react-next/src/components/Checkbox/Checkbox.scss
xugao: thanks, found it

xugao:
Converting to scss does overall increase the bundle size. Currently with my change `Checkbox.scss.js` is 18K but majority of them should go away once we remove `_themeCssVariables` and use `ThemeProvider` to provide them.

One thing I do realize is `var(--theme-variable)` is something will be used all over the place. Perhaps we can figure out ways to minify this

packages/react-next/src/components/Checkbox/Checkbox.scss
```diff
@@ -0,0 +1,261 @@
...
+    @include high-contrast {
+      background: Highlight;
+      border-color: Highlight;
+    }
+  }
+
+  &.disabled {
+    border-color: var($semanticColorsDisabledBodySubtext);
+    @include high-contrast {
+      border-color: InactiveBorder;
```
khmakoto, line 211: `GrayText`?

packages/react-next/src/components/Checkbox/Checkbox.scss
```diff
@@ -0,0 +1,261 @@
...
+    -ms-high-contrast-adjust: none;
+  }
+
+  &.checked {
+    opacity: 1;
+  }
+
+  &.disabled {
+    @include high-contrast {
+      color: InactiveBorder;
```
khmakoto, line 235: `GrayText`?

packages/react-next/src/components/Checkbox/Checkbox.scss
```diff
@@ -0,0 +1,261 @@
...
+    margin-left: 4px;
+  }
+
+  &.disabled {
+    color: var($semanticColorsDisabledText);
+
+    @include high-contrast {
+      // backwards compat for the color of the text when the checkbox was rendered
+      // using a Button.
+      color: InactiveBorder;
```
khmakoto, line 258: `GrayText`?

packages/utilities/src/classNamesFunction.ts
```diff
@@ -75,6 +83,17 @@ export function classNamesFunction<TStyleProps extends {}, TStyleSet extends ISt
 
     const disableCaching = options.disableCaching;
 
+    // If useStaticStyles is true, styleFunctionOrObject returns slot to classname mappings.
+    // If there is also no style overrides, we can skip merge styles completely and
+    // simply return the result from the style funcion
```
dzearing, line 88:
```suggestion
    // simply return the result from the style function.
```

packages/merge-styles/src/styleToClassName.test.ts
```diff
@@ -404,3 +405,72 @@ describe('styleToClassName', () => {
...
+  });
+
+  it('can repeat classname', () => {
+    const className = styleToClassName(options, { background: 'red' });
+
+    expect(className).toEqual('css-0');
+    expect(_stylesheet.getRules()).toEqual('.css-0.css-0{background:red;}');
+  });
+
+  it('can repeat classname when have child selectors', () => {
```
dzearing, line 422:
```suggestion
  it('can repeat classname with child selectors', () => {
```

packages/merge-styles/src/styleToClassName.test.ts
```diff
@@ -404,3 +405,72 @@ describe('styleToClassName', () => {
...
+    styleToClassName(options, {
+      selectors: {
+        '.foo': { background: 'red' },
+      },
+    });
+
+    expect(_stylesheet.getRules()).toEqual('.css-0.css-0 .foo{background:red;}');
+  });
+
+  it('can repeat classname when have child selectors with comma', () => {
```
dzearing, line 432:
```suggestion
  it('can repeat classname with child selectors with comma', () => {
```

packages/merge-styles/src/styleToClassName.test.ts
```diff
@@ -404,3 +405,72 @@ describe('styleToClassName', () => {
...
+    styleToClassName(options, {
+      selectors: {
+        '.foo, .bar': { background: 'red' },
+      },
+    });
+
+    expect(_stylesheet.getRules()).toEqual('.css-0.css-0 .foo{background:red;}.css-0.css-0 .bar{background:red;}');
+  });
+
+  it('can repeat classname when have child selectors with comma with pseudo selectors', () => {
```
dzearing, line 442:
```suggestion
  it('can repeat classname fwith child selectors that have a comma with pseudo selectors', () => {
```

packages/merge-styles/src/styleToClassName.test.ts
```diff
@@ -404,3 +405,72 @@ describe('styleToClassName', () => {
...
+    styleToClassName(options, {
+      selectors: {
+        ':hover, :active': { background: 'red' },
+      },
+    });
+
+    expect(_stylesheet.getRules()).toEqual('.css-0.css-0:hover{background:red;}.css-0.css-0:active{background:red;}');
+  });
+
+  it('can repeat classname when have child selectors with comma with @media query', () => {
```
dzearing, line 452:
```suggestion
  it('can repeat classname when child selectors have a comma with @media query', () => {
```

packages/merge-styles/src/styleToClassName.test.ts
```diff
@@ -404,3 +405,72 @@ describe('styleToClassName', () => {
...
+        },
+      },
+    });
+
+    expect(_stylesheet.getRules()).toEqual(
+      '@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none){.css-0.css-0{background:red;}}',
+    );
+  });
+
+  it('do not repeat classname when have global selector', () => {
```
dzearing, line 466:
```suggestion
  it('doesn't repeat classname with a global selector', () => {
```

packages/merge-styles/src/styleToClassName.ts
```diff
@@ -177,6 +177,18 @@ function getKeyForRules(options: IStyleOptions, rules: IRuleSet): string | undef
   return hasProps ? serialized.join('') : undefined;
 }
 
+function repeatString(target: string, count: number): string {
+  if (count <= 0) {
```
dzearing, line 181: `createArray(count), () => target).join('')`

dzearing: maybe the 2nd is more efficient anyway. This will be a hotspot called for every selector.

packages/merge-styles/src/styleToClassName.ts
```diff
@@ -251,12 +269,10 @@ export function applyRegistration(registration: IRegistration): void {
       const rules = rulesToInsert[i + 1];
       if (rules) {
         let selector = rulesToInsert[i];
-
-        selector = selector.replace(/&/g, '.' + registration.className);
+        selector = selector.replace(/&/g, repeatString(`.${registration.className}`, specificityMultiplier));
```
dzearing, line 272: regex is expensive. maybe it should be conditionalized only when the multiplier exists.

xugao: regex replacement logic here is the same logic before and after my change?

packages/utilities/src/classNamesFunction.ts
```diff
@@ -2,8 +2,11 @@ import { mergeCssSets, IStyleSet, IProcessedStyleSet, Stylesheet } from '@uifabr
 import { IStyleFunctionOrObject } from '@uifabric/merge-styles';
 import { getRTL } from './rtl';
 import { getWindow } from './dom';
+import { StyleFunction } from './styled';
 
 const MAX_CACHE_COUNT = 50;
+const DEFAULT_SPECIFICITY_MULTIPLIER = 5;
```
dzearing, line 8: 5 seems pretty hefty. Was this really needed?

xugao: I think we need to tweak this as we go.
Currently, if you purely look at checkbox.scss, i think this number should be 4 to ensure we don't break user. Some component needs lower and some will need higher, depends on how many modifier classes there are for each component. 

It certainly makes the raw css quite ugly/strange. Do you have a perf concern over the size of this number? (from what i read it should make a difference)

packages/merge-styles/src/styleToClassName.ts
xugao: recursion is actually the most performant for this use case. also, we should never expect a large `count` here. let me know if you disagree

packages/utilities/src/classNamesFunction.ts
dzearing: multiple longer strings injected in bulk. Slight concern but not the end of the world.

PR #13016, opened 2020-05-06
chore(Tree): Prevent string items

assuncaocharles:
#### Pull request checklist

#### Description of changes

This PR reverts the solution from #12984 and prevents strings as Tree Items

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13016)


packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -48,7 +48,7 @@ export interface TreeProps extends UIComponentProps, ChildrenComponentProps {
   exclusive?: boolean;
 
   /** Shorthand array of props for Tree. */
-  items?: ShorthandCollection<TreeItemProps>;
+  items?: TreeItemProps[];
```
mnajdova, line 51: We need to support children as fn, so these typings are not quite correct. Take a look on the `ShorthandValue`, it supports alto this: `(Props<P> & { children?: P['children'] | ShorthandRenderFunction<P> })`. Maybe we should add a special typings for cases where we don't want to support react node, only objects and children as fn..

assuncaocharles: Created `ObjectShorthandValue` and `ObjectShorthandCollection`

packages/fluentui/react-northstar/src/types.ts
```diff
@@ -65,8 +65,15 @@ type ReactNode = React.ReactChild | React.ReactNodeArray | React.ReactPortal | b
 export type ShorthandValue<P extends Props> =
   | ReactNode
   | (Props<P> & { children?: P['children'] | ShorthandRenderFunction<P> });
```
layershifter, line 67:
```suggestion
export type ShorthandValue<P extends Props> = ReactNode | ObjectShorthandValue<P>
```
??

layershifter: And it looks that we need a custom proptypes for this, too

assuncaocharles: Added

assuncaocharles:
Is it ok now? @mnajdova and @layershifter

PR #13017, opened 2020-05-06
Adding cross-platform documentation content for Build 2020

ejlayne:
#### Pull request checklist

#### Description of changes

Added documentation content to the Fluent UI website for the Windows, Mac, and Cross-platform sections of the site. This documentation is needed for Build 2020 to support our Fluent UI React Native library. Note that this PR does not include any of the images for the pages, as those will be coming in a separate PR (or depending on timing, I may add them to this one if it's still open).

#### Focus areas to test

On the fluent website, please test the "Getting Started" and "Controls" pages and sections across all platforms except web (I did not touch web).

Also, on the [ControlName]page.tsx files, I added ImplementationSection's to the contents to get the property tables. I felt like I had to do some unnatural things in this space to stop having errors (even though the content was rendering in spite of the errors), so I am open to feedback about better ways to do things in this section.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13017)


apps/fabric-website/src/SiteDefinition/SiteDefinition.pages/Controls/cross.tsx
```diff
@@ -18,11 +18,55 @@ export const controlsPagesCrossPlatform: INavPage[] = [
     pages: [
       {
         title: 'Button',
-        url: '#/controls/crossplatform/button',
+        url: '#/controls/cross/button',
```
micahgodbolt, line 21: so the plan is to use /cross for the crossplatform path? was that very intentional?

ecraig12345:
You mention images in the description--is the plan to put them in the (private) CDN repo or the main repo? I'd probably prefer the CDN repo to avoid bloating the size of our main repo even more.

apps/fabric-website/src/SiteDefinition/SiteDefinition.pages/Controls/cross.tsx
ecraig12345: I wonder if `reactnative` would be better? Saying this having not been fully following all your work, so feel free to ignore if that doesn't make sense. :)

ejlayne:
@ecraig12345 yes, I've been given instructions to put them here: https://dev.azure.com/onedrive/Design/_git/fabric-cdn/. Please let me know if there's a better place.

ecraig12345:
@ejlayne Yup that's the right place, thanks!

ejlayne:
Addressed merge conflicts from recent changes to master.

apps/fabric-website/src/SiteDefinition/SiteDefinition.pages/Controls/cross.tsx
ejlayne: I added back the "crossplatform" route. Elizabeth, you wouldn't believe the back and forth that went on around the naming of this section, but we're sticking with Cross-platform for now :).

ecraig12345: Eh I'm not too surprised... 😆

PR #13026, opened 2020-05-06
Enabling commonjs post processor script to fix node imports.

dzearing:
When importing from `@fluentui/react` from a node environment, commonjs paths will be pulled. Unfortunately, the code in our `/lib-commonjs` folder refers to `/lib/` imports in other subpackages.

This fix restores a build step which patches the import paths correctly to point to `/lib-commonjs` paths in subpackages.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13026)

### DIFF
--- a/scripts/tasks/postprocess-commonjs.js
+++ b/scripts/tasks/postprocess-commonjs.js
@@ -4,20 +4,32 @@ module.exports.postprocessCommonjsTask = function() {
   const getAllPackageInfo = require('../monorepo/getAllPackageInfo');
   const allPackages = Object.keys(getAllPackageInfo());
 
-  mod('lib-commonjs/**/*.js').asTypescript((node, modder) => {
+  mod('lib-commonjs/**/*.{ts,js}').asTypescript((node, modder) => {
     if (ts.isCallExpression(node)) {
+      console.log(node.expression.getText());
+
       if (node.expression.getText() === 'require' && node.arguments.length === 1) {
         const arg = node.arguments[0];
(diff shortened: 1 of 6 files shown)


scripts/tasks/postprocess-commonjs.js
```diff
@@ -10,7 +10,7 @@ module.exports.postprocessCommonjsTask = function() {
         const arg = node.arguments[0];
         if (ts.isStringLiteral(arg)) {
           if (belongsToFabric(arg.text) && arg.text.includes('/lib/')) {
-            modder.replace(arg, `"${arg.text.replace('/lib/', '/lib-commonjs/')}"`);
+            return modder.replace(arg, `"${arg.text.replace('/lib/', '/lib-commonjs/')}"`);
```
ecraig12345, line 13: Heh... (Anywhere else this was missing?)

ecraig12345:
Hmm we might also need to add change files for all the projects to ensure a fixed version is published. I think the "easiest" way to do this currently is:
- make some random change in all relevant projects
- commit
- yarn change
- revert random change commit

scripts/tasks/postprocess-commonjs.js
dzearing: There are only  commonjs and amd versions of this script and this was the missing one.

PR #13029, opened 2020-05-06
Prevent empty class generation

pingj:
#### Pull request checklist

- [ X ] Include a change request file using `$ yarn change`

#### Description of changes
We notice classes generated with empty style when fabric styles were cached and provided by the APP in the fabric config. The change will prevent such classnames being generated, instead will use the existing provided classname directly. 

Snapshots are updated because we removed class names with empty styles. As a result, it becomes a static classname and not processed by jest-serializer-merge-styles plugin. 
For example, in style.ts wrapper: ['ms-textField-wrapper'] ( with no other style objects), used to output "ms-textField-wrapper wrapper-123" where wrapper-123 is an empty class. In unit test jest-serializer-merge-styles will then convert this into ms-textField-wrapper w/o quotes. After the change, wrapper-123 will not be generated and the output is just "ms-textField-wrapper".

Adding @xugao @dzearing as reviewer.
#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13029)

### DIFF
--- a/packages/merge-styles/src/mergeStyleSets.test.ts
+++ b/packages/merge-styles/src/mergeStyleSets.test.ts
@@ -1,5 +1,5 @@
 import { mergeStyleSets } from './mergeStyleSets';
-import { Stylesheet, InjectionMode } from './Stylesheet';
+import { Stylesheet, InjectionMode, IStyleSheetConfig } from './Stylesheet';
 import { IStyleSet } from './IStyleSet';
 import { IStyleFunctionOrObject } from './IStyleFunction';
 import { IStyle } from './IStyle';
@@ -147,6 +147,21 @@ describe('mergeStyleSets', () => {
 
     expect(styleSet2).toEqual({ root: 'ms-Foo ms-Bar root-1', subComponentStyles: {} });
     expect(_stylesheet.getRules()).toEqual('.root-0{background:red;}' + '.root-1{background:green;}');
+  });
+
+  it('can merge correctly when class names are provided by application', () => {
+    _stylesheet.setConfig({ namespace: 'ns' });
+    _stylesheet.getClassNameCache()['ltr&displayNametestdisplayblock'] = 'test-0';
+
+    const styles = mergeStyleSets({ test: ['test-0'] });
+    const styles2 = mergeStyleSets({ test: { display: 'block' } });
+    const styles3 = mergeStyleSets({ root: [{ background: 'red' }, 'test-0'] });
+
+    expect(styles.test).toBe('test-0');
+    expect(styles2.test).toBe('test-0');
+    expect(styles3.root).toBe('test-0 ns-root-0');
+
+    _stylesheet.setConfig({ namespace: undefined });
   });
 
   describe('typings tests', () => {
--- a/packages/merge-styles/src/mergeStyleSets.ts
+++ b/packages/merge-styles/src/mergeStyleSets.ts
@@ -224,11 +224,13 @@ export function mergeCssSets(
 
       const { classes, objects } = extractStyleParts(styles);
 
-      const registration = styleToRegistration(options || {}, { displayName: styleSetArea }, objects);
+      if (objects?.length) {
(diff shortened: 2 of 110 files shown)


pingj:
@xgao is added to the review. #Closed

packages/merge-styles/src/mergeStyleSets.ts
```diff
@@ -223,12 +223,17 @@ export function mergeCssSets(
       const styles: IStyle = (concatenatedStyleSet as any)[styleSetArea];
 
       const { classes, objects } = extractStyleParts(styles);
-      const registration = styleToRegistration(options || {}, { displayName: styleSetArea }, objects);
 
-      registrations.push(registration);
+      if (objects && objects.length === 0) {
+        classNameSet[styleSetArea] = classes.join(' ');
+      } else {
+        const registration = styleToRegistration(options || {}, { displayName: styleSetArea }, objects);
```
xugao, line 230: it's better to check empty inside `styleToRegistration`

packages/merge-styles/src/mergeStyleSets.ts
```diff
@@ -223,12 +223,17 @@ export function mergeCssSets(
       const styles: IStyle = (concatenatedStyleSet as any)[styleSetArea];
 
       const { classes, objects } = extractStyleParts(styles);
-      const registration = styleToRegistration(options || {}, { displayName: styleSetArea }, objects);
 
-      registrations.push(registration);
+      if (objects && objects.length === 0) {
```
xugao, line 227: could you add unit test to cover this case

packages/merge-styles/src/mergeStyleSets.ts
xugao: hmm that does require a bit change. i'm okay leave this way

packages/merge-styles/src/mergeStyleSets.ts
dzearing, line 227: If `object` is falsey, wouldn't you also want to fall back to the optimization?

Maybe adjust the logic:

```tsx
if (objects?.length) { // if objects is truthy and has a length
  // do normal stuff
} else {
  // do new stuff 
}

Comment (pingj) [reply packages/merge-styles/src/mergeStyleSets.ts]: added

File: packages/merge-styles/src/mergeStyleSets.test.ts [hunk 3]
```diff
@@ -149,6 +149,21 @@ describe('mergeStyleSets', () => {
     expect(_stylesheet.getRules()).toEqual('.root-0{background:red;}' + '.root-1{background:green;}');
   });
 
+  it('can merge correctly when class names are provided by application', () => {
```
xugao, line 152: can you elaborate why this is testing for what you are fixing? what's the test failure looks like without your fix?

packages/merge-styles/src/styleToClassName.ts
```diff
@@ -214,6 +214,10 @@ export interface IRegistration {
 }
 
 export function styleToRegistration(options: IStyleOptions, ...args: IStyle[]): IRegistration | undefined {
+  if (args.length > 1 && (args[1] as IStyleBaseArray)?.length === 0) {
```
xugao, line 217: this check makes a lot assumption. it's cleaner to have the check in caller method like what you had originally

packages/merge-styles/src/mergeStyleSets.test.ts
pingj: talked with @xugao offline, w/o my change, it will fail on
style = mergeStyleSets({ test: ['test-0'] }),  expect(styles.test).toBe('test-0');  
Because it will generate an extra empty style, so the result will be Received: "test-0 ns-test-0"

packages/merge-styles/src/styleToClassName.ts
pingj: moved back

PR #13032, opened 2020-05-06
Slider: Add some tests along with a minor fix for controlled sliders

joschect:
#### Pull request checklist

#### Description of changes

Slider: Add some tests along with a minor fix for controlled sliders

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13032)

### DIFF
--- a/packages/react-next/src/components/Slider/Slider.base.tsx
+++ b/packages/react-next/src/components/Slider/Slider.base.tsx
@@ -188,8 +188,10 @@ export class SliderBase extends React.Component<ISliderProps, ISliderState> impl
     }
   }
 
-  // This should be the only place where value is gotten from.
-  // It ensures that the component behaves consistently whether controlled or uncontrolled.
+  /**
+   * This should be the only place where value is gotten from.
+   * It ensures that the component behaves consistently whether controlled or uncontrolled.
+   */
   public get value(): number | undefined {
     const { value = this.state.value } = this.props;
     if (this.props.min === undefined || this.props.max === undefined || value === undefined) {
--- a/packages/react-next/src/components/Slider/Slider.base.tsx
+++ b/packages/react-next/src/components/Slider/Slider.base.tsx
@@ -188,6 +188,8 @@ export class SliderBase extends React.Component<ISliderProps, ISliderState> impl
     }
   }
 
+  // This should be the only place where value is gotten from.
+  // It ensures that the component behaves consistently whether controlled or uncontrolled.
   public get value(): number | undefined {
     const { value = this.state.value } = this.props;
     if (this.props.min === undefined || this.props.max === undefined || value === undefined) {
@@ -297,6 +299,8 @@ export class SliderBase extends React.Component<ISliderProps, ISliderState> impl
 
   private _updateValue(value: number, renderedValue: number): void {
     const { step, snapToStep } = this.props;
+
+    // TODO this should be memoized based on step since it will only change when step changes
     let numDec = 0;
     if (isFinite(step!)) {
       while (Math.round(step! * Math.pow(10, numDec)) / Math.pow(10, numDec) !== step!) {
--- a/packages/react-next/src/components/Slider/Slider.base.tsx
+++ b/packages/react-next/src/components/Slider/Slider.base.tsx
@@ -294,6 +294,7 @@ export class SliderBase extends React.Component<ISliderProps, ISliderState> impl
     }
     return currentPosition;
   }
+
   private _updateValue(value: number, renderedValue: number): void {
     const { step, snapToStep } = this.props;
     let numDec = 0;
@@ -305,7 +306,7 @@ export class SliderBase extends React.Component<ISliderProps, ISliderState> impl
 
     // Make sure value has correct number of decimal places based on number of decimals in step
     const roundedValue = parseFloat(value.toFixed(numDec));
-    const valueChanged = roundedValue !== this.state.value;
+    const valueChanged = roundedValue !== this.value;
 
     if (snapToStep) {
       renderedValue = roundedValue;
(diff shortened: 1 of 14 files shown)


packages/react-next/src/components/Slider/Slider.base.tsx
```diff
@@ -294,8 +296,11 @@ export class SliderBase extends React.Component<ISliderProps, ISliderState> impl
     }
     return currentPosition;
   }
+
   private _updateValue(value: number, renderedValue: number): void {
     const { step, snapToStep } = this.props;
+
+    // TODO this should be memoized based on step since it will only change when step changes
```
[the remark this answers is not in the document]
joschect: I think that this is an interesting question. You're probably right that it's not too expensive but should it seems like it should only need to get set rarely. Do we want there to be a standard of memoizing/converting these things into state? This will happen automagically if when it get's converted to hooks if something useEffect is used.

packages/react-next/src/components/Slider/Slider.test.tsx
```diff
@@ -25,14 +23,41 @@ describe('Slider', () => {
     expect(tree).toMatchSnapshot();
   });
 
-  it('can slide to default min/max and execute onChange', () => {
+  it('calls onChange when slider changes', () => {
```
khmakoto, line 26:
```suggestion
  it('calls onChange when slider value changes', () => {
```

packages/react-next/src/components/Slider/Slider.test.tsx
```diff
@@ -119,30 +167,51 @@ describe('Slider', () => {
...
+
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.down });
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.down });
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.down });
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.up });
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.down });
+
+    expect(slider.current?.value).toEqual(9);
+
+    // onChanged should only be called after a delay
```
khmakoto, line 193: Remove this comment.

packages/react-next/src/components/Slider/Slider.test.tsx
```diff
@@ -151,7 +220,87 @@ describe('Slider', () => {
...
+    const onChange = jest.fn();
+
+    wrapper = mount(<Slider label="slider" componentRef={slider} value={3} min={0} max={100} onChange={onChange} />);
+    const sliderSlideBox = wrapper.find('.ms-Slider-slideBox');
+
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.down });
+
+    expect(slider.current?.value).toEqual(3);
+
+    // onChanged should only be called after a delay
```
khmakoto, line 236: Remove comment.

packages/react-next/src/components/Slider/Slider.test.tsx
```diff
@@ -119,30 +167,51 @@ describe('Slider', () => {
...
 
-    expect(sliderSlideBox.getAttribute('aria-valuenow')).toEqual('9');
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.down });
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.down });
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.down });
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.up });
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.down });
 
+    expect(slider.current?.value).toEqual(9);
     // onChanged should only be called after a delay
```
khmakoto, line 215:
```suggestion
    // onChange should only be called after a delay
```

packages/react-next/src/components/Slider/Slider.test.tsx
```diff
@@ -151,7 +220,87 @@ describe('Slider', () => {
...
+
+    wrapper = mount(
+      <Slider label="slider" defaultValue={10} componentRef={slider} step={-3} min={0} max={100} onChange={onChange} />,
+    );
+    const sliderSlideBox = wrapper.find('.ms-Slider-slideBox');
+
+    sliderSlideBox.simulate('keydown', { which: KeyCodes.up });
+    expect(slider.current?.value).toEqual(7);
+  });
+  it('correctly changes value with decimal steps', () => {
```
khmakoto, line 284: Add blank line between tests.

packages/react-next/src/components/Slider/Slider.test.tsx
joschect: This one is actually correct! onChanged is the prop that is called after a delay.

packages/react-next/src/components/Slider/Slider.base.tsx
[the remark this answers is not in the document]
joschect: For now I'm going to leave this, and we can decide what we want later.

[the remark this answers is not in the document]
ecraig12345: Using state to cache expensive calculations has always been an anti-pattern, but people have done it a lot anyway due to the lack of standard/easy patterns around memoization. Doing memoization right will become a lot easier with function components with `React.useMemo`. But memoization itself does have a cost, so it's still not something that should be used in all cases.

(FYI since we're on the topic: it's good to keep in mind that [per the React docs](https://reactjs.org/docs/hooks-reference.html#usememo), `useMemo` is a performance optimization, *not* a semantic guarantee--see the docs for more on that, but basically, it means it doesn't guarantee that a calculation won't be re-run. Practically this would only cause problems for exceptionally expensive calculations or calculations with side effects, the latter of which should be avoided anyway.)

Issue #13012, opened 2020-05-05
Tracking: investigate severities of Chromium Edge High Contrast regression


czearing:
Background:

The Edge update broke existing high contrast selectors. We need to do a pass to understand the full impact across all components.

Original issue is here: https://github.com/microsoft/fluentui/issues/11742

I will go down the list below, investigate Edge old vs Edge new, and fill out Severity and Notes columns.

This list is starting from the top level exports of office-ui-fabric-react, and will be reduced/expanded depending on the components (if any) exported from the export.

| Priority | Meaning |
| -|-|
| n/a | Visually the same |
| 2 | Minor differences but completely usable |
| 1 | Major changes and some features hard to see |
| 0 | Unusable |

Things to try:
* Pressed states
* Hover states
* Focus (tabbing) differences
* Disabled states if applicable
* Are interactive links indecernable?

| Component | Priority | Notes |
|-|-|-|
| Breadcrumb |  2 | Old Edge: Focus and hover caused text to go black making it unreadable. New Edge: Now readable. (Better) |
| Button | 1 | Old Edge: shows disabled button and icon. New Edge: disabled buttons aren't noticeable and icons are cut off.  |
| Calendar | 1.5 | Old Edge: Initial selection and week selection visible. New Edge:  Initial selection doesn't show up and week selection is not visible. |
| Callout | n/a | Callout box shows no visible changes. |
| Card | 2 | Old Edge: highlight around card is visible. New Edge:  highlight isn't visible. |
| ~~Checkbox~~ | ~~0~~ | Fixed by #13035 ~~Old Edge: Disabled checkbox visible and checked boxes are visible. New Edge:  Disabled buttons aren't visible and the user is unable see the boxes that they check.~~ |
| ~~ChoiceGroup~~ | ~~1~~ | Fixed by #13846 ~~Old Edge: disabled items visible. New Edge:  disabled items not visible.~~ |
| Coachmark | 2 | Old Edge: Ripple effect is blue. New Edge : Ripple effect is white |
| ColorPicker | 1.5 | It works fine but the preview box doesn't work in old or new edge.  |
| ~~ComboBox~~ | ~~2~~ | Fixed by #14273 ~~Old Edge: No lines separating items in box. New Edge: Lines separate items in the box~~ |
| CommandBar | 1.5 | Old Edge: Selected icons are green and items have hover effect. New Edge: No hover effect and icons are yellow. |
[...]

PR #13034, opened 2020-05-06
Fix MessageBar high contrast issues

### LINKED ISSUE #13019: Color of any link in messageBar should be as per accessibility guideline in High contrast mode like in black mode it should be yellow.
Color of any link in messageBar should be as per accessibility guideline in High contrast mode like in black mode it should be yellow.

### Environment Information

OS: Win 10
Browser: Edge

### Describe the issue:

Color of any link in messageBar should be as per accessibility guideline in High contrast mode like in black mode it should be yellow.

#### Actual behavior:

Color of any link in messageBar is showing in blue color

#### Expected behavior:

Color of any link in messageBar should be as per accessibility guideline in High contrast mode like in black mode it should be yellow.

Pls see in below snap in version 7.86.0
7 86 0_External

Pls see in below snap in latest version
Latest_External

### LINKED ISSUE #13113: Not able to define tooltip of clear button of messageBar(Error messageBar)
Not able to define tooltip of clear button of messageBar(Error messageBar)

### Environment Information

OS: Win10
Browser: Edge

### Describe the issue:

Not able to define tooltip of clear button of messageBar(Error messageBar)

https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar
Code pen link(Error messageBar)
https://codepen.io/Harkesh/pen/LYprqEr?editable=true%3Dhttps%3A%2F%2Fdeveloper.microsoft.com%2Fen-us%2Ffluentui

#### Actual behavior:

Not able to define tooltip of clear button of messageBar(Error messageBar)

#### Expected behavior:

Should able to define tooltip of clear button of messageBar(Error messageBar)

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13019, fixes #13113, related to #11742 and #13012

#### Description of changes

Fix high contrast issues with MessageBar:

**Background colors** were being dropped due to browser changes as described in #11742 and #13012. Fix by disabling high contrast adjustment.

**Link colors**: In high contrast mode, Links inside MessageBars should follow the OS high contrast theme's link color rather than using Fluent UI theme colors. #12821 removed the style override for this as part of a larger change

I also noticed that **button focus outlines** were missing in high contrast due to using the wrong color for the current way the MessageBar high contrast colors are handled.

Also fixes #13113, use dismissButtonAriaLabel as the dismiss button's tooltip as well.

How it looks today (incorrect):
image

How it's supposed to look (and will again after this PR):
image

#### Focus areas to test

High contrast mode in Chromium Edge and IE/old Edge

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13034)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
@@ -135,12 +135,11 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
         display: 'flex',
         wordBreak: 'break-word',
         selectors: {
-          // In high contrast mode, Links within MessageBars must revert to the default color
-          // defined by the high contrast settings rather than following the theme colors.
-          '& .ms-Link': {
+          '.ms-Link': {
+            color: semanticColors.messageLink,
             selectors: {
-              [HighContrastSelector]: {
-                MsHighContrastAdjust: 'auto',
+              ':hover': {
+                color: semanticColors.messageLinkHovered,
               },
             },
           },
@@ -148,6 +147,14 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
             background: highContrastBackgroundColor[messageBarType],
             border: '1px solid WindowText',
             color: 'WindowText',
+            // will uncomment or remove depending on how the styling override in Link works
+            // selectors: {
+            //   '.ms-Link, .ms-Link:hover': {
+            //     // In high contrast mode, Links within MessageBars must revert to the default color
+            //     // defined by the high contrast settings rather than following the theme colors.
+            //     MsHighContrastAdjust: 'auto',
+            //   },
+            // },
           },
         },
       },
--- a/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
@@ -93,8 +93,8 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
     getFocusStyle(theme, {
       inset: 1,
       highContrastStyle: {
-        outlineOffset: '-4px',
-        outlineColor: 'Window',
+        outlineOffset: '-6px',
+        outline: '1px solid Highlight',
       },
       borderColor: 'transparent',
     }),
@@ -144,17 +144,10 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
             },
           },
           [HighContrastSelector]: {
+            MsHighContrastAdjust: 'none',
             background: highContrastBackgroundColor[messageBarType],
             border: '1px solid WindowText',
             color: 'WindowText',
-            // will uncomment or remove depending on how the styling override in Link works
-            // selectors: {
-            //   '.ms-Link, .ms-Link:hover': {
-            //     // In high contrast mode, Links within MessageBars must revert to the default color
-            //     // defined by the high contrast settings rather than following the theme colors.
-            //     MsHighContrastAdjust: 'auto',
-            //   },
-            // },
           },
         },
       },
@@ -224,6 +217,8 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
         whiteSpace: 'pre-wrap',
       },
       !isMultiline && {
+        // In high contrast this causes the top and bottom of links' focus outline to be clipped
+        // (not sure of a good way around that while still maintaining text clipping)
         overflow: 'hidden',
         textOverflow: 'ellipsis',
         whiteSpace: 'nowrap',
--- a/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
@@ -135,6 +135,15 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
         display: 'flex',
         wordBreak: 'break-word',
         selectors: {
+          // In high contrast mode, Links within MessageBars must revert to the default color
+          // defined by the high contrast settings rather than following the theme colors.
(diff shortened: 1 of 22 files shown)


packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
```diff
@@ -135,6 +135,15 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
         display: 'flex',
         wordBreak: 'break-word',
         selectors: {
+          // In high contrast mode, Links within MessageBars must revert to the default color
+          // defined by the high contrast settings rather than following the theme colors.
+          '& .ms-Link': {
```
chrismohr, line 140: Thanks @ecraig12345 for fixing this. 

Wondering what the best practice is in a case like this... I assumed that using a Customizer component with scoped settings might be a little more robust than depending on the class name. Thoughts?

packages/styling/src/interfaces/ISemanticColors.ts
```diff
@@ -195,11 +195,11 @@ export interface ISemanticColors extends ISemanticTextColors {
    */
   successIcon: string;
   /**
-   * Color of links within a Message.
+   * Color of links within a message.
```
chrismohr, line 198: 👍

packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
xugao, line 140:
```suggestion
          '.ms-Link': {
```

ecraig12345: Interesting question. We've typically used class names for nested styling, but you're right that the Customizer approach is probably more robust (except in cases where the extra selector specificity is necessary for some reason). Some concerns about if we started to use that approach more widely though:
- Perf impact--may not be much, but it's hard to say for sure without testing. (Though perf impact for this specific case should be minimal since there should only be one MessageBar on a page, or worst case a couple of them.) 
- Not ideal to have some of the files in the .styles.ts and others in the main file
- Not sure whether it would conflict with the effort @xugao may be starting to move styles back to scss

Also realized thinking about it more that this style could potentially be moved to Link itself instead of being an override in MessageBar.  It wouldn't do anything except in cases like MessageBar where a parent of the Link has set MsHighContrastAdjust to none.

xugao: this made me notice... what is this code here: https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.base.tsx#L155

Let's just keep the fix simple here and not use `Customizer`. it doesn't really make this particular case more robust but it brings more complexity.

chrismohr: >  it doesn't really make this particular case more robust

@xugao I'm just passing through here, and don't know what is planned for the future, but there are always trade-offs.  

There has clearly been an effort to create a public interface for style overrides through the styles prop. Bypassing this interface means that when Link is changed in the future there are no guarantees that this type a css override will continue to work.  Or, worse in my mind is that future developers will feel like Link is too fragile to change due to all the unreferenced css overrides.

`Customizer` is the solution that consumers would need to use for this scenario.  Is it really more complex?  Meaning that it would be more difficult for future developers to understand the intentions of the code?  Can you please help me understand?

ecraig12345: Talked with @dzearing about this and while he agrees that the Customizer approach is more robust than class names, he thinks we should stick with class names for now for a few different reasons, mainly revolving around the upcoming move away from css-in-js (due to perf concerns) to scss and css variables: (this was copied from a chat)
1. we're moving to css variables so this is temporary
2. it will be easier to rationalize when we do
3. easier to upgrade users if we're not very reliant on context for passing around css

(I think what I said above about sticking with the same approach we use elsewhere in our codebase and keeping it centralized in the styles file was also a consideration.)

chrismohr: > stick with class names

Maybe I'm trying to solve pain points that don't actually exist for you all, but maybe you'd want to consider importing the string for the class name from somewhere rather than just hardcoding it.

dzearing: Just to chime in and support @ecraig12345 - we definitely want to consider the future here.

We know that css in js leads to performance problems when done in bulk; it's easy to abuse as well. We know that relying on react context to spread styling rules ends up causing react re-renders when themes change. We've worked around these pitfalls in numerous ways over the past years through multiple class caching layers. We used css in js primarily for 2 reasons: normalizing specificity and dynamic contextual theming. But everything has tradeoffs, and we're just not happy with the approach.

CSS custom properties really help with contextual theming. You can define variables at pretty much any level and they can cascade down. We could use this concept to apply standardized classes which represent themable values. Components like Link and Button could consume these tokens and provide fallbacks. This means from a perf perspective:

1. You don't need to re-evaluate the stylesheet. It can be static, but still customizable.
2. You don't depend on react context for theming. No react context means you could even render content using any other css tech (static css included) inside of the controls and still use the theme values.
3. It also means that if the theme changes, you don't need to re-render everything. You just spit out new variable values and they trickle down the dom tree.

So, that's the future. In the present, we still have Customizer, styles props, css in js, etc. If we use the approach of styling subcomponents as @ecraig12345 did here, it means:

1. As she mentioned, having all the styling logic is in one place makes it much easier to port the concern.
2. We achieve the desired outcome (links get styled correctly)
3. Porting to css variables would remove the react context dependency in delivering theme values to the styles.

It certainly has a lot of downsides too. Global selectors are bad. Reaching into internals is bad. We need to move to variables as soon as possible. This is our current focus in `@fluentui/react-next` right now (our v8 alpha build)

PR #13035, opened 2020-05-07
Checkbox: Fix High Contrast in Edge Chromium

### LINKED ISSUE #11540: Can't tell if checkbox is checked in high-contrast-mode

xugao:
#### Pull request checklist

- [ ] Mega issue: #11742
- [X] Addresses an existing issue: Fixes #11540

#### Description of changes
**Recap of the issue:**
In old Edge, the `-ms-high-contrast` media query implies `-ms-high-contrast-adjust: none` on each property inside the media query. In new/chromium Edge, this is no longer the case and `-ms-high-contrast-adjust: none` needs to be set explicitly for the properties inside HC media query to take effect.

**Heads-up:** 
The solution to this problem is not ideal. Ideally, this can be somehow addressed/improved at browser-level. However, after talking to folks in Edge, I don't see them offering anything concrete for us any time soon.

**Principal of the fix:**
1. Increase bundle size as little as possible. Add minimal amount of new css to resolve the issue.
2. No breaking change for styles set by existing use cases: 
Turing off high contrast adjustment can be a breaking change to the styles user set today on our component. For the styles user set today which are expected to be adjusted, this change potentially turns off adjustments for those case and user will need to add styles to set colors explicitly for HC.

**The fix:**
1. Added a new css selector added which turns off the HC adjustment in new Edge, and **only** new Edge (this is ensure principal No.2 above). You can read more about `forced-color-adjust` [here](https://www.w3.org/TR/css-color-adjust-1/#forced). it serves the same purpose as `-ms-high-contrast-adjust: none` in the context of this change.
2. Applies the new css selector in places where set high contrast styles using `HighContrastSelector` so our high contrast styles work in new Edge.
3. Find places where colors were previously expected to get adjusted but no longer are. Explicitly set HC styles for those cases

**To see what's changed visually**
In new Edge, before fix: https://developer.microsoft.com/en-us/fluentui#/controls/web/checkbox
In new Edge, after fix: http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/13035/merge/fabric-website-resources/dist/index.html#/examples/checkbox

New Edge + Before fix:
image

New Edge + After fix:
image

Old Edge + Before fix:
image

Old Edge + After fix:
image

#### Focus areas to test
- Old Edge
- New Chromium Edge

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13035)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.styles.ts
@@ -1,5 +1,10 @@
 import { ICheckboxStyleProps, ICheckboxStyles } from './Checkbox.types';
-import { HighContrastSelector, getGlobalClassNames, IStyle } from '../../Styling';
+import {
+  HighContrastSelector,
+  getGlobalClassNames,
+  getEdgeChromiumNoHighContrastAdjustSelector,
+  IStyle,
+} from '../../Styling';
 import { IsFocusVisibleClassName } from '../../Utilities';
 
 const GlobalClassNames = {
@@ -56,6 +61,11 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
       transitionProperty: 'border-width, border, border-color',
       transitionDuration: MS_CHECKBOX_TRANSITION_DURATION,
       transitionTimingFunction: MS_CHECKBOX_TRANSITION_TIMING,
+      selectors: {
+        [HighContrastSelector]: {
+          borderColor: 'WindowText',
+        },
+      },
     },
   ];
 
@@ -104,10 +114,6 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
                 background: checkboxBackgroundCheckedHovered,
                 borderColor: checkboxBorderColorCheckedHovered,
               },
-              [`.${classNames.checkbox}`]: {
-                background: checkboxBorderColorChecked,
-                borderColor: checkboxBorderColorChecked,
-              },
               [HighContrastSelector]: {
                 selectors: {
                   [`:hover .${classNames.checkbox}`]: {
@@ -134,6 +140,11 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
           selectors: {
             [`:hover .${classNames.checkbox}, :hover .${classNames.checkbox}:after`]: {
               borderColor: checkboxBorderIndeterminateHoveredColor,
+              selectors: {
+                [HighContrastSelector]: {
+                  borderColor: 'WindowText',
+                },
+              },
             },
             [`:focus .${classNames.checkbox}`]: {
               borderColor: checkboxBorderIndeterminateHoveredColor,
@@ -145,8 +156,14 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
         },
         {
           selectors: {
-            [`:hover .${classNames.text}`]: { color: checkboxHoveredTextColor },
-            [`:focus .${classNames.text}`]: { color: checkboxHoveredTextColor },
+            [`:hover .${classNames.text}, :focus .${classNames.text}`]: {
+              color: checkboxHoveredTextColor,
+              selectors: {
+                [HighContrastSelector]: {
+                  color: disabled ? 'GrayText' : 'WindowText',
+                },
+              },
+            },
           },
         },
       ],
@@ -219,6 +236,10 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
         overflow: 'hidden',
         selectors: {
           ':after': indeterminate ? indeterminateDotStyles : null,
+          [HighContrastSelector]: {
+            borderColor: 'WindowText',
+          },
+          ...getEdgeChromiumNoHighContrastAdjustSelector(),
         },
       },
       indeterminate && {
@@ -251,7 +272,7 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
         borderColor: checkboxBorderColorDisabled,
         selectors: {
           [HighContrastSelector]: {
-            borderColor: 'InactiveBorder',
+            borderColor: 'GrayText',
           },
         },
       },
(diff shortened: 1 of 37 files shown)


packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.styles.ts
```diff
@@ -251,14 +272,19 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
         borderColor: checkboxBorderColorDisabled,
         selectors: {
           [HighContrastSelector]: {
-            borderColor: 'InactiveBorder',
```
xugao, line 254: I am not sure why we had `InactiveBorder` here. I updated to `GrayText` since that's what being used everywhere for disabled state

packages/office-ui-fabric-react/src/components/Checkbox/Checkbox.styles.ts
```diff
@@ -104,10 +114,6 @@ export const getStyles = (props: ICheckboxStyleProps): ICheckboxStyles => {
                 background: checkboxBackgroundCheckedHovered,
                 borderColor: checkboxBorderColorCheckedHovered,
               },
-              [`.${classNames.checkbox}`]: {
```
xugao, line 107: this block is not needed as it's already defined below within `checkbox` section

khmakoto (approved):
Looks good! Thanks for doing this!

As a suggestion, you might want to add screenshots of the before and after when dealing with High Contrast related changes to let people know what to look for. Not super important for this change since it's overhauling the entire experience for the component but something to keep in mind for later 😄.

packages/styling/src/styles/CommonStyles.ts
```diff
@@ -19,3 +23,14 @@ export const ScreenWidthMinUhfMobile = 768;
 export function getScreenSelector(min: number, max: number): string {
   return `@media only screen and (min-width: ${min}px) and (max-width: ${max}px)`;
 }
+
+/**
+ * The style which turns off high contrast adjustment in (only) Edge Chromium browser.
+ */
+export function getEdgeChromiumNoHighContrastAdjustSelector(): { [EdgeChromiumHighContrastSelector]: IStyle } {
```
ecraig12345, line 30: In the interest of bundle size, I wonder if we should use a shorter name for this?

ecraig12345: Also wonder if the return type should be IRawStyle.

ecraig12345:
Thanks for doing this! +1 to Mak's comment about before/after screenshots.

One other thing you'll need to check (which unfortunately is only possible in production) is that the colors also work as expected in high contrast on the website, which has some additional obnoxious style overrides that we have no control over. It's possible that artificially increased specificity or other weirdness in the CSS was added for that reason.

xugao:
@khmakoto @ecraig12345 - thanks for the review!

I think instead of screenshot here is what I recommend you to check out:
In new Edge, before fix: https://developer.microsoft.com/en-us/fluentui#/controls/web/checkbox
In new Edge, after fix: http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/13035/merge/fabric-website-resources/dist/index.html#/examples/checkbox

Let me know if you spot any issue!

ecraig12345:
Old Edge got uninstalled from my computer and I'm not sure how to get it back (same for some others), which limits my testing ability. ☹️ It's also just much nicer for documentation/point-in-time reference purposes to have the screenshots right here in the PR.

packages/styling/src/styles/CommonStyles.ts
xugao: the function name doesn't affect bundle size as it can be minified. the size increase is mainly due to the styling code I added to checkbox, which is an unfortunate side-effect with making these fixes in our code :(

khmakoto:
@xugao like @ecraig12345 said adding screenshots is just for better documentation and reference. I did check the PR deploy links for the changes and I'd expect us to do that before approving no matter if there are or there aren't screenshots in the PR description.

xugao:
@ecraig12345 @khmakoto - agreed. added screenshots in PR description :)

packages/styling/src/styles/CommonStyles.ts
xugao: Updated typing to `IRawStyle`

ecraig12345: Actually function names are not minifiable in many cases, including when exported. (Class method names are also not minifiable.)

xugao: i verified in our `test-bundle` and it's being minified. can you elaborate when exactly you see it's not?

ecraig12345: Hmm I just looked into it and I guess I was incorrect in assuming that exported names would be handled the same as private members--searching for a similar example `getFocusOutlineStyle` in [this published version](https://unpkg.com/office-ui-fabric-react@7.110.5/dist/office-ui-fabric-react.min.js), all I find is `o.d(t,"getFocusOutlineStyle",(function(){return io}))`. So I guess it uses some optimization so exported things can still be referenced by name, but the name is minified. Sorry about the confusion.

grantrostig:
Not fixed. Running fully updated win11 pro with High Contrast set on win11 settings. Then signed up for "Bing" bonus as prompted by MS using Edge fully updated, "Confirm" checkbox is high contrast but checking it did not show check mark.  It did silently do so, and I did proceed.

PR #13044, opened 2020-05-07
fix(Popper): exclude useless updates, fix `fixed` positioning

layershifter:
### Fix `fixed` positioning

image

Due bug in `Popper` elements that are using `fixed` positioning https://github.com/popperjs/popper-core/issues/1054 may be incorrectly positioned. This issue was fixed in `2.2.2`, this PR bumps `@popperjs/core` to `2.4.0`.

### Fix flickering

Previously we used `React.useEffect` to schedule updates, but as updates can cause repositioning we should use `React.useLayoutEffect` to avoid flickering, see GIF below.

https://github.com/microsoft/fluentui/blob/31404935a5e8287ea7e5fdb491c1deb96338b710/packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx#L186

Popper flickering

### Decrease updates

https://github.com/microsoft/fluentui/blob/31404935a5e8287ea7e5fdb491c1deb96338b710/packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx#L186

`React.useEffect` & `React.useLayoutEffect` will be executed after a component's mount, but there is no sense to trigger this update. Updates should be executed only on changes in `positioningDependencies` & `computedPlacement`.

***

Also adds a visual test to be sure that the regression is fixed. Thanks to @kolaps33 for the help in debugging this issue.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13044)


ecraig12345:
@layershifter This has merge conflicts now

PR #13047, opened 2020-05-07
chore(Toolbar): use compose()

layershifter:
#### Pull request checklist

### Findings:

- [ ] `mapPropsToStylesProps()` - we need to have there `ParentProps`, see `ToolbarMenuItemIcon.ts`
- [ ] `ref` has a wrong type (for example it's `button`, but should be `HTMLButtonElement`), see `Toolbar.tsx`
- [ ] `mapPropsToStylesProps()` - should allow only primitive types
- [ ] add dev check to `useStyles` to check `mapPropsToStylesProps` for primitive types

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13047)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Box/Box.tsx
+++ b/packages/fluentui/react-northstar/src/components/Box/Box.tsx
@@ -28,7 +28,7 @@ export interface BoxProps extends UIComponentProps<BoxProps>, ContentComponentPr
   /** Accessibility behavior if overridden by the user. */
   accessibility?: Accessibility<never>;
 }
-export type BoxStylesProps = never;
+export type BoxStylesProps = {};
 
 export const boxClassName = 'ui-box';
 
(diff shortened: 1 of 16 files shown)


packages/fluentui/react-northstar/src/components/Box/Box.tsx
```diff
@@ -28,7 +28,7 @@ export interface BoxProps extends UIComponentProps<BoxProps>, ContentComponentPr
   /** Accessibility behavior if overridden by the user. */
   accessibility?: Accessibility<never>;
 }
-export type BoxStylesProps = never;
+export type BoxStylesProps = {};
```
layershifter, line 31: Merging with `never` will not work as any union that contains `never` becomes `never` 😨

PR #13048, opened 2020-05-07
fix(Carousel): Allows index as number only

assuncaocharles:
#### Pull request checklist

#### Description of changes

Carousel active index as only number to avoid cast and conflicts with animations

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13048)


packages/fluentui/react-northstar/src/components/Carousel/Carousel.tsx
```diff
@@ -306,15 +307,11 @@ export const Carousel: React.FC<WithAsProp<CarouselProps>> &
                   unmountOnExit
                   visible={active}
                   name={
-                    initialMounting
+                    initialMounting || !active
```
mnajdova, line 310: Thanks for adding this in this PR ❤️

layershifter:
Please add to PR description why this change is needed

PR #13058, opened 2020-05-08
Feature: disabledFeatures switches

phkuo:
#### Pull request checklist

#### Description of changes

Adds a disabledFeatures list of flags to the global config which allows an easy way for apps to turn off individual portions of code. This lets us respond quickly to regressions. The code is modeled after the system that SharePoint uses, where we have infrastructure in place that lets us immediately flip these switches with no code change necessary.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13058)


packages/utilities/src/IFabricConfig.ts
```diff
@@ -0,0 +1,32 @@
+import { IStyleSheetConfig } from '@uifabric/merge-styles';
```
phkuo, line 1: moved this here out of the styling package, since it's no longer exclusively used by styling

also added `theme` to it since theme.ts was using `theme` off of it anyways

packages/utilities/src/disabledFeatures.test.ts
```diff
@@ -0,0 +1,37 @@
+import { isFeatureDisabled } from './disabledFeatures';
+
+describe('isFeatureDisabled', () => {
+  it('works', () => {
+    let a = 1;
+    let b = 0;
+    let c = 0;
```
xugao, line 7: pls break this test to multiple more targeted tests

packages/utilities/src/disabledFeatures.ts
```diff
@@ -0,0 +1,21 @@
...
+export function isFeatureDisabled(guid: string, owner: string, date: string) {
+  if (!owner || !date) {
+    throw 'isFeatureDisabled() requires an owner and a date';
+  }
+
+  const fabricConfig: IFabricConfig = (getWindow() as any)?.FabricConfig; // tslint:disable-line:no-any
+  if (fabricConfig?.disabledFeatures) {
+    return fabricConfig.disabledFeatures[guid.toUpperCase()];
+  }
+  return false;
```
xugao, line 20:
```suggestion
return fabricConfig?.disabledFeatures[guid.toUpperCase()] === true;
```

packages/utilities/src/disabledFeatures.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Used to determine if a portion of code should run.
+ * See https://github.com/microsoft/fluentui/wiki/disabledFeatures-Switches for more details.
+ * @param guid - GUID associated with this switch
+ * @param owner - alias or name of the owner of this switch.
+ * @param date - date this switch was created, preferably MM/DD/YY e.g. 12/31/20, or longform e.g. Dec 31, 2020
+ */
+export function isFeatureDisabled(guid: string, owner: string, date: string) {
+  if (!owner || !date) {
+    throw 'isFeatureDisabled() requires an owner and a date';
```
xugao, line 13:
```suggestion
    throw 'isFeatureDisabled requires an owner and a date';
```

phkuo:
Will open a new PR that is enabledFeatures instead of disabledFeatures.

PR #1630, opened 2017-04-26
Use overflow-wrap: break-word instead of word-break: break-all in tooltips, labels and dropdowns

trevorsg:
Fix #1627

#### Pull request checklist

- [x] Addresses an existing issue: #1627 
- [x] Include a change request file if publishing 
- [ ] New feature, bugfix, or enhancement
  - [ ] Includes tests
- [ ] Documentation update

#### Description of changes

See #1627

PR #13064, opened 2020-05-08
fix(Popup): Avoid weird condition on popup

assuncaocharles:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes [#1630](https://github.com/microsoft/fluent-ui-react/issues/1630)

#### Description of changes

Currently we are using some weird condition to avoid Popup to break when passing `content={null|false}`. It breaks because inside of `innerRef` from Popper expects a element: 

```jsx
            <Ref
              innerRef={domElement => {
                popupContentRef.current = domElement;
                handleRef(contentRef, domElement);
                nestingRef.current = domElement;
              }}
            >
                 {popupContent}
            </Ref>
```

When passing `null | false` we are returning `null` to `popupContent` and so [breaking Popper component and Refs that expect children](https://github.com/microsoft/fluent-ui-react/issues/1630).  

The current solution just wraps `popupContent` in a fragment to keep same behavior (`innerRef.current = null` and `domElement = null`), not break and avoiding the condition.

```jsx
            <Ref
              innerRef={domElement => {
                popupContentRef.current = domElement;
                handleRef(contentRef, domElement);
                nestingRef.current = domElement;
              }}
            >
                <>{popupContent}</>
            </Ref>
```  

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13064)


assuncaocharles:
@layershifter Returning `React.Fragment` instead of null for `value: false|null` fix it, since it's non-callable and it will produce same result rendering nothing and not breaking. The question is if it would have another implications or is it even right in this case. If it is not proper solution ( that I guess it's the case ) what would you suggest? If it's ok solution ( I doubt ) I would fix the tests.

packages/fluentui/e2e/tests/popupClickHandling-example.tsx
```diff
@@ -15,16 +15,18 @@ const PopupClickHandlingExample = () => {
...
-          <Button
-            id={selectors.popupContentButtonId}
-            content="click me"
-            onClick={e => {
-              e.preventDefault();
-              setShow(!show);
-            }}
-          />
-        )
+        <div>
```
assuncaocharles, line 18: Since `popupContent` will be null when `false` is passed to `content` there would be no content wrapper in this case, what makes sense for me. To achieve the result expected in the test I am adding explicit the wrapper

pompomon: I think that then we'd need to describe this in Breaking changes, because as you show in the test, it would require users to upgrade their code, for the example to still work the same

assuncaocharles: Sure thing, makes sense for me. I will just wait @layershifter to comment if this is the proper fix and will add it on the changelog

PR #13074, opened 2020-05-08
(wip) Convert v7 to eslint

ecraig12345:
Working on converting v7 to eslint. Very very much still a work in progress. Before leaving off I was working on disabling all the rules that are causing errors, in some cases to be re-enabled in a separate PR that's less noisy already.


scripts/eslint/v7.js
```diff
@@ -0,0 +1,217 @@
+module.exports = {
```
xugao, line 1: for my own curiosity - how did you create this config?

scripts/eslint/v7.js
```diff
@@ -0,0 +1,217 @@
...
+  parser: '@typescript-eslint/parser',
+  // // can be removed once we extend airbnb config
+  // parserOptions: {
+  //   ecmaVersion: 2018,
+  //   ecmaFeatures: { jsx: true },
+  //   sourceType: 'module',
+  // },
+  plugins: ['@typescript-eslint', 'deprecation', 'import', 'jest', 'jsx-a11y', 'react', 'react-hooks'],
+  // plugins: ['@typescript-eslint', 'deprecation', 'import', 'jest', 'jsx-a11y', 'react', 'react-hooks'],
+  // // can be removed once we extend airbnb config
```
xugao, line 12: can remove already since we are extending airbnb?

scripts/eslint/v7.js
```diff
@@ -0,0 +1,217 @@
...
+  extends: ['airbnb', 'prettier'],
+  parser: '@typescript-eslint/parser',
+  // // can be removed once we extend airbnb config
+  // parserOptions: {
+  //   ecmaVersion: 2018,
+  //   ecmaFeatures: { jsx: true },
+  //   sourceType: 'module',
+  // },
+  plugins: ['@typescript-eslint', 'deprecation', 'import', 'jest', 'jsx-a11y', 'react', 'react-hooks'],
+  // plugins: ['@typescript-eslint', 'deprecation', 'import', 'jest', 'jsx-a11y', 'react', 'react-hooks'],
```
xugao, line 11: can remove line

scripts/eslint/v7.js
ecraig12345: Started from here and then modified a bunch of things https://github.com/typescript-eslint/tslint-to-eslint-config

scripts/eslint/v7.js
ecraig12345: This and the next comment are some of the things I meant when I said this is a work in progress. :) Wanted to keep this section while I'm trying out a different approach, but it will be removed before check-in for sure.

scripts/eslint/v7.js
ecraig12345: You might also be interested in this page which has a pretty detailed mapping of tslint to eslint rules. https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md

packages/react-button/src/utils/mergeProps.ts
```diff
@@ -39,6 +39,7 @@ const filterProps = (props: any): any => {
       break;
     case 'a':
       allowedProps = anchorProperties;
+      break;
```
ecraig12345, line 42: @dzearing The lint rules found a bug here

.github/CODEOWNERS
```diff
@@ -64,7 +64,6 @@ packages/react-hooks/ @ecraig12345
 packages/styling/ @dzearing
 packages/styling/src/interfaces/ @phkuo
 packages/styling/src/styles/ @phkuo
-packages/tslint-rules/ @ecraig12345
```
khmakoto, line 67: Should we add a codeowner for the `eslint-plugin` package?

apps/fabric-website/src/pages/Styles/OfficeBrandIconsPage/OfficeBrandIconsPage.tsx
```diff
@@ -173,7 +170,7 @@ function _otherSections(platform: Platforms): IPageSectionProps<Platforms>[] {
             <>
               <ul className={styles.iconList}>
                 {productIcons.map((icon, iconIndex) => (
-                  <li key={iconIndex}>
+                  <li key={icon.icon}>
```
khmakoto, line 173: Should this be:

```suggestion
                  <li key={icon.name}>
```

apps/perf-test/src/app/App.tsx
```diff
@@ -1,10 +1,9 @@
 import * as React from 'react';
-import { Dropdown, PrimaryButton, Stack, TextField, Text } from 'office-ui-fabric-react';
+import { Dropdown, PrimaryButton, Stack, TextField } from 'office-ui-fabric-react';
 import { Measurer, MeasurerTimings } from './Measurer';
-import { useTimer } from './useTimer';
 import { Scenarios } from './Scenarios';
 
-// tslint:disable
+/* eslint-disable react/jsx-no-bind */
```
khmakoto, line 6: Why are we disabling this here? I don't see any binds.

apps/perf-test/src/app/App.tsx
```diff
@@ -69,6 +68,7 @@ export const App = () => {
           ))}
         </div>
       )}
+      {/* eslint-enable react/forbid-component-props, react/jsx-no-bind */}
```
khmakoto, line 71: Seems weird that we are enabling `forbid-component-props` at the end of the file when it is disabled globally.

apps/vr-tests/src/stories/SliderNext.stories.tsx
```diff
@@ -1,11 +1,9 @@
-/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
```
khmakoto, line 1: We should probably delete these line from all vr tests 😆.

apps/fabric-website/src/pages/Styles/OfficeBrandIconsPage/OfficeBrandIconsPage.tsx
ecraig12345: Doesn't really matter since either one will be unique in this case (these are a custom type not IIconProps)

apps/perf-test/src/app/App.tsx
ecraig12345: The new rule also covers lambdas and other ways of passing functions which will mutate every render.

apps/perf-test/src/app/App.tsx
ecraig12345: This is re-enabling the rules after the area where they needed to be disabled is finished.

apps/vr-tests/src/stories/SliderNext.stories.tsx
ecraig12345: Pretty sure it's universally unnecessary, but I'd rather not add even more noise to this PR.

apps/perf-test/src/app/App.tsx
ecraig12345: Oh I see what you mean in this case. What I said was probably what was originally done with the tslint equivalents?

apps/perf-test/src/app/App.tsx
ecraig12345: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md

apps/perf-test/src/app/App.tsx
khmakoto: Probably.

packages/charting/src/components/StackedBarChart/examples/StackedBarChart.Dynamic.Example.tsx
```diff
@@ -57,12 +56,6 @@ export class StackedBarChartDynamicExample extends React.Component<{}, IExampleS
           { legend: 'fourth', data: 87, color: DefaultPalette.greenLight },
         ],
       },
-      colors: [
-        DefaultPalette.blueLight,
-        DefaultPalette.yellow,
-        DefaultPalette.neutralSecondary,
-        DefaultPalette.greenLight,
-      ],
```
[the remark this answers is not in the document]
ecraig12345: [`react/no-unused-state`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-state.md) detected that this state field was unused. (Thank you for reviewing btw!!!)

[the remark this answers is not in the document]
khmakoto: You're welcome! I don't promise I'll take a look at every file because this is massive but I'll try my best 😄

PR #13077, opened 2020-05-09
Updating OverflowSet to a function component inside React-Next

czearing:
#### Pull request checklist

#### Description of changes

Converted `OverflowSet`  to a function component inside React-Next

#### Focus areas to test
`OverflowSet`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13077)


packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -2,257 +2,221 @@ import * as React from 'react';
...
-
-  constructor(props: IOverflowSetProps) {
-    super(props);
-
-    initializeComponentRef(this);
-    warnMutuallyExclusive(COMPONENT_NAME, props, {
-      doNotContainWithinFocusZone: 'focusZoneProps',
+let Tag;
+let uniqueComponentProps;
+
```
ecraig12345, line 26: I think @xugao was doing some work related to keytips--does this conflict at all with that? And can/should it be moved to a more general location?

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -2,257 +2,221 @@ import * as React from 'react';
...
+      // Delete all persisted keytips saved
+      Object.keys(persistedKeytips).forEach((uniqueID: string) => {
+        // The "any" cast is required as a workaround due to colliding types.
+        keytipManager.unregister(persistedKeytips[uniqueID] as IKeytipProps, uniqueID, true);
+      });
+      persistedKeytips = {};
+    };
+  }, [persistedKeytips, keytipManager]);
+};
+
```
ecraig12345, line 47: Can/should this go in a more general location too?

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
xugao: Thanks for pointing this out Elizabeth. To answer you, this doesn't conflict with my changes directly. `OverflowSet` is the only component which directly uses `KeytipManger` (instead of `KeytipData`). 
So there is no need to make this code more generic for now. I will take a deeper look on if we can remove `KeytipManager` from `OverflowSet` completely

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -2,257 +2,221 @@ import * as React from 'react';
...
-    warnMutuallyExclusive(COMPONENT_NAME, props, {
-      doNotContainWithinFocusZone: 'focusZoneProps',
+let Tag;
+let uniqueComponentProps;
+
+const useKeytips = (persistedKeytips: { [uniqueID: string]: IKeytipProps }, keytipManager: KeytipManager) => {
+  React.useEffect(() => {
+    Object.keys(persistedKeytips).forEach((key: string) => {
+      const keytip = persistedKeytips[key];
+      // The "any" cast is required as a workaround due to colliding types.
```
xugao, line 31: do you mean by "The `as IKeytipProps` cast" ? Could you elaborate why this comment/the casting is needed

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -2,257 +2,221 @@ import * as React from 'react';
...
-
-    initializeComponentRef(this);
-    warnMutuallyExclusive(COMPONENT_NAME, props, {
-      doNotContainWithinFocusZone: 'focusZoneProps',
+let Tag;
+let uniqueComponentProps;
+
+const useKeytips = (persistedKeytips: { [uniqueID: string]: IKeytipProps }, keytipManager: KeytipManager) => {
+  React.useEffect(() => {
+    Object.keys(persistedKeytips).forEach((key: string) => {
```
xugao, line 29: nit: change `key: string` to `uniqueID: string`

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -2,257 +2,221 @@ import * as React from 'react';
...
-  constructor(props: IOverflowSetProps) {
-    super(props);
-
-    initializeComponentRef(this);
-    warnMutuallyExclusive(COMPONENT_NAME, props, {
-      doNotContainWithinFocusZone: 'focusZoneProps',
+let Tag;
+let uniqueComponentProps;
+
+const useKeytips = (persistedKeytips: { [uniqueID: string]: IKeytipProps }, keytipManager: KeytipManager) => {
```
xugao, line 27: pls rename `useKeytips` to `useKeytipRegistrations`, bc that's more precisely describing what this hook does

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -2,257 +2,221 @@ import * as React from 'react';
...
-  private _classNames: IProcessedStyleSet<IOverflowSetStyles>;
-
-  constructor(props: IOverflowSetProps) {
-    super(props);
-
-    initializeComponentRef(this);
-    warnMutuallyExclusive(COMPONENT_NAME, props, {
-      doNotContainWithinFocusZone: 'focusZoneProps',
+let Tag;
+let uniqueComponentProps;
```
xugao, line 25: these should be defined within the component, inside the closet scope where being used

ecraig12345:
@xugao Would you mind looking at this again now that Caleb has addressed the feedback?

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -1,257 +1,225 @@
...
-   */
-  public focus(forceIntoFirstElement?: boolean): boolean {
-    let focusSucceeded = false;
-
+    return () => {
+      // Delete all persisted keytips saved
+      Object.keys(persistedKeytips).forEach((uniqueID: string) => {
+        keytipManager.unregister(persistedKeytips[uniqueID] as IKeytipProps, uniqueID, true);
+      });
+      persistedKeytips = {};
```
ecraig12345, line 43: This only clears *this function's reference to* `persistedKeytips`, not the original object. To clear the original object, the easiest way would be to add a line inside the loop to do `delete persistedKeytips[uniqueID]`.

xugao:
found an issue with keytip:
1. go to pr deploy site [here](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/13077/merge/react-next/dist-storybook/index.html?path=/story/keytip--keytips-overflow-example)
2. click on "move overflow items" button
3. error occurs

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -1,257 +1,225 @@
...
+  focusZone: React.RefObject<IFocusZone>,
+  divContainer: React.RefObject<HTMLDivElement>,
+) => {
+  React.useImperativeHandle(
+    props.componentRef,
+    (): IOverflowSet => ({
+      focus: (forceIntoFirstElement?: boolean): boolean => {
+        let focusSucceeded = false;
+        // eslint-disable-next-line deprecation/deprecation
+        if (props.doNotContainWithinFocusZone) {
```
ecraig12345, line 59: We have a note in the `focusZoneProps` and `doNotContainWithinFocusZone` docs which says "In 8.0 the OverflowSet will no longer be wrapped in a FocusZone" (added by @micahgodbolt). I think we should go ahead and make that change now.

Removing these props makes the original comment I was going to leave here irrelevant, but the general issue is something to be careful about elsewhere: since this imperative handle has no dependencies, references to `props` will only ever refer to the *first* props passed in.

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -1,257 +1,225 @@
...
 
-  // Add keytip register/unregister handlers to lifecycle functions to correctly manage persisted keytips
-  public componentDidMount() {
-    this._registerPersistedKeytips();
-  }
-
-  public componentWillUnmount() {
-    this._unregisterPersistedKeytips();
-  }
+  const showOverflow = overflowItems && overflowItems.length > 0;
```
ecraig12345, line 111:
```suggestion
  const showOverflow = !!overflowItems && overflowItems.length > 0;
```

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -1,257 +1,225 @@
...
-        <div key={item.key} className={this._classNames.item}>
-          {this.props.onRenderItem(item)}
-        </div>
-      );
-    });
-  };
-
-  private _onRenderOverflowButtonWrapper = (items?: IOverflowSetItemProps[]): JSX.Element => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const onRenderOverflowButtonWrapper = (itemsProp: any[]): JSX.Element => {
```
ecraig12345, line 114: Can we use the proper type `IOverflowSetItemProps` here (and a couple lines down)?

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -1,257 +1,225 @@
...
+              keytipManager,
+              overflowKeytipSequences,
+              overflowItem.keytipProps.keySequences,
+            );
           } else {
-            // Nothing to change, add overflowItem to list
-            newOverflowItems.push(overflowItem);
+            // If the keytip doesn't have a submenu, just execute the original function
+            persistedKeytip.onExecute = keytip.onExecute;
           }
```
ecraig12345, line 144: Small thing but please bring back the spaces between lines; they do help with readability.

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -1,257 +1,225 @@
...
       return item.subMenuProps.items;
     }
     return undefined;
+  };
+
+  useConst(() => {
+    warnMutuallyExclusive(COMPONENT_NAME, props, {
+      doNotContainWithinFocusZone: 'focusZoneProps',
+    });
+  });
```
ecraig12345, line 187: Do `useWarnings` instead (wrapped in `if (process.env.NODE_ENV !== 'production')`)

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -1,257 +1,225 @@
...
+  useComponentRef(props, focusZone, divContainer);
+
+  return (
+    <Tag
+      role={role || 'group'}
+      aria-orientation={role === 'menubar' ? (vertical === true ? 'vertical' : 'horizontal') : undefined}
+      {...uniqueComponentProps}
+      className={classNames.root}
+    >
+      {overflowSide === 'start' && showOverflow && onRenderOverflowButtonWrapper(overflowItems!)}
```
xugao, line 214: I believe here is wha's causing the hook error. let's move `onRenderOverflowButtonWrapper` to its own FC (call it `OverflowButton`) outside of `OverflowSetBase`

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -1,257 +1,225 @@
...
+            },
+          };
+          newOverflowItems.push(newOverflowItem);
+        } else {
+          // Nothing to change, add overflowItem to list
+          newOverflowItems.push(overflowItem);
+        }
+      });
+    } else {
+      newOverflowItems = itemsProp;
```
ecraig12345, line 163: This may be a behavior change; do you know for sure that `itemsProp` is always defined?

packages/react-next/src/components/OverflowSet/OverflowSet.types.ts
```diff
@@ -31,7 +29,7 @@ export interface IOverflowSet {
 /**
  * {@docCategory OverflowSet}
  */
-export interface IOverflowSetProps extends React.ClassAttributes<OverflowSetBase> {
+export interface IOverflowSetProps extends React.ClassAttributes<HTMLElement> {
```
ecraig12345, line 32: This should probably use RefAttributes now that it's not a class component.

packages/react-next/src/components/OverflowSet/OverflowButton.base.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }, [persistedKeytips, keytipManager]);
+};
+
+export const OverflowButtonBase = (props: IOverflowSetProps) => {
+  const keytipManager: KeytipManager = KeytipManager.getInstance();
+  const { className, styles, vertical, overflowItems, keytipSequences } = props;
+  const classNames: IProcessedStyleSet<IOverflowSetStyles> = getClassNames(styles, { className, vertical });
+
+  const internalState = useConst<IOverflowInternalState>(() => ({
+    persistedKeytips: {},
```
ecraig12345, line 52: This does need to be persisted, but since it's the only piece of persisted data (see comments below) you can just do `const persistedKeytips = useConst</*type here*/>({});` and get rid of the state wrapper.

packages/react-next/src/components/OverflowSet/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './OverflowSet';
 export * from './OverflowSet.base';
 export * from './OverflowSet.types';
+export * from './OverflowButton.base';
```
ecraig12345, line 4: This shouldn't be exported since it's just an internal implementation detail.
```suggestion
```

packages/react-next/src/components/OverflowSet/OverflowButton.tsx
```diff
@@ -0,0 +1,5 @@
+import { getStyles } from './OverflowSet.styles';
+import { styled } from '../../Utilities';
+import { OverflowButtonBase } from './OverflowButton.base';
+
+export const OverflowButton = styled(OverflowButtonBase, getStyles, undefined, { scope: 'OverflowButton' });
```
xugao, line 5: The styles for OverflowButton is off because the we are missing passing the className in Base (by calling `getClassNames`)

here is my suggestion:
Remove this file (styled `OverflowButton`, and just use `OverflowButtonBase` from `OverflowSet`) but make sure pass it the className)

packages/react-next/src/components/OverflowSet/OverflowButton.base.tsx
```diff
@@ -0,0 +1,105 @@
...
+      // Checks if itemSubMenuProvider has been defined, if not defaults to subMenuProps
+      if (props.itemSubMenuProvider) {
+        return props.itemSubMenuProvider(item);
+      }
+      if (item.subMenuProps) {
+        return item.subMenuProps.items;
+      }
+      return undefined;
+    },
+    [props],
```
xugao, line 46: always put only what's absolutely required here to avoid unnecessary cache busting:
```
    [itemSubMenuProvider],
```

packages/react-next/src/components/OverflowSet/OverflowButton.base.tsx
```diff
@@ -0,0 +1,106 @@
...
+      // Checks if itemSubMenuProvider has been defined, if not defaults to subMenuProps
+      if (props.itemSubMenuProvider) {
+        return props.itemSubMenuProvider(item);
+      }
+      if (item.subMenuProps) {
+        return item.subMenuProps.items;
+      }
+      return undefined;
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
xugao, line 46: you should not need to disable the rule here.

instead using `props. itemSubMenuProvider` inside the callback, destruct the prop above
```
const { className, overflowItems, keytipSequences, itemSubMenuProvider } = props;
```
and use everywhere here

PR #13078, opened 2020-05-09
Updating Slider to a function component inside React-Next

czearing:
#### Pull request checklist

#### Description of changes

Converted `Slider` to a function component inside React-Next

#### Focus areas to test
`Slider`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13078)


packages/react-next/src/components/Slider/Slider.test.tsx
```diff
@@ -220,87 +149,7 @@ describe('Slider', () => {
     }, ONKEYDOWN_TIMEOUT_DURATION);
 
     jest.runOnlyPendingTimers();
-  });
-
-  it('does not update the value when slider is controlled', () => {
```
joschect, line 225: It looks like a lot of these test have been removed, did they not work? Should they be removed?

packages/react-next/src/components/Slider/Slider.base.tsx
```diff
@@ -1,408 +1,327 @@
...
+    min = 0,
+    showValue = true,
+    buttonProps = {},
+    vertical = false,
+    valueFormat,
+    styles,
+    theme,
+    originFromZero = false,
+  } = props;
+  const [value, setValue] = React.useState(
```
joschect, line 51: I believe that this will break the controlled slider scenario. If you use `useControllableValue` hook I think that should resolve this. You can also pass the onChange events into that hook so it will manage those as well.

joschect: It also returns a callback so you can use that instead of setValue below!

joschect:
@czearing Awesome job! I think it might break the controlled slider but there are just a few tweaks that should unblock that scenario! Thanks for submitting this 👍

joschect:
Retriggering one more time

PR #13079, opened 2020-05-09
chore(Segment): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `Segment` component to be functional. Restricting props set that will be passed to styles functions.

Related to #12237

## Prop sets

| `Segment`    |
| --------- |
| `color` |
| `inverted` |
| `disabled` |

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13079)


assuncaocharles:
@layershifter I am not sure about the prop

```typescript
rtlAttributes: typeof rtlTextContainer.getAttributes;
```

add it there because it was complaining since it's in the `Segment.PropTypes`

packages/fluentui/react-northstar/src/components/Segment/Segment.tsx
```diff
@@ -23,51 +32,86 @@ export interface SegmentProps
...
+      className,
+      design,
+      styles,
+      variables,
+    }),
+    rtl: context.rtl,
+  });
+
+  const element = (
+    <ElementType
```
jurokapsiar, line 80: don't forget to wrap in focuszone

layershifter: I don't think that it's possible to use FZ in `Segment` 🤔

assuncaocharles: Done

packages/fluentui/react-northstar/src/components/Segment/Segment.tsx
```diff
@@ -23,51 +32,86 @@ export interface SegmentProps
...
-  accessibility?: Accessibility;
+  accessibility?: Accessibility<never>;
 
   /** An segment can show it is currently unable to be interacted with. */
   disabled?: boolean;
 
   /** A segment can have its colors inverted for contrast. */
   inverted?: boolean;
+
+  rtlAttributes?: typeof rtlTextContainer.getAttributes;
```
pompomon, line 43: Please why do you need this new prop? I don't see where we use it anywhere in repo

assuncaocharles: Only added here because it was on the proptypes.... Will remove from both then

assuncaocharles, line 43:
```suggestion
```

packages/fluentui/react-northstar/src/components/Segment/Segment.tsx
```diff
@@ -23,51 +32,85 @@ export interface SegmentProps
...
+Segment.displayName = 'Segment';
+
+Segment.propTypes = {
+  ...commonPropTypes.createCommon({
+    content: 'shorthand',
+    color: true,
+  }),
+  disabled: PropTypes.bool,
+  inverted: PropTypes.bool,
+  rtlAttributes: PropTypes.func,
```
assuncaocharles, line 103:
```suggestion
```

packages/fluentui/react-northstar/src/components/Segment/Segment.tsx
```diff
@@ -23,51 +32,84 @@ export interface SegmentProps
...
+    content: 'shorthand',
+    color: true,
+  }),
+  disabled: PropTypes.bool,
+  inverted: PropTypes.bool,
+};
+Segment.handledProps = Object.keys(Segment.propTypes) as any;
+
+Segment.defaultProps = {
+  as: 'div',
```
layershifter, line 107: This is redundant, `as` defaults to `div`

assuncaocharles: Removed

PR #13086, opened 2020-05-11
fix(CarouselNavigation): enable back focus zone in carousel navigation 

kolaps33:
Arrow keys navigation in the carousel navigation list stops to work.

#### Description of changes
Root cause was that focus zone was not adding to the tablist after convert to FC
Adding function which will allow it. 

#### Focus areas to test
Navigate under the carousel into the navigation tablist
Use left/right arrow keys
Expected Result
Left/Right arrow keys navigate to previous/next tab element (bullet) 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13086)


packages/fluentui/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixes
+- `CarouselNavigation` enable back focus zone in carousel navigation @kolaps33 ([#13086](https://github.com/microsoft/fluentui/pull/13086))
```
silviuaavram, line 21: why not `Enable back focus zone in `CarouselNavigation`` ?

PR #13088, opened 2020-05-11
fix(Dropdown): prevents keydown propagation

assuncaocharles:
#### Pull request checklist

#### Description of changes

When typing in a searchable Dropdown, clicking the ESC key should clear the search (which it does).

However, if the Dropdown is a child to a component that also has ESC key behavior, the ESC key event bubbles to the parent component. For example, if the Dropdown is in a Dialog, the ESC key clears the Dropdown and closes the Dialog.

Also, when Dropdown is child of Accordion, arrow keys keydown events propagation cause the accordion panels to switch.

This PR prevents propagation on key down in the dropdown. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13088)


packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1099,6 +1099,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     const { disabled } = this.props;
 
     const handleInputKeyDown = (e: React.SyntheticEvent, searchInputProps: DropdownSearchInputProps) => {
+      e.stopPropagation();
```
jurokapsiar, line 1102: let's just stop propagation when the key was handled - for Tab, ArrowLeft, ArrowRight, Backspace

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -229,8 +229,7 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
     // and ESC is pressed, the opened Dialog should get closed and the trigger should get focus
     const lastOverlayRef = getRefs().pop();
     const isLastOpenedDialog: boolean = lastOverlayRef && lastOverlayRef.current === this.overlayRef.current;
-
-    if (keyboardKey.getCode(e) === keyboardKey.Escape && isLastOpenedDialog) {
+    if (keyboardKey.getCode(e) === keyboardKey.Escape && isLastOpenedDialog && this.overlayRef.current === e.target) {
```
layershifter, line 232: As was mentioned in chat this will not work as `this.overlayRef.current === e.target` will be always `false`

assuncaocharles: Solved

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
assuncaocharles: Done

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -229,7 +229,8 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
     // and ESC is pressed, the opened Dialog should get closed and the trigger should get focus
     const lastOverlayRef = getRefs().pop();
     const isLastOpenedDialog: boolean = lastOverlayRef && lastOverlayRef.current === this.overlayRef.current;
-    if (keyboardKey.getCode(e) === keyboardKey.Escape && isLastOpenedDialog && this.overlayRef.current === e.target) {
+    const targetIsBody = (e.target as HTMLElement).nodeName === 'BODY';
+    if (targetIsBody && keyboardKey.getCode(e) === keyboardKey.Escape && isLastOpenedDialog) {
```
layershifter, line 234:
```suggestion

    if (targetIsBody && keyboardKey.getCode(e) === keyboardKey.Escape && isLastOpenedDialog) {
```

nit

layershifter:
Let's add an E2E test to be sure that it works.

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -229,8 +229,10 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
     // and ESC is pressed, the opened Dialog should get closed and the trigger should get focus
     const lastOverlayRef = getRefs().pop();
     const isLastOpenedDialog: boolean = lastOverlayRef && lastOverlayRef.current === this.overlayRef.current;
+    const targetIsBody = (e.target as HTMLElement).nodeName === 'BODY';
 
-    if (keyboardKey.getCode(e) === keyboardKey.Escape && isLastOpenedDialog) {
+
```
layershifter, line 234:
```suggestion

```

nit: Sorry, please restore that change, my mistake I overlooked line numbers 😨

assuncaocharles:
> Let's add an E2E test to be sure that it works (with nested `Dialog`s, too).
> 
> I believe that we have similar issue in `Popup`

Tests added

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1102,21 +1102,27 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
       if (!disabled) {
         switch (keyboardKey.getCode(e)) {
           case keyboardKey.Tab:
+            e.stopPropagation();
             this.handleTabSelection(e, highlightedIndex, selectItemAtIndex, toggleMenu);
             break;
           case keyboardKey.ArrowLeft:
+            e.stopPropagation();
```
pompomon, line 1109: Should we add that to Breaking changes list, in case someone was previously listening to keypress events on dropdown?

assuncaocharles: Added

PR #13093, opened 2020-05-11
fix(Components): add Focus Zone wrapper

assuncaocharles:
#### Pull request checklist

#### Description of changes

Some components missed the `unstable_wrapWithFocusZone` wrapper when converted to RFC. This PR adds revert that adding the wrapper in the proper components. 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13093)


jurokapsiar:
Please add changelog before merging

layershifter:
If it's a case I suggest to add wrapper everywhere (as it was before) and test it via additional test `isConformant()`.

assuncaocharles:
> If it's a case I suggest to add wrapper everywhere (as it was before) and test it via additional test `isConformant()`.

@layershifter All components on this PR are already tested by `isConformant` your suggestion is to go through all FC and wrap them with FZ?

layershifter:
Yep, let's move this discussion to offline and decide.

PR #13095, opened 2020-05-11
Copying Link to react-next, in preparation for converting to function component

czearing:
#### Pull request checklist

This change does not modify Link; it simply moves it to react-next for better diffing in the actual conversion PR.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13095)

### DIFF
--- a/packages/react-next/src/components/Link/Link.doc.tsx
+++ b/packages/react-next/src/components/Link/Link.doc.tsx
@@ -1,0 +1,57 @@
+import * as React from 'react';
+import { LinkBasicExample } from './examples/Link.Basic.Example';
+
+import { IDocPageProps } from '../../common/DocPage.types';
+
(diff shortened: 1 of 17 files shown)


packages/react-next/src/components/Link/Link.doc.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+import { LinkBasicExample } from './examples/Link.Basic.Example';
+
+import { IDocPageProps } from '../../common/DocPage.types';
+
+const LinkBasicExampleCode = require('!raw-loader!office-ui-fabric-react/src/components/Link/examples/Link.Basic.Example.tsx') as string;
```
khmakoto, line 6:
```suggestion
const LinkBasicExampleCode = require('!raw-loader!@fluentui/react-next/src/components/Link/examples/Link.Basic.Example.tsx') as string;
```

PR #13096, opened 2020-05-11
Adding @fluentui/react-button package scaffolding

dzearing:
This PR sets up the initial scaffolding for the react-button package. The intent here is:

* This package will contain the shared Button component
* The PR comes with a reference implementation which may or may not be used in the final product.
* It supports v0 slots
* It uses compose from `@fluentui/react-compose`
* It uses static css 
* It exports both an unstyled `BaseButton` and a recomposed `Button` component which is styled.

What this does not cover:

* Need to add all the tests
* Need to add useFocusRects
* Fix remaining styling issues which still aren't parity with v0
* Need to test with ThemeProvider
* The `tokens` prop is still not provided

There are some tweaks in react-compose needed for passing along class maps. I will separate these out in a separate PR, but need them for this PR.

There is a temporary mergeProps helper for the hook prop parsing. This will move out separately, but added it here to get the package going.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13096)


packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -0,0 +1,84 @@
+import * as React from 'react';
+import { ComponentProps, ShorthandValue, ComposeOptions, ComposeStandardStatics } from '../../temp/tempTypes';
+
+export type SizeValue = 'smallest' | 'smaller' | 'small' | 'medium' | 'large' | 'larger' | 'largest';
+
+// tslint:disable-next-line:no-any
```
khmakoto, line 6: Remove.

packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -0,0 +1,84 @@
...
+  /** A button can show that it cannot be interacted with. */
+  disabled?: boolean;
+
+  /** A button can fill the width of its container. */
+  fluid?: boolean;
+
+  /** A button can contain only an icon. */
+  iconOnly?: boolean;
+
+  /** An icon button can format its Icon to appear before or after its content */
```
khmakoto, line 36:
```suggestion
  /** An icon button can format its Icon to appear before or after its content. */
```

packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -0,0 +1,84 @@
...
+  iconPosition?: 'before' | 'after';
+
+  /** A button that inherits its background and has a subtle appearance */
```
khmakoto, line 39:
```suggestion
  /** A button that inherits its background and has a subtle appearance. */
```

packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  /** A button can show a loading indicator. */
+  loading?: boolean;
+
+  /**
+   * Called after a user clicks the button.
+   * @param event - React's original SyntheticEvent.
+   * @param data - All props.
+   */
+  // onClick?: ComponentEventHandler<ButtonProps>;
```
khmakoto, line 50: Are these commented because you're getting them from the `HTMLAttributes`?

packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+export interface ButtonSlots {
+  icon: React.ElementType;
+  content: React.ElementType;
+  loader: React.ElementType;
+}
+
+export type ButtonSlotProps = {
+  [key in keyof ButtonSlots]: ButtonProps[key];
```
khmakoto, line 80: Not to do in this PR but we might want to create some helper types for things like this.

packages/react-button/src/components/Button/ButtonBase.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const ButtonBase = compose<'button', ButtonProps, ButtonProps, {}, {}>(
+  (props, ref, composeOptions) => {
+    const { slots, slotProps } = useButton(props, composeOptions);
+
+    return (
+      <slots.root ref={ref} {...slotProps.root}>
+        {props.loading && <slots.loader {...slotProps.loader} />}
+        {props.icon && props.iconPosition !== 'after' && <slots.icon {...slotProps.icon} />}
+        {!props.iconOnly && props.content && <slots.content {...slotProps.content} />}
+        {props.icon && props.iconPosition === 'after' && <slots.icon {...slotProps.icon} />}
```
khmakoto, line 15: Should all of this be inside a condition where `!props.loading`?

packages/react-compose/src/utils.ts
```diff
@@ -29,6 +35,16 @@ export const defaultComposeOptions: ComposePreparedOptions = {
   resolveSlotProps: () => ({}),
 };
 
+const mergeClasses = (classes1: ClassDictionary | undefined, classes2: ClassDictionary | undefined) => {
```
xugao, line 38: can we use `function` signature so return type is required

packages/react-compose/src/types.ts
```diff
@@ -47,6 +47,7 @@ export type ComposeRenderFunction<T extends React.ElementType = 'div', P = {}> =
 
 export type ComposeOptions<InputProps = {}, InputStylesProps = {}, ParentStylesProps = {}> = {
   className?: string;
+  classes?: ClassDictionary;
```
xugao, line 50: can we keep these two in sync in terms of naming?
either `class` + `classes` or `className` + `classNames`

packages/react-compose/src/utils.ts
```diff
@@ -29,6 +35,16 @@ export const defaultComposeOptions: ComposePreparedOptions = {
...
+  const result = { ...classes1 };
+
+  if (classes2) {
+    Object.keys(classes2).forEach(name => (result[name] = cx(result[name], classes2[name])));
+  }
+  console.log(result);
+  return result;
+};
+
 export function mergeComposeOptions(
```
xugao, line 48: no tests?

packages/react-button/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": ["@uifabric/tslint-rules"],
+  "rules": {
+    "interface-name": false
```
xugao, line 4:
```suggestion
    "interface-name": [true, "never-prefix"]
```

packages/react-button/src/components/Button/useButton.ts
```diff
@@ -0,0 +1,11 @@
+import { ButtonProps, ButtonSlots, ButtonSlotProps } from './Button.types';
+import { mergeProps } from '../../temp/mergeProps';
```
xugao, line 2: this is missing

packages/react-compose/src/utils.ts
dzearing: This is in a separate PR (with tests :)

dzearing: https://github.com/microsoft/fluentui/pull/13101

packages/react-compose/src/types.ts
dzearing: `className` will be removed I believe, so there'd only be a classes map.

packages/react-compose/src/utils.ts
dzearing: see other pr.

dzearing: https://github.com/microsoft/fluentui/pull/13101

packages/react-button/src/components/Button/ButtonBase.tsx
dzearing: Need to double check. I think so.

packages/react-button/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,14 @@
+import { ButtonBase } from './ButtonBase';
+import { compose } from '@fluentui/react-compose';
+import * as classes from './Button.scss';
+import { ButtonProps } from './Button.types';
+
+console.log(classes);
```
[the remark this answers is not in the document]
dzearing: done

packages/react-button/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <ButtonVariants size="large" />
+      <ButtonVariants size="larger" />
+      <ButtonVariants size="largest" />
+    </div>
+    {/*
+    <h3>A button can be formatted to show only text in order to indicate a less-pronounced action.</h3>
+    <div>
+      <ButtonVariants text />
+    </div>
+    */}
```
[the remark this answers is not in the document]
dzearing: done

packages/react-button/src/components/Button/Button.scss
```diff
@@ -0,0 +1,244 @@
...
+ * defined as literals in the same scope. If you refer to variables,
+ * they will be resolved on the :root element only; not at usage time.
+ * That means referencing things intended to be defined contextually
+ * through ThemeProvider would not resolve correctly (unless the provider
+ * defined variablesif you fall back hovered text to rested text, it won't
+ * change if the user overrides rested text.
+**/
+:root {
+  /* sizing */
+  --button-size-smallest: 20px;
```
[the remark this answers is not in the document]
dzearing: the only v0 numbers are small and regular; everything else is not defined. We'll probably want to revisit what the size should map to by default with designers.

packages/react-button/README.md
```diff
@@ -0,0 +1,11 @@
+# @fluentui/react-button
```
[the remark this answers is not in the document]
dzearing: That could be done separately. It would be ideal to add the spec content there.

packages/react-button/src/components/Button/useButton.ts
dzearing: this seems to be because git ignores folders called "temp". Renamed to utils, but it's a folder of temporary utils which will be moved out after other PRs merge.

packages/react-button/src/components/Button/ButtonBase.tsx
dzearing: v0 code still renders the icon even when loading is set.

packages/react-button/src/components/Button/Button.types.tsx
dzearing: For sure!

packages/react-button/src/components/Button/Button.types.tsx
dzearing: Not quite. I didn't understand it at first. I now do, but it's not supported yet.

```tsx
  /**
   * Called after a user clicks the button.
   * @param event - React's original SyntheticEvent.
   * @param data - All props.
   */
  onClick?: ComponentEventHandler<ButtonProps>;
```

Notice the 2nd param, data. This requires we always override `onClick` (and `onFocus`) to provide the 2nd param.

dzearing: It's convenient to be able to access props from the `onClick` but it certainly requires some factory magic to standardize. I'm still a little mixed on it.

packages/react-button/src/components/Button/Button.stories.tsx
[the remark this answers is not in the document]
dzearing: done

packages/react-button/src/components/Button/Button.scss
```diff
@@ -0,0 +1,244 @@
+/**
+ * TODO: Once theming is updated, consider removing all of this.
+ *
+ * Note: All :root rules must be literals, OR references to variables
+ * defined as literals in the same scope. If you refer to variables,
+ * they will be resolved on the :root element only; not at usage time.
+ * That means referencing things intended to be defined contextually
+ * through ThemeProvider would not resolve correctly (unless the provider
+ * defined variablesif you fall back hovered text to rested text, it won't
```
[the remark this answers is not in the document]
dzearing: fixed.

packages/react-button/README.md
```diff
@@ -8,4 +8,8 @@ To import React Button components:
 
 ```js
 import { ComponentName } from '@fluentui/react-button';
```
khmakoto, line 10:
```suggestion
import { Button } from '@fluentui/react-button';
```

packages/react-button/jest.config.js
```diff
@@ -0,0 +1,12 @@
+const { createConfig, resolveMergeStylesSerializer } = require('@uifabric/build/jest/jest-resources');
```
ecraig12345, line 1: Not sure you need the mergeStyles part here

packages/react-button/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "start": "just-scripts dev:storybook",
+    "start-test": "just-scripts jest-watch",
+    "test": "just-scripts test",
+    "update-api": "just-scripts update-api",
+    "update-snapshots": "just-scripts jest -u"
+  },
+  "devDependencies": {
+    "@types/enzyme": "3.10.3",
+    "@types/enzyme-adapter-react-16": "1.0.3",
+    "@types/es6-promise": "0.0.32",
```
ecraig12345, line 31: Remove this and probably also es6-weak-map (and probably remove them from the template too)

packages/react-button/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "react": "16.8.6",
+    "react-app-polyfill": "~1.0.1",
+    "react-dom": "16.8.6",
+    "react-test-renderer": "^16.3.0"
+  },
+  "dependencies": {
+    "@uifabric/set-version": "7.0.11",
+    "@fluentui/react-compose": "^0.2.11",
+    "@uifabric/react-hooks": "^7.3.1",
+    "@uifabric/utilities": "^7.16.2",
```
ecraig12345, line 51: Remember that uifabric/fluentui package versions will need to be updated before merging, and be careful about the timing--on one of the previous PRs I think you got a green build the night before and then merged the next morning after a publish, which broke builds after merge because the versions were mismatched.

khmakoto, line 51: nit: Order the packages alphabetically
```suggestion
    "@fluentui/react-compose": "^0.2.11",
    "@uifabric/react-hooks": "^7.3.1",
    "@uifabric/set-version": "7.0.11",
    "@uifabric/utilities": "^7.16.2",
```

packages/react-button/src/components/Button/Button.types.tsx
dzearing: I'll leave it there, lets come back to this.

packages/react-button/package.json
dzearing: eek! Thanks for the heads up.

dzearing:
@msft-github-bot merge as soon as possible

dzearing:
@msft-github-bot merge in 0 seconds

PR #13101, opened 2020-05-12
compose: Adding `classes` option to compose.

dzearing:
This change adds `classes` prop to `compose` so that a list of classes can be provided on options.

Quick example:

```tsx
import { ButtonBase } from '@fluentui/react';
import { compose } from '@fluentui/react-compose';
import * as classes from './MyButton.scss';

export const MyButton = compose(ButtonBase, { classes });
```

In a separate PR, I would like to add state resolving functionality here as well, rather than only accepting `ClassDictionary` types. This would require it to be resolved in `resolveSlotProps` or some identical callback which receives state (or props, if there is no state.) I'd propose we rename this callback to simply `mergeProps` and let it resolve slotProps, slots, and classes. This would enable us to resolve classes even with the current approach of resolving classes using css in js and caching on specific parts of the state:

```tsx
// Grab merge-styles or fela based css-in-js styles.
import { getStyles } from './Button.styles';

// Create a style function which will cache results based on limited inputs pulled from state.
// This means compose has no need for mapPropsToStylesProps, and that useStyles can stay
// out of the component itself and be resolved from within the class factory.
const classes = createClassFactory(getStyles, [ /* deps from state */ 'foo', 'bar', 'baz' ]);

compose(ButtonBase, { classes });
```

Feedback welcome!

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13101)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,7 @@
       "stopOnEntry": false,
       "args": ["-i", "--watch"],
       "runtimeExecutable": null,
-      "runtimeArgs": ["--nolazy", "--debug"],
+      "runtimeArgs": ["--nolazy", "--inspect"],
       "env": {
         "NODE_ENV": "development"
       },
--- a/change/@fluentui-react-compose-2020-05-11-18-24-56-feat-compose-classes.json
+++ b/change/@fluentui-react-compose-2020-05-11-18-24-56-feat-compose-classes.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Adding `classes` option to compose options.",
+  "packageName": "@fluentui/react-compose",
+  "email": "dzearing@microsoft.com",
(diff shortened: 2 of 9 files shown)


dzearing:
@mnajdova @layershifter Can you take a look and make sure this is the right change?

Adding this should allow the caller to provide any style system they need, whether it be merge-styles, fela, raw css, sass, etc. There should be no need to call useStylesheet, and should also set us up to move useStyles out of the component into styles funcrtion attached to it.

dzearing:
I'd like to merge this as it's blocking usage elsewhere in the convergence work. I will follow up with another PR to add functional support to `classes`, which should enable us to use it in css in js scenarios. If there's anything different you'd like to see I'm happy to incorporate it in that PR.

PR #13102, opened 2020-05-12
Bug fix to handle removal of duplicates in UnifiedPicker (SelectedItemsList)

nebhatna:
#### Pull request checklist

#### Description of changes
Updating the logic to remove selected items by not using .filter, as we end up removing all the duplicate items whereas we want to just remove that specific item. Updated the key in SelectedItemsList to be unique to handle duplicates for UnifiedPicker (selecteditemslist). And fixing minor typos
#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13102)

### DIFF
--- a/packages/experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
+++ b/packages/experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
@@ -14,7 +14,14 @@ const _SelectedItemsList = <TItem extends BaseSelectedItem>(
   }, [props.selectedItems]);
 
   const removeItems = (itemsToRemove: TItem[]): void => {
-    updateItems(items.filter(item => itemsToRemove.indexOf(item) === -1));
+    // Intentionally not using .filter here as we want to only remove a specific
+    // item in case of duplicates of same item.
+    const updatedItems: TItem[] = [...items];
+    itemsToRemove.forEach(item => {
+      const index: number = updatedItems.indexOf(item);
+      updatedItems.splice(index, 1);
+    });
+    updateItems(updatedItems);
     props.onItemsRemoved ? props.onItemsRemoved(itemsToRemove) : null;
   };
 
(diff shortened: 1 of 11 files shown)


nebhatna:
I have mistakenly selected the wrong option in the Contribution License Agreement, is there a way to update it? I tried re-creating this PR but looks like its read only now.

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
```diff
@@ -14,7 +14,12 @@ const _SelectedItemsList = <TItem extends BaseSelectedItem>(
   }, [props.selectedItems]);
 
   const removeItems = (itemsToRemove: TItem[]): void => {
-    updateItems(items.filter(item => itemsToRemove.indexOf(item) === -1));
+    const updatedItems: TItem[] = [...items];
+    itemsToRemove.forEach(item => {
```
amyngu, line 18: Let's make a comment here why we did it this way instead of the way it was originally (so anyone coming in won't change it without knowing the context)

amyngu:
The fix looks good to me! Would like some comments since this is open source that others will not go back and revert it. :)

packages/experiments/src/components/UnifiedPicker/hooks/useSelectedItems.ts
```diff
@@ -55,9 +55,12 @@ export const useSelectedItems = <T extends {}>(
   };
 
   const removeItems = (itemsToRemove: any[]): void => {
-    const currentitems: T[] = [...items];
-    const updatedItems: T[] = currentitems.filter(item => itemsToRemove.indexOf(item) === -1);
-
+    const currentItems: T[] = [...items];
```
khmakoto, line 58: If we add the comment as suggested above we should also add a comment over here indicating why we chose to do this in this way instead of how we were doing before.

PR #13103, opened 2020-05-12
Updates the just build to latest, adding rtl support to scss.

dzearing:
Thanks Xu/Mak/Elizabeth/Ken for getting the Just updates reviewed and merged. This brings them over.

Validated locally that scss directional rules now get auto flipped.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13103)


dzearing:
The bundles affected are due to directional flipping being staticly defined rather than dynamically defined. The particular regressions occur in those components because they still rely on scss, yet they likely have RTL bugs as the rules are assuming LTR and now have the rules split out.

It's interesting and worth some discussion. We are seeing a tradeoff: 

* You get the ability to externalize css into a file
* You give up the ability to dynamically alter that css sheet at runtime, and so you would accrue bundle size penalties by defining both directions.

khmakoto:
This is also missing a change file.

PR #13109, opened 2020-05-12
fix(Dropdown): Rendering no matches message when all items selected

assuncaocharles:
#### Pull request checklist

#### Description of changes

Show no matches found message when all items are selected in the `Dropdown`

Screenshot 2020-05-12 at 10 02 20

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13109)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
+++ b/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
@@ -824,7 +824,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
       };
     }
 
-    if (items && (items.length === 0 || items.length === value.length)) {
+    if (filteredItems && filteredItems.length === 0) {
       return {
         children: () =>
           DropdownItem.create(noResultsMessage, {
--- a/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
+++ b/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
@@ -809,7 +809,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
   }
 
   renderItemsListFooter(styles: ComponentSlotStylesInput) {
-    const { loading, loadingMessage, noResultsMessage, items } = this.props;
+    const { loading, loadingMessage, noResultsMessage } = this.props;
     const { filteredItems } = this.state;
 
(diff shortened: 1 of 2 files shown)


packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -810,6 +810,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
 
   renderItemsListFooter(styles: ComponentSlotStylesInput) {
     const { loading, loadingMessage, noResultsMessage, items } = this.props;
+    const { value } = this.state;
```
pompomon, line 813:
```suggestion
    const { filteredItems } = this.state;
```

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -823,7 +824,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
       };
     }
 
-    if (items && items.length === 0) {
+    if (items && (items.length === 0 || items.length === value.length)) {
```
pompomon, line 827:
```suggestion
    if (filteredItems && filteredItems.length === 0) {
```

_nit_

pompomon: I think it would work the same, and without additional condition

pompomon, line 827: Please could you add a test for that new case?

assuncaocharles: Done

assuncaocharles: Done

MaheshakaMudli:
I am facing the same issue. I am using the control in creatng teams messaging extension and the dropdown is not showing the "no matches" text when the user enters something or selects all the options from the dropdown and then searches for any text.
 
image

There is blank ul element created below the dropdown but it does not have any element created in it.

assuncaocharles:
Hey @MaheshakaMudli 

Can you open a new issue and provide an codesandbox with small reproduction?

PR #13110, opened 2020-05-12
fix(RadioGroup): Update examples to enable input for in the tab order inside Radio Group

assuncaocharles:
#### Pull request checklist

#### Description of changes

Update examples to enable input for in the tab order inside Radio Group

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13110)

### DIFF
--- a/packages/fluentui/accessibility/src/behaviors/Radio/radioGroupItemBehavior.ts
+++ b/packages/fluentui/accessibility/src/behaviors/Radio/radioGroupItemBehavior.ts
@@ -31,12 +31,6 @@ const radioGroupItemBehavior: Accessibility<RadioGroupItemBehaviorProps> = props
       },
     },
   },
-
-  focusZone: {
-    props: {
-      shouldEnterInnerZone: event => keyboardKey.getCode(event) === keyboardKey.Tab,
-    },
-  },
 });
 
 export default radioGroupItemBehavior;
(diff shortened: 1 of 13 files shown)


silviuaavram:
This example is just so weird. We removed the input from the tab order just to put some custom keyboard handling in this PR.

Why was the input removed from the tab order from the first place?

assuncaocharles:
> This example is just so weird. We removed the input from the tab order just to put some custom keyboard handling in this PR.
> 
> Why was the input removed from the tab order from the first place?
> 
> Edit: maybe because we are focusing the whole wrapper on tab instead of the radio button. Which is not a radio button. This looks just so different from what an `<input type="radio">` is.

I was also wondering if it was really a problem with the behavior or just a matter of the example that explicitly remove the input from tab order

packages/fluentui/accessibility/src/behaviors/Radio/radioGroupItemBehavior.ts
```diff
@@ -29,6 +31,12 @@ const radioGroupItemBehavior: Accessibility<RadioGroupItemBehaviorProps> = props
       },
     },
   },
+
+  focusZone: {
+    props: {
+      shouldEnterInnerZone: event => keyboardKey.getCode(event) === keyboardKey.Tab,
```
jurokapsiar, line 37: why is this needed? as far as I remember, radios do not use focus zone

assuncaocharles: Reverted

PR #13117, opened 2020-05-12
feat(MenuItem): use compose()

mnajdova:
This PR adds compose() to the MenuItem component and creates dedicated components for the MenuItem's slots.

# BREAKING CHANGES

1. Restricted set of props in the `menuItemStyles` and `MenuItemStylesProps`:
- `icon`, `menu` and `hasContent` were removed, and dedicated props inside the slot's components are added
- `icon` is replaced with `hasIcon` and `menu` is replaced with `hasMenu`.

2. `menuItemSlotClassNames.indicator` and `menuItemSlotClassNames.wrapper`  were removed, instead `menuItemIndicatorClassName` and  `menuItemWrapperClassName` shuold be used.

### Before
```tsx
import { menuItemSlotClassNames } from '@fluentui/react-northstar';
console.log(menuItemSlotClassNames.indicator);
console.log(menuItemSlotClassNames.wrapper);
```

### After
```tsx
import { menuItemIndicatorClassName, menuItemWrapperClassName } from '@fluentui/react-northstar';
console.log(menuItemIndicatorClassName);
console.log(menuItemWrapperClassName);
```

3. Move the style for the slot's of the `MenuItem`  to the dedicated component styles.

### Before

```ts
const menuItemStyles = {
  root: () => {...}, // this is ok
  menu: () => {...}, // this is ok
  wrapper: () => {...}, // move this to MenuItemWrapper styles
  icon: () => {...}, // move this to MenuItemIcon styles
  content () => {...}, // move this to MenuItemContent styles
  indicator: () => {...}, // move this to MenuItemIndicator styles
}
```

### After
```ts
const menuItemStyles = {
  root: () => {...}, // this is ok
  menu: () => {...}, // this is ok
}
const menuItemWrapperStyles = {
  root: () => {...}, // export in the theme under the MenuItemWrapper name
}
const menuItemContentStyles = {
  root: () => {...}, // export in the theme under the MenuItemContent name
}
const menuItemIconStyles = {
  root: () => {...}, // export in the theme under the MenuItemIcon name
}
const menuItemIndicatorStyles = {
  root: () => {...}, // export in the theme under the MenuItemIndicator name
}

```


packages/react-compose/src/types.ts
```diff
@@ -56,7 +64,7 @@ export type ComposeOptions<InputProps = {}, InputStylesProps = {}, ParentStylesP
 
   slots?: Record<string, React.ElementType>;
 
-  mapPropsToSlotProps?: (props: InputProps) => Record<string, object>;
+  mapPropsToSlotProps?: (props: Extendable<InputProps>) => Record<string, object>;
```
mnajdova, line 67: We need to use state here too, for now converted this to extendable

packages/fluentui/docs/src/examples/components/Menu/Visual/MenuExampleCompose.shorthand.tsx
```diff
@@ -0,0 +1,90 @@
...
+  },
+});
+
+const items = [
+  {
+    key: 'editorials',
+    content: 'Editorials',
+    icon: <BookmarkIcon />,
+    menu: ['One', 'Two', 'Three'],
+    children: (C, p) => <MenuItemBlue {...p} />,
```
layershifter, line 46:
```suggestion
    children: (C, p) => <MenuItemBlue {...p} />,
```

Should we add a todo there?

packages/fluentui/react-northstar/src/components/Menu/MenuItem.tsx
```diff
@@ -166,321 +158,403 @@ export type MenuItemStylesProps = Required<
...
+        ...props,
+        menuOpen: newValue,
+      });
+    };
+
+    const menuItemInner = (
+      <Ref
+        innerRef={node => {
+          // @ts-ignore
+          itemRef.current = node;
```
layershifter, line 393:
```suggestion
          itemRef.current = node as HTMLDivElement;
```

mnajdova: Casting is correct 👍 but I still need the ts-ignore because of 
```
TS2540: Cannot assign to 'current' because it is a read-only property.
```

Or am I missing something?

PR #13123, opened 2020-05-12
Adding keyboard-key to react-focus dependencies

dzearing:
Cloning a new repo, I'm getting build errors with react-focus due to this missing dependency.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13123)


ecraig12345:
Approved but you might want to consider making it a devDependency for now, since it's only used in tests (guessing it will be used in the component itself later).

dzearing:
The component is intended to use it, but we're holding off on integrating. I'd hate to forget to add it later down the road, so while you're correct, it seems like a pretty painless mitigation

PR #13130, opened 2020-05-13
Enabling option to not select anything in a Nav component.

### LINKED ISSUE #13129: Option to not have anything selected in the Nav component

I'd like to be able to explicitly pass something like `undefined` to the `selectedKey` prop in a `Nav` component, to tell the component to not show any items as selected.

My use case is that we have a secondary nav component that, when the user clicks on one of those, we want to have the main nav be still visible but un-selected, and the secondary nav selected instead.

Nav

No

--

paperbackdragon:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13129

#### Description of changes

This checks whether the value of `selectedKey` was explicitly set to `undefined`. If so, nothing will be shown as selected. If `selectedKey` was not set at all, it will still respect the previous functionality.

#### Focus areas to test

--

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13130)

### DIFF
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -5687,7 +5687,7 @@ export interface IModalStyles {
 // @public (undocumented)
 export interface INav {
     focus(forceIntoFirstElement?: boolean): boolean;
-    selectedKey: string | undefined;
+    selectedKey: string | null | undefined;
 }
 
 // @public (undocumented)
@@ -5748,7 +5748,7 @@ export interface INavProps {
     onRenderLink?: IRenderFunction<INavLink>;
     // @deprecated
     selectedAriaLabel?: string;
-    selectedKey?: string;
+    selectedKey?: string | null;
     styles?: IStyleFunctionOrObject<INavStyleProps, INavStyles>;
     theme?: ITheme;
 }
@@ -5762,7 +5762,7 @@ export interface INavState {
     // (undocumented)
     isLinkExpandStateChanged?: boolean;
     // (undocumented)
-    selectedKey?: string;
+    selectedKey?: string | null;
 }
 
 // @public (undocumented)
@@ -8308,7 +8308,7 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
     // (undocumented)
     render(): JSX.Element | null;
     // (undocumented)
-    readonly selectedKey: string | undefined;
+    readonly selectedKey: string | null | undefined;
     }
 
 // @public (undocumented)
--- a/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
@@ -26,7 +26,7 @@ const getClassNames = classNamesFunction<INavStyleProps, INavStyles>();
 export interface INavState {
   isGroupCollapsed: { [key: string]: boolean };
   isLinkExpandStateChanged?: boolean;
-  selectedKey?: string;
+  selectedKey?: string | null;
 }
 
 export class NavBase extends React.Component<INavProps, INavState> implements INav {
@@ -65,8 +65,9 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
     );
   }
 
-  public get selectedKey(): string | undefined {
-    return this.state.selectedKey;
+  public get selectedKey(): string | null | undefined {
+    const { selectedKey = this.state.selectedKey } = this.props;
+    return selectedKey;
   }
 
   /**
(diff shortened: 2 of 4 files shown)


ecraig12345:
Another option would be adding `null` to the `selectedKey` types (in INavProps and INav) and using that to indicate nothing is selected. Advantage of this is that it doesn't increase the API surface (and we don't have to choose a good prop name). Disadvantage is it's less explicit.

If you do this, you should also update `public get selectedKey()` to return `this.props.selectedKey` if it exists (it should have already been doing this anyway...).
```ts
  public get selectedKey(): string | null | undefined {
    const { selectedKey = this.state.selectedKey } = this.props;
    return selectedKey;
  }
```

And in `_isLinkSelected()` (the top comment can be removed because it's not very useful):
```diff
private _isLinkSelected(link: INavLink): boolean {
-   // if caller passes in selectedKey, use it as first choice or
-   // if current state.selectedKey (from addressbar) is match to the link or
-   // check if URL is matching location.href (if link.url exists)
-   if (this.props.selectedKey !== undefined) {
-     return link.key === this.props.selectedKey;
-   } else if (this.state.selectedKey !== undefined) {
-     return link.key === this.state.selectedKey;
+   if (this.selectedKey !== undefined) {
+     return link.key === this.selectedKey;
```

paperbd:
@ecraig12345 Thanks for the suggestions! I think I like the `null` option best, so that I can keep using the same prop to indicate whether it's selected or not. I'll make the improvements you suggested.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -5687,7 +5687,7 @@ export interface IModalStyles {
 // @public (undocumented)
 export interface INav {
     focus(forceIntoFirstElement?: boolean): boolean;
-    selectedKey: string | undefined;
+    selectedKey: string | null | undefined;
```
ThomasMichon, line 5690: Can we avoid `null` as something we need to consider? I still recommend testing for `'selectedKey' in this.props` rather than allowing a specific `null` assignment.

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
```diff
@@ -65,8 +65,9 @@ export class NavBase extends React.Component<INavProps, INavState> implements IN
     );
   }
 
-  public get selectedKey(): string | undefined {
-    return this.state.selectedKey;
+  public get selectedKey(): string | null | undefined {
```
ThomasMichon, line 68: Since this is a pseudo-internal method, can we just add:
```ts
public get getSelectedKeySpecifed(): boolean {
  return 'selectedKey' in this.props;
}
```
And then logic can use that for consideration along with `selectedKey`?

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
ecraig12345: I previously requested that she *not* explicitly test for `'selectedKey' in this.props` because it's brittle and runs counter to the general pattern of handling explicit and implicit undefined the same way.

Can you explain more about why you're opposed to explicit null?

packages/office-ui-fabric-react/src/components/Nav/Nav.base.tsx
ecraig12345: It's not pseudo-internal, this is part of the public interface.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
ThomasMichon: This is almost a case where I would recommend using the empty string `''` instead of `null`. The issue is that `null` tends to cascade and break stuff everywhere because code now has to differentiate it from `undefined`.

paperbackdragon:
@ecraig12345 Looks like Thomas signed off on the `null` approach, do you still like this as a solution or are there other changes I should make?

ecraig12345:
Thought about this more and partly based on what @ThomasMichon said, I think it might be better to have an explicit prop `isSelectionDisabled`. Just be sure to make it very clear in the doc comment that links are still clickable, it just won't update the visible selected state or `selectedKey`.

paperbackdragon:
Ok, that's fair. I'll make those changes.

ecraig12345:
Actually I was just discussing this with the rest of the team and @xugao was saying we want to avoid adding new props for controlling styling (though I'm not sure what a good alternative would be in this case). So probably hold off on making any changes just yet, sorry.

xugao:
@paperbackdragon - our components today face a big challenge, that is the style is not easily customizable to fit all needs and styles prop is fragile. As a result of this, our prop interface is very bloated overtime as we support more and more scenarios like yours.

We are working toward solving this problem with our upcoming releases. At the moment, we are going to push back adding new props that control our styling as much as we can.

If I understand your ask correctly, I believe this can be achieved using `styles` prop today. I definitely understand this is not ideal but i still believe is the right approach to take all things considered. 

Here is a demo of what I think you want to accomplish: https://codepen.io/xugao/pen/wvKQEVm

paperbackdragon:
@xugao I see, thanks for the update. I'll close this PR for now then.

PR #13133, opened 2020-05-13
Pin TS version used by test:projects:typings

ecraig12345:
Pin TS version used by test:projects:typings to the same one we use locally. This works around the merge-styles build errors in TS 3.9.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13133)

### DIFF
--- a/scripts/gulp/tasks/test-projects.ts
+++ b/scripts/gulp/tasks/test-projects.ts
@@ -265,7 +265,9 @@ task('test:projects:typings', async () => {
 
   logger(`✔️ Temporary directory was created: ${tmpDirectory}`);
 
-  const dependencies = ['@types/react', '@types/react-dom', 'react', 'react-dom', 'typescript'].join(' ');
+  // Install dependencies, ensuring we specify the same TS version as our projects use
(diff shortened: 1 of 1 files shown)


scripts/gulp/tasks/test-projects.ts
```diff
@@ -265,7 +265,9 @@ task('test:projects:typings', async () => {
 
   logger(`✔️ Temporary directory was created: ${tmpDirectory}`);
 
-  const dependencies = ['@types/react', '@types/react-dom', 'react', 'react-dom', 'typescript'].join(' ');
+  // Install dependencies, ensuring we specify the same TS version as our projects use
+  const tsVersion = fs.readJSONSync(paths.base('scripts', 'package.json')).devDependencies.typescript;
```
dzearing, line 269: can't you just

```tsx
const tsVersion = require('../package.json').devDependencies.typescript;
```

ecraig12345: With package.json that would work since it can't have comments, but I'm in the habit of using fs.readJSONSync since it handles commented JSON by default IIRC.

PR #13145, opened 2020-05-13
chore(bundle-size): extract classname consts to separate files

mnajdova:
This PR is experimenting with some optimization for the bundle size. I was doing some investigation of why for an application with one Button we are dragging all components in the bundle. The reason was quite surprising - because we are importing the constants for the classNames and slot's classNames from the component's file. Just extracting these constants to a separate file, removed the components from the bundle. This PR is showcasing the dropping of the Dropdown components from the bundle.
 
Before:
image

After:
image

We should probably do some more deeper investigation of why this is happening, and whether this is the default behavior of webpack (as one representative of the bundlers), but this seems like a low hanging fruit for dropping the bundle size for sure.


dzearing:
This *might* not be needed. When you import { thing } from a file, webpack will include the entire file in the bundle.

When you turn concatenation on, that file would be in the same array entry as most other modules that got imported.

When you turn on terser, it should delete unreferenced content.

I believe if you have mode=production and look at the tree maps, you should get some different results. If that's not the case, some possible causes:

1. Concatenation didn't concatenate. Anything which isn't shared runs the risk of not getting removed from the bundle. You will need to use bailout debugging to figure this out (see webpack "bailout" for more details on how this is done.) Concatenation breaks in side effect scenarios; for example, if you reference `window` or `process` or `console`, those definitions can cause a module to be separated into its own webpack arrray entry and export it, causing terser to avoid deleting it because it can't trace the usage.

2. Some hidden reference pulling it in (which i doubt is the case if separating the file makes the stuff go away.) You can look at the unexpected module and find imports to it in the stats.json.

BTW this discussion illustrates one of the reasons why we have the top level imports (like `package/lib/Foo`) - webpack randomness (like bailouts on concatenation) can cause bundles to regress unexpectedly, so if you end up minimizing how much you're actually importing, you reduce the risk of having something be pulled in due to unpredictable scenarios. Not saying top level imports, but just that there should still be some skepticism around if webpack will tree shake as you expect. It's easy to break expectations.

ecraig12345:
@mnajdova Is this intended to be checked in or is it just for demo purposes? If it's just a demo and is no longer needed we should close it.

PR #13146, opened 2020-05-13
ThemeProvider Integration

xugao:
#### Pull request checklist

#### Description of changes
The goal of this change is to use `ThemeProvider` to provide theme css variables instead of `.scss` file. But also ensure backward compatibility with existing ways of theming in V7.
I will add unit tests after a round of reviews.

##### Existing ways of theming in V7
- `Customizer`
    - `<Customizer settings={{ theme: newTheme }} />`
    - `<Customizer scopedSettings={{ [Component_Name]: {theme: newTheme}}} />`

- `Customization`
    - `applySettings`
    - `applyScopedSettings`

- `loadTheme`: calls `Customizations.applySettings({theme})`

##### Key changes to review
-  New `ThemeProvider` in `react-next`
It should be used everywhere in V8 instead of `ThemeProvider` from `react-theme-provider` package. It does 2 things:
     - Convert `ITheme` to our new Theme shape.
     - Ensure backward compatibility with `Customizations.applySettings` and `loadTheme`.
- New `Customizer` in `react-next`
This needs to be used on V8 components instead of the current `Customizer` in V7. It ensures global settings (theme) is being provided by `ThemeProvider`. 
**Caveat**: If it's used on components which are not yet converted on scss, work being done by `ThemeProvider` will be wasted work in this case.

##### Theming story in V8
https://hackmd.io/lLqrReqIQP2qAyHVXp4NcA?view

##### Backward compatibility
- [X] `<Customizer settings={{ theme: newTheme }} />`:
Use new `Customizer` from `react-next`. Need code mods for updating imports
- [ ] `<Customizer scopedSettings={{ [Component_Name]: {theme: newTheme}}} />`
My change currently does not support this. This is undocumented feature and honestly no one should be currently using. so wee are not going to support this going forward. 
- [X] `Customization.applySettings`
`ThemeProvider` will inject new theme when settings changes.
- [ ] `Customization.applyScopedSettings`
My change currently does not support this. This is undocumented feature and honestly no one should be currently using. so wee are not going to support this going forward. 
- [X] `loadTheme`
This one calls `Customization.applySettings`. so It's handled by `ThemeProvider`

##### Follow ups
- Should we add `ThemeProvider` to `Fabric`? Most likely yes. But I'm going to leave this out of scope for this PR. `Fabric` needs some cleanup. Contextual RTL logic can be removed?
- If new `Customizer` is used on components which are not yet converted on scss, work being done by `ThemeProvider` will be wasted work in this case.
- replace the usage of `CustomizerContext` using `ThemeContext` 

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13146)


packages/react-next/src/components/Checkbox/examples/Checkbox.Basic.Example.tsx
```diff
@@ -1,22 +1,25 @@
...
 
 // Used to add spacing between example checkboxes
 const stackTokens = { childrenGap: 10 };
 
 export const CheckboxBasicExample: React.FunctionComponent = () => {
   // These checkboxes are uncontrolled because they don't set the `checked` prop.
   return (
-    <Stack tokens={stackTokens}>
-      <Checkbox label="Unchecked checkbox (uncontrolled)" onChange={_onChange} />
+    <ThemeProvider>
```
khmakoto, line 12: I think it's fine for this PR but we should probably find a way to inject the `ThemeProvider` at the example layer and not in each example.

packages/react-next/src/components/Checkbox/examples/Checkbox.CustomStyle.Example.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Stack } from '@fluentui/react-next/lib/Stack';
+import { Customizer } from '@fluentui/react-next/lib/Utilities';
+import { loadTheme, ThemeProvider, FontWeights, FontSizes, IPartialTheme } from '@fluentui/react-next/lib/Styling';
+
+// Used to add spacing between example checkboxes
+const stackTokens = { childrenGap: 10 };
+
+const customizedTheme: IPartialTheme = { fonts: { medium: { fontWeight: FontWeights.bold } } };
+
+// TODO (xgao): Remove. This example is temporary for testing backward compatability.
```
khmakoto, line 13:
```suggestion
// TODO (xgao): Remove. This example is temporary for testing backward compatibility.
```

packages/react-next/src/Styling.ts
```diff
@@ -1 +1,3 @@
 export * from 'office-ui-fabric-react/lib/Styling';
+
+export * from './Styling/ThemeProvider';
```
dzearing, line 3: Hmm. Should this be in `components/ThemeProvider`?

We should probably just avoid having folders that are named the same as files which can be imported.

packages/react-next/src/Styling/ThemeProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+    setCustomizationSettings(globalSettings);
+  }, []);
+
+  React.useEffect(() => {
+    Customizations.observe(onCustomizationChange);
+
+    return () => Customizations.unobserve(onCustomizationChange);
+  }, []);
+
+  return <ReactThemeProvider {...props} theme={convertLegacyTheme(customizationSettings.theme)} />;
```
dzearing, line 29: rather than calling convert on every render, consider memoizing based on the theme pulled from customizations.

packages/react-next/src/Utilities.ts
```diff
@@ -1 +1,4 @@
 export * from 'office-ui-fabric-react/lib/Utilities';
+
+import { Customizer, ICustomizerProps } from './Utilities/Customizer';
```
dzearing, line 3: why not just export these rather than import/export?

packages/react-next/src/components/Checkbox/examples/Checkbox.Basic.Example.tsx
dzearing: Agree that it's fine to get this going here as a start, but we'll likely want to revisit before ship.

dzearing: maybe add a TODO (xugao)

packages/react-next/src/Utilities/Customizer.tsx
```diff
@@ -0,0 +1,59 @@
...
+        let newContext = mergeCustomizations(props, parentContext);
+
+        if (contextTransform) {
+          newContext = contextTransform(newContext);
+        }
+
+        const content = disableThemeProvider ? (
+          children
+        ) : (
+          <ThemeProvider theme={convertLegacyTheme(newContext.customizations.settings.theme)}>
```
dzearing, line 45: This seems backwards; shouldn't ThemeProvider (in react-next) include Customizer?

I might be wrong.

xugao: The reason I have a new `Customizer` is bc it's specifically added for keeping backward compatibility with current `Customizer`. You can think of it as a shimmer and I actually want to ship this as `@deprecated`. I don't want to tight this with `ThemeProvider` and pollute its prop interface.

should `ThemeProvider` contain `Customizer`? I think the answer is "no" because we can use `ThemeContext` instead of `CustomizerContext`

packages/react-next/src/Styling/ThemeProvider.tsx
xugao: it's memoized inside `convertLegacyTheme`

packages/react-next/src/Utilities/Customizer.tsx
```diff
@@ -0,0 +1,59 @@
...
+          </ThemeProvider>
+        );
+
+        return <CustomizerContext.Provider value={newContext}>{content}</CustomizerContext.Provider>;
+      }}
+    </CustomizerContext.Consumer>
+  );
+};
+
+function useForceUpdate(): () => void {
```
dzearing, line 56: do we already have this in react-hooks? I've needed this in at least one other place.

packages/react-next/src/components/Checkbox/examples/Checkbox.CustomStyle.Example.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  return (
+    <ThemeProvider>
+      <Stack tokens={stackTokens}>
+        <Customizer settings={{ theme: customizedTheme }}>
+          <Checkbox label="With theme" />
+        </Customizer>
+
+        <Customizer scopedSettings={{ Checkbox: { theme: customizedTheme } }}>
+          <Checkbox label="With scoped theme (not working)" defaultChecked />
```
dzearing, line 25: not working?

packages/react-next/src/Utilities/Customizer.tsx
```diff
@@ -0,0 +1,59 @@
...
+  mergeCustomizations,
+  ICustomizerContext,
+} from 'office-ui-fabric-react/lib/Utilities';
+import { ThemeProvider } from '@fluentui/react-theme-provider';
+import { convertLegacyTheme } from '../Styling/convertLegacyTheme';
+
+import { ICustomizerProps as LegacyCustomizerProps } from 'office-ui-fabric-react/lib/Utilities';
+
+export interface ICustomizerProps extends LegacyCustomizerProps {
+  disableThemeProvider?: boolean;
```
dzearing, line 14: weird boolean

dzearing:
Overall I like it. Lets chat tomorrow about it!

packages/react-next/src/components/Checkbox/examples/Checkbox.Basic.Example.tsx
xugao: fixed, now it's automatically wrapped on all example stories

packages/react-next/src/Styling.ts
xugao: What's the downside of having a folder with same name? we do this today in OUFR. also I think `ThemeProvider` fits better in `lib/Styling`

packages/react-next/src/Utilities/Customizer.tsx
xugao: moved to `react-hooks`

packages/react-next/src/components/Checkbox/examples/Checkbox.CustomStyle.Example.tsx
xugao: Nope. this is providing **theme**  (not styles) to component. As discussed, we should abort supporting this

packages/react-next/src/Utilities/Customizer.tsx
xugao: it's not ideal, but it's necessary to avoid `ThemeProvider` when it's not needed. couple scenarios:
1. theme updated with only RTL flipping (see the use case in `Fabric`)
2. when `Customizer` is used ONLY on components that are still not converted to scss yet.

xugao:
@dzearing @khmakoto - pls give another look. I have resolved/replied on all comments and added unit tests for `Customizer` and `ThemeProvider`

packages/react-next/src/Styling/ThemeProvider.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    mount(
+      <ThemeProvider>
+        <TestComponent />
+      </ThemeProvider>,
+    );
+
+    expect(resolvedTheme).toEqual(mergeThemes(convertLegacyTheme(legacyTheme)));
+  });
+
+  it('provide theme from theme prop if passed', () => {
```
dzearing, line 61:
```suggestion
  it('can provide theme from theme prop if passed', () => {
```

packages/react-next/src/Styling/ThemeProvider.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    mount(
+      <ThemeProvider theme={lightTheme}>
+        <TestComponent />
+      </ThemeProvider>,
+    );
+
+    expect(resolvedTheme).toEqual(lightTheme);
+  });
+
+  it('provide new theme over legacy theme', () => {
```
dzearing, line 78:
```suggestion
  it('can provide new theme over legacy theme', () => {
```

packages/react-next/src/Styling/ThemeProvider.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    ReactTestUtils.act(() => {
```
dzearing, line 79: curious if there's a benefit of using this over enzyme mount.

packages/react-next/src/Styling/ThemeProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const onCustomizationChange = React.useCallback(() => {
+    const globalSettings = getGlobalCustomizationSettings();
+    setCustomizationSettings(globalSettings);
+  }, []);
+
+  React.useEffect(() => {
+    Customizations.observe(onCustomizationChange);
+
+    return () => Customizations.unobserve(onCustomizationChange);
+  }, []);
```
dzearing, line 28: this could be abstracted into a nice easy to read hook:

```tsx
const customiztionSettings = useCustomizationSettings();
```

packages/react-next/src/Utilities/Customizer.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+      wrapper => {
+        expect(wrapper.html()).toEqual('<div><div>Content</div></div>');
+        const inlineStyle = wrapper.find('div').get(0).props.style;
+        expect(inlineStyle).toBeDefined();
+        expect(Object.keys(inlineStyle).length).toBeGreaterThan(0);
+      },
+    );
+  });
+
+  fit('disable ThemeProvider', () => {
```
dzearing, line 268: fit?

packages/react-next/src/Utilities/Customizer.tsx
```diff
@@ -0,0 +1,50 @@
...
+  React.useEffect(() => {
+    Customizations.observe(forceUpdate);
+
+    return () => Customizations.unobserve(forceUpdate);
+  }, []);
+
+  const { contextTransform, children, disableThemeProvider } = props;
+  return (
+    <CustomizerContext.Consumer>
+      {(parentContext: ICustomizerContext) => {
```
dzearing, line 31: this there a way to avoid merging on every render?

packages/react-next/src/Utilities/Customizer.types.ts
```diff
@@ -0,0 +1,56 @@
...
+   * mutating it.
+   */
+  contextTransform?: (context: Readonly<ICustomizerContext>) => ICustomizerContext;
+
+  /**
+   * In cases where ThemeProvider is not needed (e.g. only RTL in theme has changed,
+   * Customizer is only used on V7 components), this can be set to true to disable
+   * ThemeProvider.
+   */
+  disableThemeProvider?: boolean;
```
dzearing, line 55: Unsure of the use case here.

packages/react-next/src/Utilities/Customizer.tsx
khmakoto: Maybe this can be memoized?

packages/react-next/src/Utilities/Customizer.types.ts
```diff
@@ -0,0 +1,56 @@
...
+   * ```
+   *  <Customizer settings={{ color: 'red' }} />
+   * ```
+   * or a function that receives the current settings and returns the new ones
+   * ```
+   *  <Customizer settings={(currentSettings) => ({ ...currentSettings, color: 'red' })} />
+   * ```
+   */
+  settings?: ISettings | ISettingsFunction;
+  /**
```
khmakoto, line 19: nit: Add blank line between props.

packages/react-next/src/Utilities/Customizer.types.ts
khmakoto: Seems like it is being used in `Fabric.base.tsx`.

packages/react-next/src/Utilities/Customizer.tsx
xugao: memoization is the only way however it's not going to help.

I think most people are inlining the `settings` or `scopedSettings` props when they using the `Customizer`, as the following:
```
<Customizer settings={ theme: { fonts ... }} />
<Customizer scopedSettings={ Button: { styles: {root: ... } }} />
```

The good news is I don't think there is too big of a concern here. this "merge" logic is not expensive

packages/react-next/src/Utilities/Customizer.types.ts
xugao: I mentioned the use cases in the comment, is it still unclear why i added it? I think this can come handy since we are shooting for using this **one** `Customizer` for mixed use cases

packages/react-next/src/Utilities/Customizer.test.tsx
xugao: ugh, thanks for catching! i wish there is a way to warn me on using `fit`

packages/react-next/src/Styling/ThemeProvider.test.tsx
xugao: turns out i don't need this here as long as i clean up the mount after each test.

packages/react-next/src/Styling/ThemeProvider.tsx
xugao: added hook and tests

PR #13149, opened 2020-05-13
React-next unit test fixes

xugao:
#### Pull request checklist

#### Description of changes
- Stabilize snapshot unit tests by `resetIds` before each test run.
- Fixes Fabric tests to import from the right package.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13149)

### DIFF
--- a/change/@fluentui-react-next-2020-05-13-12-40-18-xgao-fix-tests.json
+++ b/change/@fluentui-react-next-2020-05-13-12-40-18-xgao-fix-tests.json
@@ -1,6 +1,6 @@
 {
   "type": "none",
-  "comment": "unit tests fixes",
+  "comment": "Stabilize unit tests by calling `resetIds` for each test file and fix Fabric tests to use correct import path.",
   "packageName": "@fluentui/react-next",
   "email": "xgao@microsoft.com",
   "dependentChangeType": "none",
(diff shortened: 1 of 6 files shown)


change/@fluentui-react-next-2020-05-13-12-40-18-xgao-fix-tests.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "unit tests fixes",
```
khmakoto, line 3:
```suggestion
  "comment": "Fixing unit tests by standardizing on the use of `resetIds` before each test.",
```

packages/react-next/src/components/Image/Image.test.tsx
```diff
@@ -18,6 +19,10 @@ describe('Image', () => {
     Object.defineProperty(HTMLImageElement.prototype, 'naturalWidth', { get: () => 1 });
   });
 
+  beforeEach(() => {
```
dzearing, line 22: It might be possible to just move this to config/tests.js
```
beforeAll(() => {
  resetIds()
});
```

https://jestjs.io/docs/en/setup-teardown#one-time-setup

xugao: This will help, but if we truly avoid the noise, resetting id should be done before for each test run rather than each test file run.

xugao: it's certainly a tradeoff. but i definitely like doing this once and never need to worry about it again.  🤔 i think i will take that

PR #13152, opened 2020-05-13
Add utility to apply default props for function components

MLoughry:
#### Pull request checklist

#### Description of changes

This is split off from #12928. It adds a utility to provide the same sort of functionality for function components as `defaultProps` does for class components. Suggested usage pattern is:
```ts
(propsWithoutDefaults: IFooProps) => {
    const props = getPropsWithDefaults(DEFAULT_PROPS, propsWithoutDefaults);
}
```

Alternative patterns considered and rejected:
- Parameter destructuring (`({foo, bar = ''}: IFooProps) => {}`): This prevents the ability to pass the entire props object around to hooks and the like
- In-function destructuring (`const {foo, bar = ''} = props`): This makes it easy to inadvertently dereference a non-defaulted value off of `props`, particularly if passing `props` to a hook
- Object structuring (`const props = {bar: '', ...propsWithoutDefaults}`) - This causes any property that is *specified* with an `undefined` value to remain `undefined`. This is a behavioral difference from `defaultProps` that necessitated this utility.

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13152)

### DIFF
--- a/packages/utilities/src/getPropsWithDefaults.ts
+++ b/packages/utilities/src/getPropsWithDefaults.ts
@@ -1,7 +1,7 @@
 /**
  * Function to apply default values to a component props object. This function is intended for function components,
  * to maintain parity with the `defaultProps` feature of class components. It accounts for properties that are
- * specified, but undefined
+ * specified, but undefined.
  * @param defaultProps- An object with default values for various properties
  * @param propsWithoutDefaults- The props object passed into the component
  */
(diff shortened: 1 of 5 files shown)


packages/utilities/src/getPropsWithDefaults.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Function to apply default values to a component props object. This function is intended for function components,
+ * to maintain parity with the `defaultProps` feature of class components. It accounts for properties that are
+ * specified, but undefined
```
ecraig12345, line 4:
```suggestion
 * specified, but undefined.
```

PR #13159, opened 2020-05-14
feat(schema): add tree item to the schema validator

assuncaocharles:
#### Pull request checklist

#### Description of changes

This PR adds `TreeItem` to the schema validation adding on it atomic attributes to throw errors in some cases as for example if it has other focusable element inside:

Screenshot 2020-05-14 at 13 21 19

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13159)

### DIFF
--- a/packages/fluentui/ability-attributes/schema.json
+++ b/packages/fluentui/ability-attributes/schema.json
@@ -64,6 +64,48 @@
         }
       ]
     },
+    "$setsize": {
+      "name": "setsize",
+      "attributes": [
+        {
+          "name": "aria-setsize",
+          "constraints": [
+            {
+              "xpath": "number(@aria-setsize) = @aria-setsize",
+              "description": "aria-setsize attribute value cannot be empty"
+            }
+          ]
+        }
+      ]
+    },
+    "$posinset": {
+      "name": "posinset",
+      "attributes": [
+        {
+          "name": "aria-posinset",
+          "constraints": [
+            {
+              "xpath": "number(@aria-posinset) = @aria-posinset",
+              "description": "aria-posinset attribute value cannot be empty"
+            }
+          ]
+        }
+      ]
+    },
+    "$level": {
+      "name": "level",
+      "attributes": [
+        {
+          "name": "aria-level",
+          "constraints": [
+            {
+              "xpath": "number(@aria-level) = @aria-level",
+              "description": "aria-level attribute value cannot be empty"
+            }
+          ]
+        }
+      ]
+    },
     "$describedBy": {
       "name": "describedBy",
       "attributes": [
@@ -190,6 +232,24 @@
             {
               "parameter": false,
               "attribute": [false, "false"]
+            }
+          ]
+        }
+      ]
+    },
+    "$isSelected": {
+      "name": "isSelected",
+      "attributes": [
+        {
+          "name": "aria-selected",
+          "value": [
+            {
+              "parameter": true,
+              "attribute": "true"
+            },
+            {
+              "parameter": true,
+              "attribute": "false"
             }
           ]
         }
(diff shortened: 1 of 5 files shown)


packages/fluentui/accessibility/src/behaviors/Tree/treeItemBehavior.ts
```diff
@@ -22,64 +22,73 @@ import treeTitleBehavior from './treeTitleBehavior';
...
-  childBehaviors: {
-    title: treeTitleBehavior,
-  },
-});
+    childBehaviors: {
+      title: treeTitleBehavior,
+    },
+  };
+
+  if (process.env.NODE_ENV !== 'production' && !props.hasSubtree) {
```
assuncaocharles, line 85: To avoid these conditions I suggest in another task that we refactor `Tree` and get ride of `TreeTitle` and have only `TreeItem` inside the tree. What do you think @kolaps33 @layershifter @jurokapsiar

packages/fluentui/ability-attributes/schema.json
```diff
@@ -64,6 +64,48 @@
...
     },
+    "$setsize": {
+      "name": "setsize",
+      "attributes": [
+        {
+          "name": "aria-setsize",
+          "constraints": [
+            {
+              "xpath": "number(@aria-setsize) = @aria-setsize",
+              "description": "aria-setsize attribute value cannot be empty"
```
hoshinetop, line 75: Let's do «should be a number» instead of «cannot be empty» for all three of setsize, posinset and level.

assuncaocharles: Done

packages/fluentui/accessibility/src/behaviors/Tree/treeItemBehavior.ts
```diff
@@ -22,64 +22,73 @@ import treeTitleBehavior from './treeTitleBehavior';
  * Triggers 'expand' action with 'ArrowRight' on 'root', when has a closed subtree.
  * Triggers 'focusFirstChild' action with 'ArrowRight' on 'root', when has an opened subtree.
  */
-const treeItemBehavior: Accessibility<TreeItemBehaviorProps> = props => ({
```
hoshinetop, line 25: Does this really change something other than brackets (here and below)? If not, let's make it as it was before.

mshoho, line 25: Does this really change something other than brackets (here and below)? If not, let's make it as it was before.

assuncaocharles: It changes. Instead of returning the behavior directly we are putting it in the `definition` var to change the `aa-class`

packages/fluentui/accessibility/src/behaviors/Tree/treeTitleBehavior.ts
```diff
@@ -15,28 +15,37 @@ import { Accessibility } from '../../types';
...
-        'aria-selected': props.selectable ? props.selected || false : undefined,
-      }),
+const treeTitleBehavior: Accessibility<TreeTitleBehaviorProps> = props => {
+  const definition = {
+    attributes: {
+      root: {
+        ...(!props.hasSubtree && {
+          tabIndex: -1,
+          [IS_FOCUSABLE_ATTRIBUTE]: true,
+          role: 'treeitem' as AriaRole,
```
pompomon, line 25: Could you update the `@specification` comment to reflect those changes?

pompomon: Sorry, didn't notice that it's there already :)

PR #13163, opened 2020-05-14
feat(Card): Adding new props and styles

pompomon:
#### Pull request checklist

#### Description of changes

Add `elevated`, `inverted` and `quiet` props and styles for the `Card`

image
image
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13163)


packages/fluentui/react-northstar/src/themes/teams/components/Card/cardStyles.ts
```diff
@@ -54,12 +54,66 @@ const cardStyles: ComponentSlotStylesPrepared<CardStylesProps, CardVariables> =
...
       }),
 
+      ...(p.elevated && {
+        boxShadow: v.elevatedBoxShadow,
+        ':hover': {
+          boxShadow: v.elevatedBoxShadowHover,
+        },
+
+        ...(p.actionable && {
+          ':focus-visible': {
```
mnajdova, line 64: Why are we not using the `getBorderFocusStyles` here? We can invoke it with different variables based on the need

pompomon: Good catch, thank you

Issue #13132, opened 2020-05-13
(merge-styles) TS2590: Expression produces a union type that is too complex to represent


EvanCahill:
### Environment Information

- **Package version(s)**: latest, 6.19.4
- **Browser and OS versions**: N/A build issue

### Please provide a reproduction of the bug in a codepen:
Try compiling the below code with typescript 3.9.2.
```ts
// Repro.ts
import { mergeStyleSets, concatStyleSets } from '@uifabric/merge-styles';
mergeStyleSets();
concatStyleSets();
```

Similar to #9280

#### Actual behavior:

node_modules/@uifabric/merge-styles/lib/concatStyleSets.d.ts:12:226 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/concatStyleSets.d.ts:19:318 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/concatStyleSets.d.ts:27:410 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/concatStyleSets.d.ts:35:410 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/concatStyleSets.d.ts:44:502 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/concatStyleSets.d.ts:54:594 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/mergeStyleSets.d.ts:21:225 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/mergeStyleSets.d.ts:32:317 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/mergeStyleSets.d.ts:44:409 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/mergeStyleSets.d.ts:73:239 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/mergeStyleSets.d.ts:83:320 - error TS2590: Expression produces a union type that is too complex to represent.
node_modules/@uifabric/merge-styles/lib/mergeStyleSets.d.ts:93:401 - error TS2590: Expression produces a union type that is too complex to represent.

[...]

PR #13165, opened 2020-05-14
merge-styles api input argument typings relaxed (TypeScript 3.9 now compiles)

dzearing:
#### Pull request checklist

In testing TS 3.9, customers found compiler typing errors being thrown from merge-styles apis. With the typescript compiler checks getting more accuracy, this can happen.

This should resolve the issues while still maintaining type safety on inputs/outputs.

Note that `ObjectOnly` type is needed to ensure that falsely types do not get intersected with the non-falsey result.

Background on `ObjectOnly`:

Think about this case:

```tsx
const result = concatStyleSets({ a: { ... } } , undefined)
```

The `result` type would be:

```tsx
{ a: { ... } } & undefined
```

...Which results in `never`. Which is not quite what we're looking for. The `ObjectOnly` type ensures that the undefined does not turn the intersection into a `never` and returns at least an object definition in response.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13165)


packages/merge-styles/src/IStyleSet.ts
```diff
@@ -22,7 +22,7 @@ export type __MapToFunctionType<T> = Extract<T, Function> extends never
  * It may optionally contain style functions for sub components in the special `subComponentStyles`
  * property.
  */
-export type IStyleSet<TStyleSet extends IStyleSet<TStyleSet>> = {
+export type IStyleSet<TStyleSet extends IStyleSet<TStyleSet> = any> = {
```
xugao, line 25:
```suggestion
export type IStyleSet<TStyleSet extends IStyleSet> = {
```

xugao: nvm this is not right

dzearing:
Something doesn't seem right... :)

packages/merge-styles/src/IStyleSet.ts
```diff
@@ -22,7 +22,7 @@ export type __MapToFunctionType<T> = Extract<T, Function> extends never
  * It may optionally contain style functions for sub components in the special `subComponentStyles`
  * property.
  */
-export type IStyleSet<TStyleSet extends IStyleSet<TStyleSet>> = {
+export type IStyleSet<TStyleSet extends IStyleSet<TStyleSet> = { [key: string]: any }> = {
```
xugao, line 25: why is this change needed?

packages/merge-styles/etc/merge-styles.api.md
```diff
@@ -4,29 +4,28 @@
 
 ```ts
 
+// Warning: (ae-forgotten-export) The symbol "ObjectOnly" needs to be exported by the entry point index.d.ts
```
xugao, line 7: 1. can you add in PR description why this change is needed? what scenario was failing with which change made?

xugao:
is the removal of `extends` the change that fixed the error in 3.9?

packages/merge-styles/etc/merge-styles.api.md
dzearing: Think about this case:

```tsx
const result = concatStyleSets({ a: { ... } } , undefined)
```

What is the type of result? It would be:

```tsx
{ a: { ... } } & undefined
```

...Which results in `never`. Which is not quite what we're looking for. The `ObjectOnly` type ensures that the undefined does not turn the intersection into a `never` and returns at least an object definition in response.

packages/merge-styles/src/IStyleSet.ts
dzearing: Originally when defining all of the concat/merge style definitions, I had provided the default in all of the input use cases as`IStyleSet<any>`. That felt wrong for a number of reasons, and instead I moved the default initializer here (to control the default in one place.) Then I thought `any` was too liberal and scoped it to default to an object.

packages/merge-styles/etc/merge-styles.api.md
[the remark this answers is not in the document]
dzearing: done

ecraig12345:
@dzearing Can you check whether this fixes #13132?

dzearing:
@ecraig12345 yup :)

PR #13167, opened 2020-05-14
Slider: convert to scss

joschect:
#### Pull request checklist

#### Description of changes

Convert Slider to use scss

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13167)


packages/react-next/src/components/Slider/Slider.scss
```diff
@@ -0,0 +1,264 @@
...
+@mixin thumbActiveStyles {
+  .thumb {
+    border: 2px solid var($semanticColorsInputBackgroundCheckedHovered);
+    @include high-contrast {
+      bordercolor: Highlight;
+    }
+  }
+}
+
+@mixin slideBoxctiveZeroTickStyles {
```
khmakoto, line 49:
```suggestion
@mixin slideBoxActiveZeroTickStyles {
```

packages/react-next/src/components/Slider/Slider.scss
```diff
@@ -0,0 +1,264 @@
...
+  background: transparent;
+  border: none;
+  flex-grow: 1;
+  line-height: 28px;
+  display: flex;
+  align-items: center;
+  height: 28px;
+  width: auto;
+  padding: 0 8px; // Make room for thumb at ends of line
+  @include focus-border();
```
khmakoto, line 93: To keep this in the same order shouldn't you have this include at the top of the class?

packages/react-next/src/components/Slider/Slider.scss
```diff
@@ -0,0 +1,264 @@
...
+
+  .enabled & {
```
khmakoto, line 95: I think this concept of `enabled` style does not exist on the original `.styles.ts` file right?

packages/react-next/src/components/Slider/Slider.scss
```diff
@@ -0,0 +1,264 @@
...
+  box-sizing: border-box;
+  background: var($semanticColorsInputBackground);
+  display: block;
+  width: 16px;
+  height: 16px;
+  position: absolute;
+  top: -6px;
+  transform: translateX(-50%);
+
+  // TODO either remove or implement for RTL transform: translateX(50%);
```
khmakoto, line 131: 👍

packages/react-next/src/components/Slider/Slider.scss
```diff
@@ -0,0 +1,264 @@
...
+    height: 1px;
+    transform: translateX(-6px);
+    // todo either remove or implement for rtl translateX(6px)
+  }
+}
+
+.valueLabel {
+  flex-shrink: 1;
+  width: 30px;
+  line-height: 1px; // using a string here meaning its relative to the size of the font
```
khmakoto, line 254: It's probably ok to remove this comment now.

packages/react-next/src/components/Slider/Slider.tsx
```diff
@@ -1,14 +1,74 @@
...
+    const globalClassNames = getGlobalClassNames(GlobalClassNames, theme);
+
+    const rootPropClasses = [
+      disabled ? classes.disabled : classes.enabled,
+      vertical && classes.vertical,
+      disabled ? globalClassNames.disabled : globalClassNames.enabled,
+      vertical ? globalClassNames.column : globalClassNames.row,
+    ];
+
+    const slideBoxPropClasss = [showTransition && classes.showTransitions, showValue && globalClassNames.showValue];
```
khmakoto, line 45:
```suggestion
    const slideBoxPropClasses = [showTransition && classes.showTransitions, showValue && globalClassNames.showValue];
```

packages/react-next/src/components/Slider/Slider.tsx
```diff
@@ -1,14 +1,74 @@
...
+    disabled: boolean | undefined,
+    vertical: boolean | undefined,
+    titleLabelClassName: string | undefined,
+    showTransition: boolean | undefined,
+    showValue: boolean | undefined,
+  ) => {
+    const globalClassNames = getGlobalClassNames(GlobalClassNames, theme);
+
+    const rootPropClasses = [
+      disabled ? classes.disabled : classes.enabled,
```
khmakoto, line 39: I don't think the concept of `enabled` exists in the original implementation.

packages/react-next/src/components/Slider/__snapshots__/Slider.test.tsx.snap
```diff
@@ -33,24 +22,18 @@ exports[`Slider renders correctly 1`] = `
...
           margin-top: 0px;
           overflow-wrap: break-word;
-          padding-bottom: 0px;
-          padding-left: 0px;
-          padding-right: 0px;
-          padding-top: 0px;
+          padding-bottom: 5px;
+          padding-left: 0;
+          padding-right: 0;
+          padding-top: 5px;
```
khmakoto, line 28: Interesting that we are seeing a change here in the actual attributes as opposed to just the classname being applied.

apps/vr-tests/src/stories/SliderNext.stories.tsx
```diff
@@ -0,0 +1,55 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import * as React from 'react';
+import Screener from 'screener-storybook/src/screener';
+import { storiesOf } from '@storybook/react';
+import { FabricDecoratorTall } from '../utilities';
+import { Slider } from '@fluentui/react-next';
+
+// tslint:disable:jsx-ban-props
+storiesOf('SliderNext', module)
```
xugao, line 9:
```suggestion
storiesOf('Slider Next', module)
```

packages/react-next/src/components/Slider/__snapshots__/Slider.test.tsx.snap
joschect: I spent a fair amount of time checking into it, I think that this makes sense since the underlying component is ultimately a Label, so we'd expect to see a mix of css and css-in-js. I think at least.

joschect: So all of those inline styles are the css-in-js being applied from the underlying label.

joschect:
@khmakoto I'll address the todos in the next iteration!

packages/react-next/src/components/Slider/__snapshots__/Slider.test.tsx.snap
```diff
@@ -239,22 +91,18 @@ exports[`Slider renders correctly 1`] = `
...
-            flex-shrink: 1;
             font-family: 'Segoe UI', 'Segoe UI Web (West European)', 'Segoe UI', -apple-system, BlinkMacSystemFont, 'Roboto', 'Helvetica Neue', sans-serif;
             font-size: 14px;
             font-weight: 600;
-            line-height: 1;
-            margin-bottom: 0;
-            margin-left: 8px;
-            margin-right: 8px;
-            margin-top: 0;
+            margin-bottom: 0px;
```
xugao, line 97: hmm why is this block of css-in-js  not removed

packages/react-next/src/components/Slider/Slider.tsx
joschect: It's there as a global class name and isn't one of the internal ones. I prefer the explicit use of checking for enabled vs the :not selector but that's really just my style. @xugao Do you have  an opinion of using the :not selector vs using an extra className

packages/react-next/src/components/Slider/Slider.scss
```diff
@@ -0,0 +1,264 @@
...
+    @include high-contrast {
+      backgroundcolor: GrayText;
+    }
+  }
+
+  .vertical & {
+    width: 16px;
+    height: 1px;
+    transform: translateX(-6px);
+    // todo either remove or implement for rtl translateX(6px)
```
xugao, line 247:
```suggestion
    // TODO: either remove or implement for rtl translateX(6px)
```

packages/react-next/src/components/Slider/__snapshots__/Slider.test.tsx.snap
joschect: @xugao I believe that this is still here because the underlying component is a label which still has css-in-js and so it would remain. I'm not sure about that though.

packages/react-next/src/components/Slider/Slider.scss
joschect: Resolving in favor of the talking about enabled in the other conversation

packages/react-next/src/components/Slider/__snapshots__/Slider.test.tsx.snap
xugao: gotcha, makes sense

packages/react-next/src/components/Slider/Slider.tsx
xugao: let's have one prop -> one modifier class mapping. for now at least - this can be subject to change. i think the more modifier classes we have, the more complex things can get in the css land. so i'd like to have the fewer the better. another reason would we we can do something generic for all components when creating these prop -> class mapping when key is the same:
```
    const propControlledClasses = [
      disabled && classes.disabled,
      reversed && classes.reversed,
      checked && classes.checked,
      indeterminate && classes.indeterminate,
      isUsingCustomLabelRender && classes.isUsingCustomLabelRender,
    ];
```
I get that using `:not` can feel little funky, but there shouldn't be a strong downside for that. it doesn't increase specificity and perf wise should make no difference.

joschect:
@xugao Added perf test, should be all set!

JasonGore:
I still kinda want to be codeowner on `perf-test` for a bit, but this shouldn't require or block test scenarios. `fluentui/perf-test` actually allows perf tests to reside in source, if anyone wants to look at consolidating them. For now, just let me know if I'm blocking on any PRs. ;)

xugao:
@JasonGore - i want to consolidate them, once i find a good time.. you can make me co-owner if you don't mind

JasonGore:
You can add yourself if you'd like. I'd at least like to see changes going in. Thanks!

joschect:
Reopening to trigger another build

joschect:
I have a good feeling about this build, any last minute comments before I commit?

ecraig12345:
> I have a good feeling about this build, any last minute comments before I commit?



...sorry not sorry

packages/react-next/src/components/Slider/Slider.scss
```diff
@@ -0,0 +1,265 @@
...
+
+  &.disabled {
+    background: var($semanticColorsDisabledText);
+    @include high-contrast {
+      backgroundcolor: GrayText;
+      bordercolor: GrayText;
+    }
+  }
+
+  &.showTransitions {
```
xugao, line 195: let's remove this block

packages/react-next/src/components/Slider/Slider.scss
```diff
@@ -0,0 +1,265 @@
...
+  &.disabled {
+    background: var($semanticColorsDisabledBackground);
+    @include high-contrast {
+      bordercolor: GrayText;
+    }
+  }
+
+  &.showTransitions {
+    transition: 3 3 3;
+  }
```
xugao, line 219: let's remove this block

packages/react-next/src/components/Slider/Slider.scss
```diff
@@ -0,0 +1,262 @@
...
+      bordercolor: GrayText;
+    }
+  }
+
+  /**
+   * TODO Implement animations
+   * .showTransitions && {
+   * transition: `left ${AnimationVariables.durationValue3} ${AnimationVariables.easeFunction1}`;
+   * };
+   */
```
behowell, line 155: FYI I'm going to add animation constants in my PR [#13370](https://github.com/microsoft/fluentui/pull/13370/files#diff-108d44f6cfaa5f4557efce5da2f23aed). Feel free to comment there if you think there's a better way to add the constants.

PR #13171, opened 2020-05-15
Copying SwatchColorPicker to react-next, in preparation for converting to function component

czearing:
#### Pull request checklist

This change does not modify `SwatchColorPicker`; it simply moves it to react-next for better diffing in the actual conversion PR.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13171)


packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
```diff
@@ -0,0 +1,158 @@
+import * as React from 'react';
+import { ITheme, mergeStyleSets } from '../../Styling';
+import { classNamesFunction, memoizeFunction } from '../../Utilities';
+import { getColorFromString } from 'office-ui-fabric-react/lib/utilities/color/getColorFromString';
+import { GridCell } from 'office-ui-fabric-react/lib/utilities/grid/GridCell';
+import { IGridCellProps } from 'office-ui-fabric-react/lib/utilities/grid/GridCell.types';
+import { getStyles as getActionButtonStyles } from 'office-ui-fabric-react/lib/components//Button/ActionButton/ActionButton.styles';
```
ecraig12345, line 7: This is weird...
```suggestion
import { getStyles as getActionButtonStyles } from 'office-ui-fabric-react/lib/components/Button/ActionButton/ActionButton.styles';
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.doc.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { SwatchColorPickerBasicExample } from './examples/SwatchColorPicker.Basic.Example';
+
+import { IDocPageProps } from '../../common/DocPage.types';
+
+const SwatchColorPickerBasicExampleCode = require('!raw-loader!office-ui-fabric-react/src/components/SwatchColorPicker/examples/SwatchColorPicker.Basic.Example.tsx') as string;
```
ecraig12345, line 6: Update all the imports in this file to be from this package not oufr

packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
```diff
@@ -0,0 +1,158 @@
+import * as React from 'react';
+import { ITheme, mergeStyleSets } from '../../Styling';
+import { classNamesFunction, memoizeFunction } from '../../Utilities';
+import { getColorFromString } from 'office-ui-fabric-react/lib/utilities/color/getColorFromString';
+import { GridCell } from 'office-ui-fabric-react/lib/utilities/grid/GridCell';
```
ecraig12345, line 5: You should move `utilities/grid` to react-next at the same time since it's tightly coupled to (and only used by) SwatchColorPicker.

packages/react-icons/package.json
```diff
@@ -26,7 +26,7 @@
     "update-snapshots": "just-scripts jest -u"
   },
   "devDependencies": {
-    "@types/enzyme": "3.10.3",
+    "@types/enzyme": "3.10.5",
```
ecraig12345, line 29: If you're going to update the enzyme version it needs to be updated everywhere, not just here (and probably best to do as a separate PR).

ecraig12345:
Looks like you need to update snapshots.

PR #13180, opened 2020-05-15
chore(form): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

Converting `Form` from class component to functional. Any props are not longer passed to styles function.

Related to #12237

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Form/Form.tsx
```diff
@@ -38,58 +47,88 @@ export interface FormProps extends UIComponentProps, ChildrenComponentProps {
...
-    fields: customPropTypes.collectionShorthand,
-    onSubmit: PropTypes.func,
-  };
+    rtl: context.rtl,
+  });
 
-  static defaultProps = {
-    as: 'form',
-  };
+  const getA11yProps = useAccessibility(accessibility, {
```
pompomon, line 74:
```suggestion
  const getA11yProps = useAccessibility<never>(accessibility, {
```

PR #13185, opened 2020-05-15
mergeStyles: addressing shallow partial issue found by TS 3.9.2

dzearing:
We were still seeing some scenarios where typings would not work with TS 3.9.2, due to the increase in type checking. 

The problem being reported was that subComponentStyles would require EVERY sub component in partial results. This should be address in this PR which correctly applies a deep partial typing.

### DIFF
--- a/packages/merge-styles/src/IStyleFunction.ts
+++ b/packages/merge-styles/src/IStyleFunction.ts
@@ -1,12 +1,19 @@
 import { IStyleSet } from './IStyleSet';
 
 /**
+ * Deep partial
+ */
+export type DeepPartial<T> = {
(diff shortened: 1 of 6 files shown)


packages/merge-styles/src/IStyleFunction.ts
```diff
@@ -1,12 +1,19 @@
 import { IStyleSet } from './IStyleSet';
 
+/**
+ * Deep partial
```
ecraig12345, line 4: A little more description please? 😆

dzearing: Done!

PR #13190, opened 2020-05-15
enabledFeatures switches

phkuo:
#### Pull request checklist

#### Description of changes

Adds enabledFeatures switches that lets new code be checked in "dark" and off, and only turned on using global settings.

Documentation here: https://github.com/microsoft/fluentui/wiki/enabledFeatures-Switches


packages/utilities/src/IFabricConfig.ts
```diff
@@ -0,0 +1,32 @@
...
+   */
+  theme?: IPartialTheme;
+
+  /**
+   * A list of GUIDs used to enable new features.
+   * A feature is enabled if the GUID associated with it is in the list and is set to "true".
+   * All letters in GUIDs must be uppercase letters.
+   * See further documentation at https://github.com/microsoft/fluentui/wiki/enabledFeatures-Switches
+   */
+  enabledFeatures?: { [guid: string]: boolean };
```
JustSlone, line 31: Is using a guid here overkill for this feature? I think we'll only use this in a few limited cases such that a string should be sufficient. The nice part about using a string is that the flag can be self-describing.

phkuo: that's reasonable

packages/utilities/src/IFabricConfig.ts
```diff
@@ -0,0 +1,26 @@
...
+   */
+  mergeStyles?: IStyleSheetConfig;
+
+  /**
+   * A list of features to enable.
+   * A feature is enabled if the name associated with it is in the list and is set to "true".
+   * `name`s MUST be globally unique.
+   * See further documentation at https://github.com/microsoft/fluentui/wiki/enabledFeatures-Switches
+   */
+  enabledFeatures?: { [name: string]: boolean };
```
dzearing, line 25: A few thoughts

1. I like that utilities is decoupled from knowing the flags available.
2. I don't like that a user has no clue about which flags are available.
3. Please, no guids in the code. Not even in examples.

Overall I'm not sure of a cleaner solution, other than:

OUFR package exports IFabricConfig which redefines `enabledFeatures` with type safety including flags that apply to the suite.

packages/utilities/src/enabledFeatures.ts
```diff
@@ -0,0 +1,21 @@
...
+import { getWindow } from './dom/getWindow';
+
+/**
+ * Used to determine if a portion of code should run.
+ * See https://github.com/microsoft/fluentui/wiki/enabledFeatures-Switches for more details.
+ * @param name - globally unique name associated with this switch
+ * @param owner - alias or name of the owner of this switch.
+ * @param date - date this switch was created, preferably MM/DD/YY e.g. 12/31/20, or longform e.g. Dec 31, 2020
+ */
+export function isFeatureEnabled(name: string, owner: string, date: string) {
```
dzearing, line 11: owner and date seems like there is no point.

dzearing: Why this this function needed at all?

We should consider a `getSettings().features.myFlag`, available at the suite.

dzearing: Let's do this:

1. Move or redefine the config in `oufr`. 
2. If there isn't one already add a `getSettings` helper to access FabricConfig (so renaming it later can be managed in one place)
3. The interface returned is type safe.

Then adding any features that need to be manually turned on can by typesafe.

Hope that makes sense. let me know if you have any qs or need any help.

packages/utilities/src/IFabricConfig.ts
phkuo: Type safety would mean that code enabling features will break once we graduate the feature, though, which is explicitly something we're trying to avoid.

packages/utilities/src/enabledFeatures.ts
phkuo: `owner` and `date` are there so we know who to track down about a feature that's been in the code, so we can have policies like "must graduate features in 6 months". It helps with maintenance.

packages/utilities/src/IFabricConfig.ts
dzearing: Once we graduate a feature, we have many options. In a major release we default the feature to on and mark the flag as deprecated. Things still build, but there's a whole major release to remove the flag. In the following major, we remove the flag. You clean up your code.

I don't think "no typings" is the way to make the flags useful. Nobody will know what flags can be used at all, without parsing all the code for secrets.

dzearing: Another idea would be along the lines of a provider:

@fluentui/react-feature-flags package, super lightweight, no type safety.

```tsx
export const FeatureFlagsContext = React.context<any>({});

export type FeaturesProviderType = <TProps>(props: TProps) => JSX.Element;
export const BaseFeaturesProvider = function<TProps = {}>(props: TProps) => {
 
 return (
  <FeatureFlagsContext.Provider value={rest}>{children}<FeatureFlagsContext.Provider>
  );
};

export const useBaseFeatureFlags = function<TProps> (): TProps => useContext(FeatureFlagsContext);
```

Now a package, like `@fluentui/react` can re-export this and provide the flag definitions:

```tsx
import { BaseFeaturesProvider } from '@fluentui/react-feature-flags';

// Type safety and intellisense over what features I can enable!
export interface FeaturesProps {
  flag1?: boolean;
  flag2?: boolean;
}

// Re export with type safety. Easy to combine types if managing multiple sub package features.
export const FeaturesProvider: FeaturesProviderType<FeaturesProps> = BaseFeaturesProvider;
export const FeaturesConsumer: FeaturesConsumerType<FeaturesProps> = FeaturesContext.Consumer;
export const useFeatures: FeaturesHook<FeaturesProps> = useBaseFeatureFlags;
```

Component author experience is easy:

```tsx
import { useFeatureFlags } from '../...';

const Component = (...) => {
  const { flag1 } = useFeatureFlags();

  ...do something with flag1
};

App developer experience is easy:

<FeaturesProvider flag1>
  <App .. />
</FeaturesProvider>
```
No global references or side effects.

This layering can abstract any behavior we want. Flags could be represented as an object as well rather than a boolean, so that additional metadata like generated warnings, owner names, dates, etc and such could be provided.

Would this work? Ultimately I think we want to shoot for less reliance on globals and window objects (especially with the word Fabric, when we are working on renaming) and undocumented, magic flags that are not type safe are all things that create confusion and debt.

packages/utilities/src/enabledFeatures.ts
ecraig12345: I think it would be cleaner to leave off `owner` and `date` here and add a wrapper in SP code which takes those.

PR #13193, opened 2020-05-15
CommandBar: Add optional role prop

erichdev:
#### Pull request checklist

#### Description of changes

Add optional `role` prop to `CommandBar` to override the default role of `menubar`. 

#### Focus areas to test

(optional)

### DIFF
--- a/change/office-ui-fabric-react-2020-05-15-16-48-05-commandbar-role.json
+++ b/change/office-ui-fabric-react-2020-05-15-16-48-05-commandbar-role.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "CommandBar: Add optional aria role prop",
+  "packageName": "office-ui-fabric-react",
+  "email": "erabelle@microsoft.com",
+  "commit": "e788a5dfe348f03d21a453d789bd811f25e740ca",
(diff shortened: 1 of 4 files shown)


erichdev:
Accessibility design for SPO gives the commandbar a role of region.

erichdev:
> This is interesting, we actually do already support a `role` prop on CommandBar due to `ICommandBarProps` extending from `React.HTMLAttributes<HTMLDivElement>`. However, that `role` is applied to the root ResizeGroup rather than the inner FocusZone. So to avoid a behavior change you'll need to come up with a different name for the inner role prop.
> 
> Backing up a little, I also question whether [`region`](https://www.w3.org/TR/wai-aria-1.2/#region) is the right role for a CommandBar rather than `menubar`--both semantically and because the `menuitem` child items are [required to have a `menu` or `menubar` as parent](https://www.w3.org/TR/wai-aria-1.2/#menuitem) and might behave incorrectly if they don't.

You're right, I can wrap the CommandBar in a region instead rather than changing the CommandBar's role entirely. And good reminder to look at what HTMLAttributes are already extended. Thanks @ecraig12345 ! Closing PR.

PR #13196, opened 2020-05-16
Link: Using compose in react-next version

khmakoto:
#### Pull request checklist

#### Description of changes

This PR changes the inner way in which `Link` is built by starting to use `compose` from `@fluentui/react-compose`.

#### Focus areas to test

(optional)


packages/react-next/src/components/Link/Link.tsx
```diff
@@ -1,40 +1,10 @@
...
-  },
-);
-
-const getStaticStyles = (props: ILinkStyleProps): Required<ILinkStyles> => {
-  const { className, isButton, isDisabled, theme } = props;
-
-  return getStaticStylesMemoized(theme!, className, isButton, isDisabled);
-};
-
-export const Link: React.FunctionComponent<ILinkProps> = styled<ILinkProps, ILinkStyleProps, ILinkStyles>(
```
xugao, line 33: hmm aren't you breaking a lot of logic which was in `styled` but not in `compose`? such as `Customizer` and styles pro.

packages/react-next/src/components/Link/Link.tsx
```diff
@@ -1,40 +1,10 @@
...
-import { css } from 'office-ui-fabric-react/lib/Utilities';
+import { LinkProps } from './Link.types';
 
-const GlobalClassNames = {
-  root: 'ms-Link',
-};
-
-const getStaticStylesMemoized = memoizeFunction(
-  (theme: ITheme, className?: string, isButton?: boolean, isDisabled?: boolean) => {
-    const globalClassNames = getGlobalClassNames(GlobalClassNames, theme);
```
xugao, line 15: you are also removing some static class names

packages/react-next/src/components/Link/Link.tsx
khmakoto: Yes, maybe this should have been started as a draft PR. I've added the `Do Not Merge` tag for now to account for this.

packages/react-next/src/components/Link/Link.tsx
khmakoto: Yes, I'm still trying to figure out how to completely account for all of this.

xugao: i think we need to take a smaller step. use compose without removing `styled`

packages/react-next/src/components/Link/Link.tsx
xugao: i think we need to take a smaller step. use `compose` without removing `styled`

packages/react-next/src/components/Link/Link.scss
```diff
@@ -24,7 +24,7 @@
     border-bottom: none;
   }
 
-  &.button {
+  &:not(.href) {
```
joschect, line 27: Will the root ever have a class `href`?

khmakoto: The way `compose` works right now is that it applies styling derived from props to classes that have the name of those props.

joschect: Where does that happen? I feel like that's... Not obvious right now.

packages/react-next/etc/react-next.api.md
```diff
@@ -18,6 +19,7 @@ import { IStyleFunctionOrObject } from 'office-ui-fabric-react/lib/Utilities';
 import { IStyleSet } from 'office-ui-fabric-react/lib/Styling';
 import { ITheme } from 'office-ui-fabric-react/lib/Styling';
 import { Position } from 'office-ui-fabric-react/lib/utilities/positioning';
+import * as PropTypes from 'prop-types';
```
dzearing, line 22: what is this used for?

packages/react-next/src/components/Link/Link.tsx
```diff
@@ -1,40 +1,10 @@
...
-  LinkBase,
-  getStaticStyles,
-  undefined,
-  {
-    scope: 'Link',
-  },
-);
+export const Link = compose<'a', {}, {}, LinkProps, LinkProps>(LinkBase, {
+  classes,
+  slots: {},
```
dzearing, line 8: remove?

dzearing: Problem we've discussed is:

* styles prop won't work
  * `<Link styles={{ root: { background: red }}`
  * `<Link styles={ props => { root: { background: props.theme.palette.themePrimary } } }`
* Customizer injected styles wont work
  * `<Customizer scopedSettings={{ Link: { styles: { root: { ... } } }}`

For the sake of the PR, we could keep backwards compat by just using `styled` and validate the above scenarios.

packages/react-next/src/components/Link/LinkBase.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+import { compose } from '@fluentui/react-compose';
+import { LinkProps } from './Link.types';
+import { useLink } from './useLink';
+
+export const LinkBase = compose<'a', LinkProps, LinkProps, {}, {}>(
```
xugao, line 6: Unfortunately, we need to keep `KeytipData`. no breaking changes :)

khmakoto: Done

packages/react-next/src/components/Link/Link.tsx
khmakoto: I'm using `styled` now and I added examples and visual regression tests to validate these scenarios.

packages/react-next/src/components/Link/Link.tsx
khmakoto: Resolved by reverting to use `styled`.

packages/react-next/src/components/Link/Link.tsx
khmakoto: Reverted to using `styled` for the last layer.

packages/react-next/src/components/Link/Link.doc.tsx
```diff
@@ -45,6 +47,41 @@ export const LinkPageProps: IDocPageProps = {
         };
       },
     },
+    {
+      title: 'Link Themed',
+      code: LinkThemedExampleCode,
+      view: <LinkThemedExample />,
+      styles: ({ theme }) => {
```
xugao, line 54: i don't think we want this example to be on doc site

packages/react-next/src/components/Link/Link.types.ts
```diff
@@ -1,8 +1,11 @@
 import * as React from 'react';
+import { ComponentProps, ComposeOptions, ComposeStandardStatics } from '../../utils/tempTypes';
 import { IStyle, ITheme } from '../../Styling';
 import { IRefObject, IStyleFunctionOrObject } from '../../Utilities';
 import { IKeytipProps } from '../../Keytip';
 
+// tslint:disable:interface-name
```
xugao, line 7: we should keep the I prefix until we have a clear story on how we plan to remove it

packages/react-next/src/components/Link/Link.types.ts
```diff
@@ -94,3 +92,19 @@ export interface ILinkStyleProps {
 export interface ILinkStyles {
   root: IStyle;
 }
+
+/**
+ * {@docCategory Link}
+ */
+export interface LinkSlots {
+  keytipData: React.ElementType;
```
xugao, line 100: not very sure about making this a slot, do you see a use case of this?
anything we add to the public interface, it's going to be hard to remove.

packages/react-next/src/components/Link/examples/Link.Basic.Example.tsx
```diff
@@ -1,14 +1,17 @@
 import * as React from 'react';
 import { Link } from '@fluentui/react-next/lib/Link';
+import { useFocusRects } from '@fluentui/react-next/lib/Utilities';
 
 export const LinkBasicExample: React.FunctionComponent = () => {
+  useFocusRects();
```
xugao, line 6: this should be called in LinkBase

packages/react-next/src/components/Link/useLink.ts
```diff
@@ -0,0 +1,33 @@
...
+import { LinkProps, LinkSlots, LinkSlotProps } from './Link.types';
+
+/**
+ * The useLink hook processes the Link component props and returns
+ * state, slots and slotProps for consumption by the component.
+ */
+export const useLink = (props: LinkProps, options: ComposePreparedOptions) => {
+  const { 'aria-describedby': ariaDescribedBy, as, disabled, href, keytipProps, onClick } = props;
+
+  const _onClick = (ev: React.MouseEvent<HTMLAnchorElement | HTMLButtonElement>) => {
```
xugao, line 12:
```suggestion
  const onClick = (ev: React.MouseEvent<HTMLAnchorElement | HTMLButtonElement>) => {
```

packages/react-next/src/components/Link/useLink.ts
```diff
@@ -0,0 +1,33 @@
...
+      onClick(ev);
+    }
+  };
+
+  const handledProps: LinkProps = {
+    ...props,
+    'aria-disabled': disabled,
+    as: as ? as : href ? 'a' : 'button',
+    href: disabled ? undefined : href,
+    onClick: _onClick,
```
xugao, line 25:
```suggestion
    onClick,
```

packages/react-next/src/components/Link/useLink.ts
khmakoto: I cannot name it `onClick` because I'm getting `onClick` from the props unless I rename the props `onClick` to something else. Naming suggestions appreciated.

packages/react-next/src/components/Link/examples/Link.Basic.Example.tsx
khmakoto: Why? This is not an example for `LinkBase` but for `Link`, it's just a basic example of it.

packages/react-next/src/components/Link/useLink.ts
khmakoto: See reply above.

packages/react-next/src/components/Link/examples/Link.Basic.Example.tsx
xugao: `useFocusRects` should be baked in functionality within the component. meaning when user using `Link`, they should get focus rect without explicitly do anything extra

packages/react-next/src/components/Link/useLink.ts
xugao: got it. i don't have a great suggestion. an alternative is to use `props.onClick` everywhere. but i think what you currently have is better?

packages/react-next/src/components/Link/Link.base.tsx
```diff
@@ -1,117 +0,0 @@
-import * as React from 'react';
-import { classNamesFunction } from '../../Utilities';
-import { ILink, ILinkProps, ILinkStyleProps, ILinkStyles } from './Link.types';
-import { KeytipData } from '../../KeytipData';
-
-const getClassNames = classNamesFunction<ILinkStyleProps, ILinkStyles>();
-
-const useComponentRef = (props: ILinkProps, link: React.RefObject<ILink>) => {
```
xugao, line 8: looks like this logic is missing

packages/react-next/src/components/Link/Link.doc.tsx
khmakoto: I need to put it in here to be able to see the changes when doing `yarn start`. Since the examples are easy to take out without much trouble I'd suggest keeping it here for testing purposes for the time being.

packages/react-next/src/components/Link/Link.types.ts
khmakoto: This is a weird one for me, I'd actually like to propose not bundling `Keytips` with the other components and have people wrap the components if they want this functionality. If we are indeed bundling them with the components then I don't see a reason why not to have them as slots. What do you think @xugao @dzearing @joschect?

packages/react-button/src/utils/tempTypes.ts
```diff
@@ -24,6 +26,9 @@ export type ComposeRender<TProps, TOptions> = (
 // tslint:disable-next-line:no-any
 export type ClassDictionary = any;
 
+// tslint:disable-next-line:no-any
+export type GenericDictionary = { [key: string]: any };
```
dzearing, line 30: FWIW this can be replaced with `Record<string, any>`

packages/react-next/src/components/Link/Link.base.tsx
khmakoto: Added.

ecraig12345:
Note that I just made some changes to Link as part of #13034 that will need to be ported over here (sorry).

packages/react-next/src/components/Link/Link.types.ts
khmakoto: Moved it out of slots for now.

packages/react-next/src/components/Link/Link.doc.tsx
xugao: storybook read `*.example.tsx`:
````
const req = require.context('../src/components', true, /\.Example\.tsx$/);
```

I think this `.doc.tsx` is for the website

khmakoto: Removed.

packages/react-next/src/components/Link/Link.types.ts
```diff
@@ -94,3 +91,20 @@ export interface ILinkStyleProps {
 export interface ILinkStyles {
   root: IStyle;
 }
+
+/**
+ * {@docCategory Link}
+ */
+export interface ILinkSlots {}
```
dzearing, line 98: Because `Link` is a primitive and does not have slots, I don't think this is used anywhere.

packages/react-next/src/components/Link/useLink.ts
```diff
@@ -0,0 +1,98 @@
...
+import { classNamesFunction } from '../../Utilities';
+import { ILink, ILinkProps, ILinkStyleProps, ILinkStyles } from './Link.types';
+
+const getClassNames = classNamesFunction<ILinkStyleProps, ILinkStyles>({ useStaticStyles: true });
+
+/**
+ * The useLink hook processes the Link component props and returns
+ * state, slots and slotProps for consumption by the component.
+ */
+export const useLink = (props: ILinkProps, options: ComposePreparedOptions): any => {
```
dzearing, line 12: hmm, shouldn't `any` give you a lint error?

packages/react-next/src/components/Link/useLink.ts
```diff
@@ -0,0 +1,98 @@
...
+  const { as, className, disabled, href, onClick, ref, styles, theme } = props;
+
+  useComponentRef(props, ref);
+
+  const classNames = getClassNames(styles!, {
```
dzearing, line 17: it seems weird for this to be in `useLink`.

packages/react-next/src/components/Link/useLink.ts
khmakoto: Apparently it is turned off in `react-next`. I don't know who did it but I plan to turn it on and add exceptions in a separate PR.

packages/react-next/src/components/Link/useLink.ts
khmakoto: I could move it out into the render function itself, but basically it is needed because we are still using `styled`. The drawback as we said before is that we still have styles on the base component.

khmakoto: `Toggle` also has this logic inside `useToggle` so if we're moving this outside of `useLink` we should probably move that as well.

packages/react-next/src/components/Link/useLink.ts
```diff
@@ -0,0 +1,98 @@
...
+    // Remove the disabled prop for anchor elements
+    if (RootType === 'a') {
+      return {
+        target,
+        href: disabled ? undefined : href,
+        ...restProps,
+      };
+    }
+
+    // Add the type='button' prop for button elements
```
dzearing, line 83: Just a heads up, there is a `getNativeElementProps` in utilities which will get the right props given an element type. This is called in `mergeProps` and includes more props than you have here.

packages/react-next/src/components/Link/useLink.ts
dzearing: easy to move later!

PR #13198, opened 2020-05-16
fix(Button): fix onHover colors in Teams Theme

oliwheeler:
Package affected: **@fluentui/react-northstar**

#### Description of changes

When a button is iconOnly and inverted, the onHover color
and background color should be consistent with
the colors of a button that is just inverted (iconOnly disabled).
This change makes that consistent.

**Current** behaviour:

button_inverted_iconOnly_current_behaviour

**Expected** behaviour:

button_inverted_iconOnly_expected_behaviour

##### Changed to docs site

Addition of an _inverted iconOnly_ **add button** to the [Inverted Button](https://fluentsite.z22.web.core.windows.net/components/button/definition?showCode=true&showRtl=false&showTransparent=false&showVariables=false#types-inverted) section of docs site.

#### Focus areas to test

* **@fluentui/react-northstar Button styles**


oliwheeler:
**TODO**: Update [Screener Tests](https://github.com/microsoft/fluentui/blob/master/packages/fluentui/test-a-feature.md#screener-tests)

packages/fluentui/react-northstar/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -211,6 +211,10 @@ const buttonStyles: ComponentSlotStylesPrepared<ButtonStylesProps, ButtonVariabl
           ...getIconFillOrOutlineStyles({ outline: false }),
           color: v.textColorIconOnlyHover,
```
notandrew, line 212: Wonder if these should be removed instead since your fix is to re-enforce the inverted background & foreground colors...

oliwheeler: I inverted the condition to not re-enforce the inverted background & foreground colors.

oliwheeler: @notandrew should I remove the addition of the inverted iconOnly button in the docs or should it stay and the screener tests updated?

notandrew: I'd update the screener tests - I think that's the right way to go. :)

packages/fluentui/react-northstar/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -207,11 +207,13 @@ const buttonStyles: ComponentSlotStylesPrepared<ButtonStylesProps, ButtonVariabl
...
-          color: v.textColorIconOnlyHover,
-          background: v.backgroundColorIconOnlyHover,
-        },
+        ...(!p.inverted && {
+          ':hover': {
+            ...getIconFillOrOutlineStyles({ outline: false }),
+            color: v.textColorIconOnlyHover,
+            background: v.backgroundColorIconOnlyHover,
+          },
+        }),
```
layershifter, line 216:
```suggestion
        ':hover': {
          ...getIconFillOrOutlineStyles({ outline: false }),
          color: v.textColorIconOnlyHover,
          background: v.backgroundColorIconOnlyHover,
        },
        ...(p.inverted && { ':hover': undefined })
```

Let's avoid negative conditions in styles.

oliwheeler: Actually, the change doesn't work because when `p.inverted` is true, the color and background on `:hover` should be set to:

```js
{
  color: siteVariables.colorScheme.silver.foregroundHover,
  backgroundColor: siteVariables.colorScheme.silver.backgroundHover,
}
```

which is what is was in the [first commit](https://github.com/microsoft/fluentui/pull/13198/commits/70b16921cf6dcf88aa579f29bc230fe9f6eb0762) of the PR which also did not have negative conditions, but instead reinforced the inverted background & foreground colors as @notandrew said

ecraig12345:
Is this still intended to be merged? Anything it's waiting on?

oliwheeler:
> Is this still intended to be merged? Anything it's waiting on?

There's nothing that this is waiting on. The intention is to merge, if this is considered to fix unintended behavior.

ecraig12345:
@oliwheeler Got it, just triggered a fresh build, and once that's done I'll try and get @notandrew to look at the screener changes (looks like that's what this was waiting on).

PR #13199, opened 2020-05-16
chore(Tree): Converts RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

Converting `Tree` from class component to functional. Any props are not longer passed to styles function.

Related to #12237

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -328,103 +336,106 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
       return;
     }
 
     const { id } = treeItemProps;
-    const { activeItemIds } = this.state;
+
     const siblings = getSiblings(items, id);
 
+    const nextActiveItemsIds = [...activeItemIds];
+    if (activeItemIds.indexOf(id) === -1) {
```
pompomon, line 349:
```suggestion
    if (!isActiveItem(id)) {
```
By the way, on lines 359-360 you have similar check, could they be unified?

assuncaocharles: Done

packages/fluentui/react-northstar/test/specs/components/Tree/Tree-test.tsx
```diff
@@ -71,7 +71,7 @@ const checkOpenTitles = (wrapper: ReactWrapper, expected: string[]): void => {
 };
 
 describe('Tree', () => {
-  isConformant(Tree, { autoControlledProps: ['activeItemIds'] });
+  isConformant(Tree, { constructorName: 'Tree', autoControlledProps: ['activeItemIds'] });
```
pompomon, line 74:
```suggestion
  isConformant(Tree, { constructorName: 'Tree', autoControlledProps: ['activeItemIds', 'selectedItemIds'] });
```
Let's test all autocontrolled props

assuncaocharles:
@layershifter Anything here or can I merge it?

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,108 +104,101 @@ export interface TreeItemForRenderProps {
...
-          acc,
-        );
+        return acc;
+      },
+      acc,
+    );
 
-      expandedItemsGenerator(items);
-    }
+  const iterateItems = (items, acc = []) =>
```
layershifter, line 153: Can we move out `iterateItems` and `expandedItemsGenerator` of render scope? I.e. place outside component?

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,108 +104,101 @@ export interface TreeItemForRenderProps {
...
+      design,
+      styles,
+      variables,
+    }),
+    rtl: context.rtl,
+  });
 
-  treeRef = React.createRef<HTMLElement>();
-  itemsRef = new Map<string, React.RefObject<HTMLElement>>();
+  const treeRef = React.createRef<HTMLElement>();
```
layershifter, line 197:
```suggestion
  const treeRef = React.useRef<HTMLElement>();
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -328,103 +336,103 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
+    if (item.selectableParent && item.items) {
+      return isAllGroupChecked(item.items);
+    }
+
+    return selectedItemIds && selectedItemIds.indexOf(item.id) > -1;
   };
 
-  renderContent(accessibility: ReactAccessibilityBehavior): React.ReactElement[] {
-    const { items, renderItemTitle, selectable } = this.props;
+  const contextValue: TreeRenderContextValue = {
```
layershifter, line 397: Ouch. This was previously class property and was stable between rerenders, let's use `useMemo()` to make it stable:

```tsx
const contextValue: TreeRenderContextValue = React.useMemo(() => { /* ... */ }, [onFocusParent, onSiblingsExpand, onFocusFirstChild, onTitleClick])
```

It also means that we should make stable all callbacks there via `React.useCallback()`.

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
assuncaocharles: Done

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
assuncaocharles: Done

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,129 +104,130 @@ export interface TreeItemForRenderProps {
...
-    const parentRef = this.itemsRef.get(parent);
+  const onFocusParent = React.useCallback((parent: string) => {
+    const parentRef = itemsRef.get(parent);
 
     if (!parentRef || !parentRef.current) {
       return;
     }
 
     parentRef.current.focus();
+  }, []);
```
layershifter, line 208:
```suggestion
  }, [parentRef, itemsRef]);
```

And this is dangerous as `itemsRef` are created on each render. Let's find a way to make `itemsRef` stable.

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -245,186 +254,188 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
-
-  setActiveItemIds = (e: React.SyntheticEvent, activeItemIds: string[]) => {
-    _.invoke(this.props, 'onActiveItemIdsChange', e, { ...this.props, activeItemIds });
-
-    this.setState({
-      activeItemIds,
-    });
-  };
+    setActiveItemIds(e, nextActiveItemsIds);
+  }, []);
```
layershifter, line 361:
```suggestion
  }, [items, setActiveItemIds]);
```

There variables are dependencies. `items` and `setActiveItemIds` are unstable, oops

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -245,186 +254,188 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
-    const elementToBeFocused = getNextElement(this.treeRef.current, currentElement.current);
+    const elementToBeFocused = getNextElement(treeRef.current, currentElement.current);
 
     if (!elementToBeFocused) {
       return;
     }
 
     elementToBeFocused.focus();
-  };
+  }, []);
```
layershifter, line 337:
```suggestion
  }, [treeRef, itermsRef]);
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -245,186 +254,188 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
-      selectedItemIds = selectedItemIds.filter(itemID => itemID !== id);
+      nextSelectedItemIds = nextSelectedItemIds.filter(itemID => itemID !== id);
     }
 
-    this.setSelectedItemIds(e, selectedItemIds);
+    setSelectedItemIds(e, nextSelectedItemIds);
   };
 
-  onTitleClick = (e: React.SyntheticEvent, treeItemProps: TreeItemProps, executeSelection: boolean = false) => {
+  const onTitleClick = (e: React.SyntheticEvent, treeItemProps: TreeItemProps, executeSelection: boolean = false) => {
```
layershifter, line 278: Still unstable

packages/fluentui/docs/src/examples/components/Tree/Usage/TreeMultiselectExample.shorthand.steps.ts
```diff
@@ -16,9 +16,9 @@ const config: ScreenerTestsConfig = {
         .snapshot('first expanded, not custom checkbox visible')
         .click(selectors.treeTitle(2))
         .click(selectors.treeTitle(6))
+        .snapshot('default selected states')
         .click(selectors.treeTitle(10))
         .click(selectors.treeTitle(11))
-        .snapshot('default selected states')
```
layershifter, line 21: I don't think that this change is correct

assuncaocharles: It's.

layershifter:
Issues in this PR shows that we should enable [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks).

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
assuncaocharles: Setting it in the useCallback it was breaking

packages/fluentui/docs/src/examples/components/Tree/Usage/TreeMultiselectExample.shorthand.steps.ts
assuncaocharles: Solved

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
assuncaocharles: Solved

packages/fluentui/docs/src/examples/components/Tree/Usage/TreeMultiselectExample.shorthand.steps.ts
assuncaocharles: Reverted

assuncaocharles:
@layershifter I think everything ok here now.

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -245,186 +257,197 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
+      });
 
-  getAllSelectableChildrenId = items => {
+      if (hasSubtree(treeItemProps) && !isActiveItem(id)) {
+        nextActiveItemsIds.push(id);
+      }
+
+      setActiveItemIds(e, nextActiveItemsIds);
+    },
+    [items, setActiveItemIds],
```
layershifter, line 372: `items` is unstable dependency

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,129 +104,133 @@ export interface TreeItemForRenderProps {
...
+  const setSelectedItemIds = (e: React.SyntheticEvent, selectedItemIds: string[]) => {
+    _.invoke(props, 'onSelectedItemIdsChange', e, { ...props, selectedItemIds });
 
-    parentRef.current.focus();
+    setSelectedItemIdsState(selectedItemIds);
   };
 
-  setSelectedItemIds = (e: React.SyntheticEvent, selectedItemIds: string[]) => {
-    _.invoke(this.props, 'onSelectedItemIdsChange', e, { ...this.props, selectedItemIds });
+  const setActiveItemIds = (e: React.SyntheticEvent, activeItemIds: string[]) => {
```
layershifter, line 219: Is unstable, but is used in `onSiblingsExpand` as dependency

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,129 +104,133 @@ export interface TreeItemForRenderProps {
...
-              selectedItems.push(item['id']);
-            }
-            if (item['items']) {
-              return iterateItems(item['items']);
-            }
-          });
-        };
-
-        iterateItems(items);
+const expandedItemsGenerator = (items, acc = []) =>
```
layershifter, line 111:
```suggestion
const expandedItemsGenerator = (items: TreeProps['items'] | TreeItemProps['items'], acc = []) =>
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,129 +104,133 @@ export interface TreeItemForRenderProps {
...
-          },
-          acc,
-        );
+      return acc;
+    },
+    acc,
+  );
+
+const iterateItems = (items, acc = []) =>
+  _.reduce(
```
layershifter, line 129:
```suggestion
const iterateItems = (items: TreeProps['items'] | TreeItemProps['items'], acc = []): string[] =>
  _.reduce<TreeProps['items'] | TreeItemProps['items'], string[]>(
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,129 +104,133 @@ export interface TreeItemForRenderProps {
...
+  const ElementType = getElementType(props);
+  const unhandledProps = useUnhandledProps(Tree.handledProps, props);
+
+  const [activeItemIds, setActiveItemIdsState] = useAutoControlled({
+    defaultValue: props.defaultActiveItemIds,
+    value: props.activeItemIds,
+    initialValue: expandedItemsGenerator(items),
+  });
+
+  const [selectedItemIds, setSelectedItemIdsState] = useAutoControlled({
```
layershifter, line 175:
```suggestion
  const [activeItemIds, setActiveItemIdsState] = useAutoControlled<string[]>({
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,129 +104,133 @@ export interface TreeItemForRenderProps {
...
+    defaultValue: props.defaultSelectedItemIds,
+    value: props.selectedItemIds,
+    initialValue: iterateItems(items),
+  });
+
+  const getA11yProps = useAccessibility<TreeBehaviorProps>(props.accessibility, {
```
layershifter, line 181:
```suggestion
  const getA11yProps = useAccessibility(props.accessibility, {
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,129 +104,133 @@ export interface TreeItemForRenderProps {
...
+    className,
+    design,
+    styles,
+    variables,
+  } = props;
+
+  const ElementType = getElementType(props);
+  const unhandledProps = useUnhandledProps(Tree.handledProps, props);
+
+  const [activeItemIds, setActiveItemIdsState] = useAutoControlled({
```
layershifter, line 169:
```suggestion
const [selectedItemIds, setSelectedItemIdsState] = useAutoControlled<string[]>({
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
layershifter, line 175:
```suggestion
  const [selectedItemIds, setSelectedItemIdsState] = useAutoControlled<string[]>({
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
layershifter, line 169:
```suggestion
  const [activeItemIds, setActiveItemIdsState] = useAutoControlled<string[]>({
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,129 +104,133 @@ export interface TreeItemForRenderProps {
...
+    rtl: context.rtl,
+  });
 
-    return {
-      activeItemIds,
-      selectedItemIds,
-    };
-  }
+  const treeRef = React.useRef<HTMLElement>();
+  const itemsRef = new Map<string, React.RefObject<HTMLElement>>();
```
layershifter, line 198: I already mentioned that this thing is unstable and will be recreated between rerenders

```suggestion
  const itemsRef = React.useRef(new Map<string, React.RefObject<HTMLElement>>());
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,129 +104,133 @@ export interface TreeItemForRenderProps {
...
 
-  getInitialAutoControlledState() {
-    return { activeItemIds: [], selectedItemIds: [] };
-  }
+  const onFocusParent = React.useCallback(
+    (parent: string) => {
+      const parentRef = itemsRef.get(parent);
```
layershifter, line 202:
```suggestion
      const parentRef = itemsRef.current.get(parent);
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -96,335 +112,356 @@ export interface TreeItemForRenderProps {
...
+      }
 
-  expandItems(e: React.SyntheticEvent, treeItemProps: TreeItemProps) {
-    let { activeItemIds } = this.state;
-    const { id } = treeItemProps;
-    const { exclusive, items } = this.props;
-    const siblings = getSiblings(items, id);
+      setActiveItemIds(e, nextActiveItemsIds);
+    },
+    [stableProps, setActiveItemIds, activeItemIds.length],
```
assuncaocharles, line 335: @layershifter I had to add this `activeItemsIds.length` otherwise the nested subtrees were not being expanded

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -236,7 +236,7 @@ const Tree: React.FC<WithAsProp<TreeProps>> &
 
       setSelectedItemIdsState(selectedItemIds);
     },
-    [stableProps, setSelectedItemIdsState],
+    [stableProps, setSelectedItemIdsState, selectedItemIds.length],
```
assuncaocharles, line 239: Also had to add `selectedItemIds.length ` here.

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -351,7 +351,7 @@ const Tree: React.FC<WithAsProp<TreeProps>> &
         expandItems(e, treeItemProps);
       }
     },
-    [expandItems, processItemsForSelection],
+    [hasSubtree, expandItems, processItemsForSelection],
```
layershifter, line 354: `hasSubTree` is imported, I don't think that we need it as dependency

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
layershifter: This is not cool 😯 Let's try to refactor it on Monday

assuncaocharles: Kind of makes sense, `activeItemIds` is a dep and it's changing after each expand

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -26,12 +30,24 @@ import {
...
+import {
+  getAllSelectableChildrenId,
+  isAllGroupChecked,
+  hasSubtree,
+  removeItemAtIndex,
+  getSiblings,
+  TreeContext,
+  TreeRenderContextValue,
+} from './utils';
+// @ts-ignore
```
assuncaocharles, line 45: To avoid this `//@ts-ignore` we could create a `global.d.ts` file with:

```typescript
declare module 'react-fela' {
  export const ThemeContext;
}

export {};
```

But I guess there's a reason why didn't happened... So why @layershifter?

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
assuncaocharles: Do you see any other way to solve this?

layershifter:
We should avoid `.length` dependencies via refactors otherwise these changes will break existing results with context stability.

assuncaocharles:
> We should avoid `.length` dependencies via refactors otherwise these changes will break existing results with context stability.

`activeItemIds` and `selectedItemIds` are deps for the callbacks, if we don't pass it or its length to the deps array it breaks functionality, so what would you suggest to change other than adding `activeItemids` to the deps array?

layershifter:
> > We should avoid `.length` dependencies via refactors otherwise these changes will break existing results with context stability.
> 
> `activeItemIds` and `selectedItemIds` are deps for the callbacks, if we don't pass it or its length to the deps array it breaks functionality, so what would you suggest to change other than adding `activeItemids` to the deps array?

Currently taken approach will recreate callbacks always so there is no reason to use `React.useCallback`.

image

At the same time I don't understand why `selectedItemIds.length` is a correct dependency there.

assuncaocharles:
> > > We should avoid `.length` dependencies via refactors otherwise these changes will break existing results with context stability.
> > 
> > 
> > `activeItemIds` and `selectedItemIds` are deps for the callbacks, if we don't pass it or its length to the deps array it breaks functionality, so what would you suggest to change other than adding `activeItemids` to the deps array?
> 
> Currently taken approach will recreate callbacks always so there is no reason to use `React.useCallback`.
> 
> image
> 
> At the same time I don't understand why `selectedItemIds.length` is a correct dependency there. I think that recreation of `setSelectedItemIds` triggers another dependency to be recreated. Please check this.

Same as for the `activeItemIds`, the `setSelectedItemIds` used inside is based in the last time the callback was updated, so if I select a item, I change my `setSelectedItemIds` list but the callback didn't changed so next time I trigger it will be based in a outdate `setSelectedItemIds` what will break the functionality. So since it's going to generate callback every time anyway do you agree in removing the `useCallback` here?

layershifter:
> So since it's going to generate callback every time anyway do you a

This will not work:
- `setSelectedItemIds` is a dependency for `processItemsForSelection`
- `processItemsForSelection` is a dependency  for `onTitleClick`
- `onTitleClick` is a dependency for `contextValue`

If `contextValue` will change on each render we will regress in terms of performance.

***

My proposal is try to refactor `processItemsForSelection` to make it a pure function and move it to utils. In this case `processItemsForSelection` will be no longer a dependency for `onTitleClick`.

layershifter:
Please check https://github.com/microsoft/fluentui/pull/12182 to get more context about React Context usage there.

assuncaocharles:
> > So since it's going to generate callback every time anyway do you a
> 
> This will not work:
> 
> * `setSelectedItemIds` is a dependency for `processItemsForSelection`
> * `processItemsForSelection` is a dependency  for `onTitleClick`
> * `onTitleClick` is a dependency for `contextValue`
> 
> If `contextValue` will change on each render we will regress in terms of performance.
> 
> My proposal is try to refactor `processItemsForSelection` to make it a pure function and move it to utils. In this case `processItemsForSelection` will be no longer a dependency for `onTitleClick`.

We can probably do that for `processItemsForSelection` and avoid `selectedItemIds.length` but we sill have `expandItems` that is a dep for `onItemClick` and depends on `activeItemIds`

layershifter:
I just merged #13199, it should help with dependencies stuff.

> but we sill have expandItems that is a dep for onItemClick and depends on activeItemIds

Let's make `expandItems` pure, try to find a way to refactor/solve this.

assuncaocharles:
> I just merged #13199, it should help with dependencies stuff.
> 
> > but we sill have expandItems that is a dep for onItemClick and depends on activeItemIds
> 
> Let's make `expandItems` pure, try to find a way to refactor/solve this.

I think the number of the PR is wrong?

layershifter:
> I think the number of the PR is wrong?

Oops, good call, fixed.

assuncaocharles:
@layershifter I moved `processItemsForSelection` to utils and inside `expandItems` and `onTitleClick` I am passing a callback to be used in the `setState` so when executed it will have the updated values. This way `activeItemIds` and `selectedItemIds` is not dep of any `useCallback` anymore.

assuncaocharles:
Gentle reminder @layershifter @pompomon

PR #13206, opened 2020-05-18
docs(Examples): Add hidden text for example source name

jurokapsiar:
#### Description of changes

Adds a hidden label for component example source code so that it is easier to find the source code in their preferred IDE for screen reader users.

### DIFF
--- a/packages/fluentui/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
+++ b/packages/fluentui/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
@@ -224,6 +224,10 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
     return this.kebabExamplePath;
   };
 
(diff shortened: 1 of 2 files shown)


packages/fluentui/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -224,6 +224,10 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
     return this.kebabExamplePath;
   };
 
+  getSourceFileName = () => {
+    return `Source code: ${this.props.examplePath.split('/').pop()}.tsx`;
```
miroslavstastny, line 228: This does not handle shorthand examples and the problem is that some examples have only the shorthand variant.
So instead of blindly appending `.tsx` I would keep it without that (just `ExampleName`) and the user needs to know that it is the name without optional `shorthand` and without `tsx`.

Issue #13124, opened 2020-05-12
Rename special-purpose Grid component to ButtonGrid


ecraig12345:
There's a Grid component under [`office-ui-fabric-react/src/utilities/grid`](https://github.com/microsoft/fluentui/tree/master/packages/office-ui-fabric-react/src/utilities/grid) which is only used in SwatchColorPicker and is so tightly coupled that it's unlikely to be useful in more general cases. (EDIT: or not...see below)

~~Within version 7 we should move it into the SwatchColorPicker folder to make the coupling clearer, and deprecate it.~~

~~Then in version 8 we should at least stop exporting it, and possibly also combine its logic into SwatchColorPicker itself. This will free up the name Grid for a more generally useful version we might implement in the future.~~

EDIT: Turns out it's also being used by Word Online for a generic button grid--see https://github.com/microsoft/fluentui/issues/13574#issuecomment-648436575. So unfortunately we need to keep it as a separate component, but it should definitely be renamed to `ButtonGrid` or something to free up the `Grid` name.

- [x] Version 7
- [ ] Version 8

Issue #10516, opened 2019-09-19
Make SwatchColorPicker properly controllable


whykushal93:
#### Describe the feature that you would like added
Currently, to make the `SwatchColorPicker` fully controllable we need to use a prop `isControlled`. This prop was added temporarily so that we do not introduce breaking changes. 

But when we next do a major update, we should properly fix up the `SwatchColorPicker` code so that it is consistent with how other components allow themselves to be fully controlled. 

The mechanism suggested by @ecraig12345 is to have a prop called `defaultSelectedId` which would be mutually exclusive in usage to `selectedId` prop. When the user uses the `defaultSelectedId`, we would make the component uncontrolled. Otherwise it will be fully controlled. 

#### What component or utility would this be added to
SwatchColorPicker

#### Have you discussed this feature with our team, and if so, who
Discussed the implementation with @ecraig12345. I can take up the work when we start doing work for the next upgrade.

whykushal93:
This will be a breaking change and will need to done when moving to the next major version.

aneeshack4:
@whykushal93 Thanks for discussing this with a team member already! I'll reassign this to you since you're planning on taking this up for the next major release.

ecraig12345:
@czearing Here's the whole process for fixing SwatchColorPicker controlled behavior:
- [ ] Deprecate `isControlled`, with a runtime warning that you should just provide `selectedId` instead (and ignore the `isControlled` value)
- [ ] Add a prop `defaultSelectedId`
- [ ] Add a prop `onChange?: (event?: /*type*/, id?: string, color?: string)` and deprecate `onColorChanged`
  - [ ] The actual click/change handler should call both `onChange` and `onColorChanged`
- [ ] Add a runtime controlled usage warning (`useWarnings` `controlledUsage` option)
- [ ] Use the `useControllableValue` hook for proper controlled behavior (and probably call `useMemo` on the result to map from `selectedId` to `selectedIndex`)
- [ ] Document the change in the release notes:
  - If they only want to set the initial value and don't need manual updates, use `defaultSelectedId`
  - If they want to make manual updates there are 2 options:
    - Provide `selectedId` and `onChange`, and manually update `selectedId` whenever `onChange` is called
    - Provide `defaultSelectedId` and the built-in `key` prop (set to some arbitrary value). When the color must be manually updated, change `key` to re-create the component.
  - `isControlled` is now ignored, and inferred from presence of `selectedId`/`defaultSelectedId`
[...]

PR #13214, opened 2020-05-18
Updating SwatchColorPicker and ColorPickerGridCell to a function component inside React-Next

czearing:
#### Pull request checklist

#### Description of changes
Updating `SwatchColorPicker` and `ColorPickerGridCell` to a function component within React-Next.

Note that SwatchColorPicker had setTimeout logic for disabling behaviors on mouse-wheel. After having a conversation with dzearing this was removed but we are still following up with the original author.

#### Focus areas to test
SwatchColorPicker and ColorPickerGridCell


ecraig12345:
I'll try and take a look at this later, but something to think about: can you look into converting or possibly just removing the Grid component in src/utilities/grid as part of this work? (Or at minimum moving it side-by-side with SwatchColorPicker since it's only used there.) See #13124 for background.

packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
```diff
@@ -2,8 +2,8 @@ import * as React from 'react';
 import { ITheme, mergeStyleSets } from '../../Styling';
 import { classNamesFunction, memoizeFunction } from '../../Utilities';
 import { getColorFromString } from 'office-ui-fabric-react/lib/utilities/color/getColorFromString';
-import { GridCell } from '../../Utilities/grid/GridCell';
-import { IGridCellProps } from '../../Utilities/grid/GridCell.types';
+import { GridCell } from 'office-ui-fabric-react/lib/utilities/grid/GridCell';
```
ecraig12345, line 5: Looks like this import got reverted on merge

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -13,7 +11,7 @@ import {
   ISwatchColorPickerStyleProps,
   ISwatchColorPickerStyles,
 } from './SwatchColorPicker.types';
-import { Grid } from '../../Utilities/grid/Grid';
+import { Grid } from 'office-ui-fabric-react/lib/utilities/grid/Grid';
```
ecraig12345, line 14: Revert this since Grid has been copied

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -26,226 +24,93 @@ const getClassNames = classNamesFunction<ISwatchColorPickerStyleProps, ISwatchCo
...
-  } as ISwatchColorPickerProps;
-
-  private _id: string;
-  private _cellFocused: boolean;
-
-  private navigationIdleTimeoutId: number | undefined;
-  private isNavigationIdle: boolean;
-  private readonly navigationIdleDelay: number = 250 /* ms */;
-  private async: Async;
+export const SwatchColorPickerBase = (props: ISwatchColorPickerProps) => {
```
ecraig12345, line 27:
```suggestion
export const SwatchColorPickerBase: React.FunctionComponent<ISwatchColorPickerProps> = props => {
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -26,226 +24,93 @@ const getClassNames = classNamesFunction<ISwatchColorPickerStyleProps, ISwatchCo
...
+  const [selectedIndex, setSelectedIndex] = React.useState();
```
ecraig12345, line 28:
```suggestion
  const [selectedIndex, setSelectedIndex] = React.useState<number | undefined>();
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -26,226 +24,93 @@ const getClassNames = classNamesFunction<ISwatchColorPickerStyleProps, ISwatchCo
...
+  const id = props.id || getId('swatchColorPicker');
```
ecraig12345, line 29:
```suggestion
  const id = useId('swatchColorPicker', props.id);
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -26,226 +24,93 @@ const getClassNames = classNamesFunction<ISwatchColorPickerStyleProps, ISwatchCo
...
+
+  const classNames = getClassNames(styles!, {
+    theme: props.theme!,
+    className,
+    cellMargin,
+  });
 
   // Add an index to each color cells. Memoizes this so that color cells do not re-render on every update.
-  private _getItemsWithIndex = memoizeFunction((items: IColorCellProps[]) => {
+  const getItemsWithIndex = memoizeFunction((items: IColorCellProps[]) => {
```
ecraig12345, line 56: This can go outside the component body (and convert the comment to a doc comment)

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -26,226 +24,93 @@ const getClassNames = classNamesFunction<ISwatchColorPickerStyleProps, ISwatchCo
...
-  private _cellFocused: boolean;
-
-  private navigationIdleTimeoutId: number | undefined;
-  private isNavigationIdle: boolean;
-  private readonly navigationIdleDelay: number = 250 /* ms */;
-  private async: Async;
+export const SwatchColorPickerBase = (props: ISwatchColorPickerProps) => {
+  const [selectedIndex, setSelectedIndex] = React.useState();
+  const id = props.id || getId('swatchColorPicker');
+  let cellFocused: boolean;
```
ecraig12345, line 30: To get this to persist properly between renders (the way it would have in the class), you need to store it in a ref.

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -26,226 +24,93 @@ const getClassNames = classNamesFunction<ISwatchColorPickerStyleProps, ISwatchCo
...
+  let currentSelectedIndex: number | undefined = undefined;
```
ecraig12345, line 31: Move the declaration of this to the spot where it's used (having it at the top is a bit misleading since it's essentially just a temp variable used in one spot)

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -26,226 +24,93 @@ const getClassNames = classNamesFunction<ISwatchColorPickerStyleProps, ISwatchCo
...
 
   /**
    * Callback passed to the GridCell that will manage triggering the onCellHovered callback for mouseEnter
    */
-  private _onMouseEnter = (ev: React.MouseEvent<HTMLButtonElement>): boolean => {
-    if (!this.props.focusOnHover) {
-      return !this.isNavigationIdle || !!this.props.disabled;
+  const onMouseEnter = (ev: React.MouseEvent<HTMLButtonElement>): boolean => {
+    if (!props.focusOnHover) {
+      return !!props.disabled;
```
ecraig12345, line 83: What happened to the `isNavigationIdle`, `onWheel`, and `onKeyDown` handling?

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -26,226 +24,93 @@ const getClassNames = classNamesFunction<ISwatchColorPickerStyleProps, ISwatchCo
...
     }
-
     const targetElement = ev.currentTarget as HTMLElement;
-
-    // If navigation is idle and the targetElement is the focused element bail out
-    // if (!this.isNavigationIdle || (document && targetElement === (document.activeElement as HTMLElement))) {
-    if (this.isNavigationIdle && !(document && targetElement === (document.activeElement as HTMLElement))) {
+    // If the targetElement is the focused element bail out
+    // if ((document && targetElement === (document.activeElement as HTMLElement))) {
```
ecraig12345, line 100: Remove the commented line

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -26,226 +24,93 @@ const getClassNames = classNamesFunction<ISwatchColorPickerStyleProps, ISwatchCo
...
-        onMouseMove={this._onMouseMove}
-        onMouseLeave={this._onMouseLeave}
-        onWheel={this._onWheel}
-        onKeyDown={this._onKeyDown}
-        height={props.cellHeight}
-        width={props.cellWidth}
-        borderWidth={props.cellBorderWidth}
-      />
-    );
+  const getSelectedIndex = (items: IColorCellProps[], selectedId: string): number | undefined => {
```
ecraig12345, line 73: This can go outside the component body (and bring back at least the first line of the doc comment)

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -271,107 +136,138 @@ export class SwatchColorPickerBase extends React.Component<ISwatchColorPickerPro
...
       return item ? onCellHovered(item.id, item.color) : onCellHovered();
     }
   };
 
-  /**
-   * Callback passed to the GridCell class that will trigger the onCellFocus callback of the SwatchColorPicker
-   */
-  private _onGridCellFocused = (item?: IColorCellProps): void => {
-    const { onCellFocused } = this.props;
+  // Callback passed to the GridCell class that will trigger the onCellFocus callback of the SwatchColorPicker
```
ecraig12345, line 155: Change this back to a doc ccomment

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -271,107 +136,138 @@ export class SwatchColorPickerBase extends React.Component<ISwatchColorPickerPro
...
-          selectedIndex: index,
-        });
+      if (props.isControlled !== true) {
+        setSelectedIndex(index);
       }
     }
   };
-}
+
+  // Render a color cell
```
ecraig12345, line 195: Change this to a doc comment

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -271,107 +136,138 @@ export class SwatchColorPickerBase extends React.Component<ISwatchColorPickerPro
...
+  if (props.selectedId) {
+    currentSelectedIndex = getSelectedIndex(props.colorCells, props.selectedId);
+    setSelectedIndex(currentSelectedIndex);
+  }
+
+  if (props.selectedId !== undefined) {
+    setSelectedIndex(getSelectedIndex(props.colorCells, props.selectedId));
+  }
+
+  if (process.env.NODE_ENV !== 'production') {
```
ecraig12345, line 233: This should only happen on first render, not every render. Probably best to do something like this:
```ts
if (process.env.NODE_ENV !== 'production') {
  // lint disable if needed (while this is technically in a conditional, which violates rules of hooks, it won't be at runtime)
  React.useEffect(() => {
    // warnings
  }, []);
}
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -271,107 +136,138 @@ export class SwatchColorPickerBase extends React.Component<ISwatchColorPickerPro
...
+  if (colorCells.length < 1 || columnCount < 1) {
+    return null;
+  }
+
+  if (props.selectedId) {
+    currentSelectedIndex = getSelectedIndex(props.colorCells, props.selectedId);
+    setSelectedIndex(currentSelectedIndex);
+  }
+
+  if (props.selectedId !== undefined) {
```
ecraig12345, line 229: Combine with previous block

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.test.tsx
```diff
@@ -3,8 +3,8 @@ import * as renderer from 'react-test-renderer';
 import { mount } from 'enzyme';
 import { SwatchColorPicker } from './SwatchColorPicker';
 import { IColorCellProps } from './ColorPickerGridCell.types';
-import { expectNodes, findNodes } from '../../common/testUtilities';
 import { resetIds } from '@uifabric/utilities';
```
ecraig12345, line 6: Should probably be from `../../Utilities`

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -271,107 +136,138 @@ export class SwatchColorPickerBase extends React.Component<ISwatchColorPickerPro
...
+    );
+  };
+
+  if (colorCells.length < 1 || columnCount < 1) {
+    return null;
+  }
+
+  if (props.selectedId) {
+    currentSelectedIndex = getSelectedIndex(props.colorCells, props.selectedId);
+    setSelectedIndex(currentSelectedIndex);
```
ecraig12345, line 226: Pretty sure you shouldn't be setting state while rendering. (In this case it might even cause an infinite rendering loop if `props.selectedId` is defined!)

See overall review comment for what needs to be done instead.

xugao:
@ecraig12345 - i say we make the breaking change you proposed bc those are the right things to do and major release is the time to do it. Other than marking the deprecations in code, could you also drop some notes [here](https://hackmd.io/vkr7uOmHSD6UOILxDwYnZA?view) so we don't forget we made this change

packages/react-next/src/Utilities/grid/Grid.base.tsx
```diff
@@ -1,98 +1,76 @@
...
-      items,
-      columnCount,
-      onRenderItem,
-
-      // tslint:disable:deprecation
-      ariaPosInSet = props.positionInSet,
-      ariaSetSize = props.setSize,
-      // tslint:enable:deprecation
+export const GridBase = React.forwardRef((props: IGridProps, ref: React.Ref<HTMLTableElement>) => {
+  const id = props.id || useId();
```
ecraig12345, line 9: You have to unconditionally generate the default ID first; otherwise you could get a conditional hook call error if `props.id` changes.
```suggestion
  const defaultId = useId();
  const id = props.id || defaultId;
```

packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
```diff
@@ -56,103 +57,82 @@ const getColorPickerGridCellButtonClassNames = memoizeFunction(
   },
 );
 
+// Validate if the cell's color is white or not to apply whiteCell style
```
ecraig12345, line 60: Please keep doc comment formatting in places where it was present before.
```suggestion
/** Validate if the cell's color is white or not to apply whiteCell style */
```

packages/react-next/src/components/SwatchColorPicker/examples/SwatchColorPicker.Basic.Example.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { SwatchColorPicker } from '@fluentui/react-next/lib/SwatchColorPicker';
+import { SwatchColorPicker } from 'office-ui-fabric-react/lib/SwatchColorPicker';
```
ecraig12345, line 2: Revert this

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -17,235 +9,106 @@ import { Grid } from '../../Utilities/grid/Grid';
...
-
-  // Add an index to each color cells. Memoizes this so that color cells do not re-render on every update.
-  private _getItemsWithIndex = memoizeFunction((items: IColorCellProps[]) => {
-    return items.map((item, index) => {
-      return { ...item, index: index };
-    });
+/**
+ *  Add an index to each color cells. Memoizes this so that color cells do not re-render on every update.
+ */
+const getItemsWithIndex = memoizeFunction((items: IColorCellProps[]) => {
```
ecraig12345, line 21: Instead of this, do `React.useMemo` inline within the component.

czearing:
If we want to do this correctly, here is a proposed interface shuffle:

```jsx
export interface ColorPickerProps extend HTMLDivElement {

  // Here's how you populate it with colors.
  colors: ColorCellProps[];
  
  // Here's how you control which color is selected.
  value: string;
  defaultValue: string;
  onChange: (ev: Event, value: string) => void;

  // Here's how you can apply standard modifiers.
  cellSize: string;
  shape: 'circle' | undefined;

  // Here's how you can customize the control.
  focusZone: ShorthandProps;
  cell: ShorthandProps;
}

export interface ColorCellProps {
  color?: string;
  label?: string;
}
```

ecraig12345:
@czearing Are you thinking of doing those prop changes as part of the function component conversion, or separately? I already outlined the prop changes I'd suggest (which are decently feasible for v8) in #10516.

PR #13216, opened 2020-05-18
Simplify Checkbox modifier class assignment

xugao:
#### Pull request checklist

#### Description of changes
No need to assign modifier classes onto every slot.
@joschect inspired me in his PR #13167 :) I didn't know I can do `.class &` in scss

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-next/src/components/Checkbox/Checkbox.scss
+++ b/packages/react-next/src/components/Checkbox/Checkbox.scss
@@ -132,15 +132,15 @@ $checkboxTransitionTiming: cubic-bezier(0.4, 0, 0.23, 1);
     pointer-events: none;
   }
 
-  &.isUsingCustomLabelRender {
+  .isUsingCustomLabelRender & {
     align-items: center;
   }
 
-  &.disabled {
+  .disabled & {
     cursor: default;
   }
 
-  &.reversed {
+  .reversed & {
     flex-direction: row-reverse;
     justify-content: flex-end;
   }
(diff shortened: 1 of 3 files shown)


packages/react-next/src/components/Checkbox/Checkbox.scss
```diff
@@ -269,18 +271,18 @@ $checkboxTransitionTiming: cubic-bezier(0.4, 0, 0.23, 1);
...
       color: GrayText;
     }
   }
 
-  &.reversed {
+  .reversed & {
     margin-right: 4px;
   }
-  &:not(.reversed) {
+  :not(.reversed) & {
```
joschect, line 285: You might want to double check that this works as expected. I found I had a lot of difficulties getting `:not(.name) &` to work appropriately.

xugao: could you elaborate on what exactly wasn't working for you?

joschect: Yeah unfortunately it doesn't seem to behave as expected :( https://codepen.io/slimemold/pen/YzyRvxr

joschect: I missed your response! It seems like using `:not(.classname) &` doesn't actually check the parent. In the scss "preview" in vs code it seems to parse away the `classname` so it's not applying the not to the parent properly 

image

You can see this in the codepen that I linked above, the `:not(classname) &` is applied to the classname it's contained in, rather than the parent.

image

xugao: AH! yeah this won't work of course. becasue `.notchecked` here definitely satisfies having ancestor `:not(.checked)`

xugao: will close the PR. @joschect  - could you fix your PR accordingly too?

PR #13223, opened 2020-05-19
DetailsList: Specifying preventDefaultWhenHandled to be false so that we don't break input navigation via keyboard

### LINKED ISSUE #13210: Not able to navigate input cursor with arrows inside TextField placed inside DetailsList
### Environment Information
- **Package version(s)**: 7.107.3 <

**Example:** https://codepen.io/Code-Artist/pen/KKdrmxK

Since version 7.107.4 our users are not able to use arrow keys to navigate an input cursor inside TextField placed inside DetailsList row.

I found out that the root of this problem is **preventDefaultWhenHandled** property recently added to FocusZone component (PR #12636). 
Same issue exists on your FocusZone example (https://developer.microsoft.com/en-us/fluentui#/controls/web/focuszone#usage -> _Nesting FocusZones in list rows_)

#### Actual behavior:

User is not able to use left and right arrow keys to change position of input cursor.

#### Expected behavior:

User is able to use left and right arrow keys to change position of input cursor.

Requested priority: Normal

Products/sites affected: yes

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13210

#### Description of changes

#12636 introduced the `preventDefaultWhenHandled` prop to `FocusZone`, which broke keyboard navigation on inputs inside of `DetailsList` due to its use of `FocusZone`. This PR fixes this problem by specifying `preventDefaultWhenHandled={false}` in all `DetailsList` related components.

#### Focus areas to test

(optional)

### DIFF
--- a/change/office-ui-fabric-react-2020-05-18-18-02-48-detailsListFocusZoneEventPrevention.json
+++ b/change/office-ui-fabric-react-2020-05-18-18-02-48-detailsListFocusZoneEventPrevention.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "DetailsList: Specifying preventDefaultWhenHandled to be false so that we don't break input navigation via keyboard.",
+  "packageName": "office-ui-fabric-react",
+  "email": "humbertomakotomorimoto@gmail.com",
+  "dependentChangeType": "patch",
+  "date": "2020-05-19T01:02:48.616Z"
+}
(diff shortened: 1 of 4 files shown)


agabuza:
Thanks a lot for fixing this issue as it was causing a lot of inconveniences for our clients.
Any timelines when fix potentially will be included in framework release.

Thanks,
Anatolii

ecraig12345:
@agabuza The change needs to be reviewed by @dzearing and then once it merges, it will be included in the daily release build (runs at 12pm UTC/5am PDT).

khmakoto:
Closing in favor of #13256.

PR #13228, opened 2020-05-19
fix(mixed): use useRef() hook in functional components

layershifter:
#### Pull request checklist

#### Description of changes

This PR fixes an issue with usage of `.createRef()` instead of `useRef()` hook in recently converted components.

#### Focus areas to test

(optional)


layershifter:
FYI: @assuncaocharles

PR #13234, opened 2020-05-19
Copying Modal to react-next, in preparation for converting to function component

czearing:
#### Pull request checklist

This change does not modify `Modal`; it simply moves it to react-next for better diffing in the actual conversion PR.


packages/react-next/src/components/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
+import * as renderer from 'react-test-renderer';
+
+import { Modal } from './Modal';
+import { ContextualMenu } from 'office-ui-fabric-react/lib/ContextualMenu';
```
xugao, line 5:
```suggestion
import { ContextualMenu } from '@fluentui/react-next/lib/ContextualMenu';
```

ecraig12345:
Looks like this has merged conflicts now, but once that's fixed it should be good to go.

PR #13236, opened 2020-05-20
DRAFT: replace styled with useClasses hook

xugao:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/react-next/src/components/Link/useLinkClasses.tsx
```diff
@@ -0,0 +1,42 @@
+import { memoizeFunction, css, useClasses } from '../../Utilities';
+import { getGlobalClassNames, ITheme } from '../../Styling';
+import { ILinkProps, ILinkStyleProps, ILinkStyles } from '../../Link';
+import * as classes from './Link.scss';
+
+export function useLinkClasses(props: ILinkProps): { [key in keyof ILinkStyles]: string } {
+  return useClasses<ILinkProps['styles'], ILinkStyleProps, ILinkStyles>({
+    customizationScopeName: 'Links',
```
xugao, line 8:
```suggestion
    customizationScopeName: 'Link',
```

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -42,14 +34,12 @@ export const CheckboxBase = React.forwardRef((props: ICheckboxProps, forwardedRe
...
-  const classNames: { [key in keyof ICheckboxStyles]: string } = getClassNames(styles!, {
-    theme: theme!,
-    className,
-    disabled,
+
+  // TODO: this should be called during `compose`
+  const classNames: { [key in keyof ICheckboxStyles]: string } = useCheckboxClasses({
+    ...props,
     indeterminate: isIndeterminate,
     checked: isChecked,
```
xugao, line 42: need to ensure these can be passes through `compose`

packages/react-next/src/components/Checkbox/useCheckboxClasses.tsx
```diff
@@ -0,0 +1,81 @@
...
+export function useCheckboxClasses(props: ICheckboxProps): { [key in keyof ICheckboxStyles]: string } {
+  return useClasses<ICheckboxProps['styles'], ICheckboxStyleProps, ICheckboxStyles>({
+    customizationScopeName: 'Checkbox',
+    useStaticStyles: true,
+    stylesProp: props.styles,
+    styleProps: {
+      className: props.className,
+      disabled: props.disabled,
+      indeterminate: props.indeterminate,
+      checked: props.checked,
```
dzearing, line 15: if the checkbox is uncontrolled, how is the uncontrolled `checked` state accessed? (In other words, the input to `useCheckboxClasses` should be `state: ICheckboxState` which in many (most) cases is just the same as the props with perhaps a few omissions and adds.

dzearing: But overall I think this is the right approach. I actually think this would enable us to replace `styled` with `compose` to glue them together. All the styling logic goes into the hook.

```tsx
const DetailsList = compose(DetailslListBase, {
  classes: useDetailsListClasses
});
```

dzearing: One other note, in trying to glue v0 and v7 together, in v0, the resolution of classes returns something different.

Here, your hook returns

```
{
  root: 'ms-Toggle root-123',
  pill: 'pill-124',
  etc
}
```

In v0, they need this:
```tsx
{
  classes: { ..same as above },
  style: {
     root: { background: 'red', ... }
     pill: { ... }
  }
}
```

PR #13240, opened 2020-05-20
feat(compose): add shorthandConfig to options

mnajdova:
`shorthandConfig` option was added to compose. If you are using `compose` in your component, make sure that you move the static `shorthandConfig` to the compose options. For backward compatibility, the factories are still handling the shorthandConfig defined as static prop too.

Before
```jsx
const Component = compose((props, ref, options) => {...});
Component.shorthandConfig = {
  mappedProp: 'content',
};
```

After
```jsx
const Component = compose((orios, ref, options) => {...}, {
  shorthandConfig: {
    mappedProp: 'content',
  };
);
```

### DIFF
--- a/packages/fluentui/react-northstar/src/utils/factories.ts
+++ b/packages/fluentui/react-northstar/src/utils/factories.ts
@@ -1,12 +1,11 @@
 import { mergeStyles } from '@fluentui/styles';
-import { ComponentWithAs } from '@fluentui/react-compose';
+import { ComponentWithAs, ComposePreparedOptions } from '@fluentui/react-compose';
 import cx from 'classnames';
 import * as _ from 'lodash';
 import * as React from 'react';
 import * as ReactIs from 'react-is';
 
 import { ShorthandValue, Props, PropsOf, ShorthandRenderFunction } from '../types';
-import { ComposePreparedOptions } from '@fluentui/react-compose/dist/src';
 
 type HTMLTag = 'iframe' | 'img' | 'input';
 type ShorthandProp = 'children' | 'src' | 'type';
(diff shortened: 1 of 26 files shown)


packages/fluentui/react-northstar/src/utils/factories.ts
```diff
@@ -246,27 +240,28 @@ export function createShorthandInternal<P>({
...
 export function createShorthand<TElementType extends React.ElementType>(
-  Component: TElementType & { shorthandConfig?: ShorthandConfig<PropsOf<TElementType>> },
+  Component: TElementType & { fluentComposeConfig?: ComposePreparedOptions<PropsOf<TElementType>> },
   value?: ShorthandValue<PropsOf<TElementType>>,
   options?: CreateShorthandOptions<PropsOf<TElementType>>,
 ): React.ReactElement;
 export function createShorthand<P>(Component, value?, options?) {
-  const { mappedProp = 'children', allowsJSX = true, mappedArrayProp } = Component.shorthandConfig || {};
+  const { mappedProp = 'children', allowsJSX = true, mappedArrayProp } =
+    (Component.fluentComposeConfig && Component.fluentComposeConfig.shorthandConfig) || {};
```
layershifter, line 264:
```suggestion
    Component.shorthandConfig || Component.fluentComposeConfig?.shorthandConfig) || {};
```

mnajdova: thanks, will update the types too to 
```
{ shorthandConfig?: ShorthandConfig<PropsOf<TElementType>>; fluentComposeConfig?: ComposePreparedOptions<PropsOf<TElementType>> }
```

I guess we need to have both...

PR #13241, opened 2020-05-20
Add dedicated touch invoke functionality to SelectionZone

soettl:
#### Pull request checklist

- [X] Addresses an existing issue: When using touch, it is not apparent which part of a `DetailsRow` and `Tile` can be used for invocation

#### Description of changes

This PR aims to make core item browsing controls like `DetailsList` and `TilesList` more touch-friendly by extending the invocation target to the full row/tile and not just the title _when using touch_. The behavior for non-touch devices stays as-is. The selection toggle targets also stay as-is, for both mouse and touch. The new behavior can be enabled by supplying the `enableTouchInvocationTarget` prop.

In order to mark an element as a touch invocation target, `SelectionZone` now supports a new `data-selection-touch-invoke` attribute which will be used as the invocation target when the user is (1) using a touch pointer and (2) clicks on or inside an element marked with said attribute. 

`DetailsRow` and `Tile` have been modified to make use of this new touch invoke attribute.

#### Open questions

Should this improved behavior be scoped behind a new prop?


packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
```diff
@@ -359,7 +360,10 @@ export class SelectionZone extends React.Component<ISelectionZoneProps, ISelecti
             }
           }
           break;
-        } else if (this._hasAttribute(target, SELECTION_INVOKE_ATTRIBUTE_NAME)) {
+        } else if (
```
ThomasMichon, line 363: Can you make this behavior opt-in in the `SelectionZone`? I don't want to accidentally break existing partner apps that expect a different behavior right now.

packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
```diff
@@ -34,6 +34,7 @@ const SELECTION_DISABLED_ATTRIBUTE_NAME = 'data-selection-disabled';
 const SELECTION_INDEX_ATTRIBUTE_NAME = 'data-selection-index';
 const SELECTION_TOGGLE_ATTRIBUTE_NAME = 'data-selection-toggle';
 const SELECTION_INVOKE_ATTRIBUTE_NAME = 'data-selection-invoke';
+const SELECTION_INVOKE_TOUCH_ATTRIBUTE_NAME = 'data-selection-touch-invoke';
```
dzearing, line 37: is there a test that could be added to make sure this keeps working

soettl: Have added a couple of unit tests for both the scenario where touch targets are disabled, and when they are enabled :)

packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
soettl: Have added a new prop `enableTouchInvocationTarget`

soettl:
Some advice on the naming of the prop for `SelectionZone` would be appreciated. Right now it is `enableTouchInvocationTarget`. Should it be plural, to denote the collection of invocation targets within the `SelectionZone`, i.e. `enableTouchInvocationTargets`? Or something else?

PR #13242, opened 2020-05-20
feat(screener): Enable IE11 tests

assuncaocharles:
#### Pull request checklist

#### Description of changes

This PR allows to create separated set of screener tests to run on IE11 by adding `.ie.steps` to the desired tests.

#### Focus areas to test

(optional)


scripts/screener/screener.iefiles.ts
```diff
@@ -0,0 +1,82 @@
...
+  'ListExampleHorizontal',
+  'MenuExampleVertical',
+  'MenuButtonExampleContextMenu',
+  'PopupCustomTargetExample',
+  'SplitButtonPositioningExampleShorthand',
+  'TableExampleNavigable',
+  'ToolbarExampleMenuWithSubmenu',
+  'TooltipExamplePointing',
+  'TreeInitiallyOpenExample',
+  'TreeMultiselectExample',
```
layershifter, line 79: Can we configure this under `.steps.ts` files instead?

assuncaocharles: Not all examples have `.steps` files and the way the screener config works we need a list of files to match, so the way I see it could work under the example files would be if we change files to be tested to something like `TreeMultiselectExample.shorthand.tsx` to `TreeMultiselectExample.shorthand.ie.tsx` and then have the Regex to match it. 

Or do you see other way?

scripts/screener/screener.iefiles.ts
```diff
@@ -0,0 +1,82 @@
+const defaultExamples = [
```
dzearing, line 1: I don't really know how the screener setup works in v0, and wasn't ever really consolidated with the v7 screener setup. That would be ideal at some point but just to focus here;

It seems odd to me to have component names here in a general scripts file, far away from the components. If we add another component nobody will remember to go back to scripts/screener to add it to this file.

We have the screener setup in a storybook projects in apps/vr-tests. It's pretty simple to have running and easy to know where to look for which tests will run. Not applicable to this PR specifically but I think it'd be great to just combine the projects, OR, keeps separate vr-tests app projects for multiple runs if we get parallel benefits but at least know where to look for what tests run.




We have a project

assuncaocharles: I agree that keeping the list far from the components itself is not ideal, what we were trying to achieve was to not run the IE11 for all components examples since it takes long time. So the idea would be to have a specific list. Another way to achieve the same that I can see is if for the examples that should be tested on ie we would add something in the file name like `PopupExample.ie.shorthand.tsx` this way we can match and include only these specific files to run on IE11 screener.

assuncaocharles:
@layershifter @dzearing The examples now run agains the `.steps` files that contains `browsers: ['ie11']`.

JustSlone:
@assuncaocharles, following up on this PR. 

It looks like the changes are approved, but some checks are failing. Do you have what you need to move forward?

assuncaocharles:
> @assuncaocharles, following up on this PR.
> 
> It looks like the changes are approved, but some checks are failing. Do you have what you need to move forward?

Hey @JustSlone 

I am in contact with ppl from sauce labs, the tests for screener are failing randomly for IE11 and there's no much clear understanding why. So I am applying changes they ask to try so they can debug and then try again.

PR #13247, opened 2020-05-20
Fixed more typos

tony-xia:


packages/experiments/src/components/TilesList/TilesList.tsx
```diff
@@ -543,7 +543,7 @@ export class TilesList<TItem> extends React.Component<ITilesListProps<TItem>, IT
 
     if (pageSpecificationCache.byIndex[startIndex]) {
       // If the page specification has already been calculated, return it.
-      // List recalculates all pages if any input changes, so this memoization
+      // List recalculates all pages if any input changes, so this memorization
```
ecraig12345, line 546: This was correct previously. https://en.wikipedia.org/wiki/Memoization
```suggestion
      // List recalculates all pages if any input changes, so this memoization
```

tony-xia:
thanks @ecraig12345 for your feedback. I updated them accordingly

ecraig12345:
Thanks for the updates but this is still missing a change file. Basically clone the repo and run `yarn` then `yarn change`, or see our [contributing docs](https://github.com/microsoft/fluentui/wiki/Contributing) for more details.

tony-xia:
Sorry, i might be wrong, i reverted unnecessary changes. there is only 2 typo fixes at the moment. https://github.com/microsoft/fluentui/pull/13247/files

ecraig12345:
The changes are correct now, but you need an additional file called a change file which is generated using the steps I described.

tony-xia:
Updated. Thanks @ecraig12345 for your feedback

PR #13250, opened 2020-05-20
fix(Chat): remove outline from focused items in Teams theme

notandrew:
This fixes an issue that apparently only exists in production, I can't repro in the Fluent docs site. Items with focus in Teams have orange outlines because it isn't removed in a default css location.

image


packages/fluentui/react-northstar/src/themes/teams/staticStyles/globalStyles.ts
```diff
@@ -7,6 +7,7 @@ const globalStyles: StaticStyleFunction = siteVars => ({
     fontFamily: siteVars.bodyFontFamily,
     fontSize: siteVars.bodyFontSize,
     lineHeight: siteVars.bodyLineHeight,
+    outline: 'none',
```
layershifter, line 10: I would like to fix this issue in the exact component instead of global styles.

notandrew: @layershifter - It's on almost every element. Normally I would agree, but this is probably more efficient and cleaner?

In Teamspace-Web, we have this:
image

notandrew: @layershifter  - what do you think?

layershifter: Can we at least move this style to `Chat` component itself? I.e. will affect `Chat` and everything inside instead of fully global space.

mnajdova: @notandrew can we update the PR based on the proposed comment from @layershifter ? I think it make sense that these "global" styles are moved at least to `chatStyles.ts`. Note that you can use these selectors there too:

```js
"& *": { outline: 'none'} 
```

PR #13253, opened 2020-05-20
chore(Header): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `Header` component to be functional. Restricting props set that will be passed to styles functions.

Related to #12237

## Prop sets

| `Header`    |
| --------- |
| `color` |
| `align` |
| `hasDescription` |

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Header/Header.tsx
```diff
@@ -38,56 +44,86 @@ export interface HeaderProps
...
+        }),
+      })}
+    >
+      {rtlTextContainer.createFor({ element: contentElement, condition: !!description })}
+      {!hasChildren &&
+        HeaderDescription.create(description, {
+          defaultProps: () =>
+            getA11yProps('description', {
+              variables: {
+                ...(variables?.descriptionColor && { color: variables.descriptionColor }),
```
layershifter, line 100: Oops, previously with has `resolvedVariables` so these things are not equal. I suggest to drop this statement and it as a breaking change.

assuncaocharles: Dunno what you mean, if we removed this statement our own example setting the `descriptionColor` wouldn't work. Do you mean to us to se a new prop `DescriptionColor?: string` to `Header`?

layershifter: https://github.com/microsoft/fluentui/blob/c3ed69100489a037d5e8def9296fe1055ea236c6/packages/fluentui/react-northstar/src/components/Header/Header.tsx#L62

1. In class component `variables` are coming from `renderComponent()`
2. `variables` from `renderComponent()` are `resolvedVariables`
  - are always `Object`
  - will include all `headerVariables`: https://github.com/microsoft/fluentui/blob/master/packages/fluentui/react-northstar/src/themes/teams/components/Header/headerVariables.ts
3. It means that `variables` from `renderComponent()` are not equal to `props.variables`

***

My suggestion is to remove this statement at all:

```suggestion
```

packages/fluentui/react-northstar/src/components/Header/Header.tsx
```diff
@@ -38,56 +44,85 @@ export interface HeaderProps
...
+      })}
+    >
+      {rtlTextContainer.createFor({ element: contentElement, condition: !!description })}
+      {!hasChildren &&
+        HeaderDescription.create(description, {
+          defaultProps: () =>
+            getA11yProps('description', {
+              variables: {
+              },
+            }),
```
layershifter, line 101:
```suggestion
            getA11yProps('description'),
```

packages/fluentui/react-northstar/src/components/Header/Header.tsx
layershifter: And please add this as a breaking change

packages/fluentui/react-northstar/src/components/Header/Header.tsx
```diff
@@ -38,56 +44,85 @@ export interface HeaderProps
...
+    Description: typeof HeaderDescription;
+  } = props => {
+  const context: ProviderContextPrepared = React.useContext(ThemeContext);
+  const { setStart, setEnd } = useTelemetry(Header.displayName, context.telemetry);
+  setStart();
+  const { children, content, variables, align, className, design, styles, description, color } = props;
+  const ElementType = getElementType(props);
+  const unhandledProps = useUnhandledProps(Header.handledProps, props);
+  const hasChildren = childrenExist(children);
+  const contentElement = childrenExist(children) ? children : content;
```
pompomon, line 60:
```suggestion
  const contentElement = hasChildren ? children : content;
```
_nit_

PR #13254, opened 2020-05-20
DatePicker: Copy DatePicker code from OUFR to react-next

MLoughry:
#### Pull request checklist

#### Description of changes

Copy `DatePicker` component code to `react-next` as basis for migration.

### DIFF
--- a/packages/react-next/src/components/DatePicker/examples/DatePicker.Bounded.Example.tsx
+++ b/packages/react-next/src/components/DatePicker/examples/DatePicker.Bounded.Example.tsx
@@ -1,0 +1,73 @@
+import * as React from 'react';
+import { DatePicker, DayOfWeek, IDatePickerStrings } from 'office-ui-fabric-react/lib/DatePicker';
+import { addMonths, addYears } from 'office-ui-fabric-react/lib/utilities/dateMath/DateMath';
+import { mergeStyleSets } from 'office-ui-fabric-react/lib/Styling';
+
+const today: Date = new Date(Date.now());
+const minDate: Date = addMonths(today, -1);
+const maxDate: Date = addYears(today, 1);
+const description = `When date boundaries are set (via minDate and maxDate props) the DatePicker will not allow
+out-of-bounds dates to be picked or entered. In this example, the allowed dates are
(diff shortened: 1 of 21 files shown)


MLoughry:
@dzearing @ecraig12345 As I mentioned, I'll be on mini-paternity leave for the next two weeks. If this is merged, the migration changes for `DatePicker` are in [this branch](https://github.com/MLoughry/office-ui-fabric-react/tree/functional/date-picker)

ecraig12345:
@lorejoh12 If we're converting the DatePicker to a function component, should we be starting from the OUFR one or the date-time one? How different are the APIs? (I think we're trying to avoid breaking API changes in the react-next components for now.)

lorejoh12:
> @lorejoh12 If we're converting the DatePicker to a function component, should we be starting from the OUFR one or the date-time one? How different are the APIs? (I think we're trying to avoid breaking API changes in the react-next components for now.)

I'd strongly suggest starting with the date-time one, since the OUFR one has bugs in it that have been fixed in the date-time one. That being said, there are very minor breaking API changes between them- specifically, a couple of deprecated props were removed from `ICalendarProps` type, which is the type of the `calendarProps` property.

Otherwise, I don't think there are large differences between the OUFR DatePicker and the date-time DatePicker- the biggest difference is which Calendar component it wraps.

MLoughry:
@ecraig12345 @lorejoh12 What's the plan, then? Should we migrate `@uifabric/date-picker` in place without a major version bump?

packages/react-next/src/components/DatePicker/examples/DatePicker.Bounded.Example.tsx
```diff
@@ -0,0 +1,73 @@
+import * as React from 'react';
+import { DatePicker, DayOfWeek, IDatePickerStrings } from 'office-ui-fabric-react/lib/DatePicker';
```
xugao, line 2: pls fix import to be `@fluentui/react-next` in examples

MLoughry:
What's the plan, then? Should I migrate @uifabric/date-time in-place (without a major version), or copy that one over to react-next?


-Michael

________________________________

PR #13260, opened 2020-05-21
Slider hook part 1.

joschect:
#### Pull request checklist

#### Description of changes

Part1 of moving state out of slider into a useSlider hook. I'm intentionally leaving the slide, thumb, slide value, for the next iteration to help reduce the amount that is happening in this change.
#ElectricSlide

### DIFF
--- a/packages/react-next/src/components/Slider/Slider.base.tsx
+++ b/packages/react-next/src/components/Slider/Slider.base.tsx
@@ -23,37 +23,10 @@ const useComponentRef = (props: ISliderProps, thumb: React.RefObject<HTMLSpanEle
     [value],
   );
 };
-const SliderLabel = (props: {
-  className?: string;
-  label?: string;
-  ariaLabel?: string;
-  disabled?: boolean;
-  id?: string;
-}) => {
-  const { className, label, ariaLabel, disabled, id } = props;
-
-  if (!label) {
-    return null;
-  }
-
-  return (
-    <Label className={className} {...(ariaLabel ? {} : { htmlFor: id })} disabled={disabled}>
-      {label}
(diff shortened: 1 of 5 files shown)


packages/react-next/src/components/Slider/Slider.base.tsx
```diff
@@ -28,274 +23,82 @@ const useComponentRef = (props: ISliderProps, thumb: React.RefObject<HTMLSpanEle
     [value],
   );
 };
+const SliderLabel = (props: {
+  className?: string;
+  label?: string;
+  ariaLabel?: string;
```
jurokapsiar, line 29: as discussed in https://hackmd.io/k2eJKI_JRs2Gew9dbrz00A?both, can you pls use dashed aria props? See https://reactjs.org/docs/dom-elements.html. camelCased can cause inconsistencies when spreading unhandled props to the root element.

packages/react-next/src/components/Slider/useSlider.ts
```diff
@@ -0,0 +1,264 @@
...
+    min = 0,
+    showValue = true,
+    buttonProps = {},
+    vertical = false,
+    valueFormat,
+    styles,
+    theme,
+  } = props;
+
+  const ariaLabel = props['aria-label'];
```
khmakoto, line 29: You can destructure this above by doing `'aria-label': ariaLabel.

joschect: Neat! I didn't know that.

joschect:
Any last minute comments? I'll complete the pull

joschect:
Rebuilding since it's been a bit

PR #13264, opened 2020-05-21
chore: switch to @fluentui/keyboard-key

layershifter:
#### Pull request checklist

#### Description of changes

Why we are doing this?
- `@fluentui/keyboard-key` is just a fork of `keyboard-key`
- key differences: bundle size and fact that Fluent's version is ESM

CJS module can't be treeshaken it means that we will have unused import for `keyboard-key` it will be still in bundle:

```tsx
import keyboardKey from 'keyboard-key'
```

image

```tsx
import { keyboardKey } from '@fluentui/keyboard-key'
// will keep only unsafe assignments
```

image


packages/fluentui/accessibility/src/behaviors/Card/cardsContainerBehavior.ts
```diff
@@ -15,7 +15,7 @@ const cardsContainerBehavior: Accessibility<CardsContainerBehaviorProps> = () =>
   attributes: {},
   focusZone: {
     props: {
-      shouldEnterInnerZone: event => keyboardKey.getCode(event) === keyboardKey.Enter,
+      shouldEnterInnerZone: event => getCode(event) === keyboardKey.Enter,
```
mnajdova, line 18:
```suggestion
      shouldEnterInnerZone: event => getCode(event) === EnterKey,
```

Let's keep it consistent. Is this exported?

layershifter: This is exported, but not things are available and exported as separate variables. So I don't know that is consistent in this case

mnajdova: Oh man, ok got it

PR #13272, opened 2020-05-21
Adding getNativeElementProps

dzearing:
Adding a method to `getNativeElementProps`, which requires the tagname, rather than the allow list, to resolve.

This will be used from within compose props resolution, so that components get unrecognized props mixed into the root element.

### DIFF
--- a/packages/utilities/src/properties.ts
+++ b/packages/utilities/src/properties.ts
@@ -1,4 +1,5 @@
 import { filteredAssign } from './object';
+import { ReactHTML } from 'react';
 
 /**
  * An array of events that are allowed on every html element type.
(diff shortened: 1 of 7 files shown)


packages/utilities/src/properties.ts
```diff
@@ -406,3 +407,43 @@ export function getNativeProps<T>(props: {}, allowedPropNames: string[], exclude
...
+ * element type.
+ * @param tagName - Tag name (e.g. "div")
+ * @param props - Props object
+ * @param excludedPropNames - List of props to disallow
+ */
+export function getNativeElementProps<TProps>(
+  tagName: keyof ReactHTML,
+  props: TProps,
+  excludedPropNames?: string[],
+): TProps {
```
xugao, line 445:
```suggestion
export function getNativeElementProps<TElement extends HTMLElement>(
  tagName: keyof ReactHTML,
  props: {},
  excludedPropNames?: string[],
): React.HTMLAttributes<TElement> {
```

packages/utilities/src/properties.ts
```diff
@@ -406,3 +407,43 @@ export function getNativeProps<T>(props: {}, allowedPropNames: string[], exclude
...
+  const allowedPropNames = (tagName && nativeElementMap[tagName.toLowerCase()]) || htmlElementProperties;
```
xugao, line 446:
```suggestion
  const allowedPropNames = (tagName && nativeElementMap[tagName]) || htmlElementProperties;
```

xugao: `.toLowerCase()` shouldn't be needed as we have typing to ensure

packages/utilities/src/properties.ts
dzearing: Hmmm. That doesn't seem right.

dzearing: maybe you're right. Let me try

xugao:
bundle size changes look unexpected

packages/utilities/src/properties.ts
dzearing: Updated.

packages/utilities/src/properties.ts
dzearing: removed

PR #13275, opened 2020-05-21
Add separate node template to create-package

ecraig12345:
Add a package template for things which run within Node (like build utilities) rather than in the browser. The new layout of the templates is like this (the things under `plop-templates` will always be used, plus the appropriate `plop-templates-<target>`):
```
scripts/
  create-package/
    plop-templates/       for both package types
    plop-templates-node/  for node packages
    plop-templates-react/ for react packages
```

Also fix a couple things which I noticed were wrong in the api-docs package, which should have no impact on output.


scripts/create-package/plop-templates-node/just.config.ts
```diff
@@ -0,0 +1,5 @@
+const { preset } = require('@uifabric/build');
+
+preset();
+
+task('build', 'build:node-lib').cached();
```
xugao, line 5: remove?

ecraig12345: No this is correct--for node libraries we want to only build commonjs and have it output to lib, which is what this does.

xugao: yeah i read it wrong :)

PR #13278, opened 2020-05-21
Grammar improvements for do's and don'ts documentation

czearing:
#### Pull request checklist

#### Description of changes

Grammar improvements for do's and don'ts list inside documentation.


packages/date-time/src/components/Calendar/docs/CalendarDos.md
```diff
@@ -1,5 +1,5 @@
-- Use the control as a single entity.
-- Set the default date to the current date unless a specific date is required for context (e.g. the date of the conference).
-- The control is designed to resize relative to available screen width. Allow it to render in either wide or narrow as appropriate.
-- When the control is engaged, the Calendar renders as a flyout and has defined widths (300px -narrow and 440px – wide). Plan your UI implementation accordingly.
-- The control renders date in a specific format. If allowing for manual entry of date, provide helper text in the appropriate format.
+- Do use the control as a single entity.
+- Do set the default date to the current date unless a specific date is required for context (e.g. the date of the conference).
+- Do allow Calendar to render in either wide or narrow. The control is designed to resize relative to available screen width.
+- Do when the control is engaged, the Calendar renders as a flyout and has defined widths (300px -narrow and 440px – wide). Plan your UI implementation accordingly.
```
micahgodbolt, line 4: this one doesn't work with 'do' in front of it.

packages/date-time/src/components/Calendar/docs/CalendarDos.md
```diff
@@ -1,5 +1,5 @@
-- Use the control as a single entity.
-- Set the default date to the current date unless a specific date is required for context (e.g. the date of the conference).
-- The control is designed to resize relative to available screen width. Allow it to render in either wide or narrow as appropriate.
-- When the control is engaged, the Calendar renders as a flyout and has defined widths (300px -narrow and 440px – wide). Plan your UI implementation accordingly.
-- The control renders date in a specific format. If allowing for manual entry of date, provide helper text in the appropriate format.
+- Do use the control as a single entity.
```
ecraig12345, line 1: I'm not sure we ought to be making a systematic change to the wording like this without consulting our writers or at least the website designers. I suspect they'd be against prefixing everything with "Do" in particular because it adds verbosity and 

Also, all of the dos/don'ts will be going away at some point in the future per the content team, though I don't remember the details of the reasoning (@micahgodbolt do you remember?).

micahgodbolt: Yes, dos/don'ts are going away. The question is what next steps do we take?

1. don't do anything
2. try to improve them on our own (this PR)
3. work with content team to improve what we have
4. work with content team to roll out their updated content

ecraig12345: I'm okay with trying to improve them on our own, just hesitant to make such a dramatic all-up stylistic change (prepending do/don't everywhere) without consulting anyone.

Vitalius1: I remember when writing the dos and donts for Shimmer and HoverCard someone in design team told me there is no need to add the Do and Don't into each bullet point if we already sort them in two columns with DO and DONT headers and add to that the icons on each statement has a good visual indicator if this is a good or bad thing. Adding this all over it's just redundant.

dzearing: I strongly recommend adding dos and don'ts prefixes, even if it's verbose. When you read them, it's super unclear what's being said without doing mental math that it's being said in the context of a Do or a Don't.

ecraig12345: The fact that we as devs have strong opposing opinions about this reinforces what I said about getting a writer's input before making such a dramatic change.

packages/date-time/src/components/Calendar/docs/CalendarDonts.md
```diff
@@ -1,4 +1,4 @@
 - Don't attempt to break apart year from month/day selectors. If granularity is required, use the Dropdown control or something similar.
 - Don't attempt to force resize the control in any way.
 - Don't force the control to render one mode vs. the other (year or month/day)
-- The flyout selector is a light dismiss control. Don't modify this behavior in any way.
+- Don't modify flyout selector's behavior in any way. The flyout selector is a light dismiss control.
```
dzearing, line 4: I have no idea what it's saying. @ecraig12345 ?

recommend removing.

packages/date-time/src/components/DatePicker/docs/DatePickerDonts.md
```diff
@@ -1,4 +1,4 @@
 - Don't attempt to break apart year from month/day selectors. If granularity is required, use the Dropdown control or something similar.
 - Don't attempt to force resize the control in any way.
 - Don't force the control to render one mode vs. the other (year or month/day)
-- The flyout selector is a light dismiss control. Don't modify this behavior in any way.
+- Don't modify flyout selector's behavior in any way. The flyout selector is a light dismiss control.
```
dzearing, line 4: Same as Calendar; we should consider removing this

packages/experiments/src/components/Button/docs/ButtonDos.md
```diff
@@ -1,11 +1,9 @@
...
-- Buttons should always include a noun if there is any room for interpretation about what the verb operates on.
+- Use Button labels that have a clear purpose. Labels should describe the action of the button in a single line and include a concise verb. If their is room for interpretation about what the verb operates on, a noun should also be used.
 - Consider the effect localization will have on the button and what will happen to components around it.
-- If the button’s label content is dynamic, consider how the button will resize and what will happen to components around it.
+- Consider how the button will resize if the button’s label content is dynamic.
 - Use only a single line of text in the label of the button.
-- Expose only one or two buttons to the user at a time, for example, "Accept" and "Cancel". If you need to expose more actions to the user, consider using checkboxes or radio buttons from which the user can select actions, with a single command button to trigger those actions.
+- Expose only one or two buttons to the user at a time. If you need to expose more actions to the user, consider using checkboxes or radio buttons.
```
dzearing, line 5: Replace buttons with checkboxes or radio buttons?

Remove...

packages/date-time/src/components/Calendar/docs/CalendarDonts.md
ecraig12345: Light dismiss means you can click outside it to dismiss, as opposed to only dismissable by clicking the X (or some other close button)

packages/experiments/src/components/Button/docs/ButtonDos.md
```diff
@@ -1,11 +1,9 @@
...
 - Show only one primary button that inherits theme color at rest state. In the event there are more than two buttons with equal priority, all buttons should have neutral backgrounds.
-- "Submit", "OK", and "Apply" buttons should always be styled as primary buttons. When "Reset" or "Cancel" buttons appear alongside one of the above, they should be styled as secondary buttons.
-- Default buttons should always perform safe operations. For example, a default button should never delete.
-- Use task buttons to cause actions that complete a task or cause a transitional task. Do not use buttons to toggle other UI in the same context. For example, a button may be used to open an interface area but should not be used to open an additional set of components in the same interface.
+- Style "Submit", "OK", and "Apply" buttons as primary buttons. Style "Reset" or "Cancel" buttons that appear alongside one of the above, as secondary buttons.
```
dzearing, line 7: Remove this. Adds no valuable information,

packages/date-time/src/components/Calendar/docs/CalendarDonts.md
ecraig12345: It seems to be a pretty standard term in our docs, but the meaning isn't necessarily obvious if you're not familiar with it...

packages/experiments/src/components/Button/docs/ButtonDos.md
ecraig12345: It made more sense the way it was before, but still seems a bit odd.

packages/experiments/src/components/Button/docs/ButtonDos.md
ecraig12345: Not so sure about that...maybe it's not valuable to someone who's been doing this awhile, but it could be important for someone newer or who's not actually a designer. (Though the wording should be updated to say "default buttons" instead of "secondary buttons.")

packages/date-time/src/components/Calendar/docs/CalendarDos.md
dzearing: Let me clarify; 

strongly recommend the "don't" prefixes.

I'm fine limiting the scope of the change.

packages/date-time/src/components/Calendar/docs/CalendarDonts.md
dzearing: What does modify behavior mean exactly though? I don't know how to visualize this.

packages/date-time/src/components/Calendar/docs/CalendarDos.md
Vitalius1: +1 to what @ecraig12345 said. I only put in there what a designer gave me. I don't argue that the example with Shimmer you brought is very confusing but it's a bit less confusing when you look at it from with context
image

User will get the message right away when he visually scans and sees the header and the icon.

In general I don't argue that it needs to be changed but a consulting with design + content writer is a must in my opinion when doing it like this in bulk.

packages/experiments/src/components/Button/docs/ButtonDos.md
dzearing: I don't think replacing a button with a radio button or a checkbox makes sense. They do not execute actions. They allow the user to select a state.

packages/experiments/src/components/Button/docs/ButtonDos.md
dzearing: We don't want the ask the user to style anything. That would be the wrong guidance.

We could say: Use PrimaryButtons for the primary action (e.g. the "OK" or "Apply" button) and DefaultButtons for all others.

But it already says something similar in another rule (use one primary button)

packages/date-time/src/components/Calendar/docs/CalendarDos.md
```diff
@@ -1,5 +1,5 @@
 - Use the control as a single entity.
 - Set the default date to the current date unless a specific date is required for context (e.g. the date of the conference).
-- The control is designed to resize relative to available screen width. Allow it to render in either wide or narrow as appropriate.
+- Allow Calendar to render in either wide or narrow. The control is designed to resize relative to available screen width.
```
dzearing, line 3:
```suggestion
```

packages/date-time/src/components/Calendar/docs/CalendarDos.md
```diff
@@ -1,5 +1,4 @@
 - Use the control as a single entity.
 - Set the default date to the current date unless a specific date is required for context (e.g. the date of the conference).
-- The control is designed to resize relative to available screen width. Allow it to render in either wide or narrow as appropriate.
 - When the control is engaged, the Calendar renders as a flyout and has defined widths (300px -narrow and 440px – wide). Plan your UI implementation accordingly.
-- The control renders date in a specific format. If allowing for manual entry of date, provide helper text in the appropriate format.
+- If allowing for manual entry of date, provide helper text in the appropriate format. The control renders date in a specific format.
```
dzearing, line 4:
```suggestion
- If allowing for manual entry of date, provide helper text in the appropriate format. The control renders the date in a specific format.
```

packages/experiments/src/components/Button/docs/ButtonDos.md
```diff
@@ -1,11 +1,9 @@
...
+- Ensure that Default buttons always perform safe operations. For example, a default button should never delete.
```
dzearing, line 8:
```suggestion
- Ensure that DefaultButtons always perform safe operations. For example, a default button should never delete.
```

packages/date-time/src/components/Calendar/docs/CalendarDos.md
dzearing: Blind users will be read those out-loud. Just another consideration.

JustSlone:
@czearing checking in on this PR, there was some feedback and discussion, then some changes. 

Is it ready for review again?

packages/date-time/src/components/Calendar/docs/CalendarDos.md
dzearing: The do's were removed. I think leaving the dont's makes it more clear when listening via screen reader and create some important distinction to keep things clear.

dzearing:
Yes this is ready for re-review.

micahgodbolt:
I'm talking with Jennifer about creating a branch to replace dos/don'ts with new 'overview' sections. Do we still want to try to get this change through, or just close it?

dzearing:
@micahgodbolt I'd love to get this in - there are so many busted things on there right now and overall this improves the obvious. We can always do a second pass and refine, but unless there is going to be a followup pr in a specific amount of time, I think we should merge this.

micahgodbolt:
@jennbost and I are working on this right now: https://github.com/microsoft/fluentui/tree/docs-update

The goal is to remove the dos and don'ts and replace them with layout/content specific instructions. I don't mind this code going in, even if for a short time, as long as it doesn't cause an issue for her work.

JustSlone:
@micahgodbolt @czearing It looks like this will not cause conflicts with the docs-update branch (since that branch has not deleted the do's and don'ts yet), unless that work is going to be complete next week let's take this change. We should avoid getting stuck here and letting perfect block progress. 

Looking for approvals from @dzearing @micahgodbolt and @ecraig12345 since you all have been engaged here.

PR #12681, opened 2020-04-14
Add EventAnnotation feature for Linechart

SamuelQZQ:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [X] Include a change request file using `$ yarn change`

#### Description of changes

Add a `eventAnnotationProps` parameter for line chart. With this, we can add some vertical lines with annotations to show events on specific days. And the annotation can be clicked, showing a callout with customized items.

If a few events are close to each other, or at the same date, they would be merged, like the events on 03/04 below.
image

When clicking on the label, show a customized callout.
image

Three React components added:
- `EventAnnotation`: including the **vertical lines** and the **`LabelLink`**.
- `LabelLink`: the **`TextBox`** and the **callout**
- `TextBox`:  the label

#### Focus areas to test

(optional)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/12681)

PR #13295, opened 2020-05-23
Remove implicit dep on lodash in charting package

### LINKED ISSUE #13294: lodash is referenced in charting package but isn't a dep
# Environment Information

Pull latest charting package, ensure your package doesn't have a lodash dep in it, and try to build. Build will barf on charting because there are references to it but no dependency listed.

Can work around by adding a dev dep in the consuming application but that's not a real fix.

#### Actual behavior:
Build breaks without consuming app adding dev dep on lodash

#### Expected behavior:
Build works without consumer including lodash if they don't actually require it directly.

Are you willing to submit a PR to fix? I can if required

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13294

#### Description of changes

#12681 added references to `lodash` in some files in the charting package. Charting doesn't have a dep on `lodash` (and we don't want to add one), so the references to lodash need to be removed. (@SamuelQZQ FYI for future contributions.)
- Replace `omit` with object spread
- Replace `findIndex` with one from `@uifabric/utilities` (and add a `fromIndex` arg to the utility)
- Replace `range` with inline implementation
- Replace `unionBy` with local implementation of `uniqBy` (`unionBy` does additional array flattening which is unnecessary)

Also fix some imports within the charting package which referenced the package by name (instead of using relative paths) for some reason.

We'll be adding an eslint rule soon which would have caught both of these issues.

### DIFF
--- a/change/@uifabric-charting-2020-05-22-22-12-06-charting-no-lodash.json
+++ b/change/@uifabric-charting-2020-05-22-22-12-06-charting-no-lodash.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Remove implicit dep on lodash and fix other imports",
+  "packageName": "@uifabric/charting",
+  "email": "elcraig@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-05-23T05:12:06.167Z"
+}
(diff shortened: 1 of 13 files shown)


dzearing:
We should add the charting package to bundle auditing. Bringing in a few lodash things brings a lot of things in.

PR #13297, opened 2020-05-23
chore(Loader): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `Loader` component to be functional. Restricting props set that will be passed to styles functions.

Related to #12237

## Prop sets

| `Loader`    |
| --------- |
| `inline` |
| `labelPosition` |
| `size` |

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Loader/Loader.tsx
+++ b/packages/fluentui/react-northstar/src/components/Loader/Loader.tsx
@@ -155,7 +155,7 @@ const Loader: React.FC<WithAsProp<LoaderProps>> &
           getA11yProps('label', {
             className: loaderSlotClassNames.label,
             styles: resolvedStyles.label,
-            id: labelId,
+            id: labelId.current,
           }),
       })}
     </ElementType>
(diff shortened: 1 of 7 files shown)


packages/fluentui/react-northstar/src/components/Loader/Loader.tsx
```diff
@@ -60,102 +73,126 @@ export const loaderSlotClassNames: LoaderSlotClassNames = {
...
+  } = props => {
+  const context: ProviderContextPrepared = React.useContext(ThemeContext);
+  const { setStart, setEnd } = useTelemetry(Loader.displayName, context.telemetry);
+  setStart();
+  const { delay, label, indicator, svg, inline, labelPosition, className, design, styles, variables, size } = props;
+
+  const ElementType = getElementType(props);
+  const unhandledProps = useUnhandledProps(Loader.handledProps, props);
+
+  let delayTimer: number;
```
layershifter, line 93:
```suggestion
  const delayTimer = React.useRef<number>();
```

layershifter:
Please fix PR description, props there doesn't match to `mapPropsToStyles`.

packages/fluentui/react-northstar/src/components/Loader/Loader.tsx
assuncaocharles: Done

layershifter:
> Please fix PR description, props there doesn't match to `mapPropsToStyles`.

Now it's empty, but there should be `inline`, `labelPosition` and `size`.
Please include a changelog entry.

PR #13301, opened 2020-05-23
chore(RadioGroupItem): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `RadioGroupItem` component to be functional. Restricting props set that will be passed to styles functions.

Related to #12237

## Prop sets

| `RadioGroupItem`    |
| --------- |
| `vertical` |
| `disabled` |
| `checked` |

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/RadioGroup/RadioGroupItem.tsx
+++ b/packages/fluentui/react-northstar/src/components/RadioGroup/RadioGroupItem.tsx
@@ -91,7 +91,7 @@ const RadioGroupItem: React.FC<WithAsProp<RadioGroupItemProps>> &
   const context: ProviderContextPrepared = React.useContext(ThemeContext);
   const { setStart, setEnd } = useTelemetry(RadioGroupItem.displayName, context.telemetry);
   setStart();
-  const { label, indicator, disabled, vertical, className, design, styles, variables } = props;
+  const { label, indicator, disabled, vertical, className, design, styles, variables, shouldFocus } = props;
   const elementRef = React.useRef<HTMLElement>();
   const ElementType = getElementType(props);
   const unhandledProps = useUnhandledProps(RadioGroupItem.handledProps, props);
@@ -108,9 +108,9 @@ const RadioGroupItem: React.FC<WithAsProp<RadioGroupItemProps>> &
   };
 
   React.useEffect(() => {
-    checked && props.shouldFocus && elementRef.current.focus();
+    checked && shouldFocus && elementRef.current.focus();
     _.invoke(props, 'onChange', undefined, { ...props, checked });
-  }, [checked]);
+  }, [checked, shouldFocus]);
 
   const { classes, styles: resolvedStyles } = useStyles<RadioGroupItemStylesProps>(RadioGroupItem.displayName, {
     className: radioGroupItemClassName,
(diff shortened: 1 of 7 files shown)


packages/fluentui/react-northstar/src/components/RadioGroup/RadioGroupItem.tsx
```diff
@@ -76,93 +84,124 @@ export const radioGroupItemSlotClassNames: RadioGroupItemSlotClassNames = {
...
 
-  static autoControlledProps = ['checked'];
-
-  actionHandlers = {
-    performClick: e => {
-      e.preventDefault();
-      this.handleClick(e);
+  React.useEffect(() => {
+    checked && props.shouldFocus && elementRef.current.focus();
+    _.invoke(props, 'onChange', undefined, { ...props, checked });
```
layershifter, line 112: I would rather move this to `handleClick` as this looks pretty strange and unexpected

packages/fluentui/react-northstar/src/components/RadioGroup/RadioGroupItem.tsx
```diff
@@ -76,93 +84,124 @@ export const radioGroupItemSlotClassNames: RadioGroupItemSlotClassNames = {
...
   };
 
-  static autoControlledProps = ['checked'];
-
-  actionHandlers = {
-    performClick: e => {
-      e.preventDefault();
-      this.handleClick(e);
+  React.useEffect(() => {
+    checked && props.shouldFocus && elementRef.current.focus();
```
layershifter, line 111:
```suggestion
    if(checked && shouldFocus) elementRef.current.focus();
```

`shouldFocus` is also a dependency for the hook

assuncaocharles: done

packages/fluentui/react-northstar/src/components/RadioGroup/RadioGroupItem.tsx
assuncaocharles: It was my first thought as well, but then it doesn't work... The focus never change.

layershifter: Can you please clarify what you have tried there? What change you applied?

assuncaocharles: Tried to move the condition and `.focus()` inside the click to either the event target and to the element ref. Neither applied the focus change when navigating the RadioGroupItem

assuncaocharles: This is what we get when trying to move the `focus()` to inside `handleClick`

Screenshot 2020-05-25 at 11 00 24

layershifter: Have you tried these changes?

```diff
  const handleClick = (e: React.MouseEvent | React.KeyboardEvent) => {
    _.invoke(props, 'onClick', e, props);
-    setChecked(prevChecked => !prevChecked);
+   setChecked(prevChecked => {
+        _.invoke(props, 'onChange', undefined, { ...props, checked: !prevChecked })
+        return !prevChecked
+      });
//
    if (checked && shouldFocus) elementRef.current.focus();
-   _.invoke(props, 'onChange', undefined, { ...props, checked });
```

assuncaocharles: I thought you meant to move `if (checked && shouldFocus) elementRef.current.focus();` inside the click.

assuncaocharles: This diff works.

PR #13310, opened 2020-05-25
chore: use Stylus v4

layershifter:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


layershifter:
There is no sense to bump Stylis right now as there is no perf improvements.

PR #13111, opened 2020-05-12
docs(Toolbar): add SplitButton in the custom content example

silviuaavram:
#### Description of changes

Add a SplitButton example component to the Toolbar example with custom items.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui/pull/13111)

PR #13311, opened 2020-05-25
fix(SplitButton): Stop propagation of left and right arrow keys for Toolbar integration scenario

jurokapsiar:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


jurokapsiar:
based on PR #13111

packages/fluentui/docs/src/examples/components/Toolbar/Content/ToolbarExampleCustomContent.shorthand.tsx
```diff
@@ -1,7 +1,42 @@
 import * as React from 'react';
-import { Button, Text, Toolbar } from '@fluentui/react-northstar';
+import { Button, Text, Toolbar, SplitButton } from '@fluentui/react-northstar';
 import { BoldIcon } from '@fluentui/react-icons-northstar';
 
+const ToolbarSplitButton = () => (
+  <>
+    <SplitButton
+      menu={[
```
pompomon, line 8: When navigation using keyboard and selecting menu item with Enter, the focus gets lost, do we want to handle it in this example?

jurokapsiar: good point. it is expected, I will double check if we have it in the best practices for the split button. no need to put it in the example I think

PR #13315, opened 2020-05-26
docs(Card): Add best practices

pompomon:
#### Pull request checklist

#### Description of changes

Adding `Do`s and `Don't`s to Card

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Card/BestPractices/CardBestPractices.tsx
+++ b/packages/fluentui/docs/src/examples/components/Card/BestPractices/CardBestPractices.tsx
@@ -4,16 +4,15 @@ import ComponentBestPractices from '../../../../components/ComponentBestPractice
 
 const doList = [
   'Limit use of `aria-roledescription` property to clarifying the purpose of card, or to providing a more specific description of a widget.',
-  'Use `cardBehavior` for disabled cards, so it would not be focusable.',
   'Use `cardsContainerBehavior` for cards container, so it would be navigable using arrows.',
-  'Use `cardChildrenFocusableBehavior` for focusable cards with inteactive children.',
+  'Use `cardChildrenFocusableBehavior` for focusable cards with interactive children.',
   'Use `cardFocusableBehavior` for focusable cards without inteactive children.',
   'Use `cardSelectableBehavior` for selectable cards without inteactive children.',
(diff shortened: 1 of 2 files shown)


packages/fluentui/docs/src/examples/components/Card/BestPractices/CardBestPractices.tsx
```diff
@@ -5,9 +5,16 @@ import ComponentBestPractices from '../../../../components/ComponentBestPractice
 const doList = [
   'Limit use of `aria-roledescription` property to clarifying the purpose of card, or to providing a more specific description of a widget.',
   'Use `cardBehavior` for disabled cards, so it would not be focusable.',
```
jurokapsiar, line 7: isn't `cardBehavior` default? why would the consumer need to change the behavior to make disabled work correctly?

packages/fluentui/docs/src/examples/components/Card/BestPractices/CardBestPractices.tsx
```diff
@@ -5,9 +5,16 @@ import ComponentBestPractices from '../../../../components/ComponentBestPractice
...
+  'Use `cardsContainerBehavior` for cards container, so it would be navigable using arrows.',
+  'Use `cardChildrenFocusableBehavior` for focusable cards with inteactive children.',
+  'Use `cardFocusableBehavior` for focusable cards without inteactive children.',
+  'Use `cardSelectableBehavior` for selectable cards without inteactive children.',
 ];
 
-const dontList = ['Use `onClick` property together with clickable children.'];
+const dontList = [
+  'Use `onClick` property together with clickable children.',
+  'Create selectable cards with a single text child that has `id`, that might cause incorrect narration in JAWS with virtual cursor navigation',
```
jurokapsiar, line 16: nit: two sentences might be better: 'Create selectable cards with a single text child that has `id`. It might cause incorrect narration in JAWS with virtual cursor navigation'

packages/fluentui/docs/src/examples/components/Card/BestPractices/CardBestPractices.tsx
pompomon: Oh, yes, that makes sense

PR #13319, opened 2020-05-26
feat(ToolbarCustomItem): use compose()

layershifter:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Toolbar/ToolbarCustomItem.tsx
```diff
@@ -26,7 +26,7 @@ import { ToolbarVariablesContext } from './toolbarVariablesContext';
 
 export interface ToolbarCustomItemProps extends UIComponentProps, ChildrenComponentProps, ContentComponentProps {
   /**
-   * Accessibility behavior if overridden by the user.
+   * 1111 Accessibility behavior if overridden by the user.
```
mnajdova, line 29:
```suggestion
   * Accessibility behavior if overridden by the user.
```

PR #13326, opened 2020-05-26
Re-exporting react-button in react-next

khmakoto:
#### Pull request checklist

#### Description of changes

This PR makes the `Button` in `@fluentui/react-next` export the version in `@fluentui/button` while making the changes inside the components in `@fluentui/react-next` to point to the `office-ui-fabric-react` version of the `Button` so they can continue working.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-next/etc/react-next.api.md
+++ b/packages/react-next/etc/react-next.api.md
@@ -1112,11 +1112,11 @@ export const Toggle: React.FunctionComponent<IToggleProps>;
 export const ToggleBase: React.FunctionComponent;
 
 
+export * from "@fluentui/react-button";
 export * from "office-ui-fabric-react/lib/ActivityItem";
 export * from "office-ui-fabric-react/lib/Announced";
 export * from "office-ui-fabric-react/lib/Autofill";
 export * from "office-ui-fabric-react/lib/Breadcrumb";
-export * from "office-ui-fabric-react/lib/Button";
 export * from "office-ui-fabric-react/lib/Calendar";
 export * from "office-ui-fabric-react/lib/Callout";
 export * from "office-ui-fabric-react/lib/Check";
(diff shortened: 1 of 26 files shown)


packages/react-next/etc/react-next.api.md
```diff
@@ -1112,11 +1112,11 @@ export const Toggle: React.FunctionComponent<IToggleProps>;
 export const ToggleBase: React.FunctionComponent;
 
 
+export * from "@fluentui/react-button";
 export * from "office-ui-fabric-react/lib/ActivityItem";
 export * from "office-ui-fabric-react/lib/Announced";
 export * from "office-ui-fabric-react/lib/Autofill";
 export * from "office-ui-fabric-react/lib/Breadcrumb";
-export * from "office-ui-fabric-react/lib/Button";
```
xugao, line 1119: we should still export these since it contains a lot more:

```
export * from './BaseButton';
export * from './Button.types';
export * from './Button';
export * from './ActionButton/ActionButton';
export * from './CommandBarButton/CommandBarButton';
export * from './CommandButton/CommandButton';
export * from './CompoundButton/CompoundButton';
export * from './DefaultButton/DefaultButton';
export * from './MessageBarButton/MessageBarButton';
export * from './PrimaryButton/PrimaryButton';
export * from './IconButton/IconButton';
```

khmakoto: I think we should still export them but not from next, the idea we talked about a legacy package seems like the correct place for these to go.

khmakoto:
Closing until we have a clearer path forward regarding `Button`.

dzearing:
Aug

packages/react-next/etc/react-next.api.md
dzearing: in /compat.ts probably.

PR #13327, opened 2020-05-26
chore(ItemLayout): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

Converting `ItemLayout` from class component to functional. Any props are not longer passed to styles function. Removed `state` parameter from `renderContentArea`, `renderHeaderArea` and `renderMainArea`

Related to #12237

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/ItemLayout/ItemLayout.tsx
```diff
@@ -34,9 +34,9 @@ export interface ItemLayoutProps extends UIComponentProps, ContentComponentProps
   endMedia?: any;
   headerMedia?: any;
   media?: any;
-  renderContentArea?: (props: ItemLayoutProps, state: any, classes: ComponentSlotClasses) => React.ReactNode;
```
layershifter, line 37: This looks like a breaking change, please mention this.

layershifter:
> Any props are not longer passed to styles function.

It's not true, at least `debugLayout`.

assuncaocharles:
> > Any props are not longer passed to styles function.
> 
> It's not true, at least `debugLayout`.

But it doesn't come from props. For what I understand isn't for development purposes?

layershifter:
> > > Any props are not longer passed to styles function.
> > 
> > 
> > It's not true, at least `debugLayout`.
> 
> But it doesn't come from props. For what I understand isn't for development purposes?

OMG, I see 🦀

assuncaocharles:
> > > > Any props are not longer passed to styles function.
> > > 
> > > 
> > > It's not true, at least `debugLayout`.
> > 
> > 
> > But it doesn't come from props. For what I understand isn't for development purposes?
> 
> OMG, I see 🦀

I can even remove it at all if you think it would be better.

packages/fluentui/react-northstar/src/components/ItemLayout/ItemLayout.tsx
assuncaocharles: It wasn't used at all anyway before, should I still mention?

assuncaocharles: Mentioned

packages/fluentui/react-northstar/src/components/ItemLayout/ItemLayout.tsx
```diff
@@ -64,151 +64,183 @@ export const itemLayoutSlotClassNames: ItemLayoutSlotClassNames = {
...
+  headerCSS: PropTypes.object,
+  headerMediaCSS: PropTypes.object,
+  contentCSS: PropTypes.object,
+  contentMediaCSS: PropTypes.object,
+  endMediaCSS: PropTypes.object,
+};
 
-  static deprecated_className = itemLayoutClassName;
+ItemLayout.defaultProps = {
+  as: 'div',
```
layershifter, line 171:
```suggestion
```

layershifter:
> I can even remove it at all if you think it would be better.

If it's impossible to use, let's just remove these styles and props 👍

PR #13329, opened 2020-05-26
Add storybook addon to switch between different theme

xugao:
#### Pull request checklist

#### Description of changes
Adding this to `react-next` package:
image

image

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-next-2020-05-26-16-20-54-xgao-theme-swtich.json
+++ b/change/@fluentui-react-next-2020-05-26-16-20-54-xgao-theme-swtich.json
@@ -1,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "Storybook: add theme swtiching knob",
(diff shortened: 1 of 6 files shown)


change/@fluentui-react-next-2020-05-26-16-20-54-xgao-theme-swtich.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "Storybook: add theme switching knob",
+  "packageName": "@fluentui/react-next",
```
dzearing, line 4: What about react-button?

xugao: we can certainly do the same for `react-button`. but what theme would you load?

PR #13330, opened 2020-05-26
Adding createClassResolver helper to react-compose

dzearing:
Separating a PR out for the class resolver utility. Expected usage would be:

```tsx
import * as classes from './some.module.css';

compose(Thing, {
  classes: createClassResolver(classes)
});
```

### DIFF
--- a/packages/react-compose/src/createClassResolver.test.ts
+++ b/packages/react-compose/src/createClassResolver.test.ts
@@ -1,0 +1,47 @@
+import { createClassResolver } from './createClassResolver';
+
+describe('createClassResolver', () => {
+  const classResolver = createClassResolver({
+    root: 'root',
+    slot1: 'slot1',
+    slot2: 'slot2',
+    primary: 'primary',
+    size_small: 'small',
+    size_medium: 'medium',
+  });
+  const slots = { root: null, slot1: null, slot2: null };
+
+  it('can resolve slot classes', () => {
+    expect(classResolver({}, slots)).toEqual({
+      root: 'root',
+      slot1: 'slot1',
+      slot2: 'slot2',
+    });
+  });
+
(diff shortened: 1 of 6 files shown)


packages/react-compose/src/createClassResolver.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  it('can resolve enums', () => {
+    // Can resolve
+    expect(classResolver({ size: 'small' }, slots)).toEqual({
+      root: 'root small',
+      slot1: 'slot1',
+      slot2: 'slot2',
+    });
+  });
+
+  it('can resolve mixed content, including a className in props', () => {
```
xugao, line 39: lets add test for when `state` has properties that are not mapped to classes:
```
classResolver({ className: 'foo', primary: true, size: 'medium', bar: 'bar' }, slots)
```

packages/react-compose/src/createClassResolver.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    expect(classResolver({}, slots)).toEqual({
+      root: 'root',
+      slot1: 'slot1',
+      slot2: 'slot2',
+    });
+  });
+
+  it('can resolve modifiers', () => {
+    expect(classResolver({ primary: true }, slots)).toEqual({
+      root: 'root primary',
```
xugao, line 24: with our scss conversions, we are actually adding modifier classes onto every slot. that way the scss can be written in a cleaner way.

dzearing: changed.

packages/react-compose/src/createClassResolver.test.ts
dzearing: updated.

PR #13333, opened 2020-05-27
Convert ShimmeredDetailsList to use hooks (and add useShallowMemo hook)

ThomasMichon:
Converted `ShimmeredDetailsList` to use hooks, since it does not have any advanced state to manage. This allows it to memoize its injections into `listProps`, as well as debounce rendered to the real `DetailsList` component thanks to use of a new `useShallowMemo` hook.

The new `useShallowMemo` hook is just like `useMemo` except its `deps` value is a single object of any set of fields. If the new value of `deps` fails a shallow-equal comparison to the old, then the callback is re-evaluated and the new result returned. This callback can be used in places where `useMemo` cannot, for example when comparing 'the rest of a props object' within a decorator-style component.

### DIFF
--- a/packages/office-ui-fabric-react/src/components/DetailsList/ShimmeredDetailsList.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/ShimmeredDetailsList.base.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 
 import { classNamesFunction, css } from '../../Utilities';
-import { IProcessedStyleSet } from '../../Styling';
 import { SelectionMode } from '../../utilities/selection/interfaces';
 import { DetailsList } from './DetailsList';
 import { IDetailsRowProps } from './DetailsRow.types';
@@ -12,6 +11,7 @@ import {
   IShimmeredDetailsListStyles,
 } from './ShimmeredDetailsList.types';
 import { CheckboxVisibility } from './DetailsList.types';
+import { useShallowMemo } from '@uifabric/react-hooks';
 
 import { DEFAULT_CELL_STYLE_PROPS, DEFAULT_ROW_HEIGHTS } from './DetailsRow.styles';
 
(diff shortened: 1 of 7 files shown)


packages/react-hooks/src/useShallowMemo.ts
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import { shallowCompare } from '@uifabric/utilities';
+
+/**
+ * Hook which evaluates the given expression only if the given deps object is shallow-different than
+ * for the previous invocation of the hook.
+ * This is essentially `React.useMemo`, except it shallow-compares a single deps object
+ * rather than an array of deps values.
+ * @public
```
ecraig12345, line 9: Remove this

packages/react-hooks/src/useShallowMemo.ts
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import { shallowCompare } from '@uifabric/utilities';
```
ecraig12345, line 2: For bundle size reasons it might be better to import directly from the file where shallowCompare is defined.

packages/office-ui-fabric-react/src/components/DetailsList/ShimmeredDetailsList.base.tsx
```diff
@@ -21,141 +21,145 @@ const SHIMMER_INITIAL_ITEMS = 10;
...
-    } = this.props;
-
-    const listClassName = listProps && listProps.className;
-
-    this._classNames = getClassNames(styles, {
-      theme: theme!,
-    });
-
-    const newListProps = {
+export function ShimmeredDetailsListBase(props: IShimmeredDetailsListProps): JSX.Element | null {
```
ecraig12345, line 24: Typically we've been doing `export const ShimmeredDetailsListBase: React.FunctionComponent<IShimmeredDetailsListProps> = props => {`

ecraig12345:
@dzearing or @xugao I forget, are we considering in-place function component conversions to be acceptable or do they have to be done in react-next? (since in edge cases, mainly class extension, it can be a breaking change)

PR #13346, opened 2020-05-27
feat(telemetry): add cache hits ratio

layershifter:
#### Pull request checklist

#### Description of changes

This PR improves `telemetry` object an provides more data: cache hits and differentiates renders from components' instances.

Also exposes `getFluentTelemetry()` to get telemetry data in Fluent doc site:

image
 

#### Focus areas to test

(optional)


packages/fluentui/docs/src/app.tsx
```diff
@@ -8,6 +8,7 @@ import Routes from './routes';
 
 // Experimental dev-time accessibility attributes integrity validation.
 import { setup } from '@fluentui/ability-attributes';
+import { Telemetry } from '../../react-bindings/src';
```
mnajdova, line 11:
```suggestion
import { Telemetry } from '@fluentui/react-bindings';
```
?

packages/fluentui/CHANGELOG.md
```diff
@@ -81,9 +81,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Use `compose()` in `ToolbarRadioGroup` @layershifter ([#13322](https://github.com/microsoft/fluentui/pull/13322))
 - Use `compose()` in `Toolbar` @mnajdova ([#13341](https://github.com/microsoft/fluentui/pull/13341))
 - Use `compose()` in `ToolbarItem` @mnajdova ([#13321](https://github.com/microsoft/fluentui/pull/13321))
+- Add cache hits to telemetry data & differentiate initial render @layershifter ([#13346](https://github.com/microsoft/fluentui/pull/13346))
 - Use `compose()` in `ToolbarMenuItem` @mnajdova ([#13350](https://github.com/microsoft/fluentui/pull/13350))
 - Use `compose()` in `ToolbarCustomItem` @mnajdova ([#13319](https://github.com/microsoft/fluentui/pull/13319))
 
+
```
layershifter, line 88:
```suggestion
```

mnajdova:
I think that the telemetry data showed us a problem with composed component's caching, or the calculations are wrong.
image
Basically we don't have any cache hits for composed components. I would suggest looking into this before merging, at least to ensure that the data is correct, and we can investigate if there is any issue with the caching.. Other than that, the changes look good 👍

packages/fluentui/docs/src/app.tsx
```diff
@@ -20,6 +21,21 @@ const themes = {
   teamsHighContrastTheme,
 };
 
+const TelemetryGuard: React.FC<{
```
mnajdova, line 24: 👍

PR #13352, opened 2020-05-27
compose: general cleanup

dzearing:
Splitting out a minor PR here for some `compose` general cleanup to reduce my `resolve` PR complexity. :)

* Generics follow T{Name} convention for a little more readability.
* utils.ts utilities split out into individual files to keep things simpler (and to show at a glance there are missing tests for these.)
* react-is dependency was being used but not listed in package.json. I opted to remove the usage to mitigate any extra risk (or bundle size or if forward refs change, etc), because it doesn't seem necessary to validate the composed input is a forwardref to know if it was composed previously. if the input has the super secret static `fluentComposeConfig`, that seems good enough. I can revert this and add the listing in package.json if there is some critical reason to keep it.


dzearing:
I believe `react-is` seems to be missing a `sideEffects: "false"` flag in the package.json which is bloating the bundle. I noticed this using the "import-cost" vscode addon, which was reporting it at ~3k import cost. The size auditor results confirm that.

https://unpkg.com/react-is@16.13.1/package.json

PR #13354, opened 2020-05-27
compose typings cleanup, removing react-is usage

dzearing:
Splitting out a minor PR here for some `compose` general cleanup to reduce my `resolve` PR complexity. :)

* Generics follow T{Name} convention for a little more readability.
* utils.ts utilities split out into individual files to keep things simpler (and to show at a glance there are missing tests for these.)
* react-is dependency was being used but not listed in package.json. I opted to remove the usage to mitigate any extra risk (or bundle size or if forward refs change, etc), because it doesn't seem necessary to validate the composed input is a forwardref to know if it was composed previously. if the input has the super secret static `fluentComposeConfig`, that seems good enough. I can revert this and add the listing in package.json if there is some critical reason to keep it.


packages/react-compose/src/wasComposedPreviously.ts
```diff
@@ -0,0 +1,12 @@
+import { ComposedComponent, Input } from './types';
+
+/**
+ * compose() allows you to pass two inputs:
+ * - React.forwardRef + static fluentComposeConfig, i.e. previously composed component
+ * - a function
+ */
+export function wasComposedPreviously<T extends React.ElementType = 'div', P = {}>(
```
xugao, line 8: Nit: 
T -> TElement
P -> TProps

packages/react-compose/etc/react-compose.api.md
```diff
@@ -31,43 +31,43 @@ export interface ComponentWithAs<E extends React.ElementType = 'div', P = {}> ex
 export function compose<ElementType extends React.ElementType, InputProps, InputStylesProps, ParentProps, ParentStylesProps>(input: Input<ElementType, InputProps>, inputOptions?: ComposeOptions<InputProps, InputStylesProps, ParentProps, ParentStylesProps>): ComponentWithAs<ElementType, InputProps & ParentProps>;
```
xugao, line 31: why not adding T prefix for this too?

dzearing: missed one

PR #13355, opened 2020-05-27
creating react-conformance package

sopranopillow:
#### Pull request checklist

#### Description of changes

This PR adds the react-conformance package in order to have a general isConformant function for all components. This will allow to have a general testing criteria for all components. This is a work in progress and currently this is the initial commit. 

#### Focus areas to test

(optional)


packages/react-conformance/README.md
```diff
@@ -0,0 +1,7 @@
+# @fluentui/react-conformance
+
+**React Conformance for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
```
ecraig12345, line 3: Can you update the headline and description to be a little more descriptive? Also include a prominent note that the API is unstable.

packages/react-conformance/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "url": "https://github.com/microsoft/fluentui"
+  },
+  "license": "MIT",
+  "scripts": {
+    "build": "just-scripts build",
+    "clean": "just-scripts clean",
+    "code-style": "just-scripts code-style",
+    "just": "just-scripts",
+    "lint": "just-scripts lint",
+    "start-test": "just-scripts jest-watch",
```
ecraig12345, line 19: I think for now you should remove `start-test`, `test`, and `update-snapshots` because right now the package contains tests for reuse but doesn't actually have any of its own tests. (Though later we should probably add some basic tests for the tests.)

packages/react-conformance/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "test": "just-scripts test",
+    "update-snapshots": "just-scripts jest -u"
+  },
+  "devDependencies": {
```
ecraig12345, line 23: You're missing a devDependency on `jest` (copy the version from another package). You should also add these peerDependencies, which typically use a wider range:
```
"enzyme": "^3.0.0",
"enzyme-adapter-react-16": "^1.0.0",
"jest": "^24.0.0"
```
(these dep things won't matter locally but will be important for potential external consumers once the package is published)

packages/react-conformance/src/conformance.ts
```diff
@@ -0,0 +1,29 @@
...
+import { Details } from './types';
+
+function isConformant(Component: React.ComponentType, options: Details = {}) {
+  const { constructorName } = options;
+
+  if (!constructorName) {
+    throw new Error('Constructor name is needed');
+  }
+
+  const infoJSONPath = `@fluentui/docs/src/componentInfo/${constructorName}.info.json`;
```
ecraig12345, line 11: We won't be using anything related to .info.json, so you should remove these parts.

packages/react-conformance/src/conformance.ts
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import { Details } from './types';
+
+function isConformant(Component: React.ComponentType, options: Details = {}) {
```
ecraig12345, line 4: Instead of export default below I'd prefer that you just do `export function isConformant` and remove the default. (this is a stylistic difference between v0 and v7 which we haven't quite resolved.)

packages/react-conformance/src/conformance.ts
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
```
ecraig12345, line 1: Probably call this file `isConformant.tsx` (tsx since it will eventually contain jsx probably)

packages/react-conformance/src/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface Details {
```
ecraig12345, line 1: Couple comments here:
- Instead of Details probably call this something like `IsConformantOptions` (open to other suggestions too). 
- constructorName should be required.
- Think about what other info should be included in this and go ahead and add props for them (which won't be used yet):
  - A way to indicate enabled/disabled tests
  - A way to pass in extra tests
  - Depending on what all we can cover with lint rules, we may also need:
    - `exportedAtTopLevel?: boolean` (default true) like in the old isConformant
      - They'll also need to provide the top-level export for the package, typed as `any`
    - The path to the component file

We also still haven't totally determined whether these tests should run automatically on a glob of component files or whether they should be manually called for each component. Or we could potentially have both options--one exported function which runs tests for an individual component, and one which runs them for a glob (with a larger and slightly different options object).

packages/react-conformance/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "outDir": "lib",
+    "target": "es6",
+    "module": "commonjs",
+    "declaration": true,
+    "sourceMap": true,
+    "experimentalDecorators": true,
```
ecraig12345, line 9: Need to add `"jsx": "react"` somewhere in here (the package doesn't contain react code right now but will later)

packages/react-conformance/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "importHelpers": true,
+    "noUnusedLocals": true,
+    "forceConsistentCasingInFileNames": true,
+    "strictNullChecks": true,
+    "noImplicitAny": true,
```
ecraig12345, line 14: New packages should use `"strict": true` (we should update the template for this)
```suggestion
    "strict": true,
```

[What this does](https://www.typescriptlang.org/docs/handbook/compiler-options.html):
> Enabling --strict enables --noImplicitAny, --noImplicitThis, --alwaysStrict, --strictBindCallApply, --strictNullChecks, --strictFunctionTypes and --strictPropertyInitialization.

packages/react-conformance/tslint.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": ["@uifabric/tslint-rules"],
+  "rules": {}
```
ecraig12345, line 3: For now you'll need this:
```suggestion
  "rules": {
    "interface-name": [true, "never-prefix"]
  }
```

yarn.lock
```diff
@@ -4813,6 +4813,14 @@ array.prototype.flat@^1.2.1:
     es-abstract "^1.10.0"
     function-bind "^1.1.1"
 
+array.prototype.flat@^1.2.3:
```
ecraig12345, line 4816: Revert this file

PR #13357, opened 2020-05-27
Prevent ContextualMenu from re-mounting items when they are updated

ThomasMichon:
Fixed an issue where `ContextualMenu` would re-mount (destroy and recreate the HTML Elements) each menu item if the `items` prop changed while the menu is rendered.

The fix involves wrapping the render result for each item (typed as `React.ReactNode[]` in a `Fragment` and assigning a durable key. The issue was that the reconciler could not handle a possible 'array of arrays' in the output and was instead recreating all elements in the list.

Added an example for `CommandBar` with a 'lazy-loaded' command, to show how the bar reacts when commands are dynamically loaded.

### DIFF
--- a/packages/office-ui-fabric-react/src/components/CommandBar/examples/CommandBar.Lazy.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/CommandBar/examples/CommandBar.Lazy.Example.tsx
@@ -1,0 +1,72 @@
+import * as React from 'react';
+import { CommandBar, ICommandBarItemProps } from 'office-ui-fabric-react/lib/CommandBar';
+import { ContextualMenuItemType } from 'office-ui-fabric-react/lib/ContextualMenu';
+import { useConstCallback } from '@uifabric/react-hooks';
+
+export const CommandBarLazyExample: React.FunctionComponent = () => {
+  const [menuItems, setMenuItems] = React.useState<ICommandBarItemProps[] | undefined>(undefined);
+
+  const timeoutRef = React.useRef<number | null>();
+
+  const onMenuDismissed = useConstCallback(() => {
+    setMenuItems(undefined);
+  });
+
+  const loadItems = useConstCallback(() => {
+    const itemCount = Math.floor(Math.random() * 5) + 1;
+
+    const newMenuItems: ICommandBarItemProps[] = [];
+
+    for (let i = 0; i < itemCount; i++) {
+      newMenuItems.push({
+        key: `sub-item-${i}`,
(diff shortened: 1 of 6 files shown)


packages/office-ui-fabric-react/src/components/CommandBar/examples/CommandBar.Lazy.Example.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { ContextualMenuItemType } from '../../ContextualMenu';
+
+export const CommandBarLazyExample: React.FunctionComponent = () => {
+  const [menuItems, setMenuItems] = React.useState<ICommandBarItemProps[] | undefined>(undefined);
+
+  const onMenuDismissed = useConstCallback(() => {
+    setMenuItems(undefined);
+  });
+
+  const timeoutRef = React.useRef<NodeJS.Timeout>();
```
ecraig12345, line 13: This should be a number not a NodeJS.Timeout. Unfortunately the return type for setTimeout gets messed up since we have to include the node typings in the package for test purposes. So you'll have to cast when calling setTimeout later (unless you can find a better solution).

packages/office-ui-fabric-react/src/components/CommandBar/examples/CommandBar.Lazy.Example.tsx
```diff
@@ -0,0 +1,70 @@
...
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+    }
+
+    timeoutRef.current = setTimeout(() => {
+      setMenuItems(newMenuItems);
+    }, 2000);
+  });
+
+  const items: ICommandBarItemProps[] = [
```
ecraig12345, line 36: Shouldn't these be memoized in some way to avoid re-creating the array every time? Or is that the point of the example? (If so you should call it out in a comment.)

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -597,7 +597,7 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
         break;
     }
 
-    return renderedItems;
+    return <React.Fragment key={item.key}>{renderedItems}</React.Fragment>;
```
ecraig12345, line 600: Can you add a comment explaining why it's like this?

ecraig12345:
Looks like you need to update snapshots, and there may be some other test failures as well.

PR #13358, opened 2020-05-27
fix(UseKnob): Remove disable eslint line 

assuncaocharles:
#### Pull request checklist

#### Description of changes

This PR solves the infinity update with `useKnob` without disabling eslint or the line. 

#### Focus areas to test

(optional)


assuncaocharles:
@layershifter This solution solves the problem and avoid the eslint disable line

PR #13359, opened 2020-05-27
chore(Dialog): Convert to RFC

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

This PR converts `Dialog` component to be functional. Restricting props set that will be passed to styles functions.

Related to #12237

## Prop sets

| `Dialog`    |
| --------- |
| `backdrop` |

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,324 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
+    className,
+    design,
+    styles,
+    variables,
+  } = props;
+
+  const ElementType = getElementType(props);
+  const unhandledProps = useUnhandledProps(Dialog.handledProps, props);
+
+  const contentRef = React.useRef<HTMLElement>() as React.MutableRefObject<HTMLElement>;
```
layershifter, line 168:
```suggestion
  const contentRef = React.useRef<HTMLElement>();
```

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,324 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
+  const overlayRef = React.useRef<HTMLElement>() as React.MutableRefObject<HTMLElement>;
```
layershifter, line 169:
```suggestion
  const overlayRef = React.useRef<HTMLElement>() ;
```

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,324 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
-  }
-
-  handleDialogCancel = (e: Event | React.SyntheticEvent) => {
-    _.invoke(this.props, 'onCancel', e, { ...this.props, open: false });
-    this.setState({ open: false });
+
+  const handleDialogCancel = (e: Event | React.SyntheticEvent) => {
+    _.invoke(props, 'onCancel', e, { ...props, open: false });
+    setOpen(false);
+    unlockBodyScroll();
```
layershifter, line 241: This should be under `useEffect`

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,324 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
+  onConfirm: PropTypes.func,
+  onOpen: PropTypes.func,
+  open: PropTypes.bool,
+  overlay: customPropTypes.itemShorthand,
+  trapFocus: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
+  trigger: PropTypes.any,
+};
+
+Dialog.defaultProps = {
+  accessibility: dialogBehavior as Accessibility,
```
layershifter, line 437:
```suggestion
  accessibility: dialogBehavior,
```

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,324 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
+    unlockBodyScroll();
   };
 
-  handleDialogOpen = (e: React.SyntheticEvent) => {
-    _.invoke(this.props, 'onOpen', e, { ...this.props, open: true });
-    this.setState({ open: true });
+  const handleDialogOpen = (e: React.SyntheticEvent) => {
+    _.invoke(props, 'onOpen', e, { ...props, open: true });
+    setOpen(true);
+    lockBodyScroll();
```
layershifter, line 253: This should be under `useEffect`

assuncaocharles: It didn't really under `useEffect`, previously in the componentDidUpdate we were comparing the prev and curr state, didn't worked quite the same with `useEffect`

assuncaocharles: Just pushed inside the `useEffect` but to works as expected I had to set a flag to not go unlocking the body in the first render.

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,334 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
-  };
+  const [open, setOpen] = useAutoControlled({
+    defaultValue: props.defaultOpen,
+    value: props.open,
+    initialValue: false,
+  });
 
-  static autoControlledProps = ['open'];
-  static Footer = DialogFooter;
+  const lockBodyScroll = React.useCallback(() => {
```
layershifter, line 217: Move out these functions out of render, they don't need to be wrapped with `useCallback` as they have a single param `context.target`

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,334 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
+      context.target.body.removeAttribute(dialogsCounterAttribute);
+    }
+  }, [context.target.body]);
+
+  const [openedFirst, setOpenedFirst] = React.useState(props.open || props.defaultOpen);
+
+  React.useEffect(() => {
+    if (openedFirst) {
+      if (open) {
+        lockBodyScroll();
```
layershifter, line 243:
```suggestion
        lockBodyScroll(context.target);
```

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,334 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
+      } else {
+        unlockBodyScroll();
```
layershifter, line 245:
```suggestion
        unlockBodyScroll(context.target);
```

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,334 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
+      }
+    }
+  }, [lockBodyScroll, open, openedFirst, unlockBodyScroll]);
```
layershifter, line 248:
```suggestion
  }, [open]);
```

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
```diff
@@ -117,304 +132,334 @@ export const dialogSlotClassNames: DialogSlotClassNames = {
...
+    context.target.body.setAttribute(dialogsCounterAttribute, `${openDialogs}`);
+
+    // Only enables scroll if all dialogs are closed
+    if (openDialogs === 0) {
+      enableBodyScroll(context.target.body);
+      context.target.body.removeAttribute(dialogsCounterAttribute);
+    }
+  }, [context.target.body]);
+
+  const [openedFirst, setOpenedFirst] = React.useState(props.open || props.defaultOpen);
```
layershifter, line 238: I don't understand this and it should not be required

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
assuncaocharles: Moved to utils

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
assuncaocharles: If not set, the first render open will be `false`, and then will trigger `unlockBodyScroll` and it will count -1 dialog so breaks the way it supposed to work. Before it worked because we were comparing with `prevState`

```typescript
if(prevState.open && !state.open){
    unlockBodyScroll();
}
```

layershifter: Don't use `else` there

```
React.useEffect(() => {
  if (open) lock()

  return () => {
    if (open) unlock()
  }
})
```

packages/fluentui/react-northstar/src/components/Dialog/utils/index.tsx
```diff
@@ -15,3 +15,26 @@ export function disableBodyScroll(target: HTMLElement): void {
     target.addEventListener('touchmove', _disableIosBodyScroll, { passive: false, capture: false });
   }
 }
+
+const dialogsCounterAttribute = 'fluent-dialogs-count';
+
+export const lockBodyScroll = target => {
```
layershifter, line 21:
```suggestion
export const lockBodyScroll = (target: Document) => {
```

I suggest to keep`target` as `document` and type it

assuncaocharles: Alright, no problem. Gonna change, just out of curiosity.. Why?

layershifter: https://github.com/microsoft/fluentui/blob/4eb6f33eeb419a3189b8c7d2c5e5675fd86a77fd/packages/utilities/src/scroll.ts#L105

1. I would like to converge these functions
2. `body` may not exist on `document` and be `null`: https://developer.mozilla.org/en-US/docs/Web/API/Document/body

packages/fluentui/react-northstar/src/components/Dialog/Dialog.tsx
assuncaocharles: You are right, done.

packages/fluentui/react-northstar/src/components/Dialog/utils/index.tsx
assuncaocharles: Done.

PR #13360, opened 2020-05-27
compose: adding `mergeProps` utility to react-compose

dzearing:
This PR splits out more of my draft PR, and adds `resolve` to the compose options. This helper will help resolve `slots` and `slotProps` and effectively can replace the `resolveSlotProps` that is nested within `mergeComposeOptions`. Expected usage:

```tsx
const Button = compose((props, ref, options) => {
  const state = useButton(props); // optional
  const { slots, slotProps } = options.resolve(state /* or just props */);

  return <slots.root { ...slotProps.root } />;
}, {
  /* options */
});
```

 I have not removed the existing implementation so that the v0 code is not busted, but once this is in, that code can be updated.

I am going through the shorthand factory tests right now to validate I didn't miss any, and will be adding more tests to this PR but wanted to get this going sooner for early feedback.


packages/react-compose/src/createOptionsResolver.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        foo: { className: 'foo' },
+        bar: { className: 'bar' },
+      },
+    });
+    expect(resultedState).toBe(expectedState);
+  });
+
+  // it('can resolve slot props', () => {
+  //   expect(defaultResolve);
+  // });
```
khmakoto, line 120: Remove?

packages/react-compose/src/createOptionsResolver.ts
```diff
@@ -0,0 +1,67 @@
...
+    const state = stateOrProps as OptionsResolverResult['state'];
+    const slots: OptionsResolverResult['slots'] = { ...options.slots };
+    const result: OptionsResolverResult = {
+      state,
+      slotProps,
+      slots,
+    };
+
+    // Always ensure a root slot exists.
+    slots.root = state.as || slots.root || 'div';
```
khmakoto, line 41: Shouldn't this go before the assignment to `result` in line 34?

packages/react-compose/src/createOptionsResolver.ts
```diff
@@ -0,0 +1,67 @@
...
+    return result;
+  };
+};
+
+function assignToMapObject(map: Record<string, {}>, key: string, value: {}) {
+  if (value) {
+    if (!map[key]) {
+      map[key] = {};
+    }
+    Object.assign(map[key], value);
```
khmakoto, line 65: Do we have a polyfill for `Object.assign()`?

packages/react-compose/src/mergeComposeOptions.ts
```diff
@@ -1,10 +1,10 @@
 import { ComposeOptions, ComposePreparedOptions, Input } from './types';
```
khmakoto, line 1: Remove `src/utils.ts`?

packages/react-compose/src/resolveSlotProps.ts
```diff
@@ -0,0 +1,50 @@
...
+        const { children, ...restProps } = slotProp;
+        slotProp.children = slotProp.children(slot, restProps);
+
+        // tslint:disable-next-line:no-any
+        slots[slotName] = React.Fragment;
+      }
+
+      // Assign the resolves props.
+      slotProps[slotName] = {
+        // ...configSlotProps[slotName],
```
khmakoto, line 38: Remove?

packages/react-compose/src/mergeComposeOptions.ts
dzearing: hmm. it is removed...

packages/react-compose/src/createOptionsResolver.ts
dzearing: fixed

packages/react-compose/src/createOptionsResolver.ts
dzearing: it won't make a difference, but maybe it cleans up the code. The `result` assignment will assign the `slots` reference. We're mutating values but that won't change the reference.

packages/react-compose/src/resolveClasses.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    ).toEqual({
+      state: { className: 'cn' },
+      slotProps: { root: { className: 'root cn' } },
+      slots: {},
+    });
+  });
+
+  it('can distribute classes to slotProps', () => {
+    expect(
+      resolveClasses([{ root: 'root' }, { slot1: 'slot1' }, { unrecognized: 'ignored' }], {
```
xugao, line 16: i think we can use more tests here:
1. when `classes` contain overlapped prop: 
  `resolveClasses([{ root: 'root' }, { slot1: 'slot1' }, { slot1: 'slot2' }]`
2. when a class is a function
3. duplicated classes:
  `resolveClasses([{ root: 'root' }, { slot1: 'slot1' }, { slot1: 'slot1' }]`

packages/react-compose/src/createOptionsResolver.ts
```diff
@@ -0,0 +1,67 @@
...
+      slotProps,
+      'root',
+      getNativeElementProps(state.as, state, [...options.handledProps, 'className']),
+    );
+
+    // Resolve slotProps/slots from state.
+    resolveSlotProps(result);
+
+    // Resolve classes.
+    resolveClasses(options.classes, result);
```
xugao, line 54: a bit nervous about perf - i don't see much opportunity to memoize

packages/react-compose/src/resolveSlotProps.test.ts
```diff
@@ -0,0 +1,27 @@
+import { resolveSlotProps } from './resolveSlotProps';
+
+describe('resolveSlotProps', () => {
```
xugao, line 3: this can use more tests too, perhaps after a round of review

packages/react-compose/src/types.ts
```diff
@@ -84,9 +84,9 @@ export type GenericDictionary = Record<string, any>;
  */
 export type ClassDictionary = Record<string, string>;
 
-export type ComposePreparedOptions<TProps = {}> = {
+export type ComposePreparedOptions<TProps = {}, TState = TProps> = {
   className: string;
-  classes: ClassDictionary;
+  classes: (undefined | ClassDictionary | ((state: GenericDictionary, slots: GenericDictionary) => ClassDictionary))[];
```
xugao, line 89: define type `ClassFunction = (state: GenericDictionary, slots: GenericDictionary) => ClassDictionary)` and use both here and `createClassResolver`

packages/react-compose/src/resolveSlotProps.test.ts
dzearing: Already in progress.

packages/react-compose/src/types.ts
dzearing: done

packages/react-compose/src/createOptionsResolver.ts
dzearing: the logic in this function isn't substantially more expensive than manually cx'ing all things together. it loops through classes, sticks them in arrays, joins the arrays.

I think caching isn't going to help here

packages/react-compose/src/resolveClasses.test.ts
dzearing: done

packages/react-compose/src/types.ts
dzearing: done

packages/react-compose/src/resolveSlotProps.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export function resolveSlotProps(result: OptionsResolverResult): OptionsResolverResult {
+  const { state, slots, slotProps } = result;
+
+  Object.keys(slots).forEach((slotName: string) => {
+    const slot = slots[slotName];
+    let slotProp = state[slotName];
+
+    if (slot && slotProp !== undefined && slotProp !== null) {
+      const slotPropType = typeof slotProp;
```
joschect, line 19: Don't we always want slotProp to be an object? Is this a way fo doing shorthand?

packages/react-compose/src/createOptionsResolver.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    // Mix unrecognized props onto root, excluding the handled props.
+    assignToMapObject(
+      slotProps,
+      'root',
+      getNativeElementProps(state.as, state, [...options.handledProps, 'className']),
+    );
+
+    // Resolve slotProps/slots from state.
+    resolveSlotProps(result);
```
joschect, line 49: It looks like resolveSlotProps does return result, but it doesn't need to because it mutates the result. Instead, you could pass state, slotProps, and slots down and return a result and have that get set here.

packages/react-compose/src/resolveSlotProps.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Helper utility which takes in a classes array from compose options, resolves functions,
+ * merges them into a final result, and distributes classnames to slotProps within the given
+ * resolver result object.
+ */
+export function resolveSlotProps(result: OptionsResolverResult): OptionsResolverResult {
+  const { state, slots, slotProps } = result;
+
+  Object.keys(slots).forEach((slotName: string) => {
+    const slot = slots[slotName];
+    let slotProp = state[slotName];
```
joschect, line 16: Something to think about, it seems like the classnames could get passed through here and get resolved since each className must be associated with a particular slot. At least I think I'm reading that right.

packages/react-compose/src/resolveSlotProps.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    if (slot && slotProp !== undefined && slotProp !== null) {
```
joschect, line 18: Hey, why do you check slot to see if it's falsey but check slotProp and make sure it's neither undefined or null? Is something like `false` a valid slot prop?

packages/react-compose/src/resolveSlotProps.ts
```diff
@@ -0,0 +1,51 @@
...
+
+      // If the slot prop is a literal or JSX, pass it as a child of the slot.
+      if (isLiteral || React.isValidElement(slotProp)) {
+        const mappedProp =
+          (slot && slot.shorthandConfig && slot.shorthandConfig.mappedProp) || defaultMappedProps[slot] || 'children';
+        slotProp = { [mappedProp]: slotProp };
+      }
+
+      // If the children is a function, replace the slot.
+      if (typeof slotProp.children === 'function') {
```
joschect, line 30: Is this going to be documented somewhere? This is interesting hidden behavior. Especially since the slot that I provide could supply its own data stuff down to the children.

packages/react-compose/src/createOptionsResolver.ts
joschect: @dzearing it does seem like this adds another loop through all of the slots and all of the classes. Additionally that loop will grow every-time a component is recomposed since it will have to go through all of the old classes and all of the new classes right?

packages/react-compose/src/resolveSlotProps.ts
dzearing: slot: null is valid, and should skip the logic. slotProp={null} I wasn't sure how to handle so I assumed it to mean don't render this slot.

```
<Button chevron={null} />
```

packages/react-compose/src/resolveSlotProps.ts
dzearing: slotProp can be:

* a literal value
* JSX
* an object

packages/react-compose/src/resolveSlotProps.ts
dzearing: Yep. This is documented in v0 shorthand documentation, so I'm trying to preserve the logic to converge slot approaches.

packages/react-compose/src/resolveSlotProps.test.tsx
```diff
@@ -0,0 +1,108 @@
+import * as React from 'react';
+import { resolveSlotProps, NullRender } from './resolveSlotProps';
+
+describe('resolveSlotProps', () => {
```
dzearing, line 4: Need to add `style` merging tests.

packages/react-compose/src/mergeComposeOptions.ts
```diff
@@ -42,9 +43,10 @@ export function mergeComposeOptions(
       });
       return mergedSlotProps;
     }, {});
-  return {
+
+  const preparedOptions: Partial<ComposePreparedOptions> = {
     className: inputOptions.className || parentOptions.className,
-    classes: mergeClasses(parentOptions.classes, inputOptions.classes),
+    classes: [...parentOptions.classes, inputOptions.classes],
```
xugao, line 49: i don't like how we are re-resolving all classes from all compose layers. user should be able to replace/avoid styles calculation from parent components if the composed component doesn't need

packages/react-compose/src/resolveClasses.ts
```diff
@@ -0,0 +1,60 @@
...
+  Object.keys(result.slotProps).forEach((slotName: string) => {
+    addToMapArray(classMap, slotName, result.slotProps[slotName].className);
+  });
+
+  for (const classFunctionOrObject of classes) {
+    const classObj: ClassDictionary | undefined =
+      // tslint:disable-next-line:no-any
+      typeof classFunctionOrObject === 'function' ? classFunctionOrObject(state, slots as any) : classFunctionOrObject;
+    for (const key in classObj) {
+      if (classObj.hasOwnProperty(key)) {
```
xugao, line 25: Object.keys

packages/react-compose/src/createOptionsResolver.ts
dzearing: yes, but this is not a concern. think of it like this:

In sass components like legacy Calendar, we'd have this:


```tsx
render() {
return (
  <div className={ css(styles.root, props.primary && styles.primary) }>
    <div className={ css(styles.slot1, etc) }
```

This is exactly the same performance as something like this:

```tsx
const deriveClasses = (state) => ({
  root: css(styles.root, state.primary && styles.primary),
  slot1: css(styles.slot1),
});

render() {
  const classes = deriveClasses(state);

  return (
    <div className={classes.root}>
       <div className={classes.slot1}>
 ```

In a recompose scenario, we may have multiple of these deriving functions. We need to execute each of them and join them up. There is no reducing this at the high level, but that's irrelevant because the overhead of 3 functions over 2 functions is nearly 0 cost. It's what's *in the function* that's costly.

So in css in js world, we achieve fast perf, not by optimizing it at this layer, but within the class resolving process:

```tsx
const deriveClasses = state => {
  const theme = useTheme();

  // Call the function and mergeStyleSets only if the keys within the style props (2nd arg) have mutated.
  return memoizedStyles(styleProps) => {}, {theme, state.primary});
};
```

We can now call a number of these functions. The top level logic will append classnames together. 

(Small note: there is a feature in mergeStyles where we can auto expand previous registered classnames to new rules, avoiding specificity issues. We could take advantage of this by passing along previous class names to the function and including those in the memoization keys, but again, potentially hazardous if we get into some of the edge cases.)

packages/react-compose/src/mergeProps.ts
```diff
@@ -0,0 +1,39 @@
...
+    slotProps: {},
+  };
+
+  // Resolve slotProps/slots from state.
+  resolveSlotProps(result, options);
+
+  // Resolve classes.
+  resolveClasses(result, options.classes);
+
+  // Resolve inline styles.
```
khmakoto, line 36: Add a TODO here?
```suggestion
  // TODO: Resolve inline styles.
```

packages/react-compose/src/resolveClasses.test.ts
```diff
@@ -0,0 +1,98 @@
...
+          slots: { root: 'div', slot1: 'div' },
+        },
+        [state => ({ root: 'root' }), { root: 'root' }],
+      ),
+    ).toEqual({
+      state: {},
+      slotProps: { root: { className: 'root root root' } },
+      slots: { root: 'div', slot1: 'div' },
+    });
+  });
```
khmakoto, line 97: Aren't these essentially the same test?

dzearing: hmmm not sure what happened there. good catch.

packages/react-compose/src/createOptionsResolver.ts
dzearing: I would normally do that, but I've optimized to be able to chain resolvers to mutate an ongoing result for performance reasons. Doing otherwise would mean each resolver would need to allocate a new object and mix the old result. There isn't really a need to do so in an internal function.

packages/react-compose/src/resolveSlotProps.ts
dzearing: That's true, but there's more involved with resolving classnames done in the separate resolver (resolving functions with state) so it will be done separately. But yes perhaps we could combine the two loops. Trying to keep it clean for now, can t

packages/react-compose/src/resolveClasses.test.ts
dzearing: fixed.

dzearing:
I did some sleuthing with webpack bailouts to understand why the bundle size jumped so much.

1. There is some side effect code in utilities, causing us to pull in merge-styles. I've adjusted this and it brough the min size down to 8 KB.
2. If I remove the reference to `getNativeElementProps`, which is what's responsible for filtering out the props applicable to the native element, we bring the bundle down to 4.9 KB.

packages/react-compose/src/resolveClasses.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  return result;
+}
+
+/**
+ * Ensures that map[key] exists, or is initialized as any array, and pushes value to it.
+ */
+function addToMapArray(map: Record<string, string[]>, key: string, value: string) {
+  if (value) {
```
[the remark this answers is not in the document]
dzearing: fixed :)

dzearing:
I believe this is ready to go in.

PR #13363, opened 2020-05-27
Make ts composite builds use separate buildinfo files for variants

kenotron:
#### Pull request checklist

#### Description of changes

I believe the v0 code all have project references. With that there are caches being created. The only issue is that anything using just-scripts inside v0 will need to have separate caches based on the variants. This will delineate the caches based on the variants.

#### Focus areas to test

(optional)

### DIFF
--- a/scripts/tasks/ts.js
+++ b/scripts/tasks/ts.js
@@ -5,7 +5,7 @@ const { tscTask, argv } = require('just-scripts');
 const libPath = path.resolve(process.cwd(), 'lib');
 const srcPath = path.resolve(process.cwd(), 'src');
 // Temporary hack: only use tsbuildinfo file for things under packages/fluentui
-const isV0 = /[\\/]packages[\\/]fluentui\b/.test(process.cwd());
+const isV0 = /[\\/]packages[\\/]fluentui[\\/]/.test(process.cwd());
 
 function getExtraTscParams(args) {
   return {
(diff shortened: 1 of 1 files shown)


ecraig12345:
Alternatively if you want to parse a tsconfig this is how you do it.
```js
const fs = require('fs-extra');
const path = require('path');
const ts = require('typescript');

/**
 * Helper to perform the weirdly complicated task of parsing a tsconfig file accounting for `extends`.
 * @param {string} tsconfigPath
 * @returns {import('typescript').CompilerOptions}
 */
function parseTsconfig(tsconfigPath) {
  const { config, error } = ts.readConfigFile(tsconfigPath, filename => fs.readFileSync(filename, 'utf8'));

  if (error) {
    throw new Error(
      `Cannot load custom tsconfig.json from provided path: ${tsconfigPath}, ` +
        `with error code: ${error.code}, message: ${error.messageText}`,
    );
  }

  let basePath = path.dirname(tsconfigPath);
  if (config.compilerOptions && config.compilerOptions.baseUrl) {
    basePath = path.resolve(basePath, config.compilerOptions.baseUrl);
  }
  const { options, errors } = ts.parseJsonConfigFileContent(config, ts.sys, basePath, {}, tsconfigPath);

  if (errors && errors.length) {
    throw errors[0];
  }

  return options;
}
```

PR #13365, opened 2020-05-27
SliderNext: Completely pull state out to be slot-like

joschect:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-next/src/components/Slider/__snapshots__/Slider.test.tsx.snap
+++ b/packages/react-next/src/components/Slider/__snapshots__/Slider.test.tsx.snap
@@ -66,6 +66,14 @@ exports[`Slider renders correctly 1`] = `
           }
         />
         <span
+          className="ms-Slider-enabled ms-Slider-row ms-Slider-showValue ms-Slider-inactive ms-Slider-enabled ms-Slider-row ms-Slider-showValue"
+          style={
+            Object {
+              "width": "0%",
+            }
+          }
+        />
+        <span
           className="ms-Slider-enabled ms-Slider-row ms-Slider-showValue ms-Slider-active ms-Slider-enabled ms-Slider-row ms-Slider-showValue"
           style={
             Object {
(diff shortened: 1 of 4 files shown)


packages/react-next/src/components/Slider/__snapshots__/Slider.test.tsx.snap
```diff
@@ -65,6 +65,14 @@ exports[`Slider renders correctly 1`] = `
             }
           }
         />
+        <span
```
xugao, line 68: is this expected? if so, why?

packages/react-next/src/components/Slider/useSlider.ts
```diff
@@ -250,15 +335,22 @@ export const useSlider: (props: ISliderProps, ref: React.Ref<HTMLDivElement>) =>
...
     valueLabel: valueLabelProps,
-    classNames,
-    sliderLine,
-    value,
+    thumb: thumbProps,
+    zeroTick: zeroTickProps,
+    activeTrack: trackActiveProps,
+    topInactiveTrack: trackTopInactiveProps,
+    bottomInactiveTrack: trackBottomInactiveProps,
+    sliderLine: sliderLineProps,
```
xugao, line 354: i see a long way to go from here to use `compose`. but i don't mind to take this step :)

packages/react-next/src/components/Slider/__snapshots__/Slider.test.tsx.snap
joschect: Yes this is expected. The old slider had two ways of displaying the track. When originFromZero was on, there would be three sections, one inactive section from `min -> zero | value (whichever was lower)`, the active section from `zero -> value` and a final inactive section from `zero | value (whichever was lower) -> max`. When originFromZero was off there were only 2 sections the active section from `zero -> value` and an inactive section from `value -> max`.

My change keeps the `min->zero` inactive section even if originFromZero is off because it greatly simplifies the logic. Instead there is now a const `originValue` that is either 0 if originFromZero is true or whatever the min is. So all 3 tracks always render, even though the bottom one might not appear. 

I like this because it simplifies the logic that determines what is rendered and ensures that there is only 1 set of calculations needed to determine track positioning rather than two. It also adds more flexibility since it removes the tight coupling of originFromZero and tracks. We could now specify another origin and it will work as expected. We could even have a reverse slider that goes from max->min.

xugao: thanks for the detailed explaination!

PR #13366, opened 2020-05-28
Fix examples to properly remove duplicates

### LINKED ISSUE #12960: peoplePicker onEmptyInputFocus allow duplicates
### Environment Information

- **Package version(s)**: react: 16.8.5, office-ui-fabric-react: 6.189.2,

#### Actual behavior:
Hi! 
When you click on peoplePicker https://developer.microsoft.com/en-us/fluentui#/controls/web/peoplepicker (on any of the examples) you can choose a person from a suggestion list. But, after you pick one person, and click to the peoplePicker again, this person will still exist on the suggestion list. And, if you chose this person again, it will be duplicated in the selected people area and when you will open suggestion list for the third time, this person will disappear. 

I think, that the problem is in the function returnMostRecentlyUsed which updates mostRecentlyUsed with  setMostRecentlyUsed hook but the same time returns mostRecentlyUsed in the previous condition.

The same problem can be if you would like for example implement filter user functionality on the server. The onFilterChanged function will update filtered users to the hooks state (if you will define it) but will return old version of it to the NormalPeoplePicker component in the onResolveSuggestions parameter

#### Expected behavior:
returnMostRecentlyUsed should return relevant/current version of mostRecentlyUsed from useState hook. 

Are you willing to submit a PR to fix?  No

Requested priority: High

themilanfan:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12960

#### Description of changes

Fix PeoplePicker examples so the duplicates issue is resolved

### DIFF
--- a/change/office-ui-fabric-react-2020-05-27-20-15-46-peoplepicker-duplicates.json
+++ b/change/office-ui-fabric-react-2020-05-27-20-15-46-peoplepicker-duplicates.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fix examples to properly remove duplicates",
+  "comment": "PeoplePicker: Fix examples to properly remove duplicates",
   "packageName": "office-ui-fabric-react",
   "email": "suprememilanfan@gmail.com",
   "dependentChangeType": "patch",
--- a/change/office-ui-fabric-react-2020-05-27-20-15-46-peoplepicker-duplicates.json
+++ b/change/office-ui-fabric-react-2020-05-27-20-15-46-peoplepicker-duplicates.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fix examples to properly remove duplicates",
+  "packageName": "office-ui-fabric-react",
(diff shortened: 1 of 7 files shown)


change/office-ui-fabric-react-2020-05-27-20-15-46-peoplepicker-duplicates.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fix examples to properly remove duplicates",
```
ecraig12345, line 3: It's good to include the component name in this message since it's included in release notes.
```suggestion
  "comment": "PeoplePicker: Fix examples to properly remove duplicates",
```

khmakoto:
I just split the examples up but just copied the logic as it was. I could maybe take a deeper look tomorrow but the fastest way would probably be for @joschect to write his thoughts.

themilanfan:
Any updates on this PR?

ecraig12345:
Asked @joschect to take another look. Also closing and reopening to trigger rebuild.

PR #13367, opened 2020-05-28
SpinButton: reintroduce incorrect semi-controlled behavior

ecraig12345:
#### Pull request checklist

#### Description of changes

Prior to #13174, the SpinButton incorrectly had "semi-controlled" behavior: even when the `value` prop was provided, it would respect user updates until a new `value` was passed in. (Providing `value` is supposed to mean that the component is controlled and *will not* respect updates, unless the consumer passes in a new `value` in response to change events.) 

While the SpinButton will definitely have strict controlled behavior in version 8, unintentionally introducing it in version 7 was a breaking change for some consumers. So this PR reverts to the semi-controlled behavior.

### DIFF
--- a/change/@fluentui-react-next-2020-05-27-17-09-55-spinbutton-reintroduce-wrong-behavior.json
+++ b/change/@fluentui-react-next-2020-05-27-17-09-55-spinbutton-reintroduce-wrong-behavior.json
@@ -1,6 +1,6 @@
 {
   "type": "prerelease",
-  "comment": "SpinButton: reintroduce incorrect behavior of allowing updates when value is provided",
+  "comment": "SpinButton: Reintroduce incorrect behavior of allowing updates when value is provided to prevent breaking partners.",
   "packageName": "@fluentui/react-next",
   "email": "elcraig@microsoft.com",
   "dependentChangeType": "patch",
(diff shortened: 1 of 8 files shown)


change/@fluentui-react-next-2020-05-27-17-09-55-spinbutton-reintroduce-wrong-behavior.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "SpinButton: reintroduce incorrect behavior of allowing updates when value is provided",
```
khmakoto, line 3:
```suggestion
  "comment": "SpinButton: Reintroduce incorrect behavior of allowing updates when value is provided to prevent breaking partners.",
```

Just a suggestion because right now it would look weird on our release notes.

change/office-ui-fabric-react-2020-05-27-17-09-55-spinbutton-reintroduce-wrong-behavior.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "SpinButton: reintroduce incorrect behavior of allowing updates when value is provided",
```
khmakoto, line 3:
```suggestion
  "comment": "SpinButton: Reintroduce incorrect behavior of allowing updates when value is provided to prevent breaking partners.",
```

Just a suggestion because right now it would look weird on our release notes.

PR #13369, opened 2020-05-28
Toggle: Using compose in react-next version

khmakoto:
#### Pull request checklist

#### Description of changes

This PR changes the inner way in which `Toggle` is built by starting to use `compose` from `@fluentui/react-compose`.

#### Focus areas to test

(optional)


packages/react-next/src/components/Toggle/Toggle.types.ts
```diff
@@ -183,3 +203,26 @@ export interface IToggleStyles {
    */
   text: IStyle;
 }
+
+/**
+ * {@docCategory Toggle}
+ */
+export interface IToggleSlots {
```
xugao, line 210: as discussed, let's not expose any slot related typing yet. including the shorthands in ComponentProps

packages/react-next/etc/react-next.api.md
```diff
@@ -1110,7 +1118,361 @@ export { ThemeProviderProps }
 export const Toggle: React.FunctionComponent<IToggleProps>;
 
 // @public (undocumented)
-export const ToggleBase: React.FunctionComponent;
+export const ToggleBase: import("@fluentui/react-compose").ComponentWithAs<"div", IToggleProps>;
+
+// @public (undocumented)
+export const useToggle: (props: IToggleProps, options: ComposePreparedOptions<{}>) => {
```
xugao, line 1124: wow. could you give an `any` return type to remove the crazy return type for now?

khmakoto: Done

PR #13370, opened 2020-05-28
Update pivot to use static styles

behowell:
#### Pull request checklist

#### Description of changes

Update pivot to use static styles
Replace the PivotLinkFormat/PivotLinkSize enums with string union types

#### Focus areas to test

Ensure that the styling of pivot has not changed, especially the tabs in their different states: normal vs. large, links vs. tabs.


packages/react-next/src/components/Pivot/Pivot.tsx
```diff
@@ -1,8 +1,49 @@
...
+  rootIsTabs: 'ms-Pivot--tabs',
+  text: 'ms-Pivot-text',
+};
+
+const getStaticStylesMemoized = memoizeFunction(
+  (theme: ITheme, className?: string, linkSize?: PivotLinkSizeType, linkFormat?: PivotLinkFormatType) => {
+    const globalClassNames = getGlobalClassNames(GlobalClassNames, theme);
+
+    const rootModifiers = [
+      linkSize === 'large' && css(classes.linkSize_large, globalClassNames.rootIsLarge),
```
behowell, line 25: @dzearing is there a preferred way to add these modifier classes? I.e. a way to translate `linkSize === 'large' ` into `classes.linkSize_large`

packages/react-next/etc/react-next.api.md
```diff
@@ -992,18 +992,24 @@ export class PivotItem extends React.Component<IPivotItemProps, {}> {
     render(): JSX.Element;
 }
 
-// @public (undocumented)
-export enum PivotLinkFormat {
-    links = 0,
-    tabs = 1
+// @public @deprecated (undocumented)
+export const enum PivotLinkFormat {
```
xugao, line 996: hm why did you choose to make this change now? I feel we should avoid this change since there are already enough changes for us to worry about (need codemods/shim)

xugao:
pls add a perf test and a screener test for PivotNext. you can reference: #13167

packages/react-next/etc/react-next.api.md
behowell: This is updating the `linkFormat` and `linkSize` props to use the current pattern of just being a string with type restrictions instead of an enum.

The goal was for it to be backwards compatible without requiring a codemod. By redefining the enum values to be the same strings, existing code should work without modification. Anywhere that sets e.g. `linkFormat={PivotLinkFormat.tabs}` will now effectively be setting `linkFormat="tabs"` without requiring any code changes. But let me know if I'm wrong about that? Perhaps adding `@deprecated` could cause problems?

I could avoid modifying the API and convert the enum to a string internally, but I figured that since this was backwards compatible it would be nice to improve it while I'm modifying it. Let me know if you still disagree and I'll revert it.

behowell: The motivating reason is to make this prop be seamlessly compatible with style modifiers (`linkFormat_tabs`), but it's not necessary to modify the API for just that.

packages/react-next/src/common/_constants.scss
```diff
@@ -0,0 +1,12 @@
+$fontWeightLight: 100;
```
behowell, line 1: Let me know if there's a better place to define these values instead of a new file, or if there's another preferred way to define constants like this.

behowell:
> pls add a perf test and a screener test for PivotNext. you can reference: #13167

@xugao I just submitted the tests in a separate PR #13403 that I'd like to merge first, to help catch any regressions in this PR. Could you take a look when you have a chance? Thanks!

packages/react-next/etc/react-next.api.md
dzearing: I think this seems really nice. I think it should be backwards compatible, but is more the ideal setup. Devs can just pass in strings without importing enums.

xugao: ic now. sorry i mis-read this as a breaking change

packages/react-next/src/components/Pivot/Pivot.tsx
dzearing: Talked offline, but am going to try and make this cleaner.

packages/react-next/src/components/Pivot/Pivot.types.ts
```diff
@@ -112,10 +112,9 @@ export interface IPivotProps extends React.HTMLAttributes<HTMLDivElement> {
  */
 export type IPivotStyleProps = Required<Pick<IPivotProps, 'theme'>> &
   Pick<IPivotProps, 'className'> & {
-    /** Indicates whether Pivot has large format. */
-    rootIsLarge?: boolean;
-    /** Indicates whether Pivot has tabbed format. */
-    rootIsTabs?: boolean;
+    linkSize?: PivotLinkSizeType;
+    linkFormat?: PivotLinkFormatType;
+
```
khmakoto, line 117: Instead of having them declared here again, since you already have them in `IPivotProps` you could do:

```ts
export type IPivotStyleProps = Required<Pick<IPivotProps, 'theme'>> &
     Pick<IPivotProps, 'className', 'linkFormat', 'linkSize'> & {
```

packages/react-next/src/components/Pivot/Pivot.base.tsx
```diff
@@ -5,12 +5,13 @@ import { IPivotProps, IPivotStyleProps, IPivotStyles, IPivot } from './Pivot.typ
...
+import { css } from 'office-ui-fabric-react';
 import { useId } from '@uifabric/react-hooks';
 
-const getClassNamesCall = classNamesFunction<IPivotStyleProps, IPivotStyles>();
+const getClassNamesCall = classNamesFunction<IPivotStyleProps, IPivotStyles>({
+  useStaticStyles: true,
+});
 export interface IPivotState {
   selectedKey: string | undefined;
 }
```
khmakoto, line 17: I think `IPivotState` can go now that `Pivot` is a functional component.

packages/react-next/src/components/Pivot/Pivot.base.tsx
```diff
@@ -138,7 +139,7 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef(
           {...headerButtonProps}
           id={tabId}
           key={itemKey}
-          className={isSelected ? classNames.linkIsSelected : classNames.link}
+          className={css(classNames.link, isSelected && classNames.linkIsSelected)}
           onClick={onLinkClick.bind(this, itemKey)}
           onKeyPress={onKeyPress.bind(this, itemKey)}
```
jurokapsiar, line 144: is it intentional to use onKeyPress instead of onKeyDown which is typically used for keyboard handling?

packages/react-next/src/components/Pivot/Pivot.base.tsx
```diff
@@ -138,7 +139,7 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef(
...
           ariaLabel={link.ariaLabel}
```
jurokapsiar, line 145: please use dashed aria props as with camel cased it is not obvious which one wins. See https://hackmd.io/@fluentui/rkMawiQdI

packages/react-next/src/components/Pivot/Pivot.base.tsx
jurokapsiar: since this is not convergence yet, feel free to ignore. but it might still be worth checking.

packages/react-next/src/components/Pivot/Pivot.base.tsx
jurokapsiar: since this is not convergence yet, feel free to ignore.

packages/react-next/src/components/Pivot/Pivot.base.tsx
behowell: This is out of scope for this PR, but I logged #13429 to fix this in the future. Thanks.

PR #13374, opened 2020-05-28
perf: use mergeVariablesOverrides() to avoid cache breaks when variables are not passed

layershifter:
#### Pull request checklist

#### Description of changes

Fixes an issues that we have found with improved telemetry, see https://github.com/microsoft/fluentui/pull/13346#issuecomment-635210963.

This PR introduces `mergeVariablesOverrides()` which we already uses in `Attachment` to avoid creation of variables functions even if there is no overrides as that breaks caching.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-bindings/src/utils/mergeVariablesOverrides.ts
+++ b/packages/fluentui/react-bindings/src/utils/mergeVariablesOverrides.ts
@@ -1,4 +1,4 @@
-import { ComponentVariablesInput, mergeComponentVariables } from '@fluentui/styles';
+import { ComponentVariablesInput, ComponentVariablesPrepared, mergeComponentVariables } from '@fluentui/styles';
 
 /**
  mergeComponentVariables() is always creating a function even if the arguments are undefined we have this temporary
@@ -8,6 +8,6 @@ import { ComponentVariablesInput, mergeComponentVariables } from '@fluentui/styl
 export default function mergeVariablesOverrides(
   variables: ComponentVariablesInput,
   overrides: ComponentVariablesInput,
-) {
+): ComponentVariablesPrepared {
   return (variables || overrides) && mergeComponentVariables(variables, overrides);
 }
--- a/packages/fluentui/react-bindings/src/utils/mergeVariablesOverrides.ts
+++ b/packages/fluentui/react-bindings/src/utils/mergeVariablesOverrides.ts
@@ -1,0 +1,13 @@
+import { ComponentVariablesInput, mergeComponentVariables } from '@fluentui/styles';
+
+/**
+ mergeComponentVariables() is always creating a function even if the arguments are undefined we have this temporary
+ fix in place to avoid creating empty function because it is breaking caching we should either fix
+ mergeComponentVariables(), or handle this in a more generic way.
+ */
+export default function mergeVariablesOverrides(
+  variables: ComponentVariablesInput,
+  overrides: ComponentVariablesInput,
+) {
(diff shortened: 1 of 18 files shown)


packages/fluentui/react-bindings/src/utils/mergeVariablesOverrides.ts
```diff
@@ -0,0 +1,13 @@
+import { ComponentVariablesInput, mergeComponentVariables } from '@fluentui/styles';
+
+/**
+ mergeComponentVariables() is always creating a function even if the arguments are undefined we have this temporary
+ fix in place to avoid creating empty function because it is breaking caching we should either fix
+ mergeComponentVariables(), or handle this in a more generic way.
+ */
```
mnajdova, line 7: Do we still need this comment?

packages/fluentui/react-bindings/src/utils/mergeVariablesOverrides.ts
```diff
@@ -0,0 +1,13 @@
...
+export default function mergeVariablesOverrides(
```
mnajdova, line 8: should we add this under `@fluentui/styles` with the other merge functions?

packages/fluentui/react-bindings/src/utils/mergeVariablesOverrides.ts
layershifter: I would say yes to provide a context why this is a solution

packages/fluentui/react-northstar/src/components/Table/Table.tsx
```diff
@@ -90,43 +92,33 @@ export const Table: React.FC<WithAsProp<TableProps>> &
...
-        overrideProps,
+        overrideProps: predefinedProps => ({
+          variables: mergeVariablesOverrides(variables, predefinedProps.variables),
+        }),
       });
     });
   };
 
   const renderHeader = () => {
-    if (!header) {
```
mnajdova, line 112: Should we really remove this check?

packages/fluentui/react-bindings/src/utils/mergeVariablesOverrides.ts
layershifter: I don't have there any preference, however I placed it here as it's more attached to component's specifics

packages/fluentui/react-northstar/src/components/Table/TableRow.tsx
```diff
@@ -113,13 +115,12 @@ const TableRow: React.FC<WithAsProp<TableRowProps>> & FluentComponentStaticProps
...
-    return _.map(items, (item: TableCellProps, index: number) => {
-      const overrideProps = handleVariablesOverrides(variables);
+    return _.map(items, (item: TableCellProps) => {
       return TableCell.create(item, {
-        defaultProps: () =>
-          getA11yProps('cell', {
-            ...overrideProps,
-          }),
+        defaultProps: () => getA11yProps('cell', {}),
+        overrideProps: predefinedProps => ({
```
mnajdova, line 121: 👍

packages/fluentui/react-northstar/src/components/Table/Table.tsx
layershifter: Yup, as `.create()` call few lines below handles this:

https://github.com/microsoft/fluentui/blob/e5adec2a91f1b920d1b0a2ab5e29828f2a7f4c30/packages/fluentui/react-northstar/src/components/Table/Table.tsx#L111-L124

PR #13335, opened 2020-05-27
Add autoCapitalize property

themilanfan:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13231
- [x] Include a change request file using `$ yarn change`

#### Description of changes

Added autoCapitalize property to inputProperties and textAreaProperties

PR #13387, opened 2020-05-28
Fix update release notes author name

### LINKED ISSUE #13376: Github releases authors are broken
Hey folks, I noticed that your releases are not printing the correct authors names.

### Environment Information

- **Package version(s)**: none
- **Browser and OS versions**: none

Visit the release page: https://github.com/microsoft/fluentui/releases

Notice the author's names.

#### Actual behavior:
`[object Object]` is being printed.

Example:

```
Add autoCapitalize property (PR #13335 by [object Object])
```

#### Expected behavior:

The author name should appear instead of `[object Object]`

Requested priority: Low

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13376

#### Description of changes

The update release notes script was generating descriptions like:
`Add autoCapitalize property (PR #13335 by [object Object])`

Fix that and also get rid of the warning `[@octokit/rest] const Octokit = require("@octokit/rest") is deprecated. Use const { Octokit } = require("@octokit/rest") instead` that was showing up when running `yarn change`.

(I already ran the script to fix the existing release notes from the past 50 days when this was broken.)

### DIFF
--- a/scripts/beachball/customRenderers.ts
+++ b/scripts/beachball/customRenderers.ts
@@ -1,5 +1,5 @@
 import { spawnSync } from 'child_process';
-import GitHubApi from '@octokit/rest';
+import { Octokit as GitHubApi } from '@octokit/rest';
 import { PackageChangelogRenderInfo, ChangelogEntry } from 'beachball';
 import { getPullRequestForCommit, fluentRepoDetails } from '../github';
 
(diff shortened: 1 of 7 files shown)


scripts/beachball/customRenderers.ts
```diff
@@ -1,5 +1,5 @@
 import { spawnSync } from 'child_process';
-import GitHubApi from '@octokit/rest';
+import { Octokit as GitHubApi } from '@octokit/rest';
```
xugao, line 2: nit: let's `expose { Octokit as GitHubApi } from '@octokit/rest';` somewhere and `import { GitHubApi } from 'types'` here and other places

ecraig12345: Doing that feels a bit weird since it's not actually one of our types. I guess I could also just update all the references to say Octokit since that's clearly what they want to be called.

PR #13389, opened 2020-05-28
Remove explicit blocking of Alt key for keytip start sequence

kelseyyoung:
#### Pull request checklist

#### Description of changes

When writing keytips initially we had blocked the recognition of the "Alt" key because that is reserved for putting focus on the top bar of the browser. However there are users of Keytips that are writing their app in an Electron application that looks and behaves like a native desktop app. In native desktop apps the default key for keytips is just "Alt", and because they don't run in a traditional browser they won't have any issues using it

#### Focus areas to test

Tested that with removing these lines that you can specify "Alt" to start your keytips. If done in a traditional browser this will "break" in the sense that keytips will appear but focus will also go to the browser top bar, meaning the user can't interact with keytips


kelseyyoung:
When I run "yarn change" from the root it shows that I have changes in like 20 packages, and that there is also already a change file for office-ui-fabric-react. Not sure what I should do here? yarn clean maybe?

ecraig12345:
@kelseyyoung You probably need to update the URL for your `upstream` remote to microsoft/fluentui.

kelseyyoung:
@ecraig12345 any chance you could take a look at this?

ecraig12345:
@kelseyyoung I don't really know anything about keytips and their expected behavior--so maybe get @jspurlin to review instead? (looks like he's worked on this in the past)

PR #13393, opened 2020-05-28
Add focus mode icon

AnastasiyaSarmant:
#### Pull request checklist

#### Description of changes
Added a focus mode icon
#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
+++ b/packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
@@ -1,0 +1,23 @@
+import * as React from 'react';
+import cx from 'classnames';
+import createSvgIcon from '../utils/createSvgIcon';
+import { iconClassNames } from '../utils/iconClassNames';
+
+const FocusModeIcon = createSvgIcon({
+  svg: ({ classes }) => (
+    <svg role="presentation" focusable="false" viewBox="0 0 16 16" className={classes.svg}>
+      <g>
+        <path
+          className={cx(iconClassNames.outline, classes.outlinePart)}
+          d="M1,16c-0.6,0-1-0.4-1-1V1c0-0.6,0.4-1,1-1h14c0.6,0,1,0.4,1,1v14c0,0.6-0.4,1-1,1c0,0,0,0,0,0L1,16z
+			 M0.9,1.2v13.6c0,0.2,0.1,0.3,0.3,0.3h13.6c0.2,0,0.3-0.1,0.3-0.3V1.2c0-0.2-0.1-0.3-0.3-0.3H1.2C1,0.9,0.9,1,0.9,1.2z M9.7,12.4
+			v-0.9h3.7v0.9H9.7z M4.4,12.2c-1-0.4-1.7-1.5-1.7-2.6h2.8V6.8C7,6.8,8.3,8,8.3,9.6c0,1.6-1.3,2.8-2.8,2.8
+			C5.1,12.4,4.7,12.3,4.4,12.2L4.4,12.2z M9.7,10V9.1h3.7V10H9.7z M9.7,7.6V6.8h3.7v0.9H9.7z M3,4.5v-1h10v1H3z"
+        />
+      </g>
+    </svg>
+  ),
+  displayName: 'FocusModeIcon',
+});
+
+export default FocusModeIcon;
--- a/packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
+++ b/packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
@@ -1,14 +1,11 @@
 import * as React from 'react';
-import cx from 'classnames';
 import createSvgIcon from '../utils/createSvgIcon';
-import { iconClassNames } from '../utils/iconClassNames';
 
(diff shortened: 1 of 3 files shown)


packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+          className={cx(iconClassNames.outline, classes.outlinePart)}
+          d="M1,16c-0.6,0-1-0.4-1-1V1c0-0.6,0.4-1,1-1h14c0.6,0,1,0.4,1,1v14c0,0.6-0.4,1-1,1c0,0,0,0,0,0L1,16z
+			 M0.9,1.2v13.6c0,0.2,0.1,0.3,0.3,0.3h13.6c0.2,0,0.3-0.1,0.3-0.3V1.2c0-0.2-0.1-0.3-0.3-0.3H1.2C1,0.9,0.9,1,0.9,1.2z M9.7,12.4
+			v-0.9h3.7v0.9H9.7z M4.4,12.2c-1-0.4-1.7-1.5-1.7-2.6h2.8V6.8C7,6.8,8.3,8,8.3,9.6c0,1.6-1.3,2.8-2.8,2.8
+			C5.1,12.4,4.7,12.3,4.4,12.2L4.4,12.2z M9.7,10V9.1h3.7V10H9.7z M9.7,7.6V6.8h3.7v0.9H9.7z M3,4.5v-1h10v1H3z"
+        />
+      </g>
+    </svg>
+  ),
+  displayName: 'FocusModeIcon',
```
codepretty, line 20: we have been trying to use generic icon names that describe the icon and not the usecase, but I'm not sure `FocusModeIcon` I have a good description for this one...

packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+import cx from 'classnames';
+import createSvgIcon from '../utils/createSvgIcon';
+import { iconClassNames } from '../utils/iconClassNames';
+
+const FocusModeIcon = createSvgIcon({
+  svg: ({ classes }) => (
+    <svg role="presentation" focusable="false" viewBox="0 0 16 16" className={classes.svg}>
+      <g>
+        <path
+          className={cx(iconClassNames.outline, classes.outlinePart)}
```
pompomon, line 11: @AnastasiyaSarmant , please could you either remove the classnames at all (if the icon would not have distincting styles for `outline` and `filled`), like it's done in [CallDialpadIcon](https://github.com/microsoft/fluentui/blob/ee761dace1bb944941eb0ca3500f4cbee7f9b49c/packages/fluentui/react-icons-northstar/src/components/CallDialpadIcon.tsx#L14) for example. Or add a separate class and svg for `filled` icon like we have in [AcceptIcon](https://github.com/microsoft/fluentui/blob/ee761dace1bb944941eb0ca3500f4cbee7f9b49c/packages/fluentui/react-icons-northstar/src/components/AcceptIcon.tsx#L11)

packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
AnastasiyaSarmant: There is also possibility to call it ComponentContentIcon. If it sounds better to you I'll rename it

packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
AnastasiyaSarmant: removed

packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
notandrew: This is what it looks like -- I don't see anything similar in the Fluent icon list. What's the icon used for? We should be using Fluent icons going forward...

image

packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import createSvgIcon from '../utils/createSvgIcon';
+
+const FocusModeIcon = createSvgIcon({
+  svg: ({ classes }) => (
+    <svg role="presentation" focusable="false" viewBox="0 0 16 16" className={classes.svg}>
+      <g>
```
notandrew, line 7: You can remove the _g_ tag since it's not doing anything

packages/fluentui/react-icons-northstar/src/components/FocusModeIcon.tsx
AnastasiyaSarmant: > This is what it looks like -- I don't see anything similar in the Fluent icon list. What's the icon used for? We should be using Fluent icons going forward...
> 
> image

This is the exact reason why I'm adding this icon, you don't have it in Fluent. It's used for the "Focus" menu in calling controls.

notandrew: It's ok for now, but can it be CallingControlMenuIcon or something more specific?

PR #13399, opened 2020-05-29
fix(DropdownSelectedItem): Fix render imagen passing children on DropdownSelectedItem

### LINKED ISSUE #13351: DropdownSelectedItems not rendering image using children
https://codesandbox.io/s/fluent-ui-example-ymknn?file=/example.js

image

#### Actual behavior:

SelectedItem element doesn't render image using `children` props the same way it does for with `render` props

#### Expected behavior:

SelectedItem element renders image using `children` props the same way it does for with `render` props

Requested priority: High

assuncaocharles:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13351

#### Description of changes

This PR fix the error preventing `DropdownSelectedItem` to render the selected image using `children`

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -155,19 +155,21 @@ const DropdownSelectedItem: React.FC<WithAsProp<DropdownSelectedItemProps>> &
...
-  const imageProps = _.isNil(image)
-    ? image
-    : {
-        children: (ComponentType, props) =>
-          Image.create(image, {
-            defaultProps: () => ({
-              avatar: true,
-              className: dropdownSelectedItemSlotClassNames.image,
-              styles: resolvedStyles.image,
+  const imageChildren = _.get(image, 'children');
```
assuncaocharles, line 158: I tried in previously commit to solve it using `childrenExist` but it will be true for text only and then not working as expected. If you don't agree with this solution, have any other idea? @layershifter and @pompomon

packages/fluentui/react-northstar/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -155,19 +155,20 @@ const DropdownSelectedItem: React.FC<WithAsProp<DropdownSelectedItemProps>> &
...
-    ? image
-    : {
-        children: (ComponentType, props) =>
-          Image.create(image, {
-            defaultProps: () => ({
-              avatar: true,
-              className: dropdownSelectedItemSlotClassNames.image,
-              styles: resolvedStyles.image,
+  const imageProps =
+    _.isNil(image) || image?.hasOwnProperty('children')
```
assuncaocharles, line 159: I tried in previously commit to solve it using childrenExist but it will be true for text only and then not working as expected. If you don't agree with this solution, have any other idea? @layershifter and @pompomon

layershifter: Expected fix there is to remove `Label` component from there like it was done with `DropdownItem`

https://github.com/microsoft/fluent-ui-react/pull/2382/

layershifter: In this case there is no need to call factories for `imageProps`

assuncaocharles: Done

packages/fluentui/react-northstar/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -132,57 +135,61 @@ const DropdownSelectedItem: React.FC<WithAsProp<DropdownSelectedItemProps>> &
...
+      }),
+    },
+  );
+
+  const iconElement = Box.create(icon, {
     defaultProps: () => ({
-      as: 'span',
-      className: dropdownSelectedItemSlotClassNames.header,
-      styles: resolvedStyles.header,
+      'aria-label': `Remove ${header} from selection.`, // TODO: Extract this in a behaviour.
```
jurokapsiar, line 156: how would the user customize the label? Is it through Dropdown's `getA11ySelectionMessage`?

assuncaocharles: Resolving this to focus the discussion on the PR #13410

packages/fluentui/react-northstar/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -132,57 +135,61 @@ const DropdownSelectedItem: React.FC<WithAsProp<DropdownSelectedItemProps>> &
...
-            defaultProps: () => ({
-              avatar: true,
-              className: dropdownSelectedItemSlotClassNames.image,
-              styles: resolvedStyles.image,
-            }),
-            overrideProps: props,
-          }),
-      };
+  const imageElement = Box.create(
+    Image.create(image, {
```
layershifter, line 164: This should be under `overrideProps`:

```
Box.create(media, {
  overrideProps: {
    children: Image.create()
  }
}

File: packages/fluentui/react-northstar/src/components/Dropdown/DropdownSelectedItem.tsx [hunk 5]
```diff
@@ -132,57 +135,61 @@ const DropdownSelectedItem: React.FC<WithAsProp<DropdownSelectedItemProps>> &
...
-          Image.create(image, {
-            defaultProps: () => ({
-              avatar: true,
-              className: dropdownSelectedItemSlotClassNames.image,
-              styles: resolvedStyles.image,
-            }),
-            overrideProps: props,
-          }),
-      };
+  const imageElement = Box.create(
```
layershifter, line 163:
```suggestion
  const mediaElement = Box.create(
```

packages/fluentui/react-northstar/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -132,57 +135,61 @@ const DropdownSelectedItem: React.FC<WithAsProp<DropdownSelectedItemProps>> &
...
+      defaultProps: () => ({
+        avatar: true,
+        className: dropdownSelectedItemSlotClassNames.image,
+        styles: resolvedStyles.image,
+      }),
+    }),
+    {
+      defaultProps: () => ({
+        className: dropdownSelectedItemSlotClassNames.image,
+        styles: resolvedStyles.media,
```
layershifter, line 174: I don't see styles for this element. Is it required to have it in markup?

packages/fluentui/react-northstar/src/components/Dropdown/DropdownSelectedItem.tsx
assuncaocharles: Moved

packages/fluentui/react-northstar/src/components/Dropdown/DropdownSelectedItem.tsx
assuncaocharles: Removed

assuncaocharles:
@layershifter Implemented the solution you proposed, can you have a 2nd look? 😁

packages/fluentui/react-northstar/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -132,57 +135,63 @@ const DropdownSelectedItem: React.FC<WithAsProp<DropdownSelectedItemProps>> &
...
 
   const element = (
     <Ref innerRef={itemRef}>
-      <Label
+      <ElementType
         className={classes.root}
         tabIndex={active ? 0 : -1}
-        styles={resolvedStyles.root}
-        circular
+        styles={resolvedStyles.main}
```
pompomon, line 186: Please why do you use `main` style here? And where is it defined, I don't see it in `dropdownSelectedItemStyles`

assuncaocharles: Changed by @layershifter

PR #13402, opened 2020-05-29
Adding IE11 screener config for v7

dzearing:
Letting the build go to see if this just works for validating IE11 in screener.


JustSlone:
@dzearing do we still need this PR? We should probably complete, close it, or switch it to draft.

PR #13406, opened 2020-05-29
Move eslint-plugin package and add keyboard event props rule

ecraig12345:
#### Pull request checklist

#### Description of changes

Move the `@fluentui/eslint-plugin` package from under `packages/fluentui` to just `packages`, in preparation for @sopranopillow adding more custom lint rules for component conformance. This will change the package's release cadence (immediate release instead of lockstep), but I couldn't find any external or internal usage of the plugin outside our repo, so it should be fine.

Also added an ESLint implementation of @khmakoto's `deprecated-keyboard-event-props` TSLint rule (with improved parent type verification logic) so it will be ready when we move v7 to ESLint.

In addition I updated all the ESLint-related deps in preparation for v7 moving to ESLint. This required some new rule disables in v0 code, plus I updated the v0 ESLint config with some additional `extends` configs that disable rules that are unneeded with TS and/or Prettier.


packages/fluentui/state/src/createManager.ts
```diff
@@ -4,9 +4,9 @@ const createManager = <State, Actions extends AnyActions>(
   config: ManagerConfig<State, Actions>,
 ): Manager<State, Actions> => {
   const { actions, debug, middleware = [], sideEffects = [], state } = config;
-  const _state: State = Object.assign({}, state) as State;
+  const _state: State = { ...state } as State;
 
-  const getState = (): State => Object.assign({}, _state);
+  const getState = (): State => ({ ...state } as State);
```
layershifter, line 9: Oh

```suggestion
  const getState = (): State => ({ ..._state } as State);
```

layershifter: And even

```suggestion
  const getState = (): State => ({ ..._state });
```

ecraig12345: Oops, thanks for noticing that!!

ecraig12345:
Merged this so I can use it in my big ESLint PR #13074. v0 folks in particular, if you have any further feedback just let me know and I'll integrate it in a later change.

PR #13408, opened 2020-05-30
Enabling no-any tslint rule on react-next package and disabling it on a case by case basis

khmakoto:
#### Pull request checklist

#### Description of changes

This PR enables the `no-any` tslint rule on the `react-next` package and disables it on a case by case basis where needed.

#### Focus areas to test

(optional)


ecraig12345:
Unless you're able to merge this very soon, I'd prefer to do it as part of my eslint conversion PR instead.

PR #13410, opened 2020-05-30
chore(DropdownSelectedItem): add dropdown selected item behavior

assuncaocharles:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/fluentui/accessibility/src/behaviors/Dropdown/dropdownSelectedItemBehavior.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+const dropdownSelectedItemBehavior: Accessibility<DropdownSelectedItemBehaviorProps> = props => ({
+  attributes: {
+    root: {
+      'aria-selected': true,
+      tabIndex: props.active ? 0 : -1,
+    },
+    icon: {
+      'aria-label': `Remove ${props.header} from selection.`,
```
jurokapsiar, line 15: how would user customize the label (it needs to be translated)? Is it by using Dropdown's `getA11ySelectionMessage` prop?

assuncaocharles: As it was static previously I just moved to the behavior. I don't think there's a way to change it now, should it be a prop from `DropdownSelectedItem`?

jurokapsiar: there needs to be a way to customize it, because not all of our users speak english :) can you pls look into it? maybe it is in handleIconOverrides?

assuncaocharles: What we have in `handleIconOverrides` is:

```typescript
  const handleIconOverrides = iconProps => ({
    ...iconProps,
    onClick: (e: React.SyntheticEvent, iconProps: BoxProps) => {
      e.stopPropagation();
      _.invoke(props, 'onRemove', e, iconProps);
      _.invoke(props, 'onClick', e, iconProps);
    },
    onKeyDown: (e: React.SyntheticEvent, iconProps: BoxProps) => {
      e.stopPropagation();
      if (keyboardKey.getCode(e) === keyboardKey.Enter) {
        _.invoke(props, 'onRemove', e, iconProps);
      }
      _.invoke(props, 'onKeyDown', e, iconProps);
    },
  });
```
The only way I can see to override the message with current code is if a icon would be passed:

```jsx
  {
    header: 'Bruce Wayne',
    image: 'public/images/avatar/small/matt.jpg',
    content: 'Software Engineer',
    icon: <CloseIcon aria-label="Remover" />
  },
```

pompomon: I think we have a bit bigger issue - I've tried to remove selected item using keyboard and was unable to focus on the "remove" icon

assuncaocharles: Good question, it's possible to remove the item using `backspace`. I wonder if as @pompomon  pointed out it's possible to focus using the screen reader. @kolaps33 are you aware about it?

assuncaocharles: I tested it with voice over, and there's no way to reach the `X` icon. But it seem to behave the same in other components like `<Input clearable />` where the user can only remove it by using `backspace`.

kolaps33: @pompomon this is by design from the beginning :)
But generally I see there some gaps with screen reader, how user will figure out that they can remove the user with Delete/Backspace key. 
Will create new ticket and will update comment here as well later today.

PR #13415, opened 2020-06-01
Removing side effect code from @uifabric/utilities

dzearing:
Found some super old side effect code that is causing webpack to bail out of optimizations.

Restructuring code to execute only in needed circumstances.


dzearing:
@xugao Nice catch, done!

PR #13416, opened 2020-06-01
v8 Button: Adding inline tokens prop

dzearing:
This change of adding a tokens prop to Button.

I've also cleaned up the scss in terms of naming conventions, hover/active states (which got munged by a rename).

I've also added a number of stories to the Button to demonstrate inline tokens.

There are some rough edges I've hit worth discussion:

It's annoying to have to pass in every state - if background default is provided, I expect hover to not be required. This requires some rework of the :root definition to not include default button colors.


packages/react-button/src/components/Button/Button.stories.tsx
```diff
@@ -28,14 +28,17 @@ export const ButtonCss = () => (
...
     <h3>An icon button can format its Icon to appear before or after its content.</h3>
     <div className={classes.vStack}>
       <ButtonVariants iconPosition="before" />
       <ButtonVariants iconPosition="after" />
     </div>
 
-    <h3>A button can inherit its background and have a subtle appearance using the `inverted` prop.</h3>
-    <ButtonVariants inverted />
+    {/* <h3>A button can inherit its background and have a subtle appearance using the `inverted` prop.</h3>
+    <ButtonVariants inverted /> */}
```
khmakoto, line 41: Why are you commenting this one?

packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -1,5 +1,14 @@
...
 import { ComponentProps, ShorthandValue, ComposeOptions, ComposeStandardStatics } from '../../utils/tempTypes';
+import { TokenSetType } from '@fluentui/react-theme-provider';
+
+export type RecursivePartial<T> = {
+  [P in keyof T]?: T[P] extends (infer U)[]
+    ? RecursivePartial<U>[]
+    : T[P] extends object
+    ? RecursivePartial<T[P]>
+    : T[P];
+};
```
khmakoto, line 11: This is just a temporary location right? I feel like this should be in the `compose` typings.

packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -82,3 +93,68 @@ export type ButtonSlotProps = {
...
+  default: string;
+  hovered: string;
+  focused: string;
+  pressed: string;
+  checked: string;
+  checkedHovered: string;
+  checkedFocused: string;
+  checkedPressed: string;
+  disabled: string;
+};
```
khmakoto, line 160: Same with this, it should probably go in the `compose` layer.

packages/react-button/src/components/Button/useButton.ts
```diff
@@ -1,11 +1,21 @@
...
-export const useButton = (props: ButtonProps, options: ComposePreparedOptions) =>
-  mergeProps<ButtonProps, ButtonSlots, ButtonSlotProps>(props, options);
+export const useButton = (props: ButtonProps, options: ComposePreparedOptions) => {
+  const { tokens, ...state } = props;
+
+  // This this go into mergeProps?
+  if (tokens) {
+    // tslint:disable-next-line: no-any
+    state.style = { ...props.style, ...tokensToStyleObject(tokens as Record<string, any>, '--button') };
+  }
```
khmakoto, line 18: I don't know if it should go into `mergeProps` but maybe this whole thing could be the utility that you're importing instead of having to manually spread `...props.style` here.

```suggestion
  const state = {
    ...props,
    style: tokensToStyleObject(props, '--button')
  };

File: packages/react-button/src/components/Button/Button.stories.tsx (hunk 1 above)
Comment (dzearing) [reply packages/react-button/src/components/Button/Button.stories.tsx:41]: It doesn't work.

File: packages/react-button/src/components/Button/Button.types.tsx (hunk 2 above)
Comment (dzearing) [reply packages/react-button/src/components/Button/Button.types.tsx:11]: will need to move it. I have i think 3 copies now.

File: packages/react-button/src/components/Button/Button.types.tsx (hunk 3 above)
Comment (dzearing) [reply packages/react-button/src/components/Button/Button.types.tsx:160]: probably

Comment (dzearing) [reply packages/react-button/src/components/Button/Button.types.tsx:160]: not `compose` but probably base theme

File: packages/react-button/src/components/Button/Button.scss [hunk 5]
```diff
@@ -40,42 +41,45 @@
...
-  --button-text: rgb(44, 38, 33);
-  --button-text-hovered: var(--button-text);
-  --button-text-focused: var(--button-text);
-  --button-text-pressed: var(--button-text);
-  --button-text-disabled: rgb(200, 198, 196);
+  --button-background: white;
+  --button-background-hovered: rgb(237, 235, 233);
+  --button-background-focused: var(--button-background);
+  --button-background-pressed: rgb(225, 223, 221);
+  --button-background-disabled: rgb(237, 235, 233);
```
dzearing, line 48: typo

packages/react-button/src/components/Button/Button.stories.tsx
dzearing: removed.

packages/react-button/src/components/Button/Button.types.tsx
dzearing: moved to temptypes.

packages/react-button/src/components/Button/Button.types.tsx
dzearing: moved to compose

xugao:
can you update the  ButtonNext perf-test to have tokens prop set?

dzearing:
@xugao  missed your comment, will do right now in separate pr.

PR #13308, opened 2020-05-25
perf: remove `camelCaseProperty` call in `renderRule()` inside Fela

layershifter:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [ ] Include a change request file using `$ yarn change`

#### Description of changes

[`renderRule()`](http://fela.js.org/docs/basics/Renderer.html#renderrule) is a hot function for us as we are calling it on each our cache miss. A callstack for Fela looks like this:

```
renderRule(rule, props)
  _renderStyle((rule, props)) // styles are passed resolved; performs plugins calls
    _renderStyleToClassNames(style) // performs Fela cache lookups
````

Inside `_renderStyleToClassNames()` exists a call to `generateDeclarationReference()` what creates a lookup key for Fela cache:

https://github.com/robinweser/fela/blob/97148419e7941dbd5867b35df6ac4de6c5ed3e55/packages/fela/src/createRenderer.js#L269-L277

```js
          const declarationReference = generateDeclarationReference(
            property,
            value,
            pseudo,
            media,
            support
          )

          if (!renderer.cache.hasOwnProperty(declarationReference)) {
```

Inside `generateDeclarationReference()` there is call to [`camelCaseProperty()`](https://github.com/robinweser/css-in-js-utils/blob/master/modules/camelCaseProperty.js) which performs a transform to force camelcase via regexps (and actually this makes this function slow). This is visible in our perfs and in my test environment:

```
- Fela                                        69346 ops/sec
- Fela without `camelCaseProperty()`         261801 ops/sec
```

<details>
 <summary>Test rule</summary>

```js
const fullRule = () => ({
  backgroundColor: "transparent",
  backgroundPosition: "center",
  display: "flex",
  float: 0,
  left: 0,
  right: 0,
  flexGrow: 1,
  marginLeft: 0,
  marginRight: 0,
  marginBottom: 0,
  marginTop: 0,
  paddingLeft: 0,
  paddingRight: 0,
  paddingBottom: 0,
  paddingTop: 0,
  outline: 0,
  opacity: 1,
  height: "20px",
  maxHeight: "20px",
  minWidth: "20px",
  width: "20px",
  visibility: "hidden",
  zIndex: 1,
});
```

</details>

------

__Sad part:__ This will break rehydration process from SSR content, but as we are not using it we are fine to go.

------

## Benchmarks 🚀 

A quick improvement summary, all data from benchmarks are available below:

### `new-page`
| Example                      | Improvement |
[...]

PR #13417, opened 2020-06-01
chore: add mode to perf tests

layershifter:
#### Pull request checklist

#### Description of changes

This PR adds `mode` setting to v0 perf tests to better understand caching effect as it runs test suites on the same page.

#### Focus areas to test

(optional)


layershifter:
Was merged as a part of #13308.

PR #13418, opened 2020-06-01
[WIP] feat(FormField): Add error message, error indicator and satisfactory indicator

assuncaocharles:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/accessibility/src/behaviors/Form/formFieldBehavior.tsx
+++ b/packages/fluentui/accessibility/src/behaviors/Form/formFieldBehavior.tsx
@@ -1,0 +1,24 @@
+import { Accessibility, AccessibilityDefinition } from '../../types';
+
+/**
+ * @specification
+ */
+const formFieldBehavior: Accessibility<FormFieldBehaviorProps> = props => {
+  const definition: AccessibilityDefinition = {
+    attributes: {
+      root: {},
+      control: {
+        ...(props.hasErrorMessage && { 'aria-invalid': true }),
+      },
+    },
+  };
+
+  return definition;
+};
+
+export default formFieldBehavior;
+
+export type FormFieldBehaviorProps = {
+  /** Field has error message */
+  hasErrorMessage?: boolean;
+};
--- a/packages/fluentui/react-northstar/src/components/Form/FormField.tsx
+++ b/packages/fluentui/react-northstar/src/components/Form/FormField.tsx
@@ -1,8 +1,8 @@
-import { Accessibility } from '@fluentui/accessibility';
+import { Accessibility, FormFieldBehaviorProps, formFieldBehavior } from '@fluentui/accessibility';
 import * as customPropTypes from '@fluentui/react-proptypes';
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
-
+import { ExclamationCircleIcon } from '@fluentui/react-icons-northstar';
 import {
   childrenExist,
   createShorthandFactory,
@@ -28,7 +28,7 @@ export interface FormFieldProps extends UIComponentProps, ChildrenComponentProps
   /**
    * Accessibility behavior if overridden by the user.
    */
-  accessibility?: Accessibility<never>;
+  accessibility?: Accessibility<FormFieldBehaviorProps>;
 
   /** A control for the form field. */
   control?: ShorthandValue<BoxProps>;
@@ -53,11 +53,19 @@ export interface FormFieldProps extends UIComponentProps, ChildrenComponentProps
 
   /** The HTML input type. */
   type?: string;
+
+  errorMessage?: ShorthandValue<TextProps>;
+
+  errorIndicator?: ShorthandValue<BoxProps>;
+
+  satisfactoryIndicator?: ShorthandValue<BoxProps>;
 }
 
 export const formFieldClassName = 'ui-form__field';
 
-export type FormFieldStylesProps = Required<Pick<FormFieldProps, 'type' | 'inline' | 'required'>>;
+export type FormFieldStylesProps = Required<Pick<FormFieldProps, 'type' | 'inline' | 'required'>> & {
+  hasErrorMessage: boolean;
+};
 
 const FormField: React.FC<WithAsProp<FormFieldProps>> & FluentComponentStaticProps<FormFieldProps> = props => {
   const context: ProviderContextPrepared = React.useContext(ThemeContext);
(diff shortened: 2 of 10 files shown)


packages/fluentui/react-northstar/src/components/Form/FormField.tsx
```diff
@@ -112,14 +127,31 @@ const FormField: React.FC<WithAsProp<FormFieldProps>> & FluentComponentStaticPro
...
   const controlElement = Box.create(control || {}, {
-    defaultProps: () => ({ required, id, name, type, styles: resolvedStyles.control }),
+    defaultProps: () =>
+      getA11yProps('control', {
+        required,
+        id,
+        name,
+        type,
+        icon: !!errorMessage ? iconElement : null,
+        ...(!!satisfactoryIndicator && { satisfactoryIndicator }),
```
assuncaocharles, line 152: @layershifter @pompomon I had to add this condition to only pass it when defined otherwise it would break saying it's not a valid DOM property in the box creation for different inputs, e.g `Checkbox`. `icon` works only because it's not `camelCase`. Glad to change if there's a better alternative.

packages/fluentui/react-northstar/src/components/Input/Input.tsx
```diff
@@ -281,6 +295,8 @@ Input.propTypes = {
   type: PropTypes.string,
   value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
   wrapper: customPropTypes.wrapperShorthand,
+  required: PropTypes.bool,
```
assuncaocharles, line 298: I added required here to be a handled prop, therefore not adding it to the DOM element as per spec. But like that I have no idea how we can make it clear it is a required field, my first thought was to used [`aria-required`](https://www.w3.org/TR/WCAG20-TECHS/ARIA2.html) but the spec also suggest against it. Didn't completely got the reason tho. @kolaps33 @jurokapsiar

kolaps33: I probably did mistake in summary because based on Adam comment(see below) "aria-required" should behave differently and could be used. :)
Sorry about confusion. 

The “required” HTML attribute is not used on the form fields because on the form submission, when some required field is empty, the browser’s native alert message takes priority and makes the custom live region message not be spoken. The “aria-required” appears to behave differently, as it currently does not cause the browser’s native message be narrated upon a submission of a form with an empty required field.

assuncaocharles: Great, added `aria-required` then.

kolaps33:
Charles havent' you introduced the prop "hintMessage" by purpose? or just it was forgotten ? 
if you take look here 
https://hackmd.io/NcQ6HF3OQK2STU8WT0s7Cw?view
or in the spec it is in the very bottom of it :)

kolaps33:
do you have plan to add ?
1. usage of the error message

2. add e2e tests or screener tests? or maybe screener would be enough :) to cover scenarios which can happen :)

assuncaocharles:
> Charles havent' you introduced the prop "hintMessage" by purpose? or just it was forgotten ?
> if you take look here
> https://hackmd.io/NcQ6HF3OQK2STU8WT0s7Cw?view
> or in the spec it is in the very bottom of it :)

Not really sure if we need `hintMessage` property since we already have `message` property, I think it would work the same.

assuncaocharles:
> do you have plan to add ?
> 
> 1. usage of the error message
> 2. add e2e tests or screener tests? or maybe screener would be enough :) to cover scenarios which can happen :)

Tests and example added

packages/fluentui/docs/src/examples/components/Form/Usage/FormExampleErrorAndSatisfactory.shorthand.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
```
jurokapsiar, line 1: please add children api example as well

assuncaocharles: Oh, ok, I thought was supposed to be a separated task. Will added it here.

assuncaocharles: Added.

packages/fluentui/react-northstar/src/components/Form/FormField.tsx
```diff
@@ -53,11 +54,23 @@ export interface FormFieldProps extends UIComponentProps, ChildrenComponentProps
...
   type?: string;
+
+  /** Message to be shown when input has error */
+  errorMessage?: ShorthandValue<TextProps>;
+
+  /** Indicator to be shown together with error message */
+  errorIndicator?: ShorthandValue<BoxProps>;
+
+  /** Indicator to be shown when field is required and non-empty */
+  satisfactoryIndicator?: ShorthandValue<BoxProps>;
```
pompomon, line 65:
```suggestion
  successIndicator?: ShorthandValue<BoxProps>;
```
Could we call it `successIndicator` or does that not reflect the purpose of this indicator?

assuncaocharles: I guess we can change, any objections @kolaps33 ?

assuncaocharles:
Anything else to be changed here @layershifter @kolaps33 or can I merge it?

packages/fluentui/react-northstar/src/components/Form/FormField.tsx
assuncaocharles: changed

packages/fluentui/accessibility/src/behaviors/Form/formFieldBehavior.tsx
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * @specification
+ */
+const formFieldBehavior: Accessibility<FormFieldBehaviorProps> = props => {
+  const definition: AccessibilityDefinition = {
+    attributes: {
+      root: {},
+      control: {
+        ...(props.hasErrorMessage && { 'aria-invalid': true }),
+        ...(props.messageId && { 'aria-describedby': props.messageId }),
```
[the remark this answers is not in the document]
assuncaocharles: Changed

packages/fluentui/react-northstar/src/components/Input/Input.tsx
```diff
@@ -137,6 +155,7 @@ const Input: React.FC<WithAsProp<InputProps>> & FluentComponentStaticProps<Input
     initialValue: '',
   });
   const hasValue: boolean = !!value && (value as string)?.length !== 0;
+  const requiredAndSatisfactory = required && !!successIndicator && hasValue;
```
pompomon, line 158:
```suggestion
  const requiredAndSuccess = required && !!successIndicator && hasValue;
```
Since we have `successIndicator` ?

assuncaocharles: Done

PR #13420, opened 2020-06-01
Updating spec for slider and checkbox 

kolaps33:
## Description of changes

Updating specs based on accessibility validation. 

Results for particular component can be found here:
https://jurokapsiar.github.io/open-a11y/


specs/Checkbox.md
```diff
@@ -316,12 +322,31 @@ Same behavior as above except no preview of toggled state through hover.
...
 ### Screenreader accessibility:
 
+#### `Fluent UI & Fluent UI Northstar comparison`:
+
+From accessibility point of view the main difference is in type of element used for Checkbox itself:
+
+- Fluent UI has `<input>` element with type="checkbox"
+- Fluent UI Northstar has `<div>` element with role="checkbox"
+
+  [Verification with screen reader](https://jurokapsiar.github.io/open-a11y/components/checkbox.research) doens't show any differences between these two approaches, everything was working as expected.
```
dzearing, line 332: the difference is that an `input` receives focus and manages keyboard events naturally. If you allow the input to be changed to another type of element (not sure if there is a reason to), you'll need to manage these manually.

I believe `input` is needed for an html form scenario as well. Is there any reason why v0 doesn't use `input`?

specs/Slider.md
```diff
@@ -89,26 +89,26 @@ https://developer.microsoft.com/en-us/fabric#/controls/web/slider
...
-| variables                   | any                                    | Consider only recomposition                            |
-| vertical                    | boolean                                |                                                        |
+| Name                        | Type                                   | Notes                                                                                                                                                         |
+| --------------------------- | -------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| accessibility               | "sliderBehavior" any                   | Why would a user need this as a prop?                                                                                                                         |
+| animation                   | AnimationProp                          | Why would a user need this as a prop on slider?                                                                                                               |
```
dzearing, line 95:
```suggestion
| animation                   | AnimationProp                          | Why is this needed?                                                                                                               |
```

specs/Slider.md
```diff
@@ -310,6 +310,26 @@ The same behavior as above, except that the events should concern
 
 ### Screen reader accessibility
 
+#### `Fluent UI & Fluent UI Northstar comparison`:
```
dzearing, line 313: @joschect check these and make sure you understand if we have a problem that would be solved with an input element, or a couple of input elements for range selection.

specs/Checkbox.md
jurokapsiar: no specific reason for this case - they both behave good for screen readers and yes, the benfit of using input is that it works out of the box with forms.

PR #13426, opened 2020-06-01
fixing an issue with the module classnames

dzearing:
Makes the classnames easier to read in storybook.


xugao:
note: we have diverged sass pipeline for prod vs. dev mode. ideally we can use the `sass` task somehow for dev mode

xugao:
could you remove `sass` and `ts` from  `task('dev:storybook', series('copy', 'sass', 'ts', 'storybook:start'));`

PR #13428, opened 2020-06-01
Enable Commandbar resizeGroup caching

micahgodbolt:
Commanbar cache was not working. Appears that resizeGroup requires an initial cache value in props.

### DIFF
--- a/change/office-ui-fabric-react-2020-06-01-14-39-06-commandbar-cache.json
+++ b/change/office-ui-fabric-react-2020-06-01-14-39-06-commandbar-cache.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "CommandBar: Fixed cacheing",
+  "packageName": "office-ui-fabric-react",
+  "email": "mgodbolt@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-06-01T21:39:06.626Z"
+}
--- a/change/office-ui-fabric-react-2020-06-01-14-39-06-commandbar-cache.json
+++ b/change/office-ui-fabric-react-2020-06-01-14-39-06-commandbar-cache.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "CommandBar: Fixed cacheing",
+  "comment": "CommandBar: Fixed caching to greatly improve performance for all users",
   "packageName": "office-ui-fabric-react",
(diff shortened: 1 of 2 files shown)


change/office-ui-fabric-react-2020-06-01-14-39-06-commandbar-cache.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "CommandBar: Fixed cacheing",
```
dzearing, line 3:
```suggestion
  "comment": "CommandBar: Fixed caching",
```

micahgodbolt: the built in caching was not working. Customer doesn't need to do anything. This should be a huge perf gain for anyone using command bar.

packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.base.tsx
```diff
@@ -81,7 +81,10 @@ export class CommandBarBase extends React.Component<ICommandBarProps, {}> implem
       overflowItems: [...overflowItems!],
       minimumOverflowItems: [...overflowItems!].length, // for tracking
       farItems,
-      cacheKey: '',
+      cacheKey: this._computeCacheKey({
+        primaryItems: [...items],
```
joschect, line 85: Does items get mutated by _computeCacheKey? If not it seems like you could pass items in without expanding it.

packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.base.tsx
```diff
@@ -265,7 +267,7 @@ export class CommandBarBase extends React.Component<ICommandBarProps, {}> implem
       primaryItems = shiftOnReduce ? primaryItems.slice(1) : primaryItems.slice(0, -1);
 
       const newData = { ...data, primaryItems, overflowItems };
-      cacheKey = this._computeCacheKey(newData);
+      cacheKey = this._computeCacheKey({ primaryItems, overflow: overflowItems.length > 0 });
```
joschect, line 270: I think that the computeCacheKey maybe should be responsible for determining what to do with overflowItems. Not a necessary change since all of this is private but I think it helps keep the cache generation in one spot.

PR #13430, opened 2020-06-01
Unify usage of mergeProps and delete duplicate code

dzearing:
This change unifies v0 and v7 on `mergeProps` for resolving slotProps. 

`mergeProps` handles a number of concerns:

* resolves slotProps merging
* handles shorthand resolution
* handles classes merging
* mixes in native props based on the `as` value or `root` slot, defaulting to default html props (and excluding values in `handledProps`)

Rather than passing the resolver through options, we can use a single imported utility to resolve props, slots, slotProps and classes.

Keeping the resolver as a separated import lets us change it in the future without requiring the helper to be part of the options blob. This opens up opportunities to break up utilities or reconsider the algorithm without changing the prepared options structure.

Before:
```
const Foo = compose((props, ref, options) => {
  const state = {...props, otherstuff };
  const slotProps = options.mergeSlotProps(state);
  return <> ... </>;
});
```

After:
```tsx
const Foo = compose((props, ref, options) => {
  const state = {...props, otherstuff };
  const {slotProps} = mergeProps(state, options);
  return <> ... </>;
});
```
Note this does not move the v0 code to similar slots conventions. That would be good as a separate PR/effort.


dzearing:
Cool, yup, I'll dig in!

packages/react-compose/src/resolveSlotProps.ts
```diff
@@ -25,12 +29,19 @@ export function resolveSlotProps<TProps, TState>(
...
-    'root',
-    getNativeElementProps(slots.root, state, [...(options.handledProps as string[]), 'className']),
-  );
+  //  Mix unrecognized state onto root, excluding the handled props.
+  Object.keys(state).reduce<Partial<TState>>((acc, propName: string) => {
+    if (
+      !defaultHandledProps.has(propName) &&
+      !options.slots.hasOwnProperty(propName) &&
+      propName !== 'as' &&
+      (options.handledProps as string[]).indexOf(propName) === -1
```
dzearing, line 38: this could be changed into a set lookup potentially

dzearing: This should reduce bundle size and certainly perf overhead but will put a strong reliance on handledProps.

dzearing:
I am going to come back to this momentarily; i've checked in some updates to improve perf, just need to integrate into here.

dzearing:
I've made some improvements, but let's get state right and we'll reopen this.

PR #13431, opened 2020-06-02
Compose: Adding replaceable state as part of compose

khmakoto:
#### Pull request checklist

#### Description of changes

This draft PR is a first stab at envisioning how adding a layer for a replaceable `state` would look like in `compose` and integrating it into the `Link` and `Toggle` components for testing. There probably needs to be some work regarding type safety but I'm putting this out to gather feedback.

#### Focus areas to test

(optional)


packages/react-button/src/components/Button/ButtonBase.tsx
```diff
@@ -4,8 +4,8 @@ import { compose } from '@fluentui/react-compose';
 import { useButton } from './useButton';
 
 export const ButtonBase = compose<'button', ButtonProps, ButtonProps, {}, {}>(
-  (props, ref, composeOptions) => {
-    const { slots, slotProps } = useButton(props, composeOptions);
+  (props, ref, slotsAndState) => {
```
dzearing, line 7: What if the first option is state rather than user props? I don't like the idea of constantly creating new options on every render

dzearing, line 7: i'm unsure of why this was renamed. I think it should still just be options.

Also the first value could be the computed state.

packages/react-compose/src/createClassResolver.ts
```diff
@@ -18,6 +18,7 @@ import { appendClasses } from './appendClasses';
 export const createClassResolver = (classes: ClassDictionary) => {
   // This is in creation time, so this will happen once per css file.
   const { slots, modifiers, enums } = createResolvedMap(classes);
+  console.log(slots, modifiers, enums);
```
xugao, line 21: remove

packages/react-button/src/components/ToggleButton/useToggleButton.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    setChecked(!checked);
+  };
+
+  const state = { ...props, checked, setChecked, onClick };
+
+  if (parentState) {
+    return parentState(state, options);
+  }
```
xugao, line 31: I wonder if this is better:
1. having useComponentState hook which doesn't `mergeProps`
2. have Base component to `mergeProps`

xugao: if we go what your approach, what's the advantage of this over calling `useButton` directly here?

packages/react-button/src/components/Button/ButtonBase.tsx
khmakoto: These aren't the options anymore, but I guess I could pass this as part of the options object itself, I just saw no need to pass options if we're not calling state hooks on the render function, but let me know if I'm mistaken!

packages/react-button/src/components/ToggleButton/useToggleButton.ts
khmakoto: I guess I'll answer this in 3 parts:

- For point 1, I think even without doing this work that would be preferable.
- For point 2, did you mean calling `mergeProps` on the render function? I would prefer to not do that as it would make a hard dependency on it. I'd much rather have separate `useComponent` and `useComponentState` hooks where the latter is called by the earlier.
- Regarding the benefits of this approach over calling `useButton` directly, the whole reason behind this is to be able to substitute or add state on top of what is provided by default. You can imagine a user substituting or expanding the `useButton` state on a recompose. You could still add the toggling capabilities by using this hook on recomposition of that new component without adding anything else and calling their new state. The idea here would be to basically eliminate the hard dependencies that would be created in calling the `useButton` state.

khmakoto:
Closing in favor of #13521

PR #13435, opened 2020-06-02
Checkbox: use compose and remove styled

xugao:
#### Pull request checklist

#### Description of changes
- Use `compose` on `Checkbox`. 
- Removed `styled` and added `useCheckboxClasses` hook
- **IMPORTANT**: this change breaks `styles` prop completely and we do not plan to continue supporting it going forward 

#### Focus areas to test

(optional)


packages/react-compose/src/createClassResolver.ts
```diff
@@ -14,7 +14,7 @@ import { GenericDictionary, ClassDictionary } from './types';
  * Remaining class names would be interpretted as modifiers, applied to the `root` slot when
  * the component `state` contains a truthy matching prop name.
  */
-export const createClassResolver = (classes: ClassDictionary) => (
+export const createClassResolver = (...classesList: ClassDictionary[]) => (
```
xugao, line 17: I updated changes to `createClassResolver` and created another PR #13443 for it

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -1,152 +1,60 @@
...
+              aria-describedby={mergeAriaAttributeValues(
+                slotProps.input['aria-describedby'],
+                keytipAttributes['aria-describedby'],
+              )}
+            />
+            <slots.container {...slotProps.container}>
+              <slots.checkbox {...slotProps.checkbox} data-ktp-target={keytipAttributes['data-ktp-target']}>
+                <slots.checkmark {...slotProps.checkmark} />
+              </slots.checkbox>
+              {(props.onRenderLabel || onRenderLabel)(props, onRenderLabel)}
```
jurokapsiar, line 43: shouldn't compse and shorthands allow us to stop using custom `onRender` props by using children callback instead? @dzearing

xugao: Yup! but if we remove this, it's a breaking change so i am leaving it here. i think the right thing do to is to deprecate and promote using slot

packages/react-next/src/Styling/useGlobalClassNames.ts
```diff
@@ -0,0 +1,11 @@
+import { getGlobalClassNames, GlobalClassNames, useTheme } from '../Styling';
+
+export function useGlobalClassNames<T>(classNames: GlobalClassNames<T>): GlobalClassNames<T> {
+  const { tokens: theme } = useTheme();
+
+  return getGlobalClassNames(
+    classNames,
+    // tslint:disable-next-line:no-any
+    theme as any,
```
dzearing, line 9: unclear why useTheme is required.

packages/react-next/src/Styling/useGlobalClassNames.ts
```diff
@@ -0,0 +1,11 @@
+import { getGlobalClassNames, GlobalClassNames, useTheme } from '../Styling';
+
+export function useGlobalClassNames<T>(classNames: GlobalClassNames<T>): GlobalClassNames<T> {
+  const { tokens: theme } = useTheme();
+
+  return getGlobalClassNames(
```
dzearing, line 6: If these were removed, would it improve perf?

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -1,152 +1,80 @@
...
-import { useId, useControllableValue, useMergedRefs } from '@uifabric/react-hooks';
-import { useFocusRects } from 'office-ui-fabric-react';
+import { useCheckbox } from './useCheckbox';
 
-const getClassNames = classNamesFunction<ICheckboxStyleProps, ICheckboxStyles>({
-  useStaticStyles: true,
-});
+const defaultSlots: Omit<ICheckboxSlots, 'root'> = {
+  input: 'input',
+  checkmark: Icon,
```
dzearing, line 11: The base component should not have styled slot references; would it be possible to change this to null and let the composed thing choose?

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
dzearing: I think we need to break. we should not use the new AND the old patterns, and we should instead rely on the compat layer imo to optimize perf of the new one.

dzearing: if you render a label slot, there is a clear upgrade path and it works as expected.

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -1,152 +1,80 @@
...
-  const [isIndeterminate, setIsIndeterminate] = useControllableValue(props.indeterminate, props.defaultIndeterminate);
+    const onRenderLabel = (): JSX.Element | null => {
+      return label ? <slots.text {...slotProps.text}>{label}</slots.text> : null;
+    };
 
-  useFocusRects(rootRef);
-  useDebugWarning(props);
-  useComponentRef(props, isChecked, isIndeterminate, checkBox);
+    return (
+      <KeytipData keytipProps={keytipProps} disabled={disabled}>
```
dzearing, line 27: is this being removed in a separate PR?

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -1,152 +1,80 @@
...
+          </slots.root>
+        )}
+      </KeytipData>
+    );
+  },
+  {
+    slots: defaultSlots,
+    displayName: 'CheckboxBase',
+    handledProps: [
+      'componentRef',
```
dzearing, line 54: I wish there were a more automatic way of doing this. After seeing this list. afraid that doing this in bulk will be more expensive than getNativeProps bundle size wise. There is no reusability here.

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -1,152 +1,80 @@
...
-        if (checkBox.current) {
-          checkBox.current.focus();
-        }
-      },
-    }),
-    [isChecked, isIndeterminate],
-  );
-}
+CheckboxBase.defaultProps = {
+  as: 'div',
```
dzearing, line 79: is this needed? can root slot be div?

packages/react-next/src/components/Checkbox/Checkbox.types.ts
```diff
@@ -192,3 +199,28 @@ export interface ICheckboxStyles {
...
+/**
+ * {@docCategory Checkbox}
+ */
+export interface ICheckboxSlots {
+  root: React.ElementType;
+  input: React.ElementType;
+  container: React.ElementType;
+  checkbox: React.ElementType;
+  checkmark: React.ElementType;
+  text: React.ElementType;
```
dzearing, line 212: should this be label?

packages/react-next/src/components/Checkbox/useCheckbox.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const [isChecked, setIsChecked] = useControllableValue(props.checked, props.defaultChecked, props.onChange);
+  const [isIndeterminate, setIsIndeterminate] = useControllableValue(props.indeterminate, props.defaultIndeterminate);
+
+  useFocusRects(rootRef);
+  useDebugWarning(props);
+  useComponentRef(props, isChecked, isIndeterminate, inputRef);
+
+  const onChange = (ev: React.ChangeEvent<HTMLElement>): void => {
+    if (!isIndeterminate) {
```
dzearing, line 40: onChange = (...) => {
setIsChecked(!isChecked);
setIsIndeterminate(false);
}

packages/react-next/src/components/Checkbox/useCheckbox.ts
```diff
@@ -0,0 +1,131 @@
...
+      });
+    }, []);
+  }
+}
+
+function useComponentRef(
+  props: ICheckboxProps,
+  isChecked: boolean | undefined,
+  isIndeterminate: boolean | undefined,
+  checkBox: React.RefObject<HTMLInputElement>,
```
dzearing, line 112: `checkbox`, lowercase B

packages/react-next/src/components/Checkbox/useCheckbox.ts
```diff
@@ -0,0 +1,131 @@
...
+      get checked() {
+        return !!isChecked;
+      },
+      get indeterminate() {
+        return !!isIndeterminate;
+      },
+      focus() {
+        if (checkBox.current) {
+          checkBox.current.focus();
+        }
```
dzearing, line 126:
```suggestion
  checkboxRef: React.RefObject<HTMLInputElement>,
) {
  React.useImperativeHandle(
    props.componentRef,
    () => ({
      get checked() {
        return !!isChecked;
      },
      get indeterminate() {
        return !!isIndeterminate;
      },
      focus() {
        checkboxRef.current?.focus();
```

packages/react-next/src/components/Checkbox/useCheckboxClasses.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const GlobalClassNames: Omit<ICheckboxClasses, 'input'> = {
+  root: 'ms-Checkbox',
+  container: 'ms-Checkbox-label',
+  checkbox: 'ms-Checkbox-checkbox',
+  checkmark: 'ms-Checkbox-checkmark',
+  text: 'ms-Checkbox-text',
+};
+
+const defaultClassResolver = createClassResolver(classes);
```
dzearing, line 15: If we *just* use `createClassResolver` without all this extra global classnames overhead, will that improve perf? (remove all the papercuts.)

packages/react-next/src/Styling/useGlobalClassNames.ts
dzearing: Can be validated in a separate pr

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
dzearing: Can be!

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
dzearing: I think it can be removedc.

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
xugao: yes that's the plan. i am planning to do that when @khmakoto 's state PR is in. we need to move adding this slot and related state out side of Base component

PR #13439, opened 2020-06-02
feat(Popup): add show/hide animations

mnajdova:
This PR is simply adding animations for showing/hiding popups. No customization are allowed at this moment, if we need to we can add them as a follow up


packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
```diff
@@ -540,7 +545,7 @@ const Popup: React.FC<PopupProps> &
   const element = (
     <>
       {triggerElement}
-      {open && (inline ? contentElement : <PortalInner mountNode={mountNode}>{contentElement}</PortalInner>)}
+      {inline ? contentElement : <PortalInner mountNode={mountNode}>{contentElement}</PortalInner>}
```
mnajdova, line 548: The Animation component is responsible for mounting/removing the element now, so no need for this check

mnajdova:
The `Animation` component currently is not using the `useStyles` hook, but it is using directly the `getStyles` function with all performance flags disabled, which is causing a perf issues (at least for second renderings). This PR will not be merged until that issue is solved.

packages/fluentui/e2e/server/app.tsx
```diff
@@ -22,7 +22,7 @@ const ContentList = () => (
 
 const App = () => (
   <BrowserRouter>
-    <Provider theme={teamsTheme}>
+    <Provider theme={teamsTheme} disableAnimations>
```
layershifter, line 25: Let's add a comment why it's done 🙏

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
```diff
@@ -384,17 +385,17 @@ const Popup: React.FC<PopupProps> &
     );
   };
 
-  const renderPopperChildren = ({ placement, scheduleUpdate }: PopperChildrenProps) => {
+  const renderPopperChildren = classes => ({ placement, scheduleUpdate }: PopperChildrenProps) => {
     const content = renderContent ? renderContent(scheduleUpdate) : props.content;
     const popupContent = Popup.Content.create(content || {}, {
       defaultProps: () =>
         getA11yProps('popup', {
-          ...getContentProps(),
```
layershifter, line 392: Was this removed intentionally?

packages/fluentui/react-northstar/test/__mocks__/react-transition-group.ts
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
+
+export * from 'react-transition-group';
+export const Transition = props => {
```
layershifter, line 4: Are there `TransitionProps` ? Can we use them?

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
mnajdova: No, good catch!

packages/fluentui/e2e/server/app.tsx
mnajdova: Added

PR #13440, opened 2020-06-02
Update just-scripts version to fix warning during sass build step

### LINKED ISSUE #13362: Get rid of autoprefixer/browserslist warning in sass build step
After @dzearing's recent changes to the sass build step we started getting this warning on every package's sass build:
```
■ started 'sass'
  Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.
  Using browsers option can cause errors. Browserslist config 
  can be used for Babel, Autoprefixer, postcss-normalize and other tools.
  If you really need to use option, rename it to overrideBrowserslist.
  Learn more at:
  https://github.com/browserslist/browserslist#readme
  https://twitter.com/browserslist
```
We should probably find a way to fix or suppress this.

jdhuntington:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #13362
- [X] Include a change request file using `$ yarn change` (na)

#### Description of changes

Bump version of just-scripts to include https://github.com/microsoft/just/pull/385

### DIFF
--- a/packages/fluentui/digest/package.json
+++ b/packages/fluentui/digest/package.json
@@ -18,7 +18,7 @@
     "test": "just-scripts test"
   },
   "dependencies": {
-    "just-scripts": "0.42.0",
+    "just-scripts": "0.43.0",
     "querystring": "^0.2.0",
     "webpack": "4.35.0"
   },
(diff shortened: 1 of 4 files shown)


ecraig12345:
Now you need change files.

ecraig12345:
Actually that's wrong, these packages don't use change files. Sorry!

ecraig12345:
Pushed an update to 0.43.0 to pick up https://github.com/microsoft/just/pull/390 with some new eslint options.

PR #13442, opened 2020-06-02
Pivot: Replace onKeyPress with onKeyDown and ariaLabel with aria-label

### LINKED ISSUE #13429: Investigate if Pivot should use onKeyDown instead of onKeyPress
From feedback on PR #13370, need to fix up key handling for Pivot

>> is it intentional to use onKeyPress instead of onKeyDown which is typically used for keyboard handling?

_Originally posted by @jurokapsiar in https://github.com/microsoft/fluentui/pull/13370_

behowell:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13429

#### Description of changes

Fixing PR feedback from @jurokapsiar on #13370:
* Replace the onKeyPress handler with onKeyDown
* Replace `ariaLabel` with `aria-label`

#### Focus areas to test

I've locally tested these affected scenarios
* Using the Enter key to switch tabs
* Setting ariaLabel on PivotItem

### DIFF
--- a/packages/react-next/src/components/Pivot/Pivot.base.tsx
+++ b/packages/react-next/src/components/Pivot/Pivot.base.tsx
@@ -106,7 +106,7 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef(
       );
     };
 
-    const onKeyPress = (itemKey: string, ev: React.KeyboardEvent<HTMLElement>): void => {
+    const onKeyDown = (itemKey: string, ev: React.KeyboardEvent<HTMLElement>): void => {
       if (ev.which === KeyCodes.enter) {
         ev.preventDefault();
         updateSelectedItem(itemKey);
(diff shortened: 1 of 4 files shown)


packages/react-next/src/components/Pivot/Pivot.base.tsx
```diff
@@ -106,7 +106,7 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef(
       );
     };
 
-    const onKeyPress = (itemKey: string, ev: React.KeyboardEvent<HTMLElement>): void => {
+    const onKeyDown = (itemKey: string, ev: React.KeyboardEvent<HTMLElement>): void => {
```
khmakoto, line 109: I wonder if this change, while not showing any unexpected things in our uses, could make some things stop working for partners.

xugao: looks like keypress is being deprecated in web standard. i can't think of how this being a risky change.

behowell: Also, this is listening to key events on an internal component inside Pivot. I don't know what a partner could be doing with it that this could break?

I'm going to resolve this, but let me know if you disagree @khmakoto

khmakoto: I think those two are very valid reasons for doing this. I'm fine leaving this as resolved, thanks!

PR #13443, opened 2020-06-02
createClassResolver: support resolving partial classes at runtime

xugao:
#### Pull request checklist

#### Description of changes
see use case in #13435

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-compose/src/createClassResolver.ts
+++ b/packages/react-compose/src/createClassResolver.ts
@@ -22,7 +22,7 @@ export const createClassResolver = (classes: ClassDictionary) => {
   // Everything in the function below will happen at runtime, so it's very critical that this
   // code is as minimal as possible.
   // tslint:disable-next-line:no-function-expression
-  return function classResolver(state: GenericDictionary): ClassDictionary {
+  return function classResolver(state: GenericDictionary, partialClasses?: ClassDictionary): ClassDictionary {
     const resolvedClasses: Record<string, string> = {};
 
     let modifierClasses = '';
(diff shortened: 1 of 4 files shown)


packages/react-compose/src/createClassResolver.ts
```diff
@@ -14,43 +23,74 @@ import { GenericDictionary, ClassDictionary } from './types';
...
-          state[enumName] === enumValue && modifiers.push(classValue);
-        } else {
-          state[key] && modifiers.push(classValue);
-        }
+export const createClassResolver = (staticClasses: ClassDictionary) => {
+  const cacheMap: RecursiveMemoNode = new Map();
+  let cacheMissCount = 0;
+  let disableCaching = false;
+
+  return (
```
dzearing, line 31: So, with static classes, if we know ahead of time the slots, we can actually pull out the deps automatically without them being provided. Basically every key in the class map that isn't a slot would be the key to depend on in `state`

dzearing: we'd just need to know slots at creation time.

khmakoto: @dzearing I'm guessing this is what you're doing in #13448.

xugao: I removed the memo changes from this PR for now

xugao:
thanks for reviewing @khmakoto ! David suggested a much better alternative: https://github.com/microsoft/fluentui/pull/13493
So im going to close this PR

PR #13446, opened 2020-06-03
FeaturesProvider for providing opt in experimental flags in a generalized way

phkuo:
#### Pull request checklist

#### Description of changes

Add the FeatureFlags feature, which allows users to get or set feature flags, used for enabling experimental blocks of code in shipped components. This will be used by partners to test new features before officially releasing the feature to all FluentUI consumers.


packages/utilities/src/FeatureFlags.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const _featureFlags: FeatureFlags = {};
+
+export const FeatureFlagsContext = React.createContext<FeatureFlags>(_featureFlags);
+export const getFeatureFlags = () => React.useContext(FeatureFlagsContext);
+
+/**
+ * This should only be used in a non-React environment. In React, use <FeatureFlagsContext.Provider> instead.
+ * This MUST called BEFORE the flags are consumed, because it does not trigger an update.
+ * Ideally it is called only once at the time an app boots.
```
dzearing, line 22: awesome

packages/utilities/src/FeatureFlags.ts
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+
+/**
+ * FeatureFlags allows us to ship features off by default, unless the consuming app turns on that feature. This is
+ * intended to let a partner test a given feature first before releasing it to the rest of FluentUI consumers.
+ * The flags are provided via React Context.
+ */
+
+type FeatureFlags = {
```
dzearing, line 9: This probably belongs in the suite package, not in utilities.

packages/utilities/src/FeatureFlags.ts
```diff
@@ -0,0 +1,30 @@
...
+  ModalAriaHidden?: boolean;
+  TestFeature?: boolean; // used for testing only, replace with a real feature in .test when a second feature is added
+};
+
+const _featureFlags: FeatureFlags = {};
+
+export const FeatureFlagsContext = React.createContext<FeatureFlags>(_featureFlags);
```
dzearing, line 16: This can be `React.createContext<Record<string, boolean>>`

In OUFR, you can export this:

```tsx
export interface FeatureFlagsProvider {
  /**
   * Comment here explaining feature, who added it, what scenarios need it.
   */
  modalAriaHidden?: boolean;
}

export const FeatureFlagsProvider = (props: React.PropsWithChildren<FeatureFlagsProviderProps>) => {
  return (
    <FeatureFlagsContext.Provider value={ rest }>{ children }</FeatureFlagsContext.Provider>
  );
};

This means customers simply use the provider:
```tsx
<FeatureFlagsProvider modalAriaHidden>
  <App/>
</FeatureFlagsProvider>
```

packages/react-features-provider/src/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  Object.keys(_features).forEach(key => delete _features[key]);
+
+  // tslint:disable-next-line:no-any
+  Object.keys(features).forEach(key => (_features[key] = (features as any)[key]));
+};
+
+/**
+ * Exposes a FeatOuresProvider component, which takes in the feature flags and provides them via context.
+ */
+export const FeaturesProvider = (props: React.PropsWithChildren<Features>) => {
```
ecraig12345, line 44: I wonder if it would be better to make the features type generic (rather than tied specifically to our repo), and then the version exported from OUFR can reference features specific to that package.

phkuo: What do you mean by making it generic?
If you're saying we have a set of Features for each package, David and I discussed that and decided against it. It would get too complicated/annoying to have to go through each of your dependencies and say MyFeatures implements Package1Features, Package2Features, etcetc. It would make a really weird webby shape of dependencies.

joschect:
I know I've mentioned this before but I'm still not convinced that this is the right move. It adds another API surface to manage which I suspect will cause problems in the future. We still don't have a process for how or when a killswitch will be removed once added. There

packages/react-features-provider/src/index.tsx
dzearing: It was my original proposal @ecraig12345.

It got a bit weird when we consider that suite packages export sub packages. So a subpackage exports some Features interface, and then the suite package has to export the union of the subpackage features in an uber features typing with a typesafe version of everything in the package.

jdhuntington:
> It adds another API surface to manage which I suspect will cause problems in the future.
👍  

This library looks great. However, I feel like it's an orthogonal concern to a component library and might better live in its own repo. The controls already have plenty of ways to expose functionality; it seems like integrating experimental features through props would be more testable/predictable in the long run.

dzearing:
@jdhuntington @josch the motivator here is how we ship behaviors that might be potentially risky.

The example is distributing `aria-hidden` attributes on sibling elements when modals are visible. The change should be relatively safe, but there are edge cases we're unsure will cause problems, and an opt in the behavior makes sense at least until we've vetted it in product.

How should we roll this out?

1. Use something like a FeaturesProvider to provide the experimental feature.
2. Expose an `enableAriaHiddenOnVisible` prop on `Modal`.
3. Ship it without any sort of opt in and try our best to know all use cases.

I do think Feature flags are often generators of code debt. So are experimental props on things, though maybe props are more local which can require prop drilling in nesting situations, which leads back to context or globals.

Are there any other approaches to consider?

jdhuntington:
> Expose an enableAriaHiddenOnVisible prop on Modal.

This would be my preferred approach. I know prop drilling becomes a concern, but there are lots of ways around that, especially as we look at integrating compose as a way to customize components.

dzearing:
@jdhuntington I'm not sure compose will help here;

Dialog renders Modal, so we'd have to add the prop to that as well, and anything else which renders Modal.

Is there something we'd do with compose to help here?

jdhuntington:
> Dialog renders Modal, so we'd have to add the prop to that as well, and anything else which renders Modal.

👍  From what I can tell, it's just those 2 places.

> What would we do with compose to help here?

If someone wanted to bake this in as a "global setting" for their application, compose would be a great way to plug in a value for the `enableAriaHiddenOnVisible` prop so it gets used in all locations throughout the themed uses of a modal (like in the cases where a 3rd party component is used and props aren't as easy to modify).

phkuo:
@joschect @jdhuntington _We can't keep dragging out this discussion._ We've already hashed out some of these points on Teams. I've had this workitem for _weeks_ and it was supposed to be due end of last month, and this is only part 1 of 3!!

I understand your concerns and I believe this is the best compromise.

Re: graduation/debt - yes there is debt and we don't have a clear policy in place. That's fine, we can develop the policy as we go. Graduation should be simple anyways, you just remove an `if` statement. You know where my office is (not that anyone's there at this time), so you can chase me down too.

Re: extra bytes - Yeah but any feature is extra bytes. This is absolutely worth it in terms of the control it gives us to react to live issues.

Re: extra API - any sort of feature similar to this will introduce an extra API. A hard requirement is that it can be done in one place; it's an absolute no-go if I have to go through every single dialog/modal in the product and opt them each into a feature one by one. So the logical alternative is to use some global flag. If I implement it on Modal, and then someone creates a new Modal popup component, then consumers would have to go hook that up too. Global flags makes a ton of sense.

joschect:
Is the plan to have killswitches be part of the fabric api forever? 

React allows injection of props and values and I think that the goal should be that you can inject your own killswitch api. Especially with the work that's coming down the pipe. I do understand that could take a while and this fix needs to go in now.

Could we maybe plan to have this api in place only for 2 releases? What other kinds of fixes do we see this being used for and how do we ensure they aren't too prevalent?

phkuo:
@joschect Yes, the plan would be to keep it around forever. That doesn't mean the interface grows forever, we can deprecate all retired killswitches every major release.

I definitely wouldn't be opposed to revisiting this in a major release or two if we can come up with a better solution to achieve the same goal.

My requirements are:
1) I can turn on/off a section of code somehow
2) I can flip that switch from a single place in my app

ecraig12345:
Regarding the questions a few comments back about how to handle props drilling--pretty sure you can handle this with Customizer today:
```tsx
<Customizer
  scopedSettings={{
    Modal: { enableAriaHiddenOnVisible: true }
  }}
/>
```

PR #13448, opened 2020-06-03
Perf: adjust `createClassResolver` to do more logic as compose time vs runtime

dzearing:
This change moves most of the class parsing logic to compose time. Experimenting with optimizations before we kick in memoization.

### DIFF
--- a/packages/react-button/src/components/Button/Button.scss
+++ b/packages/react-button/src/components/Button/Button.scss
@@ -120,7 +120,7 @@
   * 2. Map property definitions to variables using the format: {component}-{property}
   * 3. Avoid defining styles directly.
  **/
-  &.primary {
+  &._primary {
     --button-fill: var(--accent-fill);
     --button-fill-hovered: var(--accent-fill-hovered);
     --button-fill-pressed: var(--accent-fill-pressed);
@@ -136,17 +136,17 @@
     --button-icon-pressed: var(--accent-icon-pressed);
   }
 
-  &.iconOnly {
+  &._iconOnly {
     width: var(--button-height);
     --button-padding: 0;
   }
 
-  &.circular {
+  &._circular {
     --button-borderRadius: 50000px;
     --button-minWidth: var(--button-height);
   }
 
-  &.fluid {
+  &._fluid {
     width: 100%;
     max-width: 100%;
   }
@@ -174,7 +174,7 @@
     --button-height: var(--button-size-largest);
   }
 
-  &.disabled {
+  &._disabled {
     pointer-events: none;
     background-color: var(--button-fill-disabled);
     color: var(--button-text-disabled);
@@ -210,7 +210,7 @@
   margin-left: var(--button-contentGap);
 }
 
-.root:hovered {
+.root:hover {
   background-color: var(--button-fill-hovered, var(--button-fill));
   color: var(--button-text-hovered, var(--button-text));
   border-color: var(--button-borderColor-hovered, var(--button-borderColor));
@@ -235,7 +235,7 @@
   }
 }
 
-.root:focused::after {
+.root:focus::after {
   position: absolute;
   left: -1px;
   top: -1px;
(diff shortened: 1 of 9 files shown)


packages/react-button/src/components/Button/Button.scss
```diff
@@ -17,7 +17,7 @@
 
   --button-padding: 0 24px;
   --button-height: var(--button-size-regular);
-  --button-minWidth: 96px;
+  --button-minWidtjh: 96px;
```
jdhuntington, line 20: typo?

dzearing: fixed.

dzearing:
I see one more optimization to tweak

packages/react-compose/src/createClassResolver.ts
```diff
@@ -65,3 +66,59 @@ function addClassTo(slotProps: GenericDictionary, slotName: string, className?:
...
 }
+
+type ResolvedMap = {
+  slots: Record<string, string | string[]>;
+  modifiers: Record<string, string | string[]>;
+  enums: Record<string, Record<string, string | string[]>>;
+};
+
+/**
+ * Helper to take a css module map and translate it into { slots, modifiers,  enums } where
```
khmakoto, line 77:
```suggestion
 * Helper to take a css module map and translate it into { slots, modifiers, enums } where
```

packages/react-compose/src/createClassResolver.ts
```diff
@@ -65,3 +66,59 @@ function addClassTo(slotProps: GenericDictionary, slotName: string, className?:
...
+        addClassTo(enums[enumName], enumValue, classValue);
+      } else {
+        addClassTo(modifiers, key, classValue);
+      }
+    }
+  });
+
+  // Reduce map to strings
+  collapseArray(slots);
+  collapseArray(modifiers);
```
xugao, line 114: instead of building array then join. would it be better if we just start with string and append the classname in `addClassTo`

packages/react-compose/src/createClassResolver.ts
```diff
@@ -14,43 +14,44 @@ import { GenericDictionary, ClassDictionary } from './types';
...
 
-    // Convert the className arrays to strings.
-    Object.keys(classMap).forEach((key: string) => (classMap[key] = classMap[key].concat(modifiers).join(' ')));
-  }
+    for (const slotName of slotNames) {
+      resolvedClasses[slotName] = [
+        slotName === 'root' && state.className ? state.className : '',
+        slots[slotName] || '',
+        ...modifierClasses,
+        ...enumClasses,
```
khmakoto, line 47: So you're passing all modifier and enum classes to all the slots, correct?

packages/react-compose/etc/react-compose.api.md
```diff
@@ -73,7 +73,7 @@ export type ComposePreparedOptions<TProps = {}, TState = TProps> = {
 export type ComposeRenderFunction<TElementType extends React.ElementType = 'div', TProps = {}> = (props: TProps, ref: React.Ref<TElementType extends keyof HTMLElementTagNameMap ? HTMLElementTagNameMap[TElementType] : TElementType>, composeOptions: ComposePreparedOptions) => React.ReactElement | null;
 
 // @public
-export const createClassResolver: (classes: Record<string, string>) => (state: Record<string, any>, slots: Record<string, any>) => Record<string, string>;
+export const createClassResolver: (classes: Record<string, string>, slotNames: string[]) => (state: Record<string, any>) => Record<string, string>;
```
xugao, line 76: need to update

packages/react-compose/src/createClassResolver.ts
dzearing: yes

dzearing: although we'd get better perf only leaving them at the root

PR #13452, opened 2020-06-03
Bump beachball version and clean up from publish failure

ecraig12345:
Bump beachball to latest version with better logging of errors for publish retries.

Sync OUFR version with latest published version after publishing problems.

Delete a random change file for a private package (not needed).

### DIFF
--- a/packages/office-ui-fabric-react/package.json
+++ b/packages/office-ui-fabric-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "office-ui-fabric-react",
-  "version": "7.117.1",
+  "version": "7.117.2",
   "description": "Reusable React components for building experiences for Microsoft 365.",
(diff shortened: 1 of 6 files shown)


packages/office-ui-fabric-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "office-ui-fabric-react",
-  "version": "7.117.1",
+  "version": "7.117.2",
```
xugao, line 3: did you get this updated by running `beachball sync`? i forgot if we also need to sync the places where have oufr as dependency

ecraig12345: Yes that was `beachball sync`. It didn't update anything else, but I think that should be okay in this case.

PR #13474, opened 2020-06-04
Perf: remove getNativeProps usage in mergeProps

dzearing:
This is a first stab at moving closer to the v0 approach of mixing unhandled props into root:

Rather than allow list, which is bulky and long and slow in lookups, mix anything not included in `handledProps`, in addition to things we know aren't valid.

This means that components would need to provide `handledProps` on their config.


packages/react-button/src/components/Button/ButtonBase.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from 'react';
 import { ButtonProps } from './Button.types';
-import { compose } from '@fluentui/react-compose';
+import { compose, ComposePreparedOptions } from '@fluentui/react-compose';
 import { useButton } from './useButton';
 
 export const ButtonBase = compose<'button', ButtonProps, ButtonProps, {}, {}>(
-  (props, ref, composeOptions) => {
+  (props: ButtonProps, ref: React.Ref<HTMLElement>, composeOptions: ComposePreparedOptions<ButtonProps>) => {
```
assuncaocharles, line 7: Should it be `HTMLButtonElement`?

```suggestion
  (props: ButtonProps, ref: React.Ref<HTMLButtonElement>, composeOptions: ComposePreparedOptions<ButtonProps>) => {
```

packages/react-compose/src/types.ts
```diff
@@ -99,7 +99,7 @@ export type ComposePreparedOptions<TProps = {}, TState = TProps> = {
   mapPropsToStylesPropsChain: ((props: object) => object)[];
   render: ComposeRenderFunction;
 
-  handledProps: (keyof TProps)[];
+  handledProps: string[];
```
xugao, line 102: why do you change the typings? original looks right to me

packages/react-compose/src/types.ts
```diff
@@ -62,7 +62,7 @@ export type ComposeOptions<TInputProps = {}, TInputStylesProps = {}, TParentProp
 
   mapPropsToStylesProps?: (props: TParentStylesProps & TInputProps) => TInputStylesProps;
 
-  handledProps?: (keyof TInputProps | 'as')[];
+  handledProps?: (keyof (TInputProps & { as: React.ElementType }))[];
```
xugao, line 65:
```suggestion
  handledProps?: (keyof TInputProps | 'as')[];
```

packages/react-compose/src/types.ts
```diff
@@ -99,7 +99,7 @@ export type ComposePreparedOptions<TProps = {}, TState = TProps> = {
   mapPropsToStylesPropsChain: ((props: object) => object)[];
   render: ComposeRenderFunction;
 
-  handledProps: (keyof TProps)[];
+  handledProps: (keyof (TProps & { as: React.ElementType }))[];
```
xugao, line 102:
```suggestion
  handledProps?: (keyof TInputProps | 'as')[];
```

PR #13478, opened 2020-06-04
Changing out cssnano for postcss-clean to address scss minification bugs

dzearing:
We're seeing that cssnano is stripping out load-themed-styles tokens. We should be using `postcss-clean` instead, which doesn't have this issue.

Removing cssnano first.


dzearing:
Alright, let's let this run, we should have minification back on with less breakage.

dzearing:
updated PR, waiting for new bundle numbers

PR #13480, opened 2020-06-04
Picker/PeoplePicker: Fix issue where default value wouldn't trigger suggestions

### LINKED ISSUE #12612: PeoplePicker: Not resolving suggestions on initial run when seeded with suggestion text
### Environment Information

- **Package version(s)**: "office-ui-fabric-react": "^7.70.0",
- **Browser and OS versions**: Microsoft Edge Version 83.0.474.0 (Official build) dev (64-bit)

We are using PeoplePicker component via `NormalPeoplePickerBase`. With our scenario, we want to pre-populate the picker with a text string, using the `inputProps.defaultVisibleValue` parameter. When doing this, our `onResolveSuggestions` callback is not called until a key press happens. If the PeoplePicker is seeded with an initial value, I would expect `onResolveSuggestions` to be called on initial use (similar to how `onEmptyResolveSuggestions` is called if the initial configuration is empty). Furthermore, `onEmptyResolveSuggestions` is not called if `inputProps.defaultVisibleValue` is set.

Note -- I'm not quite sure if this is a bug, or a feature request.

#### Actual behavior:

With PeoplePicker component,  `onResolveSuggestions` callback is not called when an initial value is provided using the `inputProps.defaultVisibleValue` parameter.

#### Expected behavior:

With PeoplePicker component,  `onResolveSuggestions` callback should be called initially when an initial value is provided using the `inputProps.defaultVisibleValue` parameter.

Are you willing to submit a PR to fix? Maybe, depending on the nature of the fix

Requested priority: Normal (I'm not sure what these priorities are defined as, but this is blocking our team's feature)

joschect:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #12612

#### Description of changes

Original Behavior: If the picker's input contained some default value via `inputProps: {defaultVisibleValue: 'foo'}}` when the picker received focus, it would not show suggestions.

The fix: The picker now attempts to callback to get suggestions if it currently contains a text value and does not have any current suggestions.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
@@ -390,7 +390,7 @@ describe('BasePicker', () => {
 
     jest.runAllTimers();
 
-    expect(getSuggestions).toBeTruthy();
+    expect(getSuggestions(document)).toBeTruthy();
   });
 
   it('Opens menu when input refocused after search has happened', () => {
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
@@ -393,40 +393,6 @@ describe('BasePicker', () => {
     expect(getSuggestions).toBeTruthy();
   });
 
-  it('Opens menu on click when suggestions has been dismissed', () => {
-    jest.useFakeTimers();
-    document.body.appendChild(root);
-
-    const picker = React.createRef<IBasePicker<ISimple>>();
-
-    ReactDOM.render(
-      <BasePickerWithType
-        onResolveSuggestions={onResolveSuggestions}
-        onRenderItem={onRenderItem}
-        onRenderSuggestionsItem={basicSuggestionRenderer}
-        componentRef={picker}
-      />,
-      root,
-    );
-
-    const input = document.querySelector('.ms-BasePicker-input') as HTMLInputElement;
-    input.focus();
-    input.value = 'asdff';
-    ReactTestUtils.Simulate.input(input);
-    jest.runAllTimers();
-
-    expect(getSuggestions(document)).toBeTruthy();
-
-    ReactTestUtils.Simulate.keyDown(input, { which: 27 });
-
-    expect(getSuggestions(document)).toBeFalsy();
-    ReactTestUtils.Simulate.click(input, { button: 0 });
-
-    jest.runAllTimers();
-
-    expect(getSuggestions(document)).toBeTruthy();
-  });
-
   it('Opens menu when input refocused after search has happened', () => {
     jest.useFakeTimers();
     document.body.appendChild(root);
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
@@ -52,6 +52,7 @@ describe('BasePicker', () => {
 
   afterEach(() => {
     ReactDOM.unmountComponentAtNode(root);
+    document.body.textContent = '';
   });
   const BasePickerWithType = BasePicker as new (props: IBasePickerProps<ISimple>) => BasePicker<
     ISimple,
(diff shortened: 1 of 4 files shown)


packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
```diff
@@ -328,4 +329,183 @@ describe('BasePicker', () => {
...
+    input.focus();
+    input.value = 'asdff';
+    ReactTestUtils.Simulate.input(input);
+    jest.runAllTimers();
+
+    const suggestions = document.querySelector('.ms-Suggestions') as HTMLElement;
+    expect(suggestions).toBeTruthy();
+
+    ReactTestUtils.Simulate.keyDown(input, { which: 27 });
+    const againSugs = document.querySelector('.ms-Suggestions') as HTMLElement;
```
jdhuntington, line 359: I'm very cautious about DRYing tests, but in this case it feels like it might make sense to extract this to a 1-liner called `getSuggestionsElement`, since it happens a lot and involves an ugly cast.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -430,14 +428,28 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
...
     const emptyResolveSuggestions = this.props.onEmptyResolveSuggestions
       ? this.props.onEmptyResolveSuggestions
       : // tslint:disable-next-line:deprecation
         this.props.onEmptyInputFocus;
 
-    const suggestions = emptyResolveSuggestions!(this.state.items);
-    this.updateSuggestionsList(suggestions);
+    // Only attempt to resolve suggestions if it exists
+    if (emptyResolveSuggestions) {
+      const suggestions = emptyResolveSuggestions!(this.state.items);
```
jdhuntington, line 443: The `!` isn't needed anymore since TS can infer it via the `if` statement.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -983,6 +960,28 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
       return '';
     }
   }
+
+  /**
+   * This should be called when the user does something other than use text entry to trigger suggestions.
+   *
+   */
+  private _userTriggeredSuggestions = () => {
```
jdhuntington, line 968: 👍  for moving this. Also a valuable refactor target in the future as there are tons of paths through these branches.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -430,14 +428,28 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
...
 
-    const suggestions = emptyResolveSuggestions!(this.state.items);
-    this.updateSuggestionsList(suggestions);
+    // Only attempt to resolve suggestions if it exists
+    if (emptyResolveSuggestions) {
+      const suggestions = emptyResolveSuggestions!(this.state.items);
+
+      this.updateSuggestionsList(suggestions);
+
+      this.setState({
```
jdhuntington, line 447: This seems to be new logic and not just a refactor. Intentional?

joschect: Correct! This fixes an issue where a default value in the picker wouldn't have triggered a resolve suggestions.

dzearing:
can you fill in a description of what the behavior was before/after? how do we test?

packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
```diff
@@ -319,13 +321,179 @@ describe('BasePicker', () => {
...
+    jest.runAllTimers();
+
+    expect(getSuggestions(document)).toBeTruthy();
+
+    ReactTestUtils.Simulate.keyDown(input, { which: 27 });
+
+    expect(getSuggestions(document)).toBeFalsy();
+  });
+
+  it('Opens menu on click when suggestions has been dismissed', () => {
```
khmakoto, line 362: nit:

```suggestion
  it('Opens menu on click when suggestions have been dismissed', () => {
```

packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
```diff
@@ -319,13 +321,179 @@ describe('BasePicker', () => {
...
+    expect(getSuggestions(document)).toBeTruthy();
+
+    ReactTestUtils.Simulate.keyDown(input, { which: 27 });
+
+    expect(getSuggestions(document)).toBeFalsy();
+    ReactTestUtils.Simulate.click(input, { button: 0 });
+
+    jest.runAllTimers();
+
+    expect(getSuggestions).toBeTruthy();
```
khmakoto, line 393:
```suggestion
    expect(getSuggestions(document)).toBeTruthy();
```

packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
```diff
@@ -319,13 +321,179 @@ describe('BasePicker', () => {
...
+  });
+
+  it('Opens menu on click when suggestions has been dismissed', () => {
```
khmakoto, line 396: Is this the same test as the one above?

packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
```diff
@@ -319,13 +321,179 @@ describe('BasePicker', () => {
...
+    expect(getSuggestions(document)).toBeFalsy();
+
+    jest.runAllTimers();
+    input.focus();
+    jest.runAllTimers();
+
+    expect(getSuggestions(document)).toBeTruthy();
+  });
+
+  it('Opens calls onResolveSuggestions if it currently doesnt have suggestions', () => {
```
khmakoto, line 469: nit:
```suggestion
  it('Opens calls onResolveSuggestions if it currently doesn't have suggestions', () => {
```

packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
```diff
@@ -43,6 +43,10 @@ const basicSuggestionRenderer = (props: ISimple) => {
   return <div> {props.name} </div>;
 };
 
+const getSuggestions = (root: HTMLElement | Document) => {
+  return document.querySelector<HTMLElement>('.ms-Suggestions');
```
dzearing, line 47: you should call `getDocument` from utilities with a valid html element. Otherwise you may get the wrong document in child window rendering scenarios.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
```diff
@@ -52,6 +56,7 @@ describe('BasePicker', () => {
 
   afterEach(() => {
     ReactDOM.unmountComponentAtNode(root);
+    document.body.textContent = '';
```
dzearing, line 59: getDocument

packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
joschect: It sure is! good Catch.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -597,29 +591,13 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
    * without shifting focus.
    */
   protected onClick = (ev: React.MouseEvent<HTMLInputElement>): void => {
-    const input = this.input.current ? this.input.current.value : '';
-
     if (this.props.inputProps !== undefined && this.props.inputProps.onClick !== undefined) {
       this.props.inputProps.onClick(ev);
     }
```
dzearing, line 596:
```suggestion
    this.props.inputProps?.onClick?(ev);
```
I think that would work :)

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -535,25 +546,8 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
...
-        this.setState({
-          isMostRecentlyUsedVisible: false,
-          suggestionsVisible: true,
-        });
-      }
+      this._userTriggeredSuggestions();
+
       if (this.props.inputProps && this.props.inputProps.onFocus) {
         this.props.inputProps.onFocus(ev as React.FocusEvent<HTMLInputElement>);
       }
```
dzearing, line 553:
```suggestion
      this.props.inputProps?.onFocus?(ev as React.FocusEvent<HTMLInputElement>);
```

packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
joschect: I'm going to leave this as is for right now. It needs an escape string to make the apostrophe work.

khmakoto: You could always use `does not` instead of `doesn't`.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
joschect: Did we get these fixed? I think it was an issue with babel if I remember correctly.

PR #13482, opened 2020-06-04
Perf-test: support testing re-renders

xugao:
#### Description of changes
**This PR includes:**
- Add support to "Fabric" perf-test to test virtual re-renders. That is, we re-render the component by the number of defined iterations with exact same set of props with no layout/styling changes.
- ONLY enabled re-render tests for components we are currently working on for V8 as a starter.
- Upgrade `flamegrill` to include fix https://github.com/microsoft/flamegrill/pull/17. In this change I removed logic which sets no timeout for all tests. As a result, I need to `setDefaultTimeout` in FluentUI perf-test. 

**Note**: 
1. I understand we have 2 perf-test scripts in the repo. If I have time i would have converged them then add this feature. But I am having trouble find that time now. I still want to add the support since we are actively developing V8 and I want perf test coverage for re-render scenarios
2. master tests do not have latest `index.scenarios.ts` code so master will not work for re-render cases. I need to stage this change to 2 steps:
    - revert tests to only test mount and merge this change
    - merge another change that re-enable the tests for re-render

This iteration of perf results contain the re-rendering tests:
image

### DIFF
--- a/apps/perf-test/src/index.scenarios.tsx
+++ b/apps/perf-test/src/index.scenarios.tsx
@@ -10,6 +10,10 @@ initializeIcons();
 const div = document.createElement('div');
 document.body.appendChild(div);
 
+const renderFinishedMarkerId = 'render-done';
+const renderFinishedMarker = document.createElement('div');
+renderFinishedMarker.id = renderFinishedMarkerId;
+
 // TODO: could default to displaying list of scenarios if param is not provided.
 const defaultScenarioName = Object.keys(scenarios)[0];
 const defaultIterations = 10;
@@ -17,11 +21,46 @@ const defaultIterations = 10;
 const queryParams = qs.parse(window.location.search.substring(1));
 const iterations = queryParams.iterations ? parseInt(queryParams.iterations as string, 10) : defaultIterations;
 const scenario = queryParams.scenario ? (queryParams.scenario as string) : defaultScenarioName;
+const renderType = queryParams.renderType;
 
-// TODO: This seems to increase React (unstable_runWithPriority) render consumption from 4% to 72%!
-// const ScenarioContent = Array.from({ length: iterations }, () => scenarios[scenario]);
+const PerfTestScenario = scenarios[scenario];
 
-// TODO: Using React Fragments increases React (unstable_runWithPriority) render consumption from 4% to 26%.
-// It'd be interesting to root cause why at some point.
-// ReactDOM.render(<>{Array.from({ length: iterations }, () => (scenarios[scenario]))}</>, div);
-ReactDOM.render(<div>{Array.from({ length: iterations }, () => scenarios[scenario])}</div>, div);
+if (renderType === 'rerender') {
+  const PerfTestScenarioWrapper: React.FunctionComponent = () => {
+    const [renderCount, setRenderCounter] = React.useState(0);
+    React.useEffect(() => {
+      if (renderCount < iterations) {
+        setRenderCounter(renderCount + 1);
+      }
+    }, [renderCount]);
+
+    if (renderCount < iterations) {
+      return <PerfTestScenario />;
+    } else {
+      return (
+        <div>
+          <PerfTestScenario />
+          <div id="render-done" />
+        </div>
+      );
+    }
+  };
+
+  ReactDOM.render(<PerfTestScenarioWrapper />, div);
+} else {
+  // TODO: This seems to increase React (unstable_runWithPriority) render consumption from 4% to 72%!
+  // const ScenarioContent = Array.from({ length: iterations }, () => scenarios[scenario]);
+
+  // TODO: Using React Fragments increases React (unstable_runWithPriority) render consumption from 4% to 26%.
+  // It'd be interesting to root cause why at some point.
+  // ReactDOM.render(<>{Array.from({ length: iterations }, () => (scenarios[scenario]))}</>, div);
+  ReactDOM.render(
+    <div>
+      {Array.from({ length: iterations }, () => (
+        <PerfTestScenario />
+      ))}
+    </div>,
+    div,
+    () => div.appendChild(renderFinishedMarker),
+  );
+}
--- a/apps/perf-test/src/index.scenarios.tsx
+++ b/apps/perf-test/src/index.scenarios.tsx
@@ -25,7 +25,7 @@ const renderType = queryParams.renderType;
 
 const PerfTestScenario = scenarios[scenario];
 
-if (renderType === 'VirtualRerender') {
+if (renderType === 'virtual-renderer') {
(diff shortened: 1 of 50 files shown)


dzearing:
awesome!

apps/perf-test/tasks/perf-test.ts
```diff
@@ -233,19 +235,24 @@ module.exports = async function getPerfRegressions() {
...
-      iterations,
-    };
+    const renderTypes = scenarioRenderTypes[scenarioName] || DefaultRenderTypes;
+
+    renderTypes.forEach(renderType => {
+      const scenarioKey = `${scenarioName}-${renderType}`;
+      const testUrlParams = `?scenario=${scenarioName}&iterations=${iterations}&renderType=${renderType}`;
+
+      scenarios[scenarioKey] = {
+        baseline: `${urlForDeploy}${testUrlParams}`, // TODO: revert before merging
```
xugao, line 245: This is temporary and will not merge. I did this because master tests do not have latest perf-test script and will not work for re-render cases. Once this change is ready to merge, I need to stage this change to 2 steps:
1. revert tests to only test `mount` and merge this change
2. merge another change that re-enable the tests for `re-render`

apps/perf-test/src/index.scenarios.tsx
```diff
@@ -10,18 +10,57 @@ initializeIcons();
...
+const PerfTestScenario = scenarios[scenario];
+
+if (renderType === 'rerender') {
+  const PerfTestScenarioWrapper: React.FunctionComponent = () => {
+    const [renderCount, setRenderCounter] = React.useState(0);
+    React.useEffect(() => {
+      if (renderCount < iterations) {
+        setRenderCounter(renderCount + 1);
+      }
+    }, [renderCount]);
```
xugao, line 35: this setup code to trigger re-render seems little costly and it reflects in the flame chart

JasonGore: This is only measuring React virtual rendering time with no changes to the actual DOM hierarchy, correct? In other words, this doesn't capture layout and styling time, cost to update browser DOM, etc.

However, when we do a `mount` test, it does wait for the entire DOM hierarchy to be populated by waiting for the page load event. I think this is an important difference to call out.

If we wanted to measure the cost of layout and actual DOM updates, I could see a similar approach to watching for the `render-done` marker with a counter perhaps, and then rendering again. Although it seems a bit hacky and wonder if there are better approaches.

JasonGore: We also chatted about this but just to document: The fluentui `perf-test` is a "newer" version of this file, and also uses `digest` to create a bundle of test scenarios. I always thought it'd be relatively easy to change fabric to use fluentui's `perf-test` package as it has been.

However, now that improvements are also being made to this file, it will be harder to converge the perf-test packages. We should also make sure these changes will make sense in the context of fluentui `perf-test` and `digest`.

dzearing:
It would be nice to change this:

image

To this:

| Scenario | Mount | Re-render | Status |
|-|-|-|-|
| ButtonNext | 1234 -> 1100 (-20%) |  1234 -> 1100 (-20%) |  :white_check_mark: |

Is that possible? It is already a little noisy and is getting more noisy

apps/perf-test/src/index.scenarios.tsx
xugao: as discussed offline. I now calling the new render type "virtual re-render". also added comments in `scenarioRenderTypes` to clarify when such test is most useful.

I complete agree with the concern on having 2 perf-test scripts, I would love to make progress on that when i find chance. Right now given my priorities, I don't have time to do that ATM.

JasonGore:
There's a balance between data given and readability for sure. The problem is with a 10% variance with no code difference, it's hard to say the difference is 20%. It could actually be 10-30%. The checkmarks also seem to imply that there are no issues when there could be. The thresholds are arbitrary now. And the wording identifies that by saying "Potential regression" rather than "Actual regression" and failing PRs.

JasonGore:
The difficulty of thresholds also compound with variance. If you decide to set your threshold at 20%, for example, the exact same code changes could "pass" when 10% on one run and "fail" at 30% on another.

apps/perf-test/tasks/perf-test.ts
```diff
@@ -264,7 +271,18 @@ module.exports = async function getPerfRegressions() {
...
   /** @type {ScenarioConfig} */
-  const scenarioConfig = { outDir, tempDir };
+  const scenarioConfig = {
+    outDir,
+    tempDir,
+    pageActions: async (page, options) => {
+      page.setDefaultTimeout(60000); // set 60s timeout, default is 30s.
+
+      await page.goto(options.url);
+      // TODO: uncomment to enable re-render tests
```
JasonGore, line 281: Hmm already seems uncommented?

apps/perf-test/src/scenarioRenderTypes.js
```diff
@@ -0,0 +1,22 @@
...
+ * you want their render types to differ from the default (mount only).
+ *
+ * Note:
+ * You should not need to have re-render tests in most cases because mount test provides enough coverage.
+ * It is mostly usefual for cases where component has memoization logics. And in case of re-rendering,
+ * memoization logic help avoid certain code paths.
+ */
+
+const AllRenderTypes = ['Mount', 'VirtualRerender'];
+const DefaultRenderTypes = ['Mount'];
```
JasonGore, line 12: nit: The casing is weird for string types, I think it was better before. `mount` and `virtual-rerender` would be fine and consistent.

JasonGore: I would also advise making a `RenderType`, like `type RenderType = 'mount' | 'virtual-rerender';`. `AllRenderTypes` and `DefaultRenderTypes` would then be `RenderType[]`. The definition for `RenderType` is where I thought we would have an explanation of what `virtual-rerender` is and means.

apps/perf-test/tasks/perf-test.ts
```diff
@@ -341,10 +360,12 @@ function createScenarioTable(scenarioSettings, testResults, showAll) {
     resultsToDisplay
       .map(key => {
         const testResult = testResults[key];
-        const { iterations } = scenarioSettings[key] || {};
+        const { scenarioName, iterations, renderType } = scenarioSettings[key] || {};
+        const renderTypeFriendlyName = renderType === 'VirtualRerender' ? 'Virtual re-render' : renderType;
```
JasonGore, line 364: nit: I think the casing and `re-render` dashing is awkward here too. I think it's ok to just reflect `mount` and `virtual-rerender` in the report. Casing implies a unique object rather than a type, so I think all lower casing is actually more clear for the render type.

PR #13484, opened 2020-06-05
fix(DocSite): Fix IE11 issues

assuncaocharles:
#### Pull request checklist

#### Description of changes

The current [doc site](https://fluentsite.z22.web.core.windows.net/) it's completely not working on IE11 and this is because of [every grid child element need to be positioned by hand using -ms-grid-row and -ms-grid-column](https://stackoverflow.com/questions/52162463/css-grid-ie-11-overlap) to avoid overlap between rows. 

Screenshot 2020-06-05 at 17 08 35

This PR fixes that by replacing the Grid usage for Flex and setting styles to avoid big changes. 

#### Focus areas to test

(optional)


packages/fluentui/docs/src/components/ComponentPlayground/ComponentPlaygroundTemplate.tsx
```diff
@@ -19,43 +19,45 @@ const NoopKnobProvider: React.FunctionComponent = props => {
 };
 
 const ComponentPlaygroundTemplate: React.FunctionComponent<ComponentPlaygroundTemplateProps> = props => (
-  <Grid
-    styles={{
-      gridTemplateColumns: '1fr 300px',
-      msGridColumns: '1fr 1rem 300px',
```
layershifter, line 25: Ohoh, this breaks layout

image


vs 

image

packages/fluentui/docs/src/components/ComponentDoc/ExampleSection.tsx
```diff
@@ -18,15 +18,9 @@ const ExampleSection: React.FC<ExampleSectionProps> = ({ title, children }) => (
     <Header as="h2" align="center" styles={headerStyle} className="no-anchor">
       {title}
     </Header>
-    <Grid
-      variables={{ gridGap: '2rem' }}
-      styles={{
-        gridTemplateColumns: 'minmax(550px, 1fr)',
```
layershifter, line 24: And this breaks also:

image

image

assuncaocharles: The problem seems to be an issue with `Toolbar` component when inside a container with `display: flex`. To temporary fix it I changed the example itself, but I guess it is a toolbar issue. Right @pompomon @layershifter @jurokapsiar ?

assuncaocharles: Fixed as agreed stretching the container for now, but I think we might still wanna to report the issue for the toolbar?

packages/fluentui/docs/src/components/ComponentPlayground/ComponentPlaygroundTemplate.tsx
assuncaocharles: Fixed in the best way that keeps the same for chrome and makes it usable for IE

packages/fluentui/docs/src/components/ComponentDoc/ExampleSection.tsx
layershifter: > Fixed as agreed stretching the container for now, but I think we might still wanna to report the issue for the toolbar?

Yup, let's do it

PR #13486, opened 2020-06-05
fix(dropdown): fix display custom `clearIndicator` prop

kolaps33:
#### Description of changes

(give an overview)

#### Focus areas to test
image


packages/fluentui/react-northstar/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -36,22 +37,30 @@ const getWidth = (p: DropdownPropsAndState, v: DropdownVariables): string => {
   return v.width;
 };
 
+const isEmpty = prop => {
+  return (
+    typeof prop === 'object' &&
+    //@ts-ignore
```
assuncaocharles, line 43: Is this `ts-ignore` intent? 

I think the non-null assertion is miss used here

packages/fluentui/react-northstar/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -36,22 +37,30 @@ const getWidth = (p: DropdownPropsAndState, v: DropdownVariables): string => {
...
+    !prop.props &&
+    !_.get(prop, 'children') &&
+    !_.get(prop, 'content')
```
assuncaocharles, line 46:
```suggestion
    prop.props &&
    _.get(prop, 'children') &&
    _.get(prop, 'content')
```

kolaps33: when I tried to change it the suggestion then "default indicator" was not displayed:
image

because what we want to reach here is:
 if it is object verify it doesn't 
has any 'prop.props' what is this example: `<div>clear<div>`
has any children what is this example: `{{ children: 'clear' }}`
has any content what is this example: `{{ content: 'clear' }}`

Basically your proposal will never has "true" value. I will rather stick with our condition.

packages/fluentui/react-northstar/src/themes/teams/components/Dropdown/dropdownStyles.ts
kolaps33: thank you this good catch I forgot to remove "ts-ignore", leftover...

PR #13493, opened 2020-06-05
compose: Update classes option to allow array

xugao:
#### Pull request checklist

#### Description of changes
Update compose classes option to be array

#### Focus areas to test

(optional)


packages/react-compose/src/types.ts
```diff
@@ -56,7 +56,7 @@ export type ComposeRenderFunction<TElementType extends React.ElementType = 'div'
 export type ComposeOptions<TInputProps = {}, TInputStylesProps = {}, TParentProps = {}, TParentStylesProps = {}> = {
   className?: string;
 
-  classes?: ClassDictionary | ClassFunction;
+  classes?: (ClassDictionary | ClassFunction)[];
```
dzearing, line 59: this should be either a single item or an array. most cases you just need one. i don't think it would be much cost to check if array;

PR #13497, opened 2020-06-06
Add re-render perf test for ThemeProvider

xugao:
#### Pull request checklist

#### Description of changes
Add re-render perf test for ThemeProvider

#### Focus areas to test

(optional)


dzearing:
This is closer to what I'd expect with ThemeProvider; initial render costs time, but subsequent renders are much cheaper.

PR #13506, opened 2020-06-08
fix(Components): IE11

assuncaocharles:
#### Pull request checklist

#### Description of changes

Fix the Following components to work on IE11:

* `Form`
* `Menu` ( vertical positioning )
* `Toolbar` ( menu with submenu )
* `PopupExamplePointing`
* `PopupCustomTargetExample`

We are forcing the following cast:

```typescript
style={{ msGridRow: 1, msGridColumn: 1 } as React.CSSProperties}
```

While the grid types are not added in the `csstype`: https://github.com/frenic/csstype/issues/58

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/themes/teams/components/Form/formStyles.ts
```diff
@@ -6,12 +6,13 @@ const formStyles: ComponentSlotStylesPrepared<FormStylesProps, any> = {
...
     width: '100%',
-    display: ['grid', '-ms-grid'],
-    gridTemplateColumns: '1fr',
-    msGridColumns: '1fr',
+    display: 'flex',
+    flexDirection: 'column',
     justifyContent: 'space-evenly',
-    gridGap: pxToRem(20),
     justifyItems: 'start',
+    '> *:not(:last-child)': {
```
dzearing, line 13: nit; if you're doing gap, and you render a single item, that should not have marginBottom, correct?

I usually do `:not(:first-child) { marginTop: ... }` for gaps.

assuncaocharles: Hmm good point, I wonder if we wouldn't always want that form have the `marginBottom` I think if we remove it we might get some regression, since what I was trying to do there was just achieve same result we have before but working on IE also.. I will apply your suggestion and see if screener catches something :)

packages/fluentui/docs/src/global.d.ts
```diff
@@ -25,3 +25,9 @@ declare interface Window {
   resetExternalLayout?: () => void;
   switchTheme?: (themeName: string) => void;
 }
+
+declare module 'csstype' {
+  interface Properties {
+    msGridRow?: any;
```
pompomon, line 31: Please could we be a bit more specific?
Would this
```suggestion
    msGridRow?: number | string;
```
cover all the cases?

packages/fluentui/docs/src/examples/components/Grid/Types/GridExample.shorthand.tsx
```diff
@@ -2,16 +2,66 @@ import * as React from 'react';
...
-  <Image key="elyse" fluid src="public/images/avatar/large/elyse.png" />,
-  <Image key="helen" fluid src="public/images/avatar/large/helen.jpg" />,
-  <Image key="jenny" fluid src="public/images/avatar/large/jenny.jpg" />,
-  <Image key="joe" fluid src="public/images/avatar/large/joe.jpg" />,
-  <Image key="justen" fluid src="public/images/avatar/large/justen.jpg" />,
+  <Image
+    key="ade"
+    fluid
+    src="public/images/avatar/large/ade.jpg"
+    style={{ msGridRow: 1, msGridColumn: 1 } as React.CSSProperties}
```
assuncaocharles, line 9: @layershifter are you agains this cast while `csstype`doesn't fix this in their part?

PR #13507, opened 2020-06-08
Adding unit tests for selecteditemslist and selectedpeoplelist

nebhatna:
#### Pull request checklist

#### Description of changes
Adding unit tests for selecteditemslist and selectedpeoplelist
Also addressed handling default selected items.

#### Focus areas to test

(optional)


packages/experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
```diff
@@ -1,16 +1,15 @@
...
   ref: React.Ref<ISelectedItemsList<TItem>>,
 ) => {
-  const [items, updateItems] = React.useState(props.selectedItems || []);
+  const [items, updateItems] = React.useState(props.selectedItems || props.defaultSelectedItems || []);
   const renderedItems = React.useMemo(() => items, [items]);
 
   React.useEffect(() => {
-    updateItems(props.selectedItems || []);
+    updateItems(props.selectedItems || props.defaultSelectedItems || []);
   }, [props.selectedItems]);
```
amyngu, line 13: Do we need to add defaultSelectedItems here as well? Is this how we should be using this prop, or should it only really set the internal items state at the first render?

PR #13509, opened 2020-06-08
Feat(Animation): improve styles perf

mnajdova:
This PR improved Animation perf, by adding a caching that can be reused between multiple instances of the component.

# BREAKING CHANGES
The PR removed the `styles` and `variables` props from the `Animation` component. However does were not taken into account previously when generating the styles, so technically this is just a clean up of the props.


packages/fluentui/react-northstar/src/components/Animation/useAnimationStyles.ts
```diff
@@ -0,0 +1,71 @@
...
+  // TODO: add check for cache enabled
+  if (!animationCache.has(theme)) {
+    animationCache.set(theme, {});
+  }
+
+  const cachePerTheme = animationCache.get(theme);
+  const cacheKey = JSON.stringify(animation);
+
+  if (cachePerTheme[cacheKey]) {
+    console.log('Returning from cache...');
```
mnajdova, line 41:
```suggestion
```

packages/fluentui/react-northstar/src/components/Animation/Animation.tsx
```diff
@@ -14,14 +11,15 @@ import * as React from 'react';
...
+import useAnimationStyles from './useAnimationStyles';
+
+export { animationClassName } from './useAnimationStyles';
 
 export type AnimationChildrenProp = (props: { classes: string }) => React.ReactNode;
 
-export interface AnimationProps
-  extends StyledComponentProps,
-    ChildrenComponentProps<AnimationChildrenProp | React.ReactChild> {
+export interface AnimationProps extends ChildrenComponentProps<AnimationChildrenProp | React.ReactChild> {
```
mnajdova, line 22: We were laying that we support styles and variables, so I removed those props. This will allow us for better performance and at least caching always when it is enabled.

packages/fluentui/react-northstar/src/components/Animation/Animation.tsx
```diff
@@ -14,14 +11,15 @@ import * as React from 'react';
 import { ThemeContext } from 'react-fela';
 import { Transition } from 'react-transition-group';
 
-import { childrenExist, StyledComponentProps, commonPropTypes, ChildrenComponentProps } from '../../utils';
+import { childrenExist, commonPropTypes, ChildrenComponentProps } from '../../utils';
 import { ComponentEventHandler, ProviderContextPrepared } from '../../types';
+import useAnimationStyles from './useAnimationStyles';
+
+export { animationClassName } from './useAnimationStyles';
```
layershifter, line 18: I would like to avoid reexport there, can we reexport it in `src/index.ts` instead?

packages/fluentui/react-northstar/src/components/Animation/useAnimationStyles.ts
```diff
@@ -0,0 +1,100 @@
...
+  classes: UseStylesResult['classes'];
+  animationDuration?: string;
+  animationDelay?: string;
+};
+
+const animationCache = new WeakMap<ThemePrepared, Record<string, UseAnimationStylesResult>>();
+export const animationClassName = 'ui-animation';
+
+const useAnimationStyles = (displayName: string, props: AnimationProps): UseAnimationStylesResult => {
+  const { theme, rtl, disableAnimations, renderer, performance } = React.useContext(ThemeContext);
```
layershifter, line 24:
```suggestion
  const { theme, rtl, disableAnimations, renderer, performance }: ProviderContextInput = React.useContext(ThemeContext);
```

packages/fluentui/react-northstar/src/components/Animation/useAnimationStyles.ts
```diff
@@ -0,0 +1,100 @@
...
+} from '@fluentui/react-bindings';
+import { ThemePrepared } from '@fluentui/styles';
+import * as React from 'react';
+// @ts-ignore
+import { ThemeContext } from 'react-fela';
+import * as _ from 'lodash';
+import { AnimationProps } from './Animation';
+
+type UseAnimationStylesResult = {
+  classes: UseStylesResult['classes'];
```
layershifter, line 15: We can result only `root` there, should we just return `className` to simplify?

packages/fluentui/react-northstar/src/components/Animation/useAnimationStyles.ts
layershifter: Let's add a type there

packages/fluentui/react-northstar/src/components/Animation/useAnimationStyles.ts
mnajdova: changed

packages/fluentui/react-northstar/src/components/Animation/Animation.tsx
mnajdova: Moved

PR #13517, opened 2020-06-08
react-icons: remove dead code and merge-styles

xugao:
#### Pull request checklist

#### Description of changes
 - remove dead code
 - replace `merge-styles` with `scss`: reduce bundle size and it's more aligned with our V8 styling approach
- css fixes

#### Focus areas to test

(optional)


packages/react-icons/src/utils/SvgIcon.scss
```diff
@@ -0,0 +1,5 @@
+.svg {
+  height: 100%;
+  width: 100%;
```
dzearing, line 3: width: auto

packages/react-icons/src/utils/createSvgIcon.ts
```diff
@@ -21,7 +21,7 @@ const createSvgIcon = <TProps = {}>({ svg, displayName }: SvgIconCreateFnParams<
       {
         ...containerProps,
         ...nativeProps,
-        className: css(classes.root, className),
+        className,
```
dzearing, line 24: this doesn't seem right. 

classes.root is no longer applied to the container. We should either remove the span if the concern is an extra element, and push root to the svg, or, keep pushing classes.root to the span.

xugao: there were no styles defined on root as i removed it: https://github.com/microsoft/fluentui/blob/4e058b89af037da8b421ecc16b628238de011107/packages/react-icons/src/utils/SvgIcon.styles.ts#L11

mnajdova: We will need to add styles on the root, we discussed with @dzearing about adding the following styles:
```
speak: none;
vertical-align: middle;
display: inline-block;
```

I just didn't manage to create the PR yet. So I would suggest keeping the `classes.root` there

xugao: Added root back and also added the css fixes

packages/react-icons/src/utils/SvgIcon.scss
```diff
@@ -0,0 +1,13 @@
+.root {
+  display: inline-block;
+  vertical-align: middle;
+  speak: none;
+  width: 1em;
+  height: 1em;
```
dzearing, line 6:
```
line-height: 1; 
```
Also unsure that width/height should be set. Why not just 16px if we're going to have a default?

packages/react-icons/src/utils/SvgIcon.scss
```diff
@@ -0,0 +1,13 @@
...
+}
+
+.svg {
+  height: 100%;
```
dzearing, line 10: is `auto` the default value for `width`? I think so, not sure tho.

packages/office-ui-fabric-react/src/components/Icon/examples/Icon.SvgFactory.Example.tsx
```diff
@@ -4,21 +4,17 @@ import { PrimaryButton } from 'office-ui-fabric-react/lib/Button';
...
-  },
   cell: {
     display: 'flex',
     flexDirection: 'column',
     alignItems: 'center',
     margin: '80px',
     float: 'left',
+    height: '50px',
+    width: '50px',
+  },
```
dzearing, line 15: 1. remove merge-styles dependency from package.json
2. Add load-themed-styles because it is (currently) used when we import scss files.

xugao: this is example code.

packages/react-icons/src/utils/SvgIcon.scss
xugao: `width` default value is `auto`: https://www.w3schools.com/cssref/pr_dim_width.asp

packages/react-icons/src/utils/SvgIcon.scss
xugao: setting w & h with `1em` allows user to control the icon side using `font-size`.

PR #13519, opened 2020-06-08
getId: no need to resetIds when Stylesheet resets

xugao:
#### Pull request checklist

#### Description of changes
I don't think there is a need to `resetIds` when Stylesheet resets. If so we should remove the logic so we don't take dependency on `Stylesheet` from `getId`.

#### Focus areas to test

(optional)


dzearing:
Please don't do this. It is necessary for SSR class name predictability.

SSR loads a component into memory. On each server render, it resets the stylesheet to regenerate styles. Resetting IDs is required to ensure each render renders the exact same content.

xugao:
@dzearing  ah, got it.

PR #13521, opened 2020-06-09
Compose: Adding replaceable state as part of compose

khmakoto:
#### Pull request checklist

#### Description of changes

This PR is a first stab at envisioning how adding a layer for a replaceable `state` would look like in compose and integrating it into the `Link` and `Toggle` components for testing. A `ToggleButton` is also created to vet these ideas.

#### Focus areas to test

(optional)


packages/react-button/src/components/ToggleButton/useToggleButton.ts
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import { useControllableValue } from '@uifabric/react-hooks';
+import { ToggleButtonProps } from './ToggleButton.types';
+
+/**
+ * The useToggleButton hook processes the Button component props and returns state.
+ * @param props
+ */
+export const useToggleButton = (props: ToggleButtonProps) => {
```
xugao, line 9: input props should be ButtonProps type. pls also add return type

khmakoto: Why should it be `ButtonProps`?

I'll add the return type.

packages/react-compose/src/types.ts
```diff
@@ -109,6 +125,9 @@ export type ComposePreparedOptions<TProps = {}, TState = TProps> = {
   slots: Record<string, React.ElementType> & { __self: React.ElementType };
   slotProps: ((props: TProps) => Record<string, object>)[];
 
+  // tslint:disable-next-line:no-any
+  state: (props: TState, options: ComposePreparedOptions) => any;
```
xugao, line 129: shouldn't the typing be something like:
`(props: TParentState, options: ComposePreparedOptions) => TInputState;`

packages/react-next/src/components/Toggle/ToggleBase.tsx
```diff
@@ -5,8 +5,9 @@ import { IToggleProps } from './Toggle.types';
 import { useToggle } from './useToggle';
 
 export const ToggleBase = compose<'div', IToggleProps, IToggleProps, {}, {}>(
```
xugao, line 7:
```suggestion
export const ToggleBase = compose<'div', IToggleProps, {}, IToggleProps, {}>(
```

packages/react-compose/src/types.ts
xugao: actually, don't we need to also give  inputProps?
so something like:
`(props: TInputProps, parentState: TParentState | undefined, options: ComposePreparedOptions) => TInputState;`

packages/react-button/src/components/Button/ButtonBase.tsx
```diff
@@ -1,11 +1,13 @@
...
+import { compose, mergeProps } from '@fluentui/react-compose';
 import { useButton } from './useButton';
 
 export const ButtonBase = compose<'button', ButtonProps, ButtonProps, {}, {}>(
-  (props: ButtonProps, ref: React.Ref<HTMLButtonElement>, composeOptions: ComposePreparedOptions<ButtonProps>) => {
-    const { slots, slotProps } = useButton(props, composeOptions);
+  (props, ref, options) => {
+    const { state } = options;
+    const { slots, slotProps } = mergeProps(state, options);
+    console.log(state, slots, slotProps);
```
xugao, line 10: remove

packages/react-button/src/components/ToggleButton/useToggleButton.ts
```diff
@@ -0,0 +1,30 @@
...
+ * The useToggleButton hook processes the Button component props and returns state.
+ * @param props
+ */
+export const useToggleButton: ComposeOptions<
+  ToggleButtonProps,
+  ToggleButtonProps,
+  ButtonProps,
+  ButtonProps
+  // tslint:disable-next-line:no-any
+>['state'] = (props): any => {
```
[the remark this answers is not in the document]
khmakoto: I struggle with what it should return because in theory it the returned state could be different than `TProps`.

PR #13528, opened 2020-06-09
fix: Remove named exports of components in react-northstar

miroslavstastny:
#### Pull request checklist

#### Description of changes

react-northstar components are wrapped with `withSafeTypeForAs` and exported as default export. Those are then re-exported as named in index.ts:
```jsx
export * from './components/Accordion/Accordion';
export { default as Accordion } from './components/Accordion/Accordion';
```

If the component is exported as named, it is reexported by the first line and the name collides with the re-export on the second line.

Components should not be directly exported, I removed the exports.

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Accordion/Accordion.tsx
+++ b/packages/fluentui/react-northstar/src/components/Accordion/Accordion.tsx
@@ -105,7 +105,7 @@ export const accordionSlotClassNames: AccordionSlotClassNames = {
   title: `${accordionClassName}__title`,
 };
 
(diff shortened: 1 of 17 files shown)


layershifter:
FYI: @assuncaocharles

dzearing:
Is there any reason to just not always use named exports? What is the benefit of using a default export on a sub file in a library, just to provide the name in another file?

miroslavstastny:
> Is there any reason to just not always use named exports?

I think there is no reason. Based of bundle size investigations @layershifter did we will switch to named exports in v0.

This is just a "hotfix" to avoid the duplicate exports

PR #13530, opened 2020-06-09
[WIP]feat(Datepicker): Extracting date helper functions to utils package and PoC for FluentUI v0

pompomon:
#### Pull request checklist

##### Description of changes
Reviving https://github.com/microsoft/fluentui/pull/12981 

It's the first part of the attempt to create a react-northstar DatePicker while re-using logic from the date-time package. PR has two parts: start creating a CalendarDayGrid (just the name of the weeks for now) and copying/moving logic from date-time to utilities in order to be re-used by both date-time and react-northstar.

##### Moving code
in packages/utilties I moved the following from date-time:

CalendarDayGrid.base methods: getWeeks, getDateRangeTypeToUse, getDayInfosInRangeOfDay, getBoundedDateRange, findAvailableDate, getIsAfterMaxDate, getIsBeforeMinDate, getIsRestrictedDate
Changed them from private methods to functions, some of the accepting state+props as first argument. Ideally, these methods should be returned by a custom hook which accepts state+props, but date-time components are class components, and even sharing these functions via custom hook + render prop is not possible, as getWeeks for instance is used in constructor / lifecycle methods.

Moved them to the utilities package in the date-time folder. Also moved there:

interfaces: ICalendarStrings, IDayInfo and partial interfaces for state and props of CalendarDayGrid (ICalendarDayGridCommonState, ICalendarDayGridCommonProps). Since I don't believe both react-northstar and date-time will have 100% the same interaces for state/props I decided to copy only the commonly used properties and extend these in both date-time and react-northstar.
values: DayPickerStrings
date and time utilities: from office-ui-fabric: DateMath, DateValues, TimeConstants. These were needed by the methods we want to re-use and I cannot import them from date-time, since this means importing office-ui-fabric into react-northstar and we don't want that.
Important: at the moment I did not remove any copied code (interfaces, date math/values), but I did replace private methods use in CalendarDayGrid.base with the newly re-usable functions. I chose for now not to remove the duplicated code as this PR is already large enough and I probably need to do additional import changes as a result of deleting the code (for instance all imports of date math, values and time constants need to be changed to point to utilities instead of office-ui-fabric).

##### Creating CalendarDayGrid in react-northstar
I did not push to create the whole component as I did not want to have more changes in this PR. Right now the react-northstar component is a 7-column Grid that has the week day initials added to it and days


packages/utilities/src/date-time/dateMath/DateMath.ts
```diff
@@ -0,0 +1,418 @@
+import { DayOfWeek, MonthOfYear, FirstWeekOfYear, DateRangeType } from '../dateValues/DateValues';
```
lorejoh12, line 1: So you're copy/pasting these and some other utilities, but aren't removing the previous versions yet- what's the end goal? Is office-ui-fabric-react and date-time packages going to reference the utilities package to get these values out?

The original intention of the date-time package was to have all the date related logic wrapped up into it, so that we didn't have it spread around to multiple packages. At the moment, date-time just relies on OUFR's date logic, but the long term goal is to move all the logic out of OUFR and into date-time. 

With this solution, we end up spreading date logic into the utilities package. Is north star trying to avoid a dependency on date-time package as well as the OUFR package? Or should we take another look at getting that logic out of OUFR entirely so you can pull in just date-time without pulling in all of OUFR as well?

packages/utilities/src/index.ts
```diff
@@ -68,6 +68,13 @@ export * from './styled';
 export * from './warn';
 export * from './ie11Detector';
 export * from './getPropsWithDefaults';
+export * from './date-time/components/Calendar.types';
+export * from './date-time/components/CalendarDayGrid';
```
lorejoh12, line 72: Is this the only component we export from utilities? Just seems odd that we're pushing components into here at all, when really they should ideally live in date-time and you can reference them from there.

pompomon:
Closing PR, would re-thing the approach

PR #13537, opened 2020-06-09
ensure theme.ts is always included when importing from styling package

xugao:
#### Pull request checklist

#### Description of changes
Fix regression from #13422.
This fix ensure when styling package is being imported (no matter which code path user is importing), `theme.ts` is always included

#### Focus areas to test
doing imports such as:
```
import { FontSizes } from 'office-ui-fabric-react/lib/Styling';
import { registerIcons } from 'office-ui-fabric-react/lib/Styling';
import { FontSizes } from '@uifabric/styling';
import { registerIcons } from '@uifabric/styling';
```
includes `theme`

### DIFF
--- a/packages/styling/src/styles/theme.ts
+++ b/packages/styling/src/styles/theme.ts
@@ -18,16 +18,20 @@ let _onThemeChangeCallbacks: Array<(theme: ITheme) => void> = [];
 
 export const ThemeSettingName = 'theme';
 
-if (!Customizations.getSettings([ThemeSettingName]).theme) {
-  const win: any = getWindow(); // tslint:disable-line:no-any
+export function initializeThemeInCustomizations(): void {
+  if (!Customizations.getSettings([ThemeSettingName]).theme) {
+    const win: any = getWindow(); // tslint:disable-line:no-any
 
-  if (win?.FabricConfig?.theme) {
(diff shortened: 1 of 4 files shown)


packages/styling/src/styles/theme.ts
```diff
@@ -18,17 +18,21 @@ let _onThemeChangeCallbacks: Array<(theme: ITheme) => void> = [];
 
 export const ThemeSettingName = 'theme';
 
-if (!Customizations.getSettings([ThemeSettingName]).theme) {
-  const win: any = getWindow(); // tslint:disable-line:no-any
+export function initializeThemeInCustomizations(): void {
```
dzearing, line 21: oh one more minor change; if there is any content in this file which could be stripped in tree shaking, it should be in its own file. I'm not sure there is, so don't feel compelled to change anything but just a mention to look for global code and isolate (perhaps we should follow the filename convention `*.global.ts*` to separate side effect code from treeshakeable.)

Reason: Webpack has unpredictable results when it finds a module with side effect code and ends up pulling in more than it needs. Shift/Marija found this out a few weeks ago.

xugao: I'm no longer marking `theme.ts` as a `sideEffect`. therefore there is no need to split this file further

PR #13540, opened 2020-06-09
DetailsList: Remove unnecessary styles

### LINKED ISSUE #13539: DetailsList styles clean-up
There are some style changes I'd like to see in DetailsList, but I'd like someone with more experience with the component to weigh in, to avoid unintended consequences.

1) Change the `min-width` inline style on the `ms-DetailsHeader` element. Firstly, right now the pixel value measures incorrectly on a large internal list I'm testing with many, many columns. Secondly, it's difficult to override because it's an inline style. Instead, it should be moved to a class and be set to `min-width: 100%`.
2) Remove the white background from the `ms-DetailsList` element. It is unnecessary and messes up some custom styles I'm trying to implement. It also is locked to the viewport and scrolls away (on an internal list I'm testing on).

phkuo:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13539

#### Description of changes

Removes some styles that do nothing for the list. See attached workitem.

### DIFF
--- a/packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.base.tsx
@@ -166,7 +166,6 @@ export class DetailsHeaderBase extends React.Component<IDetailsHeaderBaseProps, 
       selectAllVisibility,
       ariaLabelForSelectionColumn,
       indentWidth,
-      rowWidth = 0,
       onColumnClick,
       onColumnContextMenu,
       onRenderColumnHeaderTooltip = this._onRenderColumnHeaderTooltip,
@@ -218,7 +217,6 @@ export class DetailsHeaderBase extends React.Component<IDetailsHeaderBaseProps, 
         ref={this._onRootRef}
         onMouseMove={this._onRootMouseMove}
         data-automationid="DetailsHeader"
-        style={{ minWidth: rowWidth }}
         direction={FocusZoneDirection.horizontal}
       >
         {showCheckbox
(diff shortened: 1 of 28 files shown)


phkuo:
@paulgildea Yes I understand the concern. We do have several visual regression tests in place here in FluentUI run as part of the PR. I've been playing around with these changes in SharePoint in different lists to verify that there won't be regressions for the big big lists there.

I've also consulted Thomas, who has more experience than me in this feature area. He's actually piloted removing the pixel-based in-width from the header already, and can confirm that it works as expected.

Regarding the removal of the background, that'd actually be buggy if anyone were taking advantage of it.

packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.base.tsx
```diff
@@ -218,7 +217,6 @@ export class DetailsHeaderBase extends React.Component<IDetailsHeaderBaseProps,
         ref={this._onRootRef}
         onMouseMove={this._onRootMouseMove}
         data-automationid="DetailsHeader"
-        style={{ minWidth: rowWidth }}
         direction={FocusZoneDirection.horizontal}
```
dzearing, line 220: I think this will shrink the header to a smaller width than rows.

phkuo: it does not in my testing, because each column has a width set anyways
i've been testing with column resize as well
i'm honestly not sure why rowWidth exists in the first place if each field's width needs to be calculated individually anyways
@ThomasMichon any thoughts?

PR #13544, opened 2020-06-10
fix height and width check in the image component

### LINKED ISSUE #13543: [Image Component] Calculate the wrong coverStyle when the height and width is 100%
### Environment Information

- **Package version(s)**:  7.x
- **Browser and OS versions**: IE

https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/Image/Image.base.tsx#L168

doesn't consider the case that the height and width can be string like '100%'.

Requested priority: Low

Products/sites affected: SharePoint Online

qianghuang94:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13543

#### Description of changes

`width as number` and `height as number` are too dangerous. There are chances that caller is providing 100% to the width and height. At that point, the coverStyle will fail

#### Focus areas to test

Image Component


qianghuang94:
@dzearing Can you help to review this PR?

PR #13551, opened 2020-06-10
Perf-test: set no timeout for puppeteer page action

xugao:
#### Pull request checklist

#### Description of changes
remove timeout for page action.
Perf tests are currently timing out occasionally with 60s timeout set:
https://uifabric.visualstudio.com/fabricpublic/_build?definitionId=146&_a=summary

#### Focus areas to test

(optional)

### DIFF
--- a/apps/perf-test/tasks/perf-test.ts
+++ b/apps/perf-test/tasks/perf-test.ts
@@ -275,8 +275,8 @@ module.exports = async function getPerfRegressions() {
     outDir,
     tempDir,
     pageActions: async (page, options) => {
-      // temporarily removing the timeout
-      // while investigating what the correct timeout should be.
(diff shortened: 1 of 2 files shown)


apps/perf-test/tasks/perf-test.ts
```diff
@@ -275,7 +275,7 @@ module.exports = async function getPerfRegressions() {
     outDir,
     tempDir,
     pageActions: async (page, options) => {
-      page.setDefaultTimeout(60000); // set 60s timeout, default is 30s.
+      page.setDefaultTimeout(0); // set no timeout
```
JasonGore, line 278: Actually it might be good to just summarize your findings so far here.. at least explain why it's (temporarily) 0.

xugao: added comments

PR #13553, opened 2020-06-10
React conformance - First groups of rules

sopranopillow:
#### Pull request checklist

#### Description of changes

Adding some of the rules. These are the tests were added: 

1a- Component file exports a valid React element type (pre-runtime test: react-docgen-typescript can find an element)
3- Has a docblock, 5 to 25 words
(use react-docgen-typescript)

1b- Component file exports a valid React element type (runtime test: require() file path, look for an export with expected name, attempt to render it)
2- Constructor/component is a named function (or has displayName)
16- Has static displayName matching constructor name
4- Constructor/component name matches filename
(similar to v0 but use react-docgen-typescript to get component info)


packages/react-conformance/package.json
```diff
@@ -25,14 +25,18 @@
     "@types/react": "16.8.25",
     "@types/react-dom": "16.8.4",
     "@types/react-test-renderer": "^16.0.0",
+    "@types/react-is": "^16.7.1",
     "@uifabric/build": "^7.0.0",
     "@uifabric/tslint-rules": "^7.2.1",
     "enzyme": "~3.10.0",
     "enzyme-adapter-react-16": "^1.15.0",
     "jest": "~24.9.0",
+    "lodash": "^4.17.15",
```
ecraig12345, line 34: lodash, react-docgen-typescript, and react-is should be dependencies (not devDependencies), because a consumer will need them when using this package.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,69 @@
...
+/** For a reference of the tests, check the file rules.md in @fluentui/react-conformance */
+
+export const defaultTests: TestObject = {
+  /** Component has a docblock with 5 to 25 words */
+  'v0-1a': (componentInfo: ComponentDoc, componentPath: string, testInfo: TestingOptions) => {
+    const maxWords = 25;
+    const minWords = 5;
+
+    // No need to check if the description is undefined, ComponentDoc.description is a "string" not "string | undefined"
+    it(`Component has a docblock with ${minWords} to ${maxWords} words`, () => {
```
ecraig12345, line 16: Generally you should omit the "Component" part in the descriptions (same in other tests)
```suggestion
    it(`has a docblock with ${minWords} to ${maxWords} words`, () => {
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { ComponentDoc } from 'react-docgen-typescript';
+import { mount } from 'enzyme';
+import * as ReactIs from 'react-is';
+import * as React from 'react';
+
+/** For a reference of the tests, check the file rules.md in @fluentui/react-conformance */
+
+export const defaultTests: TestObject = {
+  /** Component has a docblock with 5 to 25 words */
+  'v0-1a': (componentInfo: ComponentDoc, componentPath: string, testInfo: TestingOptions) => {
```
ecraig12345, line 11: We should probably come up with more descriptive names--these will be meaningless except for anyone looking at our spec.

sopranopillow: I added a test.md file with all the rules that we are going to add. I also added a comment at the top saying where to find it but I agree, I just wasn't really sure what to call them.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const defaultTests: TestObject = {
+  /** Component has a docblock with 5 to 25 words */
+  'has-docblock': (componentInfo: ComponentDoc, componentPath: string, testInfo: TestingOptions) => {
+    const maxWords = 25;
+    const minWords = 5;
+
+    // No need to check if the description is undefined, ComponentDoc.description is a "string" not "string | undefined"
+    it(`has a docblock with ${minWords} to ${maxWords} words`, () => {
+      expect(componentInfo.description).toBeGreaterThanOrEqual(minWords);
```
ecraig12345, line 17: This will count the number of characters, not words. To count words you'd have to import lodash and do `_.words()` like v0 does, or some other method of splitting into words.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,69 @@
...
+  /** Component file exports a valid React element type  */
+  'exports-react-element': (componentInfo: ComponentDoc, componentPath: string, testInfo: TestingOptions) => {
+    it(`file exports valid React Element type`, () => {
+      expect(ReactIs.isValidElementType(require(componentPath))).toBe(true);
+    });
+  },
+  /** Component file exports a valid React element and can render it */
+  'component-renders': (componentInfo: ComponentDoc, componentPath: string, testInfo: TestingOptions) => {
+    const { requiredProps } = testInfo;
+    const Component = require(componentPath);
```
ecraig12345, line 30: You should move this code within the `it()` so that it will be run within the context of the actual test (so failures are linked to the right place). Same for the setup of all the other tests.

Also, `require(componentPath)` will get you the file's entire exports object, not just the component. To get the component, which may be either a default or named export, you should do:
```
const componentFile = require(componentPath);
const Component = componentFile.default || componentFile[componentName];
expect(Component).toBeDefined();
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const maxWords = 25;
+    const minWords = 5;
+
+    // No need to check if the description is undefined, ComponentDoc.description is a "string" not "string | undefined"
+    it(`has a docblock with ${minWords} to ${maxWords} words`, () => {
+      expect(componentInfo.description).toBeGreaterThanOrEqual(minWords);
+      expect(componentInfo.description).toBeLessThanOrEqual(maxWords);
+    });
+  },
+  /** Component file exports a valid React element type  */
```
ecraig12345, line 21: Nitpicky thing but it would help readability if you add an empty line between the tests

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const mountedComponent = mount(<Component {...requiredProps} />);
+
+    it(`renders`, () => {
+      expect(mountedComponent.exists()).toBe(true);
+    });
+  },
+  /** Component's constructor is a named function or component has a displayName */
+  'constructor-is-named-function': (componentInfo: ComponentDoc, componentPath: string, testInfo: TestingOptions) => {
+    const Component = require(componentPath);
+    const constructorName = Component.prototype.constructor.name;
```
ecraig12345, line 40: `constructor` won't be defined if this is a function component, so you should use `?.` ([optional chaining operator](https://www.codeisbae.com/typescript-optional-chaining-nullish-coalescing/)):
```suggestion
    const constructorName = Component.prototype.constructor?.name;
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    it(`constructor's name is a named function or has displayName`, () => {
+      expect(constructorName || componentInfo.displayName).toBeTruthy();
+    });
+  },
+  /** Component has a static displayName matching constructor name */
+  'displayName-matches-constructor': (componentInfo: ComponentDoc, componentPath: string, testInfo: TestingOptions) => {
```
ecraig12345, line 47: You should probably combine this test with the previous one, and just change the `it()` description depending on the type of component. You can tell it's a class component if `Component.prototype.constructor` is defined.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const constructorName = Component.prototype.constructor.name;
+
+    it(`has a static displayName that matches the constructor's name`, () => {
+      expect(Component.displayName).toEqual(constructorName);
+    });
+  },
+  /** Constructor/component name matches filename */
+  'name-matches-filename': (componentInfo: ComponentDoc, componentPath: string, testInfo: TestingOptions) => {
+    const Component = require(componentPath);
+    const fileName = componentPath
```
ecraig12345, line 58: You can do this more simply with `path.basename(componentPath, path.extname(componentPath))`

packages/react-conformance/src/isConformant.ts
```diff
@@ -1,15 +1,32 @@
...
-  // }
+export function isConformant(componentPath: string, options: TestingOptions, extraTests?: TestObject) {
+  const tsconfigPath = path.join(process.cwd(), 'tsconfig.json');
+  const parser = withCustomConfig(tsconfigPath, {});
+  const components = parser.parse(componentPath);
+
+  if (components.length === 1) {
+    const componentInfo = components[0];
+
+    for (const test in defaultTests) {
```
ecraig12345, line 14: Instead of `for..in` it's better to use `for..of` on `Object.keys()` (available everywhere) or `Object.values()` or `Object.entries()` (available in Node context and newer browsers).
```
for (const test of Object.values(defaultTests) {
  test(componentInfo, componentPath, options);
}
```

packages/react-conformance/src/rules.md
```diff
@@ -0,0 +1,77 @@
+# Conformance tests rules
```
ecraig12345, line 1: This probably doesn't need to be checked in (or should be formatted differently) now that we're giving the tests descriptive names.

packages/react-conformance/tsconfig.json
```diff
@@ -7,15 +7,17 @@
     "jsx": "react",
     "declaration": true,
     "sourceMap": true,
+    "noImplicitAny": false,
```
ecraig12345, line 10: Please re-enable `noImplicitAny` and `noImplicitThis` and fix any issues.

packages/react-conformance/src/types.ts
```diff
@@ -1,9 +1,12 @@
-export interface IsConformantOptions {
+import { ComponentDoc } from 'react-docgen-typescript';
+export interface TestingOptions {
```
ecraig12345, line 2: Generally nice to have a blank line between imports and contents.
```suggestion

export interface TestingOptions {
```

packages/react-conformance/src/types.ts
```diff
@@ -1,9 +1,13 @@
-export interface IsConformantOptions {
+import { ComponentDoc } from 'react-docgen-typescript';
```
ecraig12345, line 1: Let's add the pre-loaded component as an option, as well as the component path and whether to use the default export or a named export.
```ts
import { ComponentDoc } from 'react-docgen-typescript';

export interface TestingOptions<TProps = {}> {
  componentPath: string;
  Component: React.ComponentType<TProps>;
  constructorName: string;
  requiredProps?: Partial<TProps>;
  exportedAtTopLevel?: boolean;
  useDefaultExport?: boolean;
}
```

Then update the various tests that are `require`ing the component to use the pre-loaded one.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    // No need to check if the description is undefined, ComponentDoc.description is a "string" not "string | undefined"
+    it(`has a docblock with ${minWords} to ${maxWords} words`, () => {
+      expect(_.words(componentInfo.description)).toBeGreaterThanOrEqual(minWords);
+      expect(_.words(componentInfo.description)).toBeLessThanOrEqual(maxWords);
+    });
+  },
+
+  /** Component file exports a valid React element type  */
+  'exports-react-element': (componentInfo: ComponentDoc, componentPath: string, testInfo: TestingOptions) => {
```
ecraig12345, line 23: As mentioned on chat, I think we should replace this test with the following:
```ts
'exports-component': (componentInfo: ComponentDoc, testInfo: TestingOptions) => {
  it(`exports component from file under correct name`, () => {
    const componentFile = require(testInfo.componentPath);
    if (testInfo.useDefaultExport) {
      expect(componentFile.default).toBe(testInfo.Component);
    } else {
      expect(componentFile[testInfo.constructorName]).toBe(testInfo.Component);
    }
  });
},
```

packages/react-conformance/tsconfig.json
```diff
@@ -7,15 +7,17 @@
     "jsx": "react",
     "declaration": true,
     "sourceMap": true,
+    "noImplicitAny": true,
```
ecraig12345, line 10: You can get rid of the explicit settings for `noImplicitAny` and `noImplicitThis` because both of those are implied by `strict: true`.  (see the entry for `--strict` [here](https://www.typescriptlang.org/docs/handbook/compiler-options.html))

packages/react-conformance/src/types.ts
```diff
@@ -1,9 +1,16 @@
...
-  tests: String[];
-  requiredProps?: object;
-  exportedAtTopLevel: boolean;
-  // packageRoot: any;
+import * as React from 'react';
+import { ComponentDoc } from 'react-docgen-typescript';
+
+export interface TestingOptions<TProps = {}> {
+  componentPath: string;
+  Component: React.ComponentType<TProps>;
```
ecraig12345, line 6: You should add `displayName` here too (I think that's better than `constructorName` since it's more generic) so you can reference the expected value in the tests.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,71 @@
...
+      expect(_.words(componentInfo.description)).toBeGreaterThanOrEqual(minWords);
+      expect(_.words(componentInfo.description)).toBeLessThanOrEqual(maxWords);
+    });
+  },
+
+  /** Component file exports a valid React element type  */
+  'exports-component': (componentInfo: ComponentDoc, testInfo: TestingOptions) => {
+    it(`exports component from file under correct name`, () => {
+      const componentFile = require(testInfo.componentPath);
+      const constructorName = testInfo.Component.prototype.constructor?.name;
```
ecraig12345, line 25: Use `testInfo.displayName` here instead of getting it from the component.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * if functional component: component has a displayName
+   * else: component's constructor is a named function and matches displayName
+   */
+  'component-has-displayname': (componentInfo: ComponentDoc, testInfo: TestingOptions) => {
+    const { Component } = testInfo;
+    if (Component.prototype.constructor) {
+      it(`constructor is a named function and matches displayName`, () => {
+        const constructorName = Component.prototype.constructor.name;
+        expect(constructorName).toEqual(Component.displayName);
```
ecraig12345, line 52:
```suggestion
        expect(constructorName).toEqual(testInfo.displayName);
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,71 @@
...
+      });
+    } else {
+      it(`has a displayName`, () => {
+        expect(Component.displayName).toBeTruthy();
```
ecraig12345, line 56:
```suggestion
        expect(Component.displayName).toEqual(testInfo.displayName);
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,71 @@
...
+      });
+    }
+  },
+
+  /** Constructor/component name matches filename */
+  'name-matches-filename': (componentInfo: ComponentDoc, testInfo: TestingOptions) => {
+    it(`Component/constructor name matches filename`, () => {
+      const { componentPath, Component } = testInfo;
+      const constructorName = Component.prototype.constructor?.name;
```
ecraig12345, line 65: You should probably use the displayName from testInfo here instead.

packages/react-conformance/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@fluentui/react-conformance",
   "version": "0.1.0",
-  "description": "Conformance testing",
+  "description": "Customizable conformance testing utility for React components.",
```
dzearing, line 4:
```suggestion
  "description": "Customizable conformance testing utility for Fluent UI React components.",
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,71 @@
...
+      if (testInfo.useDefaultExport) {
+        expect(componentFile.default).toBe(testInfo.Component);
+      } else {
+        expect(componentFile[constructorName]).toBe(testInfo.Component);
+      }
+    });
+  },
+
+  /** Component file exports a valid React element and can render it */
+  'component-renders': (componentInfo: ComponentDoc, testInfo: TestingOptions) => {
```
dzearing, line 35: why are we using kebab case here? I'm a little confused on the usage.

sopranopillow: I used kebab casing to make it similar to how eslint names its rules. Also at the beginning I was thinking on how to be able to disable rules in case they can be opted out and using something like: disableRules: ['component-renders', 'next-rule'] would be a nice way of doing it.

packages/react-conformance/src/isConformant.ts
```diff
@@ -1,15 +1,29 @@
...
-  //     test();
-  //   });
-  // }
+export function isConformant(options: TestingOptions, extraTests?: TestObject) {
+  const { componentPath } = options;
+  const tsconfigPath = path.join(process.cwd(), 'tsconfig.json');
+  const parser = withCustomConfig(tsconfigPath, {});
+  const components = parser.parse(componentPath);
+
+  if (components.length === 1) {
```
ecraig12345, line 12: Maybe a good way to handle the multiple components case and verify that the display name is correct would be something like this:
```ts
const mainComponents = components.filter(comp => comp.displayName === testInfo.displayName);
if (mainComponents.length === 1) {
  // do the tests
} else {
  // throw an error
}
```

packages/react-conformance/src/isConformant.ts
```diff
@@ -1,15 +1,37 @@
...
+  } else {
+    if (components.length === 0) {
+      throw Error('No exported components in path: ' + componentPath);
+    } else {
+      throw Error(
+        `No component with name '${displayName}' was found, these are the exported components: {${components
+          .map(component => {
+            return component.displayName;
+          })
+          .toString()}} in path ${componentPath}`,
```
ecraig12345, line 33: I'd change this around slightly:
```suggestion
        `No component with name '${displayName}' was found at ${componentPath}. ` +
          `These are the exported component names: ${components.map(component => component.displayName).join(', ')}`,
```

packages/react-conformance/src/isConformant.ts
```diff
@@ -1,15 +1,37 @@
...
+    if (extraTests) {
+      for (const test of Object.keys(extraTests)) {
+        extraTests[test](componentInfo, testInfo);
+      }
+    }
+  } else {
+    if (components.length === 0) {
+      throw Error('No exported components in path: ' + componentPath);
+    } else {
+      throw Error(
```
ecraig12345, line 28: When creating errors we generally use `new Error`, though at least [according to this](https://stackoverflow.com/questions/13294658/throw-errormsg-vs-throw-new-errormsg) they're equivalent.
```suggestion
      throw new Error('No exported components in path: ' + componentPath);
    } else {
      throw new Error(
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -0,0 +1,69 @@
...
+import * as _ from 'lodash';
+import * as path from 'path';
+
+export const defaultTests: TestObject = {
+  /** Component has a docblock with 5 to 25 words */
+  'has-docblock': (componentInfo: ComponentDoc, testInfo: TestingOptions) => {
+    const maxWords = 25;
+    const minWords = 5;
+
+    // No need to check if the description is undefined, ComponentDoc.description is a "string" not "string | undefined"
```
dzearing, line 14: it would be nice to see an example using the conformance test to know how it looks.

dzearing: (not specifically here, just adding a comment)

sopranopillow: I'm currently working on the example!

PR #13568, opened 2020-06-11
fix(FormExample): Update fields ID and add success indicator by default when field required

assuncaocharles:
#### Pull request checklist

# BREAKING CHANGES

This PR updates input/label ID's to correctly match and add `successIndicator` by default when the field is required. 

`Inputs` when required will by default show success indicator when fulfilled, to prevent it and have the previously behavior just use `<Input required showSuccessIndicator={false} />`or 

```typescript
const fields =  [{
    label: 'First name',
    name: 'firstName',
    id: 'first-name-error',
    key: 'first-name',
    errorMessage: 'You can not fix this error',
    required: true,
    control: { 
         as: Input,
         showSuccessIndicator: false
    }
  }]
```
#### Focus areas to test

(optional)


assuncaocharles:
@jurokapsiar @layershifter @pompomon @mnajdova @miroslavstastny As agreed, in this PR `Input` when as the flag `required` will show the `successIndicator` the question is if we want to do it always when it's required? Previously we only showed if `successIndicator` were passed, but changing this will force it for all required inputs, just wanted to confirm it's what we really want.

mnajdova:
> As agreed, in this PR Input when as the flag required will show the successIndicator the question is if we want to do it always when it's required? Previously we only showed if successIndicator were passed, but changing this will force it for all required inputs, just wanted to confirm it's what we really want.

In my opinion it would look awkward if all fields which are not required will have the success indicator.. Usually on forms, the success indicator is shown after you successfully filled the form field (which in my opinion means, that the field either was required, or required some special validation - maybe it is an email or phone number) However this is just my opinion, and we should definitely check the designs (maybe it should be shown only if the input field was focused, or the value was initialized..)

assuncaocharles:
> > As agreed, in this PR Input when as the flag required will show the successIndicator the question is if we want to do it always when it's required? Previously we only showed if successIndicator were passed, but changing this will force it for all required inputs, just wanted to confirm it's what we really want.
> 
> In my opinion it would look awkward if all fields which are not required will have the success indicator.. Usually on forms, the success indicator is shown after you successfully filled the form field (which in my opinion means, that the field either was required, or required some special validation - maybe it is an email or phone number) However this is just my opinion, and we should definitely check the designs (maybe it should be shown only if the input field was focused, or the value was initialized..)

But it's not what happening, the success indicator is only being shown if the field was required... The question here is because in several examples we were using required field ( Like in the input inside toolbar ) and having the indicator by default in required fields will generate regression because the input get bigger. Removing the required will generate regression because will remove `*` from the label...

packages/fluentui/docs/src/examples/components/Form/Usage/FormExampleErrorAndSuccessful.shorthand.tsx
```diff
@@ -1,6 +1,5 @@
 import * as React from 'react';
-import { Form, Button, Input } from '@fluentui/react-northstar';
-import { PresenceAvailableIcon } from '@fluentui/react-icons-northstar';
+import { Form, Button } from '@fluentui/react-northstar';
 
 const fields = [
   {
```
mnajdova, line 5: image
Would suggest changing the error here to something like "You cannot fix this error", or handle somehow inside the component's state when this should go away.

packages/fluentui/docs/src/examples/components/Form/Usage/FormExampleErrorAndSuccessful.tsx
```diff
@@ -9,26 +9,28 @@ const FormExampleErrorAndSatisfactory = () => (
...
       </Form.Message>
     </Form.Field>
-    <Form.Field
-      label="Last name"
-      name="LastName"
-      id="last-name-shorthand-with-error"
-      errorMessage={'ERROR'}
-      required={true}
-    />
+    <Form.Field label="Last name" name="LastName" id="last-name-shorthand-with-error" errorMessage="ERROR" required />
```
mnajdova, line 20: See https://github.com/microsoft/fluentui/pull/13568/files#r439439610

packages/fluentui/react-northstar/src/components/Input/Input.tsx
```diff
@@ -103,6 +103,9 @@ export interface InputProps extends UIComponentProps, ChildrenComponentProps, Su
 
   /** Optional Icon to display inside the Input if required and fulfilled. */
   successIndicator?: ShorthandValue<BoxProps>;
+
+  /** Set wether the successIndicator should be visible. */
```
mnajdova, line 107:
```suggestion
  /** Indicates whether the successIndicator should be visible. */
```

packages/fluentui/react-northstar/src/components/Input/Input.tsx
```diff
@@ -165,7 +170,7 @@ const Input: React.FC<WithAsProp<InputProps>> & FluentComponentStaticProps<Input
       inline,
       disabled,
       clearable,
-      hasIcon: !!icon || !!successIndicator || !!error,
+      hasIcon: !!icon || showSuccessIndicator || required || error,
```
mnajdova, line 173: Can we please send separately: `hasIcon`, `showSuccessIndicator`, `reuired` and `hasError` and do this condition in the theme? 

That way would be easier to do changes in the theme without necessarily requiring code changes.

packages/fluentui/docs/src/examples/components/Form/Usage/FormExampleErrorAndSuccessful.shorthand.tsx
assuncaocharles: Gonna change it to `You can't fix this error` we gonna handle the change probably in the prototype with some form lib

packages/fluentui/react-northstar/src/components/Input/Input.tsx
assuncaocharles: This is just the input condition for set the slot for the icon, we would be passing the props to styles only for that?

mnajdova: My point was instead of doing this decision here, why not doing that in the styles based on 4 booleans. In the end is this component's logic of theme logic? I would expect that some themes may show these differently, so if they have access to all of them, that decision can be made in the theme, otherwise it will be locked to component's logic @layershifter @jurokapsiar @miroslavstastny any other opinions?

mnajdova: After offline discussion we decided not to change this as part of this PR, we will revisit this if needed to be changed in a separete PR

packages/fluentui/docs/src/examples/components/Form/Usage/FormExampleErrorAndSuccessful.shorthand.tsx
assuncaocharles: Changed

packages/fluentui/react-northstar/src/components/Form/FormField.tsx
```diff
@@ -62,6 +62,9 @@ export interface FormFieldProps extends UIComponentProps, ChildrenComponentProps
 
   /** Indicator to be shown when field is required and non-empty */
   successIndicator?: ShorthandValue<BoxProps>;
+
+  /** Set wether the successIndicator should be visible. */
```
mnajdova, line 66:
```suggestion
  /** Indicates whether the successIndicator should be visible. */
```

packages/fluentui/react-northstar/src/components/Form/FormField.tsx
```diff
@@ -152,6 +156,7 @@ const FormField: React.FC<WithAsProp<FormFieldProps>> & FluentComponentStaticPro
         required,
         name,
         type,
+        ...(showSuccessIndicator && { showSuccessIndicator }),
```
mnajdova, line 159:
```suggestion
        showSuccessIndicator,
```
?

packages/fluentui/react-northstar/src/components/Form/FormField.tsx
```diff
@@ -62,6 +62,9 @@ export interface FormFieldProps extends UIComponentProps, ChildrenComponentProps
 
   /** Indicator to be shown when field is required and non-empty */
   successIndicator?: ShorthandValue<BoxProps>;
+
+  /** Indicates whether the successIndicator should be visible. */
+  showSuccessIndicator?: boolean;
```
mnajdova, line 67: Let's remove this

mnajdova, line 67:
```suggestion
```

packages/fluentui/react-northstar/src/components/Input/Input.tsx
mnajdova, line 173:
```suggestion
      hasIcon: !!icon || showSuccessIndicator || required || !!error,
```

packages/fluentui/docs/src/examples/components/Form/Usage/FormExampleErrorAndSuccessful.shorthand.tsx
```diff
@@ -18,9 +17,9 @@ const fields = [
     key: 'last-name',
     control: {
       as: Input,
-      successIndicator: <PresenceAvailableIcon />,
+      showSuccessIndicator: true,
+      required: true,
```
jurokapsiar, line 21: was it intentional to move it? then we do not show *

assuncaocharles: Nops, by mistake. Fixed already

PR #13570, opened 2020-06-11
Checkbox: use state option and move Icon out of Base component

xugao:
#### Pull request checklist

#### Description of changes

- compose: pass forwarded ref as param in state option
- Checkbox: use state option and move Icon out of Base component

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-compose/src/compose.ts
+++ b/packages/react-compose/src/compose.ts
@@ -20,11 +20,11 @@ function compose<
     wasComposedPreviously(input) ? input.fluentComposeConfig : undefined,
   );
 
-  const Component = (React.forwardRef<TElementType, TInputProps & TParentProps & { as?: React.ElementType }>(
+  const Component = (React.forwardRef<HTMLElement, TInputProps & TParentProps & { as?: React.ElementType }>(
     (props, ref) => {
-      return composeOptions.render(props, (ref as unknown) as React.Ref<HTMLDivElement>, {
+      return composeOptions.render(props, ref as React.Ref<HTMLDivElement>, {
         ...composeOptions,
-        state: composeOptions.state(props, composeOptions),
+        state: composeOptions.state(props, ref, composeOptions),
         slots: {
           ...composeOptions.slots,
           __self: Component,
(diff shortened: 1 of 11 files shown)


packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -1,23 +1,30 @@
...
-    const { slotProps, slots, state } = useCheckbox(props, composeOptions, forwardedRef);
+    // TODO: improve typing for mergeProps
+    const { slotProps, slots, state } = mergeProps<ICheckboxProps, ICheckboxState>(
+      composeOptions.state,
+      composeOptions,
+    ) as {
+      state: ICheckboxState;
+      slots: ICheckboxSlots;
+      slotProps: ICheckboxSlotProps;
+    };
```
khmakoto, line 26: Is the `as` needed?

packages/react-compose/src/compose.ts
```diff
@@ -20,11 +20,11 @@ function compose<
     wasComposedPreviously(input) ? input.fluentComposeConfig : undefined,
   );
 
-  const Component = (React.forwardRef<TElementType, TInputProps & TParentProps & { as?: React.ElementType }>(
+  const Component = (React.forwardRef<HTMLElement, TInputProps & TParentProps & { as?: React.ElementType }>(
```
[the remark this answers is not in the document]
xugao: no, the previous type is wrong. `TElementType` can be something like `div` but in that case what we actually want here is `React.forwardRef<HTMLDivElement` instead of `React.forwardRef<TElementType`. 
Honestly, typings here is quite messed up in general (look at how many `as` being used around this code). I don't want to change the world here so taking a smaller step to correct this

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
xugao: yes and i'm addressing this in this PR: #13570 13571

PR #13571, opened 2020-06-11
Include typings for slots and slotProps in mergeProps typing

xugao:
#### Pull request checklist

#### Description of changes
Include typings for slots and slotProps in mergeProps typing and fixes in its use cases.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-button/src/components/Button/Button.types.tsx
+++ b/packages/react-button/src/components/Button/Button.types.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { ComponentProps, ShorthandValue, ComposeStandardStatics, RecursivePartial } from '../../utils/tempTypes';
-import { ComposeOptions } from '@fluentui/react-compose';
+import { BaseSlots, ComposeOptions, SlotProps } from '@fluentui/react-compose';
 import { ColorPlateSet } from '@fluentui/react-theme-provider';
 
 export type SizeValue = 'smallest' | 'smaller' | 'small' | 'medium' | 'large' | 'larger' | 'largest';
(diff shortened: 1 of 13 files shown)


packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -73,13 +73,16 @@ export interface ButtonProps extends ComponentProps, React.HTMLAttributes<HTMLBu
...
   content: React.ElementType;
   loader: React.ElementType;
 }
 
 export type ButtonSlotProps = {
-  [key in keyof ButtonSlots]: ButtonProps[key];
+  [key in keyof Omit<ButtonSlots, 'root'>]: ButtonProps[key];
+} & {
+  root: React.ButtonHTMLAttributes<HTMLButtonElement>;
 };
```
khmakoto, line 86: We could abstract the `root` typing into `compose` and extend it.

packages/react-compose/src/mergeProps.ts
```diff
@@ -7,10 +7,13 @@ import { resolveSlotProps } from './resolveSlotProps';
...
-  options: ComposePreparedOptions<TProps, TState>,
-): MergePropsResult<TState> {
+export function mergeProps<
+  TProps,
+  TState = TProps,
+  TSlots = GenericDictionary,
+  // tslint:disable-next-line:no-any
+  TSlotProps = { [key in keyof TSlots]: any }
+>(state: TState, options: ComposePreparedOptions<TProps, TState>): MergePropsResult<TState, TSlots, TSlotProps> {
   const result: MergePropsResult<TState> = {
```
khmakoto, line 17: You could do:
```suggestion
  const result: MergePropsResult<TState, TSlots, TSlotProps> = {
```

so that you don't have to do the `as` when you return.

xugao: if you look at `result` value, it's not type `MergePropsResult<TState, TSlots, TSlotProps>`. only after resolving, it becomes `MergePropsResult<TState, TSlots, TSlotProps>`, hence i'm casting it later

packages/react-button/src/components/Button/Button.types.tsx
xugao: I added shared typing in `react-compose`, but m not very sure that's the right package to put

PR #13575, opened 2020-06-11
Creating compat layers for Button, Checkbox, Link, Pivot, Slider and Toggle components 

khmakoto:
#### Pull request checklist

#### Description of changes

This PR replaces the `Button` exported in `react-next` that was pointing to the one in `office-ui-fabric-react` to the one that is pointing to `react-button`, while creating a compatibility layer in `react-next` to still export the `office-ui-fabric-react`.

It also creates this compatibility layer for the `Checkbox`, `Link`, `Pivot`, `Slider` and `Toggle` components to have separate exports for the `office-ui-fabric-react` and `@fluentui/react-next` components.

Finally, this PR makes changes to allow for multiple `api-extractor` configs to be provided per package and makes changes to `lint-imports` so compat imports are be flagged by the deep imports rule.

#### Focus areas to test

(optional)


xugao:
this LGTM , but i will let @joschect to approve :)

packages/react-next/etc/react-next.api.md
```diff
@@ -5,9 +5,9 @@
 ```ts
 
 import { ComposePreparedOptions } from '@fluentui/react-compose';
-import { IButtonProps } from 'office-ui-fabric-react/lib/Button';
+import { IButtonProps } from '@fluentui/react-next/lib/CompatButton';
```
JustSlone, line 8: Should this be something more like `'@fluentui/react-next/lib/compat/Button'`

khmakoto: I thought we didn't want to allow for imports from internal folders so we restricted that to imports from the overall package, i.e. `@fluentui/react-next` and imports from specific files under `lib`, i.e. `@fluentui/react-next/lib/<component>`.

packages/react-next/etc/react-next.api.md
```diff
@@ -4,12 +4,66 @@
...
+import { CompoundButton } from 'office-ui-fabric-react/lib/Button';
+import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
+import { ElementType } from 'office-ui-fabric-react/lib/Button';
+import { IBaseButtonProps } from 'office-ui-fabric-react/lib/Button';
+import { IBaseButtonState } from 'office-ui-fabric-react/lib/Button';
 import { IBaseProps } from 'office-ui-fabric-react/lib/Utilities';
+import { IButton } from 'office-ui-fabric-react/lib/Button';
 import { IButtonProps } from 'office-ui-fabric-react/lib/Button';
-import { IButtonProps as IButtonProps_2 } from 'office-ui-fabric-react/lib/components/Button/Button.types';
+import { IButtonProps as IButtonProps_2 } from '@fluentui/react-next/lib/CompatButton';
```
ecraig12345, line 35: Anywhere you're getting a _2 or _3 means that you're exporting the same name twice with different definitions, which is going to cause problems for consumers.

packages/react-next/etc/react-next.api.md
ecraig12345: I would prefer `compat/Button`, and we could update our guidance to indicate paths immediately under `compat` are also considered top-level paths that are part of the public API.

Another nice thing about that approach is we could have a file `compat/index` where you could get any compat component exported by its original name. Then only do the renaming when re-exporting from the root (or maybe even better, we don't re-export those things from the root at all).

dzearing: `@fluentui/react-next/lib/compat`

ecraig12345: That's what I meant as well, I was just being lazy with typing 😆

packages/react-next/src/CompatCheckbox.ts
```diff
@@ -0,0 +1,9 @@
+export {
+  Checkbox as CompatCheckbox,
```
dzearing, line 2: src/compat/Checkbox.ts
src/compat/index.ts // (exports * from checkbox)

dzearing: Then someone importing:

/lib/Button => /lib/compat/Button
/lib/Checkbox => /lib/compat/Checkbox

packages/react-next/etc/react-next.api.md
ecraig12345: Err to clarify, what I actually meant was something like this, basically mirroring our existing structure that lets you import from either the index (all components) or individual component files:
```
@fluentui/react-next/lib/compat/
  index
  Button
  Checkbox
  ...any other component with compat layer...
```

packages/react-next/src/CompatCheckbox.ts
khmakoto: Done

packages/react-next/etc/react-next.api.md
khmakoto: Resolved!

packages/react-next/etc/react-next.api.md
khmakoto: Cool, I went with this!

packages/react-next/package.json
```diff
@@ -30,7 +30,7 @@
     "start": "just-scripts dev:storybook",
     "start-test": "just-scripts jest-watch",
     "update-snapshots": "just-scripts jest -u",
-    "update-api": "just-scripts update-api"
+    "update-api": "just-scripts update-api && just-scripts update-api --configPath ./config/api-extractor-compat.json"
```
xugao, line 33: we should add `api-extractor:verify ./config/api-extractor-compat.json` to `build` too?

xugao:
pls pull master and get a good `size-auditor` run

PR #13576, opened 2020-06-11
Checkbox: deprecate onRenderLabel and support label slot prop

xugao:
#### Pull request checklist

#### Description of changes
Supporting label shorthand prop. I am using the 
This change contains:
- fixes for providing correct props to children function
- added `SlotProp` typing (same as `ShorthandValue` V0) in `react-compose`.
- added `mergeSlotProp` helper. this helps constructing `state`, as well as using it in `resolveSlotProps`.
- deprecated `onRenderProp` without breaking it and enabled support for `label` shorthand prop.

#### Focus areas to test

(optional)


packages/react-compose/src/mergeShorthandProp.ts
```diff
@@ -0,0 +1,19 @@
+import { ShorthandValue } from './types';
+import * as React from 'react';
+
+export function mergeShorthandProp<TProps>(
```
dzearing, line 4: Why is this needed? Isn't this already done in mergeProps?

dzearing: or is the idea to move it out of mergeProps?

xugao: i think we need to have the logic in mergeProps along with this helper. see the use case in `useCheckbox`

khmakoto: I think Xu's using it as a middle ground because she's not yet moved the state to use `mergeProps`.

xugao: @khmakoto , this is not a middle ground. this is something we need and user who author `state` would need. this needs to live along with `mergeProps`  (and it's already used for `Checkbox`)

xugao: @dzearing - Updated code. now I'm using this method in `resolveSlotProps` to remove duplication

packages/react-compose/etc/react-compose.api.md
```diff
@@ -115,6 +123,14 @@ export interface ShorthandConfig<TProps> {
     mappedProp?: keyof TProps;
 }
 
+// Warning: (ae-forgotten-export) The symbol "ReactNode" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export type SlotProp<TProps> = ReactNode | ObjectSlotProp<TProps>;
```
xugao, line 129: I'm calling this `SlotProp`. It's a prop for a slot & props for the slot. Little unsure about the name but I like it more than the term `shorthand`. Via compose, we are exposing term such as slot/slotProps. We should try keeping the names close

packages/react-compose/src/resolveSlotProps.ts
```diff
@@ -30,32 +31,25 @@ export function resolveSlotProps<TProps, TState = TProps>(
...
-      // If the children is a function, replace the slot.
-      if (typeof slotProp.children === 'function') {
+      if (typeof mergedSlotProp.children === 'function') {
         const { children, ...restProps } = slotProp;
-        slotProp.children = slotProp.children(slot, restProps);
-
+        // If the children is a function, replace the slot.
         slots[slotName] = React.Fragment;
+        slotProps[slotName] = {
+          children: slotProp.children({ Component: slot, props: { ...slotProps[slotName], ...restProps } }),
```
layershifter, line 48: This is not the same as previous and should be considered as a public API change.

packages/react-next/src/components/Checkbox/examples/Checkbox.Other.Example.tsx
```diff
@@ -26,11 +26,32 @@ export const CheckboxOtherExample: React.FunctionComponent = () => {
...
+      <Checkbox label={{ className: 'label', children: 'Custom label 2' }} />
+
+      <Checkbox label={{ children: CustomLabel }} />
+
+      <Checkbox onRenderLabel={_renderLabelWithLink} />
     </Stack>
   );
 };
 
+const CustomLabel: React.FunctionComponent<{
```
layershifter, line 40: `children` are render function, they are not React component and should not be them as they implement render props pattern:

https://reactjs.org/docs/render-props.html

packages/react-compose/src/resolveSlotProps.ts
xugao: to clarify, `mergeProps` is currently only used by the V8 components we are working on, not by V0 compose yet.
And I'm trying to fix 2 things here:
1.  align the params interface for the render function interface with V0
2. we missed passing `...slotProps[slotName]` previously

xugao: I updated the code slightly. If this is not the same as how V0 handles children render prop, pls let me know

layershifter: Yep, looks the same as in v0. Thanks 👍

apps/vr-tests/src/stories/CheckboxNext.stories.tsx
```diff
@@ -51,11 +51,20 @@ storiesOf('Checkbox Next', module)
       defaultChecked={true}
     />
   ))
-  .addStory('Custom render Checkbox', () => (
+  .addStory('Custom render Checkbox using onRenderLabel', () => (
     <Checkbox
       label="Persona Checkbox"
       onRenderLabel={(props: ICheckboxProps) => {
-        return <Persona text={props!.label} size={PersonaSize.size32} />;
+        return <Persona text={props!.label as string} size={PersonaSize.size32} />;
```
xugao, line 58: hmm expanding the typing for `label` can cause this type of typing error

packages/react-compose/src/mergeSlotProp.ts
```diff
@@ -0,0 +1,20 @@
+import { SlotProp } from './types';
+import * as React from 'react';
+
+export function mergeSlotProp<TProps>(
```
dzearing, line 4: comments

packages/react-compose/src/resolveSlotProps.test.tsx
```diff
@@ -8,23 +8,23 @@ const nullRenderer = () => null;
...
 
   const defaultOptionsWithSlots: ComposePreparedOptions = {
     ...defaultComposeOptions,
-    ...defaultSlots,
+    ...getDefaultSlots(),
     classes: [{ root: 'root' }, () => ({ foo: 'foo' }), { bar: 'bar', baz: 'baz' }],
   };
 
-  it('can expand strings, booleans, and numbers as children', () => {
+  fit('can expand strings, booleans, and numbers as children', () => {
```
dzearing, line 27: remove `fit`

PR #13585, opened 2020-06-12
Scripts, config, and dependencies cleanup

ecraig12345:
#### Pull request checklist

#### Description of changes

Some recently discussed fixes plus issues that I noticed while working on the eslint conversion:

- Clarify comments in checked in `version.ts` files to indicate that they're not expected to be updated (and set the version in those files to `0.0.0` to be really obvious it's not meant to be in sync).
- Upgrade lint-staged and add a separate config file (will be more relevant once eslint change goes in).
- Add `syncpack:fix` script to easily run the syncpack command to fix version mismatches.
- Delete `recover-package-versions.js` which is now handled by `beachball sync`
- Rename create-package templates' `package.json` to `package.json.hbs` so the empty versions in those files won't get (incorrectly) automatically updated by beachball
- Clean up `dom-tests` package: remove unused config and deps
- Remove unused DOM tests from `fabric-website-resources`
- Remove dependencies on polyfills (`es6-promise` etc) other than `react-app-polyfill`, because the code added awhile back to prepend `react-app-polyfill` to bundles handles the same scenarios more comprehensively
- Remove unused dependencies including `resemblejs`, `enzyme-to-json`, `highlight.js`, `immutability-helper`
- Fix implicit dependency issues revealed by running eslint `import/no-extraneous-dependencies` in another branch


xugao:
```
  "sideEffects": [
    "lib/version.js"
  ],
```
I am still confused if we need this. We want to call `setVersion` when a package is referenced/imported right?
If so , this is not working

PR #13591, opened 2020-06-13
fix(Slider): Prevent unhandled props to be passed to Box

assuncaocharles:
#### Pull request checklist

#### Description of changes

Remove `fluid` and `vertical` that was being passed to box so console doesn't throw about non valid html attribute

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Slider/Slider.tsx
```diff
@@ -217,14 +217,12 @@ const Slider: React.FC<WithAsProp<SliderProps>> & FluentComponentStaticProps = p
         ...htmlInputProps,
```
assuncaocharles, line 217: Should we be using `Input.create` instead of `Box` here? @layershifter @mnajdova @pompomon

layershifter: Nope, as this will create a plain `input` while `Input` has different markup, styles and functionality

PR #13594, opened 2020-06-14
[Text] Component needs to respect global semantic colors

### LINKED ISSUE #13595: Fluent Text Component does not respect semantic coloring
### Environment Information

- **Package version(s)**: 7.115.3

#### Actual behavior:

The Text component does not respect the theme's bodyText color.

#### Expected behavior:

The Text component does respect the theme's bodyText color, like the TextField does.

Are you willing to submit a PR to fix? Yes, it's here: https://github.com/microsoft/fluentui/pull/13594

Requested priority: Normal

Products/sites affected: portal.azure.com

nileshp87:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13595

#### Description of changes

Looking at how TextField works, I mimicked the pattern for using semanticColors. However, to leave the old behavior intact I use the semanticColor as a fallback instead of the only option.

#### Focus areas to test

(optional)


natalieethell:
Syncing up offline

natalieethell:
I would suggest chatting with @paulgildea and/or @dzearing about the behavior here. Originally, if I remember correctly, the assumption was that the Text object would inherit the text color of its parent elements. Not sure if the FluentUI team wants to change that assumption.

ecraig12345:
@nileshp87 Looks like you need to update snapshots. `cd` to `packages/office-ui-fabric-react` then run `yarn update-snapshots`.

ecraig12345:
Actually I went ahead and did the updates in the interest of getting this merged (I've been trying to help get old PRs either merged or closed).

khmakoto:
This is happening because that `Button` is the one in the `experiments` package, or at least I think that's the case. We should probably remove the screener tests related to that one. I'll do that in a separate PR to see if that fixes the issue here.

ecraig12345:
@khmakoto Even aside from the screener issue I've started wondering if this is the right change (and in particular if it's okay to introduce in a minor version). It seems like a pretty significant behavior change that some people might not be expecting--so if we want to do it, maybe including in v8 would be better, especially since we're pretty close to releasing it. @dzearing any further thoughts?

dzearing:
I think it seems fine imo. It's a fallback behavior.

JustSlone:
@nileshp87, @khmakoto, it looks like this PR is pretty far out of date. It might be a good idea to convert to draft while things get worked out.

Also, do we really need all of these requested reviewers? It would help to tidy that up too.

khmakoto:
@JustSlone Yeah, I've been working to get rid of merge conflicts and get it back to a good state. Although it might be good to close it and do it from zero again. I'll spin up a new one tomorrow.

khmakoto:
Closed in favor of #17252.

PR #13599, opened 2020-06-15
LineChart: Add polyfill to support array.find method in callout in IE11

RajeshGoriga:
#### Pull request checklist

#### Description of changes
LineChart: Add polyfill to  support array.find method in callout  in  IE11  

#### Focus areas to test

Line chart Callout in IE 11


change/@uifabric-variants-2020-06-16-19-09-00-user-v-satgar-AddingPolyfill.json
```diff
@@ -0,0 +1,8 @@
+{
```
dzearing, line 1: something went crazy on the change files. I think you should manually delete the ones that don't apply. seems like you only care about the charting package being published here.

PR #13602, opened 2020-06-15
chore: move out Fela to a separate package

layershifter:
#### Pull request checklist

#### Description of changes

# BREAKING CHANGES

This PR changes design of `renderer` injection to Fluent UI Northstar. The `renderer` prop was removed from `Provider` component, now we accept `createRender` factory via a React context.

#### Before

```tsx
<Provider renderer={renderer} />
```

#### After

```tsx
<RendererContext.Provider value={() => renderer}>
  <Provider  />
</ProviderContext.Provider>
```

**(!)** There are no breaking changes if you never used the `renderer` prop on `Provider`.

## Details ❔ 

The previous design with `renderer` was created before we started to support multiple documents and blocks us from introducing custom renderers. That's why we decided to go with a factory that will allow us to create renderers on demand.

This PR also updates `Renderer` type and unifies its usage across Fluent Northstar code. This will allow us to introduce different renderers expect Fela. 

## Out of scope ✂️ 

This PR will not remove the dependency from `react-fela` and its `ThemeContext` which we are using everywhere. This thing should be tackled separately.


packages/fluentui/react-bindings/src/renderer/RendererContext.ts
```diff
@@ -0,0 +1,5 @@
+import { createFelaRenderer } from '@fluentui/react-northstar-fela-renderer';
+import { CreateRenderer } from '@fluentui/styles';
+import * as React from 'react';
+
+export const RendererContext = React.createContext<CreateRenderer>(createFelaRenderer);
```
layershifter, line 5: We decided to have only the context now without any additional hooks.

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
```diff
@@ -124,15 +125,13 @@ const resolveStyles = (
   // Our API should be aligned with it
   // Heads Up! Keep in sync with Design.tsx render logic
   const direction = rtl ? 'rtl' : 'ltr';
-  const felaParam: RendererParam = {
-    theme: { direction },
+  const rendererParam: RendererParam = {
+    direction,
```
layershifter, line 129: For a unified `RendererParam` it will be strange to pass `theme.direction`.

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
```diff
@@ -41,7 +43,6 @@ const stylesCache = new WeakMap<ThemePrepared, Record<string, ICSSInJSStyle>>();
 const resolveStyles = (
   options: ResolveStylesOptions,
   resolvedVariables: ComponentVariablesObject,
-  renderStylesInput?: (styles: ICSSInJSStyle) => string,
```
layershifter, line 44: This was used only for tests actually, but as we already using `renderRule` from `options.renderer` I decided to remove it to simplify the logic.

packages/fluentui/react-bindings/src/styles/types.ts
```diff
@@ -52,18 +57,6 @@ export type ComponentDesignProp = {
...
-  theme: { direction: 'ltr' | 'rtl' };
-  disableAnimations: boolean;
-  displayName: string;
-  sanitizeCss: boolean;
-};
-
-export type RendererRenderRule = (rule: () => ICSSInJSStyle, param: RendererParam) => string;
-export type Renderer = Omit<FelaRenderer, 'renderRule'> & {
-  renderRule: RendererRenderRule;
-};
```
layershifter, line 65: Move out to `@fluentui/styles`, not sure that it's a right place, but:
- it's definitely not a part of `react-bindings`
- it should be available for both renderer packages

packages/fluentui/react-bindings/src/styles/types.ts
```diff
@@ -73,28 +66,26 @@ export interface StylesContextPerformance {
...
   rtl?: boolean;
   disableAnimations?: boolean;
   performance?: StylesContextPerformanceInput;
-  renderer?: R;
+  renderer?: Renderer;
   theme?: ThemeInput;
 };
 
-export type StylesContextValue<R = Renderer> = {
+export type StylesContextValue = {
```
layershifter, line 77: Types were simplified as now we have a single `Renderer` type.

packages/fluentui/react-bindings/src/styles/types.ts
layershifter: Otherwise we can have a circular dependency:

```
@fluentui/react-northstar-emotion-renderer: error TS6202: Project references may not form a circular graph. Cycle detected: /mnt/work/1/s/packages/fluentui/react-northstar-emotion-renderer/tsconfig.json
@fluentui/react-northstar-emotion-renderer: /mnt/work/1/s/packages/fluentui/react-bindings/tsconfig.json
@fluentui/react-northstar-emotion-renderer: /mnt/work/1/s/packages/fluentui/react-northstar-fela-renderer/tsconfig.json
```

packages/fluentui/react-northstar-fela-renderer/src/felaExpandCssShorthandsPlugin.ts
```diff
@@ -1,6 +1,8 @@
 import { ICSSInJSStyle } from '@fluentui/styles';
 import * as CSS from 'csstype';
+// @ts-ignore
 import { expandProperty } from 'inline-style-expand-shorthand';
```
layershifter, line 4: This package doesn't have typings :(

packages/fluentui/react-northstar-fela-renderer/src/felaPerformanceEnhancer.ts
```diff
@@ -5,7 +5,7 @@
 //
 
 import { ICSSInJSStyle } from '@fluentui/styles';
-// eslint-disable-next-line import/no-extraneous-dependencies
+// @ts-ignore
 import cssifyDeclaration from 'css-in-js-utils/lib/cssifyDeclaration';
```
layershifter, line 9: No typings :(

packages/fluentui/react-northstar/package.json
```diff
@@ -22,19 +22,11 @@
     "classnames": "^2.2.6",
     "compute-scroll-into-view": "1.0.11",
     "downshift": "5.0.5",
-    "fela": "^10.6.1",
-    "fela-plugin-embedded": "^10.6.1",
-    "fela-plugin-fallback-value": "^10.6.1",
-    "fela-plugin-placeholder-prefixer": "^10.6.1",
-    "fela-plugin-rtl": "^10.6.1",
-    "fela-utils": "^10.6.1",
-    "inline-style-expand-shorthand": "^1.2.0",
```
layershifter, line 31: These deps were moved to `@fluentui/react-northstar-fela-renderer`

packages/fluentui/docs/src/components/VariableResolver/VariableResolver.tsx
```diff
@@ -33,11 +33,13 @@ const VariableResolver: React.FunctionComponent<VariableResolverProps> = props =
 
   useClassNamesListener(elementRef, onClassNamesChange);
 
-  return (
-    <Provider renderer={renderer}>
-      <div ref={elementRef}>{props.children}</div>
-    </Provider>
-  );
+  return null;
+  // TODO: fix this feature
```
layershifter, line 37: Variable resolver feature is now broken and should be fixed separately.

packages/fluentui/react-northstar-fela-renderer/src/felaDisableAnimationsPlugin.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Fela plugin for disabling animations. The animations are disabled or not based on the
+ * props' disableAnimations param. If the value of the prop is true, all animation related
+ * styles are removed.
+ *
+ * Caution! Infinite recursion is possible in case if style object has links to self in the props
+ * tree.
+ */
+const felaDisableAnimationsPlugin = (
```
layershifter, line 24: All plugins received a small refactor:
- were simplified where options are not required
- typings improved

Actually the main reason to do it: it's avoid fighting with typings.

packages/fluentui/react-northstar-fela-renderer/test/felaSanitizeCssPlugin-test.ts
```diff
@@ -63,16 +62,6 @@ describe('felaSanitizeCssPlugin', () => {
...
   });
 
-  test('should skip excluded CSS props', () => {
-    const style = {
-      display: 'block',
-      margin: '0 0 0 0',
-      propertyWithInvalidValue: 'rgba(',
-    };
-
-    expect(sanitize(style, { skip: ['propertyWithInvalidValue'] })).toEqual(style);
```
layershifter, line 73: This feature was removed to simplify the returned signature. Was never used anywhere.

packages/fluentui/react-northstar/src/utils/mergeProviderContexts.ts
```diff
@@ -1,34 +1,35 @@
...
 
-const registeredRenderers = new WeakMap<Document, Renderer>();
+const defaultDocument = { document: 'document' };
+const registeredRenderers = new WeakMap<Document | typeof defaultDocument, Renderer>();
 
-export const mergeRenderers = (current: Renderer, next?: Renderer, target?: Document): Renderer => {
-  if (next) {
-    return next;
-  }
+export const getOrCreateRenderer = (createRenderer: CreateRenderer, target?: Document): Renderer => {
```
layershifter, line 10: Not sure about naming there, but don't have better ideas.

packages/fluentui/react-bindings/src/hooks/useStyles.ts
```diff
@@ -9,12 +9,12 @@ import {
 import * as React from 'react';
 // @ts-ignore We have this export in package, but it is not present in typings
 import { ThemeContext } from 'react-fela';
+import { noopRenderer } from '@fluentui/styles';
```
mnajdova, line 12: I feel like we need a `@fluentui/react-northstar-renderer` package, or something similar...

packages/fluentui/react-bindings/src/hooks/useStyles.ts
```diff
@@ -9,12 +9,12 @@ import {
 import * as React from 'react';
 // @ts-ignore We have this export in package, but it is not present in typings
 import { ThemeContext } from 'react-fela';
```
mnajdova, line 11: Oops, should we still have `ThemeContext` from `react-fela` here?

layershifter: Yep, as we are receiving values from there. This will be a next and the last step of decoupling as we are using this context in our components, too.

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
```diff
@@ -124,15 +125,13 @@ const resolveStyles = (
   // Our API should be aligned with it
```
mnajdova, line 125: Should we remove/change this comment?

packages/fluentui/react-bindings/src/styles/types.ts
mnajdova: https://github.com/microsoft/fluentui/pull/13602/files#r440234899 ?

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
layershifter: Good catch, removed it and inlined `direction` to `rendererParam` object

packages/fluentui/react-bindings/src/styles/types.ts
```diff
@@ -73,28 +66,26 @@ export interface StylesContextPerformance {
...
   rtl: boolean;
   disableAnimations: boolean;
   performance: StylesContextPerformance;
-  renderer: R;
+  renderer: Renderer;
```
mnajdova, line 81: This means that if we want to support some other CSSinJS solution, we should add a wrapper render and that should be it right?

packages/fluentui/react-bindings/src/hooks/useStyles.ts
layershifter: Same feeling, but it looks strange to introduce a package for a single file...

packages/fluentui/react-bindings/src/styles/types.ts
layershifter: You're right/ We will to:
- implement a wrapper renderer that implements `Renderer`
- pass it via `RendererContext`

packages/fluentui/react-northstar-fela-renderer/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@fluentui/react-northstar-fela-renderer",
+  "description": "A CSS-in-JS renderer based on Fela for FluentUI React Northstar.",
+  "version": "0.49.0",
+  "author": "Oleksandr Fediashov <a@fedyashov.com>",
```
mnajdova, line 5: I don't think we are adding `author`

layershifter: It was copied from another package, removing it in this package

packages/fluentui/react-northstar/src/utils/mergeProviderContexts.ts
mnajdova: Just `getRenderer` should be fine in my opinion. We don't have to expose in the name the fact that we are creating it if it doesn't exists..

packages/fluentui/react-northstar/test/specs/utils/felaRenderer-test.tsx
```diff
@@ -1,16 +1,20 @@
...
+import * as React from 'react';
+
 import Animation from 'src/components/Animation/Animation';
-import Provider from 'src/components/Provider/Provider';
+import Box from 'src/components/Box/Box';
 import Text from 'src/components/Text/Text';
-import { felaRenderer } from 'src/utils';
+import { EmptyThemeProvider } from 'test/utils';
+
+const fluentRenderer = createFelaRenderer();
```
mnajdova, line 11: Should we rename this to `fluentRenderer-test.tsx`?

layershifter: I don't think so as it still tightly coupled to Fela. And it's impossible to move it `renderer` package as it's coupled with our components.

packages/fluentui/react-bindings/src/hooks/useStyles.ts
mnajdova: I think it would be nice if we can extract to a separate package: typings for the renderer + default implementation (even if it is the noopRenderer).. But anyway, not a deal breaker..

packages/fluentui/react-northstar/test/specs/utils/felaRenderer-test.tsx
layershifter: Agreed to move out this test to a `renderer` package.

layershifter: Fixed in 84a5987

packages/fluentui/CHANGELOG.md
```diff
@@ -69,7 +69,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Stop propagation of left and right arrow key in `SplitButton` for the integration with toolbar @jurokapsiar ([#13311](https://github.com/microsoft/fluentui/pull/13311))
 - Fix broken caching when `enableBooleanVariablesCaching:true` and there is no variables @layershifter ([#13345](https://github.com/microsoft/fluentui/pull/13345))
 - Fix displaying custom `clearIndicator` prop for `Dropdown` @kolaps33 ([#13486](https://github.com/microsoft/fluentui/pull/13486))
-
+- `renderer` prop on `Provider` was replaced with a custom React context @layershifter ([#13602](https://github.com/microsoft/fluentui/pull/13602))
```
miroslavstastny, line 72: breaking change?

layershifter: Oops, good catch, moved it to a proper part

packages/fluentui/react-bindings/tsconfig.json
```diff
@@ -25,6 +25,9 @@
     },
     {
       "path": "../react-component-ref"
+    },
+    {
+      "path": "../react-northstar-fela-renderer"
```
miroslavstastny, line 30: Just a question - do we have/can we add any way to enforce project references to be up-to-date?

layershifter: I don't know about it, but if there will be a missing reference it will throw.

packages/fluentui/react-northstar-fela-renderer/test/felaInvokeKeyframesPlugin-test.ts
```diff
@@ -1,35 +1,34 @@
-import felaInvokeKeyframesPlugin from 'src/utils/felaInvokeKeyframesPlugin';
-
-const renderInvokeKeyframes = felaInvokeKeyframesPlugin();
+import { ICSSInJSStyle } from '@fluentui/styles';
+import felaInvokeKeyframesPlugin from '../src/felaInvokeKeyframesPlugin';
```
mnajdova, line 2: Just out of curiosity why the paths were changed?

layershifter: We don't have (_and should not_) configured aliases under other packages, so we can't start with `src`. As this package hosts only things related to Fela I don't see any reasons to place files under `utils`.

packages/fluentui/react-northstar/src/utils/mergeProviderContexts.ts
layershifter: Renamed to `getRenderer`

packages/fluentui/react-northstar-fela-renderer/src/felaFocusVisibleEnhancer.ts
```diff
@@ -27,7 +12,7 @@ const felaFocusVisibleEnhancer = (renderer: Renderer) => {
   renderer._emitChange = (change: RendererChange) => {
     if (change.type === RULE_TYPE && change.selector.indexOf(':focus-visible') !== -1) {
       // Fela uses objects by references, it's safe to override properties
-      change.pseudo = change.pseudo ? change.pseudo.replace(':focus-visible', ':focus') : undefined;
+      change.pseudo = change.pseudo ? change.pseudo.replace(':focus-visible', ':focus') : '';
```
miroslavstastny, line 15: 🤔

layershifter: Strict typings are better as it should be a string and will default to an empty string:

https://github.com/robinweser/fela/blob/97148419e7941dbd5867b35df6ac4de6c5ed3e55/packages/fela/src/createRenderer.js#L224
https://github.com/robinweser/fela/blob/97148419e7941dbd5867b35df6ac4de6c5ed3e55/packages/fela/src/createRenderer.js#L302-L309

packages/fluentui/react-northstar/test/specs/utils/mergeProviderContexts/mergeProviderContexts-test.ts
```diff
@@ -1,22 +1,23 @@
-import mergeProviderContexts, { mergePerformanceOptions, mergeRenderers } from 'src/utils/mergeProviderContexts';
-import { felaRenderer } from 'src/utils/felaRenderer';
+import mergeProviderContexts, { mergePerformanceOptions, getRenderer } from 'src/utils/mergeProviderContexts';
+import { ProviderContextInput } from 'src/types';
 
-describe('mergeRenderers', () => {
-  test(`always uses "next" renderer`, () => {
-    const next = jest.fn();
-    expect(mergeRenderers(felaRenderer, next as any)).toBe(next);
+describe('getOrCreateRenderer', () => {
```
miroslavstastny, line 4:
```suggestion
describe('getRenderer', () => {
```

PR #13615, opened 2020-06-16
feat(FormPrototype): add form validation prototype

assuncaocharles:
#### Pull request checklist

#### Description of changes

Adds a Form Validation prototype with `Input` and `Dropdown`.

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -449,6 +460,16 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
 
+  handleOnBlur = (e: React.SyntheticEvent) => {
+    // Dropdown component doesn't present any `input` component in markup, however all of our
+    // components should handle events transparently.
+    if (e.target !== this.buttonRef.current) {
+      _.invoke(this.props, 'onBlur', e, this.props);
+    }
+  };
+
+  rootRef = React.createRef<HTMLElement>();
```
miroslavstastny, line 471: not used anywhere?

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -449,6 +460,16 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     _.invoke(this.props, 'onChange', e, { ...this.props, value: this.state.value });
   };
 
+  handleOnBlur = (e: React.SyntheticEvent) => {
+    // Dropdown component doesn't present any `input` component in markup, however all of our
+    // components should handle events transparently.
+    if (e.target !== this.buttonRef.current) {
+      _.invoke(this.props, 'onBlur', e, this.props);
```
miroslavstastny, line 467: nit: shouldn't we replace lodash with optional chaining? @layershifter

packages/fluentui/docs/src/components/Playground/renderConfig.ts
```diff
@@ -67,6 +69,16 @@ export const imports: Record<string, CodeSandboxImport> = {
     module: Classnames,
     required: false,
   },
+  formik: {
```
miroslavstastny, line 72: Why do you need this?
Do you expect `formik` to be used in anywhere in component example?

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
assuncaocharles: Removed

packages/fluentui/docs/src/components/Playground/renderConfig.ts
assuncaocharles: We are using formik in the prototype, not adding it here was throwing errors

assuncaocharles: removed

packages/fluentui/docs/package.json
```diff
@@ -26,6 +26,8 @@
     "copy-to-clipboard": "^3.2.0",
     "csstype": "^2.6.7",
     "faker": "^4.1.0",
+    "flat": "^4.1.0",
```
layershifter, line 29: Looks like a merge issue, please remove this one

packages/fluentui/docs/src/prototypes/FormValidation/FormHooks.tsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+  return (
+    <Form onSubmit={handleSubmit(onSubmit)}>
+      <Form.Input
+        label="First Name"
+        id="first-name-hooks"
+        required
+        name="firstName"
+        value={firstNameWatched}
+        inputRef={register({
```
layershifter, line 37:
```suggestion
        ref={register({
```

Why we are not using `ref` there?

assuncaocharles: It's was already in put, changing it would be a breaking change...?

packages/fluentui/docs/src/prototypes/FormValidation/FormHooks.tsx
```diff
@@ -0,0 +1,62 @@
...
+          required: { value: true, message: 'first name is required' },
+          validate: val => (val.length <= 4 ? 'Too Short' : null),
+        })}
+        errorMessage={errors.firstName?.message}
+        showSuccessIndicator={!errors.firstName && firstNameWatched?.length >= 4}
```
layershifter, line 42: I would like to avoid dependency on `firstNameWatched` in examples and use `dirty` instead:

```suggestion
        showSuccessIndicator={dirty && !errors.firstName}
```

packages/fluentui/docs/src/prototypes/FormValidation/FormHooks.tsx
assuncaocharles: Gonna handle it in separated PR

packages/fluentui/docs/src/prototypes/FormValidation/FormHooks.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const { register, handleSubmit, errors, setValue, watch, triggerValidation } = useForm({
+    mode: 'onBlur',
+  });
+
+  const onSubmit = data => {
+    console.log('errors: ', errors);
+    console.log('data: ', data);
+  };
+
+  const firstNameWatched = watch('firstName');
```
layershifter, line 15: I am not familiar with `react-hook-form`, but this breaks perf of the whole form as renders on each change:

image

Is there any chance to avoid it?

packages/fluentui/docs/src/prototypes/FormValidation/FormHooks.tsx
```diff
@@ -0,0 +1,62 @@
...
+      { name: 'city' },
+      {
+        required: { value: true, message: 'city is required' },
+      },
+    );
+  }, [register]);
+
+  const handleMultiChange = selectedOption => {
+    setValue('city', selectedOption);
+  };
```
layershifter, line 28:
```suggestion
```

packages/fluentui/docs/src/prototypes/FormValidation/FormHooks.tsx
```diff
@@ -0,0 +1,62 @@
...
+        inputRef={register({
+          required: { value: true, message: 'first name is required' },
+          validate: val => (val.length <= 4 ? 'Too Short' : null),
+        })}
+        errorMessage={errors.firstName?.message}
+        showSuccessIndicator={!errors.firstName && firstNameWatched?.length >= 4}
+      />
+      <Form.Dropdown
+        onChange={(e, { value }) => {
+          handleMultiChange(value);
```
layershifter, line 46:
```suggestion
          setValue('city', value);
```

packages/fluentui/docs/src/prototypes/FormValidation/FormHooks.tsx
```diff
@@ -0,0 +1,62 @@
...
+        name="firstName"
+        value={firstNameWatched}
+        inputRef={register({
+          required: { value: true, message: 'first name is required' },
+          validate: val => (val.length <= 4 ? 'Too Short' : null),
+        })}
+        errorMessage={errors.firstName?.message}
+        showSuccessIndicator={!errors.firstName && firstNameWatched?.length >= 4}
+      />
+      <Form.Dropdown
```
layershifter, line 44: I would like to have an additional round of discussion for this to simplify the usage.

1. We can have `Form.Select` to mimic a `select` control (`multiple` is forced to be `false`) =>
2. In this case we can easily render a hidden `input` with a value and forward ref to it =>
3. In this case the integration will be more simple and be done like with an `Input` via `ref`

packages/fluentui/docs/src/prototypes/FormValidation/Formik.tsx
```diff
@@ -0,0 +1,61 @@
+import * as React from 'react';
+import { Form } from '@fluentui/react-northstar';
+import { Formik } from 'formik';
+
+const FormValidateFormik = () => (
+  <Formik
```
layershifter, line 6: Just wondering why did you prefer this API over hooks? https://jaredpalmer.com/formik/docs/api/useFormik

yarn.lock
```diff
@@ -9685,6 +9685,13 @@ flat-cache@^2.0.1:
     rimraf "2.6.3"
     write "1.0.3"
 
+flat@^4.1.0:
```
layershifter, line 9688: Please rerun `yarn` after `flat` removal

packages/fluentui/docs/src/prototypes/FormValidation/FormHooks.tsx
assuncaocharles: removed

packages/fluentui/docs/src/prototypes/FormValidation/Formik.tsx
assuncaocharles: 2 reasons, internally our @types/react would have to be updated to support formik latest version and also just to have it as comparison agains other prototype that is already hooks.

yarn.lock
assuncaocharles: done

packages/fluentui/docs/src/prototypes/FormValidation/FormHooks.tsx
assuncaocharles: Talked with @jurokapsiar about this one... It's very common to have to handle this when using custom select, it's so that in the libraries examples this is the example default when using some select:

https://react-hook-form.com/get-started#WorkwithUIlibrary

I think we are ok with it like this for now and we can change latter if we agree on it?!

assuncaocharles: Agreed to solve it in the future.

PR #5536, opened 2018-07-12
Adding Pie chart and Donut chart to charting package 

RajeshGoriga:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [ ] Include a change request file using `$ npm run change`

#### Description of changes

Adding Pie chart and Donut chart to charting package 

#### Focus areas to test

(optional)

#### Screenshot for Donut chart
donut

#### Screenshot for Pie chart

piechart

PR #13618, opened 2020-06-16
Partially Port #5536 to calculate aria-posinset and aria-setsize in CommandBar.tsx to skip ContextualMenuItemType which is equal to Divider or Header in office-ui-fabric-react/5.79.1

Phoebe-0319:
#### Pull request checklist

- [ ] Include a change request file using `$ rush change`

#### Description of changes

Partially Port #5536 to calculate aria-posinset and aria-setsize in CommandBar.tsx to skip ContextualMenuItemType which is equal to Divider or Header in office-ui-fabric-react/5.79.1.

By default aria-posinset and aria-setsize are calculated regardless of ContextualMenuItemType. But in some case, we will encounter accessibility bugs about enumerating errors if ContextualMenu contains ContextualMenuItemType.Divider or ContextualMenuItemType.Header.

#### Focus areas to test

(optional)

### DIFF
--- a/common/changes/office-ui-fabric-react/hotfix_5-office-ui-fabric-react-5.79.1_2020-06-16-11-41.json
+++ b/common/changes/office-ui-fabric-react/hotfix_5-office-ui-fabric-react-5.79.1_2020-06-16-11-41.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "office-ui-fabric-react",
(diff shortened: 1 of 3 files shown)


dzearing:
I think this should be going into the 5.0 branch.

Phoebe-0319:
> I think this should be going into the 5.0 branch.

Yes, this change is going into the office-ui-fabric-react_v5.79.1 branch.

PR #13621, opened 2020-06-16
Icon: Rendering children if passed down as props and adding snapshot tests for these scenarios

tomi-msft:
#### Pull request checklist

#### Description of changes

Add logic to allow children prop to pass through to the Icon component, write unit tests for Icon component, and update icon snapshots

#### Focus areas to test

(optional)

### DIFF
--- a/change/office-ui-fabric-react-2020-06-16-09-19-45-react-icon-branch.json
+++ b/change/office-ui-fabric-react-2020-06-16-09-19-45-react-icon-branch.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Get children to pass through to the Icon component. Add Icon tests, and update snapshots for Icon tests",
+  "packageName": "office-ui-fabric-react",
+  "email": "ololubek@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-06-16T16:19:45.090Z"
+}
--- a/change/office-ui-fabric-react-2020-06-16-09-19-45-react-icon-branch.json
+++ b/change/office-ui-fabric-react-2020-06-16-09-19-45-react-icon-branch.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Get children to pass through to the Icon component. Add Icon tests, and update snapshots for Icon tests",
+  "comment": "Icon: Rendering children if passed down as props and adding snapshot tests for these scenarios.",
   "packageName": "office-ui-fabric-react",
   "email": "ololubek@microsoft.com",
   "dependentChangeType": "patch",
--- a/packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
@@ -65,7 +65,7 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
 
     return (
       <RootType data-icon-name={iconName} {...containerProps} {...nativeProps} className={classNames.root}>
-        {isImage ? <ImageType {...imageProps} /> : children ? children : iconContentChildren}
+        {isImage ? <ImageType {...imageProps} /> : children || iconContentChildren}
       </RootType>
     );
   }
--- a/packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
@@ -26,13 +26,13 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
   }
 
   public render() {
-    const { className, styles, iconName, imageErrorAs, theme } = this.props;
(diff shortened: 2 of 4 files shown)


change/office-ui-fabric-react-2020-06-16-09-19-45-react-icon-branch.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Get children to pass through to the Icon component. Add Icon tests, and update snapshots for Icon tests",
```
khmakoto, line 3:
```suggestion
  "comment": "Icon: Rendering children if passed down as props and adding snapshot tests for these scenarios.",
```

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
```diff
@@ -65,7 +65,7 @@ export class IconBase extends React.Component<IIconProps, IIconState> {
 
     return (
       <RootType data-icon-name={iconName} {...containerProps} {...nativeProps} className={classNames.root}>
-        {isImage ? <ImageType {...imageProps} /> : children}
+        {isImage ? <ImageType {...imageProps} /> : children ? children : iconContentChildren}
```
khmakoto, line 68: @dzearing @ecraig12345 I think there are two ways to think about this conditional. One is to do it the way this change is happening above and the other one is to do:

```ts
{children ? children : (isImage ? <ImageType {...imageProps} /> :  iconContentChildren)}
```

I wonder what are your opinion is on which of these two conditionals we should use.

packages/office-ui-fabric-react/src/components/Icon/Icon.test.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import * as renderer from 'react-test-renderer';
 
 import { Icon } from './index';
+import { TestImages } from 'office-ui-fabric-react/lib/common/TestImages';
```
khmakoto, line 5:
```suggestion
import { TestImages } from '@uifabric/example-data';
```

I think that will fix the issue you're having in the CI.

packages/office-ui-fabric-react/src/components/Icon/Icon.test.tsx
```diff
@@ -16,6 +17,22 @@ describe('Icon', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('renders Icon correctly as an image', () => {
+    const component = renderer.create(<Icon iconName="CompassNW" imageProps={{ src: TestImages.iconOne }} />);
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders Icon correctly that has children', () => {
```
khmakoto, line 26:
```suggestion
  it('renders Icon with children correctly', () => {
```

packages/office-ui-fabric-react/src/components/Icon/Icon.test.tsx
```diff
@@ -16,6 +17,22 @@ describe('Icon', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('renders Icon correctly as an image', () => {
```
khmakoto, line 20:
```suggestion
  it('renders Icon with imageProps correctly', () => {
```

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
tomi-msft: I think the first conditional is a little better from a readability perspective, because when you are reading through the ternary operator and evaluating it, you know that it will resolve to either the <ImageType> component or to the children that have been passed through to this component. Then it just becomes which children props it would be resolved to.

ecraig12345: It's not just a question of readability but of whether images can ever have children. If images *can't* have children, the conditional should stay as-is.

ecraig12345:
You should make this change in FontIcon too. Possibly also ImageIcon, but only if we think it can ever make sense for image icons to have children. (`` can't have children, but ImageIcon currently uses `<div></div>`.

packages/office-ui-fabric-react/src/components/Icon/Icon.base.tsx
dzearing: Why not just leave it and always add iconContentChildren?

dzearing: I think if Icon ever hand children and it conditionally rendered the children content based on nuance, it would be confusing. I don't see an issue just always doing it.

dzearing:
```
{ isImage ? <ImageType {...imageProps} /> : (children || iconContentChildren) }

dzearing:
@tomi-msft Nice job!

PR #13624, opened 2020-06-16
Commandbar: memoizing onRenderData

micahgodbolt:
Before memozing, resizeGroup would re-render the entire control on every resize of the window (even vertical). This is because on resize, component state changes and resizeGroup has to decide if it needs to remeasure.

before

After memoizing, resizeGroup still needs to re-render itself due to state change, but now the onRenderData method of the CommandBar is not re-evaluated unless there is a chance to the cache key. 

after


micahgodbolt:
The regression in command bar perf is interesting. Does the memo call add that much overhead on initial render?  Is that overhead worth while when considering all the additional renders that might happen over the component lifespan?

micahgodbolt:
@joschect does this memo use seem legit? it certainly helps reduce re-renders on screen resize

micahgodbolt:
Sweet! It's now passing. Getting caught up with master did the trick

micahgodbolt:
@ThomasMichon i saw you've been doing resizeGroup work lately. Have you tried using memo to reduce the re-render when screensize changes don't affect the contents of the resizeGrou?

JustSlone:
@micahgodbolt Thanks for the fix! 

Do you have a particular scenario this improves? Ideally with some before and after perf numbers showing the improvement this makes. Every fix has regression risk (especially with caching), so it's best to if we know the specific issue this is fixing.

packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.base.tsx
```diff
@@ -93,14 +93,24 @@ export class CommandBarBase extends React.Component<ICommandBarProps, {}> implem
     // TODO We may need to elevate classNames from <FocusZone> into <ResizeGroup> ?
     const nativeProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(this.props, divProperties);
 
+    // Given the same cache key, RenderedData will return the same memoized JSX Element,
+    // avoiding any re-renders of the control
+    const onRenderData = (props: ICommandBarData) => {
+      const compareData = (prevProps: ICommandBarData, nextProps: ICommandBarData) => {
+        return prevProps.cacheKey === nextProps.cacheKey;
```
JustSlone, line 100: It looks like the cacheKey on ICommandBarData does not include the farItems prop, however the _onRenderData function seems to use that prop, it seems like we might need to include farItems on compareData (note: I'm not an expert on this component).

micahgodbolt: i initially left far items out because they don't change over resize. The only time far items would change would be is user passed in new props. Guess the question is, would that invalidate the cache?

micahgodbolt:
closing this PR up for now. Can look at it again in the future if the topic comes up again.

PR #13626, opened 2020-06-16
Coachmark: Migrate Coachmark and associated components to function components

### LINKED ISSUE #14340: Coachmark: strict mode compliance (UNSAFE_)
Coachmark and related components have the following strict mode issues:
- Coachmark: `componentWillReceiveProps`
- PositioningContainer (a simplified fork of Callout):
  - `componentWillMount`
  - `componentWillUpdate`

MLoughry:
#### Pull request checklist

- [x] Fixes #14340

#### Description of changes

This change migrates `Coachmark` and its associated components `Beak` and `PositioningContainer` to function components in the `react-next` package.

#### Focus areas to test

(optional)


xugao:
I am concerned how we can better validate this change since there is almost no tests for coachmark

packages/react-next/etc/react-next.api.md
```diff
@@ -893,6 +857,8 @@ export interface IPositioningContainerProps extends IBaseProps<IPositioningConta
     ariaLabelledBy?: string;
     backgroundColor?: string;
     bounds?: IRectangle;
+    // Warning: (ae-forgotten-export) The symbol "React" needs to be exported by the entry point index.d.ts
+    children?: React_2.ReactNode;
```
xugao, line 861: this looks unexpected

MLoughry:
@xugao I've added a Screener test that needs approval

JustSlone:
@xugao can you take a look at the added screener tests?

apps/vr-tests/src/stories/CoachmarkNext.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+          ariaLabelledBy={'example-label1'}
+        >
+          Welcome to the land of Coachmarks!
+        </TeachingBubbleContent>
+      </Coachmark>
+    </>
+  );
+};
+
+storiesOf('Coachmark', module)
```
xugao, line 50:
```suggestion
storiesOf('Coachmark Next', module)
```

apps/vr-tests/src/stories/CoachmarkNext.stories.tsx
```diff
@@ -0,0 +1,71 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
+import * as React from 'react';
```
xugao, line 2: could you duplicate the this test for Coachmark in OUFR as well? similar to other tests we have. so we always have 2 results side by side

xugao:
@MLoughry - [this screener](https://screener.io/v2/states/Skgr32fUL.microsoft-fluentui/functional%2FCoachmark/1024x768/Chrome/coachmark-expanded-default-0) doesn't look correct to me

MLoughry:
> @MLoughry - [this screener](https://screener.io/v2/states/Skgr32fUL.microsoft-fluentui/functional%2FCoachmark/1024x768/Chrome/coachmark-expanded-default-0) doesn't look correct to me

Yes, there's a logic bug in the original OUFR control when you provide `isCollapsed={false}` to start with. I fixed the bug when I migrated the control.

apps/vr-tests/src/stories/Coachmark.stories.tsx
```diff
@@ -0,0 +1,71 @@
+/*! Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license. */
```
ecraig12345, line 1: Can you get rid of this header? (same in other files)

packages/react-next/src/components/Coachmark/Coachmark.types.ts
```diff
@@ -229,21 +233,11 @@ export interface ICoachmarkStyleProps {
    */
   isCollapsed: boolean;
 
-  /**
```
ecraig12345, line 232: Why were these props removed? Even in a major release, we typically try not to remove props that weren't previously marked as deprecated. Though if there's a strong argument for removal you could go ahead with it and also mark the props as deprecated in OUFR to provide at least a bit of warning.

Also, if you go ahead with removing these, you should add a note about that (and any other breaking changes) to react-next's RELEASE_NOTES.md. Same for other PRs. (We just decided on this procedure yesterday, which is why you haven't heard about it before.)

packages/react-next/src/components/Coachmark/Beak/Beak.tsx
```diff
@@ -8,78 +8,73 @@ import { RectangleEdge } from 'office-ui-fabric-react/lib/utilities/positioning'
...
-    super(props);
+export const Beak = React.forwardRef((props: IBeakProps, forwardedRef: React.Ref<HTMLDivElement>) => {
+  const { left, top, bottom, right, color, direction = RectangleEdge.top } = props;
 
-    initializeComponentRef(this);
-  }
-
-  public render(): JSX.Element {
-    const { left, top, bottom, right, color, direction = RectangleEdge.top } = this.props;
+  React.useImperativeHandle(props.componentRef, () => ({}), []);
```
ecraig12345, line 14: Is this supposed to do anything? If not you should probably just remove the componentRef prop (and deprecate it in v7).

packages/react-next/src/components/Coachmark/Coachmark.styles.ts
```diff
@@ -260,9 +258,12 @@ export function getStyles(props: ICoachmarkStyleProps): ICoachmarkStyles {
           },
         },
       },
+      !isMeasuring &&
```
ecraig12345, line 261: Why the change in logic here?

packages/react-next/src/components/Coachmark/Coachmark.base.tsx
```diff
@@ -34,126 +33,391 @@ export const COACHMARK_ATTRIBUTE_NAME = 'data-coachmarkid';
  * An interface for the cached dimensions of entity inner host.
  */
 export interface IEntityRect {
-  width: number;
-  height: number;
+  width?: number;
```
ecraig12345, line 36: Why are these optional now?

packages/react-next/src/components/Coachmark/Coachmark.base.tsx
```diff
@@ -34,126 +33,391 @@ export const COACHMARK_ATTRIBUTE_NAME = 'data-coachmarkid';
...
+  height?: number;
 }
 
-export interface ICoachmarkState {
+const DEFAULT_PROPS = {
```
ecraig12345, line 40:
```suggestion
const DEFAULT_PROPS: Partial<ICoachmarkProps> = {
```

packages/react-next/src/components/Coachmark/Coachmark.types.ts
MLoughry: They were removed as either unused or entirely redundant. `isBeaconAnimating` was always `true`, and `isMeasured` was always the inverse of `isMeasuring`.

packages/react-next/src/components/Coachmark/Coachmark.styles.ts
MLoughry: I don't remember the precise details (this work was done weeks ago), but IIRC the issue was that the measurement was now being taken such that without this change, it would determine that the size of the container to render into was 0x0, making it effectively invisible.

packages/react-next/src/components/Coachmark/Coachmark.base.tsx
MLoughry: I don't recall the precise details. Per the [commit message](https://github.com/MLoughry/office-ui-fabric-react/commit/26478ffa47d9330bf1cd165a3b17b8f3a4e0d2ee), this was to fix a bug (currently still in the OUFR version) where initializing the component with `isCollapsed={false}` would not show the expanded Coachmark.

MLoughry:
@ecraig12345 @xugao Could either of you approve the Screener additions?

ecraig12345:
It looks like someone rejected the "Coachmark: Expanded: Default" for current Coachmark because it's not actually showing the expanded coachmark. And in the CoachmarkNext expanded state, I'm not seeing the beak.

image

ecraig12345:
Actually never mind, I don't see a beak on the example on the website either, so maybe the CoachmarkNext one is okay. But I still think the Coachmark current expanded is wrong.
image

MLoughry:
@ecraig12345 I [replied](https://github.com/microsoft/fluentui/pull/13626#issuecomment-658305557) to @xugao about that very issue:
> Yes, there's a logic bug in the original OUFR control when you provide isCollapsed={false} to start with. I fixed the bug when I migrated the control.

ecraig12345:
Oh sorry, I missed that. Maybe we should remove the screener state for now so we're not checking in an incorrect state?

ecraig12345:
I accepted the CoachmarkNext expanded state.

packages/react-next/src/components/Coachmark/PositioningContainer/PositioningContainer.tsx
```diff
@@ -42,397 +42,391 @@ const SLIDE_ANIMATIONS = {
...
+const DEFAULT_PROPS = {
+  preventDismissOnScroll: false,
+  offsetFromTarget: 0,
+  minPagePadding: 8,
+  directionalHint: DirectionalHint.bottomAutoEdge,
+};
 
-  // @TODO rename to reflect the name of this class
-  private _contentHost = React.createRef<HTMLDivElement>();
+function useTargets({ target }: IPositioningContainerProps, positionedHost: React.RefObject<HTMLDivElement | null>) {
```
ecraig12345, line 52: This is optional, but since PositioningContainer is a fork of Callout, would it be possible to reuse any of the hooks from Callout here?

MLoughry: @ecraig12345 I'll do this in a PR fixing ContextualMenu's UNSAFE methods. Since this change has a `package.json` change that causes conflicts after every publish, I'd like to merge this today.

ecraig12345: Sounds good.

PR #13627, opened 2020-06-16
Fix Picker action button screen reader issue by updating activedescendant

### LINKED ISSUE #13290: Accessibility : Narrator is reading the 1st option, in the Tag Picker multiple times.
**Issue:**
Narrator is reading the 1st option, in the Tag Picker multiple times.

Issue can be reproducible in the below code pen links :
https://codepen.io/sagsah/pen/RwWxxpw?editors=1010
https://codepen.io/slimemold/pen/XWmOVpZ?editors=0010

**Browser Details: ​**
OS: Windows 10 (Version 1903, Build: 18362.175)
**Chromium Edge** Version 81.0.416.77 (Official build) (64-bit)

Please find the steps to repro the issue in the Tag Picker control:
1. Open Edge browser and switch on the Narrator.
2. Open link https://codepen.io/sagsah/full/RwWxxpw
3. Tab till "Filter items in suggestions" textfield and type "b".
4. Now, using down arrow, go to first option "black - multiple - colors - selected - here" as shown below. Do not press Enter.

Screenshot (112)
5. Listen what Narrator is reading.

**Actual:**
Narrator is reading the first option "black - multiple - colors - selected - here"  multiple times.

**Expected:**
It should read the first option only once, like reading the option 2 & 3.

Please find the steps to repro the issue in the People Picker control:
1. Open Edge browser and switch on the Narrator.
2. Open link https://codepen.io/slimemold/full/XWmOVpZ
3. Tab till text-field. The suggestion lists will get populated.
4. Now, using down arrow, go to first option "Annie Lindqvist" as shown below. Do not press Enter.

Screenshot (113)
5. Listen what Narrator is reading.

**Actual:**
Narrator is reading the first option "Annie Lindqvist"  multiple times.

**Expected:**
It should read the first option only once, like reading the option 2 & 3.

### LINKED ISSUE #13291: Accessibility : Screen reader is not reading the "searchForMoreText" prop value in the Tag Picker Suggestion pop up
**Browser Details: ​**
OS: Windows 10 (Version 1909, Build: 18363.778)
Microsoft Edge Dev Version 84.0.502.0 (Official build) dev (64-bit)
Chrome Version 81.0.4044.138 (Official Build) (64-bit)

**Issue:**
Screen reader is not reading the "searchForMoreText" prop value in the Tag Picker Suggestion pop up.

_**Note :** Issue can be reproducible in both Narrator/NVDA._

Please find the below code pen link to repro the issue:
https://codepen.io/sagsah/pen/RwWxxpw?editors=1010
https://codepen.io/slimemold/pen/XWmOVpZ?editors=0010

Please find the below NVDA speech logs. 
**Issue:** As you can see, when focus goes to "Load all results", screen reader is reading different message.
**Expected:** Screen reader should be reading "Load all results".

Screenshot (111)

smhigley:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13291 

#### Description of changes

There were two separate issues happening:
1. The action buttons in the Suggestions component (search for more and force resolve) did not have `id` attributes, and the picker input's `aria-activedescendant` attribute didn't point to them
2. The picker didn't rerender when use up/down arrows to move from an action button back to the suggestions list

This PR adds the id `sug-selectedAction` to the currently highlighted action button (managed by the Suggestions component). `BasePicker` is updated to set `aria-activedescendant` to `sug-selectedAction` when appropriate. It also adds an additional `this.forceUpdate()` to cause `activedescedant` to be recalculated and re-rendered when focus moves from the action button back to the options.

### DIFF
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
@@ -661,6 +661,7 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
           ) {
             ev.preventDefault();
             ev.stopPropagation();
+            this.forceUpdate();
           } else {
             if (
               this.suggestionElement.current &&
@@ -691,6 +692,7 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
           ) {
             ev.preventDefault();
             ev.stopPropagation();
+            this.forceUpdate();
           } else {
             if (
               this.suggestionElement.current &&
@@ -855,6 +857,11 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
 
   protected getActiveDescendant() {
     const currentIndex = this.suggestionStore.currentIndex;
+
+    if (currentIndex < 0 && this.suggestionElement.current && this.suggestionElement.current.hasSuggestedAction()) {
+      return 'sug-selectedAction';
+    }
+
     return currentIndex > -1 && !this.state.suggestionsLoading ? 'sug-' + currentIndex : undefined;
   }
 
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
@@ -856,8 +856,12 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
   };
 
   protected getActiveDescendant() {
-    const currentIndex = this.suggestionStore.currentIndex;
+    if (this.state.suggestionsLoading) {
+      return undefined;
+    }
 
+    const currentIndex = this.suggestionStore.currentIndex;
+    // if the suggestions element has actions and the currentIndex does not point to a suggestion, return the action id
     if (currentIndex < 0 && this.suggestionElement.current && this.suggestionElement.current.hasSuggestedAction()) {
       return 'sug-selectedAction';
     }
--- a/change/office-ui-fabric-react-2020-06-16-12-21-54-picker-actions-a11y.json
+++ b/change/office-ui-fabric-react-2020-06-16-12-21-54-picker-actions-a11y.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fix screen reader behavior on picker action buttons (#13291)",
+  "packageName": "office-ui-fabric-react",
+  "email": "sarah.higley@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-06-16T19:21:54.171Z"
+}
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.test.tsx
@@ -462,4 +462,51 @@ describe('BasePicker', () => {
 
     expect(getSuggestions(document)).toBeTruthy();
   });
+
+  it('navigates to search for more button', () => {
+    jest.useFakeTimers();
+    document.body.appendChild(root);
+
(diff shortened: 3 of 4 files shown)


packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -855,6 +857,11 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
 
   protected getActiveDescendant() {
     const currentIndex = this.suggestionStore.currentIndex;
+
+    if (currentIndex < 0 && this.suggestionElement.current && this.suggestionElement.current.hasSuggestedAction()) {
```
joschect, line 861: I think that there is a subtle bug, if  `this.state.suggestionsLoading` then `currentIndex < 0` will be true, because there are no suggestions. However, `hasSuggestedAction` will evaluate to true because the buttons are present in the dom. It's not a huge problem because the activeDescendant will point to a valid dom object, but it means that the buttons might get read out when it shouldn't.

It looks like if `this.suggestionsLoading` evaluates to true, then we could return undefined immediately.

joschect:
@smhigley Thanks so much for making this fix! It looks great! And thanks for adding some more tests too 👍  I made one comment but it looks like things should be good to go once that is addressed.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
smhigley: @joschect that makes sense to me! I'll add another `if` block at the top.

Another outside option is if the loading text were given an `id`, `activedescendant` could point to it when loading rather than use a live region to notify screen readers

I don't want to expand the scope of this PR too much, so I'll add `undefined` for now, but if you like that idea it could be a nice change in the future 😃

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -854,7 +856,16 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
...
 
   protected getActiveDescendant() {
+    if (this.state.suggestionsLoading) {
+      return undefined;
+    }
+
     const currentIndex = this.suggestionStore.currentIndex;
+    // if the suggestions element has actions and the currentIndex does not point to a suggestion, return the action id
+    if (currentIndex < 0 && this.suggestionElement.current && this.suggestionElement.current.hasSuggestedAction()) {
+      return 'sug-selectedAction';
```
ecraig12345, line 866: Can this and the other ID values go in a constant instead of being magic values? Also, what will happen if there's more than one picker on the page?

smhigley: @ecraig12345 thanks for the review! I followed the format of the picker item id values when I created this one -- they're also magic values of the format `sug-${index}` (controlled on line 869 immediately below this block).

I can't speak to the original reason that approach was chosen, but I'm guessing it's because the id is shared across the BasePicker and Suggestions components, and using a constant like `this._id` would require an additional property passed into the Suggestions component. On the plus side, multiple pickers per page shouldn't be an issue, since it shouldn't be possible to have more than one open at once.

Let me know if that answers your concern :)

JustSlone:
@joschect, @ecraig12345 looks like the PR has address the outstanding comments, can you take a look again?

joschect:
Closing and reopening to rebuild!

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
ecraig12345: That makes sense. Probably not the best design originally (the constants should at least be in a shared file somewhere) but this change doesn't make it significantly worse.

PR #13629, opened 2020-06-16
Add eslint configuration + expand eslint-plugin package

ecraig12345:
In preparation for actually switching to ESLint for v7, I'm adding some of the configuration separately so it's easier to review. (Any commented additions will be uncommented in the next PR, and all references to tslint removed at that point.)

#### Shared configs

Following common patterns for sharing ESLint configs, I moved our internal configuration variants into the `@fluentui/eslint-plugin` package and will make each package's config extend from that (both v0 and v7). This could also potentially be consumed by partners if they want. 

The configs are consumed like `"extends": ["plugin:@fluentui/eslint-plugin/react"]` where `react` is the name of the config within the package. These are the different configs (open to naming suggestions):
- `react`: v7/"converged" rules
  - Right now "converged" just means no `I` prefix for interfaces, but we'll eventually converge more rules
  - This is mostly a direct port of the v7 tslint config (to reduce noise), with the addition of the `import/no-extraneous-dependencies` rule
  - It extends from [`eslint-config-airbnb`](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb) like v0, to work towards convergence, but most of the added rules are disabled for now 
- `react--legacy`: v7 older rules (like `I` prefix on interfaces)
  - I don't really like this way of adding a modifier to the name, but couldn't think of something better. `react-legacy` sounds like it's for a *package* called `react-legacy` (which is inaccurate), and `react/legacy` doesn't work with tooling.
- `react-northstar`: existing v0 rules, moved from `scripts` plus added `ignorePatterns`
- `node` and `node--legacy`: like `react` but for packages which run in a Node environment

#### Custom rules

There weren't close enough ESLint equivalents of [`import-blacklist`](https://palantir.github.io/tslint/rules/import-blacklist/) from TSLint 
~~or `jsx-ban-props` from [`tslint-react`](https://github.com/palantir/tslint-react) ([`eslint-plugin-react`](https://github.com/yannickcr/eslint-plugin-react)'s `react/forbid-component-props` only works on components, not DOM elements)~~. *EDIT: David just removed jsx-ban-props from our TSLint in another PR.*

Since both of those rules are fairly simply to implement, I wrote custom replacements: `@fluentui/ban-imports` (renamed to avoid the term "blacklist") ~~and `@fluentui/jsx-ban-props`~~.

#### Performance concerns

A complication of this switch was that [`typescript-eslint` performance is slow](https://duncanleung.com/why-slow-plugin-typescript-eslint-performance-issues/) compared to TSLint--especially bad for rules which require type information, but still not great even for ones that don't. They're done some work to improve this (for example [this misleadingly named PR](https://github.com/typescript-eslint/typescript-eslint/pull/412)) but it's still not great. (related

Running with ESLint's [`--timing` option](https://eslint.org/docs/developer-guide/working-with-rules-deprecated#per-rule-performance) which shows the 10 slowest rules revealed that **the deprecation rule takes over 50% of runtime.** 😢  This is because it requires type info and has to do a bunch of complicated type inference stuff in TS for *most identifiers in the entire code*. It probably still adds enough value that we want to keep it, but it's at a significant cost.

The performance issues were most noticeable when running pre-commit (`lint-staged`), taking commit times up to ~10s! So based on [this comment](https://github.com/typescript-eslint/typescript-eslint/issues/389#issuecomment-524626231) I added logic to disable all rules which need type info while running `lint-staged` (mainly the deprecation rule), which makes it more comparable to the commit time with TSLint. Violations are still guaranteed to be caught in CI, even if the contributor doesn't notice them sooner via an editor extension or running `yarn lint`.

ESLint also has a [`--cache` option](https://eslint.org/docs/2.0.0/user-guide/command-line-interface#caching) we can try using (only lint things that changed since last lint run), which I have enabled experimentally for now. We can turn it off if it causes problems. The cache is stored in the project directory and therefore pollution in CI shouldn't be an issue.

### Project-level config

To avoid noise, I'm initially only including the updated configs for a few packages for demonstration purposes. Eventually all packages will have:
- a dev dependency on `@fluentui/eslint-plugin`
- `.eslintrc.json` or `.eslintrc.js` extending from `@fluentui/eslint-plugin/<something>`
- no `.eslintignore` since this can be handled in one place with `ignorePatterns` in the shared config

Note that I currently have ESLint set to run separately for each project, rather than on the whole monorepo. This may be possible to set up but not sure how hard it would be or which is preferable. ([related issue](https://github.com/typescript-eslint/typescript-eslint/issues/1681) with some info)

### Other configuration/scripts

- add `eslint` just task (not yet used)
- add eslint to lint-staged (commented for now)
- update editor config (commented for now)
- other minor script updates


ecraig12345:
Note that this isn't expected to build right now, but please review anyway (I have the other changes that will make it build ready to go, they'd just make the change hard to read).

packages/eslint-plugin/README.md
```diff
@@ -1,6 +1,18 @@
 # @fluentui/eslint-plugin
 
-**Custom ESLint rules for Fluent UI**
+**ESLint configuration and custom rules for Fluent UI**
+
+## Configs
+
+Usage: `{ "extends": ["plugin:@fluentui/<name>"] }` or `{ "extends": ["plugin:@fluentui/eslint-plugin/<name>"] }`
```
dzearing, line 7: Add this to... `eslint.config.js`?

packages/eslint-plugin/README.md
```diff
@@ -1,6 +1,18 @@
...
+
+- `react`: For `@fluentui/react` (`office-ui-fabric-react`) and related packages
+  - `react--legacy`: Like `react` but requiring an `I` prefix for interfaces
```
dzearing, line 10: I don't understand why this is here.

When we deprecate a rule, shouldn't we just add an exclusion to all the packages and then remove it over time?

packages/eslint-plugin/README.md
```diff
@@ -1,6 +1,18 @@
...
+  - `node`: Like `react` but for packages which run in a Node environment (not the browser)
+  - `node--legacy`: Like `node` but requiring an `I` prefix for interfaces
```
dzearing, line 12: Same as above; can we just not have this?

packages/eslint-plugin/README.md
```diff
@@ -1,6 +1,18 @@
...
+- `react-northstar`: For `@fluentui/react-northstar` and related packages
```
dzearing, line 13: Actually... same as above. Can we just have one mode and then per package exclusions that can be removed?

packages/eslint-plugin/src/configs/react.js
```diff
@@ -0,0 +1,324 @@
...
+    'react/no-string-refs': 'error',
+    'react/self-closing-comp': 'error',
+
+    // TODO: find equivalents to these tslint rules
+    // used in certain packages:
+    // "import-blacklist": [true, { "../../Styling": ["FontSizes"] }]
+
+    // airbnb or other config overrides (some temporary)
+    // TODO: determine which rules we want to enable, and make needed changes (separate PR)
+    'class-methods-use-this': 'off',
```
dzearing, line 115: A dev won't know what each of these do. It would be helpful to add docs on these. Separate PR is fine.

scripts/just.config.ts
```diff
@@ -95,7 +97,7 @@ module.exports = function preset() {
 
   task('lint', parallel('lint-imports', 'tslint'));
```
dzearing, line 98: will eslint be added here eventually?

packages/eslint-plugin/README.md
ecraig12345: Oh right...details... 😆

packages/eslint-plugin/README.md
ecraig12345: The biggest reason is that the rule which is different ([`@typescript-eslint/naming-convention`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md)) is a MASSIVE pain to configure because it covers way too many different things--to the point that I made a helper function getNamingConventionRule to handle the variations. (I *really* don't like this rule, but they've removed all the simpler alternatives...) So the alternative to the method you see here would be to make all the ESLint configs JS files and import the helper function. Having a "legacy" config to extend seemed closer to an idiomatic way to do it as far as I could tell from the ESLint ecosystem (though I could be wrong).

The other reason is that the number of different rules may increase over time (including possibly some rules with complex configuration requirements) if we start moving the main config towards the true converged set of rules we want but don't have time to update everything right away (or don't want to update certainly rules at all for legacy code due to philosophical differences). Having a "legacy" config to extend seemed easier than copying all the configuration everywhere.

packages/eslint-plugin/src/configs/react.js
ecraig12345: I don't think we need to add docs on every disabled rule right now. Eventually the goal would be to go through the rules and either enable them, or permanently disable, and at that point probably add a comment explaining why.

scripts/just.config.ts
ecraig12345: Yes, it will replace tslint on the final PR--I just didn't do that yet because it would cause tons of lint errors without the corresponding updates to local disables and such.

packages/office-ui-fabric-react/.eslintrc.js
```diff
@@ -0,0 +1,28 @@
...
+  rules: {
+    'react/forbid-component-props': 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
+    'import/no-extraneous-dependencies': [
+      'error',
+      {
+        devDependencies: [...configHelpers.devDependenciesFiles, 'src/common/{shallowUntilTarget,testUtilities}.ts'],
+      },
+    ],
+    // "import-blacklist": [true, { "../../Styling": ["FontSizes"] }]
```
khmakoto, line 17: Should we remove this?

ecraig12345: This is one that I haven't managed to find an eslint alternative for yet, so it was left there kind of as a reminder (though not a very useful one if I forget to follow up on it)

ecraig12345: I realized this rule would be pretty easy to reimplement, so I wrote a version which is now added to the PR. Same with `jsx-ban-props`.

ecraig12345:
Added alllllllllll the boilerplate. If you only care about seeing the last round of more interesting updates (including adding the custom rules) I'd recommend looking at 5e7c6a53fc5afc34679d223a027bd0a38a07935e.

PR #13631, opened 2020-06-16
Implement Pivot overflow: move tabs into a menu when they don't fit onscreen

### LINKED ISSUE #4066: Pivot: Support overflowing pivots

behowell:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #4066

#### Description of changes

Add a `useOverflow` hook that handles calculating if items fit within their container. Any item that doesn't fit within the container is moved to the overflow menu.

Attach the hook to the Pivot tabs, and update their styles to handle the overflow.

#### Focus areas to test

The Pivot control when the window is too narrow to fit all of the tabs.


xugao:
there seems a regression on PivotNext perf. it's not obvious to me what is causing it from the flamechart. but the ticks count went up quite significantly from the 2 perf-test runs we have on this PR

packages/react-next/src/components/Pivot/useOverflow.ts
```diff
@@ -0,0 +1,158 @@
...
+ */
+export const useOverflow = (
+  overflowMenuButtonRef: React.RefObject<HTMLElement | undefined>,
+  callbacks: OverflowCallbacks,
+) => {
+  let overflowIndex: number | undefined = undefined;
+
+  // Cache the computed styles from this render
+  let cachedStyle: { rtl: boolean; containerPadding: number; menuMargin: number } | undefined = undefined;
+  const getCachedStyle = (container: HTMLElement, menuButton: HTMLElement, win: Window) => {
```
xugao, line 52: use `React.useMemo`

packages/react-next/src/components/Pivot/useOverflow.ts
```diff
@@ -0,0 +1,158 @@
...
+      const requestId = win.requestAnimationFrame(updateOverflow);
+      win.addEventListener('resize', updateOverflowDebounced, false);
+
+      return () => {
+        win.cancelAnimationFrame(requestId);
+        win.removeEventListener('resize', updateOverflowDebounced, false);
+        resetOverflowItems(menuButton);
+      };
+    }
+  });
```
xugao, line 157: i dont think we need to run this hook on every render?
can't we have deps `[overflowMenuButtonRef.current]` here?

behowell: I'm looking into how to capture the dependencies. It's more than just overflowMenuButtonRef.current; it depends on at least:
* Which tab is selected (the selected tab never goes into the overflow)
* Whether the HTMLElements of any of the tabs has been re-rendered by React (which would reset the dynamically-set properties about whether it's in the overflow)
* The width of each tab item, which generally won't change, but technically depends on the header text and any styling of the tab header itself.

packages/react-next/src/components/Pivot/useOverflow.ts
behowell: useMemo doesn't really work here; the values need to be calculated once the refs have resolved, so it needs to be done during useLayoutEffect. I've updated useOverflow to cache some of the element's size and style info whenever useLayoutEffect runs. If I can improve the deps for useLayoutEffect, it'll end up being similar to useMemo.

packages/react-next/src/components/Pivot/useOverflow.ts
behowell: I looked into it some more, and it turns out that useEffect/useLayoutEffect don't work properly if you use `[ref.current]` as the dependency. It uses the ref's value from the _previous_ render, which ends up not re-running the effect when it should be.
The proper solution is to do all this work in a ref setter function instead of useLayoutEffect, so it gets triggered whenever the ref itself changes. That is complicated by the fact that we have to "listen" for multiple refs (the overflowMenu as well as any of the individual tab items), and coalesce all of those ref updates into a single recalculation after they've all been set. It's all doable but requires a pretty significant change to how it works now. I'm not sure I'll get to it.

packages/react-next/src/components/Pivot/observeResize.ts
```diff
@@ -0,0 +1,66 @@
...
+ *    (e.g. clientWidth/clientHeight or getBoundingClientRect).
+ *
+ * @returns A function to dispose the observer/listener.
+ */
+export const observeResize = (
+  target: Element | Element[],
+  onResize: (entries: readonly ResizeObserverEntry[] | undefined) => void,
+): (() => void) => {
+  if (ResizeObserver) {
+    const observer = new ResizeObserver(onResize);
```
xugao, line 22: feels wrong that we need to re-create resizeObserver every render

packages/react-next/src/components/Pivot/Pivot.tsx
```diff
@@ -8,13 +8,15 @@ import * as classes from './Pivot.scss';
...
-  linkIsSelected: 'is-selected',
   link: 'ms-Pivot-link',
+  linkInMenu: 'ms-Pivot-linkInMenu',
+  linkIsSelected: 'is-selected',
   linkContent: 'ms-Pivot-linkContent',
   root: 'ms-Pivot',
   rootIsLarge: 'ms-Pivot--large',
   rootIsTabs: 'ms-Pivot--tabs',
   text: 'ms-Pivot-text',
+  overflowMenuButton: 'ms-Pivot-overflowMenuButton',
```
khmakoto, line 19: nit: Move it up so that these are in alphabetical order.

packages/react-next/src/components/Pivot/useOverflow.ts
```diff
@@ -0,0 +1,127 @@
+import * as React from 'react';
+import { getWindow } from '../../Utilities';
+import { observeResize } from './observeResize';
+
+/**
+ * Callback to notify the user that the items in the overflow have changed. This should ensure that the overflow menu
+ * is visible, and contains all of the overflowing items.
+ *
+ * @param overflowIndex The index of the first item in the overflow, or items.length if nothing is overflowing.
+ * @param items The list of all items in the container, and whether that particular item is in the overflow
```
khmakoto, line 10: nit:
```suggestion
 * @param overflowIndex - The index of the first item in the overflow, or items.length if nothing is overflowing.
 * @param items - The list of all items in the container, and whether that particular item is in the overflow.
```

packages/react-next/src/components/Pivot/useOverflow.ts
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @param rtl - True if the element containing overflowMenuButtonRef is in right-to-left order
+ *
+ * @param pinnedIndexes - Optional: List of item indexes that should never go into the overflow menu.
+ */
+export const useOverflow = (
+  overflowMenuButtonRef: React.RefObject<HTMLElement | undefined> | null | undefined,
+  onOverflowItemsChanged: OverflowItemsChangedCallback,
+  rtl: boolean,
+  pinnedIndexes?: readonly number[],
```
xugao, line 44: I think it's better to change this to 1 param, that is `options: OverflowOptions`. then overtime, we can expand/shrink the params more easily
```
type OverflowOptions = { overflowMenuButtonRef ;  ... }
```

packages/react-next/src/common/_constants.scss
```diff
@@ -10,3 +10,5 @@ $animationDuration1: 0.167s;
 $animationDuration2: 0.267s;
 $animationDuration3: 0.367s;
 $animationDuration4: 0.467s;
+
+$CONTEXTUAL_MENU_ITEM_HEIGHT: 36px;
```
xugao, line 14: camelCase?

packages/react-next/src/components/Pivot/Pivot.types.ts
```diff
@@ -69,6 +69,16 @@ export interface IPivotProps extends React.HTMLAttributes<HTMLDivElement> {
    */
   linkFormat?: PivotLinkFormatType;
 
+  /**
+   * Overflow behavior when there is not enough room to display all of the links/tabs
+   *
+   * none: Pivot links will overflow the container and may not be visible
+   * menu: Display an overflow menu that contains the tabs that don't fit
```
xugao, line 76: i think we should move these comments to where we defined `PivotOverflowBehaviorType`

packages/react-next/src/components/Pivot/useOverflow.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    let prevOverflowIndex = items.length;
+    const setOverflowIndex = (overflowIndex: number) => {
+      if (prevOverflowIndex !== overflowIndex) {
+        prevOverflowIndex = overflowIndex;
+        onOverflowItemsChanged(
+          overflowIndex,
+          items.map((ele, index) => ({
+            ele,
+            isOverflowing: index >= overflowIndex && !pinnedIndexes?.includes(index),
```
xugao, line 109: use `indexOf` instead of `includes`: https://caniuse.com/#feat=array-includes

packages/react-next/src/components/Pivot/Pivot.base.tsx
```diff
@@ -190,15 +195,46 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef(
...
+    const renderedSelectedIndex = renderedSelectedKey ? linkCollection.keyToIndexMapping[renderedSelectedKey] : 0;
+
+    const items = linkCollection.links.map(l =>
+      renderPivotLink(linkCollection, l, renderedSelectedKey, classNames.link),
+    );
+
+    const overflowMenuButtonComponentRef = React.useRef<IButton>();
+    const overflowMenuButtonRef = React.useRef<HTMLElement>();
+    const setOverflowMenuButtonRef = (button: ActionButton | null) => {
+      const node = ReactDOM.findDOMNode(button);
```
shadowglen, line 214: This would not work once the buttons are migrated to function components in future, since ReactDOM.findDOMNode cannot be used on function components.

packages/react-next/src/common/_constants.scss
powerchelle, line 14: Can this value be influenced by the theme or the size of the pivot item?

packages/react-next/src/components/Pivot/Pivot.types.ts
behowell: IntelliSense only shows the comment from above the param itself, so it's more useful in practice to have it here. The type definition isn't even necessary, so I'll just remove that.

packages/react-next/src/common/_constants.scss
behowell: This needs to match CONTEXTUAL_MENU_ITEM_HEIGHT in ContextualMenu.cnstyles.ts. I'll add a comment to that effect.

packages/react-next/src/components/Pivot/useOverflow.ts
behowell: Thanks for catching that! I'm just going to change it to only support one pinnedIndex to keep it simpler, since the extra complexity of an array of indexes isn't necessary for now.

packages/react-next/src/components/Pivot/Pivot.base.tsx
behowell: The overflow algorithm won't work unless it can get to the DOM node to measure it. It looks like the new Button exposes a ref directly to the <button> Element, which is even better, and will simplify this code so it doesn't need to use findDOMNode. I'll leave this as-is for now until Pivot uses the new Button.

packages/react-next/src/common/_constants.scss
behowell: This is just copying a hardcoded value in the old Fabric ContextualMenu code (I'll add a comment about that), so it can't be changed for now. Eventually Pivot will be moved to use the new Button and ContextualMenu; I'm not sure if the new ContextualMenu will support theming this value.

packages/react-next/src/components/Pivot/observeResize.ts
behowell: This turned out to be challenging to solve because `useLayoutEffect` can't depend on a ref's current value, so I can't hook this up in a `useLayoutEffect` hook. I instead created a `useRefEffect` hook that works around this problem, so the ResizeObserver can reliably be created/destroyed only when the container ref changes.

packages/react-next/src/components/Pivot/useRefEffect.ts
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+
+/**
+ * Creates a ref, and calls a callback whenever the ref changes to a non-null value.
+ * The callback can optionally return a cleanup function that'll be called before the value changes or is set to null.
```
xugao, line 5:
```suggestion
 * The callback can optionally return a cleanup function that'll be called before the value changes.

```

i think can remove "or is set to null" here as it's also a value change

packages/react-next/src/components/Pivot/useRefEffect.ts
```diff
@@ -0,0 +1,45 @@
...
+ * https://github.com/facebook/react/issues/14387#issuecomment-503616820).
+ *
+ * @param callback - Called whenever the ref's value changes to non-null. Can optionally return a cleanup function.
+ * @param initialValue - (Optional) The initial value for the ref.
+ *
+ * @returns A tuple of `[ref, setRef]`:
+ * * `ref` - Read-only view of the ref's current value
+ * * `setRef` - A ref setter function that can be used to change the ref's current value.
+ */
+export function useRefEffect<T>(
```
xugao, line 17: I like it :)

let's put this in `react-hooks` package? would be awesome if you can add unit test too

packages/react-next/src/components/Pivot/useRefEffect.ts
```diff
@@ -0,0 +1,45 @@
...
+  callback: (value: T) => (() => void) | void,
+  initialValue: T | null = null,
+): [/*ref:*/ React.RefObject<T>, /*setRef:*/ (value: T | null) => void] {
+  type RefData = {
+    readonly ref: React.MutableRefObject<T | null>;
+    readonly setRef: (value: T | null) => void;
+    currentCallback: (value: T) => (() => void) | void;
```
xugao, line 24: nit: I think we should just call this `callback`. the "current" confuses me

packages/react-next/src/components/Pivot/useRefEffect.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const data = React.useRef<RefData>({
+    ref: { current: initialValue },
+    setRef: value => {
+      if (data.ref.current !== value) {
+        if (data.cleanup) {
+          data.cleanup();
+        }
+        data.ref.current = value;
+        data.cleanup = value !== null ? data.currentCallback(value) : undefined;
```
xugao, line 36: shouldn't we still call the callback when `value` is set to null ?

packages/react-next/src/components/Pivot/useRefEffect.ts
xugao: Feel free to do this in a follow-up PR.

packages/react-next/src/components/Pivot/useRefEffect.ts
behowell: The cleanup function can be used to do whatever you need to do when the ref is set to null.

React sets refs to `null` when the component is unmounted. That's how useRefEffect guarantees that `cleanup` will always run on unmount. If I call the callback when it's set to `null`, then the callback could return another `cleanup` function, which would never be run.

Thinking about it more, I could just run `cleanup` immediately after calling `callback` if the ref is null. Seems like potentially confusing behavior though. Thoughts?

xugao: Here is how i think (correct me if wrong) - the behavior here should be consistent with how `useEffect` behaves. for `useEffect`, every time deps changes, cleanup will be called and callback will be called. I'm curious when `useEffect` calls the final `cleanup`.

xugao: ic. `useEffect` calls the final clean up when component is unmounted. see: https://codesandbox.io/s/useeffect-cleanup-qjdmb

behowell: React sets refs to null when the component is unmounted. So the current behavior *is* aligned with `useEffect` that has a dependency on `[ref.current]`:
* The first time the ref is set to non-null, `callback` is called.
* When the ref's value changes (non-null), `cleanup` is called, then `callback` is called.
* When the ref's value changes to null (i.e. on unmount), then only `cleanup` is called.

See: https://codesandbox.io/s/misty-flower-sfbno?file=/src/App.tsx

packages/react-next/src/components/Pivot/useRefEffect.ts
behowell: Yeah, I was planning to move it to react-hooks in a future PR. Also to potentially move useOverflow to react-hooks as well, and observeResize to utilities.

PR #13640, opened 2020-06-17
feat(Datepicker): Extract utility functions from date-time/Calendar*

pompomon:
#### Pull request checklist

#### Description of changes

Extracting the following methods
Default formatters:

- `formatDay` 
- `formatMonthDayYear`
- `formatMonthYear`
- `formatYear`

Day grid generation:

- `getDayGrid`

Date comparison:

- `getBoundedDateRange`
- `getDateRangeArray`
- `isAfterMaxDate`
- `isBeforeMinDate`
- `isInDateRangeArray`
- `isRestrictedDate`

Other:
- `getDateRangeTypeToUse`


packages/date-time/etc/date-time.api.md
```diff
@@ -121,8 +123,8 @@ export interface ICalendarDayProps extends IBaseProps_2<ICalendarDay>, ICalendar
 // @public (undocumented)
 export interface ICalendarFormatDateCallbacks {
     formatDay: (date: Date) => string;
-    formatMonthDayYear: (date: Date, strings?: ICalendarStrings) => string;
-    formatMonthYear: (date: Date, strings?: ICalendarStrings) => string;
+    formatMonthDayYear: (date: Date, strings?: IGridStrings) => string;
```
pompomon, line 126: @lorejoh12 please I've limited the localization strings object we are sending to formatters to have only `days`, `months`, `shortDays` and `shortMonths` - would that be OK?
Or should I add some other fields there?

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.types.ts
```diff
@@ -8,12 +8,13 @@ import {
...
 } from '../Calendar/Calendar.types';
 import { IStyle, ITheme, IProcessedStyleSet } from '@uifabric/styling';
+import { IDayGridOptions } from '@fluentui/date-time-utilities/lib/dateGrid/DateGrid.types';
 
 export interface ICalendarDayGrid {
   focus(): void;
 }
 
-export interface ICalendarDayGridProps extends IBaseProps<ICalendarDayGrid> {
+export interface ICalendarDayGridProps extends IDayGridOptions, IBaseProps<ICalendarDayGrid> {
```
pompomon, line 17: I've kept all the fields present here for the comments with defaults, since I'm not providing any defaults in `date-time-utilities`, would that be OK?

packages/date-time/src/components/Calendar/Calendar.types.ts
```diff
@@ -3,6 +3,7 @@ import { IRefObject, IBaseProps, IStyleFunctionOrObject } from '@uifabric/utilit
 import { IStyle, ITheme } from '@uifabric/styling';
 import { ICalendarDayProps, ICalendarDayGridStyles } from './CalendarDay/CalendarDay.types';
 import { ICalendarMonthProps } from './CalendarMonth/CalendarMonth.types';
+import { IGridStrings } from '@fluentui/date-time-utilities/lib/dateGrid/DateGrid.types';
```
dzearing, line 6: Please use package imports rather than path based imports.

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.base.tsx
```diff
@@ -16,31 +15,28 @@ import {
...
   getDateRangeArray,
-  isInDateRangeArray,
   getWeekNumbersInMonth,
 } from '@fluentui/date-time-utilities/lib/dateMath/DateMath';
+import {
+  getDayGrid,
+  getBoundedDateRange,
+  findAvailableDate,
+  isRestrictedDate,
+} from '@fluentui/date-time-utilities/lib/dateGrid/DateGrid';
```
dzearing, line 26: same

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.types.ts
dzearing: sure i think that makes sense.

packages/date-time-utilities/src/dateGrid/DateGrid.test.ts
```diff
@@ -0,0 +1,717 @@
...
+  describe('getDayGrid', () => {
+    const defaultOptions: IDayGridOptions = {
+      selectedDate: new Date(2016, Months.Apr, 1),
+      navigatedDate: new Date(2016, Months.Apr, 1),
+      firstDayOfWeek: DayOfWeek.Sunday,
+      firstWeekOfYear: FirstWeekOfYear.FirstDay,
+      dateRangeType: DateRangeType.Day,
+    };
+
+    const defautlResult: IDay[][] = [
```
dzearing, line 50: It seems like this could be somewhat simplified by just taking a jest snapshot of the output (and validating it's what you typed here.)

packages/date-time/src/components/Calendar/Calendar.types.ts
pompomon: Oh, sure :)
I thought that /lib/ imports are preferred way, will update those :)

packages/date-time-utilities/src/dateGrid/DateGrid.ts
```diff
@@ -0,0 +1,184 @@
...
+export const isRestrictedDate = (date: Date, options: IRestrictedDatesOptions): boolean => {
+  const { restrictedDates, minDate, maxDate } = options;
+  if (!restrictedDates && !minDate && !maxDate) {
+    return false;
+  }
+  const inRestrictedDates =
+    restrictedDates &&
+    !!find(restrictedDates, (rd: Date) => {
+      return compareDates(rd, date);
+    });
```
ecraig12345, line 104: You can get rid of the uifabric/utilities reference here too (also in this case, switching to a more idiomatic way to achieve the desired result):
```suggestion
    restrictedDates.some((rd: Date) => compareDates(rd, date));
```

packages/date-time-utilities/src/dateGrid/DateGrid.test.ts
pompomon: Great idea, thank you :)

packages/date-time-utilities/src/dateGrid/DateGrid.ts
```diff
@@ -0,0 +1,184 @@
+import { IDay, IDayGridOptions, IRestrictedDatesOptions, IAvailableDateOptions, IGridStrings } from './DateGrid.types';
+import { addDays, getDateRangeArray, compareDates, isInDateRangeArray, compareDatePart } from '../dateMath/DateMath';
+import { find } from '@uifabric/utilities';
+import { DAYS_IN_WEEK, DateRangeType, DayOfWeek } from '../dateValues/DateValues';
+
+export const getDayGrid = (options: IDayGridOptions): IDay[][] => {
```
ecraig12345, line 6: One thing we need to resolve at some point is whether to use `const myFunc = () => { ... }` or `function myFunc() { ... }`. I very strongly prefer `function` (better for debugging, seems more semantically correct) and this has mostly been the convention in v7 aside from when creating React function components, but the convention in v0 appears to be using `const` for everything (I've tried to ask about this but nobody has ever explained why).

ecraig12345, line 6: Can you add a basic doc comment to each of the new functions?

pompomon: Sure, would add those :)

packages/date-time/etc/date-time.api.md
lorejoh12: it seems reasonable to me. If we need additional strings from the full object for later formatters we'll add them.

packages/date-time-utilities/src/dateGrid/DateGrid.ts
```diff
@@ -0,0 +1,184 @@
+import { IDay, IDayGridOptions, IRestrictedDatesOptions, IAvailableDateOptions, IGridStrings } from './DateGrid.types';
```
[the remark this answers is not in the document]
pompomon: Yeah, I'm OK with renaming the files, I was just following the examples of date utilities files like https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/utilities/dateMath/DateMath.ts
@dzearing , please WDYT?

packages/date-time-utilities/src/dateGrid/DateGrid.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    // we don't check shouldGetWeeks before pushing because we want to add one extra week for transition state
+    weeks.push(week);
+  }
+
+  return weeks;
+};
+
+/**
+ * Checks if the input @date param falls into the restricted options
```
ecraig12345, line 107: `@date` in the context of a doc comment could be misinterpreted by some tools as a JSDoc tag. Typically we just use code formatting like this (same in any other comment with `@someParam`):
```suggestion
 * Checks if `date` falls into the restricted options
```

packages/date-time-utilities/src/dateGrid/DateGrid.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  return dateRangeType;
+};
+
+/**
+ * Returns closest available date given the restriction options, or undefined otherwise
+ * @param options - list of options
+ * initialDate - date from which we start the search
```
ecraig12345, line 189: These should be documented in the options interface rather than here.

packages/date-time-utilities/src/dateGrid/DateGrid.ts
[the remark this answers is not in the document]
dzearing: So the convention is to try and match the file with the content.TitleCase implies a class or component. A lot of the older code was class based. I would suggest for all files which only contain functional helpers we stick with camelCasing and be done. :)

I would also suggest that we try and have 1 function per file (with exceptions for types or non-exported helpers) and match the name. Reasoning is that you can visually scan: is there a .test file which covers that helper?

So:

getDayGrid.ts
getDayGrid.test.ts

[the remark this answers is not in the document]
ecraig12345: That's kind of what I was thinking too--though I'd add that if you don't have time to move things around before going on vacation, it's okay to just merge with the current filename (to keep other changes moving) and move things when you get back.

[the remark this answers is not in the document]
pompomon: OK, sure, will update the names and split the functions then
Thank you!

PR #13645, opened 2020-06-17
Persona: Migrate Persona and PersonaCoin to function components

MLoughry:
#### Pull request checklist

#### Description of changes

Migrate `Persona` and `PersonaCoin` to function components in the `react-next` package

#### Focus areas to test

(optional)


packages/react-next/src/components/Persona/PersonaCoin/PersonaCoin.test.tsx
```diff
@@ -70,12 +68,4 @@ describe('PersonaCoin', () => {
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
-
-  it('sets isImageLoaded to false if imageUrl changes from valid to undefined', () => {
```
xugao, line 74: instead of removing the test, can you add a replacement? maybe something like - when imageUrl changes,`onPhotoLoadingStateChange` gets called with the correct value

MLoughry: That would ultimately be testing `Image`, not `PersonaCoin`.

packages/react-next/src/components/Persona/examples/Persona.PresenceColor.Example.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react';
-import { IPersonaSharedProps, Persona, PersonaSize, PersonaPresence } from '@fluentui/react-next/lib/Persona';
-import { Label } from '@fluentui/react-next/lib/Label';
-import { Stack } from '@fluentui/react-next/lib/Stack';
-import { mergeStyles } from '@fluentui/react-next/lib/Styling';
+import { IPersonaSharedProps, Persona, PersonaSize, PersonaPresence } from '@fluentui/react-next';
+import { Label } from 'office-ui-fabric-react/lib/Label';
```
dzearing, line 3: no need to change back to oufr for these imports.

dzearing: (there are a number of them here)

PR #13651, opened 2020-06-17
Scrub and pre-commit fixes

ecraig12345:
The scrub script needed a few fixes to work better on Windows:
- Use `npx.cmd` not `npx` on Windows
- Check `./node_modules` and `packages/fluentui/*/node_modules` for symlinks before running `git clean`

Also add another case to the pre-commit error message for if lint-staged isn't installed at all (because they need to run `yarn`).

### DIFF
--- a/scripts/scrub.js
+++ b/scripts/scrub.js
@@ -2,6 +2,7 @@
 
 const child_process = require('child_process');
 const fs = require('fs');
+const os = require('os');
 const path = require('path');
 // This script MUST NOT have any deps aside from Node built-ins, because it deletes all node_modules!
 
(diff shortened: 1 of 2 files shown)


scripts/scrub.js
```diff
@@ -53,8 +54,11 @@ function deleteNodeModulesSymlinks(packagePath, failedPaths) {
     return;
   }
   // Check node_modules for symlinks and manually remove those
+  // (odd way of iterating is since we're adding more modules to the list as we go)
```
khmakoto, line 57:
```suggestion
  // (using this odd way of iterating since we're adding more modules to the list as we go)
```

Maxxisthename:
> The scrub script needed a few fixes to work better on Windows:
> - Use `npx.cmd` not `npx` on Windows
> - Check `./node_modules` and `packages/fluentui/*/node_modules` for symlinks before running `git clean`
> 
> Also add another case to the pre-commit error message for if lint-staged isn't installed at all (because they need to run `yarn`).

PR #13656, opened 2020-06-18
@fluentui/react-conformance - adding isConformant() example to slider and fixing bugs 

sopranopillow:
#### Pull request checklist

#### Description of changes

This PR adds the new isConformant test to the Slider component in the following packages: react-next, react-northstar, and office-ui-fabric-react. This PR also fixes the following bugs in @fluentui/react-conformance:
* 'has-docblock': Instead of checking for words, the test was checking for characters.
* 'component-renders': Test now allows custom mount, specifically for react-northstar
* 'has-displayName': Now handles wrapped and styled components
* isConformant(): Receives and run extra tests and tests can be opted out if needed.


packages/react-conformance/src/defaultTests.tsx
```diff
@@ -45,16 +47,18 @@ export const defaultTests: TestObject = {
...
-      it(`has a displayName`, () => {
-        expect(Component.displayName).toEqual(testInfo.displayName);
-      });
-    }
+
+    it(`has a displayName or constructor name`, () => {
+      const constructorName = Component.prototype?.constructor.name;
+      let displayName = Component.displayName || constructorName;
+
+      if (constructorName === 'Wrapped') {
```
ecraig12345, line 55: Can you add a comment explaining what case this is for?

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -13,14 +13,14 @@ export const defaultTests: TestObject = {
...
-      expect(_.words(componentInfo.description)).toBeLessThanOrEqual(maxWords);
+      expect(_.words(componentInfo.description).length).toBeGreaterThanOrEqual(minWords);
+      expect(_.words(componentInfo.description).length).toBeLessThanOrEqual(maxWords);
     });
   },
 
   /** Component file exports a valid React element type  */
   'exports-component': (componentInfo: ComponentDoc, testInfo: TestingOptions) => {
-    it(`exports component from file under correct name`, () => {
+    it(`exports component from file under correct name ${testInfo.componentPath}`, () => {
```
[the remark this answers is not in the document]
sopranopillow: Was testing something and forgot to remove it 😅

packages/fluentui/react-northstar/test/specs/components/Slider/Slider-test.tsx
```diff
@@ -1,7 +1,18 @@
...
+import { isConformant as newIsConformant } from '@fluentui/react-conformance';
+import { mountWithProvider as mount } from 'test/utils';
 import Slider from 'src/components/Slider/Slider';
 
 describe('Slider', () => {
+  newIsConformant({
+    componentPath: __dirname.replace(/test.*/, 'src/components/Slider/Slider.tsx'),
+    Component: Slider,
+    displayName: 'Slider',
+    isInternal: false,
```
ecraig12345, line 11: You can remove this since the default is false

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -45,16 +47,20 @@ export const defaultTests: TestObject = {
...
-        expect(Component.displayName).toEqual(testInfo.displayName);
-      });
-    }
+
+    it(`has a displayName or constructor name`, () => {
+      const constructorName = Component.prototype?.constructor.name;
+      let displayName = Component.displayName || constructorName;
+
+      // This check is needed in case the Component is wrapped therefore the constructor's name is Wrapped,
+      // so we have to look at the display name and remove Base or Styled if the component was styled or is a base
```
ecraig12345, line 56: I'd go with something like this to be clear that the wrapping is using styled():
```suggestion
      // This check is needed in case the Component is wrapped with the v7 styled() helper, which returns a wrapper
      // component with constructor name Wrapped, and adds a Styled prefix to the displayName. Components passed to
      // styled() typically have Base in their name, so remove that too.
```

PR #6244, opened 2018-09-05
Commandbar: remove invalid aria attributes

kenotron:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #6232 
- [x] Include a change request file using `$ npm run change`

#### Description of changes

Commandbar: remove invalid aria attributes
- removed function that modifies the aria attribs
- kept the cacheKey calculation
- adding back the clone array functionality of set aria function

#### Focus areas to test

(optional)

Issue #6232, opened 2018-09-05
Remove aria-setsize and aria-posinset from CommandBar


micahgodbolt:
There has been a longstanding conflict between a bug found in Keros due to a fixed added to solve a narrator bug (fun!). 

The contested change was adding aria-setsize and aria-posinset to the command bar.

From what I've seen the only vote "for" the props is from an accessibility vendor opening a bug that states

> SizeOfSet & PositionInSet" value is not programmatically defined for the menu items

This was 7 months ago, and fixed in the original CommandBar. I have since ported the change to the new CommandBar and we've gotten several different teams reporting this as being flagged in Keros, and incorrect.

On top of that, the specification for aria-setsize/posinset states that it only needs to be present if not all of the items in the list are in the DOM at the same time (i.e. the list starts at item 7 of 16)

![image](https://user-images.githubusercontent.com/1434956/45099274-bbce8380-b0db-11e8-9e17-9afbdbe50bb7.png)

So with all of this research, and agreement from the the dev who initially added aria-setsize, I'm proposing we remove aria-setsize/posinset from command bar. (these values can always be added manually if required).

Removing these properties will also remove a bug where we are mutating props to add these values in, so it'll be great to not have to do that anymore.

PR #13681, opened 2020-06-18
Partially Port #5536 to calculate aria-posinset and aria-setsize in CommandBar.tsx to skip ContextualMenuItemType which is equal to Divider or Header in office-ui-fabric-react/5.0

Phoebe-0319:
#### Pull request checklist

- [ ] Include a change request file using `$ rush change`

#### Description of changes

CommandBar: Caculate aria-posinset and aria-setsize to skip ContextualMenuItemType which is equal to Divider or Header

#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.tsx
+++ b/packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.tsx
@@ -10,7 +10,7 @@ import {
 } from '../../Utilities';
 import { ICommandBar, ICommandBarProps, ICommandBarItemProps } from './CommandBar.types';
 import { IFocusZone, FocusZone, FocusZoneDirection } from '../../FocusZone';
-import { ContextualMenu, IContextualMenuProps, IContextualMenuItem } from '../../ContextualMenu';
+import { ContextualMenu, IContextualMenuProps, IContextualMenuItem, ContextualMenuItemType } from '../../ContextualMenu';
 import { hasSubmenu } from '../../utilities/contextualMenu/index';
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.tsx
```diff
@@ -538,4 +538,27 @@ export class CommandBar extends BaseComponent<ICommandBarProps, ICommandBarState
...
+        }
+
+        ++totalItemCount;
+      }
+    }
+
+    return totalItemCount;
+  }
+
+  private _needReviseAriaPosinset(item: IContextualMenuItem, indexCorrection: number): boolean {
```
joschect, line 557: Did you mean to have `indexCorrection` handled here?

joschect:
@Phoebe-0319 It looks like the #5536 doesn't point to a pull request related to the commandbar. I found a few other ones but it looks like ultimately the forward fix was to remove posinset entirely.  See issue #6232 and pr #6244 .

@micahgodbolt might be able to speak to more why that removal makes the most sense but we might want to consider removing all the posinset from command bar.

Phoebe-0319:
> @Phoebe-0319 It looks like the #5536 doesn't point to a pull request related to the commandbar. I found a few other ones but it looks like ultimately the forward fix was to remove posinset entirely. See issue #6232 and pr #6244 .
> 
> @micahgodbolt might be able to speak to more why that removal makes the most sense but we might want to consider removing all the posinset from command bar.

Thank you @joschect . I will abandon this PR and send another PR to remove posinset entirely through porting pr #6244.

PR #13683, opened 2020-06-18
LineChart: import find from office-ui-fabric-react/lib/Utilities to support IE11

RajeshGoriga:
#### Pull request checklist

#### Description of changes

LineChart: Add polyfill to support array.find method in callout in IE11

#### Focus areas to test

Line chart Callout in IE11


yarn.lock
```diff
@@ -6146,14 +6146,6 @@ chainsaw@~0.1.0:
   dependencies:
     traverse ">=0.3.0 <0.4"
 
-chalk@^4.0.0:
```
ecraig12345, line 6149: Not sure why this changed. If the CI build fails you may have to manually revert it. (`git checkout master -- yarn.lock`)

PR #13687, opened 2020-06-18
Keytips: storybook example fixes

xugao:
#### Pull request checklist

#### Description of changes
1. Added render `KeytipLayer` using storybook decorator as Keytips are actually shown when using storybook examples
2. Added `keytipProps` as part of handled props for link

#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/.storybook/decorators/withKeytipLayer.js
+++ b/packages/office-ui-fabric-react/.storybook/decorators/withKeytipLayer.js
@@ -1,0 +1,11 @@
+import React from 'react';
+import { KeytipLayer } from 'office-ui-fabric-react';
+
+export const withKeytipLayer = storyFn => {
+  return (
+    <>
+      <KeytipLayer content="Alt Windows" />
+      {storyFn()}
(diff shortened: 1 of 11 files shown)


packages/office-ui-fabric-react/.storybook/decorators/withKeytipLayer.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { KeytipLayer } from 'office-ui-fabric-react';
+
+export const withKeytipLayer = storyFn => {
```
ecraig12345, line 4: Should we put this in a shared location so other packages can use it too? Or do we only care about the current version of keytips working in OUFR?

xugao: maybe, but current use cases do not require this to be duplicated (hence, to be shared). 
We might need a new package to put shared decorators (I don't think `/scripts` is the right place). I'd to postpone this until we actually use this in multiple places

PR #13688, opened 2020-06-18
KeytipData: convert to function component

xugao:
#### Pull request checklist

#### Description of changes
1.  Convert `KeytipData` to function component and extracted out `useKeytipData` hook to use in V8 for removing `KeytipData` from individual component. Note since we don't support `componentRef` on `KeytipData`, this change should be safe to make in V7.

2. I removed `IBaseProps<IKeytip>` (`componentRef`) from `IKeyDataProps` because it's not being initialized inside `KeytipData` component.

3. Added `usePrevious` hook

4. Avoid render `KeytipData` component when `keytipProps` is not defined

#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/KeytipData/useKeytipData.ts
+++ b/packages/office-ui-fabric-react/src/components/KeytipData/useKeytipData.ts
@@ -3,7 +3,7 @@ import { useConst } from '@uifabric/react-hooks';
 import { mergeAriaAttributeValues } from '../../Utilities';
 import { IKeytipDataProps } from './KeytipData.types';
 import { IKeytipProps } from '../../Keytip';
-import { KeytipManager, mergeOverflows, sequencesToID, getAriaDescribedBy } from '../../utilities/keytips';
+import { KeytipManager, mergeOverflows, sequencesToID, getAriaDescribedBy } from '../../utilities/keytips/index';
 
 export type KeytipDataOptions = IKeytipDataProps;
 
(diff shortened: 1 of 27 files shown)


packages/office-ui-fabric-react/src/components/KeytipData/useKeytipData.ts
```diff
@@ -0,0 +1,87 @@
...
+  keytipProps: IKeytipProps,
+  describedByPrepend?: string,
+): IKeytipData {
+  // Add the parent overflow sequence if necessary
+  const newKeytipProps = keytipManager.addParentOverflow(keytipProps);
+
+  // Construct aria-describedby and data-ktp-id attributes
+  const ariaDescribedBy = mergeAriaAttributeValues(
+    getAriaDescribedBy(newKeytipProps.keySequences),
+    describedByPrepend,
```
khmakoto, line 69: I don't know if it's relevant but you're switching the order here from the original.

packages/office-ui-fabric-react/src/components/KeytipData/useKeytipData.ts
```diff
@@ -0,0 +1,87 @@
+import * as React from 'react';
+import { useConst } from '@uifabric/react-hooks';
+import { mergeAriaAttributeValues } from '../../Utilities';
+import { IKeytipDataProps } from './KeytipData.types';
+import { IKeytipProps } from '../../Keytip';
+import { KeytipManager, mergeOverflows, sequencesToID, getAriaDescribedBy } from '../../utilities/keytips';
```
khmakoto, line 6:
```suggestion
import { KeytipManager, mergeOverflows, sequencesToID, getAriaDescribedBy } from '../../utilities/keytips/index';
```

PR #13712, opened 2020-06-19
useMergedRefs should return an immutable function

dzearing:
```tsx
const ref1 = React.useRef();
const ref2 = React.useRef();

return <div ref={useMergedRefs(ref1, ref2)} />;
```

In the current code, the ref function mutates every time, causing prop mutation unnecessarily. Tweaking the function to return the same function every time.

### DIFF
--- a/packages/react-hooks/src/useMergedRefs.test.tsx
+++ b/packages/react-hooks/src/useMergedRefs.test.tsx
@@ -1,8 +1,31 @@
 import * as React from 'react';
-import { mount } from 'enzyme';
+import { mount, ReactWrapper } from 'enzyme';
 import { useMergedRefs } from './useMergedRefs';
 
 describe('useMergedRefs', () => {
+  let wrapper: ReactWrapper | undefined;
+
+  afterEach(() => {
+    wrapper?.unmount();
+    wrapper = undefined;
+  });
+
+  it('always returns the same ref (refs should be immutable)', () => {
+    let lastMergedRef;
+
+    const TestComponent: React.FunctionComponent = () => {
+      lastMergedRef = useMergedRefs<boolean>(() => ({}));
+      return null;
+    };
+
+    wrapper = mount(<TestComponent />);
+    const ref1 = lastMergedRef;
+    wrapper.setProps({});
+    const ref2 = lastMergedRef;
+
+    expect(ref1).toBe(ref2);
+  });
+
   it('updates all provided refs', () => {
     const refObject: React.RefObject<boolean> = React.createRef<boolean>();
     let refValue: boolean | null = null;
@@ -11,7 +34,7 @@ describe('useMergedRefs', () => {
       mergedRef(true);
       return null;
     };
-    mount(<TestComponent />);
+    wrapper = mount(<TestComponent />);
 
     expect(refObject.current).toBe(true);
     expect(refValue).toBe(true);
@@ -29,7 +52,7 @@ describe('useMergedRefs', () => {
       return null;
     };
 
-    const wrapper = mount(<TestComponent />);
+    wrapper = mount(<TestComponent />);
 
     const firstRefCallback = refCallback;
 
@@ -51,7 +74,7 @@ describe('useMergedRefs', () => {
       return null;
     };
 
-    const wrapper = mount(<TestComponent />);
+    wrapper = mount(<TestComponent />);
 
     let secondRefValue: boolean | null = null;
     refValueFunc = (val: boolean) => (secondRefValue = val);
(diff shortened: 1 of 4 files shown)


packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,19 +1,27 @@
...
 /**
  * React hook to merge multiple React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
  * updates all provided refs
  * @param refs- Refs to collectively update with one ref value.
  */
-export function useMergedRefs<T>(...refs: React.Ref<T>[]): (instance: T) => void {
-  return React.useCallback((value: T) => {
-    refs.forEach(ref => {
+export function useMergedRefs<T>(...refObjects: React.Ref<T>[]): (instance: T) => void {
+  const [state] = useState<{ refs: (Ref<T> | undefined)[] }>({ refs: [] });
```
MLoughry, line 9: Why not use a ref instead of state?

dzearing: is there a benefit?

dzearing: That's cleaner, updated!

packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,19 +1,25 @@
...
  * updates all provided refs
  * @param refs- Refs to collectively update with one ref value.
  */
-export function useMergedRefs<T>(...refs: React.Ref<T>[]): (instance: T) => void {
-  return React.useCallback((value: T) => {
-    refs.forEach(ref => {
+export function useMergedRefs<T>(...refObjects: Ref<T>[]): (instance: T) => void {
+  const state = useRef<(Ref<T> | undefined)[]>();
+
+  // Update refs list in immutatable state object.
```
MLoughry, line 11: Nit: comment is wrong :P

packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,19 +1,25 @@
...
+  state.current = refObjects;
+
+  // Return cached callback which refers to mutable refs array within the immutable state.
```
MLoughry, line 14: Nit: same thing with comment here

packages/react-hooks/src/useMergedRefs.test.tsx
```diff
@@ -1,8 +1,31 @@
...
+    let lastMergedRef;
+
+    const TestComponent: React.FunctionComponent = () => {
+      lastMergedRef = useMergedRefs<boolean>(() => ({}));
+      return null;
+    };
+
+    wrapper = mount(<TestComponent />);
+    const ref1 = lastMergedRef;
+    wrapper.setProps({});
```
xugao, line 23: don't you need to call `wrapper.update` for the component to actually re-render?

dzearing: nope. locally updated to count renders, and it passes.

```
    let lastMergedRef;
    let renderCount = 0;
    const TestComponent: React.FunctionComponent = ({ val }) => {
      renderCount++;
      lastMergedRef = useMergedRefs<boolean>(() => ({}));
      return null;
    };

    wrapper = mount(<TestComponent />);
    const ref1 = lastMergedRef;
    expect(renderCount).toEqual(1);

    wrapper.setProps({});
    expect(renderCount).toEqual(2);
    const ref2 = lastMergedRef;
```

dzearing: In fact, changing that to `update` will update the wrapper state to match the rendered state - not cause a rerender. The test above fails if you change `setProps` to `update`.

dzearing: https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/update.html

packages/react-hooks/src/useMergedRefs.ts
dzearing: fixed.

dzearing:
Build has passed, so i'm moving this one in.

packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,19 +1,24 @@
...
-export function useMergedRefs<T>(...refs: React.Ref<T>[]): (instance: T) => void {
-  return React.useCallback((value: T) => {
-    refs.forEach(ref => {
+export function useMergedRefs<T>(...refs: Ref<T>[]): (instance: T) => void {
+  const state = useRef<(Ref<T> | undefined)[]>();
+
+  // Update refs list.
+  state.current = refs;
+
+  return useCallback((value: T) => {
```
ecraig12345, line 14: Coming in late here but this is exactly what `useConstCallback` was made for. It guarantees the callback never mutates. (The current implementation of `React.useCallback` also does that, but the React docs are clear that this may change in the future.)

PR #13713, opened 2020-06-19
Jest test: add and use act wrapped create

xugao:
#### Pull request checklist

#### Description of changes
Some context:
- I realize this change is needed in PR #13688
- I found this PR https://github.com/facebook/react/pull/16013 in `react-test-renderer` repo

This change contains:
1. Adding (react-test-renderer) `create` with `act` wrapped in `reactTestRenderer.ts`: see how to use `act` with `create` [here](https://reactjs.org/docs/test-renderer.html#testrendereract)
2. Use new `create` in `ComponentExamples.test` (has expected snapshots update) and `react-next` where components are function components

#### Focus areas to test

(optional)

### DIFF
--- a/packages/utilities/src/test/reactTestRenderer.ts
+++ b/packages/utilities/src/test/reactTestRenderer.ts
@@ -1,0 +1,14 @@
+import { act, TestRendererOptions, ReactTestRenderer, create as defaultCreate } from 'react-test-renderer';
+import { ReactElement } from 'react';
+
+/**
+ * Wrapping `create` from `react-test-renderer' with `act`.
+ */
+export function create(nextElement: ReactElement, options?: TestRendererOptions): ReactTestRenderer {
+  let component: ReactTestRenderer;
+  act(() => {
+    component = defaultCreate(nextElement, options);
+  });
+
+  return component!;
+}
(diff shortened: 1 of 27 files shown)


packages/utilities/src/test/reactTestRenderer.ts
```diff
@@ -0,0 +1,14 @@
+import { act, TestRendererOptions, ReactTestRenderer, create as defaultCreate } from 'react-test-renderer';
+import { ReactElement } from 'react';
+
+/**
+ * Wrapping `create` from `react-test-renderer' with `act`.
+ */
+export function create(nextElement: ReactElement, options?: TestRendererOptions): ReactTestRenderer {
+  let component: ReactTestRenderer;
+  act(() => {
```
dzearing, line 9: hmm. is wrapping the whole thing in `act` the right thing? From what I've seen you want to wrap event simulates only with `act`, then assert outside of the act scope.

packages/utilities/src/test/reactTestRenderer.ts
```diff
@@ -0,0 +1,14 @@
+import { act, TestRendererOptions, ReactTestRenderer, create as defaultCreate } from 'react-test-renderer';
+import { ReactElement } from 'react';
+
+/**
+ * Wrapping `create` from `react-test-renderer' with `act`.
+ */
+export function create(nextElement: ReactElement, options?: TestRendererOptions): ReactTestRenderer {
```
MLoughry, line 7: Nit: you can make this more "future-proof" by using the typings of `defaultCreate`

```ts
export function create(...args: Parameters<typeof defaultCreate>): ReturnType<typeof defaultCreate>{
  let component: ReturnType<typeof defaultCreate>;
  act(() => {
    component = defaultCreate(...args);
  });

  return component!;
}

xugao: great suggestion, fixed!

PR #13714, opened 2020-06-19
React conformance - Adding new tests and updating slider test

sopranopillow:
#### Pull request checklist

#### Description of changes

This PR adds new test to the `react-conformance` default tests and updates the current tests for slider. `react-northstar`'s slider needs to be updated to disable `'has-top-level-file'` since components do not have a file following `package/src/component`. The rules added are the following:

* 'exported-top-level': Is exported at top level
* 'has-top-level-file': Has a corresponding top-level file
* 'is-static-property': For sub-component, is a static property of its parent component. This ensures CardBody is exposed as Card.Body.
* 'kebab-aria-attributes': Aria attributes should be kebab-cased and not camelCased
* 'consistent-callback-names': Ensure custom callback names are consistent (e.g. on[Part][Event] == onItemClick)


packages/react-conformance/src/defaultTests.tsx
```diff
@@ -73,4 +75,89 @@ export const defaultTests: TestObject = {
...
+        ? customMount(<Component {...requiredProps} />)
+        : mount(<Component {...requiredProps} />);
+      const props = mountedComponent.props();
+
+      if (props) {
+        for (const prop of Object.keys(props)) {
+          if (prop.toLowerCase().includes('aria')) {
+            expect(prop.split(/[A-Z][a-z]+/)?.length).toBeLessThanOrEqual(1);
+            expect(prop.split('-').length).toBeGreaterThan(1);
+          }
```
ecraig12345, line 137: This check can be a bit more specific: `aria` must be at the very beginning, followed by a hyphen then lowercase letters only.
```ts
          if (propName.startsWith('aria')) {
            expect(propName).toMatch(/^aria-[a-z]+$/);
          }
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -73,4 +75,89 @@ export const defaultTests: TestObject = {
...
+        }
+      }
+    });
+  },
+
+  /** Ensures that components have consistent custom callback names i.e. on[Part][Event] */
+  'consistent-callback-names': (componentInfo: ComponentDoc, testInfo: TestingOptions) => {
+    it(`has consistent custom callback names`, () => {
+      const { Component, customMount, requiredProps } = testInfo;
+      const mountedComponent = customMount
```
ecraig12345, line 147: Same as above, this needs to be tested on the declared props (`componentInfo.props`) not the rendered ones.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -73,4 +75,89 @@ export const defaultTests: TestObject = {
...
+        ? customMount(<Component {...requiredProps} />)
+        : mount(<Component {...requiredProps} />);
+      const props = mountedComponent.props();
+
+      if (props) {
+        for (const prop of Object.keys(props)) {
+          if (prop.slice(0, 3) === 'on') {
```
ecraig12345, line 154: This is probably not quite what we want to test. Instead I'd recommend checking `if (/^on[A-Z]/.test(propName))` and then:
1. Remove the "on" prefix
2. Get the last word. It should *not* end with "ed". (We might also want to test it against a valid list of verbs, but not sure how to get that list.)
3. The remaining words (with first letter converted back to lowercase) should probably match a prop name. 

Due to existing violations of these rules, at this point we'll probably need to introduce test-specific options: a mapping from test name to options object. Initially it would be like:
```
interface ???? {
  'consistent-callback-names'?: {
    ignoreProps?: string[];
  };
}
```
The ??? is because I think the best name for this interface would be `TestOptions`, but that's already taken right now by the interface for conformance test options all-up. So we'd need to either come up with a different name, or rename the all-up options interface to something like `IsConformantOptions` (which I think would make more sense anyway).

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +67,95 @@ export const defaultTests: TestObject = {
...
+        const SubComponent = useDefaultExport ? subComponentFile.default : subComponentFile[subComponentName];
+
+        expect(ReactIs.isValidElementType(SubComponent)).toBe(true);
+      }
+    });
+  },
+
+  /** Ensures aria attributes are kebab cased */
+  'kebab-aria-attributes': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    it(`Aria attributes are kebab-cased`, () => {
```
ecraig12345, line 123:
```suggestion
    it(`uses kebab-case for aria attributes`, () => {
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +67,95 @@ export const defaultTests: TestObject = {
...
+      const props = Object.keys(componentInfo.props);
+
+      for (const prop of Object.keys(props)) {
```
ecraig12345, line 126: I'm pretty sure this isn't working as expected--`props` is an array, so `Object.keys(props)` will return the indices. Instead you can just do `for (const prop of props)`.

In general, for each test, it's a good idea to temporarily make a component violate the criteria to ensure that the test actually catches violations.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +67,95 @@ export const defaultTests: TestObject = {
...
+    });
+  },
+
+  // TODO: Test last word of callback name against list of valid verbs
+  /** Ensures that components have consistent custom callback names i.e. on[Part][Event] */
+  'consistent-callback-names': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    it(`has consistent custom callback names`, () => {
+      const propKeys = Object.keys(componentInfo.props);
+      const { testingOptions } = testInfo;
+      const ignoreProps = testingOptions ? testingOptions['consistent-callback-names']?.ignoreProps : undefined;
```
ecraig12345, line 140: If you do this you can get rid of the `?` when accessing ignoreProps later.
```suggestion
      const ignoreProps = testingOptions?.['consistent-callback-names']?.ignoreProps  || [];
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +67,95 @@ export const defaultTests: TestObject = {
...
+        }
+      }
+    });
+  },
+
+  // TODO: Test last word of callback name against list of valid verbs
+  /** Ensures that components have consistent custom callback names i.e. on[Part][Event] */
+  'consistent-callback-names': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    it(`has consistent custom callback names`, () => {
+      const propKeys = Object.keys(componentInfo.props);
```
ecraig12345, line 138: Might be better to call this propNames for clarity

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +67,95 @@ export const defaultTests: TestObject = {
...
       expect(displayName).toEqual(fileName);
     });
   },
+
+  /** Ensures component is exported at top level allowing import { Component } from 'packageName' */
+  'exported-top-level': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    if (!testInfo.isInternal) {
+      it(`is exported at top-level`, () => {
+        const { displayName, componentPath, Component } = testInfo;
+        const rootPath = componentPath.replace(/src.*/, '');
```
ecraig12345, line 82: To be safe you should make this a little more specific (same in other places):
```suggestion
        const rootPath = componentPath.replace(/[\\/]src[\\/].*/, '');
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +65,95 @@ export const defaultTests: TestObject = {
...
+
+        expect(topLevelFile.default || topLevelFile[displayName]).toBe(Component);
+      });
+    }
+  },
+
+  /** If the component is a subcomponent, ensure its parent has the subcomponent as static property */
+  'is-static-property-of-parent': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    it(`is a static property of its parent`, () => {
+      const { componentPath, displayName, useDefaultExport } = testInfo;
```
ecraig12345, line 106: This is one case where it might be better to do part of the setup (the `isParent` calculation) before declaring the `it()`--this test ideally shouldn't run if the component is a parent.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +65,95 @@ export const defaultTests: TestObject = {
...
+      const ignoreProps = testingOptions ? testingOptions['consistent-callback-names']?.ignoreProps : undefined;
+
+      for (const propName of Object.keys(propKeys)) {
+        if (!ignoreProps?.includes(propName) && /^on[A-Z]/.test(propName)) {
+          const words = propName.slice(2).match(/[A-Z][a-z]+/g);
+
+          if (words) {
+            // Make sure last word doesn't end with ed
+            const lastWord = words[words.length - 1];
+            expect(lastWord.lastIndexOf('ed')).not.toBe(lastWord.length - 2);
```
ecraig12345, line 149: You can simplify this since `endsWith` is available in Node:
```suggestion
            expect(lastWord.endsWith('ed')).toBe(false);
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +65,95 @@ export const defaultTests: TestObject = {
...
+  // TODO: Test last word of callback name against list of valid verbs
+  /** Ensures that components have consistent custom callback names i.e. on[Part][Event] */
+  'consistent-callback-names': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    it(`has consistent custom callback names`, () => {
+      const propKeys = Object.keys(componentInfo.props);
+      const { testingOptions } = testInfo;
+      const ignoreProps = testingOptions ? testingOptions['consistent-callback-names']?.ignoreProps : undefined;
+
+      for (const propName of Object.keys(propKeys)) {
+        if (!ignoreProps?.includes(propName) && /^on[A-Z]/.test(propName)) {
```
ecraig12345, line 143: Just realized we need to exclude render callbacks. (`(?!whatever)` is negative lookahead)
```suggestion
        if (!ignoreProps?.includes(propName) && /^on(?!Render[A-Z])[A-Z]/.test(propName)) {
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +65,95 @@ export const defaultTests: TestObject = {
...
+          const words = propName.slice(2).match(/[A-Z][a-z]+/g);
+
+          if (words) {
+            // Make sure last word doesn't end with ed
+            const lastWord = words[words.length - 1];
+            expect(lastWord.lastIndexOf('ed')).not.toBe(lastWord.length - 2);
+
+            // Checking on[Part][Event] [Part] matches a prop
+            const firstWord = words[0].toLowerCase;
```
ecraig12345, line 152: Doing a search through existing props, we should probably get rid of this part of the test--there are actually a lot more (valid) `on[Part][Event]` cases where `[Part]` is *not* the name of a prop than cases where it is.

(search in VS Code for case-sensitive regex `\bon(?!Render)([A-Z][a-z]+){2,}`, limited to `*.types.ts`)

packages/react-conformance/src/types.ts
```diff
@@ -5,10 +5,16 @@ import { mount } from 'enzyme';
 
 export type Tests = keyof typeof defaultTests;
 
+export interface TestingOptions {
```
ecraig12345, line 8: One more nitpicky naming thing (sorry), this should probably just be `TestOptions`. And add a doc comment noting that it's options for individual tests.

packages/react-conformance/src/types.ts
```diff
@@ -5,10 +5,16 @@ import { mount } from 'enzyme';
...
+  'consistent-callback-names'?: {
+    ignoreProps?: string[];
+  };
+}
+
 /**
  * {@docCategory isConformant}
```
ecraig12345, line 15: These interfaces don't need the docCategory tag (that's only for things that will be on the public doc site)

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -65,12 +65,95 @@ export const defaultTests: TestObject = {
...
+  /** If the component is a subcomponent, ensure its parent has the subcomponent as static property */
+  'is-static-property-of-parent': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    it(`is a static property of its parent`, () => {
+      const { componentPath, displayName, useDefaultExport } = testInfo;
+      const componentFolder = componentPath.replace(path.basename(componentPath) + path.extname(componentPath), '');
+      const dirName = path.basename(componentFolder).replace(path.extname(componentFolder), '');
+      const isParent = displayName === dirName;
+
+      if (!isParent) {
+        const subComponentName = displayName.replace(dirName, '');
```
ecraig12345, line 112: I'm not sure this is doing the right thing--what we need to do here is determine the path to the *parent* component, require that, and then `expect(ParentComponent[displayName]).toBe(Component)`

packages/react-conformance/src/defaultTests.tsx
sopranopillow: Forgot that lookageads are allowed since it doesn't matter in node 😅

ecraig12345:
Looks like you either need to add some excludes or entirely disable the aria test for slider. (This will be the case with a lot of v7 components.)
https://dev.azure.com/uifabric/fabricpublic/_build/results?buildId=85282&view=logs&j=92058b04-f16a-5035-546c-cae3ad5e2f5f&t=17e772b0-f0c3-53ab-7aac-ad6df80e30d7&l=652

sopranopillow:
> Looks like you either need to add some excludes or entirely disable the aria test for slider. (This will be the case with a lot of v7 components.)
> https://dev.azure.com/uifabric/fabricpublic/_build/results?buildId=85282&view=logs&j=92058b04-f16a-5035-546c-cae3ad5e2f5f&t=17e772b0-f0c3-53ab-7aac-ad6df80e30d7&l=652

I noticed 😕yesterday, oufr and react-next do not follow kebab cased for aria. I will disable them for now.

PR #13715, opened 2020-06-19
Button: Replacing content with children

### LINKED ISSUE #13665: @fluentui/react-button: does not render `children`
### Environment Information

```tsx
<Button>I am a button</Button>
```

https://codesandbox.io/s/mystifying-morse-mxf67?file=/src/App.js

#### Actual behavior:

Renders nothing.

#### Expected behavior:

A button with "I am a button" text.

Requested priority: Normal

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13665

#### Description of changes

This PR addresses an issue where `children` passed into the `Button` in `@fluentui/react-button` were not being rendered. The PR fixes this by rendering the `children` always and dropping the `content` slot to follow html expectations.

The PR also adds a test so that this issue doesn't regress in the future.

#### Focus areas to test

(optional)


khmakoto:
Closing and reopening to trigger size auditor job again.

packages/react-button/src/components/Button/ButtonBase.tsx
```diff
@@ -10,10 +10,16 @@ export const ButtonBase = compose<'button', ButtonProps, ButtonProps, {}, {}>(
 
     return (
       <slots.root ref={ref} {...slotProps.root}>
-        {props.loading && <slots.loader {...slotProps.loader} />}
-        {props.icon && props.iconPosition !== 'after' && <slots.icon {...slotProps.icon} />}
-        {!props.iconOnly && props.content && <slots.content {...slotProps.content} />}
-        {props.icon && props.iconPosition === 'after' && <slots.icon {...slotProps.icon} />}
+        {state.children ? (
+          state.children
```
layershifter, line 14: This becomes extremely confusing 🤔 

- What I should use `state.children`, `props.children` or `slotProps.children`?
- What I should use `state.icon`, `props.icon` or `slotProps.icon`?

khmakoto: I think this makes sense. I'm going to repurpose this PR to get rid of the `content` slot since that itself is confusing with `children` and we shouldn't reinvent things here and just go with what `HTML` is already doing. I'll think about how this should look.

packages/react-button/src/components/Button/ButtonBase.tsx
```diff
@@ -6,22 +6,22 @@ import { useButton } from './useButton';
...
+        {!iconOnly && children && <span>{children}</span>}
+        {icon && iconPosition === 'after' && <slots.icon {...slotProps.icon} />}
       </slots.root>
     );
   },
   {
     displayName: 'ButtonBase',
     handledProps: [
       'circular',
-      'content',
```
mnajdova, line 24: I don't understand why we are `removing` the content, just because we want to support `children`. In my opinion, if there are children, we should just render children (that way we are allowing the users to opt out of our rendering logic). If they want to use shorthand, they can use the `content` in combination with other slots.

khmakoto: The reasoning we had was that we shouldn't go against the standard expectations. When talking about an `HTML` button we expect the content to be passed as `children`. Having both a `content` slot and `children` rendering support seems unintuitive and would most probably lead to confusion, especially for first-time users of our components.

An example: If we just let them opt out of our rendering logic what's the expectation? Should we still render the `icons` or not? Where's the position of the `children` if the answer to the previous question is yes? It all can become very confusing very fast.

dzearing: I think replacing `content` with only `children` is the right call - it simplifies the API and is more predictable.

If you try other component library buttons, or just vanilla, Rendering a button with some text is always this:

```tsx
<Button>Hello</Button>
```

Take a look at other frameworks as well; this seems to be a consistent pattern. 

A user will then add an icon:

```tsx
<Button>
  <AddIcon />
  Hello
</Button>
```

This creates a wrong looking button. The icon size isn't contained, may not be the right color, doesn't have the right gap. So they realize we have a slot for that:

```tsx
<Button icon={ <AddIcon/> }>Hello</Button>
```

This feels ergonomic and matches other libraries. So, I believe having both content and children definitely creates redundancy, but just having children seems like a pattern more consistent with native.

There are cases like `Checkbox`, where `children` makes less sense. I would expect cases like this prohibiting `children` because it tends to follow the native `input` element expectations.

`Link` is another analogous example which wouldn't make sense to prioritize `content` over `children`. I think `Button` and `Link` should feel pretty symmetrical.

packages/react-button/src/components/Button/ButtonBase.tsx
dzearing: So, `state` should === superset of `props`, if any. That's it. The point is totally spot on though, we need to simplify and possibly remove slotProps in favor of state. It's too confusing. I don't think slotProps is necessary when state[slot] is good enough.

The only answer should be `state.children`. By default, with no state hooks, state === props.

khmakoto:
Closing this because changes have been made that mostly address this already.

PR #13717, opened 2020-06-19
Button: Adding correct accessibility attributes when Button is rendered as something else

### LINKED ISSUE #13671: @fluentui/react-button: missing accessibility attributes when `as` prop is used
### Environment Information

```tsx
<Button
  as="div"
  content={{
    children: "Foo"
  }}
/>
```

https://codesandbox.io/s/funny-moon-xde2j?file=/src/App.js

#### Actual behavior:

`role="button"`, keyboard handler for enter/space, and `tabIndex="0"` are missing.

#### Expected behavior:

`role="button"` and `tabIndex="0"` are pressed when an element is `div`.

Requested priority: Normal

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13671

#### Description of changes

This PR adds the correct attributes that the component needs to behave as `Button` when being rendered as something else, such as a `div`. These attributes include `role="button"`, `tabIndex=0` and correct handling of `Spacebar`/`Enter` keys.

A test is also added so this does not regress in the future.

#### Focus areas to test

(optional)


packages/react-button/src/components/Button/useButton.ts
```diff
@@ -6,8 +7,33 @@ import { ButtonProps, ButtonState } from './Button.types';
  * @param props - Button props to derive state from.
  */
 export const useButton = (props: ButtonProps): ButtonState => {
```
ecraig12345, line 9: How does all this work with `as="a"`?

packages/react-button/src/components/Button/useButton.ts
```diff
@@ -6,8 +7,33 @@ import { ButtonProps, ButtonState } from './Button.types';
...
+        onClick(ev);
+      }
+    }
+  };
+
+  const isButtonRootType = !as || as === 'button';
+  const rootTypeProps = isButtonRootType
+    ? {}
+    : {
+        onKeyDown,
```
jurokapsiar, line 29: would it be possible to use v0 behaviors in some form? another use case - if this is composed into a toggle button, aria-pressed needs to be applied. this logic can get hairy and it would be good to have it externalized/centralized so that the components are not polluted with it

packages/react-button/src/components/Button/useButton.ts
khmakoto: This is accounted for now.

packages/react-button/src/components/Button/useButton.ts
khmakoto: @jurokapsiar I'm making use of it now, please take a look again.

packages/react-button/src/components/Button/useButtonBehavior.ts
```diff
@@ -0,0 +1,31 @@
...
+export const useButtonBehavior = (props: ButtonProps): Partial<ButtonState> => {
+  const { as, disabled, loading, onClick, onKeyDown } = props;
+
+  const { attributes, keyActions } = buttonBehavior({ as: as as string, disabled, loading });
+
+  let _onKeyDown = onKeyDown;
+
+  const keyCombinations = keyActions?.root?.performClick?.keyCombinations;
+  if (keyCombinations) {
+    _onKeyDown = (ev: React.KeyboardEvent<HTMLButtonElement> & React.MouseEvent<HTMLButtonElement>) => {
```
layershifter, line 14:
```suggestion
    _onKeyDown = (ev: React.KeyboardEvent<HTMLButtonElement>) => {
```

I don't think that it's valid. I suggest to update `ButtonProps` instead to include `KeyboardEvent` as a possible option to `onClick` handler

packages/react-button/src/components/Button/useButton.ts
```diff
@@ -14,14 +15,16 @@ export const useButton = (
...
   useImperativeHandle(props.componentRef, () => ({
     focus: () => buttonRef.current?.focus(),
   }));
   useFocusRects(ref as React.RefObject<HTMLElement>);
 
+  const buttonBehaviorProps = useButtonBehavior(props);
+
   return {
     ...props,
+    ...buttonBehaviorProps,
```
layershifter, line 27:
```suggestion
    ...buttonBehaviorProps,
    ...props,
```

User's input should always win, I suggest to have a dedicated test for this

packages/react-button/src/components/Button/useButtonBehavior.ts
```diff
@@ -0,0 +1,31 @@
+import { buttonBehavior } from '@fluentui/accessibility';
+import { getCode } from '@fluentui/keyboard-key';
+import { ButtonProps, ButtonState } from './Button.types';
+
+export const useButtonBehavior = (props: ButtonProps): Partial<ButtonState> => {
+  const { as, disabled, loading, onClick, onKeyDown } = props;
+
+  const { attributes, keyActions } = buttonBehavior({ as: as as string, disabled, loading });
```
layershifter, line 8: V0 code has the same issue. `as` usage is unsafe to use there as it will not work for cases where a component will be passed. In this case it's preferable to handle it via `ref` so will always know an actual `tagName`:

https://github.com/reakit/reakit/blob/9bd6882fb50948c616e757d750b436decc36ddfc/packages/reakit/src/Button/Button.ts#L40

packages/react-button/src/components/Button/useButtonBehavior.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (keyCombinations) {
+    _onKeyDown = (ev: React.KeyboardEvent<HTMLButtonElement> & React.MouseEvent<HTMLButtonElement>) => {
+      if (onKeyDown) {
+        onKeyDown(ev);
+      }
+
+      if (onClick) {
+        const eventCode = getCode(ev);
+        for (const keyCombination of keyCombinations) {
+          if (eventCode === keyCombination.keyCode) {
```
xugao, line 22: i feel this logic to check key match will be needed in many places in the future. can we extract? or maybe we already have a helper

packages/react-button/src/components/Button/useButtonBehavior.ts
khmakoto: Made the change.

packages/react-button/src/components/Button/useButton.ts
khmakoto: Made the change and added some tests.

packages/react-button/src/components/Button/useButtonBehavior.ts
khmakoto: Thanks for this suggestion! I've made the change.

packages/react-button/src/components/Button/useButtonBehavior.ts
khmakoto: I extracted it into its own hook in `@uifabric/react-hooks`.

packages/react-button/src/components/Button/useButton.ts
```diff
@@ -14,14 +15,18 @@ export const useButton = (
   options: ComposePreparedOptions,
 ): ButtonState => {
   const buttonRef = useRef<HTMLButtonElement | null>(null);
-
   useImperativeHandle(props.componentRef, () => ({
     focus: () => buttonRef.current?.focus(),
   }));
   useFocusRects(ref as React.RefObject<HTMLElement>);
 
+  const buttonBehaviorProps = useButtonBehavior(props, ref);
```
layershifter, line 23:
```suggestion
  const buttonBehaviorProps = useButtonBehavior(props, buttonRef);
```

I expect there `buttonRef` as it will always exist

packages/react-button/src/components/Button/useButtonBehavior.ts
```diff
@@ -0,0 +1,16 @@
...
+import { ButtonProps, ButtonState } from './Button.types';
+
+export const useButtonBehavior = (props: ButtonProps, ref: React.Ref<HTMLElement>): Partial<ButtonState> => {
+  const { as, disabled, loading, onClick, onKeyDown: _onKeyDown } = props;
+
+  const { attributes, keyActions } = buttonBehavior({
+    as: (ref !== null && typeof ref !== 'function' && ref.current?.tagName) || (as as string),
+    disabled,
+    loading,
+  });
```
layershifter, line 12: This will not work 😭 

If you will check the original suggestion it works based on `elementType` that is stored under the state (https://github.com/reakit/reakit/blob/9bd6882fb50948c616e757d750b436decc36ddfc/packages/reakit/src/Button/Button.ts#L40). I made a simplified CSB to highlight it: https://codesandbox.io/s/thirsty-galois-tdz7s?file=/src/App.js

`ref.current?.tagName` will be nothing during the first render as refs are resolved during mount stage.

---

```suggestion
  const [elementTag, setElementTag] = React.useState("button");

  const { attributes, keyActions } = buttonBehavior({
    as: tagName,
    disabled,
    loading,
  });
  
    React.useEffect(() => {
      const element = ref.current;

      if (!element) {
        console.warn(
          "Can't determine whether the element is a native button because `ref` wasn't passed to the component",
        );
        return;
      }

      setElementTag(element.tagName.toLowerCase())
    }, []);
```

packages/react-hooks/src/useBehaviorKeyActions.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const useBehaviorKeyActions = <TElement>(
+  keyActions?: KeyActions,
+  eventHandlers?: React.DOMAttributes<TElement>,
+): React.DOMAttributes<TElement> => {
+  const { onClick, onKeyDown } = eventHandlers || {};
+
+  let _onKeyDown = onKeyDown;
+
+  const keyCombinations = keyActions?.root?.performClick?.keyCombinations;
```
layershifter, line 12: Is it expected to have there hardcoded `performClick`?

khmakoto: From what I saw that seems to be the standard in v0, although it doesn't just apply to root but also other slots. I'll account for this in the hook but let me know if any of this is a wrong assumption to make.

Also, I'm just taking care of this scenario for now but I'd expect this hook to grow as more is needed from the v0 behaviors.

packages/react-button/src/components/Button/useButton.ts
khmakoto: Changed!

packages/react-button/src/components/Button/useButtonBehavior.ts
khmakoto: Changed!

packages/react-button/src/components/Button/Button.test.tsx
```diff
@@ -1,14 +1,93 @@
...
+  it('renders Button with the correct default accessibility props', () => {
+    const onClick = jest.fn();
+
+    wrapper = mount(<Button onClick={onClick}>Button</Button>);
+
+    const button = wrapper.find('button');
+
+    expect(button.getElements().length).toBe(1);
+    expect(button.props().role).toBe(undefined);
+    expect(button.props().tabIndex).toBe(undefined);
```
layershifter, line 30:
```suggestion
    expect(button.prop('role')).toBe(undefined);
    expect(button.prop('tabIndex')).toBe(undefined);
```

_nit:_ Enzyme has support of `.prop()`

packages/react-button/src/components/Button/useButton.ts
```diff
@@ -14,14 +15,18 @@ export const useButton = (
...
   }));
   useFocusRects(ref as React.RefObject<HTMLElement>);
 
+  const buttonBehaviorProps = useButtonBehavior(props, buttonRef);
+  const { onKeyDown } = buttonBehaviorProps;
+
   return {
+    ...buttonBehaviorProps,
     ...props,
+    onKeyDown,
```
layershifter, line 29: This looks fragile as will require the same action across all components...

`useAccessibility()` has the same issue but handles it in a single place:
https://github.com/microsoft/fluentui/blob/346cd61df2ce2268f4126eab1b31a173904ff4c2/packages/fluentui/react-bindings/src/hooks/useAccessibility.ts#L68-L73

khmakoto: You're right! I'm actually already doing this in place and just didn't tackle it correctly. Fixing it.

packages/react-button/src/components/Button/Button.test.tsx
khmakoto: Went with your suggestion.

khmakoto:
Closing for now until Button is more stable since changes have been causing lots of conflicts right now.

PR #13718, opened 2020-06-19
Button: Adding various tests and updating component to include componentRef

dzearing:
Setting up componentRef focus method, adding tests.


change/@uifabric-react-hooks-2020-06-19-12-26-25-fix-button-fixes.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
dzearing, line 2: remove after merge

packages/react-button/src/components/Button/Button.scss
```diff
@@ -273,6 +273,7 @@
   right: -1px;
   bottom: -1px;
   content: '';
+  pointer-events: none;
```
dzearing, line 276: confused why this is showing up

packages/react-button/src/components/Button/ButtonBase.tsx
```diff
@@ -31,8 +34,10 @@ export const ButtonBase = compose<'button', ButtonProps, ButtonProps, {}, {}>(
       'loading',
       'primary',
       'secondary',
+      'tokens',
       'size',
-    ],
+      // tslint:disable-next-line:no-any
+    ] as any,
```
dzearing, line 40: This needs to be discussed. `handledProps` is a mess with hooks.

I needed to add `buttonRef` to the state. `handledProps` only takes in `TProps`. The `compose` typings don't even take in `TState`, but even if it did, the hook is what knows about the state being returned and should own the state. So I'm perplexed here on what the right thing should be.

I am very much thinking that getNativeProps might need to be used, and perhaps explicitly called in each component.

change/@fluentui-react-button-2020-06-19-15-32-16-fix-button-fixes.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Button: Adding focus method and tests.",
```
khmakoto, line 3:
```suggestion
  "comment": "Button: Adding componentRef, focus method and tests.",
```

packages/react-button/src/components/Button/Button.stories.tsx
```diff
@@ -2,64 +2,74 @@ import * as React from 'react';
...
+/**
+ * Temporary Stack until there's one in its own package.
+ */
+const Stack = (props: React.PropsWithChildren<{ horizontal?: boolean }>) => {
+  const { horizontal, ...rest } = props;
+
+  return <div {...rest} className={horizontal ? classes.hStack : classes.vStack} />;
+};
+
+const Text = (props: React.PropsWithChildren<{}>) => <h2 {...props} className={classes.text} />;
```
khmakoto, line 15: nit: You should either indicate in the comment above `Temporary Stack and Text components ...` or add a similar comment here for `Text`.

packages/react-button/src/components/Button/Button.stories.tsx
```diff
@@ -2,64 +2,74 @@ import * as React from 'react';
...
 
 const ButtonVariants = (props: ButtonProps) => (
-  <div className={classes.hStack}>
-    <Button {...props} content="Hello, world" icon="X" />
-    <Button {...props} primary content="Hello, world" icon="X" />
-    <Button {...props} disabled content="Hello, world" icon="X" />
-    <Button {...props} primary disabled content="Hello, world" icon="X" />
-  </div>
+  <Stack horizontal>
+    <Button content="Hello, world" icon="O" {...props} />
```
khmakoto, line 19: Are you changing the icon here just to have at least one different?

packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -5,7 +5,22 @@ import { ColorPlateSet } from '@fluentui/react-theme-provider';
...
+export interface ButtonRef {
+  /**
+   * Sets focus to the button.
+   */
+  focus: () => void;
+}
+
 export interface ButtonProps extends ComponentProps, React.HTMLAttributes<HTMLButtonElement> {
+  /**
+   * Access the imperative api of the button.
```
khmakoto, line 20: nit:
```suggestion
   * Access the imperative API of the Button.
```

packages/react-button/src/components/Button/ButtonBase.tsx
```diff
@@ -2,14 +2,15 @@ import * as React from 'react';
...
+import { useMergedRefs } from '@uifabric/react-hooks';
 
 export const ButtonBase = compose<'button', ButtonProps, ButtonProps, {}, {}>(
   (props, ref, options) => {
     const { state } = options;
     const { slots, slotProps } = mergeProps(state, options);
 
     return (
-      <slots.root ref={ref} {...slotProps.root}>
+      <slots.root ref={useMergedRefs(ref, state.buttonRef)} {...slotProps.root}>
```
khmakoto, line 13: Shouldn't this merging of the refs be done in `useButton` instead?

packages/react-button/src/components/Button/Button.stories.tsx
dzearing: I'm moving the props mixing to the end so that I can override things later.

packages/react-button/src/components/Button/ButtonBase.tsx
dzearing: let me think about this.

dzearing: Another q: should ref even be here, and should it have been in slotProps.root already

packages/react-button/src/components/Button/ButtonBase.tsx
dzearing: OK I've merged a getNativeProps performance improvement, so I'll revisit this in a separate PR.

packages/react-button/src/components/Button/ButtonBase.tsx
dzearing: I think this is probably correct for now as the ref is being set here already. Let's keep as is for now.

layershifter:
Hey folks!

Can you please clarify why you have decided to go this approach? As I don't see any benefits from it.

### Issue 1: Additional API

It adds an additional API about what customer should know and learn. In the same time there is an existing API with `ref`s and it works. It's also a standard thing even for CSS-in-JS libraries, https://styled-components.com/docs/advanced#refs

At the same time there is no additional value from it as points to the same element: https://github.com/microsoft/fluentui/blob/4746eb1c2dd4846510353f839bc4079511a75e22/packages/react-button/src/components/Button/ButtonBase.tsx#L13

### Issue 2: Endless story

As customer I would like to have access to other DOM methods like `.blur()`, `.select()` (https://github.com/Semantic-Org/Semantic-UI-React/pull/2928) and even `.querySelector()`. What I should use and why? Are we going to add all DOM methods or guide user to use `componentRef` & `ref`?

```tsx
function App() {
  const ref = React.useRef() // Should I use `.focus()` there?
  const componentRef = React.useRef() // Or there?

  return <Button componentRef={componentRef} ref={ref} />
}
```

### Issue 3: Internal usage

There are cases where we will need refs to call some DOM methods like `.focus()`. We can't use `.findDOMNode()` as it will throw a warning in Strict Mode and will make components incompatible with Concurrent Mode of React.

So what we should prefer internally `componentRef` or `ref`? If we will prefer a `componentRef` how we are going to detect that it is supported?

```tsx
const isComponentRefSupported = !!slots.button.type.IS_FLUENT_COMPONENT
// ??? But we know that this will not work because of HoCs at least
```

dzearing:
> At the same time there is no additional value from it as points to the same element:
>
> `<slots.root ref={useMergedRefs(ref, state.buttonRef)} {...slotProps.root}>`

I don't think you're following it quite right. `buttonRef` is not the imperative API handle. It is simply a ref, owned by state management, and used in the imperative api surface. `ref` is the optional ref, provided by the user and forwarded by the `forwardRef` call, possibly. We can not re-use the `ref` to gain access to the root element from the state hook as it's optional and we don't own it. So we make a `buttonRef` to enable us to identify it and use it.

The reasoning for `componentRef` pattern is this:

* `ref` should return the root element predictably. This gives you an escape hatch that predictably gains you access to the DOM, so that we can avoid `findDOMNode` due to deprecation like you called out.
* `componentRef` should return the abstracted, *limited* imperative API surface. Typically this is limited to `focus`, but there are a few other things which are awkward to do declaratively.
* devs should not reach into private DOM element structure to execute an imperative thing. Either it should be a declarative prop, a declarative callback, or if imperative API surface is the right solution, expose a limited api surface for that.

By having `componentRef` we can provide a predictable way to do things like set focus. That way things like `SplitButton` or `FocusZone`, which have special behavior to determine the right focus target, still show up to customers in a consistent, predictable format: provide a ref to `componentRef`, which is the imperative api handle.

> As customer I would like to have access to other DOM methods like .blur(), .select() (Semantic-Org/Semantic-UI-React#2928) and even .querySelector(). What I should use and why? Are we going to add all DOM methods or guide user to use componentRef & ref?

No, we should not expose every possible API as an abstraction on the DOM. We should prioritize declarative ways to expose functionality. If that doesn't make sense (e.g. an Input's `setSelectedRange` method may be a good candidate for abstracting), we add it to the imperative handle. 

We've found limiting and abstracting common imperative APIs is a good thing:

* It creates a consistent, predictable API for accessing common imperative things without reaching into the DOM guts.
* It protects consumers from shifting DOM (either from code changes or from slot/as reconfiguration)
* It gives a window of opportunity to refine behavior (e.g. if `focus` is called and there are multiple options, we can refine where to go. Or if we want to asyncFocus to avoid layout events, we can do that.) 

> So what we should prefer internally componentRef or ref? If we will prefer a componentRef how we are going to detect that it is supported?

`ref` should return the root element, always. This creates symmetry with native element expectations. But using it is an escape hatch to the DOM. Once you're there, it's free for all, no guarantees.

`componentRef` should return the imperative limited api. Every component should decide if it is required; not everything has an imperative API. I don't think there is a need to detect that it is supported, other than from conformance tests.

And yes, now there are multiple ways to do the same thing, which in principle we want to avoid. You could set focus to a Button by calling `componentRef.current.focus()` which internally will abstract calling `focus` on the correct element, or you could go directly to the DOM like `ref.current.focus()` which breaks as soon as the DOM changes from your original expectations, but also triggers a layout event, and may not be applicable in all states. And, as you move to other components like `SplitButton` where the DOM changes, instead of focusing via `ref.current.firstChild.focus()` which is structurally sensitive and error prone, you predictably access the imperative API and call `focus()` and it will set focus to the right thing. Same with `FocusZone`, where setting focus must be abstracted because focus should move to the active element (a stateful value), rather than the first focusable child or the root.

`componentRef` provides a pattern for when imperative APIs are necessary. `focus` tends to be one that is common, necessary, and should be abstracted for predictability in API surface.

Hope this makes sense.  If you have alternatives on how to provide imperative apis when needed without requiring customers reaching into DOM structure and doing things we may want to change later, feel free to suggest alternatives.

layershifter:
> `ref` should return the root element, always. This creates symmetry with native element expectations. But using it is an escape hatch to the DOM. Once you're there, it's free for all, no guarantees.
> 
> `componentRef` should return the imperative limited api. Every component should decide if it is required; not everything has an imperative API. I don't think there is a need to detect that it is supported, other than from conformance tests.

What we are going to use internally? You mentioned `SplitButton` as an example, if `SplitButton` is used as a slot are we going to use `ref` or `componentRef` to execute `.focus()`?

layershifter:
What is a purpose to have this API for `Button`? Only to have unified approach for all components?

levithomason:
Let's have a sync on this one.  I think we moved a little too fast both in merging this and in objecting to it. Let's take a look at the use cases we have, refresh ourselves on the latest React docs dealing with [refs and the DOM](https://reactjs.org/docs/refs-and-the-dom.html) (especially [exposing DOM refs to parents](https://reactjs.org/docs/refs-and-the-dom.html#exposing-dom-refs-to-parent-components) and [forwardRef](https://reactjs.org/docs/forwarding-refs.html)), and hop on a call to discuss.  Who's in?

### Reminders

👉 We discussed this to a decent degree in our last off site as well, notes aren't too telling but here they are:

>componentRef currently supports 2 cases unpredictably - will continue to support both cases but with 2 difft APIs: have a way to always return imperative API but it shouldn’t be abused, separately have something that always gets DOM node.
>- return native DOM element in Stardust vs rn in Fabric, it’s returning the React instance or be line with hooks - imperative api?
>- current Ref in Fluent = findDOMNode
>
>-- https://hackmd.io/8WievoYRQpiAXbP_rBAecA?both

👉  We also have slots, so you can already get a ref to any part of the component, including the root:

```jsx
const MyComponent = () => {
  const rootRef = React.useRef();
  const iconRef = React.useRef();
  const headerRef = React.useRef();
  const descriptionRef = React.useRef();
  const actionRef = React.useRef();
  const progressRef = React.useRef();

  return (
    <Attachment
      ref={{ ref: rootRef }}
      icon={<MyIcon ref={iconRef} />}
      header={{ ref: headerRef }}
      description={{ ref: descriptionRef }}
      action={{ ref: actionRef }}
      progress={{ ref: progressRef }}
    />
  );
};
```

PR #13723, opened 2020-06-20
Button: Mixing style coming from slotProps

### LINKED ISSUE #13675: @fluentui/react-button: `style` values for slots in `compose()` are not passed to `root`
### Environment Information

```tsx
const Composed = compose(
  Button,
  {
    slotProps: () => ({
      root: {
        id: "button2",
        style: { background: "red" }
      }
    })
  }
);
```

```tsx
<Composed content={{ children: "Composed" }} />
```

https://codesandbox.io/s/divine-fast-uf8bi?file=/src/App.js:363-411

#### Actual behavior:

`style` values are passed.

#### Expected behavior:

`style` values are not passed.

Requested priority: Normal

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13675

#### Description of changes

Previously, if you passed the `style` prop to the `root` as part of `slotProps` in compose this wouldn't be reflected in the composed component because `Button` was not correctly mixing it. This PR fixes this while also making this into a separate function `getStyleFromPropsAndOptions` that serves as a utility that can be reused by more components and not only `Button`, eliminating duplication.

#### Focus areas to test

(optional)


packages/react-button/src/components/Button/useButton.ts
```diff
@@ -1,13 +1,18 @@
...
-export const useButton = (props: ButtonProps): ButtonState => {
+export const useButton = (
+  props: ButtonProps,
+  ref: React.Ref<HTMLElement>,
+  options: ComposePreparedOptions,
+): ButtonState => {
   return {
     ...props,
-    style: { ...props.style, ...tokensToStyleObject(props.tokens, '--button') },
+    style: getStyleFromPropsAndOptions(props, options),
```
xugao, line 16: shouldn't we do this in `resolveSlotProps`?

xugao, line 16: looks like you are dropping `--button` prefix here

packages/react-theme-provider/etc/react-theme-provider.api.md
```diff
@@ -39,6 +39,1523 @@ export type FontTokens = Partial<{
     fontWeight: string;
 }>;
 
+// @public (undocumented)
+export const getStyleFromPropsAndOptions: <TProps extends StyleProps<import("./types").ColorPlateSet>, TOptions extends StyleOptions<TProps>>(props: TProps, options: TOptions, prefix?: string | undefined) => {
```
xugao, line 43: pls fix return type if relevant

packages/react-button/src/components/Button/useButton.ts
khmakoto: We probably should, I wanted to take a stab and solve this in the `Button` and get some feedback before applying it at the compose layer.

xugao: probably better to keep them separated for now

PR #13725, opened 2020-06-20
getNativeProps: using set to improve perf.

dzearing:
Testing through ci how changing to Set will affect performance.

### DIFF
--- a/change/@uifabric-experiments-2020-06-20-06-00-46-fix-getNativeProps.json
+++ b/change/@uifabric-experiments-2020-06-20-06-00-46-fix-getNativeProps.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Updated to use Sets",
+  "packageName": "@uifabric/experiments",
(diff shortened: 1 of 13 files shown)


dzearing:
Do not merge; I want to assess perf here. I am not sure how the tsconfig downlevelIteration will affect things, but typescript doesn't like Set iteration without it. Might be only when we build ES5 or when used with some other setting.

ecraig12345:
Per [this page](https://kangax.github.io/compat-table/es6/#ie11) `new Set(arr)` **does not** work in IE 11:
image

For downlevel iteration, enabling this might be okay but is a little scary due to subtle behavior changes it could cause. [New TS docs](https://www.staging-typescript.org/tsconfig#downlevelIteration) have an explanation of what it does, and [this more extensive exploration](https://mariusschulz.com/blog/downlevel-iteration-for-es3-es5-in-typescript) is a bit older but may still be applicable.

dzearing:
Closing for now.

Issue #12953, opened 2020-04-30
@fluentui/react-northstar: importing a single component adds >1MB to bundle size


astegmaier:
### Environment Information

- **Package version(s)**: `"@fluentui/react-northstar": "0.48.0"`
- **Browser and OS versions**: All Browsers / Mac OS X 10.15.4

### Please provide a reproduction of the bug in a codepen:

Create a simple app with the following files:
***index.tsx***
```
import React from "react";
import ReactDOM from "react-dom";
import App from "./components/App";

ReactDOM.render(<App />, document.getElementById("root"));
```

***components/App.tsx***
```
import React from "react";
import { Provider, themes, Button } from '@fluentui/react-northstar';

const App = () => (
  <Provider theme={themes.teams}>
    <Button content="Hello from FluentUI" />
  </Provider>
);

export default App;
```

Build it with webpack (or Parcel2) in production mode (which enables tree-shaking and minification), and you'll see a 1.3-1.4MB increase in bundle size over a simple `Hello World` react app that doesn't use `@fluentui/react-northstar`.

Here's a visualization of the minified, tree-shaken bundle from `webpack`, using `source-map-explorer`:

![image](https://user-images.githubusercontent.com/1000369/80762146-f20b1300-8b09-11ea-96c1-3e9344f46d5e.png)

You can see a repro for the issue in the `fluentui-button` branch of [this repository](https://github.com/astegmaier/parcel-webpack-comparer/tree/fluentui-button). Run `yarn build:webpack` (to see webpack bundle output), `yarn build:parcel` (to see parcel2 bundle output).

This is related to issue [#2113](https://github.com/microsoft/fluent-ui-react/issues/2113), [#2324](https://github.com/microsoft/fluent-ui-react/issues/2324), and [#2324](https://github.com/microsoft/fluent-ui-react/issues/2325) on the legacy repo. Those are still marked as "open", and I don't see any issues or PRs addressing it in the new repo.

#### Actual behavior:

The bundle size increase of using a single button component is really, really big (1.3-1.4 MB). It looks like it's importing just about every component and icon into the bundle, even though the app uses no icons and only a single `button` component.

#### Expected behavior:

The bundle size increase should be "pay-to-play" - the more components/icons that I leverage, the more space it takes, and it should start small.

### Priorities and help requested:

Are you willing to submit a PR to fix? No

Requested priority: High (before a v1 release, would think it's important to get this under control).

mnajdova:
[...]

PR #13729, opened 2020-06-21
Resolve side effects in keyboard-key/codes.ts

connorjsmith:
#### Pull request checklist

- [x] Addresses an existing issue: Resolves part of #12953

#### Description of changes

Removes the side effect in codes.ts, which should help improve treeshaking for our packages


layershifter:
I think that 432 bytes is an acceptable to avoid side-effects.

PR #13739, opened 2020-06-22
docs: adding screen reader labeling examples

kolaps33:
Adding examples to the documentation with the most common usage.

Added text can be found here:
http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/13739/merge/react-northstar/accessibility#useful-examples-of-labelling

The table here:
http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/13739/merge/react-northstar/button-name-computation


packages/fluentui/docs/src/views/Accessibility.tsx
```diff
@@ -48,7 +48,6 @@ export default () => (
       <li>{link('High Contrast', '#high-contrast')}</li>
       <li>{link('Zoom', '#zoom')}</li>
     </ul>
-
```
mnajdova, line 51: Can we avoid all these line removals? Is this some formatting issue

kolaps33: I would like to avoid but not sure how? :)
when I switch off formatting these lines were removed when I did commit.

kolaps33: issue was that I didn't wrap text in <p> tag

packages/fluentui/docs/src/views/Accessibility.tsx
```diff
@@ -383,6 +383,37 @@ export default () => (
       `}
     />
 
+    <Header as="h3" content="Useful examples of labelling" />
+    <Header as="h4" content="Aria-labelledby refers to itself " />
+    <p>
+      In following example we want to get narrated from screen reader "delete Active members" button. Solution is
```
jurokapsiar, line 389: There are cases, when `aria-label` and `aria-labelledby` when you need to add names of other elements in the accessible name of an element. In the example below, the button has accessible name composed of its own name (delete) and header name (Active members). Solution is to refer to the element itself by its `id` in the `aria-labelledby` attribute. Screen reader
      computes the accessible name in the order which is provided in `aria-labelledby`. The strategy was taken from the
      {link(' Accessible Name and Description Computation 1.1', 'https://www.w3.org/TR/accname-1.1/#terminology')} page.

packages/fluentui/docs/src/views/Accessibility.tsx
```diff
@@ -383,6 +383,37 @@ export default () => (
...
+        aria-label="delete"
+        aria-labelledby="delete-button table-header"
+      />
+    </>
+    `}
+    />
+
+    <Header as="h3" content="Accessibility name computation" />
+    <p>
+      Recommendation how screen reader is computing accessibility name is mentioned in
```
jurokapsiar, line 410: The way how screen readers compute name and description is described in {link(' Accessible Name and Description Computation 1.1', 'https://www.w3.org/TR/accname-1.1/')} page. As the rules are quite complex, follow the page {' '}
      <Link to="button-name-computation"> Button name computation</Link> to see how these rules apply in concrete examples.

packages/fluentui/docs/src/views/Accessibility.tsx
kolaps33: applied changes

packages/fluentui/docs/src/views/Accessibility.tsx
kolaps33: applied changes

ecraig12345:
@kolaps33 Looks like this has a merge conflict.

PR #13741, opened 2020-06-22
fix(Divider): Fix rtl styles

### LINKED ISSUE #13337: Divider: No spaces between the lines and content in RTL context
### Environment Information

- **Package version(s)**: 0.49.0
- **Browser and OS versions**: Not relevant

#### Sandbox: 
https://codesandbox.io/s/divider-content-rtl-jzwql
#### Screenshot:
Divider with RTL content issue

#### Actual behavior:
No spaces between the lines and the content

#### Expected behavior:
There should be spaces between the lines and the content (as the same of LTR context)

Are you willing to submit a PR to fix? **Yes**

assuncaocharles:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13337

#### Description of changes

Fix RTL styles for `Divider` in RTL context

Screenshot 2020-06-23 at 07 44 59

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/themes/teams/components/Divider/dividerStyles.ts
```diff
@@ -40,11 +40,13 @@ const dividerStyles: ComponentSlotStylesPrepared<DividerStylesProps, DividerVari
             lineHeight: variables.textLineHeight,
             '::before': {
               ...beforeAndAfter(size, variables, colors, props),
-              marginRight: pxToRem(20),
+              ...(!rtl && { marginRight: pxToRem(20) }),
+              marginInlineEnd: pxToRem(20),
```
miroslavstastny, line 44: `margin-inline-end` is [not supported in IE11](https://caniuse.com/#search=margin-inline-end).

assuncaocharles: The alternative could be create a slot for the text and apply the margins on it, but @layershifter mentioned it can be too much to solve this...

Initially I didn't saw a problem in the related issue itself, the RTL text works in RTL context and LTR in the LTR... But if we are going to solve it what can be the best solution then?

assuncaocharles: I pushed new code solving it with a slot.

packages/fluentui/react-northstar/src/themes/teams/components/Divider/dividerStyles.ts
```diff
@@ -19,7 +19,7 @@ const beforeAndAfter = (
 });
 
 const dividerStyles: ComponentSlotStylesPrepared<DividerStylesProps, DividerVariables> = {
-  root: ({ props, variables }): ICSSInJSStyle => {
+  root: ({ props, variables, rtl }): ICSSInJSStyle => {
```
assuncaocharles, line 22:
```suggestion
  root: ({ props, variables }): ICSSInJSStyle => {
```

assuncaocharles:
> Please add screener example to catch regressions.

Screener added

packages/fluentui/react-northstar/src/themes/teams/components/Divider/dividerStyles.ts
abdullahsalem: You are right @assuncaocharles, initially it looks like there is no an issue (when the direction of the language meets the direction of the context flow), but in some cases there may be a need to fill that content with a language that has a different direction than the context flow direction. 
BTW, I've mentioned the main reason that leads to this issue, [here](https://github.com/microsoft/fluentui/issues/13337#issuecomment-634954807).

In other words:
> The `auto` value tells the browser to look at the first strongly typed character in the element and work out from that what the base direction of the element should be.
[w3.org](https://www.w3.org/International/questions/qa-html-dir#dirauto)

Therefore, I would suggest removing that `auto` if there is no any other side effect should be considered.

packages/fluentui/docs/src/examples/components/Divider/Visuals/DividerExampleRtlAndLtr.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { Divider, Provider, Card, CardHeader, CardBody, Text } from '@fluentui/react-northstar';
+
+const DividerExampleFittedShorthand = () => (
+  <>
+    <Provider>
```
layershifter, line 6:
```suggestion
```

packages/fluentui/docs/src/examples/components/Divider/Visuals/DividerExampleRtlAndLtr.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { Divider, Provider, Card, CardHeader, CardBody, Text } from '@fluentui/react-northstar';
+
+const DividerExampleFittedShorthand = () => (
```
layershifter, line 4:
```suggestion
const DividerExampleRtlAndLtr = () => (
```

packages/fluentui/docs/src/examples/components/Divider/Visuals/index.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+
+import ComponentExample from '../../../../components/ComponentDoc/ComponentExample';
+import NonPublicSection from '../../../../components/ComponentDoc/NonPublicSection';
+
+const UsageVisual = () => (
```
layershifter, line 6:
```suggestion
const Visual = () => (
```

packages/fluentui/react-northstar/src/components/Divider/Divider.tsx
```diff
@@ -91,11 +95,17 @@ const Divider: React.FC<WithAsProp<DividerProps>> & FluentComponentStaticProps<D
     rtl: context.rtl,
   });
 
+  const content = createShorthand(DividerContent, props.content, {
+    defaultProps: () => ({
+      styles: resolvedStyle.content,
+    }),
```
layershifter, line 101:
```suggestion
```

Let's move styles to `dividerContentStyles`

packages/fluentui/react-northstar/src/themes/teams/components/Divider/dividerStyles.ts
```diff
@@ -54,6 +52,10 @@ const dividerStyles: ComponentSlotStylesPrepared<DividerStylesProps, DividerVari
           }),
     };
   },
+  content: () => ({
```
layershifter, line 55: Please move this to `dividerContentStyles`

PR #13742, opened 2020-06-22
Remove KeytipData usage from Link, Checkbox, Toggle (react-next)

xugao:
#### Pull request checklist

#### Description of changes
Removed `KeytipData` usage from Link, Checkbox, Toggle components:
- Deprecated `keytipProps`
- A new way to support Keytip for a given component: use `useKeytipRef` hook and pass the ref to the targeted component

How to apply `Keytip` to a component after this change, here is an example:
```
import { Checkbox, useKeytipRef } from '@fluentui/react-next';

const ref = useKeytipRef({...});
<Checkbox ref={ref} />
```
You can also find usage examples in `Keytips.Basic.Example.tsx`

#### Focus areas to test

(optional)


packages/react-next/src/components/Keytip/examples/Keytips.Basic.Example.tsx
```diff
@@ -0,0 +1,61 @@
...
+        are shown.
+      </p>
+      <Pivot>
+        <PivotItem headerText="Pivot 1" keytipProps={keytipMap.Pivot1Keytip} style={pivotItemStyle}>
+          <Stack tokens={stackTokens}>
+            <SpinButton label="Spin Button" keytipProps={keytipMap.SpinButtonKeytip} styles={spinButtonStyles} />
+            <Toggle onText="Yes" offText="No" keytipData={toggleKeytipData} />
+            <span>
+              Go to{' '}
+              <Link keytipData={linkKeytipData} href="http://www.bing.com" target="_blank">
```
xugao, line 41: Before:
```
<Link keytipProps={keytipMap.LinkKeytip} href="http://www.bing.com" target="_blank">
```

After:
```
const linkKeytipData = useKeytipData({ keytipProps: keytipMap.LinkKeytip });
<Link keytipData={linkKeytipData} href="http://www.bing.com" target="_blank">
```

change/@fluentui-react-next-2020-06-22-11-18-19-xgao-useKeytipData.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Remove KeytipData for Checkbox,Toggle,Link",
```
khmakoto, line 3: nit:
```suggestion
  "comment": "Remove KeytipData for Checkbox, Toggle and Link.",
```

change/office-ui-fabric-react-2020-06-22-11-18-19-xgao-useKeytipData.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "export useKeytipData from index",
```
khmakoto, line 3: nit:
```suggestion
  "comment": "Export useKeytipData from index.",
```

packages/office-ui-fabric-react/src/components/KeytipData/useKeytipData.ts
```diff
@@ -13,6 +13,9 @@ export interface IKeytipData {
   executeElementAttributes: { [key: string]: string | undefined };
 }
 
+/**
+ * Hook that create attributes for component which enabled with Keytip.
```
khmakoto, line 17: nit:
```suggestion
 * Hook that creates attributes for components which are enabled with Keytip.
```

packages/react-next/src/compat/withKeytipData.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { IKeytipProps } from '../Keytip';
+import { KeytipDataOptions, useKeytipData, IKeytipData } from '../KeytipData';
+
+/**
+ * HOC to output wrapped component that supports keytipProps prop,
+ * given an input component that has keytipProps prop deprecated.
```
khmakoto, line 7: nit:
```suggestion
 * given an input component that has deprecated keytipProps prop.
```

packages/react-next/src/components/Checkbox/Checkbox.types.ts
```diff
@@ -129,10 +130,17 @@ export interface ICheckboxProps extends React.ButtonHTMLAttributes<HTMLElement |
...
+   * Optional keytip.
+   *
+   * @deprecated Use keytipData instead.
    */
   keytipProps?: IKeytipProps;
 
+  /**
+   * Optional keytip data.
+   */
+  keytipData?: IKeytipData;
```
khmakoto, line 142: Are we ok with still having `keytip` related props in the components?

xugao: the question here is - how can we give the keytip data to a component. one alternative way to do this is by creating a composed component. but `compose` won't be ready for all the components. the way i'm approaching (and the migration story) is applicable to all functional components we converted in V8

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -15,30 +13,20 @@ const defaultSlots: Omit<ICheckboxSlots, 'root'> = {
...
-          {...slotProps.input}
-          data-ktp-execute-target={keytipAttributes['data-ktp-execute-target']}
-          aria-describedby={mergeAriaAttributeValues(
-            slotProps.input['aria-describedby'],
-            keytipAttributes['aria-describedby'],
-          )}
-        />
+        <slots.input {...slotProps.input} />
         <slots.container {...slotProps.container}>
-          <slots.checkbox {...slotProps.checkbox} data-ktp-target={keytipAttributes['data-ktp-target']}>
```
kelseyyoung, line 41: does this mean that this data- attribute no longer gets put on the element?

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -15,30 +13,20 @@ const defaultSlots: Omit<ICheckboxSlots, 'root'> = {
...
       return <slots.label {...slotProps.label} />;
     };
 
-    // tslint:disable-next-line:no-any
-    const renderContent = (keytipAttributes: any = {}) => (
+    return (
       <slots.root {...slotProps.root}>
-        <slots.input
-          {...slotProps.input}
-          data-ktp-execute-target={keytipAttributes['data-ktp-execute-target']}
```
kelseyyoung, line 34: same question about the data- attribute here

packages/react-next/src/components/Checkbox/Checkbox.base.tsx
xugao: it still is! and exactly the same attributes as what we have currently. code is here: https://github.com/microsoft/fluentui/pull/13742/files#diff-b2490de9a8f87eb3eb767dcfd41a69ed

packages/react-next/src/components/Checkbox/Checkbox.types.ts
dzearing: this doesn't seem correct.

Keytips should be designed to be able to work with any component regardless of them knowing about keytip. The requirement is that the keytip is able to identify which element to target the tooltip towards, and which element to click when keytip is executed.

In composed components, that may mean you pass ids into the particular slots. Or you add refs. But I think we need a model that doesn't require keytipData to be on every component in the world.

packages/react-next/src/components/Keytip/examples/Keytips.Basic.Example.tsx
dzearing: Instead why not this?

```tsx
<Keytip {...keytipProps}>
  <Link href="..." target="...">...</Link>
</Keytip>
```

packages/react-next/src/components/Checkbox/Checkbox.types.ts
xugao: @dzearing  - i do agree having the `keytipData` pollutes prop interface. here is what i was thinking - the way to achieve not having `keytipData` prop is via `compose`. `compose` will allow user to pass the attributes down to any slot they want. however before we can announce `compose` to user to use, having `keytipData` prop is a nice stepping stone. 

On what you proposed:
- if we define IDs within our component, how do we communicate these "IDs" from our component to Keytip users so they know what ID to set?
- other than target/execute element ids, we also need to set `aria-describedby` within the targeted component. the `aria-describedby` is ID of element which created by `KeytipLayer` that describes the keytip sequences
image


image

packages/react-next/src/components/Keytip/examples/Keytips.Basic.Example.tsx
xugao: Updated the code:

Before:
```
<Link keytipProps={keytipMap.LinkKeytip} href="http://www.bing.com" target="_blank">
```

After:
```
const linkRef = useKeytipRef({ keytipProps: keytipMap.LinkKeytip });
<Link ref={linkRef} href="http://www.bing.com" target="_blank">
```

xugao:
@dzearing @khmakoto @kelseyyoung -  I updated the code and added `useKeytipRef` hook.

The idea is that the hook will return a ref which then user can pass the ref to whichever component keytip is targeting. The ref will allow the hook to access the root element that it's targeting and sets proper keytip attributes based on (optional) data attributes set within the root element.

kelseyyoung:
@xugao would the ref approach be problematic if the 'ref' attribute for the component was already used?

xugao:
@kelseyyoung - user can merge multiple refs into one. we have an implementation in `@fluentui/react-hooks` package [here](https://github.com/microsoft/fluentui/blob/master/packages/react-hooks/src/useMergedRefs.ts)

packages/react-next/src/components/Link/Link.types.ts
```diff
@@ -72,7 +72,9 @@ export interface ILinkProps extends ILinkHTMLAttributes<HTMLAnchorElement | HTML
   theme?: ITheme;
 
   /**
-   * Optional keytip for this Link
+   * Optional keytip.
+   *
+   * @deprecated This no longer works. Use `useKeytipData` hook instead.
```
dzearing, line 77: This doesn't seem like the right deprecation message - is there any way to have this value deprecated in v7 with an alternative implementation for customers?

dzearing: If we can deprecate it (but keep it functional)_ with the viable alternative in v7, and then remove it in v8, that's a better user experience. They can update to wrapping things before the upgrade, right now, in v7.

packages/react-next/src/components/Toggle/useToggle.ts
```diff
@@ -128,6 +130,7 @@ export const useToggle = (
       'aria-labelledby': labelledById,
       className: classNames.pill,
       'data-is-focusable': true,
+      'data-ktp-target': true,
```
dzearing, line 133: hmm not clear why this is here.

packages/react-next/src/components/Link/useLink.ts
```diff
@@ -69,7 +72,19 @@ const adjustPropsForRootType = (
...
+    as,
+    disabled,
+    target,
+    href,
+    theme,
+    getStyles,
+    styles,
+    componentRef,
+    // tslint:disable-next-line:deprecation
+    keytipProps,
```
dzearing, line 85: is this really still needed?

packages/react-next/src/components/Toggle/__snapshots__/Toggle.test.tsx.snap
```diff
@@ -12,6 +12,7 @@ exports[`Toggle renders hidden toggle correctly 1`] = `
       aria-checked={false}
       className="ms-Toggle-background"
       data-is-focusable={true}
+      data-ktp-target={true}
```
dzearing, line 15: this seems unexpected

packages/react-next/src/components/Toggle/useToggle.ts
xugao: this is needed for specifying target element when it's not the root element

packages/react-next/src/components/Link/Link.types.ts
xugao: Unfortunately user can't use the new approach on V7 components, because they don't support passing `ref` to the root element

PR #13743, opened 2020-06-22
revving CDN url and adding 2 filetype icons

bigbadcapers:
#### Pull request checklist

- [x] Addresses an existing issue: Adding missing icons for filetype


Jahnp:
Hi @JustSlone , just checking, is it ok if we go ahead and merge this? There're some product teams waiting on this change and the net bundle impact is quite low. Thanks!

bigbadcapers:
uh oh. i saw the comment _after_ i instinctively clicked the merge. @JustSlone please let me know if you have concerns and if so what additions need to be done here. I am healthily suspicious of what this perf test is baselining against, which was the only flag raised by the checks. It's hard to understand how the added png icons that are scoped to filetypes mightve impacted things like slider perf. Happy to dig in! lmk.

JustSlone:
Was just looking into this now. Nope no concerns, we look good. The perf possible regression appears to be just normal fluctuation. 

Thanks!

PR #13745, opened 2020-06-22
Adding a template in the repo for codesandbox examples

tomi-msft:
#### Pull request checklist

#### Description of changes

Creates a template to test some v7 components in codesandbox  so that we can test changes in sandboxes, which will help catch issues and bugs a lot more easily.

#### Focus areas to test

(optional)


apps/codesandbox-react-template/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@types/webpack-env": "1.15.1",
+    "@uifabric/tslint-rules": "^7.2.1",
+    "@uifabric/build": "^7.0.0"
+  },
+  "dependencies": {
+    "@uifabric/set-version": "^7.0.13",
+    "@microsoft/load-themed-styles": "^1.10.26",
+    "es6-promise": "^4.1.0",
+    "immutability-helper": "~2.8.1",
+    "office-ui-fabric-react": "^7.121.1",
```
dzearing, line 29: minor nit; I would suggest this depends on `@fluentui/react` instead of `office-ui-fabric-react`.

apps/codesandbox-react-template/src/index.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+
+import { Fabric } from 'office-ui-fabric-react/lib/Fabric';
```
dzearing, line 4: Change all the imports to one import which pulls from `@fluentui/react`.

ecraig12345:
Would you mind updating the PR description with a little more info about what this is and how it will be used?

apps/codesandbox-react-template/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@microsoft/load-themed-styles": "^1.10.26",
+    "es6-promise": "^4.1.0",
+    "immutability-helper": "~2.8.1",
+    "@fluentui/react": "7.121.2",
+    "react": "16.8.6",
+    "react-app-polyfill": "~1.0.1",
+    "react-dom": "16.8.6",
+    "typescript": "3.7.2",
+    "tslib": "^1.10.0"
+  }
```
ecraig12345, line 35: This looks like it was created with an outdated create-package template (or copy-pasted). As a starting point, the deps should be more like this. Once I understand more about the package purpose it will probably be possible to refine the list a bit more.
```suggestion
  "devDependencies": {
    "@types/react": "16.8.25",
    "@types/react-dom": "16.8.4",
    "@types/webpack-env": "1.15.1",
    "@uifabric/tslint-rules": "^7.2.1",
    "@uifabric/build": "^7.0.0"
  },
  "dependencies": {
    "@uifabric/set-version": "^7.0.13",
    "@microsoft/load-themed-styles": "^1.10.26",
    "@fluentui/react": "7.121.2",
    "react": "16.8.6",
    "react-app-polyfill": "~1.0.1",
    "react-dom": "16.8.6",
    "tslib": "^1.10.0"
  }
```

ecraig12345:
Still kind of confused, sorry--which files from this template will show up in the sandbox, and how are they used? In particular I'm wondering if these are actually needed:
- `.storybook`
- `index.html` (and `pre-copy.json` which copies `index.html` to `dist`)
- webpack configs

tomi-msft:
I don't think pre-copy.json is needed. I am not sure where index.html gets used. For .storybook, I kept it when I was running the app locally to make sure every component rendered properly. 

I can remove these, and I think it should still be functional.

ecraig12345:
Is there a way to see what the sandbox generated based on this will look like?

tomi-msft:
Besides running it locally? I'm not sure. There is supposed to be some codesandbox integration for our PRs but there have been issues with that lately

dzearing:
@lbogdan Would you at all be able to help here? We're trying to add a custom template in the repo, but running into fail.

1. We've added a template. We see that it is added, but upon clicking the link, it doesn't render. It seems that tslib fails because it's trying to extend something that doesn't exist. This seems like a webpack kind of error where something is missing in the bundle.

2. As a side note, is there a way to give the template a name so that it shows up in the PR as a named thing?

CompuIves:
Hey @dzearing. Thanks for letting us know! I did some fixes:

1. This should work properly now, we weren't handling the `baseUrl` in `tsconfig.json` correctly, but now it should work as intended!
2. That's a really good point, we'll make sure to update it in CI, and I'm thinking of changing our name generation in general to include the sandbox name it was forked from when making a fork from a template. Should be live this week!

dzearing:
@CompuIves awesome!! Thank you!

PR #13748, opened 2020-06-22
SelectedItmesList: Migrate to function components in react-next

MLoughry:
#### Pull request checklist

#### Description of changes

Migrated `SelectedItemsList` and associated components to function components in the `react-next` package.

There was a lot of very loose typing, inhertience patterns instead of composition, and extraneous code going on; so there was some cleanup in that area as well.

#### Focus areas to test

(optional)


ecraig12345:
Do you know if/how this SelectedItemsList is related to the one in experiments and the work @nebhatna has been doing there?

nebhatna:
> Do you know if/how this SelectedItemsList is related to the one in experiments and the work @nebhatna has been doing there?

Thanks! I will sync with Michael offline.

MLoughry:
> > Do you know if/how this SelectedItemsList is related to the one in experiments and the work @nebhatna has been doing there?
> 
> Thanks! I will sync with Michael offline.

It sounds like they're already working on a revamp that includes migrating to function components. I'll close (but not delete) this branch under the assumption their change will make it in prior to v8.

PR #13751, opened 2020-06-23
merge-styles: allowing css variables to be registered without being kebabcased.

dzearing:
Updating `merge-styles` to allow for css variable registration without kebab casing:

Before, type error and erroneous output:
```tsx
mergeStyles({ '--button-borderColor': 'red' }) // output: .css-0{--button-border-color: 'red'}
```
After, no red squiggles, correct output:

```tsx
mergeStyles({ '--button-borderColor': 'red' }) // output: .css-0{--button-borderColor: 'red'}
```

### DIFF
--- a/packages/merge-styles/src/IStyle.ts
+++ b/packages/merge-styles/src/IStyle.ts
@@ -7,6 +7,8 @@ import { IRawStyleBase } from './IRawStyleBase';
  * {@docCategory IRawStyle}
  */
 export interface IRawStyle extends IRawStyleBase {
+  [key: string]: string | number | undefined | Record<string, IStyle>;
+
   /**
    * Display name for the style.
    */
--- a/packages/merge-styles/src/transforms/kebabRules.test.ts
+++ b/packages/merge-styles/src/transforms/kebabRules.test.ts
@@ -16,4 +16,12 @@ describe('kebabRules', () => {
 
     expect(rules).toEqual(['-webkit-font-smoothing', 'antialiased']);
   });
+
+  it('can avoid kebabing css variables', () => {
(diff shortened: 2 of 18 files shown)


packages/merge-styles/src/IStyle.ts
```diff
@@ -7,6 +7,8 @@ import { IRawStyleBase } from './IRawStyleBase';
  * {@docCategory IRawStyle}
  */
 export interface IRawStyle extends IRawStyleBase {
+  [key: string]: string | number | undefined | Record<string, IStyle>;
```
xugao, line 10: add this in `IRawStyleBase`?

xugao: side note, would be nice to have this one day: https://github.com/Microsoft/TypeScript/issues/6579

packages/merge-styles/src/transforms/kebabRules.test.ts
```diff
@@ -16,4 +16,12 @@ describe('kebabRules', () => {
 
     expect(rules).toEqual(['-webkit-font-smoothing', 'antialiased']);
   });
+
+  it('can avoid kebabing css variables', () => {
```
xugao, line 20: let's also add a test for 
```
const rules = ['-webkit-font-smoothing', 'antialiased'];
```

packages/merge-styles/src/IStyle.ts
dzearing: i wish

dzearing: hmm. I did this first, but then it put the knowledge of the selectors in `IRawStyleBase`. I *think* this is more correct.

PR #13763, opened 2020-06-23
refactor: minor refactors and small perf improvements

layershifter:
#### Pull request checklist

### `@fluentui/react-bindings`

Separate `props` to `componentProps` and `stylesProps` to avoid destructions on objects and clarify the usage. Not affecting public API, only internal `getStyles()` function.

### `@fluentui/react-northstar`

- use improved `round()` to increase perf (credits to @miroslavstastny):

image
- small refactor in `getBorderFocusStyles.ts` to avoid useless destructions/spreads on objects

### `@fluentui/react-telemetry`

Fix to iterate on `defaultPerformanceFlags` instead of information from `localStorage`.

### `@fluentui/styles`

In `mergeThemes.ts` avoid creating useless frames and calls to `deepmerge()` until there is something to merge.


packages/fluentui/react-telemetry/src/TelemetryPerfFlags.tsx
```diff
@@ -10,7 +10,7 @@ type TelemetryPerfFlagsProps = {
 
 export const TelemetryPerfFlags: React.FC<TelemetryPerfFlagsProps> = props => (
   <div style={styles.performanceFlags()}>
-    {Object.keys(props.flags).map(flag => (
+    {Object.keys(defaultPerformanceFlags).map(flag => (
```
layershifter, line 13: It's required to iterate on `defaultPerformanceFlags` as currently we're reading them `localStorage`: it will be an issue if we will introduce a new perf flag.

packages/fluentui/react-northstar/src/utils/fontSizeUtility.ts
```diff
@@ -16,6 +15,8 @@ const getFontSizeValue = (size?: string | null): number | null => {
   return (size && parseFloat(size)) || null;
 };
 
+export const round = (n: number) => ((n * 10000 + (n > 0 ? 0.5 : -0.5)) << 0) / 10000;
```
layershifter, line 18:
```suggestion
// A replacement for a Lodash's one, way more faster on our cases as handles specific scenario
export const round = (n: number) => ((n * 10000 + (n > 0 ? 0.5 : -0.5)) << 0) / 10000;
```

packages/fluentui/react-northstar/src/themes/teams/components/Button/buttonVariables.ts
```diff
@@ -1,7 +1,9 @@
 import { FontWeightProperty } from 'csstype';
 
 import { pxToRem } from '../../../../utils';
-
+// function pxToRem(v) {
+//   return v + 'px';
+// }
```
layershifter, line 6:
```suggestion

```

packages/fluentui/react-northstar/src/themes/teams/components/Attachment/attachmentVariables.ts
```diff
@@ -56,7 +56,7 @@ export default (siteVariables: any): AttachmentVariables => ({
   focusColor: undefined,
 
   progressColor: siteVariables.colorScheme.green.background,
-  progressHeight: 4,
+  progressHeight: pxToRem(4),
```
layershifter, line 59: nit change as it was really strange to see `pxToRem(v.progressHeight)` in `attachmentStyles.ts`

packages/fluentui/react-northstar/src/utils/renderComponent.tsx
```diff
@@ -80,7 +80,8 @@ const renderComponent = <P extends {}>(
     className,
     disableAnimations: context.disableAnimations || false,
     primaryDisplayName: displayName,
-    props: stateAndProps,
+    componentProps: stateAndProps,
+    stylesProps: stateAndProps,
     renderer: context.renderer || noopRenderer,
```
layershifter, line 85: Cache usage is blocked for class components so there is no need to separate them somehow.

packages/fluentui/styles/src/mergeThemes.ts
```diff
@@ -68,7 +68,7 @@ export const mergeComponentStyles__PROD = (
       // no target means source doesn't need to merge onto anything
       // just ensure source is callable (prepared format)
       if (typeof originalTarget === 'undefined') {
-        partStylesPrepared[partName] = callable(originalSource);
+        partStylesPrepared[partName] = typeof originalSource === 'function' ? originalSource : callable(originalSource);
```
mnajdova, line 71: If the `originalSource` is not a function, then I do not understand why we are using `callable`.

packages/fluentui/react-bindings/src/hooks/useStyles.ts
```diff
@@ -113,11 +113,11 @@ const useStyles = <StyleProps extends PrimitiveProps>(
     allDisplayNames: composeOptions?.displayNames || [displayName],
     className: composeOptions?.className || className,
     primaryDisplayName: composeOptions?.displayName || displayName,
-    props: {
+    componentProps: {
```
mnajdova, line 116: Would recommend mentioning in the PR description, that the changes are referring to the `getStyles` function.

packages/fluentui/react-bindings/src/hooks/useStyles.ts
```diff
@@ -113,11 +113,11 @@ const useStyles = <StyleProps extends PrimitiveProps>(
...
       ...componentStylesProps,
-      ...mapPropsToInlineStyles(),
       ...composeStylesProps,
     },
+    stylesProps: mapPropsToInlineStyles(),
```
mnajdova, line 120: Maybe `inlineStylesProps` would be better name?

packages/fluentui/react-bindings/src/styles/resolveVariables.ts
```diff
@@ -37,9 +37,18 @@ const resolveVariables = (
     const handlingDisplayName = effectiveDisplayNames[effectiveDisplayNames.length - 1];
 
     if (!variablesThemeCache[handlingDisplayName]) {
-      variablesThemeCache[handlingDisplayName] = mergeComponentVariables(
-        ...effectiveDisplayNames.map(displayName => theme.componentVariables[displayName]),
-      )(theme.siteVariables);
+      // A short circle to avoid additional merging for non-composed components
+      if (effectiveDisplayNames.length === 1) {
+        variablesThemeCache[handlingDisplayName] =
+          typeof theme.componentVariables[handlingDisplayName] === 'function'
```
mnajdova, line 43: `callable(theme.componentVariables[handlingDisplayName])(theme.siteVariables)`?

packages/fluentui/react-bindings/src/hooks/useStyles.ts
layershifter: Done 👍

packages/fluentui/react-bindings/src/hooks/useStyles.ts
layershifter: Agree, will rename

packages/fluentui/react-bindings/src/styles/resolveVariables.ts
layershifter: Yup, it will create an additional frame but it's not important

packages/fluentui/styles/src/mergeThemes.ts
layershifter: As we always expect a function as a result of merging.

An additional check these is because otherwise each theme merging will add a useless closure on each merge and then it looks in a profiler like:

```
anonymous()
anonymous()
anonymous()
anonymous()
```

PR #13765, opened 2020-06-23
fix(form): fix narration of the label for other elements than input

kolaps33:
#### Description of changes
Recognized that when we added Dropdown or RadioGroup into the form, then screen reader was not narrated the label.

Issue was that construction of label works only with input and not with other elements. 
<label dir="auto" class="ui-form__label ci" **for="first-name-inline"** id="form-label-147">First name</label>

Therefore fix would be to add `aria-labelledby` on each element in the form. This attribute refers to the `<label>` element.

This is generic(I know we don't need it for input) but should cover different types of elements which can be in the form. 
As well this fix will be consistent. :)

As well amending the example to be able test it properly. 

#### Focus areas to test
Navigate with screen reader to each type of form field.
Expected: Screen reader narrates the label


packages/fluentui/accessibility/src/behaviors/Form/formFieldBehavior.tsx
```diff
@@ -9,7 +9,9 @@ const formFieldBehavior: Accessibility<FormFieldBehaviorProps> = props => {
       root: {},
       control: {
         ...(props.hasErrorMessage && { 'aria-invalid': true }),
-        ...(props.messageId && { 'aria-labelledby': `${props.labelId} ${props.messageId}` }),
+        ...(props.messageId
+          ? { 'aria-labelledby': `${props.labelId} ${props.messageId}` }
+          : { 'aria-labelledby': `${props.labelId}` }),
```
assuncaocharles, line 14:
```suggestion
         { 'aria-labelledby': `${props.labelId}` },
        ...(props.messageId && { 'aria-labelledby': `${props.labelId} ${props.messageId}` }),
```

packages/fluentui/docs/src/examples/components/Form/Variants/FormExampleComponents.tsx
```diff
@@ -27,8 +27,8 @@ const FormExampleComponents = () => (
   <Form>
     <Form.Input label="First name" name="firstName" id="first-name-inline" required />
     <Form.Dropdown label="City" id="cities-id" items={['prague', 'new york']} />
-    <Form.RadioGroup label="Pizza" vertical defaultCheckedValue="prosciutto" items={items} />
-    <Form.Slider />
+    <Form.RadioGroup label="Pizza" id="pizza-selection" vertical defaultCheckedValue="prosciutto" items={items} />
```
assuncaocharles, line 30: Should `id` be required prop?

kolaps33: Now I am looking deeper to it and recognized, as I change the logic for screen reader narration we would not need  id. Because this was used for `<label for='id'>`
So I could change from
` labelId.current = getOrGenerateIdFromShorthand('form-label-', id, labelId.current);`   
into 
`  labelId.current = getOrGenerateIdFromShorthand('form-label-', label, labelId.current);`

and remove all dependencies on id.

If @jurokapsiar will agree with the new reader approach then I would remove `id` prop. 
@assuncaocharles what do you think about removing id ? do we need it for some other reason like for usage in `label`  element?

jurokapsiar: from where does `label` in `getOrGenerateIdFromShorthand('form-label-', label, labelId.current);` come from?

kolaps33: sorry I didn't mention the file where we generate it, it is in file: formFieldBase.tsx

assuncaocharles: In this case we are never going to have 
```
<label for="id" />
<input id="id" />
```

And always 

```
<label id="id-pizza">pizza</label>
<input aria-labelledby="id-pizza" />
```

Right? Sounds ok for me

kolaps33: @assuncaocharles 
yes exactly. We will always narrate label with `aria-labelledby`. This should works always and consistently. :)

kolaps33: I removed ID from props.

packages/fluentui/accessibility/src/behaviors/Form/formFieldBehavior.tsx
kolaps33: done :)

PR #13769, opened 2020-06-24
 Button: Adding stories of a Button themed like Fluent using tokens

khmakoto:
#### Pull request checklist

#### Description of changes

This PR leverages tokens to create a storybook add-on that changes the theme of the `Button` between `Teams` and `Fluent`.

__Teams theme:__
image

__Fluent theme:__
image

#### Focus areas to test

(optional)


xugao:
how about we allow switching theme using storybook add on and decorator similar to here: https://github.com/microsoft/fluentui/pull/13394/files

khmakoto:
@xugao I'll look into it.

packages/react-button/.storybook/preview.js
```diff
@@ -25,6 +29,16 @@ function loadStories() {
...
+  // Wrap examples with ThemeProvider
+  for (let [key, story] of stories) {
+    Object.keys(story).forEach(exampleName => {
+      const example = story[exampleName];
+      if (typeof example === 'function') {
+        story[exampleName] = () => example();
+      }
+    });
+  }
+
```
xugao, line 41: we should be able to remove this block here and in `react-next` too. could you do that and double check nothing breaks? thanks!

PR #13770, opened 2020-06-24
React conformance - Adding as prop tests and adding isConformant to MenuItem

sopranopillow:
#### Pull request checklist

#### Description of changes

This PR adds the as prop tests:
* 'as-renders-fc': If the component has an "as" prop - Renders as functional component or passes as to the next component.
* 'as-renders-react-class': If the component has an "as" prop - Renders as ReactClass or passes as to the next component.
* 'as-passes-as-value': If the component has an "as" prop - If the component does not render any DOM, ensure it passes the as value to the next component.
* 'as-renders-html': If the component has an "as" prop - Renders component as HTML tags.

This PR also adds the consoleUtil and  getComponent utility needed for the tests above. 

As part of testing if the test do work, isConformant was added to MenuItem since it needs the wrappedComponent property and the current tests did not.


packages/react-conformance/src/defaultTests.tsx
```diff
@@ -148,4 +151,108 @@ export const defaultTests: TestObject = {
...
+          expect(component.prop('as')).toEqual(MyComponent);
+        }
+      });
+    }
+  },
+
+  /** If it has "as" prop: If the component does not render any DOM,
+   * ensure it passes the as value to the next component */
+  'as-passes-as-value': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    if (componentInfo.props.as) {
```
ecraig12345, line 210: This looks the same as the first `as` test? Unless I'm missing something.

packages/react-conformance/src/isConformant.ts
```diff
@@ -13,7 +13,9 @@ export function isConformant(testInfo: IsConformantOptions) {
     throw new Error(`Path ${componentPath} does not exist`);
   }
 
-  const parser = withCustomConfig(tsconfigPath, {});
+  const parser = withCustomConfig(tsconfigPath, {
+    propFilter: prop => !/@types[\\/]react[\\/]/.test(prop.parent?.fileName || ''),
```
ecraig12345, line 17: Can you add a comment about why we're doing this?

packages/react-conformance/src/utils/getComponent.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const getDisplayName = (Component: React.ElementType) => {
+  return (
+    (Component as React.ComponentType).displayName ||
+    (Component as React.ComponentType).name ||
+    (typeof Component === 'string' && Component.length > 0 ? Component : 'Unknown')
+  );
+};
+
+// tslint:disable-next-line:no-any
+export const toNextNonTrivialChild = (from: ReactWrapper, wrapperComponent: React.ElementType | undefined): any => {
```
ecraig12345, line 14: Would be nice if you could figure out the actual return type here (or just don't specify it and it might be inferred correctly)

packages/react-conformance/src/isConformant.ts
```diff
@@ -13,7 +13,11 @@ export function isConformant(testInfo: IsConformantOptions) {
     throw new Error(`Path ${componentPath} does not exist`);
   }
 
-  const parser = withCustomConfig(tsconfigPath, {});
+  // Props need to be filtered since react-docgen shows all the props including props
+  // that aren't specified in the component.
```
ecraig12345, line 17: I'd modify this slightly--the problem isn't all inherited props (there are actually some cases where we want inherited props), just native props (ones from DOM elements) or certain React built-ins like ref.
```suggestion
  // inherited native props or React built-in props.
```

packages/react-conformance/src/utils/getComponent.tsx
```diff
@@ -0,0 +1,37 @@
+import { ReactWrapper } from 'enzyme';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
+// import { FocusZone } from '@fluentui/react-bindings';
```
xugao, line 3: remove commented out code

packages/react-conformance/src/utils/getComponent.tsx
```diff
@@ -0,0 +1,37 @@
+import { ReactWrapper } from 'enzyme';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
```
xugao, line 2: need to add `@fluentui/react-component-ref` as dependency in package.json

ecraig12345: Good catch, I read straight over this. It's probably why the build was failing (though I'm not sure why it failed with the error messages it did...). 

I think that rather than adding a dep on react-component-ref and directly referencing these components, it would be more cross-package-friendly to make `wrapperComponent` an array (`wrapperComponents`) and pass in the components that way. That will be the solution for this PR.

In a separate PR, I think at this point it also makes sense to add a new utility under `packages/fluentui/react-northstar/test/utils` which wraps the new `isConformant` and mixes in options which will always be applicable (such as `wrapperComponents: [Ref, RefFindNode]`, or certain test disables if needed). And a similar utility for OUFR and react-next, probably under `src/common` in each package.

sopranopillow:
> See comment

I wasn't able to make the array of `React.ElementType` work for wrapperComponents. It brought some weird bugs that didn't make sense. For now I will push it with the `react-component-ref` but will come back to it later. Decided to not spend much of the time working on that to get some other more stuff done like the second part of the comment above.

xugao:
Regarding Elizabeth's comment, pls follow up with her once she back from vacation. Other than that the changes LGTM

PR #13777, opened 2020-06-24
chore: improve lazy loading in v0 docs

layershifter:
This PR improves lazy loading in v0 docs:
- prototypes and their dependencies are in a separate chunk
- playground is lazily loaded in a separate chunk as it has dependency on `faker`
- `SourceRender` and Babel for it is loaded lazily

And replaces `IgnorePlugin`:

### Before

image

### After

image

## Measures

On the main page of docs.

### Before

image

### After

image


packages/fluentui/docs/src/components/ComponentDoc/ComponentDoc.tsx
```diff
@@ -11,11 +11,14 @@ import ComponentProps from './ComponentProps';
 import { ComponentDocAccessibility } from './ComponentDocAccessibility';
 import { ThemeContext } from '../../context/ThemeContext';
 import ExampleContext from '../../context/ExampleContext';
-import ComponentPlayground from '../ComponentPlayground/ComponentPlayground';
 import { ComponentInfo } from '../../types';
 import ComponentBestPractices from './ComponentBestPractices';
 import * as _ from 'lodash';
 
+const ComponentPlayground = React.lazy(
+  /* webpackChunkName: "playground" */ () => import('../ComponentPlayground/ComponentPlayground'),
```
ecraig12345, line 19: Does the chunk name comment work correctly with this positioning? I'm used to seeing it look like `import(/* webpackChunkName: "playground" */ '../ComponentPlayground/ComponentPlayground')`

packages/fluentui/docs/src/components/ComponentDoc/ComponentPerfExample/ComponentPerfChart.tsx
```diff
@@ -6,7 +6,9 @@ const ExtremesFilter = ({ onChange }) => (
   <Checkbox label="Show extremes" onChange={(e, { checked }) => onChange({ extremes: checked })} />
 );
 
-export const ComponentPerfChart = ({ perfTestName }) => {
+const ComponentPerfChart = ({ perfTestName }) => {
```
ecraig12345, line 9: Thought you were moving away from default exports?

layershifter: Yup, good catch

packages/fluentui/docs/src/components/ComponentDoc/ComponentDoc.tsx
layershifter: Good catch, fixed 👍

layershifter:
> Have you checked the docsite still loads in IE11?

Yup, still works :)

PR #13781, opened 2020-06-24
GroupedVerticalBarChart - The second bar chart disappear when the number of charts are greater than 2 in one page -Issue fixed

jameelakowsar:
#### Pull request checklist

#### Description of changes

ID's added to chart SVG and G elements.
When we have 2 charts on same page, because of same ID's second graph getting disappeared. By Adding unique ID's, issue resolved.

#### Focus areas to test

Grouped vertical bar chart

### Before fix
image

### After fix (adding 2 charts in same page)
image

### DIFF
--- a/packages/charting/src/components/GroupedVerticalBarChart/GroupedVerticalBarChart.base.tsx
+++ b/packages/charting/src/components/GroupedVerticalBarChart/GroupedVerticalBarChart.base.tsx
@@ -72,6 +72,7 @@ export class GroupedVerticalBarChartBase extends React.Component<
   private _xScale0: any;
   // tslint:disable-next-line:no-any
   private _xScale1: any;
+  private _uniqLineText: string;
   private _dataset: IGVDataPoint[];
   private _keys: string[];
   private _isGraphDraw: boolean = true;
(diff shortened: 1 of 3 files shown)


packages/charting/src/components/GroupedVerticalBarChart/GroupedVerticalBarChart.base.tsx
```diff
@@ -156,12 +158,12 @@ export class GroupedVerticalBarChartBase extends React.Component<
 
     return (
       <div
-        id="d3GroupedChart"
+        id={`d3GroupedChart_${this._uniqLineText}`}
```
joschect, line 161: I might make this a const or method for consistency.

packages/charting/src/components/GroupedVerticalBarChart/GroupedVerticalBarChart.base.tsx
```diff
@@ -174,10 +176,14 @@ export class GroupedVerticalBarChartBase extends React.Component<
               className={this._classNames.yAxis}
               transform={`translate(40, 0)`}
             />
-            <g id="barGElement" />
+            <g id={`barGElement_${this._uniqLineText}`} className="barGElement" />
```
joschect, line 179: See above comment. If you put the id generation in a centralized place, it will be easier to change if you need to.

PR #13786, opened 2020-06-24
Add Tile props to streamline drag/drop behaviors

ThomasMichon:
The `Tile` component has some issues with drag/drop support, notably that dragging from its foreground and background elements can sometimes disrupt the drag. In addition, there is no way to mark a `Tile` with a custom style indicating that a drag or drop is happening. This change introduces some new props to make it easier to implement drag/drop for `Tile` smoothly.


xugao:
@ThomasMichon - sorry for the delayed review, approved! next time feel free to ping me directly if it's not getting attention.

ecraig12345:
@ThomasMichon Looks like this is ready to merge. Feel free to merge once the build succeeds if it's still needed (I think you're an admin, or ask me) or please close if not.

PR #13790, opened 2020-06-24
Button: Using correct focus borders from utilities

khmakoto:
#### Pull request checklist

#### Description of changes

This PR makes it so that the `Button` uses the correct focus rectangles from our utilities but adds extra methods to these for allowing for box-shadow support as well.

#### Focus areas to test

(optional)


packages/react-button/src/common/_focusBorder.scss
```diff
@@ -0,0 +1,101 @@
...
+
+  @include focus($onFocus) {
+    &:after {
+      @include after-box-shadow($hOffset, $vOffset, $blur, $thickness, $color)
+    }
+  }
+}
+
+// When focus is set using the keyboard, apply a boxShadow.
+@mixin after-box-shadow ($hOffset: 0, $vOffset:0, $blur: 0, $thickness: 1px, $color: $focusBorderColor) {
```
dzearing, line 84: feels like this file should be in something like a `@fluentui/common-sass` package. Not in this PR but obviously other packages will need some helpers.

dzearing: I believe you're moving this file here to avoid circular deps, but we'll need to re-address this.

PR #13794, opened 2020-06-25
Improve bundle test perf

xugao:
#### Pull request checklist

#### Description of changes
Draft - Need to see if this helps 

#### Focus areas to test

(optional)

### DIFF
--- a/scripts/webpack/webpack-resources.js
+++ b/scripts/webpack/webpack-resources.js
@@ -79,7 +79,7 @@ module.exports = {
    * @param {boolean} [excludeSourceMaps] - whether to skip generating source maps
    * @returns {webpack.Configuration[]} array of configs
    */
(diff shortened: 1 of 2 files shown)


scripts/webpack/webpack-resources.js
```diff
@@ -235,24 +235,18 @@ module.exports = {
...
         reportFilename: bundleName + '.stats.html',
         openAnalyzer: false,
-        generateStatsFile: true,
-        statsOptions: {
-          source: false,
-          reasons: false,
-          chunks: false,
-        },
-        statsFilename: bundleName + '.stats.json',
+        generateStatsFile: false,
```
dzearing, line 249: I believe the stats file is what sizeauditor consumes.

xugao: yup. so what i realized is generating the stats file is the slow part. i moved it to size-auditor. but i don't have a solution to reduce that cost yet :(

PR #13803, opened 2020-06-25
CalendarDayGrid: migrate component to function component

MLoughry:
#### Pull request checklist

#### Description of changes

Migrates CalendarDayGrid to (a set of) functional components, per the ongoing effort in `@fluentui/react-next` to make all components React Strict-mode-compliant. Since the OUFR version of Calendar/DateGrid are being deprecated by this package, I am migrating the components in place.

#### Focus areas to test

(optional)


lorejoh12:
Tested the functionality, only major bug I found was that the animation direction seems messed up. First Calendar example, navigate to next month, navigate back to this month, and select a day. The animation for navigating backwards will "replay" during selection. Does not repro in http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/13801/merge/date-time/dist/index.html#/examples/calendar (some other random build without your changes, because the date-time package doesn't get deployed to the normal testing site)

PR #13808, opened 2020-06-25
Re-enable bundle tests for react-next package without stats

xugao:
#### Pull request checklist

#### Description of changes
Stats generation is time consuming and it slows down CI quite a lot. Not the most ideal, but I am going to re-enable `react-next` (which previously disabled due to perf) without stats generation. This means we can still get bundle size results but won't be able to drill down what's causing the size change from size auditor website. Developers can still drill down failures by locally running the bundle tests.  I think this is the best work around for now to still have basic bundle size gate for `react-next` without spending too much time on its perf issue. 

I also see we are "profiling" for any webpack production build (in our shared webpack config script). I removed that and made it opt-in only.

here is result from CI: https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=86435&view=results

#### Focus areas to test

(optional)


apps/test-bundles/webpack.nostats.config.js
```diff
@@ -0,0 +1,11 @@
+const { createWebpackConfig, buildEntries } = require('./webpackUtils');
+
+/**
+ * Stats is generated and used in size auditor report for checking more details on what affected bundle size change.
+ * Due to stats generation is slow and it slows down CI significantly, we recommend to configure experimental packages without stats generation.
+ * If size is changed unexpectedly, devs can drill deep more locally.
```
khmakoto, line 6:
```suggestion
 * Stats are generated and used by the size auditor report to check more details on what caused the bundle size change.
 * Due to stats generation being slow, and therefore slowing down CI significantly, we recommend to configure experimental packages without stats generation.
 * If bundle size is changed unexpectedly, developers can drill down deeper on the problem by locally running bundle tests.
```

apps/test-bundles/webpackUtils.js
```diff
@@ -0,0 +1,100 @@
...
+      true,
+      {
+        entry: {
+          [entryName]: entries[entryName],
+        },
+        externals: {
+          react: 'React',
+          'react-dom': 'ReactDOM',
+        },
+        plugins: [new BundleAnalyzerPlugin(anaylizerPluginOptions)],
```
joschect, line 53: This is a slight tweak that is definitely not necessary: 

Since this is the only part of each entry that could change via `includeStats` you could pass anaylizerPluginOptions in via each entry. That means you only need 1 webpack file and you would configure the react-next entry to pass the "no-stats" options through.

xugao: i thought about this. but i went with having 2 configs (as i thought that's cleaner). let me double check

xugao: fixed!

PR #13809, opened 2020-06-25
Replace the use of Lerna with Lage for CI builds

kenotron:
#### Description of changes

This PR is both a proposal and demo of what is like to have `lage` replace CI builds in place of `lerna`.

`lage` is a task runner for monorepos that was created as a brainchild between the FAST team and the Office Online (Flywheel) teams. The main reasons why one would choose `lage` over `lerna` are:

1. Scoping per package tasks given a repo wide understanding of the dependency graph of tasks
2. Caching task results via backfill for a build-over-build incremental build capability
3. Proper scheduling of tasks (pipelining)

You can see how this is impacted in this chart-heavy page: https://microsoft.github.io/lage/guide/levels.html

As this PR stands, we apply several changes to the build process:

1. jest is set to "runInBand" when running under lage since they would both compete with CPU - 1 concurrency causing thrashing
2. rather than relying on build artifacts, all the steps start independent of each other doing yarn + build - subsequent builds of PRs can simply take advantage of prebuilt caches
3. lage is turned on for all the steps that are topologically run by lerna - scope and cache are turned on for these
4. delete files is replaced with the clean workspace functionality

#### Initial Results

image

For this PR so far, the builds are NOT CACHED and NOT SCOPED yet since the change itself prompts the build to actually rebuild everything and nothing has been cached just yet. Barring the screener failure (which is a strange timeout issue on the part of SauceLabs), it looks like the build would complete around 18 minutes. The current avg build time is ~38 minutes. This is a significant reduction. 

#### Focus areas to test

(optional)


azure-pipelines.tools.yml
```diff
@@ -2,7 +2,7 @@
 steps:
   - task: NodeTool@0
     inputs:
-      versionSpec: '10.x'
+      versionSpec: '12.x'
```
ecraig12345, line 5: Do we have to switch to 12? Switching in CI increases changes that we'll accidentally break 10, which a lot of people are still using. (Unless we want to make everyone switch to 12--I see 10 is now in "maintenance" status rather than side-by-side LTS with 12.)

azure-pipelines.yml
```diff
@@ -11,12 +11,12 @@ variables:
 pool: 'Self Host Ubuntu'
 
 jobs:
-  - job: Build
+  - job: Validation
```
ecraig12345, line 14: There might be a better name for this--just "Validation" makes it sound like it's running tests, which it isn't.

.devops/templates/pr-target-branch.yml
```diff
@@ -0,0 +1,4 @@
+steps:
```
ecraig12345, line 1: Can the other pipeline files also be moved into .devops?

azure-pipelines.yml
```diff
@@ -25,22 +25,24 @@ jobs:
...
+      - template: .devops/templates/pr-target-branch.yml
 
-      - publish: $(Build.ArtifactStagingDirectory)
-        artifact: Build-PR-$(Build.BuildNumber)
+      - script: |
+          yarn lage build test lint --since $(target_branch) --verbose
+        displayName: build, test, lint
+        env:
+          BACKFILL_CACHE_PROVIDER: 'azure-blob'
+          BACKFILL_CACHE_PROVIDER_OPTIONS: '{"connectionString":"$(BACKFILL_CONNECTION_STRING)", "container":"$(BACKFILL_CONTAINER)"}'
```
ecraig12345, line 45: Is this going to be a security issue to have these exposed in a public pipeline? (meaning is there anything else someone could do if they have the connection string?)

lage.config.js
```diff
@@ -0,0 +1,37 @@
+module.exports = {
+  pipeline: {
```
ecraig12345, line 2: Is there documentation on what these settings do?

.devops/templates/pr-target-branch.yml
kenotron: probably only after we DUPLICATE them and then letting the build jobs get changed in the definition... but not for this PR.

azure-pipelines.yml
kenotron: I actually created a separate storage account just for this purpose. The leak would be carefully crafted hashed cache to bring about a code injection. Although this is already possible as a change in yml. 

For ultimate security, we should be doing a pipeline feature to not allow non collaborators forks to run by default... I think...

azure-pipelines.yml
```diff
@@ -203,33 +157,26 @@ jobs:
...
+      - template: .devops/templates/pr-target-branch.yml
+
       - script: |
           git config --global user.email "fabrictactical@microsoft.com"
           git config --global user.name "Fabric Tactical"
-          yarn vrtest -- --debug
+        displayName: config git user and email
+
+      - script: |
+          yarn lage vrtest --debug --since $(target_branch) --no-deps --verbose
```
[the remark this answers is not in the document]
kenotron: This gets us building & getting to vrtest as quickly as possible. Not all things need to be built, so it's scoped to that (naturally as well because only one project has vrtest task I think)

lage.config.js
kenotron: Yes: https://microsoft.github.io/lage/guide/config.html

It's not as comprehensive as I would like, but we do have some docs on this.

azure-pipelines.yml
ecraig12345: This one could be pretty easily leaked by changing the yml or a build script and making a PR. (By comparison, that's not such a concern for tokens used in the release pipeline because you'd have to get a bad change checked in.)

If a feature exists which limits variable availability to only collaborators' PRs, that would be great, but I kind of doubt that exists. :(

azure-pipelines.tools.yml
kenotron: I'm gonna try going back to 10... I think I had tried this for speed's sake, but I'll just stick with 10 for reasons you raised.

azure-pipelines.yml
kenotron: I got rid of this extra job. Putting back inside Build, Test

lage.config.js
ecraig12345: Oh okay. Maybe put a link in a file comment for ease of finding it?

Also am I understanding right that leading `^` means "do dependencies first"?

ecraig12345: Oh also can you add docs to the page for `ignore` and `cacheOptions`? (I'm not sure what the `ignore` things are being ignored from)

azure-pipelines.yml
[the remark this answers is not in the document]
ecraig12345: Okay, so this is basically like our current `yarn buildto`? (I couldn't immediately tell based on experience with other tools what this would do.)

lage.config.js
kenotron: > Oh okay. Maybe put a link in a file comment for ease of finding it?
> 
> Also am I understanding right that leading `^` means "do dependencies first"?

Yes that's it!

> Oh also can you add docs to the page for ignore and cacheOptions? (I'm not sure what the ignore things are being ignored from)

I'll add some comments here in this file first - will add docs later. I'm organizing myself to be able to work on these docs this week.

azure-pipelines.yml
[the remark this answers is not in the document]
kenotron: yes, that's right.

.devops/templates/pr-target-branch.yml
```diff
@@ -0,0 +1,4 @@
...
+  - script: |
+      node -e "process.stdout.write('##vso[task.setvariable variable=target_branch]origin/$(System.PullRequest.TargetBranch)')"
```
ecraig12345, line 3: Does `echo` not work here?

azure-pipelines.yml
```diff
@@ -11,36 +11,28 @@ variables:
...
-          # Include root folder so archive can just be extracted into $(Pipeline.Workspace)
-          includeRootFolder: true
-          archiveType: 'tar'
-          tarCompression: 'gz'
-          archiveFile: $(Build.ArtifactStagingDirectory)/build.tar.gz
-          replaceExistingArchive: true
-
-      - publish: $(Build.ArtifactStagingDirectory)
-        artifact: Build-PR-$(Build.BuildNumber)
+          yarn lage build test lint --verbose --no-cache
```
ecraig12345, line 31: The use of lage has the same benefit or better compared to doing these in separate jobs, right?

azure-pipelines.yml
```diff
@@ -122,6 +71,13 @@ jobs:
         env:
           DANGER_GITHUB_API_TOKEN: $(DANGER_GITHUB_API_TOKEN)
 
+      - script: |
+          yarn lage build --since $(target_branch) --scope @fluentui/* --no-deps --verbose --no-cache
```
ecraig12345, line 75: This seems mostly redundant with the main build/test/lint step--is all of that really needed here?

scripts/tasks/jest.js
```diff
@@ -5,7 +5,7 @@ const path = require('path');
 
 exports.jest = () =>
   jestTask({
-    ...(process.env.TF_BUILD && { runInBand: true }),
+    ...((process.env.TF_BUILD || process.env.LAGE_PACKAGE) && { runInBand: true }),
```
ecraig12345, line 8: Are there any other usages of TF_BUILD where things need to be changed?

JustSlone:
@kenotron and @ecraig12345 we're really close on this PR! Let's see if we can get it through and give it a go in the repo. Would love to see the speed benefits in our daily workflow.

PR #13810, opened 2020-06-26
Checkbox: use CheckMark icon from react-icons

xugao:
#### Pull request checklist

#### Description of changes
Use icon from `react-icons` brings faster perf and smaller bundle size and added `checkmark` slot prop support.
Migration story for user who previously using `checkmarkIconProps` is to use `checkmark` slot prop.

#### Focus areas to test

(optional)


apps/vr-tests/src/stories/CheckboxNext.stories.tsx
```diff
@@ -67,6 +68,12 @@ storiesOf('Checkbox Next', module)
         },
       }}
     />
+  ))
+  .addStory('Custom render CheckMark icon', () => (
+    <Checkbox
+      label="Checkbox with customized icon"
+      checkmark={{ children: () => <Icon iconName="CheckboxComposite" /> }}
```
xugao, line 75: Note: we should be able to do 
```
checkmark={ <Icon iconName="CheckboxComposite" /> }
```
here, but i found we are not handling the case when parent slot is a function component (in checkbox's case, we have `checkmark` slot being `CheckMarkIcon`) correctly in `resolveSlotProps`.
I will get a fix going for this in a separate PR.

cc @dzearing

xugao: here is the fix: https://github.com/microsoft/fluentui/pull/13827

dzearing: Interesting scenario; Slot is an icon, so having JSX in the prop would make it a child.

Slots in compose are the container. The container can choose how to manage children.

Passing JSX in makes the content children of the slot, rather than a replacement.

So, technically, what we could do to support this case is for the CheckmarkIcon to support children:

```tsx
const CheckmarkIcon = (p) => (
  p.children || (<svg/>)
);
```
...but then maybe this is a bad idea in that now there are 2 components in the React heirarchy rather than a single one for the icon.

But your fix is correct as well in that to replace the slot, you have to provide the children function.

Is the design right? Is it completely unexpected? One other consideration is standards:

https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_templates_and_slots

In web components, slots get replaced, which is more like how you expect it to work.

xugao: @dzearing - i think when you pass JSX, you want it to *replace* the slot completely. if you do want to keep the outer parent slot layer, you would use `children` function. let's continue our discussion in #13827 :)

PR #13811, opened 2020-06-26
Plumb event information through for spin button increment/decrement

makopch-ms:
#### Description of changes

There is no way for a caller of spin button to determine if the spinner was incremented or decremented by mouse or keyboard.  We already have all the event information so this allows us to pass it through

### DIFF
--- a/change/office-ui-fabric-react-2020-06-25-16-57-56-passEventSpinButton.json
+++ b/change/office-ui-fabric-react-2020-06-25-16-57-56-passEventSpinButton.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "pass through event information if the user spins up or down",
+  "packageName": "office-ui-fabric-react",
+  "email": "makopch@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-06-25T23:57:56.349Z"
+}
--- a/change/office-ui-fabric-react-2020-06-25-16-57-56-passEventSpinButton.json
+++ b/change/office-ui-fabric-react-2020-06-25-16-57-56-passEventSpinButton.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "pass through event information if the user spins up or down",
+  "comment": "SpinButton: pass through event information if the user spins up or down",
   "packageName": "office-ui-fabric-react",
   "email": "makopch@microsoft.com",
   "dependentChangeType": "patch",
--- a/packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
+++ b/packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
@@ -355,9 +355,12 @@ export class SpinButton extends React.Component<ISpinButtonProps, ISpinButtonSta
     return String(newValue);
   };
 
-  private _onIncrement = (value: string): string | void => {
+  private _onIncrement = (
+    value: string,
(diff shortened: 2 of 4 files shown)


change/office-ui-fabric-react-2020-06-25-16-57-56-passEventSpinButton.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "pass through event information if the user spins up or down",
```
khmakoto, line 3: nit:
```suggestion
  "comment": "SpinButton: Pass through event information if the user spins up or down.",
```

packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.types.ts
```diff
@@ -121,15 +121,25 @@ export interface ISpinButtonProps extends React.HTMLAttributes<HTMLDivElement> {
 
   /**
    * Callback for when the increment button or up arrow key is pressed.
+   * @param value - The current value to be incremeneted
```
khmakoto, line 124:
```suggestion
   * @param value - The current value to be incremeneted
```
```suggestion
   * @param value - The current value to be incremented
```

packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
```diff
@@ -545,11 +556,11 @@ export class SpinButton extends React.Component<ISpinButtonProps, ISpinButtonSta
     }
   };
 
-  private _onIncrementMouseDown = (): void => {
-    this._updateValue(true /* shouldSpin */, this._initialStepDelay, this._onIncrement!);
+  private _onIncrementMouseDown = (event: React.MouseEvent<HTMLElement>): void => {
```
khmakoto, line 559: Should this be?
```suggestion
  private _onIncrementMouseDown = (event?: React.MouseEvent<HTMLElement>): void => {
```

packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
```diff
@@ -545,11 +556,11 @@ export class SpinButton extends React.Component<ISpinButtonProps, ISpinButtonSta
...
+    this._updateValue(true /* shouldSpin */, this._initialStepDelay, this._onIncrement!, event);
   };
 
-  private _onDecrementMouseDown = (): void => {
-    this._updateValue(true /* shouldSpin */, this._initialStepDelay, this._onDecrement!);
+  private _onDecrementMouseDown = (event: React.MouseEvent<HTMLElement>): void => {
```
khmakoto, line 563: Should this be?
```suggestion
  private _onDecrementMouseDown = (event?: React.MouseEvent<HTMLElement>): void => {
```

packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
makopch-ms: Since this is directly passed to the onMouseDown, it is required if you don't just omit it altogether.

packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.tsx
makopch-ms: see above comment

PR #13812, opened 2020-06-26
MenuButton: Creating component using compose

khmakoto:
#### Pull request checklist

#### Description of changes

This is a first stab at a `MenuButton` component using `compose`. At the moment I'm making using of `ContextualMenu` to build this but only on the opinionated version and we can swap it for whatever once we have a more refined new `Menu` component.

MenuButton

#### Focus areas to test

(optional)


packages/react-button/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { ContextualMenu } from 'office-ui-fabric-react';
+import { MenuButtonBase } from './MenuButtonBase';
+import { MenuButtonProps } from './MenuButton.types';
+import * as classes from './MenuButton.scss';
+
+export const MenuButton = compose<'button', MenuButtonProps, MenuButtonProps, {}, {}>(MenuButtonBase, {
+  classes: createClassResolver(classes),
+  displayName: 'MenuButton',
+  slots: {
+    menu: ContextualMenu,
```
dzearing, line 11: menuIcon should be opinionated here I believe.

dzearing: import the chevron from react-icons.

packages/react-button/src/components/MenuButton/MenuButton.types.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+import { ComposeOptions } from '@fluentui/react-compose';
+import { ComposeStandardStatics, ShorthandValue } from '../../utils/tempTypes';
+import { ButtonProps, ButtonTokens } from '../Button/Button.types';
+
+export interface MenuButtonProps extends Omit<ButtonProps, 'icon' | 'iconPosition' | 'loader'> {
+  /**
```
joschect, line 7: Out of curiosity, why can't a menu button have an icon to the left?

packages/react-button/src/components/MenuButton/MenuButton.types.tsx
```diff
@@ -0,0 +1,56 @@
...
+   * Shorthand menu that is displayed when the button is pressed.
+   */
+  menu?: ShorthandValue<{}>;
```
xugao, line 10: I added proper typing for `ShorthandValue` in `react-compose`. it's called `SlotProp`. pls use that and remove the ones from tempTypes

packages/react-button/src/components/MenuButton/MenuButton.types.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export interface MenuButtonState extends MenuButtonProps {
+  /**
+   * Defines the React reference to the MenuButton element.
+   */
+  ref: (instance: HTMLElement) => void;
+}
+
+export interface MenuButtonSlots {
```
xugao, line 44: extends BaseSlots

packages/react-button/src/components/MenuButton/MenuButton.types.tsx
```diff
@@ -0,0 +1,56 @@
...
+  menu: React.ElementType;
+  menuIcon: React.ElementType;
+}
+
+export type MenuButtonSlotProps = {
```
xugao, line 49: use `SlotProps` to construct this. see button.types

packages/react-button/src/components/MenuButton/MenuButton.types.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
```
xugao, line 1: this file should have extension `ts`, not `tsx`

packages/react-button/src/components/MenuButton/MenuButton.types.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export type MenuButtonSlotProps = {
+  [key in keyof MenuButtonSlots]: MenuButtonProps[key];
+};
+
+export interface MenuButtonOptions
+  extends ComposeOptions<MenuButtonProps, MenuButtonSlots, MenuButtonSlotProps, ComposeStandardStatics> {}
+
+export type MenuButtonTokens = ButtonTokens & {};
```
xugao, line 56: what's `& {}` for?

packages/react-button/src/components/MenuButton/MenuButtonBase.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const MenuButtonBase = compose<'button', MenuButtonProps, MenuButtonProps, {}, {}>(
+  (props, ref, options) => {
+    const { state } = options;
+    const { slots, slotProps } = mergeProps(state, options);
+
+    const { children, expanded, iconOnly } = state;
+
+    return (
+      <>
```
xugao, line 14: remove `<>` layer, not needed

packages/react-button/src/components/MenuButton/MenuButtonBase.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <slots.root ref={ref} {...slotProps.root}>
+          {!iconOnly && <span>{children}</span>}
```
xugao, line 16: shouldn't we also check ` && children`?

packages/react-button/src/components/MenuButton/MenuButtonBase.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
+import { compose, mergeProps } from '@fluentui/react-compose';
+import { MenuButtonProps } from './MenuButton.types';
+import { useMenuButton } from './useMenuButton';
+
+export const MenuButtonBase = compose<'button', MenuButtonProps, MenuButtonProps, {}, {}>(
+  (props, ref, options) => {
+    const { state } = options;
+    const { slots, slotProps } = mergeProps(state, options);
```
xugao, line 9: pass generic types for `mergeProps<...>`, so you can actually get typing for slots/slotProps

packages/react-button/src/components/MenuButton/useMenuButton.ts
```diff
@@ -0,0 +1,92 @@
...
+          return;
+        }
+      }
+
+      setExpanded(!expanded);
+    },
+    [expanded, onClick],
+  );
+
+  const onDismiss = React.useCallback(() => {
```
xugao, line 46: don't need to `useCallback` for these handlers. creating functions should not be costly. memoize only when needed

packages/react-button/src/components/ToggleButton/useToggle.ts
```diff
@@ -10,22 +10,29 @@ interface ToggleProps {
...
 ): TState & ToggleState => {
-  const { checked: controlledChecked, defaultChecked = false, onClick: onButtonClick } = props;
+  const { checked: controlledChecked, defaultChecked = false, onClick, ...rest } = props;
   const [checked, setChecked] = useControllableValue(controlledChecked, defaultChecked);
 
-  const onClick = (ev: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
-    if (onButtonClick) {
-      onButtonClick(ev);
-    }
+  const _onClick = React.useCallback(
```
xugao, line 22: same here. is `useCallback` really needed?

packages/react-button/src/components/MenuButton/useMenuButton.ts
```diff
@@ -0,0 +1,92 @@
...
+  const state: MenuButtonState = {
+    ...rest,
+    'aria-expanded': expanded,
+    expanded,
+    onClick: _onClick,
+    onKeyDown: _onKeyDown,
+    ref: useMergedRefs(ref, buttonRef),
+
+    // Menu slot props.
+    menu: {
```
xugao, line 80: you need to use `mergeSlotProp` from `react-compose` here. so the shorthand/slot prop is properly merged

packages/react-button/src/components/MenuButton/useMenuButton.ts
```diff
@@ -0,0 +1,92 @@
...
+      ...(menu as object),
+      directionalHint: DirectionalHint.bottomRightEdge,
+      onDismiss,
+      target: buttonRef && buttonRef.current,
+    },
+
+    // Menu icon slot props.
+    menuIcon: menuIcon ? menuIcon : { children: '˅' },
```
xugao, line 88: use `react-icons` for chevron here?

packages/react-button/src/components/ToggleButton/useToggle.ts
```diff
@@ -10,22 +10,29 @@ interface ToggleProps {
 interface ToggleState extends ToggleProps {}
 
 /**
- * The useToggleButton hook processes the Button component props and returns state.
+ * The useToggleButton hook processes the ToggleButton component props and returns state.
  * @param props - ToggleButton props to derive state from.
  */
 export const useToggle = <TProps, TState extends TProps = TProps>(
   props: TProps & ToggleProps,
 ): TState & ToggleState => {
```
xugao, line 18: I might have asked this before. why is `TState` needed here?

packages/react-button/src/components/MenuButton/MenuButton.types.tsx
khmakoto: I wanted to omit extra icons when having a `MenuButton` besides the `menuIcon` itself.

This brings a good point of discussion though:
- Should we allow the `menuIcon` to be on the left as well?
- Should we allow for an `icon` besides the `menuIcon`?
- If so, should we allow that `icon` to be able to exist before and after the content or just before?

packages/react-button/src/components/MenuButton/MenuButton.types.tsx
khmakoto: I was expecting to add more tokens here but didn't need to for now. I'll remove it.

packages/react-button/src/components/MenuButton/useMenuButton.ts
khmakoto: I'll use it in the opinionated version but not here.

packages/react-button/src/components/ToggleButton/useToggle.ts
khmakoto: Because we're going to be returning something that is more than just the `defaultChecked`, `checked`, `onClick` props we're using here. The idea is that we can use this `useToggle` hook on other components besides the `ToggleButton` later on.

packages/react-button/src/components/MenuButton/MenuButton.scss
```diff
@@ -0,0 +1,5 @@
+@import '../Button/Button.scss';
+
+.menuIcon + * {
```
xugao, line 3: i think  we should just do something like
```
.menu {
  margin: 0;
}
```

packages/react-compose/src/types.ts
```diff
@@ -141,6 +141,17 @@ export type ComposePreparedOptions<TProps = {}, TInputState = any, TParentState
 // Component types
 //
 
+export interface ComponentProps {
+  // Removing these props:
+  // design - use style or className instead
```
xugao, line 146: i don't understand these comments?

packages/react-compose/src/types.ts
```diff
@@ -141,6 +141,17 @@ export type ComposePreparedOptions<TProps = {}, TInputState = any, TParentState
...
+
+  as?: React.ElementType;
+
+  className?: string;
+
+  root?: SlotProp<React.HTMLAttributes<HTMLElement>>;
```
xugao, line 152: i don't think `root` is quite right. what if we want to be accurate?
`SlotProp<React.ButtonHTMLAttributes<HTMLButtonElement>>`

packages/react-button/src/components/MenuButton/useMenuButton.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    // Menu slot props.
+    menu: mergeSlotProp(menu, {
+      directionalHint: DirectionalHint.bottomRightEdge,
+      onDismiss,
+      target: buttonRef && buttonRef.current,
+    }),
+  };
+
+  return useButton(state, ref, options) as MenuButtonState;
```
xugao, line 81: i think instead of `as MenuButtonState` here. we should fix typing of `useButton` to be similar to what you had for `useToggle`

packages/react-button/src/components/MenuButton/MenuButton.types.ts
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react';
+import { BaseSlots, SlotProp, SlotProps } from '@fluentui/react-compose';
+import { ButtonProps, ButtonTokens } from '../Button/Button.types';
+
+export interface MenuButtonProps extends Omit<ButtonProps, 'icon' | 'iconPosition' | 'loader'> {
+  /**
+   * Shorthand menu that is displayed when the button is pressed.
+   */
+  // tslint:disable-next-line:no-any
+  menu?: SlotProp<any>;
```
xugao, line 10: shouldn't here be `SlotProp<IContextualMenuProps>`?

packages/react-button/src/components/MenuButton/MenuButton.types.ts
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react';
+import { BaseSlots, SlotProp, SlotProps } from '@fluentui/react-compose';
+import { ButtonProps, ButtonTokens } from '../Button/Button.types';
+
+export interface MenuButtonProps extends Omit<ButtonProps, 'icon' | 'iconPosition' | 'loader'> {
+  /**
+   * Shorthand menu that is displayed when the button is pressed.
```
xugao, line 7:
```suggestion
   *Menu that is displayed when the button is pressed.
```

packages/react-button/src/components/MenuButton/MenuButton.types.ts
```diff
@@ -0,0 +1,55 @@
...
+   */
+  // tslint:disable-next-line:no-any
+  menu?: SlotProp<any>;
+
+  /**
+   * Shorthand menu icon that is displayed when the button is pressed.
```
xugao, line 13:
```suggestion
   * Menu icon that opens menu when pressed.
```

packages/react-button/src/components/MenuButton/MenuButton.scss
khmakoto: This does not work, at least if we use `ContextualMenu`, because it is created in a separate `Layer`.

packages/react-compose/src/types.ts
khmakoto: Removed.

packages/react-compose/src/types.ts
khmakoto: These `typings` are at the compose layer so I tried to do the most general as possible. I've since changed this to use generic typings.

packages/react-button/src/components/MenuButton/useMenuButton.ts
khmakoto: I think it is a different scenario since `useToggle` is supposed to be a more generic thing that could be reusable across different components while `useButton` is very much linked to the `Button` component. However, I'm happy to discuss this if need be.

packages/react-button/src/components/MenuButton/MenuButton.types.ts
khmakoto: I don't think we want to marry `Button` and `ContextualMenu` just yet. Specially since there is a new `Menu` planned somewhere in the future. Happy to discuss this as well.

packages/react-button/src/components/MenuButton/useMenuButton.ts
xugao: i don't quite follow. where i am getting at is we should not need to do `as MenuButtonState`

packages/react-button/src/components/MenuButton/MenuButton.types.ts
xugao: but right now it is using `ContextualMenu`. so we should just use the correct typing? we can also change it when we actually using a new menu

khmakoto: I'll make the change for now since we can revisit later as you said.

packages/react-button/src/components/MenuButton/useMenuButton.ts
khmakoto: I've improved the typings so that we don't need this cast.

packages/react-button/src/components/MenuButton/MenuButton.types.ts
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+import { BaseSlots, SlotProp, SlotProps } from '@fluentui/react-compose';
+import { IContextualMenuProps } from 'office-ui-fabric-react';
+import { ButtonProps, ButtonState, ButtonTokens } from '../Button/Button.types';
+
+export interface MenuButtonProps extends Omit<ButtonProps, 'icon' | 'iconPosition' | 'loader'> {
+  /**
+   * Menu that is displayed when the button is pressed.
+   */
+  // tslint:disable-next-line:no-any
```
xugao, line 10: tslint exception can be removed

packages/react-button/src/components/MenuButton/MenuButton.types.ts
```diff
@@ -0,0 +1,51 @@
...
+  menu?: SlotProp<Partial<IContextualMenuProps>>;
```
xugao, line 11: hm why is `Partial` needed here?

khmakoto: It was needed because using `mergeSlotProp` in `useMenuButton` on the `menu` slot required for me to send an `items` objects since that is required in the `IContextualMenuProps` interface. I've changed this to specify partial only on the `mergeSlotProp` call and not on the `menu` slot itself.

PR #13813, opened 2020-06-26
ContextualMenu: Fix several issue with focus

joschect:
#### Pull request checklist

#### Description of changes

Fix 2 issues:
1. Fix an issue where focus would not return to the previously focused element if the contextualmenu was hidden instead of being removed from the dom.
2. Fix an issue where an error could sometimes throw if the previously focused element was removed from the dom before the contextualmenu was dismissed.

#### Focus areas to test

(optional)

### DIFF
--- a/change/office-ui-fabric-react-2020-06-25-22-59-46-minor-focus-fixes.json
+++ b/change/office-ui-fabric-react-2020-06-25-22-59-46-minor-focus-fixes.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "ContextualMenu: Fix issue where focus would not return to button. Also fix issue where error might occur if the previously focused element didn't have a focus method.",
+  "packageName": "office-ui-fabric-react",
+  "email": "joschect@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-06-26T05:59:46.400Z"
+}
--- a/change/office-ui-fabric-react-2020-06-25-22-59-46-minor-focus-fixes.json
+++ b/change/office-ui-fabric-react-2020-06-25-22-59-46-minor-focus-fixes.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "ContextualMenu: Fix issue where focus would not return to button. Also fix issue where error might occur if the previously focused element didn't have a focus method.",
+  "comment": "ContextualMenu: Fixing issue where focus would not return to previously focused element and issue where error might occur if said previously focused element didn't have a focus method.",
   "packageName": "office-ui-fabric-react",
   "email": "joschect@microsoft.com",
   "dependentChangeType": "patch",
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -1215,7 +1215,7 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
       this.dismiss(ev, true);
 
       // This should be removed whenever possible.
-      // This ensures that the hidden dismissal behavior maintains the same.
+      // This ensures that the hidden dismissal action maintains the same behavior.
       // If the menu is being dismissed then the previously focused element should
       // get focused since the dismiss was triggered by a user click on an item
       // Rather than focus being lost.
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -452,6 +452,7 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     if (options && options.containsFocus && this._previousActiveElement) {
       // Make sure that the focus method actually exists
       // In some cases the object might exist but not be a real element.
+      // This is primarily for IE 11 and should be removed once IE 11 is no longer in use.
       if (this._previousActiveElement.focus) {
         this._previousActiveElement && this._previousActiveElement.focus();
       }
--- a/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
@@ -427,6 +427,7 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
       containsFocus: this._focusingPreviousElement,
       originalElement: this._previousActiveElement,
     });
+    this._focusingPreviousElement = false;
 
     if (this.props.onMenuDismissed) {
       this.props.onMenuDismissed(this.props);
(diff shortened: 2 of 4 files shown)


khmakoto:
Do these fixes have any issues that need to be attached?

change/office-ui-fabric-react-2020-06-25-22-59-46-minor-focus-fixes.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "ContextualMenu: Fix issue where focus would not return to button. Also fix issue where error might occur if the previously focused element didn't have a focus method.",
```
khmakoto, line 3: nit:
```suggestion
  "comment": "ContextualMenu: Fixing issue where focus would not return to previously focused element and issue where error might occur if said previously focused element didn't have a focus method.",
```

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -1205,7 +1210,16 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
       dismiss = !!this.props.onItemClick(ev, item);
     }
 
-    (dismiss || !ev.defaultPrevented) && this.dismiss(ev, true);
+    if (dismiss || !ev.defaultPrevented) {
+      this.dismiss(ev, true);
+
+      // This should be removed whenever possible.
+      // This ensures that the hidden dismissal behavior maintains the same.
```
khmakoto, line 1217: This sentence is confusing to me, did you mean:
```suggestion
      // This ensures that the hidden dismissal action maintains the same behavior.
```

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.test.tsx
```diff
@@ -1039,6 +1039,36 @@ describe('ContextualMenu', () => {
       button.current!.openMenu();
       expect(document.querySelector('.SubMenuClass')).toEqual(null);
     });
+
+    it('Menu should correctly return focus to button when dismissed', () => {
```
khmakoto, line 1043: nit:
```suggestion
    it('Menu should correctly return focus to previously focused element when dismissed', () => {
```

packages/office-ui-fabric-react/src/components/Popup/Popup.tsx
```diff
@@ -173,6 +173,10 @@ function defaultFocusRestorer(options: { originalElement?: HTMLElement | Window;
   const { originalElement, containsFocus } = options;
 
   if (originalElement && containsFocus && originalElement !== window) {
-    originalElement.focus();
+    // Make sure that the focus method actually exists
+    // In some cases the object might exist but not be a real element.
+    if (originalElement.focus) {
```
khmakoto, line 178: nit: Collapse both if conditions into one.

joschect: I want to keep them separate because this shouldn't need to exist once IE 11 is gone. I'll add that as a comment.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -449,7 +450,11 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     originalElement: HTMLElement | Window | undefined;
   }) => {
     if (options && options.containsFocus && this._previousActiveElement) {
-      this._previousActiveElement && this._previousActiveElement.focus();
+      // Make sure that the focus method actually exists
+      // In some cases the object might exist but not be a real element.
+      if (this._previousActiveElement.focus) {
+        this._previousActiveElement && this._previousActiveElement.focus();
+      }
```
xugao, line 457: can simplify:
```
if (.... && this. _previousActiveElement?.focus) {
  this._previousActiveElement.focus();
}
```

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -1205,7 +1210,16 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
...
-    (dismiss || !ev.defaultPrevented) && this.dismiss(ev, true);
+    if (dismiss || !ev.defaultPrevented) {
+      this.dismiss(ev, true);
+
+      // This should be removed whenever possible.
+      // This ensures that the hidden dismissal behavior maintains the same.
+      // If the menu is being dismissed then the previously focused element should
+      // get focused since the dismiss was triggered by a user click on an item
+      // Rather than focus being lost.
+      this._focusingPreviousElement = true;
```
xugao, line 1221: shouldn't this logic be inside `this.dismiss`?

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -960,7 +965,7 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     let handled = false;
 
     if (shouldHandleKey(ev)) {
-      this._focusingPreviousElement = false;
+      this._focusingPreviousElement = true;
```
xugao, line 969: this change looks wrong to me?

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
joschect: Sadly no, there are other things that can trigger dismiss other than a click. That's why it's in this click method and in keydown.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.test.tsx
```diff
@@ -1039,6 +1039,36 @@ describe('ContextualMenu', () => {
       button.current!.openMenu();
       expect(document.querySelector('.SubMenuClass')).toEqual(null);
     });
+
+    it('Menu should correctly return focus to previously focused element when dismissed', () => {
+      const temp = ReactTestUtils.renderIntoDocument<HTMLDivElement>(
+        <div>
```
xugao, line 1045: nit: can remove <div> wrapper

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
xugao: gotcha

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
joschect: This was broken before. The original was setting it to true, I set it to false here incorrectly in the previous change.

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.test.tsx
joschect: I cannot, at least not if I still want to get Temp and use it as an HTMLElement. At least that's how the typings seem to work out.

joschect:
Any last minute comments? This should be ready to go in!

PR #13821, opened 2020-06-26
HoverCard: Expose Optional native props for the host div of HoverCard

RajeshGoriga:
#### Pull request checklist

#### Description of changes

HoverCard: Expose Optional native props for the host div of HoverCard

 Need to pass accessibility props for host like role and aria-label and aria expanded to resolve this issue https://office.visualstudio.com/OC/_workitems/edit/4221810 in this bug 7 more is the outer host element i want pass aria props for this host element.

example :
`<HoverCard
        type={HoverCardType.plain}
        plainCardProps={plainCardProps}
        instantOpenOnClick={true}
        onCardHide={onHoverCardHideHandler}
        setInitialFocus={true}
        trapFocus={true}
        onCardVisible={this._hoverCardVisible}
        styles={classNames.subComponentStyles.hoverCardStyles}
        cardDismissDelay={300}
      >
        <div
          role={'combobox'}
          aria-expanded={this.state.isHoverCardVisible}
          aria-label={`${items.length} ${overflowString}`}
          data-is-focusable={true}
        >
          <div className={classNames.overflowIndicationTextStyle}>
            {items.length} {overflowString}
          </div>
        </div>
      </HoverCard>`

Here i have added role aria-expanded and aria-label but since it is wrapped with hover card host i am unable to use this as expected. 

#### Focus areas to test
Hover card 
        https://github.com/microsoft/fluentui/blob/fb9e9f54a7a5ef35e56bc9962c20e6f2dedebf86/packages/charting/src/components/Legends/Legends.base.tsx#L243

### DIFF
--- a/change/office-ui-fabric-react-2020-06-26-17-04-04-HostNativePropsForHoverCard.json
+++ b/change/office-ui-fabric-react-2020-06-26-17-04-04-HostNativePropsForHoverCard.json
@@ -1,5 +1,5 @@
 {
-  "type": "patch",
+  "type": "minor",
   "comment": "Expose Optional native props for the host div of HoverCard",
   "packageName": "office-ui-fabric-react",
   "email": "v-gorraj@microsoft.com",
--- a/change/office-ui-fabric-react-2020-06-26-17-04-04-HostNativePropsForHoverCard.json
+++ b/change/office-ui-fabric-react-2020-06-26-17-04-04-HostNativePropsForHoverCard.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Expose Optional native props for the host div of HoverCard",
+  "packageName": "office-ui-fabric-react",
+  "email": "v-gorraj@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-06-26T11:34:04.337Z"
+}
--- a/packages/office-ui-fabric-react/src/components/HoverCard/HoverCard.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/HoverCard/HoverCard.base.tsx
@@ -138,6 +138,7 @@ export class HoverCardBase extends React.Component<IHoverCardProps, IHoverCardSt
       plainCardProps,
       trapFocus,
       setInitialFocus,
+      hostNativeProps,
     } = this.props;
     const { isHoverCardVisible, mode, openMode } = this.state;
     const hoverCardId = id || getId('hoverCard');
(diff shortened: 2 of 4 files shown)


packages/office-ui-fabric-react/src/components/HoverCard/HoverCard.types.ts
```diff
@@ -97,6 +97,12 @@ export interface IHoverCardProps extends React.HTMLAttributes<HTMLDivElement> {
    */
   onCardExpand?: () => void;
 
+  /**
+   * Optional Native props (s) for the host div of HoverCard
+   */
+
+  hostNativeProps?: React.HTMLAttributes<HTMLDivElement>;
```
Vitalius1, line 104: We already pass these props to the card element. What is the scenario for needing to pass the native props to the outer host element? I could not find anything in the PR description nor any issue it's fixing.

packages/office-ui-fabric-react/src/components/HoverCard/HoverCard.base.tsx
```diff
@@ -167,6 +168,7 @@ export class HoverCardBase extends React.Component<IHoverCardProps, IHoverCardSt
         ref={this._hoverCard}
         aria-describedby={setAriaDescribedBy && isHoverCardVisible ? hoverCardId : undefined}
         data-is-focusable={!Boolean(this.props.target)}
+        {...(hostNativeProps && { ...getNativeProps(hostNativeProps!, divProperties) })}
```
Vitalius1, line 171: Same question here... what is your use case for needing these?

change/office-ui-fabric-react-2020-06-26-17-04-04-HostNativePropsForHoverCard.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
Vitalius1, line 2: The API is changed so it needs to be `minor` type.

packages/office-ui-fabric-react/src/components/HoverCard/HoverCard.types.ts
RajeshGoriga: Yes we pass these props to card element but i needed for outer host why because i need to pass accessibility props for host like role and aria-label and aria expanded to resolve this issue https://office.visualstudio.com/OC/_workitems/edit/4221810

packages/office-ui-fabric-react/src/components/HoverCard/HoverCard.base.tsx
RajeshGoriga: To pass accessibility props for outer host element

packages/office-ui-fabric-react/src/components/HoverCard/HoverCard.types.ts
Vitalius1: Not sure how you use the HoverCard in your case. I assume, and correct me if I am wrong, from what you say you use a plain text wrapped by the `HoverCard` component and you don't provide a `target` prop. In that case you just need to wrap `7 more` with an html element and then wrap it with the HoverCard, make it focusable and put the accessibility attributes on that element. Not sure if adding another prop is the right solution. 

@micahgodbolt, who is leading the accessibility of controls in the core team? Would like their opinion on adding such a wide open prop that allows for so many attributes to be set.

PR #13827, opened 2020-06-26
resolveSlotProps: when slot/shorthand prop has JSX element as children, slot should be completely replaced by it

xugao:
#### Pull request checklist

#### Description of changes
In case we have
```
<Button icon={<MyIcon>} />
```
we should have icon slot completely replaced by `<MyIcon>`, not 
```
<ParentIconSlot>
  <MyIcon>
</ParentIconSlot>
```

In case if you do want to keep the `<ParentIconSlot>` wrapper, you should do
```
<Button icon={{
  children: (Component, props) => (
    <Component>
      <MyIcon>
    </Component>
  )
}} />
```

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-compose-2020-06-26-11-31-26-xgao-compose-fix.json
+++ b/change/@fluentui-react-compose-2020-06-26-11-31-26-xgao-compose-fix.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "resolveSlotProps: when slot/shorthand prop has JSX element as children, slot should be completely replace by it.",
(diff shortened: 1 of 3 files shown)


change/@fluentui-react-compose-2020-06-26-11-31-26-xgao-compose-fix.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "resolveSlotProps: when slot/shorthand prop has JSX element as children, slot should be completely replace by it.",
```
khmakoto, line 3: nit:
```suggestion
  "comment": "resolveSlotProps: Making change so that if slot/shorthand prop has JSX element as children, slot is completely replaced by it.",
```

PR #13830, opened 2020-06-26
Deeply merge props within foundation slots APIs.

### LINKED ISSUE #13829: getSlots doesn't deeply merge props
When building components with Slots APIs, props are not deeply merged. So if you have deep props that need this there is no way to solve.

Example: You have a HoverCard with some default props set but still want to allow overrides through slots without losing defaults unless the consumer explicitly does so.

#### Actual behavior:
No deep props merging. If there is a deeper prop, overriding it in any way will blow out any defaults for the whole thing.

#### Expected behavior:
Props merge deeply

Requested priority: Normal

brandonthomas:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #13829

#### Description of changes
When building out components with slots utilities it was discovered that props would not deeply merge. If there are deep default props in the underlying implementation, overriding anything will blow the whole object away. So to support this, this change updates slots to use merge instead of assign.

- Update merge to properly handle arrays and merge arrays if the object keys across two merging objects are both arrays. Also updated tests.
- Update CollapsibleSection Accordian example text. I needed to do this because I wasn't sure if there was some issue with the object merging etc.

#### Focus areas to test
Slots components still work.


packages/utilities/src/merge.test.ts
```diff
@@ -87,4 +87,23 @@ describe('merge', () => {
       foo: { bar: 'baz' },
     });
   });
+
+  it('can handle arrays', () => {
```
dzearing, line 91: add a case to verify what happens when you mix and array in target with any of these values in source:

* null/undefined (does it clear the array? I *think so*)
* string, number, boolean (does it append the value to the array or replace? I think it replaces the array.)
* object (does it replace the array? I think replace)

brandonthomas: Updated various tests and removed concat option since I think it could negatively impact scenarios like props.children.

JustSlone:
Closing and re-opening to re-spin the checks

JustSlone:
Closing and re-opening to re-spin the checks

ecraig12345:
@brandonthomas Is this still needed? If so @dzearing can you review? Also looks like the most recent PR build failed, so I'm triggering a new one.

brandonthomas:
@ecraig12345 yup still needed just haven't been able to make time to go back in and fix this. It's failing some tests but not sure why at this point. So needs debugging.

JustSlone:
@brandonthomas circling back here. Do you mind converting this PR to draft while it's on hold? Feel free to come back to it when you find time, and let us know if you need someone to review.

khmakoto (changes requested):

PR #13839, opened 2020-06-29
fix(FormInput): Forwards ref properly 

assuncaocharles:
#### Pull request checklist

#### Description of changes

Forwards ref to input control

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Input/Input.tsx
```diff
@@ -274,7 +275,7 @@ const Input = compose<'div', InputProps, InputStylesProps, {}, {}>(
                       type,
                       required,
                       value: value || '',
-                      ref,
+                      // ref,
```
assuncaocharles, line 278:
```suggestion
```

PR #13841, opened 2020-06-29
fix(compose): update types to provide better infers

layershifter:
#### Pull request checklist

#### Description of changes

See comments on code inside the PR.


packages/react-compose/src/compose.ts
```diff
@@ -5,7 +5,7 @@ import { wasComposedPreviously } from './wasComposedPreviously';
 import { mergeComposeOptions } from './mergeComposeOptions';
 
 function compose<
-  TElementType extends React.ElementType,
+  TElementType extends keyof JSX.IntrinsicElements,
```
layershifter, line 8: `TElementType` should be always an HTML tag as refs and props will be inferred from it.

packages/react-compose/src/types.ts
```diff
@@ -13,23 +13,27 @@ export interface ShorthandConfig<TProps> {
 
 export type PropsOfElement<
   // tslint:disable-next-line:no-any
-  E extends keyof JSX.IntrinsicElements | React.JSXElementConstructor<any>
-> = JSX.LibraryManagedAttributes<E, React.ComponentPropsWithRef<E>>;
+  E extends keyof JSX.IntrinsicElements | React.JSXElementConstructor<any> | ComponentWithAs
+> = E extends { __PRIVATE_PROPS: any }
+  ? E['__PRIVATE_PROPS']
+  : JSX.LibraryManagedAttributes<E, React.ComponentPropsWithRef<E>>;
```
layershifter, line 19: It's a hack for `ComponentWithAs` as otherwise it will be resolved to:

```
(props: Record<any, any> & { as?: React.ElementType } & SomeProps) => JSX.Element
```

As TS struggles to resolve `TExtendedElementType`.

***

The idea of `__PRIVATE_PROPS` is based on the assumption that it's possible to use `as` only once:

```tsx
<Button /> => // button 
<Button as='div' /> => // div
<Button as={Link} /> => // a
// ??? No way to pass `as` for `Link` 👍 
```

In this case we can assign `__PRIVATE_PROPS` to a default value of `TElementType` to simplify the resolutions.

packages/react-compose/src/types.ts
```diff
@@ -13,23 +13,27 @@ export interface ShorthandConfig<TProps> {
...
 // tslint:disable-next-line:interface-name
-export interface ComponentWithAs<TElementType extends React.ElementType = 'div', TProps = {}>
-  extends React.FunctionComponent {
-  <TExtendedElementType extends React.ElementType = TElementType>(
-    props: Omit<PropsOfElement<TExtendedElementType>, 'as' | keyof TProps> & { as?: TExtendedElementType } & TProps,
-  ): JSX.Element | null;
-  displayName?: string;
-
-  defaultProps?: Partial<TProps & { as: TElementType }>;
+export type ComponentWithAs<TElementType extends keyof JSX.IntrinsicElements = 'div', TProps = {}> = (<
```
layershifter, line 22: I reverted this thing to be case because of strange resolutions of TS: https://github.com/microsoft/TypeScript/issues/39313

***

`ComponentWithAs` is good enough to be used of other our components even without `compose()`, but before this change it marked `props` as `any`:

```tsx
const Divider: ComponentWithAs<'div', DividerProps> = (props /* becomes any 💣 */) => {}
```

PR #13843, opened 2020-06-29
ContextualMenu accessibility: Fix high contrast style for submenu icon

erichdev:
#### Pull request checklist

#### Description of changes

Fix high contrast style for ContextualMenu submenu icon. In "high contrast white", the icon currently does not have enough contrast with background on focus/hover.

**Before**

high contrast

**After**

high contrast fix

#### Focus areas to test

(optional)


packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.cnstyles.ts
```diff
@@ -24,6 +24,11 @@ const getItemHighContrastStyles = memoizeFunction(
           borderColor: 'Highlight',
           color: 'HighlightText',
           MsHighContrastAdjust: 'none',
+          selectors: {
```
xugao, line 27: can you add this style to `subMenuIcon: { ... }` block below instead targeting the icon using className?

erichdev: Thanks @xugao I've moved the change down to `subMenuIcon`

xugao: Thanks!

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.cnstyles.ts
```diff
@@ -225,6 +225,9 @@ export const getMenuItemStyles = memoizeFunction(
           [MediumScreenSelector]: {
             fontSize: IconFontSizes.medium, // 16px
           },
+          [HighContrastSelector]: {
+            color: 'HighlightText',
```
xugao, line 229: This should be WindowText color

erichdev: @xugao  The background in high contrast is `Highlight`, so should the text color be `HighlightText`, similar to how it is for the root text and for `iconColor`? 

I tested locally setting it to `WindowText`, and in high contrast white, the icon looks similar to the "Before" picture in the description, where there's not much contrast with the background.

xugao: Ah you are right.

xugao: I merged the change for you:)

PR #13845, opened 2020-06-29
chore: remove withSafeTypeForAs

layershifter:
#### Pull request checklist

#### Description of changes

This PR replaces `withSafeTypeForAs()` calls with usage to:
- avoid potential side-effects that may be caused by these calls (https://github.com/microsoft/fluentui/issues/12953#issuecomment-633970215)
- open a way to remove default exports from V0 code
#### Focus areas to test

(optional)


packages/fluentui/docs/src/components/ComponentDoc/ComponentDocSee.tsx
```diff
@@ -15,12 +15,13 @@ const ComponentDocSee: any = ({ displayName }) => {
   return (
     <List styles={listStyle}>
       {/* Heads up! Still render empty lists to reserve the whitespace */}
-      <List.Item>
+      <List.Item index={0}>
```
layershifter, line 18: `index` is a required prop on `ListItem` for Children API, https://github.com/microsoft/fluent-ui-react/pull/2207

Previous typings were not correct because we exported `ListItemProps` only:
https://github.com/microsoft/fluentui/blob/c46010712f3515f5f6fba9966b5c18546ae94cea/packages/fluentui/react-northstar/src/components/List/ListItem.tsx#L92
https://github.com/microsoft/fluentui/blob/c46010712f3515f5f6fba9966b5c18546ae94cea/packages/fluentui/react-northstar/src/components/List/ListItem.tsx#L287

packages/fluentui/docs/src/prototypes/SearchPage/SearchPage.tsx
```diff
@@ -81,7 +81,6 @@ class SearchPage extends React.Component<SearchPageState, any> {
 
         <p>
           <Input
-            ref="input"
```
layershifter, line 84: Strings can be used only for deprecated refs API: https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs. But there is no such usage 🧛

packages/fluentui/docs/src/prototypes/hexagonalAvatar/index.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
-import { Flex, StatusProps, WithAsProp, Extendable, Text } from '@fluentui/react-northstar';
+import { Flex, StatusProps, Extendable, Text } from '@fluentui/react-northstar';
 import CustomAvatar from './CustomAvatar';
 import { AcceptIcon } from '@fluentui/react-icons-northstar';
 
-const statusProps: Extendable<WithAsProp<StatusProps>> = {
```
layershifter, line 7: There is no sense in it as `as` is not used inside

packages/fluentui/react-northstar/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -126,7 +134,7 @@ const AccordionTitle: React.FC<WithAsProp<AccordionTitleProps>> &
     mapPropsToBehavior: () => ({
       hasContent: !!content,
       canBeCollapsed,
-      as,
+      as: String(as),
```
layershifter, line 137: https://github.com/microsoft/fluentui/blob/6525619bf5e6c844406ed3cb9ad3103237eff89f/packages/fluentui/accessibility/src/behaviors/Accordion/accordionTitleBehavior.ts#L48

String is expected in the behavior

packages/fluentui/react-northstar/src/components/Portal/Portal.tsx
```diff
@@ -79,14 +79,10 @@ export interface PortalProps extends ChildrenComponentProps, ContentComponentPro
...
 }
 
-export interface PortalState {
-  open?: boolean;
-}
-
 /**
  * A Portal allows to render children outside of their parent.
  */
-const Portal: React.FC<WithAsProp<PortalProps>> & FluentComponentStaticProps<PortalProps> = props => {
```
layershifter, line 89: `Portal` does not have `ElementType`...

packages/fluentui/react-northstar/src/components/Portal/Portal.tsx
```diff
@@ -79,14 +79,10 @@ export interface PortalProps extends ChildrenComponentProps, ContentComponentPro
   onOutsideClick?: (e: React.MouseEvent) => void;
 }
 
-export interface PortalState {
```
layershifter, line 82: Oops, cruft

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -161,7 +160,16 @@ const iterateItems = (items: TreeProps['items'] | TreeItemProps['items'], acc =
...
+/**
+ * A Tree displays data organised in tree hierarchy.
+ *
+ * @accessibility
+ * Implements [ARIA TreeView](https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView) design pattern.
+ * @accessibilityIssues
+ * [Treeview - JAWS doesn't narrate position for each tree item](https://github.com/FreedomScientific/VFO-standards-support/issues/338)
+ * [Aria compliant trees are read as empty tables](https://bugs.chromium.org/p/chromium/issues/detail?id=1048770)
+ */
+const Tree: ComponentWithAs<'div', TreeProps> &
```
miroslavstastny, line 172: The change from `ul` to `div` is intentional bugfix, correct?

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1621,4 +1614,11 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
  */
-export default withSafeTypeForAs<typeof Dropdown, DropdownProps>(Dropdown);
+export default (Dropdown as unknown) as ComponentWithAs<'div', DropdownProps> & {
+  Item: typeof DropdownItem;
+  SelectedItem: typeof DropdownSelectedItem;
+
+  a11yStatusCleanupTime: number;
+  charKeyPressedCleanupTime: number;
+};
+/* ^ A temporary typing until Dropdown will not be converted to functional component. */
```
miroslavstastny, line 1624:
```suggestion
/* ^ A temporary typing until Dropdown is converted to functional component. */
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
layershifter: Yep, as `Tree` and its subcomponents render `div`s:

image

I also double checked this with @jurokapsiar and we will avoid `ul`/`li` as they are breaking markup once you add custom scrollbar or virtualization

PR #13847, opened 2020-06-29
mergestyles: IStyle: loosening typings for objects embedded within the style.

dzearing:
The `IStyle` interface was recently tweaked to allow for CSS variables to be provided. The typings were too restrictive, loosening the indexer to make things more compatible with previous versions.


dzearing:
I am not convinced this was a regression; the typings are correct. The issues we've found so far were where typings were not explicit and thus inferred by TypeScript either as an ambiguous value requiring explicit typing (by casting a style object to IStyle), or just wrong. 

However, it's important that we support css variables. The original change loosened typings. It seems the only way we can do that and preserve backwards compatibility is to loosen the typings even more.

PR #13849, opened 2020-06-30
Updating Shimmer and Stack examples to use function components

czearing:
#### Pull request checklist

#### Description of changes

1. Updated `Shimmer` and `Stack` examples to use Function Components.
2. Added `useSetInterval` and `useSetTimeout` within `react-hooks` package.

#### Focus areas to test
`Shimmer` and `Stack`


packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -1,8 +1,18 @@
...
 import { ShimmeredDetailsList } from 'office-ui-fabric-react/lib/ShimmeredDetailsList';
+import { useSetInterval } from '@uifabric/react-hooks';
+
+export interface IShimmerApplicationExampleState {
+  lastIntervalId: number;
+  visibleCount: number;
+}
+
+const ITEMS_COUNT = 200;
+const INTERVAL_DELAY = 1200;
```
khmakoto, line 13: Is the change from 2500 to 1200 here intended?

dzearing: yes. It is minor, but I think it makes it more obvious what's going on.

packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -31,140 +41,92 @@ const fileIcons: { name: string }[] = [
...
+export const ShimmerApplicationExample: React.FunctionComponent = () => {
+  const safeSetInterval = useSetInterval();
+  const [items, setItems] = React.useState<IExampleItem[] | undefined>(undefined);
 
-    return (
-      <div>
-        <Toggle
-          label="Toggle to load content"
-          style={{ display: 'block', marginBottom: '20px' }}
-          checked={isDataLoaded}
```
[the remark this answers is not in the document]
dzearing: Yes, I worked with @czearing to simplify this whole example. I think it does what was originally intended with less code now. We don't need a separate state to track this; we can simply say: when we don't have items, render the list with shimmer enabled. That was the original logic but due to the overall complexity of the example it was hard to understand what it was demonstrating.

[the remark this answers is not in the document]
dzearing: I'm still thinking it probably could be even more condensed. There is no need to use file type icons or even custom columns to demonstrate how shimmer works.

packages/react-hooks/src/useSetInterval.ts
```diff
@@ -0,0 +1,30 @@
+import { useConst } from './useConst';
+import { useEffect } from 'react';
+
+/**
+ * Returns a self disposing setTimeout function.
```
xugao, line 5:
```suggestion
 * Returns a self disposing setInterval function.
```

packages/react-hooks/src/useSetInterval.ts
```diff
@@ -0,0 +1,30 @@
...
+    () => () => {
+      for (const id of intervalIds) {
+        clearInterval(id);
+      }
+      intervalIds.splice(0);
+    },
+    [],
+  );
+
+  return (func: () => void, duration: number) => {
```
xugao, line 20: you should `useConst` here so we don't create a new setInterval function on every render

xugao, line 20: you can make the typing better:
```
type setIntervalType = typeof window.setInterval;

....

return (...args: Parameters<setIntervalType>): ReturnType<setIntervalType> => {
  ...
}
```

packages/react-hooks/src/useSetInterval.ts
```diff
@@ -0,0 +1,30 @@
...
+    const id = (setInterval(() => {
+      intervalIds.splice(intervalIds.indexOf(id), 1);
+      func();
+    }, duration) as unknown) as number;
```
xugao, line 24: not sure why you need to cast `as unknown` here?

packages/react-hooks/src/useSetTimeout.ts
```diff
@@ -0,0 +1,30 @@
+import { useConst } from './useConst';
+import { useEffect } from 'react';
+
+/**
+ * Returns a self disposing setTimeout function.
+ */
+export const useSetTimeout = () => {
```
xugao, line 7: similar comments from `useSetInterval` applies here too

packages/react-hooks/src/useSetInterval.ts
ecraig12345: I think I added a similar comment to this function in another PR. @czearing Can you please make a separate PR with just these hooks (integrating the comments I made elsewhere)?

packages/react-hooks/src/useSetInterval.ts
ecraig12345: Do you need a `...` for passing `rest` to the real `setInterval`?

packages/react-hooks/src/useSetInterval.ts
ecraig12345: The types for setInterval get messed up due to presence of `@types/node` in the package. In Node it returns some other type, not a number like in the browser.

packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -1,8 +1,18 @@
...
+
+export interface IShimmerApplicationExampleState {
+  lastIntervalId: number;
+  visibleCount: number;
+}
+
+const ITEMS_COUNT = 200;
+const INTERVAL_DELAY = 1200;
+const toggleStyling = { marginBottom: '20px' };
+const shimmeredDetailsListProps = { renderedWindowsAhead: 0, renderedWindowsBehind: 0 };
```
ecraig12345, line 15: Add a type?

packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -31,140 +41,92 @@ const fileIcons: { name: string }[] = [
...
+      break;
+    }
   }
+  return columns;
+};
 
-  public componentWillUnmount(): void {
-    this._async.dispose();
-  }
+const shimmerColumns: IColumn[] = buildShimmerColumns();
```
ecraig12345, line 68: This and the items should be done within the component using `useConst(() => /*initialize*/)`

packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -31,140 +41,92 @@ const fileIcons: { name: string }[] = [
...
+const exampleItems: IExampleItem[] = createListItems(ITEMS_COUNT).map((item: IExampleItem) => {
+  const randomFileType = randomFileIcon();
+  return { ...item, thumbnail: randomFileType.url };
+});
 
-  public render(): JSX.Element {
-    const { items, columns, isDataLoaded } = this.state;
+export const ShimmerApplicationExample: React.FunctionComponent = () => {
+  const safeSetInterval = useSetInterval();
+  const [items, setItems] = React.useState<IExampleItem[] | undefined>(undefined);
```
ecraig12345, line 76: Based on the code below it looks like this should be `(IExampleItem | null)[] | undefined`.

packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -31,140 +41,92 @@ const fileIcons: { name: string }[] = [
...
+      clearInterval(state.lastIntervalId);
     }
-    this.setState({
-      isDataLoaded: checked,
-      items: items,
-    });
   };
 
-  private _onRenderItemColumn = (item: IExampleItem, index: number, column: IColumn): JSX.Element | string | number => {
+  const onRenderItemColumn = (item: IExampleItem, index: number, column: IColumn): JSX.Element | string | number => {
```
ecraig12345, line 101: This can go outside the component body

packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -31,140 +41,92 @@ const fileIcons: { name: string }[] = [
...
+  return (
+    <div>
+      <Toggle
+        label="Toggle to load content"
+        style={toggleStyling}
+        onChange={onLoadData}
+        onText="Content"
+        offText="Shimmer"
+      />
+      <div>
```
ecraig12345, line 117: This wrapping div might not be needed

packages/office-ui-fabric-react/src/components/Stack/examples/Stack.Horizontal.Configure.Example.tsx
```diff
@@ -118,23 +148,8 @@ const HorizontalStackConfigureExampleContent: React.FunctionComponent<IExampleOp
...
-const sectionStackTokens: IStackTokens = { childrenGap: 10 };
-const configureStackTokens: IStackTokens = { childrenGap: 20 };
-
-const shadowItemCheckboxStyles: Partial<ICheckboxStyles> = { root: { marginRight: 10 } };
-const wrapItemCheckboxStyles: Partial<ICheckboxStyles> = { root: { marginBottom: 10 } };
-
-export class HorizontalStackConfigureExample extends React.Component<{}, IExampleOptions> {
-  public state: IExampleOptions = {
+export const HorizontalStackConfigureExample: React.FunctionComponent = () => {
+  const [state, setState] = React.useState<IExampleOptions>({
```
ecraig12345, line 152: It seems kind of un-idiomatic for function components to have a single large state object like this. You should probably consider breaking it up into individual `useState` calls (or other patterns if more appropriate for certain members).

packages/office-ui-fabric-react/src/components/Stack/examples/Stack.Horizontal.WrapAdvanced.Example.tsx
```diff
@@ -64,123 +84,101 @@ const HorizontalStackWrapAdvancedExampleContent: React.FunctionComponent<IExampl
...
     result.push(i);
   }
   return result;
-}
+};
 
-export class HorizontalStackWrapAdvancedExample extends React.Component<{}, IExampleOptions> {
-  public state: IExampleOptions = {
+export const HorizontalStackWrapAdvancedExample: React.FunctionComponent = () => {
+  const [state, setState] = React.useState<IExampleOptions>({
```
ecraig12345, line 96: Same comment about splitting up the state object.

packages/office-ui-fabric-react/src/components/Stack/examples/Stack.Horizontal.Configure.Example.tsx
```diff
@@ -33,6 +33,36 @@ export interface IExampleOptions {
...
+];
+
+// Tokens definition
+const sectionStackTokens: IStackTokens = { childrenGap: 10 };
+const configureStackTokens: IStackTokens = { childrenGap: 20 };
+
+const shadowItemCheckboxStyles: Partial<ICheckboxStyles> = { root: { marginRight: 10 } };
+const wrapItemCheckboxStyles: Partial<ICheckboxStyles> = { root: { marginBottom: 10 } };
+
+const range = (start: number, end: number): number[] => {
```
ecraig12345, line 58: Can you add this to `packages/utilities/src/array.ts` and use that everywhere?

packages/office-ui-fabric-react/src/components/Stack/examples/Stack.Vertical.Configure.Example.tsx
```diff
@@ -120,8 +134,8 @@ function _range(start: number, end: number): number[] {
 const sectionStackTokens: IStackTokens = { childrenGap: 10 };
 const configureStackTokens: IStackTokens = { childrenGap: 20 };
 
-export class VerticalStackConfigureExample extends React.Component<{}, IExampleOptions> {
-  public state: IExampleOptions = {
+export const VerticalStackConfigureExample: React.FunctionComponent = () => {
+  const [state, setState] = React.useState<IExampleOptions>({
```
ecraig12345, line 138: Same comment about the state object

packages/office-ui-fabric-react/src/components/Stack/examples/Stack.Vertical.WrapAdvanced.Example.tsx
```diff
@@ -61,115 +61,112 @@ const VerticalStackWrapAdvancedExampleContent: React.FunctionComponent<IExampleO
...
+  { key: 'space-evenly', text: 'Space evenly' },
+];
+const overflowOptions: IDropdownOption[] = [
+  { key: 'visible', text: 'Visible' },
+  { key: 'auto', text: 'Auto' },
+  { key: 'hidden', text: 'Hidden' },
+];
+
+export const VerticalStackWrapAdvancedExample: React.FunctionComponent = () => {
+  const [state, setState] = React.useState<IExampleOptions>({
```
ecraig12345, line 85: same

packages/office-ui-fabric-react/src/components/Stack/examples/Stack.Horizontal.Configure.Example.tsx
dzearing: This helper creates an array of incrementing numbers. It probably should be named `createNumberList` or something, but I don't think we would need a helper like that in array utilities; possibly test-utilities though.

ecraig12345: `range` is the name used for this in [lodash](https://docs-lodash.com/v4/range/) and python, and likely elsewhere too. So I think we should keep the name.

change/@uifabric-react-hooks-2020-06-29-17-19-12-fix-example-14-update.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
ecraig12345, line 2: Remove this change file since these got added separately

packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -1,8 +1,18 @@
 import * as React from 'react';
-import { Async } from 'office-ui-fabric-react/lib/Utilities';
 import { createListItems, IExampleItem } from '@uifabric/example-data';
 import { IColumn, buildColumns, SelectionMode, Toggle } from 'office-ui-fabric-react/lib/index';
 import { ShimmeredDetailsList } from 'office-ui-fabric-react/lib/ShimmeredDetailsList';
+import { useSetInterval, useConst } from '@uifabric/react-hooks';
+
+export interface IShimmerApplicationExampleState {
```
ecraig12345, line 7: This can be internal now
```suggestion
interface IShimmerApplicationExampleState {
```

packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -31,140 +41,91 @@ const fileIcons: { name: string }[] = [
...
+
+export const ShimmerApplicationExample: React.FunctionComponent = () => {
+  const { current: state } = React.useRef<IShimmerApplicationExampleState>({
+    lastIntervalId: 0,
+    visibleCount: 0,
+  });
+
+  const [items, setItems] = React.useState<(IExampleItem | null)[] | undefined>(undefined);
+
+  const buildShimmerColumns: IColumn[] = useConst(() => {
```
ecraig12345, line 72: Just call this `shimmerColumns`

packages/office-ui-fabric-react/src/components/Shimmer/examples/Shimmer.Application.Example.tsx
```diff
@@ -31,140 +41,91 @@ const fileIcons: { name: string }[] = [
...
-      _items = createListItems(ITEMS_COUNT);
-      _items.forEach((item: IExampleItem) => {
-        const randomFileType = this._randomFileIcon();
-        item.thumbnail = randomFileType.url;
-      });
-    }
-
-    let items: IExampleItem[];
-    const randomQuantity: number = Math.floor(Math.random() * 10) + 1;
+  const onLoadData = React.useCallback((ev: React.MouseEvent<HTMLElement>, checked: boolean): void => {
```
ecraig12345, line 96: Best to do `useConstCallback` here since you don't care about dep changes. (Also, that should prevent it from getting flagged by the "rules of hooks" eslint rule I'm about to enable for not having exhaustive deps.)

packages/office-ui-fabric-react/src/components/Stack/examples/Stack.Horizontal.Configure.Example.tsx
```diff
@@ -118,270 +150,204 @@ const HorizontalStackConfigureExampleContent: React.FunctionComponent<IExampleOp
...
+          <Stack>
+            <Slider
+              label="Number of items:"
+              min={1}
+              max={30}
+              step={1}
+              defaultValue={5}
+              showValue
+              // eslint-disable-next-line react/jsx-no-bind
+              onChange={value => setNumItems(value)}
```
ecraig12345, line 184: Pretty sure that since `value` is the first argument you can just do `onChange={setNumItems}` (and same for all the other sliders in this and other examples)

packages/office-ui-fabric-react/src/components/Stack/examples/Stack.Horizontal.Configure.Example.tsx
```diff
@@ -33,6 +34,35 @@ export interface IExampleOptions {
...
+  { key: 'end', text: 'Bottom' },
+];
+
+// Tokens definition
+const sectionStackTokens: IStackTokens = { childrenGap: 10 };
+const configureStackTokens: IStackTokens = { childrenGap: 20 };
+const shadowItemCheckboxStyles: Partial<ICheckboxStyles> = { root: { marginRight: 10 } };
+const wrapItemCheckboxStyles: Partial<ICheckboxStyles> = { root: { marginBottom: 10 } };
+
+const range = (start: number, end: number): number[] => {
```
ecraig12345, line 58: Just realized, if you want to put this in a shared example-only spot instead of repeating, it could go in the example-data package.

PR #13641, opened 2020-06-17
feat: update `design` prop to support style overrides

mnajdova:
This PR is alternative to https://github.com/microsoft/fluentui/pull/13642

Here we are trying the `@emotion/serialize` for hasing the design prop

PR #13867, opened 2020-06-30
chore(design): design prop using caching by hash from emotion [47.x]

miroslavstastny:
Backport of #13641 to @fluentui/react-northstar@0.47.x


layershifter:
We decided to go with `useCSS()` hook, closing for housekeeping.

PR #13884, opened 2020-07-01
Add new codemods package

joschect:
#### Pull request checklist

#### Description of changes

Adds the basis for codemods. Work that this completes:
* Searches for codeMod files, loads and runs them
* Searches for tsconfig files, and loads them to get the files on which codemods should run
* Sets up for filtering codemods based on some criteria TBD
* Adds the majority of the import-to-compat codemod
* Sets up a framework for defining new codemods.
 
The primary files to focus on are `upgrade.ts` and `runnerUtilities`. These setup the framework for how codemods will run. Additional work still needs to be done to:

- [ ] Add logging
- [ ] Add more filtering to tests
- [ ] Break out functions into utilities.

#### Focus areas to test

(optional)


packages/codemods/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+To build the upgrade.js file
+
+Run
+
+```sh
+npm pack
+```
+
+from the codemods root to create a tar file for testing. Move the created tar file to the package you want to test and run
```
khmakoto, line 36:
```suggestion
from the codemods package's root to create a tar file for testing. Move the created tar file to the package you want to test and run
```

packages/codemods/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+```sh
+npx <tarFileName>
+```
+
+## Executing a codemod (Prototype)
+
+Make sure that fluent is installed in your package.
```
khmakoto, line 44:
```suggestion
Make sure that Fluent UI is installed in your package.
```

packages/codemods/README.md
```diff
@@ -0,0 +1,83 @@
...
+```
+npx @fluentui\codemods
+```
+
+## Todos
+
+- Write a `flag` utility that will enable devs to note when a part of a file needs to be changed, but cannot be done via codemod.
+- Implement a command that will execute all listed codemods on a single file.
+  - Will need to think of a way to specify the order. Maybe something like tasks in Just.
+  - This could be helpful when you want to run a set of codemods based on a single condition, like the presence of a specific import
```
khmakoto, line 56: Should these be tracked here? It feels like these should probably be tracked in an issue instead of in the README.

packages/codemods/src/codeMods/ICodeMod.ts
```diff
@@ -0,0 +1,28 @@
+import { SourceFile } from 'ts-morph';
```
khmakoto, line 1: Should this be a `.types.ts` file?

packages/codemods/src/codeMods/ICodeMod.ts
```diff
@@ -0,0 +1,28 @@
...
+   * Must be valid semver.
+   * Currently has no impact. Maybe remove?
+   */
+  version: string;
+  /**
+   * A string to help identify the codemod.
+   */
+  name: string;
+  /**
+   * The actual function that should be run on any given file
```
khmakoto, line 19:
```suggestion
   * The actual function that should be run on any given file.
```

packages/codemods/src/codeMods/ICodeMod.ts
```diff
@@ -0,0 +1,28 @@
...
+   * TODO, is there a possibility of codemods that would need to execute over mutiple files?
+   */
+  run: (file: T) => ICodeModResult;
+  /**
+   * If not enabled, then this mod will not be conisdered to run. Only enable it once it's ready for
+   * the real world.
```
khmakoto, line 25:
```suggestion
   * If not enabled, then the mod will not be executed. Only enable it once it's ready to be applied in
   * real world scenarios.
```

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
```diff
@@ -0,0 +1,84 @@
+import { Project } from 'ts-morph';
```
khmakoto, line 1: Is there a special reason we're using `__tests__` instead of `tests` in the path?

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
```diff
@@ -0,0 +1,84 @@
...
+import * as Root from '../mock/compat/mockIndex';
+import {
+  IRawCompat,
+  runComponentToCompat,
+  getNamedExports,
+  buildCompatHash as buildHash,
+  repathNamedImports,
+  repathPathedImports,
+} from '../../mods/ComponentToCompat/CompatHelpers';
+import * as Button from 'office-ui-fabric-react/lib-commonjs/Button';
```
khmakoto, line 11: Why are you importing from `lib-commonjs`?

joschect: This is only runs via node and common-js is the only type of import that is supported.

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
joschect: I liked keeping them separate this way, especially since a lot of them require mocks. I'm not sure if it's the right approach. For right now I think this is okay to checkin and we can rearrange if things prove too difficult in the future.

packages/codemods/src/codeMods/ICodeMod.ts
joschect: I don't think so. At least not yet. This doesn't correspond to an actual object/component so I don't think that it makes sense to have a types file. I'm open to it though.

packages/codemods/README.md
```diff
@@ -0,0 +1,83 @@
+# ts-morphin-migration
+
+ts-morphin-migration is a package created and managed with the `create-just` utility.
```
ecraig12345, line 3: Can you update this with the current package name and description? (remove the reference to create-just unless you did actually use that, and add a real description)

packages/codemods/README.md
```diff
@@ -0,0 +1,83 @@
...
+npx <tarFileName>
+```
+
+## Executing a codemod (Prototype)
+
+Make sure that fluent is installed in your package.
+Run
+
+```
+npx @fluentui\codemods
```
ecraig12345, line 48: This should probably use a forward slash?

packages/codemods/README.md
ecraig12345, line 44: You should list the specific package(s) here. I assume it works with fui/react but also add OUFR if it works with that.
```suggestion
Make sure that `@fluentui/react` is installed as a dependency of your package.
```

packages/codemods/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Test with the actual package:
+
+Run
+
+```sh
+yarn build
+```
+
+To build the upgrade.js file
```
ecraig12345, line 28: What is the upgrade.js file?

packages/codemods/README.md
```diff
@@ -0,0 +1,83 @@
...
+## Next Steps
+
+Run
+
+```sh
+yarn
+yarn start-test
+```
+
+To start testing the codemods
```
ecraig12345, line 14: In general I think the description of what the commands do should go above the commands not below

packages/codemods/jest.config.js
```diff
@@ -0,0 +1,8 @@
+const { createConfig } = require('@uifabric/build/jest/jest-resources');
+
+const config = createConfig({
+  testMatch: ['<rootDir>/src/**/__tests__/test/**/*.[jt]s?(x)', '<rootDir>/src/**/?(*.)+(spec|test).[tj]s?(x)'],
+  testRegex: undefined,
```
ecraig12345, line 5: Not sure you need to override this--all your tests appear to be named `*.test.ts`, which should be picked up by the default glob. (Or consider fixing the default glob if it's wrong.)

packages/codemods/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@fluentui/codemods",
+  "version": "0.1.0",
+  "description": "package to enable easy upgrades to new versions",
```
ecraig12345, line 4:
```suggestion
  "description": "Tool enabling easy upgrades to new versions of Fluent UI",
```

packages/codemods/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "private": true,
```
ecraig12345, line 5: Note you'll have to remove this before anyone can use the package

packages/codemods/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "license": "MIT",
+  "scripts": {
+    "build": "just-scripts build",
+    "clean": "just-scripts clean",
+    "code-style": "just-scripts code-style",
+    "just": "just-scripts",
+    "lint": "just-scripts lint",
+    "start-test": "just-scripts jest-watch",
+    "test": "just-scripts test",
+    "update-snapshots": "just-scripts jest -u"
```
ecraig12345, line 24: You can remove this one if the package doesn't use snapshots (haven't finished looking yet)

packages/codemods/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "devDependencies": {
+    "@types/jest": "~24.9.0",
+    "@types/node": "^10.3.2",
+    "@uifabric/build": "^7.0.0",
+    "@uifabric/tslint-rules": "^7.2.2"
+  },
+  "dependencies": {
+    "tslib": "^1.10.0",
+    "@types/react": "16.8.25",
```
ecraig12345, line 34: Typically types should be dev dependencies.

packages/codemods/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "@types/semver": "^5.5.0",
+    "acorn": "^6.4.1",
+    "minimist": "^1.2.2",
```
ecraig12345, line 37: Not seeing acorn or minimist used anywhere?

packages/codemods/src/codeMods/ICodeMod.ts
ecraig12345: Often in this case we just call the file `types.ts` (no component/object prefix).

packages/codemods/src/codeMods/ICodeMod.ts
```diff
@@ -0,0 +1,28 @@
+import { SourceFile } from 'ts-morph';
+
+export interface ICodeModResult {
```
ecraig12345, line 3: New packages should probably omit the I from interfaces. create-package templates ought to be setting it up that way automatically.

packages/codemods/src/codeMods/__tests__/mock/compat/Button.tsx
```diff
@@ -0,0 +1,2 @@
+export const Button = {};
```
ecraig12345, line 1: Not sure if it would work in this case, but you should consider following the automatic mocking folder structure described here. https://jestjs.io/docs/en/manual-mocks.html

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
ecraig12345: It seems like both conventions in open source, though in this repo just `tests` seems more common (including in a few cases having a central `tests` folder directly under `src`). So it might be better to omit the underscores for consistency.

packages/codemods/src/codeMods/ICodeMod.ts
```diff
@@ -0,0 +1,28 @@
...
+  success?: boolean;
+}
+
+export interface ICodeMod<T = SourceFile> {
```
ecraig12345, line 7: Depending on how T is used (haven't finished reading yet) you may need to put an `extends` constraint on it instead of just supplying a default. Also what other types would you expect it to potentially be?

packages/codemods/src/codeMods/ICodeMod.ts
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * The version before which the codemod should run.
+   * Must be valid semver.
+   * Currently has no impact. Maybe remove?
```
ecraig12345, line 11: If it has no impact you probably shouldn't have it marked as required :)

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
ecraig12345: Typically we import from `lib` anyway and handle the remapping with `moduleNameMapper` settings (which should be configured automatically after a recent change I did)

packages/codemods/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "office-ui-fabric-react": "^7.121.7",
```
ecraig12345, line 38: If this is only used in tests, it should be a dev dependency.

packages/codemods/jest.config.js
joschect: I definitely need to override this. At least since I'm not using .test because of 
image

It looks like it will run all files under __test__ OR those that have .test. Once I switch to just a test folder I can do away with this.

packages/codemods/src/codeMods/ICodeMod.ts
xugao: agree with @ecraig12345 here. also we should remove `I` prefix on interfaces.

packages/codemods/README.md
```diff
@@ -0,0 +1,83 @@
+# ts-morphin-migration
+
+ts-morphin-migration is a package created and managed with the `create-just` utility.
+
+## Next Steps
```
dzearing, line 5: It would be great to start the README.md to explain what the package is for and how it should be used, rather than how to test it.

packages/codemods/README.md
dzearing: Even then, I'm confused. What does this mean? In your "application"? What "Fluent UI" package? Why is this package separate from other packages?

E.g. if `react-button` wants to ship a mod with the package, can the mod live with the package? What pattern are we suggesting to use for partners (like react-date-time) that want to write codemods? If they are separate from the package they apply to, how are they locked to the package they apply to (how does versioning work?)

packages/codemods/jest.config.js
ecraig12345: Hmm okay. Can you use just `testRegex` though? And with a glob that only reflects your specific tests.

packages/codemods/src/codeMods/mods/PersonaToAvatar/PersonaToAvatar.mod.ts
```diff
@@ -0,0 +1,179 @@
+import {
```
dzearing, line 1: I don't think this kind of file should exist. If we wanted to rename 'Foo' to 'Bar' in the future, I would not want a dev copy pasting this file. 

I think we want to have something more like:

codemods.config.js:
```
export default const config = {
  actions: [
    {
       type: 'rename',
       from: {
         importPaths: [ 'office-ui-fabric-react', 'office-ui-fabric-react/lib/Persona', @fluentui/react', '@fluentui/react/lib' ],
         name: 'PersonaCoin',
       },
       to: {
         path: '@fluentui/react',
         name: 'Avatar'
       }
    }
  ]
};
```

Or if there is a way to make it even simpler, awesome. But we need to keep it the least error prone system for doing simple API surface updates. That way we can easily review codemods and have full confidence they will work without some simple mistake.

packages/codemods/src/codeMods/mods/PersonaToAvatar/PersonaToAvatar.mod.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+const PersonaToAvatarMod: ICodeMod = {
+  run: (file: SourceFile) => {
+    try {
+      replacePersonaImport(file);
+      replaceIPersonaPropsImport(file);
+      replacePersonaSizeImport(file);
+      renamePrimaryTextProp(file);
+      renameRenderCoin(file);
```
dzearing, line 168: These right here; in my suggestion above, could these be just replacement types in a config?

Some of them probably not. E.g. `renameRenderCoin` finds `Persona` jsx elements and renames `onRenderCoin` to `onRenderAvatarCoin` (which I think is wrong btw.)

dzearing: So, you may need a custom `type` which does some sort of manual work programatically. But whatever we can do to simplify and keep the codemod data driven is good.

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  it('correctly repaths from exact known import', () => {
+    const file = project.getSourceFileOrThrow('ImportsStuff.tsx');
+    const hash = buildHash([{ componentName: 'Button', namedExports: Root }], createComponentToCompat);
+    repathPathedImports(file, hash);
+    expect(file.getFullText()).toContain('compat/Button');
+  });
+
+  it('correctly gets maps imports from an external library', () => {
```
ecraig12345, line 35: Not sure what this means/does?

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    const buttonExports = getNamedExports(Button);
+    expect(buttonExports).toContain('DefaultButton');
+    expect(buttonExports).toContain('CommandButton');
+  });
+
+  it('correctly repaths from index', () => {
+    const file = project.getSourceFileOrThrow('ImportsStuff.tsx');
```
ecraig12345, line 42: I'm having trouble understanding what some of the rest of these do too, since you're using fake files which don't actually reference OUFR.

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  it('correctly removes index if no imports left', () => {
+    const file = project.getSourceFileOrThrow('ImportsStuff.tsx');
+    const hash = buildHash([{ componentName: 'Button', namedExports: Root }], createComponentToCompat);
+    repathNamedImports(file, hash, './DefaultButton');
+    expect(file.getFullText()).not.toContain('./DefaultButton');
+  });
+
+  it('correctly builds a hash map', () => {
```
ecraig12345, line 78: This should probably come first since it's relied on by other tests.

packages/codemods/README.md
joschect: Right now, for v1, this will run all codemods present in this package. There is no reason we can't change where the execution file looks to get codemods but that can be a feature down the line.

joschect: Right now the codemods don't have a concept of package or root. That's probably a good thing to add as a filter or next steps.

packages/codemods/src/codeMods/ICodeMod.ts
joschect: This is more of an escape hatch right now. We're expecting it to basically always be of type SourceFile, but if you wanted to test, say a codemod which just takes a string, you can do that.

packages/codemods/src/codeMods/__tests__/personaToAvatar/componentMod.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      imp.getNamedImports().forEach(name => {
+        expect(name.getText()).not.toEqual('IPersonaProps');
+      });
+      file.forEachDescendant(val => {
+        switch (val.getKind()) {
+          case SyntaxKind.InterfaceDeclaration: {
+            const tVal = val as InterfaceDeclaration;
+            const struct = tVal.getStructure();
+            expect(
+              (struct.extends as string[])?.some(str => {
```
ecraig12345, line 42: What happens if `struct.extends` isn't defined?

packages/codemods/src/codeMods/mods/PersonaToAvatar/PersonaToAvatar.mod.ts
joschect: This is more of a proof of concept for right now that I should make more clear. 

As for the config, we could potentially allow some of these to take in a json blob. For the start we should get the api of the functions correct. Trip and I are working on that.

packages/codemods/src/codeMods/__tests__/personaToAvatar/componentMod.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      });
+    });
+  });
+
+  it('can replace PersonaSize with AvatarSize', () => {
+    const file = project.getSourceFileOrThrow('mWithPersonaSize.tsx');
+    replacePersonaSizeImport(file);
+
+    file.forEachDescendant(val => {
+      // I believe that these are really the only two cases that a reference to PersonaSize
```
ecraig12345, line 61: In theory you could do `PersonaSize[something]` (not sure if that's caught by PropertyAccessExpression) or `Object.keys(PersonaSize)`. Also seems like TS or ts-morph ought to have a built-in utility for finding all references to a type within a file.

packages/codemods/src/codeMods/__tests__/personaToAvatar/propsMod.test.ts
```diff
@@ -0,0 +1,71 @@
...
+// @TODO ensure that props are not renamed for non fabric personas if they exist
+
+const personaPropsFile = 'mPersonaProps.tsx';
+const personaSpreadPropsFile = 'mPersonaSpreadProps.tsx';
+
+describe('Persona props mod tests', () => {
+  let project: Project;
+
+  beforeEach(() => {
+    project = new Project();
```
ecraig12345, line 14: Do you actually need a new project each time? (is it mutating anything?)

packages/codemods/src/codeMods/__tests__/personaToAvatar/componentMod.test.ts
```diff
@@ -0,0 +1,77 @@
+import { Project, SyntaxKind, InterfaceDeclaration, TypeReferenceNode, PropertyAccessExpression } from 'ts-morph';
+import {
+  replacePersonaImport,
+  replaceIPersonaPropsImport,
+  replacePersonaSizeImport,
+} from '../../mods/PersonaToAvatar/PersonaToAvatar.mod';
+import { findJsxTag } from '../../utilities';
+const personaPath = '/**/__tests__/mock/**/persona/**/*.tsx';
+
+describe('Persona component mod tests', () => {
```
ecraig12345, line 10: nit, you don't need "tests"/similar in the describe name

packages/codemods/src/codeMods/__tests__/personaToAvatar/propsMod.test.ts
joschect: Unfortunately, yes. It mutates the virtual files and there doesn't appear to be a way to revert them.

packages/codemods/src/codeMods/__tests__/utilities/importUtils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const file = project.getSourceFileOrThrow(fileName);
+    const imp = getImportsByPath(file, rootPath);
+    expect(imp.length).toEqual(0);
+  });
+
+  it('can find a single import based on a regex string match', () => {
+    const file = project.getSourceFileOrThrow(fileName);
+    const imp = getImportsByPath(file, /office\-ui\-fabric\-react.+Button/);
+    expect(imp).toBeTruthy();
+    expect(imp?.length).toEqual(1);
```
ecraig12345, line 33: Probably ! is more appropriate in cases like this

packages/codemods/src/codeMods/__tests__/utilities/importUtils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  it('only finds exact matches', () => {
+    const file = project.getSourceFileOrThrow(fileName);
+    const imp = getImportsByPath(file, rootPath);
+    expect(imp.length).toEqual(0);
+  });
+
+  it('can find a single import based on a regex string match', () => {
+    const file = project.getSourceFileOrThrow(fileName);
+    const imp = getImportsByPath(file, /office\-ui\-fabric\-react.+Button/);
```
ecraig12345, line 31: I don't think the - absolutely has to be escaped in a regex, outside of a character class?

packages/codemods/src/codeMods/mods/ComponentToCompat/ComponentToCompat.mod.ts
```diff
@@ -0,0 +1,56 @@
...
+  };
+}
+
+const ComponentToCompat: ICodeMod = {
+  run: (file: SourceFile) => {
+    runComponentToCompat(file, buildCompatHash(exportMapping, createComponentToCompat), fabricindex);
+    return { success: true };
+  },
+  name: 'ComponentToCompat',
+  version: '1.0.0',
```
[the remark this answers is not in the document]
joschect: Good question, it's the version that the codemod should run against. I wonder if it would be more appropriate to have a filter callback.

packages/codemods/src/codeMods/mods/PersonaToAvatar/PersonaToAvatar.mod.ts
joschect: @dzearing I agree, we won't have a file like this. This was a proof of concept mainly. I want to get the utility functions nailed down. I'll get this refactored for an example in the future.

packages/codemods/src/codeMods/__tests__/utilities/importUtils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+const fileName = 'mockImports.tsx';
+const rootPath = 'office-ui-fabric-react';
+const buttonPath = `${rootPath}/lib/Button`;
+
+describe('Import Utilities test', () => {
+  let project: Project;
+
+  beforeEach(() => {
+    project = new Project();
+    project.addSourceFilesAtPaths(`${process.cwd()}/**/__tests__/mock/utils/*.tsx`);
```
xugao, line 13: we should remove these files after test?

packages/codemods/README.md
xugao, line 48: just want to confirm these will be supported:
1. User can pick a specific codemod to run, e.g: repath to compat path
2. User can target a specific component in case of repathing

packages/codemods/package.json
xugao, line 5: Just FYI - this means the package won't be automatically published when you merge this PR.

At one point, we need to test these codemods in an app codebase such as OOUI

packages/codemods/src/codeMods/mods/ComponentToCompat/CompatHelpers.ts
```diff
@@ -0,0 +1,79 @@
...
+      return acc;
+    },
+    { namedExportsMatch: {}, exactPathMatch: {} },
+  );
+}
+
+export function repathPathedImports(file: SourceFile, hash: ICompatHash) {
+  file.getImportDeclarations().forEach(dec => {
+    const str = dec.getModuleSpecifierValue();
+    if (hash.exactPathMatch[str]) {
```
xugao, line 70: if I'm reading the code right, i don't think this works for the following scenario?

Before:
```
import { DefaultButton, DetailsList } from 'office-ui-fabric-react'
```

After:
```
import { DetailsList } from 'office-ui-fabric-react';
import { DefaultButton } from 'office-ui-fabric-react/lib/compat'`
```

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  beforeEach(() => {
+    project = new Project();
+    project.addSourceFilesAtPaths(`${process.cwd()}/**/__tests__/mock/compat/*.tsx`);
+  });
+
+  it('correctly repaths from exact known import', () => {
+    const file = project.getSourceFileOrThrow('ImportsStuff.tsx');
+    const hash = buildHash([{ componentName: 'Button', namedExports: Root }], createComponentToCompat);
+    repathPathedImports(file, hash);
+    expect(file.getFullText()).toContain('compat/Button');
```
xugao, line 32: this is not a robust enough check. meaning we could been doing wrong things can still have the test passing.

I think we should take a snapshot for `file.getFullText()` instead

xugao: similar for other tests

packages/codemods/src/codeMods/__tests__/utilities/importUtils.test.ts
ecraig12345: "Add source files" doesn't mean actually creating files, just adding references to pre-existing mock files.

packages/codemods/README.md
joschect: I don't believe that the goal is to support this for V1. We can talk more about it, however.

packages/codemods/src/codeMods/mods/ComponentToCompat/CompatHelpers.ts
joschect: Correct! This doesn't, but that's why there is the function `repathNamedImports`. Realistically They should be named better but they do two fundamentally different things. One goes and just checks for an exact match and then replaces that with the new path. That's the one referenced here, `repathPathedImports`. 

The other one finds a specific root path (like office-ui-fabric-react) and checks all of the named imports it contains and then moves those. There's a test for this one too :D

packages/codemods/src/codeMods/utilities/imports.ts
```diff
@@ -0,0 +1,74 @@
+import { SourceFile, ImportDeclaration, ImportSpecifierStructure } from 'ts-morph';
+
+export function renameImport(file: SourceFile, originalImport: string, renamedImport: string) {
+  const imps = file.getImportDeclarations().filter(cond => {
+    return cond.getNamedImports().some(val => {
+      return val.getText() === originalImport;
+    });
+  });
+  imps[0].getNamedImports().forEach(name => {
```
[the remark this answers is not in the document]
joschect: Potentially. But unlikely. It could happen if you imported from the same path twice. I can make that more flexible.

packages/codemods/src/command.ts
```diff
@@ -0,0 +1,18 @@
+export class CommandParserResult {
+  // path: string = ""; TODO this doesn't work with jest; why?
+  constructor(public shouldExit = false, public path = '') {}
+}
+
+export class CommandParser {
```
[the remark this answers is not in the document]
joschect: If we need a more robust command parser I can certainly add one, but for now this is nice little one to just help out.

packages/codemods/src/modRunner/__tests__/modRunner.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        run: runcallBack,
+        version: '1.1.1',
+      },
+      {
+        name: 'c',
+        run: runcallBack,
+        version: '1.1.1',
+      },
+    ];
+    runMods(mods, [''], () => void 0);
```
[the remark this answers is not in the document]
joschect: This is the purpose of making the ICodeMods generic. They can take anything into their run as an argument and it will work fine.

packages/codemods/src/upgrade.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  console.log('getting configs');
+  const configs = getTsConfigs();
+  const projects: Project[] = configs.map(configString => new Project({ tsConfigFilePath: configString }));
+
+  projects.forEach(project => {
+    let error = false;
+    try {
+      const files = project.getSourceFiles();
+      runMods(mods, files, result => {
```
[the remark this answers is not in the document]
joschect: But runMods runs on each file/mod so the return value would be 1 giant log array. Unless I'm misunderstanding.

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
joschect: I'll iterate through the imports and check it directly instead of using a snapshot.

packages/codemods/src/codeMods/mods/ComponentToCompat/CompatHelpers.ts
xugao: Ic now. Looks to me no test is covering this case explicitly. That is to make sure `import { DetailsList } from 'office-ui-fabric-react';` reminds untouched. If i missed it, could you point me to it? If not, could you add the test coverage?

packages/codemods/src/upgrade.ts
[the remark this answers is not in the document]
ecraig12345: Oh okay, actually I was misunderstanding 😆

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
xugao: i'd suggest at least have one test that takes a full snapshot of the file. perhaps a test for `runComponentToCompat`

packages/codemods/README.md
joschect: I'm going to keep them here for now, since this is a prototype/alpha I think it's fine to have more stuff here. It will be removed before it's published.

packages/codemods/package.json
joschect: I'll make it public.

packages/codemods/README.md
xugao: I left this as an Ask in hackmd. we can discuss further with the team. regardless, this is something outside of the scope of this PR

packages/codemods/package.json
joschect: It is used in real code, at least for now.

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
joschect: I'm going to leave it as commonjs for now, I can test removing that later once this has been checked in.

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
joschect: It's making sure that the utilities/each individual piece does the job that it's supposed to do. That way when they all get put together it will work as expected.

packages/codemods/src/codeMods/__tests__/mock/compat/Button.tsx
joschect: It probably won't since we just care the files exist to be parsed into typescript asts.

packages/codemods/src/codeMods/__tests__/personaToAvatar/componentMod.test.ts
joschect: It has to be defined, at least that's a guarantee when you call getStructure()

packages/codemods/src/codeMods/__tests__/personaToAvatar/componentMod.test.ts
joschect: Yes, you can call getReferences or something like that. But this is just the test, I believe that in the actual prototype mod that we use the rename function to rename all the PersonaSizes within the file without iterating through like that.

packages/codemods/src/codeMods/mods/ComponentToCompat/CompatHelpers.ts
```diff
@@ -0,0 +1,79 @@
+import { SourceFile } from 'ts-morph';
+import { getImportsByPath, appendOrCreateNamedImport, repathImport } from '../../utilities/index';
+
+export interface IComponentToCompat {
+  // Old exact path
```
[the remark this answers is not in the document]
joschect: I'll do another pass later once all of this has stabilized.

packages/codemods/src/modRunner/__tests__/mocks/MockProject/projects/subProject/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "commonjs",
+    "jsx": "react",
+    "declaration": true,
+    "outDir": "./lib",
+    "strict": true,
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "typeRoots": ["node_modules/@types", "node_modules/just-stack-single-lib/node_modules/@types"]
```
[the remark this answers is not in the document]
joschect: Its just a brainless copy of a tsconfig. It's used to validate that we can find them.

packages/codemods/src/modRunner/__tests__/modRunner.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  it('finds the right ts-config files', () => {
+    const files = getTsConfigs(`${process.cwd()}/src/modRunner/__tests__/mocks/MockProject`);
+
+    expect(files.length).toEqual(2);
+    expect(files[0]).toContain('tsconfig.json');
+  });
+
+  it('gets a mod when given a path', () => {
```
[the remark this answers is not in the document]
joschect: I agree.

packages/codemods/src/command.ts
```diff
@@ -0,0 +1,18 @@
...
+  constructor(private logger = console.log) {}
+
+  public parseArgs(argv: string[]) {
+    if (argv.slice(2).indexOf('--help') >= 0) {
```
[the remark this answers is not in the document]
joschect: ITT this is okay for now.

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
ecraig12345: Part of what I meant by that was that it would be helpful to have more comments explaining what each one does. And I'm still having trouble how it works with fake files which don't actually reference OUFR.

packages/codemods/src/codeMods/mods/ComponentToCompat/CompatHelpers.ts
joschect: I had missed that test! Just added it.

joschect: I had ones that were close to it, but none that explicitly checked that it left some imports untouched.

packages/codemods/src/codeMods/__tests__/ComponentToCompat/ComponentToCompat.test.ts
joschect: I'd prefer to not add snapshots just yet. It might be a good idea in the future but for now I the checks to be explicit.

packages/codemods/src/codeMods/__tests__/personaToAvatar/componentMod.test.ts
ecraig12345: In that case can you use `!` instead of `?`

packages/codemods/src/codeMods/mods/ComponentToCompat/CompatHelpers.ts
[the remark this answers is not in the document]
ecraig12345: My concern (not with you personally but in general) is that the second pass will be forgotten and then we just have undocumented code. Lack of comments also makes it harder for us who are trying to review things in the meantime. Or if you don't want to add full docs for things that are likely to change, it would be helpful to at least add a comment at the top of the file saying this is subject to change and don't review too closely (though I'm still a bit concerned about the cleanup/doc pass being forgotten).

packages/codemods/src/modRunner/__tests__/mocks/MockProject/projects/subProject/tsconfig.json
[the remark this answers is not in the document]
ecraig12345: In that case it's probably better to use a very minimal valid tsconfig, with a comment that it's just for a test of being able to find them (tsconfigs can have comments)

packages/codemods/README.md
```diff
@@ -1,6 +1,20 @@
-# ts-morphin-migration
+# @fluentui/codemods
 
-ts-morphin-migration is a package created and managed with the `create-just` utility.
+This is a utility package to assist with the upgrading of libraries and apps that rely on fluentui.
+
+## How it works
+
+If you have a typescript application or library that relies on fluentui then you can run `npx @fluentui/codemods` which will immediately begin an upgrade of your codebase so that you don't have to! It works by finding all the tsconfig files and then using those to find the relevant files to upgrade before running the updates on each of them!
```
ecraig12345, line 7: I still think it would be helpful to specify the particular package names it's capable of upgrading (in particular that OUFR is supported too)

packages/codemods/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "@fluentui/codemods",
   "version": "0.1.0",
-  "description": "package to enable easy upgrades to new versions",
-  "private": true,
+  "description": "Tool enabling easy upgrades to new versions",
+  "private": false,
```
ecraig12345, line 5: Shouldn't this specifically say it's for Fluent UI, unless you plan to make it more general? Also we typically just remove the `private` setting.
```suggestion
  "description": "Tool enabling easy upgrades to new Fluent UI versions",
```

packages/codemods/src/codeMods/types.ts
```diff
@@ -1,28 +1,30 @@
 import { SourceFile } from 'ts-morph';
 
-export interface ICodeModResult {
+// tslint:disable-next-line: interface-name
```
ecraig12345, line 3: Instead of disabling per line you can modify the package's tslint.json
```
{
  "extends": ["@uifabric/tslint-rules"],
  "rules": {
  }
}
```

packages/codemods/README.md
joschect: Currently nothing is explicitly supported, we're still figuring out how all of that will work.

packages/codemods/src/codeMods/types.ts
joschect: I'll plan do that once this has been checked in.

xugao: why postpone to next checkin? it's more suitable to do it now/in this pr

packages/codemods/src/codeMods/types.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface CodeModResult {
+  success?: boolean;
+}
+
+// tslint:disable-next-line: interface-name
+export interface Codemod<T = SourceFile> {
+  /**
+   * The version before which the codemod should run.
+   * Must be valid semver.
+   * Currently has no impact. Maybe remove?
```
xugao, line 13:
```suggestion
   * Each type of codemod can have multiple versions which work on different versions of its targeted package. 
   * Must be valid semver.
   * TODO: Currently version is not being checked. Implement this fully or remove.
```

packages/codemods/src/codeMods/mods/PersonaToAvatar/PersonaToAvatar.mod.ts
```diff
@@ -0,0 +1,181 @@
...
+    } catch (e) {
+      return { success: false };
+    }
+    return { success: true };
+  },
+  version: '100000',
```
xugao:
```suggestion
  version: '1.0.0',
```

change/@fluentui-codemods-2020-07-07-17-46-16-codemods.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Initial checkin for codemods",
```
khmakoto, line 3:
```suggestion
  "comment": "Codemods: Creating package and initial check-in.",
```

packages/codemods/README.md
```diff
@@ -0,0 +1,88 @@
+# @fluentui/codemods
+
+This is a utility package to assist with the upgrading of libraries and apps that rely on fluentui.
+
+## How it works
+
+If you have a typescript application or library that relies on fluentui then you can run `npx @fluentui/codemods` which will immediately begin an upgrade of your codebase so that you don't have to! It works by finding all the tsconfig files and then using those to find the relevant files to upgrade before running the updates on each of them!
```
khmakoto, line 7:
```suggestion
If you have a typescript application or library that relies on a non-current version of Fluent UI then you can run `npx @fluentui/codemods` to immediately begin an upgrade of your codebase, saving you the trouble of doing so manually! This works by finding all the tsconfig files and then using those to find the relevant files to upgrade before running the updates on each of them!
```

packages/codemods/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Executing a codemod (Prototype)
+
+If your application relies on any package in fluentui simply run
```
khmakoto, line 11:
```suggestion
If your application relies on any Fluent UI package simply run
```

packages/codemods/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+```
+npx @fluentui/codemods
+```
+
+and the upgrade will begin if there are any relevant codemods!
```
khmakoto, line 17:
```suggestion
and the upgrade will begin if there are any relevant codemods to apply to your codebase!
```

packages/codemods/README.md
```diff
@@ -0,0 +1,88 @@
...
+  - This could be helpful when you want to run a set of codemods based on a single condition, like the presence of a specific import
+
+## Notes
+
+ts-morph:
+
+- ts-morph does most of the heavy lifting. Don't be afraid to use it directly rather than trying to abstract into a utility.
+- One of the most useful types of the syntax tree to get is the `SyntaxKind.block` it is the equivalent of the `{ stuff }` that is located in a function declaration and is where a lot of code lives.
+- You can only access JSX props on syntax kinds of `SyntaxKind.JSXOpeningElement` and `SyntaxKind.JSXSelfClosingElement`
+- `getChildIndex` returns the child index respective to the immediate parent. It resets each level. So consider the following:
```
khmakoto, line 70:
```suggestion
- `getChildIndex` returns the child index respective to the immediate parent. It resets at each level. So consider the following:
```

packages/codemods/package.json
xugao: @joschect - pls fix this

packages/codemods/src/codeMods/types.ts
xugao: @joschect - pls fix this

packages/codemods/src/codeMods/types.ts
xugao: @joschect - we should fix this in this round

packages/codemods/src/codeMods/types.ts
```diff
@@ -0,0 +1,30 @@
+import { SourceFile } from 'ts-morph';
+
+// tslint:disable-next-line: interface-name
```
xugao, line 3: can remove , other places too

PR #13886, opened 2020-07-01
Toggle: remove double label for

### LINKED ISSUE #12708: There are two "<Label htmlFor" in Toggle component
### Environment Information

- **Package version(s)**: 7.94.0

### Describe the issue:

Using Toggle component and open developer tool to inspect component. Notice that there are two "Lable for="Toggle1" in the Toggle component. It’s incorrect from HTML perspective and accessibility. 

Here is a link for reference, https://www.w3.org/TR/html4/interact/forms.html#edef-LABEL.

"The LABEL element may be used to attach information to controls. Each LABEL element is associated with exactly one form control." Is what stated in the link.

Some browsers in fact override label when multiple LABEL FOR is added. So some screenreaders in a different browser might just read On/Off instead of first label.

https://codepen.io/chihch/pen/bGVERpw

#### Actual behavior:

html renders two "Label for="Toggle1"" within the same Toggle component.

#### Expected behavior:

Only one "Label for" associate with the same toggle can be rendered

tmaster628:
Fixes #12708 

#### Description of changes

Removed mutual exclusivity of LABELID and STATETEXTID in LABELLEDBYID in the toggle component.

Previous behavior: If both LABEL and STATETEXT were present, LABELLEDBYID would only store LABELID, causing the STATETEXT to be ignored by the screen reader.

Current behavior: LABELID and STATETEXTID are no longer mutually exclusive; a screen reader will now read the extant properties, no longer omitting STATETEXT. 

Behavioral note: Now that the screen reader will read both LABEL and STATETEXT if present, the narrator (on Edge at least) will read LABEL and STATETEXT as well as "ON" or "OFF" depending on the checked state of the toggle.

Note: The initial fix pointed out that there are two "Label forhtml" in toggle, but they are both required so that clicking on either will toggle a state change. Discussed further in #8597

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Toggle/Toggle.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Toggle/Toggle.base.tsx
@@ -109,7 +109,7 @@ export class ToggleBase extends React.Component<IToggleProps, IToggleState> impl
         labelledById = labelId;
       }
       if (stateText) {
-        labelledById = labelledById ? (labelledById += ' ' + stateTextId) : stateTextId;
+        labelledById = labelledById ? labelledById + ' ' + stateTextId : stateTextId;
       }
     }
 
--- a/packages/office-ui-fabric-react/src/components/Toggle/Toggle.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/Toggle/Toggle.base.tsx
@@ -146,8 +146,8 @@ export class ToggleBase extends React.Component<IToggleProps, IToggleState> impl
         {(keytipAttributes: any): JSX.Element => renderPill(keytipAttributes)}
       </KeytipData>
     ) : (
-      renderPill()
-    );
+        renderPill()
+      );
 
     return (
       <RootType className={classNames.root} hidden={(toggleNativeProps as any).hidden}>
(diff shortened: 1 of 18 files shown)


xugao:
@tmaster628 - for future reference, (when relevant), start PR title with format: `{ComponentName}: ` . that way our bot will auto label `Component: Toggle` on your PR

packages/office-ui-fabric-react/src/components/Toggle/Toggle.base.tsx
```diff
@@ -101,14 +101,15 @@ export class ToggleBase extends React.Component<IToggleProps, IToggleState> impl
...
+
     let labelledById: string | undefined = undefined;
     if (!ariaLabel && !badAriaLabel) {
       if (label) {
         labelledById = labelId;
-      } else if (stateText) {
-        labelledById = stateTextId;
+      }
+      if (stateText) {
+        labelledById = labelledById ? labelledById + ' ' + stateTextId : stateTextId;
```
xugao, line 112:
```suggestion
        labelledById = labelledById ? `${labelledById} ${stateTextId}` : stateTextId;
```

xugao: minor nit^

packages/office-ui-fabric-react/src/components/Toggle/Toggle.base.tsx
```diff
@@ -101,14 +101,15 @@ export class ToggleBase extends React.Component<IToggleProps, IToggleState> impl
...
+    // 3. label AND stateText, if existent
+
     let labelledById: string | undefined = undefined;
     if (!ariaLabel && !badAriaLabel) {
       if (label) {
         labelledById = labelId;
-      } else if (stateText) {
-        labelledById = stateTextId;
+      }
+      if (stateText) {
```
xugao, line 111: could you pls make the same fix in `useToggle` in `react-next` package? thanks!!

PR #13891, opened 2020-07-01
Migrate styled utility function to use functional component

MLoughry:
#### Pull request checklist

#### Description of changes

Oddly, `styled` was typed to return a function component, despite the fact that it really returned a class component. This change refactors the code to return a function component using hooks, and also allow for any refs to be forwarded to the underlying base component.

Additionally, as a fix for seeming regressions in the `styled` tests, I updated `safeCreate()` to wrap the `create()` call in [`act()`](https://reactjs.org/docs/test-renderer.html#testrendereract). This is needed for any underlying component hooks to fire properly.

(The impetus for this work was the inability to forward refs across `styled` boundaries)

#### Focus areas to test

(optional)


packages/utilities/src/styled.tsx
```diff
@@ -59,79 +64,85 @@ export function styled<
...
 
   const { scope, fields = DefaultFields } = customizable;
-  const ParentComponent = pure ? React.PureComponent : React.Component;
 
-  class Wrapped extends ParentComponent<TComponentProps, {}> {
-    // Function.prototype.name is an ES6 feature, so the cast to any is required until we're
-    // able to drop IE 11 support and compile with ES6 libs
-    // tslint:disable-next-line:no-any
-    public static displayName = `Styled${Component.displayName || (Component as any).name}`;
+  const Wrapped = React.forwardRef((props: TComponentProps, forwardedRef: React.Ref<TRef>) => {
```
xugao, line 96: i dont think fowardRef is needed

packages/utilities/src/styled.tsx
```diff
@@ -30,6 +30,11 @@ export type StyleFunction<TStyleProps, TStyleSet> = IStyleFunctionOrObject<TStyl
   __noStyleOverride__: boolean;
 };
 
+function useForceUpdate() {
```
xugao, line 33: we have this hook already in `react-hooks` package

packages/utilities/src/styled.tsx
MLoughry: The entire impetus for this change was the fact that I *do* need to forward a ref across a `styled` boundary.

packages/utilities/src/styled.tsx
MLoughry: Yes, but `@uifabric/react-hooks` depends on `@uifabric/utilities`, so I can't create a circular dependency.

packages/utilities/src/styled.tsx
xugao: icic

packages/utilities/src/styled.tsx
xugao: ah that sucks

MLoughry:
@dzearing @xugao There were some knock-on effects in tests that made assumptions about the class-based nature of styled components that I've fixed. As part of one of them, I updated the `Fabric` component to pass a ref.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -63,7 +63,7 @@ export const defaultTests: TestObject = {
         displayName = displayName.replace(/^Styled/, '');
       }
 
-      expect(displayName).toEqual(testInfo.displayName);
+      expect(displayName).toMatch(new RegExp(`^(Customized|Styled)?${testInfo.displayName}(Base)?$`));
```
xugao, line 66: instead of changing this. I think we can just replace 
```
const constructorName = Component.prototype?.constructor.name;
if (constructorName === 'Wrapped') {
    displayName = displayName.replace(/^Styled/, '');
}
```
with 
```
displayName = displayName.replace(/^Styled/, '');
```
?

cc @sopranopillow

MLoughry: It’s functionally the same either way. I feel it’s cleaner with the RegExp.

If you feel strongly enough, I can change it. Otherwise, I’ll merge tomorrow.

xugao: I don't mind to use regex. if so, we should remove the logic above. not need to have both
```
      const constructorName = Component.prototype?.constructor.name;
      let displayName = Component.displayName || constructorName;

      // This check is needed in case the Component is wrapped with the v7 styled() helper, which returns a wrapper
      // component with constructor name Wrapped, and adds a Styled prefix to the displayName. Components passed to
      // styled() typically have Base in their name, so remove that too.
      displayName = displayName.replace(/Base$/, '');
      displayName = displayName.replace(/^Customized/, '');
      if (constructorName === 'Wrapped') {
        displayName = displayName.replace(/^Styled/, '');
      }
```

MLoughry: Whoops! Fixed. 😝

PR #13896, opened 2020-07-02
Line chart: Expose chart margins prop for line chart

RajeshGoriga:
#### Pull request checklist

#### Description of changes

Line chart:Expose chart margins prop for line chart
This will fix if we have larger digit numbers having in y axis we can pass margins according our requirements
Add also fix hover should appear while hover on lines 

#### Focus areas to test

Line chart hover card and chart margins


change/@uifabric-charting-2020-07-02-14-22-06-user-v-gorraj-Task_4265281.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
Raghurk, line 2: >patch [](start = 11, length = 5)

should this be patch? You are adding a new props.

RajeshGoriga: I have changed this to minior as per fabric guidelines

PR #13901, opened 2020-07-02
fix(components): Fix input and menu styles for IE11

assuncaocharles:
#### Pull request checklist

#### Description of changes

[Deployed Site](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/13901/merge/react-northstar)

This PR fixes styles for `Input` and `Menu` in IE11

`Input` before:

Screenshot 2020-07-07 at 09 28 44

after: 

Screenshot 2020-07-07 at 10 12 04

`Menu` before:

Screenshot 2020-07-07 at 09 29 48

after:

Screenshot 2020-07-07 at 10 11 26

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Menu/Visual/MenuExampleCompose.shorthand.tsx
+++ b/packages/fluentui/docs/src/examples/components/Menu/Visual/MenuExampleCompose.shorthand.tsx
@@ -107,6 +107,7 @@ const themeOverrides = {
     MenuItemBlue: {
       root: {
         backgroundColor: 'lightblue',
+        height: '100%',
       },
     },
     MenuItemIconGreen: {
@@ -122,8 +123,6 @@ const themeOverrides = {
     MenuitemIndicatorSaturated: {
       root: {
         filter: 'saturate(8)',
-        height: '20px',
-        width: '20px',
       },
     },
     MenuItemWrapperDashed: {
--- a/packages/fluentui/docs/src/examples/components/Menu/Visual/MenuExampleCompose.shorthand.tsx
+++ b/packages/fluentui/docs/src/examples/components/Menu/Visual/MenuExampleCompose.shorthand.tsx
@@ -122,6 +122,7 @@ const themeOverrides = {
     MenuitemIndicatorSaturated: {
       root: {
         filter: 'saturate(8)',
+        backgroundSize: '20px',
       },
     },
     MenuItemWrapperDashed: {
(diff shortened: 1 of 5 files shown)


packages/fluentui/react-northstar/src/themes/teams/components/Input/inputStyles.ts
```diff
@@ -58,6 +58,10 @@ const inputStyles: ComponentSlotStylesPrepared<InputStylesProps, InputVariables>
     ...(p.hasIcon && {
       padding: p.iconPosition === 'start' ? v.inputPaddingWithIconAtStart : v.inputPaddingWithIconAtEnd,
     }),
+
+    '::-ms-clear ': {
```
mnajdova, line 62:
```suggestion
    '::-ms-clear': {
```

packages/fluentui/react-northstar/src/themes/teams/components/Menu/menuItemIndicatorStyles.ts
```diff
@@ -26,8 +26,9 @@ const menuItemIndicatorStyles: ComponentSlotStylesPrepared<MenuItemIndicatorStyl
       content: '" "',
       display: 'block',
       overflow: 'hidden',
-      height: pxToRem(16),
+      height: pxToRem(14),
```
assuncaocharles, line 29: @layershifter @mnajdova Before the element being float wasn't affecting the height of the container, now I removed the float and reduced the height to avoid visual regression, this affects only the `ComposeExample` but I think we are ok with it because it was overriding the height/width in the example

packages/fluentui/docs/src/examples/components/Menu/Visual/MenuExampleCompose.shorthand.tsx
```diff
@@ -122,8 +124,10 @@ const themeOverrides = {
     MenuitemIndicatorSaturated: {
       root: {
         filter: 'saturate(8)',
+        backgroundSize: '20px',
         height: '20px',
         width: '20px',
+        float: 'right' as FloatProperty,
```
assuncaocharles, line 130: @mnajdova @layershifter I am here only making it was before the changes in the menuItemStyles so we avoid screener regressions.

packages/fluentui/react-northstar/src/themes/teams/components/Menu/menuItemIndicatorStyles.ts
mnajdova: A bit skeptical about this change, as all icons are 16x16, so it looks awkward why this should require 14 for avoiding visual regressions. Why I am mentioning this is, if the clients override this to use some different icon, we may end up with clipped icon

assuncaocharles: We could leave it has it was and accept the menu item will have 2 px more of height. What do you think?

PR #13903, opened 2020-07-02
chore: remove default exports [P1]

layershifter:
#### Description of changes

This PR is a first round of `default` exports removal, see https://github.com/microsoft/fluentui/issues/12953#issuecomment-633970215.


packages/fluentui/react-bindings/src/index.ts
```diff
@@ -11,25 +11,24 @@ export { default as FocusZone } from './FocusZone/FocusZone';
 export * from './FocusZone/FocusZone.types';
```
miroslavstastny, line 11: There are still default exports in `react-bindings`:
image

layershifter: Yep, they will be handled separately as they require changes in `getComponentInfo()`: https://github.com/microsoft/fluentui/blob/d55eecf434dfe269ecbf6f412dcde73a13dace3e/scripts/gulp/plugins/util/getComponentInfo.ts#L64

All `export default` will be catched by ESLint rule https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-default-export.md. I will enable it in the final PR

packages/fluentui/react-component-nesting-registry/src/index.ts
```diff
@@ -1,2 +1,2 @@
-export { default as Unstable_NestingAuto } from './NestingAuto';
```
miroslavstastny, line 1: default exports in `react-component-nesting-registry`:
image

layershifter: Fixed 👍

PR #13904, opened 2020-07-02
Calendar: Migrate components to function components

MLoughry:
#### Pull request checklist

#### Description of changes

As part of the continuing effort to migrate components to function components, especially those with UNSAFE methods, this change does so for the Calendar and associated components.

Additionally, changed the `@fluentui/react-next` package to export `Calendar` and `DatePicker` from `@uifabric/date-time`

#### Focus areas to test

(optional)


lorejoh12:
Calendar example "Inline Calendar with week selection", clicking the next and previous buttons change the selected date, but the component doesn't actually visually update to reflect that.

lorejoh12:
Calendar example "Calendar launched from a button"

Once a date has already been selected, after reopening, navigation is completely broken. Unable to navigate to past or future months using either the keyboard, or clicking on the arrow keys (or go to today button or buttons from the month picker).

MLoughry:
> Calendar example "Inline Calendar with week selection", clicking the next and previous buttons change the selected date, but the component doesn't actually visually update to reflect that.

@lorejoh12  It looks like this is because the example doesn't specify a controlled value initially, and then does if you use the navigation buttons. When I first implemented `useControllableValue()`, @dzearing said that the hook/component should be controlled or uncontrolled for its entire lifetime, rather than alternating.

lorejoh12:
> > Calendar example "Inline Calendar with week selection", clicking the next and previous buttons change the selected date, but the component doesn't actually visually update to reflect that.
> 
> @lorejoh12 It looks like this is because the example doesn't specify a controlled value initially, and then does if you use the navigation buttons. When I first implemented `useControllableValue()`, @dzearing said that the hook/component should be controlled or uncontrolled for its entire lifetime, rather than alternating.

makes sense, sounds like the component is doing what it needs to and the example needs to be updated to be fully controlled.

PR #13908, opened 2020-07-02
Allow overriding ContextualMenu's internal List's role

### LINKED ISSUE #13971: ContextualMenu control hard-codes the role of internal List element

Currently, the `ContextualMenu` control hard-codes the `role` attribute of the internal `List` as `menu`. In my usage, I want to set a different role, specifically `grid`. There's currently no way to achieve this.

I propose adding a `role` property to to the `IContextualMenuListProps` type so that the default List renderer knows what role it should set on the List element. If this property is undefined, we fall back to `menu`, which preserves the existing behavior.

I have submitted a [PR](https://github.com/microsoft/fluentui/pull/13908) for this.

**ContextualMenu**

No

In the `ContextualMenu.base.tsx` file, the `_onRenderMenuList()` function renders a `<ul>` element as:

```ts
 <ul className={this._classNames.list} onKeyDown={this._onKeyDown} onKeyUp={this._onKeyUp} role="menu">
```

Notice that it hard codes the role as `menu`.

dotnetjunky:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #13971 

#### Description of changes

Currently, the `ContextualMenu` control hard-code the `role` attribute of the internal List as `menu`. In my usage, I want to set a different role, specifically `grid`. There's currently no way to achieve this.

In this PR, I'm adding a `role` property to the `IContextualMenuListProps` struct. The idea is that the consumer of `ContextualMenu` would provide a custom implementation of `onRenderMenuList` callback. Inside the callback, it can set the `role` attribute to `grid`, before invoking the default renderer.

The default renderer inside `ContextualMenu` is modified to honor the `role` property. If `role` is not set, it falls back to `menu`, hence preserving the existing behavior. 

#### Focus areas to test

Added unit tests.

### DIFF
--- a/change/office-ui-fabric-react-2020-07-02-14-53-01-contextual-menu-list-role.json
+++ b/change/office-ui-fabric-react-2020-07-02-14-53-01-contextual-menu-list-role.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Allow consumer of ContextualMenu control to override the role of internal list.",
+  "packageName": "office-ui-fabric-react",
+  "email": "kinhln@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-07-02T21:53:01.420Z"
+}
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -3019,6 +3019,8 @@ export interface IContextualMenuListProps {
     // (undocumented)
     items: IContextualMenuItem[];
     // (undocumented)
+    role?: string;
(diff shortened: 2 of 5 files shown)


dotnetjunky:
@joschect Hi Jon, could you please review this PR? Thanks

ecraig12345:
Triggering new build. Also @joschect can you please review?

ecraig12345:
Trying again with the build...? Never seen this one before... https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=90635&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=8d0a7aee-8090-5653-72a7-7c81ab87cde7

PR #13912, opened 2020-07-03
FC example conversion for Announced, GroupedList, Coachmark, and ExtendedPeoplePicker

czearing:
#### Pull request checklist

#### Description of changes
1. Updated `announced`, `groupedList`, `coachmark`, `extendedPeoplePicker`examples to use function components.
2. Added useSetInterval and useSetTimeout within react-hooks package.

#### Focus areas to test
`announced`, `groupedList`, `coachmark`, `extendedPeoplePicker`


ecraig12345:
Triggering new build.

packages/office-ui-fabric-react/src/components/Coachmark/examples/Coachmark.Basic.Example.tsx
```diff
@@ -34,123 +29,100 @@ export interface ICoachmarkBasicExampleStyles {
...
+
+const buttonProps: IButtonProps = {
+  text: 'Try it',
+};
+
+const buttonProps2: IButtonProps = {
+  text: 'Try it again',
+};
+
+const IDropdownOption = [
```
ecraig12345, line 53: Probably not the name you meant to use :)
```suggestion
const dropdownOptions: IDropdownOption[] = [
```

packages/office-ui-fabric-react/src/components/Coachmark/examples/Coachmark.Basic.Example.tsx
```diff
@@ -34,123 +29,100 @@ export interface ICoachmarkBasicExampleStyles {
...
-      isCoachmarkVisible: false,
-    });
-  };
-
-  private _onDropdownChange = (event: React.FormEvent<HTMLDivElement>, option: IDropdownOption): void => {
-    this.setState({
-      coachmarkPosition: option.data,
-      dropdownSelectedOptionKey: option.key,
-    });
+const getClassNames = classNamesFunction<{}, ICoachmarkBasicExampleStyles>();
```
ecraig12345, line 32: While you're here, can you convert this to just use `mergeStyleSets` (and get rid of `ICoachmarkBasicExampleStyles`)? There's no reason to use `classNamesFunction` here, so it's just unnecessarily complicating the example.
```
const classNames = mergeStyleSets(/*styles object*/);
```

packages/office-ui-fabric-react/src/components/ExtendedPicker/examples/ExtendedPeoplePicker.Basic.Example.tsx
```diff
@@ -1,298 +1,242 @@
...
-import { IFocusZoneProps, FocusZoneTabbableElements } from 'office-ui-fabric-react/lib/FocusZone';
-import { mergeStyleSets, getTheme, IStyle, IProcessedStyleSet } from 'office-ui-fabric-react/lib/Styling';
+import { SuggestionsStore, FloatingPeoplePicker } from 'office-ui-fabric-react/lib/FloatingPicker';
+import { SelectedPeopleList, IExtendedPersonaProps } from 'office-ui-fabric-react/lib/SelectedItemsList';
+import { FocusZoneTabbableElements } from 'office-ui-fabric-react/lib/FocusZone';
+import { mergeStyleSets, getTheme } from 'office-ui-fabric-react/lib/Styling';
 import { people, mru, groupOne, groupTwo } from '@uifabric/example-data';
-
+import { useConst } from '@uifabric/react-hooks';
 export interface IPeoplePickerExampleState {
```
ecraig12345, line 11: Get rid of this interface

packages/office-ui-fabric-react/src/components/ExtendedPicker/examples/ExtendedPeoplePicker.Basic.Example.tsx
```diff
@@ -1,298 +1,242 @@
...
   };
 
-  private _onRemoveSuggestion = (item: IPersonaProps): void => {
-    const { peopleList, mostRecentlyUsed: mruState } = this.state;
+  const onRemoveSuggestion = (item: IPersonaProps): void => {
     const itemIndex = peopleList.indexOf(item);
-    const itemMruIndex = mruState.indexOf(item);
-
+    const itemMruIndex = mostRecentlyUsed.indexOf(item);
     const stateUpdate = {} as IPeoplePickerExampleState;
```
ecraig12345, line 65: Instead of making a stateUpdate object, just do the `setPeopleList` and `setMostRecentlyUsed` calls inline

packages/office-ui-fabric-react/src/components/GroupedList/examples/GroupedList.CustomCheckbox.Example.tsx
```diff
@@ -2,33 +2,30 @@ import * as React from 'react';
...
 
-  constructor(props: {}) {
-    super(props);
+const onRenderGroupHeaderCheckbox = (props: IGroupHeaderCheckboxProps) => <Toggle checked={props.checked} />;
 
-    this._items = createListItems(Math.pow(groupCount, groupDepth + 1));
-    this._columns = Object.keys(this._items[0])
+export const GroupedListCustomCheckboxExample: React.FunctionComponent = () => {
+  const items = useConst(createListItems(Math.pow(groupCount, groupDepth + 1)));
+  const groups = useConst(createGroups(groupCount, groupDepth, 0, groupCount));
```
ecraig12345, line 26: To avoid calling these every time, you should use initializer functions.
```suggestion
  const items = useConst(() => createListItems(Math.pow(groupCount, groupDepth + 1)));
  const groups = useConst(() => createGroups(groupCount, groupDepth, 0, groupCount));
```

packages/office-ui-fabric-react/src/components/GroupedList/examples/GroupedList.CustomCheckbox.Example.tsx
```diff
@@ -2,33 +2,30 @@ import * as React from 'react';
...
+  const columns = useConst(
```
ecraig12345, line 27: This should use an initializer function too
```suggestion
  const columns = useConst(() =>
```

packages/office-ui-fabric-react/src/components/Announced/examples/Announced.BulkOperations.Example.tsx
```diff
@@ -38,143 +50,116 @@ const _names: string[] = [
...
-    if (numberOfItems > 0) {
-      return (
-        <Announced message={`${numberOfItems} item${numberOfItems === 1 ? '' : 's'} moved`} aria-live={'assertive'} />
-      );
+  return item[column.key as keyof IFileExampleItem];
+};
+
+const stackTokens: IStackTokens = { childrenGap: 10 };
+
+export const AnnouncedBulkOperationsExample: React.FunctionComponent = () => {
```
ecraig12345, line 71: I started thinking about this example while testing it with Narrator and realized that it doesn't even make much sense...pretty sure a real Narrator user would not typically be moving things with drag/drop!! Ideally we should change this to demonstrate deleting multiple items or something like that (and get rid of the MarqueeSelection for simplicity), but it would also be okay to just leave it for now since this isn't a new problem.

packages/office-ui-fabric-react/src/components/Announced/examples/Announced.QuickActions.Example.tsx
```diff
@@ -53,177 +52,162 @@ export interface IAnnouncedQuickActionsExampleItem {
...
+        filesize: Math.floor(Math.random() * 30).toString() + ' MB',
+      });
+    }
+    return itemsList;
+  });
+  const [renameDialogOpen, setRenameDialogOpen] = React.useState<boolean>(false);
+
+  const [dialogContent, setDialogContent] = React.useState<JSX.Element | undefined>(undefined);
+  const [announced, setAnnounced] = React.useState<JSX.Element | undefined>(undefined);
+  const [previousAnnouncedValue, setPreviousAnnouncedValue] = React.useState<JSX.Element | undefined>(undefined);
```
ecraig12345, line 79: Instead of manually tracking this you can use `usePrevious` from `react-hooks`.

packages/office-ui-fabric-react/src/components/Announced/examples/Announced.QuickActions.Example.tsx
```diff
@@ -53,177 +52,162 @@ export interface IAnnouncedQuickActionsExampleItem {
...
+        key: i,
+        name: 'Item ' + i,
+        modified: getMockDateString(),
+        modifiedby: names[Math.floor(Math.random() * names.length)],
+        filesize: Math.floor(Math.random() * 30).toString() + ' MB',
+      });
+    }
+    return itemsList;
+  });
+  const [renameDialogOpen, setRenameDialogOpen] = React.useState<boolean>(false);
```
ecraig12345, line 75: Thought about this part a bit more and you could potentially get rid of this boolean and just check `!!dialogContent` instead (and set `dialogContent` back to undefined when closed)

packages/office-ui-fabric-react/src/components/Announced/examples/Announced.QuickActions.Example.tsx
```diff
@@ -53,177 +52,162 @@ export interface IAnnouncedQuickActionsExampleItem {
...
+
+  const [dialogContent, setDialogContent] = React.useState<JSX.Element | undefined>(undefined);
+  const [announced, setAnnounced] = React.useState<JSX.Element | undefined>(undefined);
+  const [previousAnnouncedValue, setPreviousAnnouncedValue] = React.useState<JSX.Element | undefined>(undefined);
+
+  const { setTimeout } = useSetTimeout();
+
+  const onRenderRow = useConstCallback(
```
ecraig12345, line 83: This one can go outside the component body

packages/office-ui-fabric-react/src/components/Announced/examples/Announced.QuickActions.Example.tsx
```diff
@@ -179,67 +112,87 @@ export class AnnouncedQuickActionsExample extends React.Component<{}, IAnnounced
...
+          filesize: Math.floor(Math.random() * 30).toString() + ' MB',
+        });
+      }
     }
-  }
-}
+  }, []);
+
+  // componentDidUpdate
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
ecraig12345: Did you figure out if there's any reason why we need to clear the announced value after 2s?

packages/office-ui-fabric-react/src/components/Announced/examples/Announced.QuickActions.Example.tsx
```diff
@@ -179,67 +112,87 @@ export class AnnouncedQuickActionsExample extends React.Component<{}, IAnnounced
...
 
-    switch (selectionCount) {
-      case 0:
-        return 'No items selected';
-      case 1:
-        return '1 item selected: ' + (this._selection.getSelection()[0] as any).name;
-      default:
-        return `${selectionCount} items selected`;
+  // Populate with items for demos.
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
ecraig12345: What's the reasoning for still doing this with useEffect not useConst?

packages/office-ui-fabric-react/src/components/Announced/examples/Announced.QuickActions.Example.tsx
```diff
@@ -53,177 +52,162 @@ export interface IAnnouncedQuickActionsExampleItem {
...
-  public render(): JSX.Element {
-    const { items, renameDialogOpen, dialogContent } = this.state;
-
-    return (
-      <>
-        {this._renderAnnounced()}
+  return (
+    <>
+      {announced}
+      <MarqueeSelection selection={selection}>
```
ecraig12345, line 193: Get rid of the MarqueeSelection here since none of the actions actually work on multi-select (I did this recently in the class version of the example)

packages/office-ui-fabric-react/src/components/Announced/examples/Announced.QuickActions.Example.tsx
```diff
@@ -53,177 +52,162 @@ export interface IAnnouncedQuickActionsExampleItem {
...
+      }
+    },
+    [deleteItem, renameItem],
+  );
+
+  const updateItemName = useConstCallback((index: number) => {
+    if (textField && textField.current) {
+      const renamedItems = items;
+      renamedItems[index].name = textField.current.value || renamedItems[index].name;
+      setItems([...renamedItems]);
```
ecraig12345, line 154: You should copy the items before mutating. Also don't mutate the original item in place.
```suggestion
      const renamedItems = [...items];
      renamedItems[index] = { ...renamedItems[index], name: textField.current.value || renamedItems[index].name };
      setItems(renamedItems);
```

packages/office-ui-fabric-react/src/components/Coachmark/examples/Coachmark.Basic.Example.tsx
```diff
@@ -1,156 +1,107 @@
...
+    },
+    [],
+  );
+
+  return (
+    <>
+      <div className={classNames.dropdownContainer}>
+        <Dropdown
+          label="Coachmark position"
+          selectedKey={dropdownSelectedOptionKey}
```
ecraig12345, line 68: It's simpler to just make the dropdown uncontrolled in this case.
```suggestion
  const onDropdownChange = React.useCallback(
    (event: React.FormEvent<HTMLDivElement>, option: IDropdownOption): void => {
      setCoachmarkPosition(option.data);
    },
    [],
  );

  return (
    <>
      <div className={classNames.dropdownContainer}>
        <Dropdown
          label="Coachmark position"
          defaultSelectedKey="H"
```

packages/office-ui-fabric-react/src/components/GroupedList/examples/GroupedList.CustomCheckbox.Example.tsx
```diff
@@ -37,54 +34,44 @@ export class GroupedListCustomCheckboxExample extends React.Component<{}, IGroup
...
-
-    this.state = {};
-  }
+      ),
+  );
+  const selection = useConst(() => {
+    const s = new Selection();
+    s.setItems(items);
+    return s;
+  });
```
ecraig12345, line 43: Pretty sure it's possible to just do this now (Selection didn't have an `items` prop in the past, which is why it was creating and updating separately before)
```suggestion
  const selection = useConst(() => new Selection({ items }));
```

packages/office-ui-fabric-react/src/components/GroupedList/examples/GroupedList.CustomCheckbox.Example.tsx
```diff
@@ -37,54 +34,44 @@ export class GroupedListCustomCheckboxExample extends React.Component<{}, IGroup
...
-    return <GroupHeader onRenderGroupHeaderCheckbox={this._onRenderGroupHeaderCheckbox} {...props} />;
-  };
-
-  private _onRenderGroupHeaderCheckbox = (props: IGroupHeaderCheckboxProps) => {
-    return <Toggle checked={props.checked} />;
-  };
-
-  private _onRenderCell = (nestingDepth: number, item: IExampleItem, itemIndex: number): JSX.Element => {
-    return (
+  const onRenderCell = React.useCallback(
```
ecraig12345, line 45: You can do `useConstCallback` here since neither `items` nor `selection` will ever change.

ecraig12345:
I just pushed an update majorly simplifying the Announced examples (I'd noticed so many issues that it was easier to just make the changes than trying to explain them) and fixing a couple issues related to stale captured values in lambdas. 

However the lazy loading example is still broken, and not quite for the reason I previously thought--the real issue is that the interval is being re-created every time `total` changes, so then it never actually runs. Maybe you can work with @dzearing to figure out a fix.

apps/fabric-website/src/pages/Styles/OfficeBrandIconsPage/docs/web/OfficeBrandIconsImplementation.md
```diff
@@ -8,17 +8,16 @@ The following code shows you how to specify a 96px product icon by brand using t
   rel="stylesheet"
   href="https://static2.sharepointonline.com/files/fabric/office-ui-fabric-core/11.0.0/css/fabric.min.css"
 />
-
 <div class="ms-BrandIcon--icon96 ms-BrandIcon--word"></div>
 ```
 
-This following code shows you how to specify a 48px product icon by brand using the [office-ui-fabric-core](https://github.com/OfficeDev/office-ui-fabric-core) SVG and an `<img>` element: 
+This following code shows you how to specify a 48px product icon by brand using the [office-ui-fabric-core](https://github.com/OfficeDev/office-ui-fabric-core) SVG and an `<img>` element:
```
ecraig12345, line 14: This should have been fixed already by #14388. Or if not I can fix it and push to master, and then you can merge.

ecraig12345:
Looks like it's loading now, but instead of announcing every 2 seconds it's announcing much more frequently. So depending on the user's configured speech speed they hear something like "1 of 15 photos l-- 2 of 15 pho-- 3 of 15 photos--" or "1 of 15-- 2 of 15-- 3 of 15--".

PR #13934, opened 2020-07-07
chore: remove default exports [P5]

layershifter:
#### Description of changes

This PR is a fifth round of `default` exports removal, see https://github.com/microsoft/fluentui/issues/12953#issuecomment-633970215.


packages/fluentui/e2e/.eslintrc.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": ["plugin:@fluentui/eslint-plugin/react-northstar"],
-  "root": true
+  "root": true,
+  "rules": {
+    "import/no-default-export": "off"
```
pompomon, line 5: Please why it's "off" for e2e?

layershifter: `e2e` is not a part of distributed code so it's not important to fix things there.

packages/fluentui/react-northstar/src/index.ts
```diff
@@ -17,304 +23,184 @@ export * from './themes/teams/types';
...
 export * from './components/Button/ButtonGroup';
-export { default as ButtonGroup } from './components/Button/ButtonGroup';
 export * from './components/Button/ButtonContent';
-export { default as ButtonContent } from './components/Button/ButtonContent';
 
 export * from './components/Chat/Chat';
-export { default as Chat } from './components/Chat/Chat';
 export * from './components/Chat/ChatItem';
-export { default as ChatItem } from './components/Chat/ChatItem';
+export * from './components/Chat/ChatItem';
```
pompomon, line 49:
```suggestion
```
Duplicate line

packages/fluentui/docs/src/config.ts
```diff
@@ -1,4 +1,5 @@
 const config = {
+  repoURL: 'https://github.com/microsoft/fluentui',
```
pompomon, line 2: Should we import it from `packages/fluentui/react-northstar/src/utils/constants.ts` ? So we don't have to change URL in two places

layershifter: `constants` are no longer reexported under `@fluentui/react-northstar` (I think that this variable was exported by a mistake).

PR #13939, opened 2020-07-07
React conformance - Adding wrappers to packages

sopranopillow:
#### Pull request checklist

#### Description of changes

This draft PR adds wrappers for the following packages:
* react-northstar
* react-next [not done]
* office-ui-fabric-react [not done]


packages/fluentui/react-northstar/test/specs/components/Accordion/AccordionContent-test.tsx
```diff
@@ -1,20 +1,18 @@
 import * as React from 'react';
-
 import AccordionContent from 'src/components/Accordion/AccordionContent';
+
 import { isConformant, handlesAccessibility, getRenderedAttribute } from 'test/specs/commonTests';
-import { mountWithProviderAndGetComponent, mountWithProvider as mount } from 'test/utils';
-import { isConformant as newIsConformant } from '@fluentui/react-conformance';
+import { mountWithProviderAndGetComponent } from 'test/utils';
+import { sharedIsConformant } from 'test/utils';
```
ecraig12345, line 6:
```suggestion
import { mountWithProviderAndGetComponent, sharedIsConformant } from 'test/utils';
```

packages/fluentui/react-northstar/test/specs/components/Accordion/AccordionContent-test.tsx
```diff
@@ -1,20 +1,18 @@
...
 
 describe('AccordionContent', () => {
   isConformant(AccordionContent, { constructorName: 'AccordionContent' });
 
-  newIsConformant({
+  sharedIsConformant({
     Component: AccordionContent,
     componentPath: __filename.replace(/test[/\\]specs/, 'src').replace('-test.tsx', '.tsx'),
     displayName: 'AccordionContent',
     useDefaultExport: true,
```
ecraig12345, line 15: You can probably move this to the package defaults.

packages/fluentui/react-northstar/test/utils/sharedIsConformant.tsx
```diff
@@ -0,0 +1,15 @@
+import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
+import { mountWithProvider as mount } from 'test/utils';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
+// import { FocusZone } from '@fluentui/react-bindings';
+
+export function sharedIsConformant(testInfo: IsConformantOptions) {
+  const { disabledTests = [], wrapperComponents = [] } = testInfo;
+
+  isConformant({
+    ...testInfo,
```
ecraig12345, line 10: This won't handle merging disabledTests or wrapperComponents arrays properly. In this package you can use `_.merge` from lodash.

Also, you should put the defaults first and then merge in the custom options, so the custom options take precedence. `_.merge(/*default options*/, testInfo)`

packages/office-ui-fabric-react/src/common/sharedIsConformant.tsx
```diff
@@ -0,0 +1,12 @@
+import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
+
+export function sharedIsConformant(testInfo: IsConformantOptions) {
+  const { disabledTests = [], wrapperComponents = [] } = testInfo;
+
+  isConformant({
```
ecraig12345, line 7: For correct deep merging of arrays, you need to use `merge` from `../Utilities`, and pass in the user options second. `merge(/*defaults*/, testInfo)`

packages/react-next/src/common/sharedIsConformant.tsx
```diff
@@ -0,0 +1,12 @@
+import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
+
+export function sharedIsConformant(testInfo: IsConformantOptions) {
+  const { disabledTests = [], wrapperComponents = [] } = testInfo;
+
+  isConformant({
```
ecraig12345, line 7: Same comment here as in OUFR.

packages/react-conformance/package.json
```diff
@@ -28,8 +28,7 @@
     "@types/react-test-renderer": "^16.0.0",
     "@types/react-is": "^16.7.1",
     "@uifabric/build": "^7.0.0",
-    "@fluentui/react-bindings": "^0.50.0",
-    "@fluentui/react-component-ref": "^0.50.0",
+    "@uifabric/utilities": "^7.22.0",
```
ecraig12345, line 31: You can remove this now

packages/react-conformance/src/utils/merge.tsx
```diff
@@ -0,0 +1,41 @@
+import * as _ from 'lodash';
+
+// tslint:disable-next-line:no-any
```
ecraig12345, line 3: You'll need to merge with master and switch these to `eslint-disable @typescript-eslint/no-explicit-any`

packages/react-conformance/src/utils/merge.tsx
```diff
@@ -0,0 +1,41 @@
+import * as _ from 'lodash';
```
ecraig12345, line 1: You shouldn't need to use lodash for merging here. Probably best to just copy the implementation from scripts/tasks/merge.js unless that didn't work for some reason.

packages/react-conformance/src/utils/merge.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Merge function with specific type
+ * @param mergeArrayPrimitive Merge array with primitive value, i.e.:
+ *  a { ob: ['foo', 'bar'] }
+ *  b { ob: 'a' }
+ *  merge(true, a, b) === { ob: ['foo', 'bar', 'a'] }
+ * @param objs
+ */
+export function merge<T = Object>(mergeArrayPrimitive: boolean, ...objs: Object[]): T {
```
ecraig12345, line 32: Can you add a test for this?

packages/react-conformance/src/utils/merge.tsx
sopranopillow: sorry forgot about that

packages/react-conformance/src/utils/merge.tsx
sopranopillow: I'm only using lodash to check for types, if a property exists in the object and to insert an attribute that doesn't exist in the first object. I'm basing everything in obj1 and copying everything else from obj2.

packages/react-conformance/src/utils/merge.tsx
sopranopillow: Okay! 😃

packages/react-conformance/src/utils/merge.tsx
ecraig12345: You don't need lodash for either of those things. For `has` you can do `!!obj1 && obj1.hasOwnProperty(k)`. For `isArray` you can do `Array.isArray(foo)`.

sopranopillow: I see, will change it!

packages/fluentui/react-northstar/test/utils/sharedIsConformant.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { mountWithProvider } from 'test/utils';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
+import { FocusZone } from '@fluentui/react-bindings';
+import * as _ from 'lodash';
+
+export function sharedIsConformant(testInfo: IsConformantOptions) {
+  const { componentPath } = testInfo;
+  const defaultConfig = {
+    customMount: mountWithProvider,
+    componentPath: componentPath.replace(/test[/\\]specs/, 'src').replace('-test.tsx', '.tsx'),
```
ecraig12345, line 11: This is a good idea, but I think with the current implementation the raw filename would still override the corrected one. Instead, maybe you could add a new param `testFilePath` to `sharedIsConformant`?

ecraig12345: Another benefit of the separate param is it's a bit misleading to be passing the test filename as **component**Path

packages/office-ui-fabric-react/src/common/sharedIsConformant.ts
```diff
@@ -0,0 +1,12 @@
+import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
+import * as _ from 'lodash';
+
+export function sharedIsConformant(testInfo: IsConformantOptions) {
+  const defaultOptions = {
+    disabledTests: ['has-docblock'],
+    wrapperComponents: [Ref, RefFindNode],
```
ecraig12345, line 8: These aren't used in OUFR, so don't list them here.

packages/react-next/src/common/sharedIsConformant.ts
```diff
@@ -0,0 +1,12 @@
+import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
+import * as _ from 'lodash';
+
+export function sharedIsConformant(testInfo: IsConformantOptions) {
+  const defaultOptions = {
+    disabledTests: ['has-docblock', 'kebab-aria-attributes'],
+    wrapperComponents: [Ref, RefFindNode],
```
ecraig12345, line 8: These aren't used currently in react-next and shouldn't be listed here

packages/react-conformance/src/utils/merge.tsx
```diff
@@ -0,0 +1,41 @@
...
```
ecraig12345, line 32: Why is merging arrays opt-in? At least to me it seems like that should be the default behavior, with no need to override it in this case.

sopranopillow: Arrays are always merged. This flag is so that when you have something like an integer or a value that isn't an array but the field with the same name in the second object is an array  you can decide if you want to merge the array with the single value. Something like:

`
obj1 = {
  foo: 1
}
obj2 = {
  foo: [2, 3, 4]
}

merge(true, obj1, obj2) === { foo: [1, 2, 3, 4] }
merge(false, obj1, obj2) === { foo: 1 }
`

sopranopillow: I thought this could be useful but I can remove it if it makes sense.

ecraig12345: Hmm interesting. That's a tricky case to know what the "right" thing to do is. But it's also not something we're very likely to encounter here, so it's probably okay to take a simple approach: if the first item and second item are different types, the second entirely overwrites the first.

packages/office-ui-fabric-react/src/common/sharedIsConformant.ts
```diff
@@ -0,0 +1,11 @@
+import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
+import { Ref } from '@fluentui/react-component-ref';
+
+export function sharedIsConformant(testInfo: IsConformantOptions) {
+  const defaultOptions = {
+    disabledTests: ['has-docblock'],
+    helperComponents: [Ref],
```
ecraig12345, line 7: OUFR and react-next don’t use Ref. This should be either not provided or just have FocusZone.

packages/fluentui/react-northstar/test/specs/components/Menu/MenuItem-test.tsx
```diff
@@ -23,14 +22,15 @@ describe('MenuItem', () => {
...
-    customMount: mount,
-    componentPath: __filename.replace(/test[/\\]specs/, 'src').replace('-test.tsx', '.tsx'),
-    wrapperComponent: MenuItemWrapper,
-    disabledTests: ['has-top-level-file'],
-  });
+  sharedIsConformant(
+    {
+      Component: MenuItem,
+      displayName: 'MenuItem',
+      componentPath: '',
```
ecraig12345, line 29: Why the empty string here?

packages/fluentui/react-northstar/test/utils/sharedIsConformant.tsx
```diff
@@ -0,0 +1,15 @@
+import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
+import { mountWithProvider } from 'test/utils';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
+import { FocusZone } from '@fluentui/react-bindings';
+
+export function sharedIsConformant(testInfo: IsConformantOptions, testFilePath: string) {
+  const defaultConfig = {
+    customMount: mountWithProvider,
+    componentPath: testFilePath.replace(/test[/\\]specs/, 'src').replace('-test.tsx', '.tsx'),
+    disabledTests: ['has-docblock', 'has-top-level-file'],
```
ecraig12345, line 10: has-docblock should be enabled for react-northstar

packages/office-ui-fabric-react/package.json
```diff
@@ -69,6 +68,8 @@
     "@fluentui/react-focus": "^7.12.25",
     "@fluentui/react-icons": "^0.1.40",
     "@microsoft/load-themed-styles": "^1.10.26",
+    "@fluentui/react-conformance": "^0.1.0",
+    "@fluentui/react-component-ref": "^0.50.0",
```
ecraig12345, line 72: Remove the dep on react-component-ref and move react-conformance back to devDependencies (test-only stuff should never be in dependencies). If that makes a lint rule complain we’ll fix the lint rule config.

packages/react-next/package.json
```diff
@@ -60,6 +59,7 @@
   "dependencies": {
     "@fluentui/react-button": "^0.7.7",
     "@fluentui/react-compose": "^0.12.5",
+    "@fluentui/react-conformance": "^0.1.0",
```
ecraig12345, line 62: Move this back to devDependencies

packages/fluentui/react-northstar/test/specs/components/Accordion/AccordionContent-test.tsx
```diff
@@ -2,19 +2,19 @@ import * as React from 'react';
...
-    Component: AccordionContent,
-    componentPath: __filename.replace(/test[/\\]specs/, 'src').replace('-test.tsx', '.tsx'),
-    displayName: 'AccordionContent',
-    customMount: mount,
-    disabledTests: ['has-docblock', 'has-top-level-file'],
-  });
+  sharedIsConformant(
+    {
+      Component: AccordionContent,
+      componentPath: '',
```
ecraig12345, line 13:
```suggestion
```

packages/fluentui/react-northstar/test/specs/components/Slider/Slider-test.tsx
```diff
@@ -1,16 +1,9 @@
...
 
 describe('Slider', () => {
-  newIsConformant({
-    componentPath: __filename.replace(/test[/\\]specs/, 'src').replace('-test.tsx', '.tsx'),
-    Component: Slider,
-    displayName: 'Slider',
-    customMount: mount,
-    disabledTests: ['has-docblock', 'has-top-level-file'],
-  });
+  sharedIsConformant({ Component: Slider, componentPath: '', displayName: 'Slider' }, __filename);
```
ecraig12345, line 6:
```suggestion
  sharedIsConformant({ Component: Slider, displayName: 'Slider' }, __filename);
```

packages/fluentui/react-northstar/test/utils/sharedIsConformant.tsx
```diff
@@ -0,0 +1,15 @@
+import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
+import { mountWithProvider } from 'test/utils';
+import { Ref, RefFindNode } from '@fluentui/react-component-ref';
+import { FocusZone } from '@fluentui/react-bindings';
+
+export function sharedIsConformant(testInfo: Partial<IsConformantOptions>, testFilePath: string) {
+  const defaultConfig = {
```
ecraig12345, line 7: If you explicitly type this, that will potentially make it easier to find errors or handle refactorings in the future.
```suggestion
  const defaultConfig: Partial<IsConformantOptions> = {
```

packages/office-ui-fabric-react/src/common/sharedIsConformant.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
ecraig12345, line 1: In this case it's good to give an explanation for why the rule is being disabled (note that you specifically need to use `--` with a space on each side for eslint to accept it)
```suggestion
// eslint-disable-next-line import/no-extraneous-dependencies -- this file is for testing
```

packages/react-conformance/src/isConformant.ts
```diff
@@ -4,9 +4,11 @@ import * as fs from 'fs';
 import { IsConformantOptions } from './types';
 import { withCustomConfig } from 'react-docgen-typescript';
 import { defaultTests } from './defaultTests';
+import { merge } from './utils/merge';
 
-export function isConformant(testInfo: IsConformantOptions) {
-  const { componentPath, displayName, disabledTests = [], extraTests, isInternal } = testInfo;
+export function isConformant(...testInfo: (IsConformantOptions | Object)[]) {
```
ecraig12345, line 9:
```suggestion
export function isConformant(...testInfo: Partial<IsConformantOptions>[]) {
```

packages/react-conformance/src/utils/merge.tsx
```diff
@@ -0,0 +1,33 @@
+import * as _ from 'lodash';
```
ecraig12345, line 1: I think this is unused now?

packages/react-next/src/common/sharedIsConformant.ts
```diff
@@ -0,0 +1,9 @@
+import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
```
ecraig12345, line 1:
```suggestion
// eslint-disable-next-line import/no-extraneous-dependencies -- this file is for testing
import { isConformant, IsConformantOptions } from '@fluentui/react-conformance';
```

packages/react-conformance/package.json
```diff
@@ -38,10 +38,11 @@
     "react-dom": "16.8.6"
   },
   "dependencies": {
-    "tslib": "^1.10.0",
-    "react-is": "^16.6.3",
+    "doctrine": "^2.0.0",
```
layershifter, line 41: Why not the latest version? In any case, this package was deprecated two years ago https://github.com/eslint/doctrine

packages/react-conformance/src/utils/merge.tsx
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function merge<T = Object>(...objs: Object[]): T {
+  const merged = { ...objs[0] };
+  objs = objs.slice(1);
+
+  objs.forEach(option => {
+    _merge(merged, option);
+  });
+
+  return merged as T;
```
layershifter, line 30: Just wondering why this is better than any merge function, I
e. from Lodash?

packages/react-conformance/package.json
ecraig12345: The docblock conformance test was copied from react-northstar, so we used the same deps and versions as you have. We can use something else if you have another suggestion, but I'd prefer to fix that separately because this PR has been held up way too long.

packages/react-conformance/src/utils/merge.tsx
ecraig12345: IIRC Lodash's merge may not have handled arrays the way we needed (concatenating).

packages/react-conformance/package.json
layershifter: It's not a good reason to introduce deprecated packages to a new code. Please at least add a to replace this dependency.

packages/react-conformance/src/utils/merge.tsx
ecraig12345: Confirmed here: https://runkit.com/ecraig12345/5f1b54969d6aa400192f4eb7

layershifter: According to unit tests in this PR I don't see any difference as Lodash performs deep merge, too.

layershifter: Oops, I see 🙃 However, `mergeWith()` the exact thing that you want, https://lodash.com/docs/4.17.15#mergeWith. An example in docs covers this use case

packages/react-conformance/package.json
ecraig12345: Despite the repo being archived, the [npm package](https://www.npmjs.com/package/doctrine) is not marked as deprecated and is extremely widely used (21 million downloads/week).

For alternatives, the next most widely used and similarly lightweight alternative that I could find is [`comment-parser`](https://www.npmjs.com/package/comment-parser) at 344k/week.

packages/react-conformance/src/utils/merge.tsx
ecraig12345: Just tried out a bunch of merge functions and `webpack-merge` is the only thing I could find with the desired array behavior. Seems weird to use something with "webpack" in the name for this generic purpose, but I guess we could? Probably better than maintaining our own.

ecraig12345: Oops just saw the mergeWith comment. Tracking that and the dependency change in #14205.

PR #13944, opened 2020-07-08
Use ESLint everywhere

ecraig12345:
#### Pull request checklist

#### Description of changes

Use ESLint everywhere:
- Change `tslint:disable` to `eslint-disable` and change rule names
- A few instances of actually fixing violations. Most commonly:
  - Use `forEach` instead of `map` when no result is returned (especially common in charting package)
  - Remove unnecessary `!!`
  - Add missing underscores to private method names (we had this rule previously but it wasn't being enforced)
  - prefer const over let in a few places
  - single variable declaration per line
  - no `void 0`
  - Replace `alert` with `console.log` in some tests and examples
  - Remove string literal property access when not needed
  - Remove unused component state properties, only in examples or internal stuff
  - Add public visibility modifiers to getters which had no visibility marking
  - fabric-website: fix typing and imports of json metadata files, and svg property naming violations
- Add deps on `@fluentui/eslint-plugin` and remove `@uifabric/tslint-rules` and tslint.json
- A few more config fixes: 
  - Don't check for unused disable directives during lint-staged (rules like `deprecation/deprecation` which are off for lint-staged will cause false positives)
  - Use helper function for generating naming convention rule (it didn't merge settings properly otherwise)
  - Add `no-tslint-comment` rule which bans `tslint:*` comments
- Add lint and lint-staged scripts to prevent reintroduction of tslint stuff


packages/example-app-base/src/components/ApiReferencesTable/index.ts
```diff
@@ -1,3 +1,3 @@
-export * from './ApiReferencesTable';
+export { ApiReferencesTable, IApiReferencesTableState, gapTokens } from './ApiReferencesTable';
```
khmakoto, line 1: What's the reason for changing this export?

ecraig12345: ApiReferencesTable.tsx re-exports IApiReferencesTableProps (from the types file) for some reason. I think this got flagged by a no duplicate exports rule at some point, though I might have turned that rule off due to issues in other places or bad perf.

packages/tsx-editor/src/components/Editor.tsx
```diff
@@ -64,7 +63,7 @@ export const Editor: React.FunctionComponent<IEditorProps> = (props: IEditorProp
       }
 
       if (currDebounceTime) {
-        debounceTimeout = setTimeout(() => currOnChange(model.getValue()), currDebounceTime);
+        debounceTimeout = (setTimeout(() => currOnChange(model.getValue()), currDebounceTime) as unknown) as number;
```
khmakoto, line 66: I think you can do `window.setTimeout` to get the correct typing here and not have to do that `as unknown as number` cast.

packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.base.tsx
```diff
@@ -16,7 +16,6 @@ const getClassNames = classNamesFunction<IMessageBarStyleProps, IMessageBarStyle
 
 export interface IMessageBarState {
   labelId?: string;
-  showContent?: boolean;
```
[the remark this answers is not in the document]
ecraig12345: Almost certainly, but I can bring it back to be extra safe for now (since it will be gone in v8 anyway).

packages/tsx-editor/src/components/Editor.tsx
ecraig12345: Good point, thanks!

PR #13948, opened 2020-07-08
Fix: Normalize date in tests without setting timezone

### LINKED ISSUE #13858: @fluentui/date-time-utilities tests do not run on Windows.
### Environment Information

- **Package version(s)**: 7.1.2
- **Browser and OS versions**: Windows build machine

Relates to build so the codepen is irrelevant.

#### Actual behavior:

`process.env.TZ = 'UTC';` jest [setup](https://github.com/microsoft/fluentui/blob/6525619bf5e6c844406ed3cb9ad3103237eff89f/packages/date-time-utilities/config/timezone-setup.js#L2) does not work on Windows. There are nodejs issues ([1](https://github.com/nodejs/node/issues/4230), [2](https://github.com/nodejs/node/issues/31478)) explaining why this is the case. 

#### Expected behavior:

The [following tests](https://github.com/microsoft/fluentui/blob/6525619bf5e6c844406ed3cb9ad3103237eff89f/packages/date-time-utilities/src/dateGrid/getDayGrid.test.ts#L21) should succeed on all types of build machines.

pompomon:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13858

#### Description of changes

Set timzone for tests via environment variables instead of `process.env.TZ` in Jest config

### DIFF
--- a/change/@fluentui-date-time-utilities-2020-07-08-09-37-29-fix-date-time-utils-tz-setup.json
+++ b/change/@fluentui-date-time-utilities-2020-07-08-09-37-29-fix-date-time-utils-tz-setup.json
@@ -1,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "Set timezone for tests via env variable",
+  "packageName": "@fluentui/date-time-utilities",
+  "email": "pompomon@gmail.com",
+  "dependentChangeType": "none",
+  "date": "2020-07-08T07:37:29.758Z"
+}
--- a/change/@fluentui-date-time-utilities-2020-07-08-09-37-29-fix-date-time-utils-tz-setup.json
+++ b/change/@fluentui-date-time-utilities-2020-07-08-09-37-29-fix-date-time-utils-tz-setup.json
@@ -1,6 +1,6 @@
 {
   "type": "none",
-  "comment": "Set timezone for tests via env variable",
+  "comment": "Normalize dates in tests instead of setting timezone in the config",
   "packageName": "@fluentui/date-time-utilities",
   "email": "pompomon@gmail.com",
   "dependentChangeType": "none",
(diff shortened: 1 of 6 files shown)


change/@fluentui-date-time-utilities-2020-07-08-09-37-29-fix-date-time-utils-tz-setup.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "Set timezone for tests via env variable",
```
ecraig12345, line 3: Should probably update this comment now?

packages/date-time-utilities/src/dateGrid/getDayGrid.test.ts
```diff
@@ -1,39 +1,38 @@
...
-    selectedDate: new Date(2016, Months.Apr, 1),
-    navigatedDate: new Date(2016, Months.Apr, 1),
+    selectedDate: defaultDate,
+    navigatedDate: defaultDate,
     firstDayOfWeek: DayOfWeek.Sunday,
     firstWeekOfYear: FirstWeekOfYear.FirstDay,
     dateRangeType: DateRangeType.Day,
   };
 
+  const normalizeDay = (day: IDay) => {
```
ecraig12345, line 16: Would be good to add some comments about why this is necessary so it doesn't get accidentally removed in the future.

change/@fluentui-date-time-utilities-2020-07-08-09-37-29-fix-date-time-utils-tz-setup.json
pompomon: Yeah, good catch, thank you! :)

PR #13956, opened 2020-07-08
[Date-Time] refactor date time examples

lorejoh12:
#### Pull request checklist

#### Description of changes

Refactoring the Calendar and WeeklyDatePicker examples so that they all show up in the local storybook build. Storybook just looks for any *.Example.tsx files, which means no props can be passed in to the component, so this separates all the various configurations into separate files with the props hardcoded. As a bonus, it also makes the code examples linked much more clear about exactly what configuration is being used.

While working on this, found a bug where an externally controlled Calendar component would update the selected date but would not navigate when the prop value changed. Added a fix for it.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/date-time/src/components/Calendar/Calendar.base.tsx
+++ b/packages/date-time/src/components/Calendar/Calendar.base.tsx
@@ -81,14 +81,22 @@ const DEFAULT_PROPS: Partial<ICalendarProps> = {
 };
 
 function useDateState({ value, today = new Date(), onSelectDate }: ICalendarProps) {
+  /** The currently selected date in the calendar */
+  const [selectedDate = today, setSelectedDate] = useControllableValue(value, today);
+
   /** The currently focused date in the day picker, but not necessarily selected */
   const [navigatedDay = today, setNavigatedDay] = React.useState(value);
 
   /** The currently focused date in the month picker, but not necessarily selected */
   const [navigatedMonth = today, setNavigatedMonth] = React.useState(value);
 
-  /** The currently selected date in the calendar */
-  const [selectedDate = today, setSelectedDate] = useControllableValue(value, today);
+  /** If using a controlled value, when that value changes, navigate to that date */
+  const [lastSelectedDate = today, setLastSelectedDate] = React.useState(value);
+  if (value && lastSelectedDate.valueOf() !== value.valueOf()) {
+    setNavigatedDay(value);
+    setNavigatedMonth(value);
+    setLastSelectedDate(value);
+  }
 
   const navigateMonth = (date: Date) => {
     setNavigatedMonth(date);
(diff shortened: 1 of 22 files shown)


packages/date-time/src/components/Calendar/examples/Calendar.Inline.ContiguousWorkWeekDays.Example.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Calendar, DateRangeType, DayOfWeek, defaultDayPickerStrings } from '@uifabric/date-time';
+
+import * as styles from './Calendar.Example.scss';
+
+export interface ICalendarInlineExampleState {
+  selectedDate?: Date;
+  selectedDateRange?: Date[];
+}
+
+export class CalendarInlineContiguousWorkWeekDaysExample extends React.Component<{}, ICalendarInlineExampleState> {
```
dzearing, line 11: Is this just a file move, or a new example? It would be good to make this a simple function component. Could happen separately in a PR.

packages/office-ui-fabric-react/src/components/Calendar/examples/Calendar.Inline.Example.tsx
```diff
@@ -108,7 +108,7 @@ export const CalendarInlineExample: React.FunctionComponent<ICalendarInlineExamp
     <div style={divStyle}>
       {
         <div>
-          Selected date(s): <span>{!selectedDate ? 'Not set' : selectedDate.toLocaleString()}</span>
+          Selected 2date(s): <span>{!selectedDate ? 'Not set' : selectedDate.toLocaleString()}</span>
```
dzearing, line 111: not sure this change was intended.

packages/date-time/src/components/Calendar/examples/Calendar.Inline.ContiguousWorkWeekDays.Example.tsx
lorejoh12: I'd like to get these checked in sooner rather than later, but will make a note to convert them to functional components in a future PR. This is basically just breaking apart the old example (which accepted props) into multiple example files that don't have props, to support storybook scenario. The conversion to functional components will require a little more time

packages/office-ui-fabric-react/src/components/Calendar/examples/Calendar.Inline.Example.tsx
lorejoh12: :D good call, will fix

packages/date-time/src/components/Calendar/Calendar.base.tsx
```diff
@@ -81,14 +81,22 @@ const DEFAULT_PROPS: Partial<ICalendarProps> = {
...
   const [navigatedDay = today, setNavigatedDay] = React.useState(value);
 
   /** The currently focused date in the month picker, but not necessarily selected */
   const [navigatedMonth = today, setNavigatedMonth] = React.useState(value);
 
-  /** The currently selected date in the calendar */
-  const [selectedDate = today, setSelectedDate] = useControllableValue(value, today);
+  /** If using a controlled value, when that value changes, navigate to that date */
+  const [lastSelectedDate = today, setLastSelectedDate] = React.useState(value);
+  if (value && lastSelectedDate.valueOf() !== value.valueOf()) {
```
ecraig12345, line 95: Coming in late here (sorry) but I'm pretty sure we shouldn't be setting state while rendering; seems like that could lead to unpredictable behavior. Could you do this with a ref (for the old value) and useEffect instead?

PR #13961, opened 2020-07-09
fix: make getShorthandInfo compatible with `compose()`

layershifter:
#### Description of changes

This PR updates `getShorthandInfo()` to support the extraction for `shorthandConfig` from composed components, too. I.e.

```tsx
const Button = compose(
  any,
  {
    shorthandConfig: { mappedProp: "content" }
  }
);
```

#### Focus areas to test

(optional)

### DIFF
--- a/scripts/gulp/plugins/util/getShorthandInfo.ts
+++ b/scripts/gulp/plugins/util/getShorthandInfo.ts
@@ -47,6 +47,19 @@ const isShorthandExpression = (componentName: string, path: NodePath<t.Assignmen
   );
 };
 
+/**
+ * Checks that an expression is a compose() call.
+ */
+function isCompose(path: NodePath<t.Expression>): path is NodePath<t.CallExpression> {
+  if (path.isCallExpression()) {
+    if (path.get('callee').isIdentifier({ name: 'compose' })) {
+      return true;
+    }
+  }
+
+  return false;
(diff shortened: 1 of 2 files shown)


scripts/gulp/plugins/util/getShorthandInfo.ts
```diff
@@ -81,6 +94,35 @@ const getShorthandInfo = (componentFile: t.File, componentName: string): Shortha
...
+              t.ObjectProperty
+            >[];
+            const shorthandMappedPropertyPath = shorthandConfigPropertiesPaths.find(property =>
+              (property.get('key') as NodePath<t.Identifier>).isIdentifier({ name: 'mappedProp' }),
+            );
+
+            if (shorthandMappedPropertyPath) {
+              // @ts-ignore
+              t.assertStringLiteral(shorthandMappedPropertyPath.node.value);
+              mappedShorthandProp = (shorthandMappedPropertyPath.node.value as t.StringLiteral).value;
```
mnajdova, line 120: Did we ever showed here the `mappedArrayProp`?

layershifter: Nope. `mappedShorthandProp` is used for our playground to guess the mapping to a proper prop and then get its type.

PR #13964, opened 2020-07-09
fix(Button): Respect hidden HTML attribute on components

### LINKED ISSUE #13733: Button: hidden prop overriden by styles
### Environment Information

- **Package version(s)**: 0.50

`hidden` prop does not hide the `Button` as `display: inline-flex` is aways applied

https://codesandbox.io/s/stq2b?module=/example.js

pompomon:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #13733

#### Description of changes

Add priority CSS rule for `[hidden]` HTML attribute in global styles, so it's applied for all components, which don't handle `hidden` prop

### DIFF
--- a/packages/fluentui/react-northstar/src/themes/teams/staticStyles/globalStyles.ts
+++ b/packages/fluentui/react-northstar/src/themes/teams/staticStyles/globalStyles.ts
@@ -17,4 +17,8 @@ export const globalStyles: StaticStyleFunction = siteVars => ({
   '*:after': {
     boxSizing: 'border-box',
   },
+  /* Adding priority for HTML `hidden` attribute to be applied correctly */
+  '[hidden]': {
+    display: 'none!important',
+  },
 });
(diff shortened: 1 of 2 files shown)


packages/fluentui/react-northstar/src/themes/teams/staticStyles/globalStyles.ts
```diff
@@ -17,4 +17,8 @@ export const globalStyles: StaticStyleFunction = siteVars => ({
   '*:after': {
     boxSizing: 'border-box',
   },
+  /* Adding priority for HTML `hidden` attribute to be applied correctly */
```
pompomon, line 20: Based on the discussions here https://css-tricks.com/the-hidden-attribute-is-visibly-weak/ and here https://meowni.ca/hidden.is.a.lie.html - I'm opting to add !important to the `[hidden]` style as the most robust solution

packages/fluentui/react-northstar/src/themes/teams/staticStyles/globalStyles.ts
```diff
@@ -17,4 +17,8 @@ export const globalStyles: StaticStyleFunction = siteVars => ({
...
+  '[hidden]': {
+    display: 'none!important',
```
mnajdova, line 22: A bit scary to use !important in global styles, but I don't have a better idea for this..

pompomon: Yeah, from what I've found there is nothing much else to do for that case :( 
Some of other CSS frameworks are doing the same

mnajdova:
Can you update the PR description to reflect that we are adding something in the global styles, this is not a fix only for the Button component

pompomon:
> Can you update the PR description to reflect that we are adding something in the global styles, this is not a fix only for the Button component

Done :)

PR #13965, opened 2020-07-09
CodeMods: Implement Maybe to help deal with null/undefined checks.

joschect:
#### Pull request checklist

#### Description of changes

Maybe makes it very easy to transform and deal with possibly undefined values. It can greatly help clarify and streamline chains of code that operate on something that may or may not exist.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/codemods/src/maybe.ts
+++ b/packages/codemods/src/maybe.ts
@@ -32,10 +32,7 @@ export interface Nothing<T> extends MB<T> {
 
 export type Maybe<T> = Just<T> | Nothing<T>;
 
-export const Nothing = <T>(): Nothing<T> => ({
-  ..._makeMaybe(),
-  just: false,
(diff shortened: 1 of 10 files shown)


packages/codemods/src/maybe.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * Maybe is a pattern that helps ensure that you don't need null checks everywhere.
+ * I endevoured to make it work the way promises do so it should feel familiar.
+ * Using Maybe.then means that you can chain many functions without needing to check for null/undefined
+ * constantly.
+ *
+ * CodeMods is an ideal place to use Maybe because there are numerous cases where something may or may
+ * not exist.
+ */
```
jdhuntington, line 9: 👍  🎉

PR #13967, opened 2020-07-09
Website icon fix - give icons unique IDs

micahgodbolt:
SVG icons had hard coded ids. made them unique.


apps/fabric-website/src/utilities/svgIcons.tsx
```diff
@@ -13,6 +13,7 @@ export interface ISVGIconProps {
 
 export const AndroidLogo = (props?: ISVGIconProps) => {
   props = props || {};
+  const id = getId('android');
```
ecraig12345, line 16: In a function component you should do `useId` from `@uifabric/react-hooks`

micahgodbolt:
oh...that useId is a hook, and i'm using the function in an exported SVG. I'll revert back to getId

PR #13969, opened 2020-07-09
t-dama/add prop-renaming utility

tmaster628:
#### Description of changes

Added a basic utility to rename props in primitives and objects, with basic support for enums.
Includes a small testing file and introduces a new interface to map old enum values to new ones.
Big props (no pun intended) to Jon for an excellent base / support thru the process!

#### Focus areas to test
More robustness tests will be added soon!


packages/codemods/src/codeMods/tests/utilities/propsUtilities.test.ts
```diff
@@ -0,0 +1,91 @@
...
+        (tag.getAttribute('type') as JsxAttribute)
+          .getFirstChildByKind(SyntaxKind.JsxExpression)
+          .getFirstChildByKind(SyntaxKind.PropertyAccessExpression)
+          .getLastChildByKind(SyntaxKind.Identifier)
+          .getText(),
+      );
+    });
+    /* Icky reverse call here -- we want O(1) removal */
+    oldEnumValues.reverse();
+    renameProp(tags, 'type', 'size', 'SpinnerSize', spinnerMap);
```
joschect, line 79: Since you setup the mock file, you can probably have some expectations about where things will be so that you can more easily go straight to the attribute/thing that you care about and check that!

packages/codemods/src/codeMods/utilities/props.ts
```diff
@@ -0,0 +1,116 @@
...
+          let enumValue = (foundProp as JsxAttribute)
+            .getFirstChildByKind(SyntaxKind.JsxExpression)
+            .getFirstChildByKind(SyntaxKind.PropertyAccessExpression);
+          let firstEnumElem = enumValue.getFirstChildByKind(SyntaxKind.Identifier);
+          let secondEnumElem = enumValue.getLastChildByKind(SyntaxKind.Identifier);
+          firstEnumElem.rename(valueName);
+          secondEnumElem.rename(changeValueMap[secondEnumElem.getText()]);
+        } else {
+          /* If we don't have an enum, but we do have a value, only change the value of the prop to valueName. */
+          foundProp.set({ initializer: valueName });
```
joschect, line 38: I'm not positive, but I suspect that this initializer might not quite work. At the very least, I think that the dev would need to know that they need to put in `FullEnumName.Value` otherwise it will resolve to `foo={<whatever was in valueName`

packages/codemods/src/codeMods/types.ts
```diff
@@ -26,3 +26,7 @@ export interface Codemod<T = SourceFile> {
    */
   enabled?: boolean;
 }
+
+export interface EnumMap {
```
joschect, line 30: Not necessary, but if you have time you could change this to `export type Dictionary<T> = {[key: string]: T}` so that it can be used generically :D

packages/codemods/src/codeMods/utilities/props.ts
tmaster628: You're right -- I think I wasn't explicit enough here. This isn't actually a case for enums, it's actually a case for full-on value change like a bool, where you need to flip the value!

PR #13973, opened 2020-07-09
Adding useSetInterval and useSetTimeout within react-hooks package

czearing:
#### Pull request checklist

#### Description of changes

1. Adding `useSetInterval` and `useSetTimeout` hooks within react-hooks package.
- Added descriptions and examples within the README.md.
- Added unit tests for the hooks.

2. Fix for the test-utility safeMount to check if wrapper should unmount.

#### Focus areas to test
useSetInterval and useSetTimeout


packages/react-hooks/README.md
```diff
@@ -140,3 +140,47 @@ const Example = React.forwardRef(function Example(props:{}, forwardedRef: React.
   return <div>Example</div>;
 })
 ```
+
+## useSetInterval
+
+`function useSetInterval()`
```
ecraig12345, line 146: Add the return type (for useSetTimeout too)

packages/react-hooks/README.md
```diff
@@ -140,3 +140,47 @@ const Example = React.forwardRef(function Example(props:{}, forwardedRef: React.
...
+
+Hook that returns a wrapper function for `setInterval` which automatically handles disposal.
+
+### Example
+
+```jsx
+import { useSetInterval } from '@uifabric/react-hooks';
+
+const MyComponent = buttonContent => {
```
ecraig12345, line 155: For demo purposes it's probably best to just get rid of the props and hard-code a string inside the button (same in useSetTimeout)

packages/react-hooks/src/useSetInterval.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    });
+  });
+
+  it('does not execute the interval when unmounted', () => {
+    safeMount(<TestComponent />, wrapper => {
+      expect(timesCalled).toEqual(0);
+
+      wrapper.unmount();
+
+      jest.runAllTimers;
```
ecraig12345, line 45:
```suggestion
      jest.runAllTimers();
```

packages/react-hooks/src/useSetTimeout.tsx
```diff
@@ -0,0 +1,33 @@
...
+      for (const id of timeoutIds) {
+        clearTimeout(id);
+      }
+      timeoutIds.splice(0);
+    },
+    [],
+  );
+
+  return (func: () => void, duration: number) => {
+    const id = useConst(
```
ecraig12345, line 21: Why the useConst here?

packages/react-hooks/src/useSetTimeout.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      ref.current!.clearTimeout();
+
+      jest.runOnlyPendingTimers();
+
+      expect(timesCalled).toEqual(1);
+    });
+  });
+});
+
+// import * as React from 'react';
```
dzearing, line 100: this doesn't look right...

packages/react-hooks/src/useSetTimeout.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export type UseSetTimeoutReturnType = {
+  setTimeout: (callback: () => void, duration: number) => number;
+  clearTimeout: (id: number) => void;
+};
+
+/**
+ *  Returns a wrapper function for `setTimeout` which automatically handles disposal.
+ */
+export const useSetTimeout = (): UseSetTimeoutReturnType => {
```
dzearing, line 13: My only suggestion would be that the useSetTimeout and useSetInterval are almost identical; it seems you could abstract these into one function:
```tsx
const createUseTimer = (set, clear) => { ... everything above };

export const useSetTimeout = createUseTimer(setTimeout, clearTimeout);
export const useSetInterval = createUseTimeout(setInterval, clearInterval);
```

This would reduce the overall bundle hit and redundant code.

ecraig12345: I like this too, though I'd call it `useTimers` instead.

ecraig12345: Oh never mind I see what you mean. I think I actually prefer to leave it as-is because that feels like a confusing amount of indirection, and the bundle hit and amount of redundant code is pretty small.

dzearing: The whole block is duplicate though. We should try and right DRY code if possible :) Perhaps comments would help to clarify. I am going to guess it would shave off another 500 bytes, also keeps the code condensed so that fixing or optimizing it is done one time. Every bit counts.

dzearing: On second thought, keeping them separate is probably cleaner and not worth the savings. The API surface returned is different and it would be less readable.

packages/react-hooks/README.md
```diff
@@ -140,3 +140,47 @@ const Example = React.forwardRef(function Example(props:{}, forwardedRef: React.
...
+`useSetInterval: () => { setInterval, clearInterval }`
+
+Hook which returns safe `setInterval` and `clearInterval` methods. Intervals set up using this hook will be automatically cleared when the component is unmounted.
+
+### Example
+
+```jsx
+import { useSetInterval } from '@uifabric/react-hooks';
+
+const MyComponent = buttonContent => {
```
ecraig12345, line 155:
```suggestion
const MyComponent =() => {
```

packages/react-hooks/README.md
```diff
@@ -140,3 +140,47 @@ const Example = React.forwardRef(function Example(props:{}, forwardedRef: React.
...
+`const useSetTimeout: () => { setTimeout, clearTimeout }`
+
+Hook which returns safe `setTimeout` and `clearTimeout` methods. Timeout callbacks set up using this hook will be automatically cleared when the component is unmounted.
+
+### Example
+
+```jsx
+import { useSetInterval } from '@uifabric/react-hooks';
+
+const MyComponent = buttonContent => {
```
ecraig12345, line 177:
```suggestion
const MyComponent = () => {
```

PR #13979, opened 2020-07-10
Add multi select delimiter to combobox

### LINKED ISSUE #13978: Add multiSelectDelimiter property to ComboBox

ComboBox does not currently have a property to specify a delimiter for multi-select labels.  The default of ', ' is always used.  It would be useful to be able to specify an alternate delimiter string much like can be done in Dropdown.

ComboBox.

No.

Both ComboBox and Dropdown are being used in a Microsoft Teams + Power Apps integration project and we would like the two controls to be swapped interchangeably depending of if searchability of items is required.  The former component does not currently support alternate multi-select delimiters whereas the latter component does.  We just want to bring ComboBox up to parity with Dropdown in this respect.

image

c0r32:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #13978

#### Description of changes

This change adds a `multiSelectDelimiter` property to ComboBox that is analogous to Dropdown's `multiSelectDelimiter` property.

#### Focus areas to test

ComboBox.

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
@@ -706,14 +706,8 @@ export class ComboBox extends React.Component<IComboBoxProps, IComboBoxState> {
           : this._normalizeToString(suggestedDisplayValue),
       );
     }
-    let displayString = '';
-    for (let idx = 0; idx < displayValues.length; idx++) {
-      if (idx > 0) {
-        displayString += ', ';
-      }
-      displayString += displayValues[idx];
-    }
-    return displayString;
+    const { multiSelectDelimiter = ', ' } = this.props;
+    return displayValues.join(multiSelectDelimiter);
   }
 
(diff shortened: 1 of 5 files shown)


packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -706,14 +706,8 @@ export class ComboBox extends React.Component<IComboBoxProps, IComboBoxState> {
...
-    let displayString = '';
-    for (let idx = 0; idx < displayValues.length; idx++) {
-      if (idx > 0) {
-        displayString += ', ';
-      }
-      displayString += displayValues[idx];
-    }
-    return displayString;
+    const { multiSelectDelimiter = ', ' } = this.props;
+    return displayValues.join(multiSelectDelimiter);
```
c0r32, line 710: This is borrowed, almost verbatim, from Dropdown.base.tsx.

dzearing:
@c0r32 thanks!

PR #13981, opened 2020-07-10
DatePicker: showDatePickerPopup should be public

shrallap:
#### Pull request checklist

#### Description of changes

Make showDatePickerPopup public so that, date callout can be opened.

(give an overview)

#### Focus areas to test

(optional)

### DIFF
--- a/packages/date-time/src/components/DatePicker/DatePicker.base.tsx
+++ b/packages/date-time/src/components/DatePicker/DatePicker.base.tsx
@@ -311,6 +311,15 @@ export class DatePickerBase extends React.Component<IDatePickerProps, IDatePicke
     this.setState(this._getDefaultState());
   }
 
+  private showDatePickerPopup(): void {
+    if (!this.state.isDatePickerShown) {
+      this._preventFocusOpeningPicker = true;
+      this.setState({
+        isDatePickerShown: true,
+      });
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


packages/date-time/src/components/DatePicker/DatePicker.base.tsx
```diff
@@ -311,6 +311,15 @@ export class DatePickerBase extends React.Component<IDatePickerProps, IDatePicke
     this.setState(this._getDefaultState());
   }
 
+  public showDatePickerPopup(): void {
```
dzearing, line 314: Just a heads up: this prop would need to be added to `IDatePicker` and only accessed using `componentRef`, rather than `ref`.

We should not assume that `ref` should be accessing public class based APIs, so please make sure to add this to the imperative interface. In function components, `ref` should return the root element, so converting this to a function component would likely cause some churn on customers unless this is only officially accessible through `componentRef` (yes, for now, ref would also return it, but let's call that unsupported until this is functional.)

React in general tends to avoid imperative api in favor of declarative props. Is there a way to achieve this as a prop?

If you absolutely see no other way to do this, then, yes, add to the imperative api surface.

srhyderabad: I think you are right this needs to be added to IDatePicker

PR #13989, opened 2020-07-10
Website hook fix

micahgodbolt:
Fixes bug in #13967

useId is only valid in react components. our icons are just functions returning jsx elements for the registerIcons function

### DIFF
--- a/apps/fabric-website/src/utilities/svgIcons.tsx
+++ b/apps/fabric-website/src/utilities/svgIcons.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { css, getId } from 'office-ui-fabric-react';
+import { css } from 'office-ui-fabric-react';
 import * as styles from './svgIcons.module.scss';
 
 export interface ISVGIconProps {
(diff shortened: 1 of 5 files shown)


apps/fabric-website/src/utilities/svgIcons.tsx
```diff
@@ -215,7 +214,7 @@ export const WebLogo = (props?: ISVGIconProps) => {
 
 export const CrossPlatformLogo = (props?: ISVGIconProps) => {
   props = props || {};
-  const id = useId('crossplat');
+  const id = getId('crossplat');
```
dzearing, line 217: I'm not sure that makes sense. It make more sense to do any of the following:

1. Change them to be used as components. E.g. change the register process to render them:
```tsx
registerIcons({
  icons: {
     CrossPlatformLogo: <CrossPlatformLogo />
  }
});
```

2. If they truely are just functions, they should be renamed and not capitalized.

micahgodbolt: yeah, i changed them to non caps. turns out we don't need that ID on the offending items (we dont' use the color version of the icon anymore). The homepage icons (color) are different icons and are only used once.

PR #13992, opened 2020-07-10
Calendar: Fix onBlur event firing on every date click

### LINKED ISSUE #13633: Calendar: onBlur callback is fired on every date click
### Environment Information

- **Package version(s)**: 7.120.2 
- **Browser versions**: Chrome Version 83.x (or any browser)
- **OS versions**: Windows 10 (or any OS)

### Codepen example:
https://codepen.io/pragya91/pen/NWxRZJa

#### Actual behavior:

**onBlur** event gets fired whenever a new date is focused. If a user keeps clicking on dates or uses tab to navigate within the Calendar component, then onBlur event gets called.
#### Expected behavior:

Ideally, since the user is still within the bounds of calendar, onBlur even should not get fired.

Requested priority: (Normal)

Products/sites affected: https://fluidpreview.office.net/

tomi-msft:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13633 

#### Description of changes

-Fix onBlur event firing on every date click in Calendar component

**Before**
https://codepen.io/pragya91/pen/NWxRZJa

**After**
-onFocus should be fired when focus comes from outside of the Calendar to inside of the Calendar
-onBlur should be fired when focus leaves from inside of the Calendar to outside of the Calendar

#### Focus areas to test

(optional)


packages/office-ui-fabric-react/src/components/Calendar/Calendar.tsx
```diff
@@ -277,6 +278,10 @@ export class Calendar extends React.Component<ICalendarProps, ICalendarState> im
     }
   }
 
+  private _onBlur = (event: React.FocusEvent<HTMLDivElement>) => {
+    event.stopPropagation();
```
xugao, line 282: you need to call the original props.onBlur here. otherwise anything passed by user from `onBlur` prop is an no-op because you are entirely overwriting it with your own logic here

packages/office-ui-fabric-react/src/components/Calendar/Calendar.tsx
```diff
@@ -84,6 +84,7 @@ export class Calendar extends React.Component<ICalendarProps, ICalendarState> im
 
   private _dayPicker = React.createRef<ICalendarDay>();
   private _monthPicker = React.createRef<ICalendarMonth>();
+  private isFocusInside = false;
```
xugao, line 87:
```suggestion
  private _hasFocus = false;
```
1. pls add `_` prefix for private field
2. i think hasFocus is a simpler name that better described its use?

packages/office-ui-fabric-react/src/components/Calendar/Calendar.tsx
```diff
@@ -277,6 +279,20 @@ export class Calendar extends React.Component<ICalendarProps, ICalendarState> im
     }
   }
 
+  private _onBlur = (event: React.FocusEvent<HTMLElement>) => {
+    if (!event.currentTarget.contains(event.relatedTarget as Node)) {
```
xugao, line 283: element.contains is not supported in IE11. you can use `elementContains` we have in utilities package. and make sure after you update the code you test this in IE11 too :)

PR #13993, opened 2020-07-10
Support for enums in props

tmaster628:
#### Description of changes

Add support for a prop-renaming utility that has the ability to modify simple primitives and objects, including enums.

Includes undefined checks with the new Maybe feature.


packages/codemods/src/codeMods/utilities/props.ts
```diff
@@ -0,0 +1,132 @@
...
+  instances.forEach(val => {
+    /* For each instance, first see if desired prop exists in the open. */
+    let foundProp = val.getAttribute(toRename);
+    if (foundProp) {
+      /* If found, do a simple name-replacementName. */
+      foundProp.set({ name: replacementName });
+      if (changeValueMap) {
+        /* Change the value of the enum via map conversion. */
+        let enumExp = (foundProp as JsxAttribute)
+          .getFirstChildByKind(SyntaxKind.JsxExpression)
```
joschect, line 28: I think that there need to be some more null/undefined checks. The getFirstChildByKind can return undefined I believe.

tmaster628: Good point! I'll fix this first thing tomorrow!

PR #13994, opened 2020-07-11
ESLint: add faster max-len rule

ecraig12345:
When running ESLint with timing data to investigate why it's slower, one outlier I noticed was the `max-len` rule. I looked into it and the reason it's slow is because it supports a lot of different options, and it **unconditionally** does extra calculations related to those options even if they're not being used. It also handles tabs and multi-byte unicode characters when calculating line length (rather than using raw string length), and neither of those is very important to us.

Since this is a pretty simple rule, I made a custom rule `@fluentui/max-len` with the following optimizations:
- The only options are `max` and a list of `ignorePatterns` (ignore lines which match these regexes)
- In the implementation, check the line length FIRST. If it's less than `max`, skip other checks.
- Don't account for tab expansion or multi-byte unicode characters when calculating length (just use string length)

### DIFF
--- a/packages/eslint-plugin/src/configs/react.js
+++ b/packages/eslint-plugin/src/configs/react.js
@@ -47,6 +47,21 @@ const config = {
     '**/*.scss.ts',
   ],
   rules: {
+    '@fluentui/max-len': [
+      'error',
+      {
+        ignorePatterns: [
+          'require(<.*?>)?\\(',
+          'https?:\\/\\/',
+          '^(import|export) \\{ \\w+( as \\w+)? \\} from',
+          '^import \\* as',
+          '^\\s+(<path )?d=',
+          '!raw-loader!',
+          '\\bdata:image/',
+        ],
+        max: 120,
+      },
+    ],
     '@fluentui/no-tslint-comments': 'error',
 
     // tslint: function-name, variable-name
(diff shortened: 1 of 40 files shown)


dzearing:
any way to contribute back to the source?

ecraig12345:
> any way to contribute back to the source?

Looking into it, but I'm not sure how long it will take to get approved (assuming they'd even take it) and published. And even then, by necessity it would still be slower than this version due to the additional options available. So I'd like to check this in for now.

ecraig12345:
Also, thanks to the real rule's requirement of counting multi-byte unicode characters as a single character (which we don't really need to care about), that seriously limits the amount of optimization that can be done.

packages/eslint-plugin/src/configs/react.js
```diff
@@ -203,11 +201,18 @@ const config = {
     'no-restricted-syntax': 'off',
 
     // permanently disable because we disagree with these rules
+    'import/prefer-default-export': 'off',
     'no-await-in-loop': 'off', // contrary to rule docs, awaited things often are NOT parallelizable
     'no-restricted-globals': 'off', // airbnb bans isNaN in favor of Number.isNaN which is unavailable in IE 11
     'react/jsx-props-no-spreading': 'off',
     'react/prop-types': 'off',
 
+    // permanently disable due to redundancy with custom rules
```
xugao, line 210:
```suggestion
    // permanently disable due to performance reason. use custom rule `@fluentui/max-len` instead.
```

ecraig12345:
```suggestion
    // permanently disable due to performance issues (using custom rule `@fluentui/max-len` instead)
```

ecraig12345:
Update re: @dzearing's comment about contributing back, the most significant perf optimization here is the simple string length check before the `ignorePatterns` regex check. Unfortunately due to the real rule's tab and unicode handling, it can't do a simple string length check, which probably removes a lot of the benefit. It might also *de-optimize* the rule for those who don't use `ignorePatterns` by introducing the line length calculation when it wouldn't have previously happened.

PR #14009, opened 2020-07-13
fix: suppress warnings in `createProvider()`

### LINKED ISSUE #13684: Warning: Cannot update a component MenuItem while rendering a different component ContextSelector.Provider
By the stack trace, it seems my own components are not related to the problem. Appears on v0.50, was working before just fine.

Captura de Pantalla 2020-06-18 a la(s) 13 04 12

layershifter:
- [x] Addresses an existing issue: Fixes #13684

#### Description of changes

Ports a fix from https://github.com/dai-shi/use-context-selector/commit/526bea3b43bfc97de966e6125449b46c9a27e17a.


layershifter:
Based on internal agreement we decided to keep existing warning to avoid different code for development and production. 

React will continue to raise this warning, but everything will continue work as it is. Once it will be possible to provide a proper fix, we will do it.

PR #14010, opened 2020-07-13
chore: use own React context

### LINKED ISSUE #13836: Component rendered outside Provider throws
### Environment Information

- **Package version(s)**: @fluentui/react-northstar@0.50.0

https://codesandbox.io/s/button-without-provider-error-gbcyr?file=/src/App.tsx

Render a component out of a `Provider`.

#### Actual behavior:

image

#### Expected behavior:

No error thrown.

layershifter:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13836.

#### Description of changes

Removes usage of `ThemeContext` from `react-fela` and introduces our own context instead.

#### Focus areas to test

(optional)


packages/fluentui/react-bindings/src/index.ts
```diff
@@ -32,3 +32,7 @@ export * from './telemetry/types';
 export { getElementType } from './utils/getElementType';
 export { getUnhandledProps } from './utils/getUnhandledProps';
 export { mergeVariablesOverrides } from './utils/mergeVariablesOverrides';
+
+export * from './context';
```
assuncaocharles, line 36: dup?
```suggestion
```

packages/fluentui/react-bindings/src/hooks/useStyles.ts
```diff
@@ -67,27 +53,11 @@ type InlineStyleProps<StyleProps> = {
...
-  theme: emptyTheme,
-  telemetry: undefined,
-};
-
 export const useStyles = <StyleProps extends PrimitiveProps>(
   displayName: string,
   options: UseStylesOptions<StyleProps>,
 ): UseStylesResult => {
-  const context: StylesContextValue = React.useContext(ThemeContext) || defaultContext;
+  const context = useFluentContext();
```
assuncaocharles, line 60:
```suggestion
  const context = useFluentContext() || defaultContext;
```

?

layershifter: Nope, it's not a case anymore. Previously we had this because we need to have a default value, but it was impossible to define it as `.createContext()` was called in `react-fela`. But now we are defining it under own `.createContext()` call:

```tsx
const FluentContext = React.createContext<ProviderContextPrepared>(defaultContext);
```

PR #14018, opened 2020-07-14
SplitButton: Initial implementation

khmakoto:
#### Pull request checklist

#### Description of changes

This PR contains the first implementation of a `SplitButton` component using our latest stack of technologies.

#### Focus areas to test

(optional)


packages/react-button/src/components/SplitButton/useSplitButton.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const menuButtonState = useMenuButton(menuButtonProps, ref, options);
+  const { expanded, menu, onClick: onMenuButtonClick, onKeyDown, ...rest } = menuButtonState;
+
+  const state = {
+    ...rest,
+    'aria-expanded': expanded,
+    expanded,
+    onKeyDown,
+    tabIndex: 0,
```
jurokapsiar, line 26: what is the role or element type of the root? as far as I can see, we are making a `div` focusable, without any role. That does not sound consistent. Also, onClick seems to be on the button slot inside - Voiceover used to have probems with this, when element with onClick was not directly focused. For virtual cursor, does it mean that there are 3 stops - root, button and menu button?

In v0, the approach we landed on was to make the root just a presentation container, and allowing the focus to go to only `button` and `menuButton` (the slots are called differently). This way we avoided all of the problems above.

khmakoto: I am missing the role, thanks for noticing that out!

Regarding tab/focus stops, there is an ongoing discussion about it being 2 tab stops or 1 tab stop (which honestly just drives me crazy at this point hahaha). @dzearing have you heard anything about a decision on that?

jurokapsiar: I have watched that discussion - as far as I understand, it is leaning towards having one tab stop. but correct me if I am worng :)

PR #14028, opened 2020-07-14
Checkbox: allow focus when disabled

### LINKED ISSUE #13704: [ Check-box-Disabled-Won't Get Focus ] : Focus won't go on Disabled check box, So It is not convey message to  Narrator/NVDA .
Actual Behavior:
In Disabled option checkbox won't get focus. 

Repro: https://developer.microsoft.com/en-us/fabric#/controls/web/checkbox

Repro Step:
Open Edge browser with above link.
Go to checkbox, tab through page try to reach on disabled checkbox.
Now observe that it's difficult to focus on checkbox which is disabled.

Expected 
Focus should go on disabled check box and narrator should speak it is disabled.

Can you please let us know is any property available for this . As on official web i am not able see.

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13704

#### Description of changes
Allow focus when Checkbox is disabled

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-next-2020-07-14-11-43-58-xgao-checkbox-focus.json
+++ b/change/@fluentui-react-next-2020-07-14-11-43-58-xgao-checkbox-focus.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Checkbox: allow focus when disabled.",
+  "packageName": "@fluentui/react-next",
+  "email": "xgao@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-07-14T18:43:58.085Z"
+}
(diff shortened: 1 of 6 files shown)


khmakoto:
Why do we want to allow this?

xugao:
@khmakoto - why not allow? i think it helps convey information to screen reader user similar how it conveys information for sighted users. northstar Checkbox has this behavior and our V7 buttons as well.

xugao:
@khmakoto - chatted with @betrue-final-final . we don't want to sacrifice key-nav efficiency by enabling the focus. will close the PR and document this decision somewhere. we still have the general problem that is components behaves differently across our library :(

PR #14030, opened 2020-07-14
DatePicker: Refactor DatePicker as a function-based component

### LINKED ISSUE #14339: DatePicker: strict mode compliance (UNSAFE_)
DatePicker has an `UNSAFE_componentWillReceiveProps` implementation preventing strict mode compliance.

MLoughry:
#### Pull request checklist

- [x] Fixes #14339

#### Description of changes

As part of the [ongoing effort](https://github.com/microsoft/fluentui/projects/35) to convert components to function components, this change does so for DatePicker.

As part of that work, I exposed the existing utility function `getDatePartHashValue()` so it could be used in a `React.useEffect()` dependency array.

#### Focus areas to test

(optional)


packages/date-time/src/components/DatePicker/DatePicker.base.tsx
```diff
@@ -205,91 +243,263 @@ export class DatePickerBase extends React.Component<IDatePickerProps, IDatePicke
...
               className: css(classNames.icon, iconProps && iconProps.className),
-              onClick: this._onIconClick,
+              onClick: onIconClick,
             }}
-            onKeyDown={this._onTextFieldKeyDown}
-            onFocus={this._onTextFieldFocus}
-            onBlur={this._onTextFieldBlur}
-            onClick={this._onTextFieldClick}
-            onChange={this._onTextFieldChanged}
+            // eslint-disable-next-line react/jsx-no-bind
```
lorejoh12, line 452: any reason to leave these functions as pure inline functions (and therefore need to disable the lint rule on them) rather than convert them to useCallbacks with empty dependency arrays?

MLoughry: I'm been instructed by @xugao not to preemptively optimize with `useCallback` in Fabric code. 🤷‍♂️

xugao: @lorejoh12 - if we need to keep these static in order to avoid re-rendering for the component that receives these, we can `useCallback`. but we don't want to over-memoize by `useCallback` everywhere.

lorejoh12: Sounds reasonable. Seems to me like we shouldn't have a lint rule suggesting that we use it in that case though.

lorejoh12:
Now that the build is working I'm able to test the deploy change. http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/14030/merge/date-time/dist/index.html#/examples/datepicker

all the examples that let you type a custom date work fine on first selection, but are broken from second selection onwards. All the examples that do not (e.g. the first few examples) don't allow you to select a date at all. As usual, probably an issue with the example rather than the refactor...

MLoughry:
@lorejoh12 I fixed the regressions you found, which came about as I was trying to fix test failures. 😝

PR #14031, opened 2020-07-14
fix(FocusZone): fix EventListener leak

### LINKED ISSUE #14025: Resource leak in FocusZone
### Environment Information

- **Package version(s)**: @fluentui/react-northstar@0.50.0

https://codesandbox.io/s/v050-focus-zone-resource-leak-31gpt

#### Repro steps
1. Have a component with two sibling `FocusZone` components:
```jsx
<FocusZone>first</FocusZone>
<FocusZone>second</FocusZone>
```
2. Mount and unmount the component 5 times.
3. Check keydown event listeners on html node.

#### Actual behavior:

After the last unmount, there are 5 keydown event listeners in FocusZone:
image

#### Expected behavior:

After the last unmount, there should be NO event listeners in FocusZone.

Requested priority: (Blocking, High, Normal, Low) **High**

miroslavstastny:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14025 

#### Description of changes

Converted key down event handler in `FocusZone` class from member arrow function to a regular member function to make it equal among class instances to avoid the leak.
Fixed in `react-focus` (v7, v8).
`react-bindings` (v0) was fixed in a separate PR #14044 

#### Focus areas to test

(optional)


packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -462,12 +462,13 @@ export class FocusZone extends React.Component<FocusZoneProps> implements IFocus
 
   /**
    * Handle global tab presses so that we can patch tabindexes on the fly.
+   * HEADS UP: This must not an arrow function in order to be referentially equal among instances for ref counting to work correctly!
```
dzearing, line 465:
```suggestion
   * HEADS UP: This must not be an arrow function in order to be referentially equal among instances for ref counting to work correctly!
```

dzearing:
Was a bit unclear why referential equality helped the memory leak.

miroslavstastny:
> Was a bit unclear why referential equality helped the memory leak.

Please check #14025 for details.

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -462,12 +462,13 @@ export class FocusZone extends React.Component<FocusZoneProps> implements IFocus
...
    */
-  _onKeyDownCapture = (ev: KeyboardEvent) => {
+  _onKeyDownCapture(ev: KeyboardEvent) {
```
layershifter, line 467: I suggest to move out to be a function instead as it does not have any dependencies to `FocusZone` instance. In this case there is no need to have a dedicated comment.

packages/fluentui/react-bindings/test/FocusZone/FocusZone-EventHandler-test.tsx
```diff
@@ -0,0 +1,92 @@
+import * as ReactDOM from 'react-dom';
+import * as React from 'react';
+import { FocusZone } from '@fluentui/react-bindings';
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+function _deepEquals(a: any, b: any) {
```
layershifter, line 6: This looks for me unnecessary complicated to compare event listeners:

```
function compareEventListeners(a: SIGNATURE, b: SIGNATURE) {
  return a.listener === b.listener && _.isEqual(a.options, b.options)
}
```

packages/react-focus/src/components/FocusZone/FocusZone.tsx
```diff
@@ -107,6 +107,18 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
     return _outerZones.size;
   }
 
+  /**
+   * Handle global tab presses so that we can patch tabindexes on the fly.
+   * HEADS UP: This must not an arrow function in order to be referentially equal among instances
```
dzearing, line 112:
```suggestion
   * HEADS UP: This must not be an arrow function in order to be referentially equal among instances
```

dzearing:
Thank you @miroslavstastny !

Do we need to backport this fix to Fabric 6?

dzearing:
CC @khmakoto

khmakoto:
@miroslavstastny when this merges I'll backport this to v6.

KevinTCoughlin:
Thanks @miroslavstastny for the fix. We have seen this leak in heap snapshots of SharePoint Online. Much appreciated 👍.

PR #14032, opened 2020-07-15
Calendar accessibility fixes

### LINKED ISSUE #11508: [Accessibility][DatePicker] Datepicker cannot be expanded by Alt+Down arrow
### Environment Information

- **Package version(s)**: 6.195.1

### Describe the issue:

Our product uses the DatePicker control that also allows for text input. Since the component is both a textfield and calendar callout, the component was identified by our accessibility testers as a combobox, which should be expandable from the textfield using Alt+Down arrow key. 

This was filed as an MAS bug on our product.

https://fabricweb.z5.web.core.windows.net/oufr/7.74.2/#/examples/datepicker

#### Actual behavior:

DatePicker calendar callout can only be opened from the textfield using enter key or click. 

#### Expected behavior:

DatePicker calendar callout can be opened from the textfield using enter key, click, and Alt+Down arrow key. 

Keyboard interaction documentation for textboxes within comboboxes: https://www.w3.org/TR/wai-aria-practices-1.1/#textbox-keyboard-interaction

### LINKED ISSUE #12444: [DatePicker] Background highlight on today's date not visible in high contrast mode in Chromium Edge
### Environment Information

- **Package version(s)**:  Observable in version used in our product (v6.211.2) and current version 7 used in Fabric documentation website/codepen below.
- **Browser and OS versions**: Edge (Chromium)

### Describe the issue:
In Edge (Chromium), using DatePicker in high contrast mode, the today date in the month view is not highlighted as it is when using in legacy Edge. 

### Please provide a reproduction of the issue in a codepen: https://codepen.io/stmil/pen/RwPqqvy
(Use HCM in new Edge and compare against legacy Edge)

#### Actual behavior:

Today date in month picker is not highlighted in high contrast mode in Edge Chromium.

Example screenshots of issue (Today is March 26):

Edge (Legacy):
image

Edge (New):
image

#### Expected behavior:

Today date in month picker is highlighted in high contrast mode in Edge Chromium, the same as in Edge Legacy.

### LINKED ISSUE #13848: Date picker shows current month as a button but not clickable which conveys incorrect info to screenreader users
### Environment Information

- **Package version(s)**: uifabric/date-time@7.10.0
- **Browser and OS versions**: Edge + Narrator, Windows 2004 (19041.329)

### Describe the issue:

On date picker, there is a header showing the current selected month, like "June 2020". This header is implemented by <button> but it is not clickable. When Narrator is on, it reads this as "June 2020 button". This conveys wrong info to users that this button is clickable.

https://codepen.io/yanbxice/pen/BajwqWo?editable=true%3Dhttps%3A%2F%2Ffabricweb.z5.web.core.windows.net%2Foufr%2F7.121.4%2F

#### Actual behavior:

Narrator reads the header as "June 2020 button"

#### Expected behavior:

It should be reads out as "June 2020"

lorejoh12:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11508, Fixes #12444, Fixes #13848 

#### Description of changes

3 accessibility fixes for Calendar and DatePicker.
- for DatePicker, alt+down arrow should open if the span has focus. Added keydown listener case for it
- for button in header of day picker in Calendar, the button does nothing if the month and day pickers are side by side. We mark the tabIndex = -1 already, but if the user navigates to the header using Narrator's functions, it reports the button as clickable. Changing button to a div so that doesn't happen.
- Added all the various high contrast selectors to match Ben's design for Calendar.

#### Focus areas to test

Can test high contrast only in Win10 + Edge, turn on High Contrast settings from windows settings.

### DIFF
--- a/packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
+++ b/packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
@@ -7,6 +7,7 @@ import {
   AnimationStyles,
   IRawStyle,
   keyframes,
+  HighContrastSelector,
 } from '@uifabric/styling';
 import { DateRangeType } from 'office-ui-fabric-react/lib/utilities/dateValues/DateValues';
 import { AnimationDirection } from '../Calendar/Calendar.types';
@@ -104,12 +105,33 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
       fontWeight: FontWeights.regular,
       color: palette.neutralPrimary,
       cursor: 'pointer',
+      position: 'relative',
       selectors: {
+        [HighContrastSelector]: {
+          color: 'WindowText!important',
+          backgroundColor: 'Window!important',
+          zIndex: 0,
+        },
         ['&.' + classNames.hoverStyle]: {
           backgroundColor: palette.neutralLighter,
+          selectors: {
+            [HighContrastSelector]: {
+              zIndex: 3,
+              overflow: 'visible',
+              outline: '1px solid Highlight!important',
+              MsHighContrastAdjust: 'none',
+            },
+          },
         },
         ['&.' + classNames.pressedStyle]: {
           backgroundColor: palette.neutralLight,
+          selectors: {
+            [HighContrastSelector]: {
+              borderColor: 'Highlight!important',
+              color: 'Highlight!important',
+              MsHighContrastAdjust: 'none',
+            },
+          },
         },
       },
     },
@@ -119,6 +141,18 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
         selectors: {
           ['&:hover, &.' + classNames.hoverStyle + ', &.' + classNames.pressedStyle]: {
             backgroundColor: palette.neutralLight + '!important',
+            selectors: {
+              [HighContrastSelector]: {
+                color: 'HighlightText!important',
+                background: 'Highlight!important',
+              },
+            },
+          },
+          [HighContrastSelector]: {
+            background: 'Highlight!important',
+            color: 'HighlightText!important',
+            borderColor: 'Highlight!important',
+            MsHighContrastAdjust: 'none',
           },
         },
       },
@@ -171,6 +205,14 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
       borderRadius: '100%',
       color: palette.white + '!important',
       fontWeight: (FontWeights.semibold + '!important') as 'initial',
+      selectors: {
+        [HighContrastSelector]: {
+          background: 'WindowText!important',
+          color: 'Window!important',
+          borderColor: 'WindowText!important',
+          MsHighContrastAdjust: 'none',
+        },
+      },
     },
     firstTransitionWeek: {
       position: 'absolute',
--- a/packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
+++ b/packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
@@ -108,8 +108,8 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
       position: 'relative',
       selectors: {
         [HighContrastSelector]: {
-          color: 'WindowText!important',
-          backgroundColor: 'Window!important',
+          color: 'WindowText',
+          backgroundColor: 'Window',
           zIndex: 0,
         },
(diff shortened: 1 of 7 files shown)


packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
```diff
@@ -104,12 +105,33 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
       fontWeight: FontWeights.regular,
       color: palette.neutralPrimary,
       cursor: 'pointer',
+      position: 'relative',
       selectors: {
+        [HighContrastSelector]: {
+          color: 'WindowText!important',
+          backgroundColor: 'Window!important',
```
xugao, line 112: we should avoid setting `!important`. we should use minimal specificity to achieve the goal here. this makes very hard for user to override (not sure if they ever need, but still it's not a good practice in general).

lorejoh12:
> Actually, all our fixes are made only on `data-time` package. we should make the same fixes in OUFR too? since that's what they are using/bug is filed against.

Fixed. There was some existing high contrast code in OUFR Calendar, I updated it to work in Edge Chromium.

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
lorejoh12: I removed a few of them. I run into css specificity issues with merge-styles trying to remove the rest. Specifically, I have a classname `dayCell`, which applies a backgroundColor. I have another class `daySelected`, which applies a different background color. Both classes get put on the same <div>.

The color that actually ends up getting applied depends on the order the classNames end up getting written to the generated CSS. If you open the storybook example, and flip from the Calendar to WeeklyDayPicker examples, you'll see that the order gets flipped for the two classes, whereas I would prefer if `daySelected` styles always win. In pure CSS, you can use a specificity hack like defining `.daySelected.daySelected`, but as far as I know you can't do that in mergestyles. So I end up just putting !important on the backgroundColor, and therefore also put !important on the backgroundColor when defining the high contrast settings.

If you know of a way to override the specificity in merge styles defined in our own component, please let me know and I'll try to remove the other !important tags in a future change.

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
```diff
@@ -104,12 +105,46 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
...
+              backgroundColor: 'Window',
+              MsHighContrastAdjust: 'none',
+            },
+          },
+        },
+        ['&.' + classNames.pressedStyle + '.' + classNames.hoverStyle]: {
+          selectors: {
+            [HighContrastSelector]: {
+              color: 'Highlight',
+              backgroundColor: 'Window',
```
xugao, line 143: look like you can remove these styles. it's covered by the block above

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
```diff
@@ -104,12 +105,46 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
...
+          zIndex: 0,
+        },
         ['&.' + classNames.hoverStyle]: {
           backgroundColor: palette.neutralLighter,
+          selectors: {
+            [HighContrastSelector]: {
+              zIndex: 3,
+              overflow: 'visible',
+              color: 'WindowText',
+              backgroundColor: 'Window',
```
xugao, line 122: do you need these 2 lines or we can relay on cascaded styles from above

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
```diff
@@ -104,12 +105,46 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
...
+              outline: '1px solid Highlight',
+              MsHighContrastAdjust: 'none',
```
xugao, line 124: is it possible to move this line from individual blocks to inside
```
 [HighContrastSelector]: {
          color: 'WindowText',
          backgroundColor: 'Window',
          zIndex: 0,
        },
```

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
lorejoh12: can remove color, backgroundColor gets overridden right above this so need to re-override

PR #14048, opened 2020-07-15
Fix useMergedRefs behaves differently from useRef due to returning a static ref

### LINKED ISSUE #14041: @uifabric/react-hooks: useMergedRefs() behaves differently than native refs
### Environment Information

https://codesandbox.io/s/bold-montalcini-gwzgo

1. Open CSB
2. Click a rerender button

An implementation of `useForkRef()` handles referential equality so it behaves like native refs.

#### Actual behavior:

Only `Button BUTTON` was logged.

#### Expected behavior:

`Button BUTTON` and `Button DIV` & `MergedButton div DIV` should be logged.

Requested priority: Normal

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14041

#### Description of changes
`useMergedRef` should return ref that behaves the same as `useRef` (native ref). 
The hook needs to return new ref instance on every hook call / render. Otherwise `ref` that passed down to element won't be called, see repro [here](https://codesandbox.io/s/ref-changes-on-every-render-o7szs?file=/src/App.tsx).

This change basically reverts changes in #13712, as well as removing `useCallback` logic all together.

#### Focus areas to test

(optional)

### DIFF
--- a/change/@uifabric-react-hooks-2020-07-15-14-39-51-xgao-useMergedRef.json
+++ b/change/@uifabric-react-hooks-2020-07-15-14-39-51-xgao-useMergedRef.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fix useMergedRefs behaves differently from useRef due to returning a static ref.",
+  "packageName": "@uifabric/react-hooks",
+  "email": "xgao@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-07-15T21:39:51.885Z"
+}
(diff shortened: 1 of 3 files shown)


MLoughry:
I'm not sure I understand the problem. In the repro, the staticRef is not called because the underlying component/element  doesn't change, so there's no need to invoke the ref callback again.

Moving to a dynamic ref will cause unnecessary re-renders and invocations of ref functions.

xugao:
Thanks @MLoughry! closing down this PR in favor of #14053

PR #14051, opened 2020-07-16
react-flex) Creating new react-flex package

sopranopillow:
#### Pull request checklist

#### Description of changes

Initial commit for react-flex package, currently there's no functionality.


packages/react-flex/README.md
```diff
@@ -0,0 +1,15 @@
+# @fluentui/react-flex
+
+**React Flex components for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
```
ecraig12345, line 3: Can you update this description? Probably something like "Flexbox abstraction for..."

packages/react-flex/src/Flex.tsx
```diff
@@ -0,0 +1,5 @@
+import * as React from 'react';
```
ecraig12345, line 1: If we're following the same layout as the other packages (which I'd recommend), it should be like this:
```
src/
  Flex.ts           // exports * from ./components/flex/index
  index.ts          // exports * from ./Flex
  components/
    Flex/
      Flex.tsx
      Flex.types.ts
      index.ts      // exports * from Flex and Flex.types
```

packages/react-flex/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "module": "commonjs",
+    "jsx": "react",
+    "declaration": true,
+    "sourceMap": true,
+    "experimentalDecorators": true,
+    "importHelpers": true,
+    "noUnusedLocals": true,
+    "forceConsistentCasingInFileNames": true,
+    "strictNullChecks": true,
+    "noImplicitAny": true,
```
ecraig12345, line 15: Should be like this. Would you mind making the change in the templates too? (scripts/create-package/plop-templates-react/tsconfig.json and scripts/create-package/plop-templates-node/tsconfig.json)
```suggestion
    "strict": true,
```

packages/react-flex/tslint.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": ["@uifabric/tslint-rules"],
```
ecraig12345, line 2: Delete this file

packages/react-flex/tsconfig.json
sopranopillow: sure!

PR #14053, opened 2020-07-16
useMergedRefs: mergedRefs should mutate the function if any of the merged refs mutate

### LINKED ISSUE #14041: @uifabric/react-hooks: useMergedRefs() behaves differently than native refs

dzearing:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #14041

#### Description of changes

This change makes `useMergedRefs` mutate only when merging refs that mutate.

By default `useRef` returns immutable refs, so in normal usage using `useRef` or `createRef`, your refs will not mutate. When using functions as refs, they can easily mutate, causing inconsistencies in behaviors.

#### Focus areas to test

Validated in a @layershifter 's codesandbox forked here:
https://codesandbox.io/s/exciting-golick-8o0gj?file=/src/App.tsx

### DIFF
--- a/packages/react-hooks/src/useMergedRefs.ts
+++ b/packages/react-hooks/src/useMergedRefs.ts
@@ -1,4 +1,4 @@
-import { useRef, useCallback, Ref, MutableRefObject } from 'react';
+import { useCallback, Ref, MutableRefObject } from 'react';
 /**
  * React hook to merge multiple React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
  * updates all provided refs
--- a/packages/react-hooks/src/useMergedRefs.ts
+++ b/packages/react-hooks/src/useMergedRefs.ts
@@ -1,21 +1,24 @@
 import { useRef, useCallback, Ref, MutableRefObject } from 'react';
+
 /**
  * React hook to merge multiple React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
  * updates all provided refs
  * @param refs- Refs to collectively update with one ref value.
  */
 export function useMergedRefs<T>(...refs: Ref<T>[]): (instance: T) => void {
-  return useCallback(
-    (value: T) => {
-      for (const ref of refs) {
-        if (typeof ref === 'function') {
-          ref(value);
-        } else if (ref) {
-          // work around the immutability of the React.Ref type
-          ((ref as unknown) as MutableRefObject<T>).current = value;
-        }
+  const state = useRef<(Ref<T> | undefined)[]>();
+
+  // Update refs list.
+  state.current = refs;
+
+  return useCallback((value: T) => {
+    for (const ref of state.current!) {
+      if (typeof ref === 'function') {
+        ref(value);
+      } else if (ref) {
+        // work around the immutability of the React.Ref type
+        ((ref as unknown) as MutableRefObject<T>).current = value;
       }
-    },
-    [...refs],
-  );
+    }
+  }, []);
 }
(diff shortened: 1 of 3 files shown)


packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,24 +1,21 @@
...
+          ref(value);
+        } else if (ref) {
+          // work around the immutability of the React.Ref type
+          ((ref as unknown) as MutableRefObject<T>).current = value;
+        }
       }
-    }
-  }, []);
+    },
+    [...refs],
```
layershifter, line 19: This breaks rules of hooks, https://github.com/facebook/react/blob/fed4ae0247883050aab6d77d9a83f9b6399d9a45/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L737

packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,24 +1,21 @@
...
-        ref(value);
-      } else if (ref) {
-        // work around the immutability of the React.Ref type
-        ((ref as unknown) as MutableRefObject<T>).current = value;
+  return useCallback(
+    (value: T) => {
+      for (const ref of refs) {
+        if (typeof ref === 'function') {
+          ref(value);
+        } else if (ref) {
```
layershifter, line 13: Refs can be also a string: https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs

Under `handleRef` we have a condition for this:
https://github.com/microsoft/fluentui/blob/8247d915e42b98aceeeaaf0061c1ce3773bd17fb/packages/fluentui/react-component-ref/src/utils.ts#L20

MLoughry: The `LegacyRef` type can be a string, but not `Ref`. `forwardRef`, which is the only way to pass refs to function components, does not accept string refs.

layershifter: https://codesandbox.io/s/interesting-cloud-4s7en?file=/src/App.js

Are you sure?

MLoughry: From the console in your link:
```
Warning: A string ref, "button", has been found within a strict mode tree. String refs are a source of potential bugs and should be avoided. We recommend using useRef() or createRef() instead. Learn more about using refs safely here: https://fb.me/react-strict-mode-string-ref
    in div (at App.js:15)
    in App (at src/index.js:9)
    in StrictMode (at src/index.js:8)
```

I don't think we care to support deprecated legacy features in this hook.

packages/react-hooks/src/useMergedRefs.ts
MLoughry: It may violate the ESLint rule logic, but it doesn't violate the intention or spirit of the rule. The alternative is either having N overloads handling N refs, or forcing users to chain `useMergedRefs()`

packages/react-hooks/src/useMergedRefs.ts
layershifter: So, it's a case for class components only and as React transforms string refs to functions internally:

https://github.com/facebook/react/blob/b23ea02be507cc08747d49c1994f283514c5aeea/packages/react-reconciler/src/ReactChildFiber.new.js#L187-L199

---

https://codesandbox.io/s/goofy-glitter-2gu0g?file=/src/App.js

PR #14054, opened 2020-07-16
Move useRefEffect hook to the react-hooks package

behowell:
#### Pull request checklist

#### Description of changes

Move the hook useRefEffect to the react-hooks package and add tests.


change/@fluentui-react-next-2020-07-15-21-26-39-useRefEffect.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Move useRefEffect to hook to the react-hooks package",
```
khmakoto, line 3:
```suggestion
  "comment": "Move useRefEffect hook to the react-hooks package.",
```

change/@uifabric-react-hooks-2020-07-15-21-26-39-useRefEffect.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Add useRefEffect to hook to the react-hooks package",
```
khmakoto, line 3:
```suggestion
  "comment": "Add useRefEffect hook to the react-hooks package.",
```

change/@fluentui-react-next-2020-07-15-21-26-39-useRefEffect.json
behowell: Oops, thanks for catching the typo!

ecraig12345:
Can you also add this to the package readme? Otherwise looks good.

packages/react-hooks/src/useRefEffect.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    wrapper.setProps({});
+
+    expect(ref.current).toBe(prevRef);
+    expect(log).not.toHaveBeenCalled();
+  });
+
+  it('clears ref.current and calls cleanup when the component is unmounted', () => {
+    const prevRef = ref.current;
+
+    wrapper.unmount();
```
xugao, line 56: hmm we shouldn't write individual test this way which depends on the previous test to run. i always expect i can run one test and it should be pass by itself. in your case it's better to combine these into one test. adding comments between the steps will help readability

behowell:
I updated the API for `useRefEffect` to return a single ref callback function, which also has a `.current` member. This simplifies the API and reduces the ability to misuse it by setting an element's ref to the `RefObject` rather than the `RefCallback` (they are now one in the same).

PR #14063, opened 2020-07-16
fix: useFocusRects should resolve window inside useEffect

### LINKED ISSUE #14062: @fluentui/react-button: useFocusRects() uses a wrong ref
### Environment Information

- **Package version(s)**: `@fluentui/react-button@0.7.2`

https://codesandbox.io/s/summer-feather-13lb5

1. Open a CSB
2. Tab to a frame
3. Mention that frame body doesn't have classes for focus rects

image

#### Actual behavior:

No classes on `body`.

#### Expected behavior:

Classes on `body` 👍 

Requested priority: Normal
Products/sites affected: N/A

layershifter:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14062

#### Description of changes

1. `useFocusRects` uses a wrong `ref` in `useButton`:

https://github.com/microsoft/fluentui/blob/8247d915e42b98aceeeaaf0061c1ce3773bd17fb/packages/react-button/src/components/Button/useButton.ts#L21

Previously it passed a `ref` from the second param of `React.forwardRef()` which will be `null` a user will provide an input.

2. `useFocusRects` resolved `win` during a render

https://github.com/microsoft/fluentui/blob/8247d915e42b98aceeeaaf0061c1ce3773bd17fb/packages/utilities/src/useFocusRects.ts#L42-L46

The resolution occured during render when `rootRef.current` will be `null` 🧷 

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-button/src/components/Button/useButton.ts
+++ b/packages/react-button/src/components/Button/useButton.ts
@@ -1,4 +1,4 @@
-import { useImperativeHandle, useRef } from 'react';
+import * as React from 'react';
 import { getStyleFromPropsAndOptions } from '@fluentui/react-theme-provider';
 import { useFocusRects } from '@uifabric/utilities';
 import { ComposePreparedOptions } from '@fluentui/react-compose';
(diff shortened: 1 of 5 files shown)


packages/react-button/src/components/Button/useButton.ts
```diff
@@ -1,4 +1,4 @@
-import { useImperativeHandle, useRef } from 'react';
```
layershifter, line 1: I suggest to avoid such imports as previously `React` was not imported to this scope however it was used in ` React.RefObject`

packages/utilities/src/useFocusRects.ts
```diff
@@ -40,9 +40,9 @@ type AppWindow = (Window & { FabricConfig?: { disableFocusRects?: boolean } }) |
  * @param rootRef - A Ref object. Focus rectangle can be applied on itself and all its children.
  */
 export function useFocusRects(rootRef?: React.RefObject<HTMLElement>): void {
```
layershifter, line 42: This is still not enough safe it will be never reexecuted if `rootRef.current` will change

PR #14072, opened 2020-07-16
CodeMods: Add filtering and improve maybe

joschect:
#### Pull request checklist

#### Description of changes

Adds in filtering by adding in Yargs to help parse the command line and then creating a getFilter function. Also moves some things around to streamline upgrade command.
#### Focus areas to test

(optional)


packages/codemods/src/codeMods/tests/command.test.ts
```diff
@@ -1,47 +1,30 @@
...
-    it('returns the path when called with a single argument', () => {
-      expect(result.path).toEqual('bar/baz');
+    it('has two strings parsed', () => {
+      expect(result.stringFilter.orElse([]).length).toEqual(2);
     });
 
-    it('does not direct the program to exit', () => {
-      expect(result.shouldExit).toBeFalsy();
+    it('first string is one', () => {
+      expect(result.stringFilter.then(v => v[0]).orElse('')).toEqual('one');
```
jdhuntington, line 27: Any reason this is `stringFilter.then` instead of `stringFilter.map` ?

packages/codemods/src/command.ts
```diff
@@ -1,18 +1,41 @@
...
 
 export class CommandParser {
-  constructor(private logger = console.log) {}
-
-  public parseArgs(argv: string[]) {
-    if (argv.slice(2).indexOf('--help') >= 0) {
-      this.logger(['Usage:', [argv[0], argv[1], '<path in project>'].join(' ')].join('\n'));
-
-      return new CommandParserResult(true);
+  public parseArgs(passedIn: string[]): CommandParserResult {
```
jdhuntington, line 14: 👍 having this encapsulated

packages/codemods/src/command.ts
```diff
@@ -1,18 +1,41 @@
-export class CommandParserResult {
-  // path: string = ""; TODO this doesn't work with jest; why?
-  constructor(public shouldExit = false, public path = '') {}
+import yarr from 'yargs';
+import { Maybe, Nothing } from './maybe';
+export interface CommandParserResult {
+  shouldExit?: boolean;
```
jdhuntington, line 4: Recommend making this a `Maybe<ExitReason>`. ExitReason could be `type = 'success' | 'help' | 'badCommand' | 'programFailure'`

packages/codemods/src/command.ts
```diff
@@ -1,18 +1,41 @@
...
-      return new CommandParserResult(true);
+  public parseArgs(passedIn: string[]): CommandParserResult {
+    const parsed = yarr([])
+      .help()
+      .exitProcess(false)
+      .option('sfilter', {
+        alias: 'sf',
+        type: 'string',
+        array: true,
+        description: 'pass a list of strings to filter by mod name. Does substring search',
```
jdhuntington, line 22: Maybe rephrase?
> only includes mods matching `sfilter`

packages/codemods/src/maybe.ts
```diff
@@ -8,8 +8,8 @@
  * not exist.
  */
 export interface MB<T> {
-  then: <N>(fn: (v: T) => N | Maybe<N>) => Maybe<N>;
-  orElse: (mElse: T) => T;
+  then: <N>(fn: (v: NonNullable<T>) => N | Maybe<N>) => Maybe<N>;
```
jdhuntington, line 11: :+1:

packages/codemods/src/command.ts
```diff
@@ -1,18 +1,41 @@
...
+      .parse(passedIn);
+    if (parsed.help) {
+      return { shouldExit: true, regexFilter: Nothing(), stringFilter: Nothing() };
     }
 
-    return new CommandParserResult(false, argv[2]);
+    return {
+      shouldExit: false,
+      stringFilter: Maybe<string[]>(parsed.sfilter),
+      regexFilter: Maybe<string[]>(parsed.rfilter),
```
jdhuntington, line 38: If both `sfilter` and `rfilter` are passed I imagine things might be complicated downstream. Maybe have the command parser do the right thing and just pass a `filter` as a result?

packages/codemods/src/modRunner/modFilter.ts
```diff
@@ -0,0 +1,33 @@
+import { CodeMod } from '../codeMods/types';
+import { Maybe } from '../maybe';
+
+export function getStringFilter(filter: string) {
+  const filterString = filter;
+  return (modString: string) => {
+    return modString === filterString;
+  };
+}
```
jdhuntington, line 9: Can be shortened to
`export const getStringFilter = (f: string) => (modString: string) => f === modString;`

Is that better? Maybe not, definitely denser though.

packages/codemods/src/modRunner/modFilter.ts
```diff
@@ -0,0 +1,33 @@
...
+  let filts: ((modString: string) => boolean)[] = [];
+  filts = filts.concat(filters.stringFilter.then(v => v.map(getStringFilter)).orElse([]));
+  filts = filts.concat(filters.regexFilter.then(v => v.map(getRegexFilter)).orElse([]));
+
+  // Return default filter if no other filters provided
+  if (filts.length === 0) {
+    return (mod: CodeMod<T>) => true;
+  }
+  return (mod: CodeMod<T>) => {
+    return filts.some(v => v(mod.name));
```
jdhuntington, line 31: As a consumer, I would expect this to be `all` instead of `some`.

packages/codemods/src/modRunner/runnerUtilities.ts
```diff
@@ -72,18 +71,18 @@ export function loadMod(path: string, errorCallback: (e: Error) => void): Maybe<
   return Nothing();
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function filterMods(codeMods: CodeMod<any>[], semverRange: Range) {
-  return codeMods.filter(mod => shouldRunMod(mod, semverRange));
+export function getEnabledMods(getPaths = getModsPaths, loadM = loadMod) {
+  return getPaths()
+    .map(pth => {
+      console.log('fetching codeMod at ', pth);
```
jdhuntington, line 77: Long run you'll probably want to inject a logger for use during testing.

packages/codemods/src/modRunner/runnerUtilities.ts
```diff
@@ -72,18 +71,18 @@ export function loadMod(path: string, errorCallback: (e: Error) => void): Maybe<
...
+      return loadM(pth, e => {
+        console.error(e);
```
jdhuntington, line 79: This is a weird signature. Any reason not to throw if fetching fails? Or have an `Either` type instead of a `Maybe`?

packages/codemods/src/codeMods/tests/command.test.ts
joschect: StringFilter is a maybe. Maybe doesn't have a straight map/bind yet.

packages/codemods/src/modRunner/modFilter.ts
joschect: I'll change it to have includeMods and excludeMods and have them be mutually exclusive to clear that up.

packages/codemods/src/command.ts
joschect: Per my other comment I'll address this when I make an either. Not because I'll use either but because I think this should be batched there.

packages/codemods/src/modRunner/runnerUtilities.ts
joschect: I'll add either in another pr along with monads.

packages/codemods/src/command.ts
```diff
@@ -1,18 +1,59 @@
...
+    .help()
+    .exitProcess(false)
+    .option('modNames', {
+      alias: 'n',
+      type: 'string',
+      array: true,
+      description: 'A list of strings of mod names to run',
+    })
+    .option('modNameRegexs', {
+      alias: 'nr',
```
ecraig12345, line 25: I think the alias is typically a single letter

packages/codemods/src/command.ts
```diff
@@ -1,18 +1,59 @@
...
+  return yarr([])
+    .help()
+    .exitProcess(false)
+    .option('modNames', {
+      alias: 'n',
+      type: 'string',
+      array: true,
+      description: 'A list of strings of mod names to run',
+    })
+    .option('modNameRegexs', {
```
ecraig12345, line 24: I'm pretty sure this should be `modNameRegexes`? Or to make it easier to type just `modRegexes`. Or `modPatterns` to get around the issue of how to pluralize "regex"? ("pattern" is commonly used in regex options elsewhere)

packages/codemods/src/maybe.ts
```diff
@@ -21,27 +21,24 @@ const _makeMaybe = <T>(): MB<T> => {
...
 }
 
-export type Maybe<T> = Just<NonNullable<T>> | Nothing<NonNullable<T>>;
+export type Maybe<T> = Just<T> | Nothing<T>;
 
 // Need to use assign so that the object remains correctly bound to the methods.
 export const Nothing = <T>(): Nothing<T> => Object.assign<MB<T>, { just: false }>(_makeMaybe(), { just: false });
 
-export const Just = <T>(value: NonNullable<T>): Just<NonNullable<T>> => {
+export const Just = <T>(value: NonNullable<T>): Just<T> => {
```
ecraig12345, line 37: I wonder if it would be better to call this `Only` since we already have an unrelated thing called `just`? (missed this in the PR where it was added.) Though I guess you could argue that the confusion risk is not super high since the context is different.

joschect: The just property is different than the Just function. I like having Just as both a function and a type. I'd be happy to rename it to something more explicit, but I'll do that in another PR.

joschect: Also the just property is a way of ensuring that the types, Just and Nothing are distinguishable. So that property name would end up changing too. 

It's following the way that Maybe works in some other libraries, but another naming convention would be to rename it Some which follows Rust's option type.

PR #14073, opened 2020-07-16
fix aria labels in WeeklyDayPicker

lorejoh12:
#### Pull request checklist

#### Description of changes

Fix the aria labels in the WeeklyDayPicker component in date-time. The next/previous buttons were using the week aria label, but were populating the month, so you'd end up with weird tooltips like "next week August".

Fixing it to use the week strings if not expanded and the month strings if expanded. Also correctly calculating the range info for next/previous week and populating like "next week July 19 2020 - July 25 2020", and "next month August".

Also exporting the default weekly day picker strings from the index and updating the test case.

#### Focus areas to test

(optional)

### DIFF
--- a/change/@uifabric-date-time-2020-07-16-16-00-27-lorejoh12-fixAriaLabelsWeeklyDayPicker.json
+++ b/change/@uifabric-date-time-2020-07-16-16-00-27-lorejoh12-fixAriaLabelsWeeklyDayPicker.json
@@ -1,6 +1,6 @@
 {
-  "type": "patch",
-  "comment": "fix next and previous month/week aria labels in weekly day picker",
+  "type": "minor",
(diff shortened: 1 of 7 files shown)


change/@uifabric-date-time-2020-07-16-16-00-27-lorejoh12-fixAriaLabelsWeeklyDayPicker.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "fix next and previous month/week aria labels in weekly day picker, export new default strings",
+  "packageName": "@uifabric/date-time",
+  "email": "lorejoh12@gmail.com",
+  "dependentChangeType": "patch",
```
lorejoh12, line 6: does this need to be minor as well?

PR #14074, opened 2020-07-17
Add mark to Calendar component days

lorejoh12:
#### Pull request checklist

#### Description of changes

Adding ability to mark days in the Calendar date-time component using an optional callback. The default mark shows up as a small circle on the day grid itself (there isn't room for much more than that).

Also converting the default inline calendar example to functional component along with the new example, a follow up PR will be needed to convert the remaining.

image

#### Focus areas to test

(optional)


packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
```diff
@@ -241,6 +241,38 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
...
+      height: 4,
+      backgroundColor: palette.neutralSecondary,
+      borderRadius: '100%',
+      bottom: 1,
+      left: 0,
+      right: 0,
+      position: 'absolute',
+      margin: 'auto',
+      selectors: {
+        '[class*=dayIsToday] &': {
```
lorejoh12, line 255: anyone have suggestions of better ways to do this? this is normally where I'd just stick an "!important" on the styles, but this is an alternative way to force specificity when both "dayIsToday" and "dayMarker" classes are present

packages/date-time/src/components/CalendarDayGrid/CalendarGridDayCell.tsx
```diff
@@ -14,6 +14,7 @@ import { ICalendarGridRowProps } from './CalendarGridRow';
 export interface ICalendarGridDayCellProps extends ICalendarGridRowProps {
   day: IDayInfo;
   dayIndex: number;
+  shouldMarkDay: boolean;
```
pompomon, line 17: _nit_

```suggestion
  marked: boolean;
```

or 

```suggestion
  isMarked: boolean;
```

packages/date-time/src/components/CalendarDayGrid/CalendarGridDayCell.tsx
```diff
@@ -201,6 +203,11 @@ export const CalendarGridDayCell: React.FunctionComponent<ICalendarGridDayCellPr
     }
   };
 
+  let ariaLabel = dateTimeFormatter.formatMonthDayYear(day.originalDate, strings);
+  if (shouldMarkDay) {
+    ariaLabel = ariaLabel + ', ' + strings.dayMarkedAriaLabel;
```
pompomon, line 208: Would comma sign make sense in all the localizations? 
Should we maybe just have a space between two labels?

packages/date-time/src/components/CalendarDayGrid/CalendarGridRow.tsx
```diff
@@ -64,7 +66,13 @@ export const CalendarGridRow = (props: ICalendarGridRowProps): JSX.Element => {
         </th>
       )}
       {week.map((day: IDayInfo, dayIndex: number) => (
-        <CalendarGridDayCell {...props} key={day.key} day={day} dayIndex={dayIndex} />
+        <CalendarGridDayCell
+          {...props}
+          key={day.key}
+          day={day}
+          dayIndex={dayIndex}
+          shouldMarkDay={markedDays.filter(markedDay => compareDates(day.originalDate, markedDay)).length > 0}
```
pompomon, line 74:
```suggestion
          shouldMarkDay={markedDays.some(markedDay => compareDates(day.originalDate, markedDay))}
```
_nit_

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.base.tsx
```diff
@@ -349,6 +353,7 @@ export const CalendarDayGridBase = React.forwardRef(
       weekCorners,
       getDayInfosInRangeOfDay,
       getRefsFromDayInfos,
+      markedDays,
```
pompomon, line 356: Please have you considered to add this `shouldMarkDay` prop to `IDay` from `date-time-utilities` and then you won't need to pass anything to CalendarGridRow and have a check in `useWeeks` instead?

lorejoh12: sure I hadn't considered that but I can make that change

packages/date-time/src/components/CalendarDayGrid/CalendarGridDayCell.tsx
lorejoh12: I thought about having the space but having the comma forces narrator to have a pause between the phrases. We have at least one other instance of that in this control (formatMonthDayYear), and I haven't seen any localization issues raised with it yet

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
xugao: you should be defining and using global classname: https://github.com/microsoft/fluentui/blob/88d22c29939759ac38c15333f8efa5d3f2d93fbb/packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts#L15

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
```diff
@@ -241,6 +243,38 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
...
+      height: 4,
+      backgroundColor: palette.neutralSecondary,
+      borderRadius: '100%',
+      bottom: 1,
+      left: 0,
+      right: 0,
+      position: 'absolute',
+      margin: 'auto',
+      selectors: {
+        ['.' + GlobalClassNames.dayIsTodayStyle + ' &']: {
```
xugao, line 257:
```suggestion
        ['.' + classNames.dayIsTodayStyle + ' &']: {
```

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.styles.ts
```diff
@@ -241,6 +243,38 @@ export const styles = (props: ICalendarDayGridStyleProps): ICalendarDayGridStyle
...
+          backgroundColor: palette.white,
+          selectors: {
+            [HighContrastSelector]: {
+              backgroundColor: 'Window',
+            },
+          },
+        },
+        ['.' + GlobalClassNames.daySelectedStyle + ' &']: {
```
xugao, line 265:
```suggestion
        ['.' + classNames.daySelectedStyle + ' &']: {
```

PR #14083, opened 2020-07-17
feat: add @fluentui/react-image

layershifter:
#### Pull request checklist

#### Description of changes

This PR sets up the initial scaffolding for the react-image package.

#### Focus areas to test

(optional)


packages/react-image/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "react-test-renderer": "^16.3.0"
+  },
+  "dependencies": {
+    "@fluentui/react-compose": "^0.12.2",
+    "@fluentui/react-theme-provider": "^0.3.8",
+    "@microsoft/load-themed-styles": "^1.10.26",
+    "@uifabric/react-hooks": "^7.5.3",
+    "@uifabric/set-version": "^7.0.17",
+    "object-fit-images": "^3.2.4",
+    "office-ui-fabric-react": "^7.123.2",
```
ecraig12345, line 55: Remove this

packages/react-image/src/components/Image/useImage.ts
```diff
@@ -0,0 +1,38 @@
+import { getStyleFromPropsAndOptions } from '@fluentui/react-theme-provider';
+import { ComposePreparedOptions } from '@fluentui/react-compose';
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+// @ts-ignore Typings require esModuleInterop
```
ecraig12345, line 4: Would it be possible to fix the typings, or does this use a module structure where that's not possible? (I remember that being an issue with some structures but forget which ones specifically)

packages/react-image/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "module": "commonjs",
+    "jsx": "react",
+    "declaration": true,
+    "sourceMap": true,
+    "experimentalDecorators": true,
+    "importHelpers": true,
+    "noUnusedLocals": true,
+    "forceConsistentCasingInFileNames": true,
+    "strictNullChecks": true,
+    "noImplicitAny": true,
```
ecraig12345, line 15: Can you change to this? (it will be fixed in the template soon too)
```suggestion
    "strict": true,
```

packages/react-image/src/components/Image/ImageBase.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { useImage } from './useImage';
+
+export const ImageBase = compose<'img', ImageProps, ImageProps, {}, {}>(
+  (props, ref, options) => {
+    const { state } = options;
+    const { slots, slotProps } = mergeProps<ImageProps, ImageState, ImageSlots, ImageSlotProps>(state, options);
+
+    const { buttonRef } = state;
+
+    return <slots.root ref={useMergedRefs(ref, buttonRef)} {...slotProps.root} />;
```
khmakoto, line 15: Should this be?

```suggestion
    const { imageRef } = state;

    return <slots.root ref={useMergedRefs(ref, imageRef)} {...slotProps.root} />;
```

packages/react-image/src/components/Image/useImage.ts
```diff
@@ -0,0 +1,38 @@
+import { getStyleFromPropsAndOptions } from '@fluentui/react-theme-provider';
+import { ComposePreparedOptions } from '@fluentui/react-compose';
```
khmakoto, line 2: nit: Can we put these in alphabetical order?

```suggestion
import { ComposePreparedOptions } from '@fluentui/react-compose';
import { getStyleFromPropsAndOptions } from '@fluentui/react-theme-provider';
```

layershifter: Definitely, however I would like to avoid spending time on such things :)

@ecraig12345 Is it possible to enable [import/order](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md)? As I remember it contains even an autofix...

packages/react-image/package.json
layershifter:
```suggestion
```

packages/react-image/src/components/Image/useImage.ts
layershifter:
```ts
import objectFitImages from 'object-fit-images';
```
```
Error:(3, 8) TS1259: Module '"office-ui-fabric-react/node_modules/@types/object-fit-images/index"' can only be default-imported using the 'esModuleInterop' flag
```

**vs.**

```ts
import * as objectFitImages from 'object-fit-images';
```
```
Error:(3, 34) TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.
```

---

- Option 1: enable `esModuleInterop` on this/all packages
- Option 2: fix typings (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/object-fit-images/index.d.ts), but I am familiar enough to do this. Can you please help me there?

packages/react-image/src/components/Image/useImage.ts
ecraig12345: Potentially. I like that the order is configurable. The default of `["builtin", "external", "parent", "sibling", "index"]` mostly seems reasonable to me, though we might want to use `pathGroups` to put monorepo things (`@fluentui/*`, `@uifabric/*`) between actual externals and relative paths.

packages/react-image/src/components/Image/useImage.ts
ecraig12345: Hmm in this case it looks like the published version of `object-fit-images` uses `module.exports =`, which unfortunately means the typings do also need to use `export =`. I'm surprised TS complains about the `import *` in that case though...

PR #14085, opened 2020-07-17
change EventGroup.declare to string literal

jeremy-coleman:
because typescript supports class declarations such as `class X {
  public declare x: number
}` , using declare has the potential to cause problems for ts parsers supporting classes declaring types. Specifically, this causes an issue with the latest version sucrase. Although the responsibility falls on the parser, it seems pertinent to just avoid this. Changing to ["declare"] is an immediate fix, while renaming it to EventGroup.declareEvent is probably a better long term change. See https://github.com/alangpierce/sucrase/issues/545 for reference.

#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/utilities/src/EventGroup.ts
```diff
@@ -307,10 +307,10 @@ export class EventGroup {
     return EventGroup.raise(this._parent, eventName, eventArgs, bubbleEvent);
   }
 
+  //see https://github.com/alangpierce/sucrase/issues/545
   /** Declare an event as being supported by this instance of EventGroup. */
-  public declare(event: string | string[]): void {
+  public ['declare'](event: string | string[]): void {
```
ecraig12345, line 312: I'd recommend adding a short inline explanation for the use of quotes, in addition to the link. Otherwise it's not clear what the link is referring to.

ecraig12345 (approved):

ecraig12345:
@jeremy-coleman Looks like you'll need to sign the CLA before this can merge--you should have gotten a comment about that from the CLA bot when you first opened the PR, but I think it was broken for awhile.

PR #14090, opened 2020-07-17
UnifiedPicker accessibility fixes

elisabethcvs:
#### Pull request checklist

#### Description of changes

Fixing three errors given by the edge accessibility tooling

### DIFF
--- a/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
+++ b/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
@@ -239,7 +239,11 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
                   onClick={_onInputClick}
                   onInputValueChange={_onInputChange}
                   /* eslint-enable react/jsx-no-bind */
-                  aria-activedescendant={isSuggestionsShown ? 'FloatingSuggestionsItemId-' + focusItemIndex : undefined}
+                  aria-activedescendant={
+                    isSuggestionsShown && focusItemIndex >= 0
+                      ? 'FloatingSuggestionsItemId-' + focusItemIndex
+                      : undefined
+                  }
                   aria-owns={isSuggestionsShown ? 'suggestion-list' : undefined}
                   aria-expanded={isSuggestionsShown}
                   aria-haspopup="true"
--- a/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
+++ b/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
@@ -230,30 +230,34 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
               {headerComponent}
               {_renderSelectedItemsList()}
               {_canAddItems() && (
-                <Autofill
-                  {...(inputProps as IInputProps)}
-                  className={css('ms-BasePicker-input', classNames.pickerInput)}
-                  ref={input}
-                  /* eslint-disable react/jsx-no-bind */
-                  onFocus={_onInputFocus}
-                  onClick={_onInputClick}
-                  onInputValueChange={_onInputChange}
-                  /* eslint-enable react/jsx-no-bind */
-                  aria-activedescendant={
-                    isSuggestionsShown && focusItemIndex >= 0
-                      ? 'FloatingSuggestionsItemId-' + focusItemIndex
-                      : undefined
-                  }
+                <div
                   aria-owns={isSuggestionsShown ? 'suggestion-list' : undefined}
                   aria-expanded={isSuggestionsShown}
-                  aria-haspopup="true"
+                  aria-haspopup="listbox"
                   role="combobox"
-                  disabled={false}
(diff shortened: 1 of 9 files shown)


packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -242,9 +239,13 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
                   /* eslint-enable react/jsx-no-bind */
-                  aria-activedescendant={activeDescendant}
-                  aria-owns={isExpanded ? 'suggestion-list' : undefined}
-                  aria-expanded={isExpanded}
+                  aria-activedescendant={
+                    isSuggestionsShown && focusItemIndex >= 0
+                      ? 'FloatingSuggestionsItemId-' + focusItemIndex
+                      : undefined
+                  }
+                  aria-owns={isSuggestionsShown ? 'suggestion-list' : undefined}
```
philipzloh, line 247: Somewhere in here I think we also need `aria-autocomplete='list'`

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -242,9 +239,13 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
+                  aria-expanded={isSuggestionsShown}
                   aria-haspopup="true"
                   role="combobox"
```
philipzloh, line 250: Aria 1.1 guidance is to have `role='combobox'` on a container and then have the typable `<input>` retain its textbox role

e.g., https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -242,9 +239,13 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
-                  aria-owns={isExpanded ? 'suggestion-list' : undefined}
-                  aria-expanded={isExpanded}
+                  aria-activedescendant={
+                    isSuggestionsShown && focusItemIndex >= 0
+                      ? 'FloatingSuggestionsItemId-' + focusItemIndex
+                      : undefined
+                  }
+                  aria-owns={isSuggestionsShown ? 'suggestion-list' : undefined}
+                  aria-expanded={isSuggestionsShown}
                   aria-haspopup="true"
```
philipzloh, line 249: `aria-haspopup="listbox"` is preferred it seems

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
elisabethcvs: Updated. Autofill was setting that value to both, but I believe you're correct and it should be list here

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
elisabethcvs: Fixed in the new iteration

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
elisabethcvs: Updated

nebhatna:
The changes look good to me. I would get another set of eyes on this since I am new to accessibility as well :)

packages/experiments/src/components/UnifiedPicker/__snapshots__/UnifiedPicker.test.tsx.snap
```diff
@@ -57,50 +57,53 @@ exports[`UnifiedPicker renders correctly with no items 1`] = `
...
-            onCompositionEnd={[Function]}
-            onCompositionStart={[Function]}
-            onCompositionUpdate={[Function]}
-            onFocus={[Function]}
-            onInput={[Function]}
-            onKeyDown={[Function]}
-            onPaste={[Function]}
-            role="combobox"
-            value=""
+          <div
```
philipzloh, line 61: curious what this `<div>` is doing here now..

philipzloh: I know it's a test snap, but is it a hint that the code generated poor HTML?

philipzloh: ah, the original had
```
<div role="list">
    <input role="combobox" />
</div>
```
but looks like this change turned it into 
```
<div>
    <div role="list" />
    <div role="combobox">
        <input />
    </div>
</div>
```
probably the empty `<div role="list" />` shouldn't be there...

elisabethcvs: There was an empty div if the list had no items.  Not ideal. Fixed in the next update

PR #14092, opened 2020-07-17
Removing unnecessary link from website footer

dzearing:
Hiding unnecessary link from the site footer, while we understand the right fix.

### DIFF
--- a/apps/fabric-website/src/styles/_base.scss
+++ b/apps/fabric-website/src/styles/_base.scss
@@ -124,4 +124,9 @@
       width: 125px;
     }
   }
+
+  /* Hide the YouTube TOC link being shown incorrectly by the shared footer. */
+  [itemtype="http://schema.org/Organization"] a.c-hyperlink {
+    display: none !important;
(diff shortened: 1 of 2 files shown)


apps/fabric-website/src/styles/_base.scss
```diff
@@ -124,4 +124,9 @@
       width: 125px;
     }
   }
+
+  /* Hide the YouTube TOC link being shown incorrectly by the shared footer. */
+  [itemtype="http://schema.org/Organization"] a.c-hyperlink {
```
ecraig12345, line 129: Not sure this is specific enough--`document.querySelectorAll('[itemtype="http://schema.org/Organization"]')` returns 4 things. Could we hide it by href instead, or is that type of query too expensive (or too specific to current implementation)?

ecraig12345: Rather, it appears to be specific enough now (the full query only returns one thing) but I wonder if it might not stay that way.

Issue #13945, opened 2020-07-08
Dropdowns (incl menu button) may not be accessible within modal/panel on iOS/Mac due to use of aria-modal - spec unclear?


BrendanMcK:
- **Package version(s)**: office-ui-fabric-react@7.121.10
- **Browser and OS versions**:  Current safari on Mac, current safari on iOS, both with VoiceOver

### Describe the issue:

Popups (Dropdown control, Button with attached menu) don't work when in a Modal or Panel on iOS or Mac. Unclear from ARIA docs who's at fault here, so starting here to get input from Fluent team.

### Please provide a reproduction of the issue in a codepen:

Codepen: https://codepen.io/brendanmckeon/full/qBboNqy

Issue is that controls like Modal, Dialog and Panel all use `aria-modal="true"` on their frame element, which is expected. This essentially tells screenreaders to ignore anything outside the modal.

However, popup elements in Fluent are implemented as `div`s in a different part of the DOM, outside the modal. This appears to cause some screenreaders to ignore them, making it impossible to navigate to them, or track focus when navigating through them.

Some browser scenarios tried:
* Windows 10: Edge+Narrator: all works fine. (Edge classic, not Edge chromium tried here.)
* Mac OSX: Safari + VoiceOver: the items outside the modal track fine, but can't access content in the dropdowns inside the modal.
* iOS: Safari + VoiceOver: similar behavior to Mac; dropdowns within the modal can't be accessed.
* Android: Chrome + TalkBack: Tracking works fine - though there's other (unrelated) issues navigating to items in the dropdown

WAI docs are somewhat ambiguous re what should happen here: On the one hand, the doc for [aria-modal](https://www.w3.org/TR/wai-aria-1.1/#aria-modal) say:

>  If focus moves to an element outside the modal element, assistive technologies SHOULD NOT limit navigation to the modal element.

This suggests that VoiceOver is at fault here - at least for not tracking focus to these items. (Though it would still prevent navigating to those items w/o them first receiving focus.)

However, the doc then goes on to say:

> When a modal element is displayed, authors MUST ensure the interface can be controlled using only descendants of the modal element

Is the drop-down part of a Dropdown or menubutton considered to be something that 'controls' the UI? If so, it suggests that a best practice is to ensure that all relevant UI elements are within the aria-modal element.

[...]

PR #14096, opened 2020-07-18
Dialog: VoiceOver / aria-modal fix

smhigley:
#### Pull request checklist

- [ ] Addresses an existing issue

#### Description of changes

Adds `role="document"` to the child of the modal if `aria-modal` is set to true. This fixes a known VoiceOver behavior with modals: https://bugs.webkit.org/show_bug.cgi?id=174667

Related to #13945

### DIFF
--- a/change/office-ui-fabric-react-2020-07-17-16-06-51-dialog-voiceover-bug.json
+++ b/change/office-ui-fabric-react-2020-07-17-16-06-51-dialog-voiceover-bug.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "add document role to child of modal to fix VoiceOver issue",
+  "packageName": "office-ui-fabric-react",
+  "email": "sarah.higley@microsoft.com",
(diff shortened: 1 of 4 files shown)


xugao:
@smhigley - you meant to fix #13945 in this PR correct? if so pls update the PR description accordingly :)
Also for future reference, we normally name the PR title in this format: {component_name}: {short_description}. Our bot will kick in and add the component label too

smhigley:
@xugao ah, my bad with the PR naming. Apparently this is what I get for doing it late on a friday 😅.

It doesn't fully fix #13945, but it does address one of the problems going on. There are roughly two bugs with VoiceOver in that issue: one is the document role thing fixed in this PR that prevents moving the VO cursor normally within a modal dialog (but tabbing would still work).

The second issue is that the dropdowns are outside the aria-modal element, and that I'm still not sure is possible to fix :P. Aria-owns doesn't help, and the only other thing I can think of is managing aria-modal between components somehow. I'm going to poke at that a bit later :)

PR #14097, opened 2020-07-18
Start enabling "rules of hooks" lint rules

ecraig12345:
#### Pull request checklist

#### Description of changes

Start enabling the `react-hooks` "rules of hooks" lint rules.

This PR enables the rules by default and contains fixes (mostly simple/safe ones) for these packages:
- office-ui-fabric-react
- perf-test
- foundation
- react-stylesheets
- tsx-editor

It also disables the rules for the following packages which had more problematic violations that will be fixed ASAP separately:
- react-next (WIP)
- date-time (see #14098)
- lists (see #14098)
- experiments (see #14100)
- react-hooks (see #14099)
- utilities (see #14099)

One common fix in this PR is adding a "missing" dependency on a value which will never actually change, such as those returned by `useConst`, `useConstCallback`, `useId`, or certain other custom hooks. This is because the hooks rules only handle React's built-in hooks, not our custom hooks. I chose to add the unnecessary-but-harmless dependencies rather than disabling the rule because that way the rule will still catch any newly-added missing deps.

### DIFF
--- a/packages/foundation/src/createComponent.tsx
+++ b/packages/foundation/src/createComponent.tsx
@@ -47,7 +47,7 @@ export function createComponent<
   const { factoryOptions = {} } = options;
   const { defaultProp } = factoryOptions;
 
-  const result: React.FunctionComponent<TComponentProps> = (
+  const ResultComponent: React.FunctionComponent<TComponentProps> = (
     componentProps: TComponentProps & IStyleableComponentProps<TViewProps, TTokens, TStyleSet>,
   ) => {
     const settings: ICustomizationProps<TViewProps, TTokens, TStyleSet> = _getCustomizations(
@@ -62,6 +62,7 @@ export function createComponent<
       // Don't assume state will return all props, so spread useState result over component props.
       componentProps = {
         ...componentProps,
+        // eslint-disable-next-line react-hooks/rules-of-hooks
         ...useState(componentProps),
       };
     }
@@ -89,19 +90,19 @@ export function createComponent<
     return view(viewProps);
   };
 
-  result.displayName = options.displayName || view.name;
+  ResultComponent.displayName = options.displayName || view.name;
 
   // If a shorthand prop is defined, create a factory for the component.
   // TODO: This shouldn't be a concern of createComponent.. factoryOptions should just be forwarded.
   //       Need to weigh creating default factories on component creation vs. memoizing them on use in slots.tsx.
   if (defaultProp) {
-    (result as ISlotCreator<TComponentProps, any>).create = createFactory(result, { defaultProp });
+    (ResultComponent as ISlotCreator<TComponentProps, any>).create = createFactory(ResultComponent, { defaultProp });
   }
 
-  assign(result, options.statics);
+  assign(ResultComponent, options.statics);
 
   // Later versions of TypeSript should allow us to merge objects in a type safe way and avoid this cast.
-  return result as React.FunctionComponent<TComponentProps> & TStatics;
+  return ResultComponent as React.FunctionComponent<TComponentProps> & TStatics;
 }
 
 /**
--- a/packages/foundation/src/createComponent.tsx
+++ b/packages/foundation/src/createComponent.tsx
@@ -56,14 +56,13 @@ export function createComponent<
       options.fields,
     );
 
-    const useState = options.state;
+    const stateReducer = options.state;
 
-    if (useState) {
+    if (stateReducer) {
       // Don't assume state will return all props, so spread useState result over component props.
       componentProps = {
         ...componentProps,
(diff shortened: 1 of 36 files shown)


packages/foundation/src/createComponent.tsx
```diff
@@ -47,7 +47,7 @@ export function createComponent<
   const { factoryOptions = {} } = options;
   const { defaultProp } = factoryOptions;
 
-  const result: React.FunctionComponent<TComponentProps> = (
+  const ResultComponent: React.FunctionComponent<TComponentProps> = (
```
ecraig12345, line 50: This fixes a false positive where the rule complains about hooks being called in a function which doesn't appear to be a component or hook (this function is actually a component, it just didn't look like a component before).

packages/office-ui-fabric-react/src/components/HoverCard/HoverCard.doc.tsx
```diff
@@ -20,27 +20,27 @@ export const HoverCardPageProps: IDocPageProps = {
     'https://github.com/microsoft/fluentui/tree/master/packages/office-ui-fabric-react/src/components/HoverCard',
   examples: [
     {
-      title: 'Example 1: Expanding HoverCard wrapping an element',
+      title: 'Expanding HoverCard wrapping an element',
```
ecraig12345, line 23: One of the HoverCard examples needed cleanup, which made me notice that the docs did too...

packages/office-ui-fabric-react/src/components/HoverCard/examples/HoverCard.Target.Example.tsx
```diff
@@ -53,10 +54,15 @@ const onRenderExpandedCard = (item: IExampleItem): JSX.Element => {
     </div>
   );
 };
-const onRenderItemColumn = (item: IExampleItem, index: number, column: IColumn): JSX.Element | React.ReactText => {
+
+/**
+ * Used for the hoverable "key" cell. In this case, the implementation uses hooks to control
+ * open/closed state, so it must be a function component (not just code within a render callback).
+ */
+const KeyCellWithHoverCard: React.FunctionComponent<{ item: IExampleItem }> = ({ item }) => {
```
ecraig12345, line 62: In this case the rule caught an actual issue: `onRenderItemColumn` (which is not a component or a hook) contained hook calls. I think it happened to work okay in this case, but in general that's not recommended since it can lead to hooks being called conditionally or other strange, hard-to-find issues. 

As a fix, I extracted the logic that needs hooks into this sub-component. (@czearing FYI)

packages/office-ui-fabric-react/src/components/KeytipData/useKeytipData.ts
```diff
@@ -15,12 +15,15 @@ export interface IKeytipData {
  */
 export function useKeytipData(options: KeytipDataOptions): IKeytipData {
   const uniqueId = React.useRef<string>();
-  const keytipProps: IKeytipProps | undefined = options.keytipProps
-    ? {
-        disabled: options.disabled,
-        ...options.keytipProps,
-      }
-    : undefined;
+  // TODO: verify with @xugao whether this should ever change
```
ecraig12345, line 18: @xugao I changed this because the rule pointed out that the `useEffect` to register the keytip was missing dependencies. Do you think this is the right approach, or should I just disable the rule in this case?

khmakoto: There's a test that's failing because of this change, so I think the way it was before was the intended behavior.

ecraig12345: Hmm interesting. I'm not sure the previous way was entirely correct either--if there's a different keytip, shouldn't it be re-registered?

xugao: So here was the original logic before the conversion to FC/hook:
https://github.com/microsoft/fluentui/pull/13688/files#diff-ea9c0d7c983c01ae447822e6dd74bc0d

That's why i didn't have any deps here. 
I am a bit more lean towards disabling the rule for now. reason is that in existing user code, `keytipProps` can be passed as inlined prop (which busts memoization). I am nervous that this causes unintended side-effects if we add the deps

ecraig12345: Sounds good. Looking at the update logic, I do wonder if that will need to be handled a bit differently due to something @layershifter said in https://github.com/microsoft/fluentui/pull/14107#discussion_r457259549--I'm guessing the issue he mentioned has to do with concurrent mode but waiting for clarification. But if a fix is needed there, it can be handled in a separate PR.

PR #14098, opened 2020-07-18
Enable "rules of hooks" lint rules in date-time and lists

ecraig12345:
#### Pull request checklist

#### Description of changes

Enable the `eslint-plugin-react-hooks` "rules of hooks" lint rules in `@uifabric/date-time` and `@uifabric/lists`.

Currently includes some outstanding questions for code authors.

### DIFF
--- a/packages/date-time/src/components/Calendar/Calendar.base.tsx
+++ b/packages/date-time/src/components/Calendar/Calendar.base.tsx
@@ -142,7 +142,6 @@ function useFocusLogic({ componentRef }: ICalendarProps, isDayPickerVisible: boo
   const monthPicker = React.useRef<ICalendarMonth>(null);
   const focusOnUpdate = React.useRef(false);
 
-  // TODO: verify fix with @MLoughry
   const focus = React.useCallback(() => {
     if (isDayPickerVisible && dayPicker.current) {
       focusAsync(dayPicker.current);
--- a/packages/date-time/src/components/Calendar/Calendar.base.tsx
+++ b/packages/date-time/src/components/Calendar/Calendar.base.tsx
@@ -142,13 +142,14 @@ function useFocusLogic({ componentRef }: ICalendarProps, isDayPickerVisible: boo
   const monthPicker = React.useRef<ICalendarMonth>(null);
   const focusOnUpdate = React.useRef(false);
 
-  const focus = () => {
+  // TODO: verify fix with @MLoughry
+  const focus = React.useCallback(() => {
     if (isDayPickerVisible && dayPicker.current) {
       focusAsync(dayPicker.current);
     } else if (isMonthPickerVisible && monthPicker.current) {
       focusAsync(monthPicker.current);
     }
-  };
+  }, [isDayPickerVisible, isMonthPickerVisible]);
 
   React.useImperativeHandle(componentRef, () => ({ focus }), [focus]);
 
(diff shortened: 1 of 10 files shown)


packages/lists/src/components/Viewport/Viewport.tsx
```diff
@@ -87,12 +87,18 @@ export const Viewport = (props: IViewportProps): JSX.Element => {
 
       return () => {
         if (scrollContainerRef.current) {
+          // TODO: check with @soettl on what to do in response to this warning:
```
ecraig12345, line 90: @soettl How would you recommend handling this? Just disabling the rule for this area is an option, but if there's a "right" way to fix it that's better.

ecraig12345:
@MLoughry Since you did the function component conversions for date-time, can you take a look at this and verify the changes I made seem correct? There are also a few cases where I wasn't sure what the "right" fix was (called out in code comments).

packages/date-time/src/components/Calendar/Calendar.base.tsx
```diff
@@ -142,13 +142,14 @@ function useFocusLogic({ componentRef }: ICalendarProps, isDayPickerVisible: boo
   const monthPicker = React.useRef<ICalendarMonth>(null);
   const focusOnUpdate = React.useRef(false);
 
-  const focus = () => {
+  // TODO: verify fix with @MLoughry
```
MLoughry, line 145: Looks fine here.

packages/date-time/src/components/CalendarDayGrid/CalendarDayGrid.base.tsx
```diff
@@ -77,7 +77,11 @@ function useWeeks(
     }
 
     return returnValue;
-  }, [props]);
+    // TODO: check with @MLoughry -- this was missing deps on getSetRefCallback and onSelectDate.
+    // It was easy to prevent getSetRefCallback from changing every render, but onSelectDate is
+    // harder (and likely props will change a lot too). Should we ignore the violation, add a dep
+    // on onSelectDate, or fix it "properly"?
+  }, [props, getSetRefCallback]);
```
MLoughry, line 84: We should add `onSelectDate` here, and probably make `onSelectDate` a memo'd callback as well.

ecraig12345: I tried making `onSelectDate` memo'd, but that turns out to require a huge cascade of changes throughout Calendar as well. That seemed likely to be out of scope for a change like this, so I just disabled the rule for this line instead.

I also noticed while looking through the calendar code that some of the hooks (particularly `useDateState`) call state setters while rendering, which seems unsafe, particularly for concurrent mode.

MLoughry: @ecraig12345 Regarding calling state setters during render: https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

> ... you can update the state right during rendering. React will re-run the component with updated state immediately after exiting the first render so it wouldn’t be expensive.

ecraig12345: Oh interesting, that is not what I would have expected from experience with class components.

PR #14099, opened 2020-07-18
Enable "rules of hooks" lint rules in react-hooks and utilities; hook fixes

ecraig12345:
#### Pull request checklist

#### Description of changes

Enable `eslint-plugin-react-hooks` "rules of hooks" lint rules in `@uifabric/react-hooks` and `@uifabric/utilities`.

Make `useForceUpdate` always return the same function instead of mutating every render. This will be important with more exhaustive listings of deps requested by the lint rules.

Make `useSetInterval` and `useSetTimeout` always return the same entire result object (they already returned probably-constant functions with `useCallback`).

Add missing hooks to `@uifabric/react-hooks` readme.

### DIFF
--- a/packages/react-hooks/src/useSetInterval.test.tsx
+++ b/packages/react-hooks/src/useSetInterval.test.tsx
@@ -11,10 +11,13 @@ describe('useSetInterval', () => {
     const { setInterval, clearInterval } = useSetInterval();
     const { current: state } = React.useRef<{ id: number }>({ id: 0 });
 
-    // This should not have a dependency array because `clearInterval` will always be the same
-    React.useImperativeHandle(ref, () => ({
-      clearInterval: () => clearInterval(state.id),
-    }));
+    React.useImperativeHandle(
+      ref,
+      () => ({
+        clearInterval: () => clearInterval(state.id),
+      }),
+      [clearInterval, state],
+    );
 
     state.id = setInterval(() => {
       timesCalled++;
--- a/packages/react-hooks/src/useSetInterval.test.tsx
+++ b/packages/react-hooks/src/useSetInterval.test.tsx
@@ -11,13 +11,10 @@ describe('useSetInterval', () => {
     const { setInterval, clearInterval } = useSetInterval();
     const { current: state } = React.useRef<{ id: number }>({ id: 0 });
 
-    React.useImperativeHandle(
-      ref,
-      () => ({
-        clearInterval: () => clearInterval(state.id),
-      }),
-      [clearInterval],
-    );
+    // This should not have a dependency array because `clearInterval` will always be the same
+    React.useImperativeHandle(ref, () => ({
+      clearInterval: () => clearInterval(state.id),
+    }));
 
     state.id = setInterval(() => {
       timesCalled++;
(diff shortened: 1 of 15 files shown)


packages/utilities/src/styled.tsx
```diff
@@ -105,7 +105,9 @@ export function styled<
 
         return () => Customizations.unobserve(forceUpdate);
       }
-    }, []);
+      // TODO: check with @MLoughry about whether to add dep on
```
ecraig12345, line 108: @MLoughry would appreciate your input on this

packages/react-hooks/src/useSetInterval.test.tsx
```diff
@@ -11,13 +11,10 @@ describe('useSetInterval', () => {
...
     const { current: state } = React.useRef<{ id: number }>({ id: 0 });
 
-    React.useImperativeHandle(
-      ref,
-      () => ({
-        clearInterval: () => clearInterval(state.id),
-      }),
-      [clearInterval],
-    );
+    // This should not have a dependency array because `clearInterval` will always be the same
```
MLoughry, line 14: I can't find confirmation in docs; however, if this acts like `useEffect`, then if you don't pass a dependency array you'll end up re-assigning the ref every render. If you want to do it only once, pass an empty deps array.

packages/react-hooks/src/useSetTimeout.test.tsx
```diff
@@ -11,13 +11,10 @@ describe('useSetTimeout', () => {
...
 
-    React.useImperativeHandle(
-      ref,
-      () => ({
-        clearTimeout: () => clearTimeout(state.id),
-      }),
-      [clearTimeout],
-    );
+    // This should not have a dependency array because `clearInterval` will always be the same
+    React.useImperativeHandle(ref, () => ({
```
MLoughry, line 15: Same comment about passing in an empty deps array instead.

packages/utilities/src/styled.tsx
MLoughry: It shouldn't have any deps, as it should only run on mount/unmount. Deps in `useEffect` control *when* it's run, and thus needn't be exhaustive. Is the rule throwing in this case?

packages/react-hooks/src/useSetInterval.test.tsx
ecraig12345: I think I misinterpreted the plugin's warning message here. Updated to remove the comment and (similar to other places) add a dependency array `[clearInterval, state]` even though neither will change, to avoid disabling the rule.

packages/utilities/src/styled.tsx
ecraig12345: The rule flags *any* time you have a non-exhaustive deps list, because that often (but not always) points to an error in the logic.

ecraig12345: I changed it to disable the rule in this case. The reason I asked though instead of just disabling the rule to begin with was because I wondered if it would actually be appropriate to re-configure if `inCustomizerContext` changed (which I'm not sure if it can due to not being familiar with Customizer).

MLoughry: I don't think it's possible. It's set to true if a parent component is a `<CustomizerContext>`, and I don't think a React component instance can be re-homed under a separate tree.

ecraig12345:
Build failure was from an unrelated v0 screener timeout issue. Merging.

PR #14100, opened 2020-07-18
Enable "rules of hooks" lint rules in experiments

ecraig12345:
#### Pull request checklist

#### Description of changes

Enable `eslint-plugin-react-hooks` "rules of hooks" lint rules in `@uifabric/experiments`.

Most of the issues were in the SelectedItemsList and UnifiedPicker, where I'll need @nebhatna's help verifying the fixes or in some cases figuring out the right fix.

### DIFF
--- a/packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
+++ b/packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { ISelectedItemProps } from '../SelectedItemsList.types';
 import { ItemCanDispatchTrigger, Item } from './ItemTrigger.types';
+import { useBoolean } from '@uifabric/react-hooks';
 
 export type EditingItemComponentProps<T> = {
   item: T;
@@ -19,19 +20,14 @@ export type EditableItemProps<T> = {
 // `extends any` to trick the parser into parsing as a type decl instead of a jsx tag
 export const EditableItem = <T extends any>(editableItemProps: EditableItemProps<T>): Item<T> => {
   return React.memo((selectedItemProps: ISelectedItemProps<T>) => {
-    const [isEditing, setIsEditing] = React.useState(false);
-    const setEditingTrue = React.useCallback(() => {
-      setIsEditing(true);
-    }, [setIsEditing]);
-    const setEditingFalse = React.useCallback(() => {
-      setIsEditing(false);
-    }, [setIsEditing]);
+    const [isEditing, { setTrue: setEditingTrue, setFalse: setEditingFalse }] = useBoolean(false);
+
     const onItemEdited = React.useCallback(
       (_oldItem: T, newItem: T) => {
         selectedItemProps.onItemChange && selectedItemProps.onItemChange(newItem, selectedItemProps.index);
-        setIsEditing(false);
+        setEditingFalse;
       },
-      [selectedItemProps.onItemChange],
+      [selectedItemProps, setEditingFalse],
     );
 
     const ItemComponent = editableItemProps.itemComponent;
--- a/packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
+++ b/packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
@@ -20,14 +20,15 @@ export type EditableItemProps<T> = {
 // `extends any` to trick the parser into parsing as a type decl instead of a jsx tag
 export const EditableItem = <T extends any>(editableItemProps: EditableItemProps<T>): Item<T> => {
   return React.memo((selectedItemProps: ISelectedItemProps<T>) => {
+    const { onItemChange, index } = selectedItemProps;
     const [isEditing, { setTrue: setEditingTrue, setFalse: setEditingFalse }] = useBoolean(false);
 
     const onItemEdited = React.useCallback(
       (_oldItem: T, newItem: T) => {
-        selectedItemProps.onItemChange && selectedItemProps.onItemChange(newItem, selectedItemProps.index);
+        onItemChange && onItemChange(newItem, index);
         setEditingFalse;
       },
-      [selectedItemProps, setEditingFalse],
+      [onItemChange, index, setEditingFalse],
     );
 
     const ItemComponent = editableItemProps.itemComponent;
--- a/packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
+++ b/packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
@@ -26,7 +26,7 @@ export const EditableItem = <T extends any>(editableItemProps: EditableItemProps
     const onItemEdited = React.useCallback(
       (_oldItem: T, newItem: T) => {
         onItemChange && onItemChange(newItem, index);
-        setEditingFalse;
+        setEditingFalse();
       },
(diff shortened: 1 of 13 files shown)


packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
```diff
@@ -19,19 +20,15 @@ export type EditableItemProps<T> = {
...
   return React.memo((selectedItemProps: ISelectedItemProps<T>) => {
-    const [isEditing, setIsEditing] = React.useState(false);
-    const setEditingTrue = React.useCallback(() => {
-      setIsEditing(true);
-    }, [setIsEditing]);
-    const setEditingFalse = React.useCallback(() => {
-      setIsEditing(false);
-    }, [setIsEditing]);
+    const { onItemChange, index } = selectedItemProps;
+    const [isEditing, { setTrue: setEditingTrue, setFalse: setEditingFalse }] = useBoolean(false);
```
ecraig12345, line 24: The change to `useBoolean` will have the exact same behavior as before, with less verbosity (it was made for cases like this)

packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
```diff
@@ -19,19 +20,15 @@ export type EditableItemProps<T> = {
...
+
     const onItemEdited = React.useCallback(
       (_oldItem: T, newItem: T) => {
-        selectedItemProps.onItemChange && selectedItemProps.onItemChange(newItem, selectedItemProps.index);
-        setIsEditing(false);
+        onItemChange && onItemChange(newItem, index);
+        setEditingFalse;
       },
-      [selectedItemProps.onItemChange],
+      [onItemChange, index, setEditingFalse],
```
ecraig12345, line 31: I'm guessing adding a dependency on `selectedItemProps.index` here will be safe but I can revert if desired.

packages/experiments/src/components/SelectedItemsList/Items/TriggerOnContextMenu.tsx
```diff
@@ -5,13 +5,15 @@ import { ISelectedItemProps } from '../SelectedItemsList.types';
...
 export const TriggerOnContextMenu = <T extends any>(ItemComponent: React.ComponentType<ISelectedItemProps<T>>) => {
   return (props: TriggerProps<T>) => {
+    const { onTrigger } = props;
+    // eslint-disable-next-line react-hooks/rules-of-hooks -- this is a component
     const trigger = React.useCallback(
       e => {
         e.preventDefault();
         e.stopPropagation();
-        props.onTrigger && props.onTrigger();
+        onTrigger?.();
```
ecraig12345, line 14: Same behavior as before but achieved with new optional chaining syntax https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
```diff
@@ -5,11 +5,11 @@ const _SelectedItemsList = <TItem extends BaseSelectedItem>(
   props: ISelectedItemsListProps<TItem>,
   ref: React.Ref<ISelectedItemsList<TItem>>,
 ) => {
-  const [items, updateItems] = React.useState(props.selectedItems || props.defaultSelectedItems || []);
+  const [items, setItems] = React.useState(props.selectedItems || props.defaultSelectedItems || []);
```
ecraig12345, line 8: Renamed this because the non-standard callback name confused the lint rule

packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
```diff
@@ -19,19 +20,15 @@ export type EditableItemProps<T> = {
...
-    }, [setIsEditing]);
+    const { onItemChange, index } = selectedItemProps;
+    const [isEditing, { setTrue: setEditingTrue, setFalse: setEditingFalse }] = useBoolean(false);
+
     const onItemEdited = React.useCallback(
       (_oldItem: T, newItem: T) => {
-        selectedItemProps.onItemChange && selectedItemProps.onItemChange(newItem, selectedItemProps.index);
-        setIsEditing(false);
+        onItemChange && onItemChange(newItem, index);
+        setEditingFalse;
```
khmakoto, line 29:
```suggestion
        setEditingFalse();
```

packages/experiments/src/components/UnifiedPicker/UnifiedPeoplePicker/UnifiedPeoplePicker.tsx
```diff
@@ -11,43 +10,29 @@ import {
 import { UnifiedPicker } from '../UnifiedPicker';
 
 export const UnifiedPeoplePicker = (props: IUnifiedPeoplePickerProps): JSX.Element => {
-  // update the suggestion like componentWillReceiveProps
```
ecraig12345, line 14: From looking at this code, `peopleSuggestions` and `peopleSelectedItems` were **only** being used to trigger generation of new render callbacks. The same thing can be achieved in a simpler way by just making the callbacks directly dependent on those props.

packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
ecraig12345: Oops thank you... 😬

ecraig12345:
Waiting for verification from @nebhatna that the picker changes are okay.

packages/experiments/src/components/SelectedItemsList/Items/TriggerOnContextMenu.tsx
nebhatna: nice!

packages/experiments/src/components/SelectedItemsList/Items/EditableItem.tsx
nebhatna: Just trying to understand the need for adding index here?

ecraig12345: The lint rule wants you to exhaustively list all the values the callback depends on, so that a new callback can be generated if one of those values changes. Forgetting to list a dependency value (and therefore keeping a callback with stale values captured from its closure) is a common source of bugs.

The downside of this is that if any of those dependency values change frequently (like if you take callbacks as props, and the component that provides those callbacks re-creates them every render due to not using `useCallback`/similar), your callback also ends up changing frequently, which can cause a chain of effects down the line.

In this case, I'm guessing `index` won't change very frequently, but guessing you'd have a better idea of whether that's actually true based on how it's used in practice. ("Very frequently" only becomes a real problem if it's something that can happen many times per second.)

nebhatna: Got it, thanks for the clarification.. index does not change that frequently so should be okay to list here. Sounds good. I am curious then if we need to add setEditingTrue here as well?

ecraig12345: Not in this case, since it's not used in that particular callback.

nebhatna: oh yes you are right , my bad. :)

PR #14105, opened 2020-07-20
fix: useMergedRefs() to accept only two params

### LINKED ISSUE #14104: @uifabric/react-hooks: useMergedRefs() violates rules of hooks and produces a warning
### Environment Information

- **Package version(s)**: @uifabric/react-hooks@7.5.4

https://codesandbox.io/s/festive-raman-g8e8g

`useMergedRefs()` allows to accept random amount of refs i.e. nothing restricts me from passing on first render two params and on second three:

```tsx
const refs = [ref, mayBeRef ? inputRef : undefined].filter(Boolean)

const mergedRef = useMergedRefs(...refs)
```

- violates rules of hooks (can be catched by `eslint-plugin-react-hooks`)
https://github.com/microsoft/fluentui/blob/561bb587ebba766eb0cec41533aaf9ebd576e150/packages/react-hooks/src/useMergedRefs.ts#L19
```
React Hook React.useCallback has a spread element in its dependency array. This means we can't statically verify whether you've passed the correct dependencies. (react-hooks/exhaustive-deps)
```

- it produces a React warning
```
index.js:27 Warning: The final argument passed to useCallback changed size between renders. The order and size of this array must remain constant.

Previous: [[object Object], [object Object]]
Incoming: [[object Object], [object Object], [object Object]]
    in App (at src/index.js:9)
    in StrictMode (at src/index.js:8)
```

The similar `useForkRef()` from Material UI takes only two params:
https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/utils/useForkRef.js

#### Actual behavior:

No warnings

#### Expected behavior:

React warnings, unpredictable behavior

Requested priority: Normal
Products/sites affected: N/A

layershifter:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14104

#### Description of changes

See the original issue for more details. This PR restricts params for `useMergedRefs()` hook to accept only two params as otherwise there is no obvious way to handle dependencies for `useCallback()`.

Technically as it's a public API changes there are **breaking**.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-hooks/src/useMergedRefs.ts
+++ b/packages/react-hooks/src/useMergedRefs.ts
@@ -1,13 +1,15 @@
 import { useCallback, Ref, MutableRefObject } from 'react';
+
 /**
  * React hook to merge multiple React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
  * updates all provided refs
- * @param refs- Refs to collectively update with one ref value.
+ * @param refA- A first ref to collectively update with one ref value.
+ * @param refB- A second ref to collectively update with one ref value.
  */
-export function useMergedRefs<T>(...refs: Ref<T>[]): (instance: T) => void {
+export function useMergedRefs<T>(refA: Ref<T>, refB: Ref<T>): (instance: T) => void {
   return useCallback(
     (value: T) => {
-      for (const ref of refs) {
+      for (const ref of [refA, refB]) {
         if (typeof ref === 'function') {
           ref(value);
         } else if (ref) {
@@ -16,6 +18,6 @@ export function useMergedRefs<T>(...refs: Ref<T>[]): (instance: T) => void {
         }
       }
     },
-    [...refs],
+    [refA, refB],
   );
 }
--- a/change/@uifabric-react-hooks-2020-07-20-11-43-15-fix-merged-refs.json
+++ b/change/@uifabric-react-hooks-2020-07-20-11-43-15-fix-merged-refs.json
@@ -1,0 +1,8 @@
+{
+  "type": "major",
+  "comment": "fix: useMergedRefs() to accept only two params",
+  "packageName": "@uifabric/react-hooks",
+  "email": "olfedias@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-07-20T09:43:15.410Z"
+}
--- a/packages/react-hooks/etc/react-hooks.api.md
+++ b/packages/react-hooks/etc/react-hooks.api.md
@@ -46,7 +46,7 @@ export function useForceUpdate(): () => void;
 export function useId(prefix?: string, providedId?: string): string;
 
 // @public
-export function useMergedRefs<T>(...refs: Ref<T>[]): (instance: T) => void;
+export function useMergedRefs<T>(refA: Ref<T>, refB: Ref<T>): (instance: T) => void;
 
 // @public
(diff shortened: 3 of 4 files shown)


packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,13 +1,15 @@
...
- * @param refs- Refs to collectively update with one ref value.
+ * @param refA- A first ref to collectively update with one ref value.
+ * @param refB- A second ref to collectively update with one ref value.
  */
-export function useMergedRefs<T>(...refs: Ref<T>[]): (instance: T) => void {
+export function useMergedRefs<T>(refA: Ref<T>, refB: Ref<T>): (instance: T) => void {
   return useCallback(
     (value: T) => {
-      for (const ref of refs) {
+      for (const ref of [refA, refB]) {
```
layershifter, line 12: If there will be an agreement with change for me it makes sense to eliminate this loop

ecraig12345:
Not sure what you mean by this? 
> otherwise there is no obvious way to handle dependencies for useCallback().

cc @MLoughry since I think he added this hook.

MLoughry:
I really don't think this change is necessary. Spreading the rest parameter of the function as `useCallback` deps may trip up the ESLint rule, but it's not wrong semantically. It's only a "problem" if someone is spreading arguments in a call to `useMergedRefs()`.

Restricting it to two parameters removes flexibility.

layershifter:
> Spreading the rest parameter of the function as `useCallback` deps may trip up the ESLint rule, but it's not wrong semantically.

Sorry, but I can't agree with this. Spreading may lead to a different number of dependencies and then to a React error.

> It's only a "problem" if someone is spreading dynamic arguments in a call to `useMergedRefs()`.

If there an option to use something someone will definitely will try to do it. That's why as API designers we should avoid error prone APIs. The approach taken by other libraries is representative.

> Restricting it to two parameters removes flexibility.

Totally agree, but at the same time there is no usages of this flexibility in actual code. "Safe is always better than flexible".

/cc @levithomason

layershifter:
> Not sure what you mean by this?
> 
> > otherwise there is no obvious way to handle dependencies for useCallback().

Sorry, it was definitely a bit unclear.

I meant that I don't know a simple way to handle dynamic number of dependencies for `useCallback()` / `useMemo()`, i.e. for a current case:

```tsx
useCallback(() => {}, [...refs]);
```

---

As number of dependencies is unclear, to support such scenario we will need to write a custom hook (any other ideas?):

```tsx
const dependencyId = useDependencies(rest);

useCallback(() => {}, [dependencyId]);
```

ecraig12345:
Instead of restricting the number of refs, could we add validation that the number of passed in refs doesn't change?
```ts
const refCount = useConst(refs.length);
if (refs.length !== refCount) {
  throw new Error(...); // or something
}
```

packages/react-hooks/src/useMergedRefs.ts
xugao: The change LGTM

ecraig12345:
@layershifter What do you think of the idea of runtime validation that the number of merged refs doesn't change?

ecraig12345:
Also there's a merge conflict now.

layershifter:
> @layershifter What do you think of the idea of runtime validation that the number of merged refs doesn't change?

It adds more complexity rather than solves the issue.
Closing, feel free to propose another fix for #14104.

PR #14107, opened 2020-07-20
chore: enable eslint-plugin-react-hooks

layershifter:
#### Pull request checklist

#### Description of changes

Has changes from #14105.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/foundation/src/createComponent.tsx
+++ b/packages/foundation/src/createComponent.tsx
@@ -52,17 +52,18 @@ export function createComponent<
   ) => {
     const settings: ICustomizationProps<TViewProps, TTokens, TStyleSet> = _getCustomizations(
       options.displayName,
+      // eslint-disable-next-line react-hooks/rules-of-hooks
       React.useContext(CustomizerContext),
       options.fields,
     );
 
-    const useState = options.state;
+    const stateReducer = options.state;
 
-    if (useState) {
+    if (stateReducer) {
       // Don't assume state will return all props, so spread useState result over component props.
       componentProps = {
         ...componentProps,
-        ...useState(componentProps),
+        ...stateReducer(componentProps),
       };
     }
 
--- a/packages/foundation/src/next/composed.tsx
+++ b/packages/foundation/src/next/composed.tsx
@@ -158,17 +158,18 @@ export function composed<
   ) => {
     const settings: ICustomizationProps<TViewProps, TTokens, TStyleSet> = _getCustomizations(
       options.displayName,
+      // eslint-disable-next-line react-hooks/rules-of-hooks
       React.useContext(CustomizerContext),
       options.fields,
     );
 
-    const useState = options.state;
+    const stateReducer = options.state;
 
-    if (useState) {
+    if (stateReducer) {
       // Don't assume state will return all props, so spread useState result over component props.
       componentProps = {
         ...componentProps,
-        ...useState(componentProps),
+        ...stateReducer(componentProps),
       };
     }
 
--- a/packages/react-hooks/src/useForceUpdate.test.tsx
+++ b/packages/react-hooks/src/useForceUpdate.test.tsx
@@ -7,7 +7,7 @@ describe('useForceUpdate', () => {
     let renderCount = 0;
     const TestComponent: React.FunctionComponent = () => {
       const forceUpdate = useForceUpdate();
-      React.useEffect(() => forceUpdate(), []);
+      React.useEffect(() => forceUpdate(), [forceUpdate]);
 
       renderCount++;
       return <>Test Component</>;
--- a/packages/react-hooks/src/useMergedRefs.ts
+++ b/packages/react-hooks/src/useMergedRefs.ts
@@ -1,13 +1,15 @@
 import { useCallback, Ref, MutableRefObject } from 'react';
+
 /**
  * React hook to merge multiple React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
  * updates all provided refs
- * @param refs- Refs to collectively update with one ref value.
+ * @param refA- A first ref to collectively update with one ref value.
+ * @param refB- A second ref to collectively update with one ref value.
  */
-export function useMergedRefs<T>(...refs: Ref<T>[]): (instance: T) => void {
+export function useMergedRefs<T>(refA: Ref<T>, refB: Ref<T>): (instance: T) => void {
   return useCallback(
     (value: T) => {
-      for (const ref of refs) {
+      for (const ref of [refA, refB]) {
         if (typeof ref === 'function') {
           ref(value);
         } else if (ref) {
@@ -16,6 +18,6 @@ export function useMergedRefs<T>(...refs: Ref<T>[]): (instance: T) => void {
         }
       }
     },
-    [...refs],
+    [refA, refB],
   );
 }
--- a/packages/react-hooks/src/useOnEvent.ts
+++ b/packages/react-hooks/src/useOnEvent.ts
@@ -16,17 +16,19 @@ export function useOnEvent<TElement extends Element, TEvent extends Event>(
 ) {
   // Use a ref for the callback to prevent repeatedly attaching/unattaching callbacks that are unstable across renders
   const callbackRef = React.useRef(callback);
-  callbackRef.current = callback;
 
   React.useEffect(() => {
-    if (element && 'current' in element) {
-      element = element.current;
-    }
-    if (!element) {
+    callbackRef.current = callback;
+  });
+
+  React.useEffect(() => {
+    const actualElement = element && 'current' in element ? element.current : element;
+
+    if (!actualElement) {
       return;
     }
 
-    const dispose = on(element, eventName, (ev: TEvent) => callbackRef.current(ev), useCapture);
+    const dispose = on(actualElement, eventName, (ev: TEvent) => callbackRef.current(ev), useCapture);
     return dispose;
   }, [element, eventName, useCapture]);
 }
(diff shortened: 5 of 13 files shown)


packages/foundation/src/createComponent.tsx
```diff
@@ -52,17 +52,18 @@ export function createComponent<
   ) => {
     const settings: ICustomizationProps<TViewProps, TTokens, TStyleSet> = _getCustomizations(
       options.displayName,
+      // eslint-disable-next-line react-hooks/rules-of-hooks
       React.useContext(CustomizerContext),
       options.fields,
     );
 
-    const useState = options.state;
+    const stateReducer = options.state;
```
layershifter, line 60: Renamed to avoid name collision with `useState` hook.

packages/foundation/src/next/composed.tsx
```diff
@@ -158,17 +158,18 @@ export function composed<
...
+      // eslint-disable-next-line react-hooks/rules-of-hooks
       React.useContext(CustomizerContext),
       options.fields,
     );
 
-    const useState = options.state;
+    const stateReducer = options.state;
 
-    if (useState) {
+    if (stateReducer) {
```
layershifter, line 168: Renamed to avoid name collision with `useState` hook.

packages/foundation/src/createComponent.tsx
```diff
@@ -52,17 +52,18 @@ export function createComponent<
   ) => {
     const settings: ICustomizationProps<TViewProps, TTokens, TStyleSet> = _getCustomizations(
       options.displayName,
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
layershifter, line 55: I don't think that is another way to fix this as the plugin raises a warning because `useContext()` is used outside of a real component

packages/foundation/src/next/composed.tsx
```diff
@@ -158,17 +158,18 @@ export function composed<
   ) => {
     const settings: ICustomizationProps<TViewProps, TTokens, TStyleSet> = _getCustomizations(
       options.displayName,
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
layershifter, line 161: https://github.com/microsoft/fluentui/pull/14107/files#r457258228

packages/react-hooks/src/useForceUpdate.test.tsx
```diff
@@ -7,7 +7,7 @@ describe('useForceUpdate', () => {
     let renderCount = 0;
     const TestComponent: React.FunctionComponent = () => {
       const forceUpdate = useForceUpdate();
-      React.useEffect(() => forceUpdate(), []);
+      React.useEffect(() => forceUpdate(), [forceUpdate]);
```
layershifter, line 10: Another option is to silence this warning as obviously `forceUpdate` is a stable function

packages/react-hooks/src/useOnEvent.ts
```diff
@@ -16,17 +16,19 @@ export function useOnEvent<TElement extends Element, TEvent extends Event>(
...
   // Use a ref for the callback to prevent repeatedly attaching/unattaching callbacks that are unstable across renders
   const callbackRef = React.useRef(callback);
-  callbackRef.current = callback;
 
   React.useEffect(() => {
-    if (element && 'current' in element) {
-      element = element.current;
-    }
-    if (!element) {
+    callbackRef.current = callback;
```
layershifter, line 21: Oops, real issue: it's unsafe to write to `.current` during render. See React's examples: https://reactjs.org/docs/hooks-faq.html#how-to-read-an-often-changing-value-from-usecallback

packages/react-hooks/src/useSetInterval.tsx
```diff
@@ -20,21 +20,27 @@ export const useSetInterval = (): UseSetIntervalReturnType => {
...
-    clearInterval: React.useCallback((id: number): void => {
-      delete intervalIds[id];
-      clearInterval(id);
-    }, []),
+    clearInterval: React.useCallback(
+      (id: number): void => {
+        delete intervalIds[id];
+        clearInterval(id);
+      },
+      [intervalIds],
```
layershifter, line 43: `intervalIds` are stable so the another option is to disable rules for these lines

packages/react-hooks/src/useSetTimeout.tsx
```diff
@@ -21,22 +21,28 @@ export const useSetTimeout = (): UseSetTimeoutReturnType => {
...
-    clearTimeout: React.useCallback((id: number): void => {
-      delete timeoutIds[id];
-      clearTimeout(id);
-    }, []),
+    clearTimeout: React.useCallback(
+      (id: number): void => {
+        delete timeoutIds[id];
+        clearTimeout(id);
+      },
+      [timeoutIds],
```
layershifter, line 45: `timeoutIds` are stable so the another option is to disable rules for these lines

packages/react-stylesheets/src/StylesheetProvider.tsx
```diff
@@ -17,6 +17,7 @@ export const StylesheetProvider = (props: React.PropsWithChildren<StylesheetProv
     }),
     // Only recompute the context to pass down if the parent passes a new one. Props should not
     // be mutating dynamically for a provider, or may be doing so accidentally. Avoid recomputations.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
layershifter, line 20: I disabled a rule there, however I don't that `useMemo()` does the right job there

packages/utilities/src/styled.tsx
```diff
@@ -105,7 +105,7 @@ export function styled<
 
         return () => Customizations.unobserve(forceUpdate);
       }
-    }, []);
+    }, [context.customizations.inCustomizerContext, forceUpdate]);
```
layershifter, line 108: Not sure about `forceUpdate()`, but `context.customizations.inCustomizerContext` looks like a valid dependency

packages/react-hooks/src/useForceUpdate.test.tsx
ecraig12345: forceUpdate actually isn't stable as checked in, though it obviously should be. I'm fixing this in ##14099.

packages/foundation/src/createComponent.tsx
ecraig12345: Good idea, I'm going to borrow this for #14097 instead of disabling the warning.

packages/react-hooks/src/useOnEvent.ts
ecraig12345: I see that page wrapping the write to `.current` in `useEffect`, but I don't see anything explaining why writing to it during render is wrong? And if that is a problem, we need to fix it in a lot of places and should add a lint rule.

ecraig12345: Is the problem related to concurrent mode? (The page vaguely mentions this but doesn't go into detail.) I guess the analogous concurrent mode problem from class components would be that you should be careful about mutating state during render, since a render method may be called multiple times. @MLoughry do you have any thoughts on this one?

packages/foundation/src/createComponent.tsx
ecraig12345: This is a real component, it's just not named as one, which confuses the rule.

MLoughry: If it acts like the TSLint version, you can fix it by PascalCasing the variable `Result`

packages/foundation/src/next/composed.tsx
MLoughry: Same comment about PascalCasing `Result`

packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,13 +1,15 @@
...
+
 /**
  * React hook to merge multiple React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
  * updates all provided refs
- * @param refs- Refs to collectively update with one ref value.
+ * @param refA- A first ref to collectively update with one ref value.
+ * @param refB- A second ref to collectively update with one ref value.
  */
-export function useMergedRefs<T>(...refs: Ref<T>[]): (instance: T) => void {
+export function useMergedRefs<T>(refA: Ref<T>, refB: Ref<T>): (instance: T) => void {
```
MLoughry, line 9: Again, I don't think this change is necessary or warranted

packages/react-hooks/src/useOnEvent.ts
MLoughry: I don't see in the docs where it says that writing to `.current` during render is unsafe; it's just an unobserved mutable object. Regardless, I think the change is benign.

packages/utilities/src/styled.tsx
MLoughry: It's a benign change, but the logic is such that `inCustomizerContext` would never change for a given instance of this component.

packages/react-hooks/src/useOnEvent.ts
layershifter: Yes, it's about concurrent mode:

https://medium.com/better-programming/how-to-properly-use-the-react-useref-hook-in-concurrent-mode-38c54543857b

ecraig12345: Ugh...we are already doing this wrong in a LOT of places and it's going to be a big headache to fix. :( I'm surprised the hooks lint rules don't try to cover this...

MLoughry: Reading that article, it's less about "never assign to current during render", and more "realize that render may be run multiple times per commit".

For example, our implementation of `useConst` should be fine, even in concurrent mode.

ecraig12345: Hmm yeah I guess it might not be as bad as I thought--I'd thought `usePrevious` was doing this wrong, but actually it's fine. But I will still do a search through the code and check for any instances of where this isn't okay, and try to make the rest of the team aware of this issue.

PR #14118, opened 2020-07-20
Add aria label to website search box

micahgodbolt:
website search was missing a label

image


micahgodbolt:
unrelated screener change. pushing through

PR #13873, opened 2020-06-30
PR deployed sites: Adding react-button and react-next to list of deployed sites

khmakoto:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [ ] Include a change request file using `$ yarn change`

#### Description of changes

This PR adds the `react-button` and `react-next` packages to the list of deployed sites that appears in the PR checks for other people to be able to test the changes that have been done.

#### Focus areas to test

(optional)

PR #14125, opened 2020-07-20
Move Pivot to @fluentui/react-tabs package

### LINKED ISSUE #14034: Move Pivot into its own package
Pivot needs to be moved into its own package `@fluentui/react-tabs`. All components are being moved into individual tabs so they can be versioned and released separately from each other. This will also involve renaming `Pivot` to `Tabs`, leaving the old deprecated name `Pivot` exported as well, for now.

behowell:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14034

#### Description of changes

Move Pivot to @fluentui/react-tabs package


packages/react-tabs/src/common/_constants.scss
```diff
@@ -0,0 +1,14 @@
...
+$fontWeightBold: 700;
+
+$animationEaseFunction1: cubic-bezier(0.1, 0.9, 0.2, 1.0);
+$animationEaseFunction2: cubic-bezier(0.1, 0.25, 0.75, 0.9);
+$animationDuration1: 0.167s;
+$animationDuration2: 0.267s;
+$animationDuration3: 0.367s;
+$animationDuration4: 0.467s;
+
+$CONTEXTUAL_MENU_ITEM_HEIGHT: 36px; // Must match CONTEXTUAL_MENU_ITEM_HEIGHT in ContextualMenu.cnstyles.ts
```
behowell, line 14: I had to copy a bunch of _*.scss files from react-next/src/common into react-tabs/src/common. I'm not a fan of copying code, but I didn't see another established way to handle it for now. Any ideas? (cc @dzearing, @xugao)

xugao: other than having a shared package to put these shared styles ('@fluentui/common-styles`?), i don't see other options. I think i am okay with tracking this as a TODO for now until another package needs these styles

khmakoto: Well, this is already a thing in `react-button` and possibly `react-next` so I think we'd better think about a solution sooner rather than later.

packages/react-next/src/components/Keytip/examples/Keytips.Basic.Example.tsx
```diff
@@ -6,7 +6,7 @@ import { Dropdown } from '@fluentui/react-next/lib/Dropdown';
 import { Link } from '@fluentui/react-next/lib/Link';
 import { SpinButton, ISpinButtonStyles } from '@fluentui/react-next/lib/SpinButton';
 import { Toggle } from '@fluentui/react-next/lib/Toggle';
-import { Pivot, PivotItem } from '@fluentui/react-next/lib/Pivot';
+import { Pivot, PivotItem } from '@fluentui/react-tabs';
```
khmakoto, line 9: Since you are already re-exporting from `react-tabs` in `Pivot.ts` you can probably leave this unchanged.

packages/react-tabs/src/components/Pivot/Pivot.base.tsx
```diff
@@ -121,9 +129,9 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef(
           key={itemKey}
           className={css(className, isSelected && classNames.linkIsSelected)}
           // eslint-disable-next-line react/jsx-no-bind
-          onClick={onLinkClick.bind(this, itemKey)}
+          onClick={(ev: React.MouseEvent<HTMLElement>) => onLinkClick(itemKey!, ev)}
           // eslint-disable-next-line react/jsx-no-bind
-          onKeyDown={onKeyDown.bind(this, itemKey)}
+          onKeyDown={(ev: React.KeyboardEvent<HTMLElement>) => onKeyDown(itemKey!, ev)}
```
khmakoto, line 134: Can we move into `consts` above so that we don't have lambdas in the render function?

behowell: That wouldn't actually change the fact that the function enclosures are newly created on each call to render. We'd need to use `Rect.useCallback` to prevent them from changing... However, using a hook in renderPivotLink isn't technically safe, since the number or order of pivots could change between renders, resulting in the hooks being called in a different order between renders.

We'll need to extract the header links to their own component in order to fix this in the future. It'll likely be a re-composed Button.

packages/react-tabs/.storybook/decorators/withThemeProvider.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { ThemeProvider } from '@fluentui/react-next';
+import { useCustomizationOptions } from '../knobs/theme';
+
+export const withThemeProvider = storyFn => {
```
xugao, line 5: Hmm we should make these shared too since now they are being duplicated in 2 places. `@uifabric/build` package has a folder called `storybook`, i think that's a good place to put

packages/react-tabs/src/components/Pivot/examples/Pivot.Basic.Example.tsx
```diff
@@ -1,7 +1,6 @@
 import * as React from 'react';
-import { Label, ILabelStyles } from '@fluentui/react-next/lib/Label';
```
xugao, line 2: We should add `react-tabs` to PR deploy site, similar to button: https://github.com/microsoft/fluentui/pull/13873

packages/react-tabs/src/common/_constants.scss
xugao: I didn't realize until now how much styles we are duplicating. We should create that shared package now instead of postponing :)

behowell: Logged #14143 to track putting them in a common package. I'll get to it next, but I'd prefer not to delay this PR for it, if possible.

packages/react-tabs/.storybook/decorators/withThemeProvider.js
behowell: Logged #14144 to move storybook files to `@uifabric/build`.

packages/react-tabs/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@fluentui/react-tabs",
+  "version": "0.1.0",
+  "description": "Fluent UI React Tabs component",
+  "private": true,
```
xugao, line 5: remove this line, otherwise package won't be published to npm

packages/react-tabs/.eslintrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  // OUFR has some overrides this package needs to mirror
+  "extends": ["../office-ui-fabric-react/.eslintrc"],
+  "root": true,
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "error"
+  }
```
ecraig12345, line 7: Pretty sure you can just use the default config here:
```suggestion
  "extends": ["plugin:@fluentui/eslint-plugin/react"],
  "root": true
```

behowell: Done. I had to suppress some interface naming warnings that came up as a result.

ecraig12345: I assume in a future change you're planning on actually renaming to Tabs? If so, that would also be a good time to fix the interface names (at least in the renamed versions).

behowell: Yup!

PR #14139, opened 2020-07-21
Syntax highlighting a11y

micahgodbolt:
Updating syntax highlighting to resolve some a11y contrast issues. Bit of a stopgap until we can do a full pass. 

image


ecraig12345:
I thought I'd addressed this previously, using an unofficial "gray15" (manual interpolation between gray10 and gray20)? Not sure what happened to that.

micahgodbolt:
moving to gray 10 didn't fix the gray and blue text sitting on top of it

micahgodbolt:
@ecraig12345 silly console logs fixed.....are you good with this proposal? I'm fine if you want to take another pass at it later. I'm not attached to these colors....just want to get i a11y passing.

micahgodbolt:
yeah, i agree we could take a good design pass on this and really hone it in.

PR #14140, opened 2020-07-21
Add simple transform functions to change prop values

tmaster628:
#### Description of changes

Added support for powerful transforms that allow for developers to have more control over the prop renaming process.

-Enums are now completely designated to a transform case.
-If a dev wants to simply change a prop's value to a default, they can either use a transform OR use the REPLACEMENTVALUE string in renameProp.
-If a dev wants to apply a mapping of values to change (dangerous is the domain of prop values is unbounded), they can provide a string literal mapping in the prop transform.
-Examples to writing one's own transform exist in `transforms.ts`. To handle a custom case, the developer needs to know what kind of SyntaxKind their desired value-to-change is, but beyond that, the process is relatively straightforward if they match the existing patterns. There is currently minimal support for developers who want to modify the spread-handling case -- the expected behavior is that it should change values in the same way that a non-spread case would.


packages/codemods/src/codeMods/utilities/helpers/propHelpers.ts
```diff
@@ -0,0 +1,159 @@
...
+          .getProperties()
+          .some(name => name.getName() === toRename)
+      ) {
+        /* If we've found our prop inside the object,
+          continue attempt to rename it. */
+        const propKind = spreadProp.getDefinitions();
+        if (propKind.length === 1) {
+          const newSpreadName = '__mig' + propKind[0].getName()[0].toUpperCase() + propKind[0].getName().substring(1);
+          const newMapName = '__migEnumMap';
+          switch (propKind[0].getKind()) {
```
joschect, line 39: I know it was like this initially, but I believe that it's time for this switch statement to get put into somethign like a `shouldUpdateElementProp` function that takes in the kind of the prop and then determines whether or not to run it. The same is true for the spreadProp checks in the above if.

PR #14141, opened 2020-07-21
[GroupedList] Copy code from OUFR to react-next

MLoughry:
#### Pull request checklist

#### Description of changes

Copy code from `GroupedList` from OUFR to `react-next`. In the process, I had to fix several `any` typings, and create a stub `DetailsList` index that doesn't export `GroupedList` exports.

#### Focus areas to test

(optional)


ecraig12345:
If it's okay with you I think we should close this one for now since we've decided to delay converting or fixing GroupedList.

PR #14145, opened 2020-07-21
useCustomizationSettings: get settings from CustomizerContext if exists.

xugao:
#### Pull request checklist

#### Description of changes
get local settings from context by default within the hook, instead of being passed as a param.

#### Focus areas to test

(optional)

### DIFF
--- a/change/@uifabric-utilities-2020-07-21-11-56-13-xgao-useCustomization.json
+++ b/change/@uifabric-utilities-2020-07-21-11-56-13-xgao-useCustomization.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "useCustomizationSettings: get settings from CustomizerContext if exists.",
+  "packageName": "@uifabric/utilities",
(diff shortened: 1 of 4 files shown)


micahgodbolt:
Is there any way for the returned theme to have a type? getTheme returns a typed theme, but useCustmizationSettings does not.

Also, is there any reason we wouldn't also expose a useTheme hook that simply returns `useCustomizationSettings(['theme']);`

Thats a VERY verbose line compared to `const theme = getTheme()`

xugao:
@micahgodbolt - is [this](https://github.com/microsoft/fluentui/blob/master/packages/react-next/src/Styling/ThemeProvider.tsx#L24) what you looking for? I didn't put this in a shared location when i added this because nowhere else is using it. Also, we have a new `useTheme` in V8 so i hesitate to create a public one for V7 now. 
What exactly is your use case? can we not adding `useTheme` for now and you just use this implementation i had?

PR #14152, opened 2020-07-21
T-dama/Added More Transforms

tmaster628:
#### Description of changes

Added number and string transforms
Added documentation on how to write a transform
Abstracted out some ts-morph logic from the transform process into helpers so that devs don't have to know too much about AST's.
Reordered a few tests.

### DIFF
--- a/packages/codemods/src/codeMods/utilities/transforms.ts
+++ b/packages/codemods/src/codeMods/utilities/transforms.ts
@@ -1,6 +1,5 @@
 import { PropTransform, ValueMap } from '../types';
 import { JsxExpression, SyntaxKind, JsxOpeningElement, JsxSelfClosingElement } from 'ts-morph';
-import { Maybe } from '../../helpers/maybe';
 import { renamePropInSpread } from './helpers/propHelpers';
 
 /*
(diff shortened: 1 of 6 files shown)


packages/codemods/src/codeMods/utilities/transforms.ts
```diff
@@ -57,21 +126,29 @@ export function enumTransform(map: ValueMap<string>): PropTransform {
...
+   ELEMENT is a JsxExpression, meaning that we have close access to
+   the prop, or if it's a Jsx opening element, meaning we can't access it. */
+function elementNotInSpread(element: JsxExpression | JsxOpeningElement | JsxSelfClosingElement): boolean {
+  return element.getKind() === SyntaxKind.JsxExpression;
+}
+
+/* Returns the piece of the AST that represents the value that the
+   developer can overwrite. */
+function getValueToChange(element: JsxExpression) {
+  const toChange = Maybe(element.getChildAtIndex(1)); // Child between {} operators.
```
joschect, line 152: Since tochange isn't ever operated on it's okay to just do `return element.getChildAtIndex(1)`

PR #14154, opened 2020-07-22
Result: Add either to help with creation of codemods

joschect:
#### Pull request checklist

#### Description of changes

Either makes it easy to handle things like failure states since you can return the error and handle it later after preforming several computations.

#### Focus areas to test

(optional)


packages/codemods/src/helpers/tests/either.test.ts
```diff
@@ -0,0 +1,109 @@
+import { Left, Right, Either, DoEither } from '../either';
+
+const getEither = <T, Z>(right: T, left: Z, returnLeft: boolean = false): Either<T, Z> => {
+  if (returnLeft) {
+    return Left<T, Z>(left!);
+  }
+  return Right<T, Z>(right!);
+};
+
+describe('Maybe', () => {
```
xugao, line 10:
```suggestion
describe('Either', () => {
```

packages/codemods/src/helpers/either.ts
```diff
@@ -0,0 +1,89 @@
...
+  public right: boolean;
+  public value: R | L;
+
+  public constructor(options: { left: boolean; right: boolean; value: R | L }) {
+    this.left = options.left;
+    this.right = options.right;
+    this.value = options.value;
+  }
+
+  public fmap<T>(this: Either<R, L>, fn: (v: R) => T): Either<T, L> {
```
xugao, line 25: could we please add comments for each public method

packages/codemods/src/helpers/tests/either.test.ts
```diff
@@ -0,0 +1,109 @@
+import { Left, Right, Either, DoEither } from '../either';
+
+const getEither = <T, Z>(right: T, left: Z, returnLeft: boolean = false): Either<T, Z> => {
```
xugao, line 3: let's split this to `goLeft` and `goRight` , that is going to make the test a lot easier to read

packages/codemods/src/helpers/either.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  public flatten(): Flattened<R, Either<R, L>> {
+    if (this.value && this.value instanceof EitherInternal) {
+      return this.value as Flattened<R, Either<R, L>>;
+    }
+    return (this as unknown) as Flattened<R, Either<R, L>>;
+  }
+
+  public then<F>(this: Either<R, L>, fnR: (v: R) => F | Either<F, L>, fnL?: (v: L) => F | Either<F, L>): Either<F, L> {
```
xugao, line 47: i don't quite understand this `then` method. why support 2 functions here?

Should it be something like
```
if (this.left) {
   return Left(...)
} 

return Right(..)
```

packages/codemods/src/helpers/either.ts
```diff
@@ -0,0 +1,89 @@
...
+  return new EitherInternal<R, L>({ value: value, left: true, right: false }) as Left<R, L>;
+};
+
+export const Right = <R, L>(value: R): Right<R, L> => {
+  return new EitherInternal<R, L>({ value: value, left: false, right: true }) as Right<R, L>;
+};
+
+export type Either<R, L> = Left<R, L> | Right<R, L>;
+
+export const DoEither = <R, L, T>(either: Either<R, L>, fnR: (v: R) => T, fnL: (v: L) => T): T => {
```
xugao, line 83: 1. why not camelCasing?
2. doEither is a weird name to me. how about something like `run` or `execute`

packages/codemods/src/helpers/either.ts
joschect: This can't support 2 single functions due to the expect input of those lambdas. Having a single function means that the type required to assess is `R | L` which then means that you must know something about the type of the Either which negates a lot of its usefulness. I could potentially rename this whole thing as result. Another way to look at the typings could be `Return Value | Error Value` and you might want to preform several operations on the return value which could return a new Error Value at any time. But the error cannot then return a valid return value.

xugao: I don't quite follow. my understanding for `then` is to proceed another operation on the same direction `this` currently is on, something like: `public then<F>(this: Either<R, L>, fn: (v: R | L) => F): Either<F, L> | Either<R, L>`

This is why having JSDoc for these methods can really help.
Also we don't have test case that covers all the `then` code paths, therefore I am confused what it's designed to do :)

packages/codemods/src/helpers/result.ts
```diff
@@ -0,0 +1,127 @@
+import { Chainable, Flattened } from './chainable';
+
+export interface Ok<R, E> extends ResultInternal<R, E> {
+  ok: true;
+  value: R;
+}
+
+export interface Err<R, E> extends ResultInternal<R, E> {
```
xugao, line 8: sorry for the nit pick, but i think we should use full word `Error` here, and other places too

joschect: I chose err because the Error constructor already exists and I don't want to overload it to much.

packages/codemods/src/helpers/result.ts
```diff
@@ -0,0 +1,127 @@
...
+class ResultInternal<R, E> implements Chainable<R> {
+  public ok: boolean;
+  public value: R | E;
+
+  public constructor(options: { ok: boolean; value: R | E }) {
+    this.ok = options.ok;
+    this.value = options.value;
+  }
+
+  public map<T>(this: Result<R, E>, fn: (v: R) => T): Result<T, E> {
```
xugao, line 22: why do we need `chain`? isn't `map` alone enough?

packages/codemods/src/helpers/result.ts
joschect: Actually Error is not a constructor, it's used the exact same way that this error would be, `Error("Error is thrown")`

packages/codemods/src/helpers/result.ts
```diff
@@ -0,0 +1,127 @@
...
+      return fn(this.value);
+    }
+    return Err(this.value);
+  }
+
+  /**
+   * Works just like chain, but is only called if this Result is an error.
+   * This returns a new Result with type Result<R, NewType>
+   */
+  public orChain<T>(this: Result<R, E>, fn: (v: E) => Result<R, T>): Result<R, T> {
```
xugao, line 41: missing test

packages/codemods/src/helpers/result.ts
```diff
@@ -0,0 +1,127 @@
...
+    return Err(this.value);
+  }
+
+  /**
+   * Works just like then, but is only called if this Result is an error.
+   * This returns a new Result with type Result<R, NewType>
+   */
+  public orThen<F>(this: Result<R, E>, fnErr: (v: E) => F | Result<R, F>): Result<R, F> {
+    if (!this.ok) {
+      return Ok(fnErr(this.value)).flatten() as Result<R, F>;
```
xugao, line 77: shouldn't this be Err?

packages/codemods/src/helpers/result.ts
joschect: Map is sufficient, but I think that chain is a nice to have. Take the instance where I have a function like `processString(str): Result<string, E>`  and I have some result like `fetchedString: Result<string, E>` that was potentially obtained from a rest call. I want to process the string and I cannot do `fetchedString.map(processString)` without calling flatten afterwards. Having something like chain to just handle that is nice. There are also cases where you might want it to be nested, in which case having map is still useful.

packages/codemods/src/helpers/tests/result.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  it('flattens correctly', () => {
+    expect(
+      getOkay(3, '4')
+        .map(v => getError(9, 'good'))
+        .flatten()
+        .errorOrElse('bad'),
+    ).toBe('good');
+  });
+
+  it('fMaps correctly on Okay', () => {
```
xugao, line 50:
```suggestion
  it('map correctly on Okay', () => {
```

packages/codemods/src/helpers/tests/result.test.ts
```diff
@@ -0,0 +1,109 @@
+import { Err, Ok, Result } from '../result';
+
+const getOkay = <T, Z>(okay: T, left: Z): Result<T, Z> => {
```
xugao, line 3: nit: `okay` -> `ok` everywhere

packages/codemods/src/helpers/tests/result.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  it('fMaps correctly on Okay', () => {
+    expect(
+      getOkay(3, '4')
+        .map(v => 'Good')
+        .okOrElse('Bad'),
+    ).toBe('Good');
+  });
+
+  it('fMaps correctly on Err', () => {
```
xugao, line 58:
```suggestion
  it('map correctly on error', () => {
```

packages/codemods/src/helpers/result.ts
joschect: Yes it should be. Or at least I need to change the typings.

packages/codemods/src/helpers/tests/result.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  it('Thens correctly on Err', () => {
+    expect(
+      getError(3, '4')
+        .then(v => 30)
+        .then(v => v.toString())
+        .okOrElse('Bad'),
+    ).toBe('Bad');
+  });
+
+  it('DoEither calls Okay function with Okay object value', () => {
```
xugao, line 84: "DoEither" -> "resolve"

packages/codemods/src/helpers/result.ts
xugao: just pointing out couple things for you to consider:
1. the names `map` and `chain` really confuse me.  I think in the world we only have one of them, either name works for me. having both exist forced me to read into implementation to understand the difference
2. you have `orChain`, not `orMap`

PR #14155, opened 2020-07-22
Create @fluentui/common-styles package, for styling files shared between components

### LINKED ISSUE #14143: Move common scss files into their own package
From #14125, many .scss files had to be copied from react-next to react-tabs. See discussion here: https://github.com/microsoft/fluentui/pull/14125#discussion_r457760987

This tracks moving them to their own package so that the files aren't duplicated.

behowell:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14143

#### Description of changes

Create @fluentui/common-styles package, for styling files shared between components
Move shared .scss files from react-next and react-tabs


packages/common-styles/just.config.ts
```diff
@@ -0,0 +1,3 @@
+const { preset } = require('@uifabric/build');
+
+preset();
```
ecraig12345, line 3: You can simplify this:
```suggestion
const { preset, just } = require('@uifabric/build');
const { task, series } = just;

preset();

task('build', series('clean', 'copy')).cached();
```

packages/common-styles/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "license": "MIT",
+  "scripts": {
+    "build": "just-scripts build",
+    "bundle": "just-scripts bundle",
+    "clean": "just-scripts clean",
+    "code-style": "just-scripts code-style",
+    "just": "just-scripts",
+    "lint": "just-scripts lint",
+    "start": "just-scripts dev:storybook",
+    "update-api": "just-scripts update-api"
```
ecraig12345, line 24: Can be simplified as follows. You could potentially also remove `lint` since it won't do anything right now (though in the future, linting will probably include config files).
```suggestion
    "build": "just-scripts build",
    "clean": "just-scripts clean",
    "code-style": "just-scripts code-style",
    "just": "just-scripts",
    "lint": "just-scripts lint"
```

packages/common-styles/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "@uifabric/set-version": "^7.0.18",
+    "tslib": "^1.10.0"
+  },
+  "peerDependencies": {
+    "@types/react": ">=16.8.0 <17.0.0",
+    "@types/react-dom": ">=16.8.0 <17.0.0",
+    "react": ">=16.8.0 <17.0.0",
+    "react-dom": ">=16.8.0 <17.0.0"
+  }
+}
```
ecraig12345, line 46:
```suggestion
  "devDependencies": {
    "@fluentui/eslint-plugin": "^0.53.1",
    "@uifabric/build": "^7.0.0",
    "office-ui-fabric-core": "^11.0.0"
  }
}
```

packages/common-styles/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@fluentui/common-styles",
+  "version": "0.1.0",
+  "description": "Common styling definitions for react components",
+  "main": "lib-commonjs/index.js",
+  "module": "lib/index.js",
+  "typings": "lib/index.d.ts",
+  "sideEffects": [
+    "lib/version.js"
+  ],
```
ecraig12345, line 10: None of these are relevant in this case
```suggestion
```

packages/common-styles/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
```
ecraig12345, line 1: Remove this file

scripts/create-package/plop-templates-react/package.json.hbs
```diff
@@ -2,7 +2,6 @@
   "name": "{{{packageNpmName}}}",
   "version": "0.1.0",
   "description": "{{{description}}}",
-  "private": true,
```
behowell, line 5: @ecraig12345 I updated the template to remove "private", so that new packages are published by default. Happy to revert this if it's not a good idea.

ecraig12345, line 5: You should bring this back--we actually still want to avoid publishing new packages by default.

packages/common-styles/README.md
```diff
@@ -0,0 +1,11 @@
+# @fluentui/common-styles
+
+**Common style definitions for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
+
+These are not production-ready components and **should never be used in product**. This space is useful for testing new components whose APIs might change before final release.
```
ecraig12345, line 5: Probably modify this to reflect the package contents better

packages/common-styles/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@fluentui/common-styles",
+  "version": "0.1.0",
+  "description": "Common style definitions for react components",
```
ecraig12345, line 4:
```suggestion
  "description": "Common style definitions for Fluent UI React components",
```

behowell:
> Thanks for doing this! There are some issues with the config. Also I'd be interested to hear others' input on the package name.
> 
> I think it would be best if we copy the logic from OUFR where we copy our common files plus fabric-core files to `dist/sass`. In that case you'll need a file `config/pre-copy.json` similar to OUFR's:
> 
> ```
> {
>   "copyTo": {
>     "dist/sass": [
>       "office-ui-fabric-core/dist/sass/**/*",
>       "./src/common/_highContrast.scss",
>       "./src/common/_i18n.scss",
>       "./src/common/_semanticSlots.scss",
>       "./src/common/_themeOverrides.scss",
>       "./src/common/_legacyThemePalette.scss",
>       "./src/common/_effects.scss",
>       "./src/common/_themeVariables.scss",
>       "./src/common/ThemingSass.scss"
>     ]
>   }
> }
> ```

@ecraig12345 Any reason not all of the .scss files are included? (missing _common.scss, _constants.scss, _focusBorder.scss, _themeCssVariables.scss). And if we should include all of them, should it just be `"./src/*.scss"`?

packages/react-next/package.json
```diff
@@ -34,6 +34,7 @@
     "update-api": "just-scripts update-api"
   },
   "devDependencies": {
+    "@fluentui/common-styles": "^0.1.0",
```
behowell, line 37: Moved to devDependencies, since OUFR's package.json lists office-ui-fabric-core under devDependencies. Just wanted to double-check that's correct @ecraig12345?

ecraig12345:
> @ecraig12345 Any reason not all of the .scss files are included? (missing _common.scss, _constants.scss, _focusBorder.scss, _themeCssVariables.scss). And if we should include all of them, should it just be `"./src/*.scss"`?

"Historical reasons"...? Either that or the copy task doesn't support globs, not sure. But if the copy task *does* support globs, it's probably fine to include them all.

packages/react-next/package.json
ecraig12345: I think so? Not the most familiar with scss build but I'm pretty sure these common scss files are only processed at build time.

change/@fluentui-common-styles-2020-07-21-18-01-09-styles-package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
khmakoto, line 2:
```suggestion
  "type": "minor",
```

behowell: It's starting at v0.1.0 now, so `patch` will make the released version v0.1.1. `minor` would increment it to start at v0.2.0, right?

ecraig12345: Right. Minor would generally be correct since you're adding new APIs, but practically I don't think it matters in this case.

PR #14167, opened 2020-07-22
Support onChange callback in FluentUI Datepicker Input Field

rymeskar:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Datepicker/DatepickerCalendar.tsx
```diff
@@ -309,20 +235,47 @@ DatepickerCalendar.propTypes = {
...
   accessibility: datepickerCalendarBehavior,
-  firstDayOfWeek: DayOfWeek.Monday,
+  firstDayOfWeek: DayOfWeek.Sunday,
   firstWeekOfYear: FirstWeekOfYear.FirstDay,
   dateRangeType: DateRangeType.Day,
   header: {},
   calendarCell: {},
   calendarHeaderCell: {},
-  localizedStrings: DEFAULT_CALENDAR_LOCALIZED_STRINGS,
+  ...DEFAULT_CALENDAR_STRINGS,
```
pompomon, line 278: Please could you update DatepickerCalendarHeader.tsx with the localized string for next/previous month (there are TODO: marks)

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -234,17 +240,24 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
             ...unhandledProps,
           })}
         >
-          <Input readOnly onClick={showCalendarGrid} value={valueFormatter(selectedDate)} />
+          <Input
```
pompomon, line 243: Please could you update this part to also use `createShorthand`, like other parts are doing?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -165,31 +94,65 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
   const context = useFluentContext();
   const { setStart, setEnd } = useTelemetry(Datepicker.displayName, context.telemetry);
   setStart();
+  const dateSelectionErrorString = 'A date selection is required';
   const datepickerRef = React.useRef<HTMLElement>();
-  const [open, setOpen] = React.useState<boolean>(false);
+  const [openState, setOpenState] = useAutoControlled<OpenState>({
```
layershifter, line 99: This should be just `boolean`

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -165,31 +94,65 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+    defaultValue: props.defaultCalendarOpenState ? OpenState.Open : OpenState.Closed,
+    value: undefined,
+    initialValue: OpenState.Closed,
```
layershifter, line 102:
```suggestion
    initialValue: false,
```

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -165,31 +94,65 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
   const context = useFluentContext();
   const { setStart, setEnd } = useTelemetry(Datepicker.displayName, context.telemetry);
   setStart();
+  const dateSelectionErrorString = 'A date selection is required';
   const datepickerRef = React.useRef<HTMLElement>();
-  const [open, setOpen] = React.useState<boolean>(false);
+  const [openState, setOpenState] = useAutoControlled<OpenState>({
+    defaultValue: props.defaultCalendarOpenState ? OpenState.Open : OpenState.Closed,
```
layershifter, line 100:
```suggestion
    defaultValue: props.defaultOpen,
```

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -165,31 +94,65 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+    value: undefined,
```
layershifter, line 101:
```suggestion
    value: props.open,
```

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -165,31 +94,65 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
-  const calendarOptions: IDayGridOptions = {
-    selectedDate: selectedDate ?? props.today ?? new Date(),
-    navigatedDate: selectedDate ?? props.today ?? new Date(),
-    firstDayOfWeek,
-    firstWeekOfYear,
-    dateRangeType,
-  };
+  const [formattedDate, setFormattedDate] = React.useState<string>('');
+  const [error, setError] = React.useState<string>(() =>
+    props.isRequired && !selectedDate ? dateSelectionErrorString : '',
```
layershifter, line 107:
```suggestion
    props.required && !selectedDate ? dateSelectionErrorString : '',
```

We don't use `is` prefix

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -165,31 +94,65 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+  );
 
-  const showCalendarGrid = () => {
-    setOpen(true);
+  const { calendar, popup, className, design, styles, variables, formatMonthDayYear } = props;
+  const valueFormatter = date => (date ? formatMonthDayYear(date) : '');
```
layershifter, line 111: Please move out it from component's body

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -165,31 +94,65 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+    daysToSelectInDayView: props.daysToSelectInDayView,
+    today: props.today,
+    showWeekNumbers: props.showWeekNumbers,
+    workWeekDays: props.workWeekDays,
+    minDate: props.minDate,
+    maxDate: props.maxDate,
+    restrictedDates: props.restrictedDates,
+  } as IDayGridOptions;
+
+  const dateFormatting = {
```
layershifter, line 130: Is there a type for this?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -213,10 +176,15 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
 
   const overrideDatepickerCalendarProps = (predefinedProps: DatepickerCalendarProps): DatepickerCalendarProps => ({
     ...calendarOptions,
+    ...dateFormatting,
     onDateChange: (e, itemProps) => {
-      setSelectedDate(itemProps.value.originalDate);
-      setOpen(false);
-      _.invoke(predefinedProps, 'onDateChange', e, itemProps);
+      const targetDay = itemProps.value;
+      setSelectedDate(targetDay.originalDate);
```
layershifter, line 182: Please switch state management to `useReducer`, it's hard to follow this

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -165,31 +94,65 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+  );
 
-  const showCalendarGrid = () => {
-    setOpen(true);
+  const { calendar, popup, className, design, styles, variables, formatMonthDayYear } = props;
+  const valueFormatter = date => (date ? formatMonthDayYear(date) : '');
+
+  const nonNullSelectedDate = selectedDate ?? props.today ?? new Date();
+
+  const calendarOptions = {
```
layershifter, line 115: Is there a type for this?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -225,6 +193,44 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+      } else {
+        setOpenState(OpenState.Open);
+      }
+    }
+  };
+
+  const onInputChange = (e, target: { value: string }) => {
+    const parsedDate = props.parseDate(target.value);
+
+    setFormattedDate(target.value);
```
layershifter, line 216: Validation is pure and should be done outside of a component

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -276,25 +296,49 @@ Datepicker.propTypes = {
...
+  shortMonths: PropTypes.arrayOf(PropTypes.string),
+  days: PropTypes.arrayOf(PropTypes.string),
+  shortDays: PropTypes.arrayOf(PropTypes.string),
+
+  isRequiredErrorMessage: PropTypes.string,
+  invalidInputErrorMessage: PropTypes.string,
+  isOutOfBoundsErrorMessage: PropTypes.string,
   goToToday: PropTypes.string,
-  placeholder: PropTypes.string,
+  prevMonthAriaLabel: PropTypes.string,
```
layershifter, line 315: Why all of these are not in `LocalizedStrings`?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -264,6 +277,13 @@ Datepicker.propTypes = {
   calendar: customPropTypes.itemShorthand,
   popup: customPropTypes.itemShorthand,
 
+  disabled: PropTypes.bool,
+  isRequired: PropTypes.bool,
+  onDateChange: PropTypes.func,
+  placeholder: PropTypes.string,
+  allowTextInput: PropTypes.bool,
```
layershifter, line 284: @pompomon is there a better name for this prop?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Should I then also add slots for Input?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: @layershifter Do I understand that I should rather add the localizedStrings type as property and not extend the localizedStrings type?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: @layershifter  I copied the name from the [FabricUI](https://github.com/microsoft/fluentui/blob/fdc095fd93f528381650e1215bcf2bfe1ff9acf6/packages/date-time/src/components/DatePicker/DatePicker.base.tsx#L477).

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: @layershifter just outside of the component's body or completely outside of this file?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Added type.

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Added type.

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: @layershifter  in this same file or in different one?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Fixed.

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: @layershifter  what would then be the difference between 'props.defaultOpen' and 'props.open'?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: @layershifter I use the enum to mimic the behavior of FabricUI/Teams. https://github.com/microsoft/fluentui/blob/0b666d9bc2e4b20e9267a78de4327998eda5f6b4/packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx#L204

Should I then simplify to boolean but lose the described behavior?

FYI: @pompomon

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
pompomon: Yeah, let's do that :)

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
pompomon: ManualInput/AllowManualInput was my other option, but I think we can do that like it

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -225,6 +193,44 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+    }
+  };
+
+  const onInputChange = (e, target: { value: string }) => {
+    const parsedDate = props.parseDate(target.value);
+
+    setFormattedDate(target.value);
+    if (parsedDate) {
+      if (isRestrictedDate(parsedDate, calendarOptions)) {
+        setError('The selected date is from the restricted range.');
```
pompomon, line 219: That should be a prop, so consumers can put the localized version of the error

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -225,6 +193,44 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+      } else {
+        setError('');
+        setSelectedDate(parsedDate);
+        _.invoke(props, 'onDateChange', e, { ...props, value: parsedDate });
+      }
+    } else if (target.value) {
+      setError('Manually entered date is not in correct format.');
```
pompomon, line 226: Same here, we should be able to localize all the labels

rymeskar: Good point, changed.

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Good point, changed.

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Is it then fair to also create a composite state object { error, formattedDate, selectedDate }  or you would keep the states separate?

packages/fluentui/docs/src/components/ComponentPlayground/componentGenerators.tsx
```diff
@@ -36,6 +37,9 @@ export const Box: KnobComponentGenerators<BoxProps> = {
   children: () => null,
 };
 
+export const Input: KnobComponentGenerators<InputProps> = {
+  children: () => null,
+};
```
layershifter, line 42: It should not be required.

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
layershifter: I don't see any reason to have enum state, `open` is am `boolean` value

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
layershifter: The same

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
layershifter: > Is it then fair to also create a composite state object { error, formattedDate, selectedDate } or you would keep the states separate?

Let's keep as it is for now.

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
layershifter: Probably even in a separate file. I expect something like this:

```
const { error } = validateDate()

setError(error)
```

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: @layershifter switched in [the following commit](https://github.com/microsoft/fluentui/pull/14563/commits/f4c038eeed968dfec4a00cc20d2194bc72072387).

packages/fluentui/docs/src/components/ComponentPlayground/componentGenerators.tsx
rymeskar: When removed, I got the following error on viewing examples:
image

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Should I move the `valueFormatter` away from component's body even though it depends on props?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Updated in [the following commit](https://github.com/microsoft/fluentui/pull/14563/commits/f1dce86c84c5559c3e397d00f4b20bd814c3fb7f).

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
layershifter: I see, now it depends on props. Ignore it

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
layershifter: Oh, I see. All strings are props actually... Let's discuss this offline

PR #14169, opened 2020-07-22
Feature/new tooltip design supporting subcounts in line chart

jisooj:
#### Pull request checklist

Edit: Looks like some already did refactoring work on charts. ChartHelper component was created so that it can be utilized across different types of charts. Right now, LineChart and AreaChart use this base component to render charts. That's why I moved tooltip code to ChartHelper when I was resolving conflicts.

#### Description of changes
I'm working a feature in M365 Security Center and this new feature requires line chart to show sub-counts in tooltip. These sub-count are in form of a dictionary of string (sub-count name as key) to number (numeric value as value) where all the values sum up to y-value at its corresponding point. Suppose we have a point (x_i, y_i). The sub-counts at this point sum up to y_i.

Tooltip behavior:
- For a point with subcounts the new design is used
- For a point without subcounts the original design is used

**Current tooltip:**
image

**Updated tooltip:**

Point with subcount:
image

Point without subcounts:
image

#### Focus areas to test

(optional)


packages/charting/src/components/LineChart/LineChart.base.tsx
```diff
@@ -263,37 +270,30 @@ export class LineChartBase extends React.Component<
             id={`toolTip${this._uniqueCallOutID}`}
           >
             <div className={this._classNames.calloutContentRoot}>
-              <div className={this._classNames.calloutDateTimeContainer}>
+              <div
+                className={this._classNames.calloutDateTimeContainer}
+                style={yValueHoverSubCountsExists ? { marginBottom: '11px' } : {}}
```
Raghurk, line 275: >{yValueHoverSubCountsExists ? { marginBottom: '11px' } : {} [](start = 22, length = 59)

yValueHoverSubCountsExists &&{ marginBottom: '11px' }

packages/charting/src/components/LineChart/LineChart.base.tsx
```diff
@@ -263,37 +270,30 @@ export class LineChartBase extends React.Component<
...
+              >
                 <div className={this._classNames.calloutContentX}>{this.state.hoverXValue} </div>
                 {/*TO DO  if we add time for callout then will use this */}
                 {/* <div className={this._classNames.calloutContentX}>07:00am</div> */}
               </div>
-              <div className={this._classNames.calloutInfoContainer}>
+              <div
+                className={this._classNames.calloutInfoContainer}
+                style={yValueHoverSubCountsExists ? { display: 'flex' } : {}}
```
Raghurk, line 283: >{yValueHoverSubCountsExists ? { display: 'flex' } : {}} [](start = 22, length = 55)

yValueHoverSubCountsExists && { display: 'flex' }

packages/charting/src/components/LineChart/LineChart.base.tsx
```diff
@@ -263,37 +270,30 @@ export class LineChartBase extends React.Component<
...
+                    const isLast: boolean = index + 1 === yValues.length;
+                    return (
                       <div
-                        key={index}
-                        id={`${index}_${xValue.y}`}
-                        className={mergeStyles(this._classNames.calloutBlockContainer, {
-                          borderLeft: `4px solid ${xValue.color}`,
-                        })}
+                        key={`callout-content-${index}`}
+                        style={yValueHoverSubCountsExists ? { display: 'inline-block' } : {}}
```
Raghurk, line 291: >{yValueHoverSubCountsExists ? { display: 'inline-block' } : {}} [](start = 30, length = 63)

same as above

packages/charting/src/components/LineChart/LineChart.base.tsx
```diff
@@ -302,6 +302,80 @@ export class LineChartBase extends React.Component<
...
+    },
+    index: number,
+    yValueHoverSubCountsExists: boolean,
+    isLast: boolean,
+  ): React.ReactNode {
+    const marginStyle: React.CSSProperties = isLast ? {} : { marginRight: '16px' };
+
+    if (!xValue.yAxisCalloutData || typeof xValue.yAxisCalloutData === 'string') {
+      return (
+        <div style={yValueHoverSubCountsExists ? marginStyle : {}}>
```
Raghurk, line 335: >style={yValueHoverSubCountsExists ? marginStyle : {} [](start = 13, length = 52)

same as above

packages/charting/src/components/LineChart/LineChart.base.tsx
jisooj: I also tried to shorten with the your suggestion but this was what I was getting:

Type 'false | { marginBottom: string }' is not assignable to type 'CSSProperties | undefined'. Type 'false' is not assignable to type 'CSSProperties | undefined'

React prop style was evaluated into boolean when we have yValueHoverSubCountsExists &&{ marginBottom: '11px' } and that's why it wasn't happy about it. The closest I can get to is to pass undefined instead of empty object. i.e. style={yValueHoverSubCountsExists ? { marginBottom: '11px' } : undefined}

Let me know if there's a better approach

PR #14171, opened 2020-07-22
Updating FocusTrapZone to a function component inside React-Next

czearing:
#### Pull request checklist

#### Description of changes
1. Converted `FocusTrapZone` to a function component inside React-Next
2. Added perf-test tests for `FocusTrapZoneNext` and `FocusTrapZone`
3. Added vr-tests for `FocusTrapZoneNext`

#### Focus areas to test
`FocusTrapZone`

### DIFF
--- a/apps/perf-test/src/scenarios/FocusTrapZone.tsx
+++ b/apps/perf-test/src/scenarios/FocusTrapZone.tsx
@@ -1,0 +1,19 @@
+import * as React from 'react';
+import { FocusTrapZone } from 'office-ui-fabric-react';
+
+const FocusTrapZoneContent: React.FunctionComponent = () => {
+  const items: React.ReactNode[] = [];
+
+  for (let i = 0; i < 20; i++) {
+    items.push(<li key={i} data-is-focusable />);
+  }
+  return <>{items}</>;
+};
+
+const Scenario = () => (
+  <FocusTrapZone>
+    <FocusTrapZoneContent />
+  </FocusTrapZone>
+);
+
+export default Scenario;
--- a/packages/react-next/src/components/FocusTrapZone/FocusTrapZone.test.tsx
+++ b/packages/react-next/src/components/FocusTrapZone/FocusTrapZone.test.tsx
@@ -3,7 +3,7 @@ import * as ReactTestUtils from 'react-dom/test-utils';
 import { KeyCodes } from '../../Utilities';
 import { FocusZone, FocusZoneDirection } from '../../FocusZone';
 import { FocusTrapZone } from './FocusTrapZone';
-import { IFocusTrapZoneProps } from './FocusTrapZone.types';
+import { IFocusTrapZoneProps, IFocusTrapZone } from './FocusTrapZone.types';
 
 // rAF does not exist in node - let's mock it
 window.requestAnimationFrame = (callback: FrameRequestCallback) => {
@@ -23,7 +23,7 @@ class FocusTrapZoneTestComponent extends React.Component<{}, { isShowingFirst: b
   public render() {
     return (
       <div>
-        <FocusTrapZone forceFocusInsideTrap={true} isClickableOutsideFocusTrap={false}>
+        <FocusTrapZone forceFocusInsideTrap isClickableOutsideFocusTrap={false}>
           <button className={'a'} onClick={this._toggleFirst}>
             a
           </button>
@@ -33,7 +33,7 @@ class FocusTrapZoneTestComponent extends React.Component<{}, { isShowingFirst: b
         </FocusTrapZone>
 
         {this.state.isShowingFirst && (
-          <FocusTrapZone forceFocusInsideTrap={false} isClickableOutsideFocusTrap={false}>
+          <FocusTrapZone forceFocusInsideTrap isClickableOutsideFocusTrap={false}>
             <FocusZone data-is-visible={true}>First</FocusZone>
           </FocusTrapZone>
         )}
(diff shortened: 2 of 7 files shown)


apps/perf-test/src/scenarios/FocusTrapZone.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
```
ecraig12345, line 1: Do we really need a perf test for this? Each perf test does have a cost, so I'm not sure we necessarily need them for all components by default.

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.test.tsx
```diff
@@ -534,22 +543,29 @@ describe('FocusTrapZone', () => {
...
+            /*noop*/
+          },
+          stopPropagation: () => {
+            /*noop*/
+          },
+        });
       });
-      expect(lastFocusedElement).toBe(buttonB);
+
+      expect(lastFocusedElement).toBe(buttonA);
```
ecraig12345, line 568: Why this change?

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.test.tsx
```diff
@@ -704,44 +720,41 @@ describe('FocusTrapZone', () => {
...
+          <div>
+            <FocusTrapZoneTestComponent />
+          </div>,
+        ) as unknown) as HTMLElement;
+      });
+      const buttonA = topLevelDiv!.querySelector('.a') as HTMLElement;
+      const buttonB = topLevelDiv!.querySelector('.b') as HTMLElement;
 
       expect(focusTrapZoneFocusStack.length).toBe(2);
-      const baseFocusTrapZone = focusTrapZoneFocusStack[0];
```
ecraig12345, line 729: Did the behavior change such that this test didn't work, or was this just not necessary to cover here? (same for other changes in this test)

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -8,334 +8,389 @@ import {
...
-    }
-  }
+  const updateEventHandlers = (): void => {
+    const { isClickableOutsideFocusTrap = false, forceFocusInsideTrap = true } = props;
+    const win = getWindow(root.current)!;
 
-  private _onFocusCapture = (ev: React.FocusEvent<HTMLDivElement>) => {
-    if (this.props.onFocusCapture) {
-      this.props.onFocusCapture(ev);
+    if (forceFocusInsideTrap && !state.disposeFocusHandler) {
```
ecraig12345, line 246: Is there any way we can handle click and focus handler creation and disposal with effects or something? While a 1-1 mapping from private members to internal state may technically work, it would be nice if we can take advantage of automatic updating and disposal patterns provided by hooks.

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -8,334 +8,389 @@ import {
...
+    disabled,
+  } = props;
+  const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(props, divProperties);
+  const { current: state } = React.useRef<IFocusTrapZoneState>({
+    previouslyFocusedElementOutsideTrapZone: undefined,
+    previouslyFocusedElementInTrapZone: undefined,
+    disposeFocusHandler: undefined,
+    disposeClickHandler: undefined,
+    hasFocus: false,
+    prevProps: props,
```
ecraig12345, line 81: Don't put this in internal state; instead use `usePrevious(props)`

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -8,334 +8,389 @@ import {
...
+};
+
+interface IFocusTrapZoneState {
+  disposeFocusHandler: (() => void) | undefined;
+  disposeClickHandler: (() => void) | undefined;
+  previouslyFocusedElementOutsideTrapZone: HTMLElement | undefined;
+  previouslyFocusedElementInTrapZone: HTMLElement | undefined;
+  hasFocus: boolean;
+  prevProps: IFocusTrapZoneProps;
+  hasPreviouslyRendered: boolean;
```
ecraig12345, line 54: Remove this--looks like it's being updated but not actually read anywhere

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -8,334 +8,389 @@ import {
...
+      state.hasFocus = true;
+    },
+    [props.onFocus],
+  );
+
+  const onRootBlur = React.useCallback(
+    (ev: React.FocusEvent<HTMLDivElement>) => {
+      if (props.onBlur) {
+        props.onBlur(ev);
+      }
```
ecraig12345, line 154:
```suggestion
      props.onBlur?.(ev);
```

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -8,334 +8,389 @@ import {
...
+        state.hasFocus = false;
+      }
+    },
+    [props.onBlur, elementContains, root.current],
+  );
 
-  private _onFirstBumperFocus = () => {
-    this._onBumperFocus(true);
-  };
+  const onFirstBumperFocus = useConstCallback(() => {
```
ecraig12345, line 171: These can't use `useConstCallback` because `onBumperFocus` isn't constant.

JustSlone:
@khmakoto can you help out and review this PR. Thanks!

PR #14173, opened 2020-07-22
Moving Autofill to react-next, in preparation for converting to a function component

czearing:
#### Pull request checklist

#### Description of changes
This change does not modify Autofill; it simply moves it to react-next for better diffing in the actual conversion PR.


ecraig12345:
Unless someone has immediate plans to convert Autofill, I think we should hold off moving it so that any new fixes will be picked up automatically for as long as possible. Same applies to other components.

PR #14179, opened 2020-07-23
fix(Popup): allow close controlled popup when have hover trigger

### LINKED ISSUE #14080: Popup opened on hover cannot be closed by changing the open state
### Environment Information

- **Package version(s)**: 0.50.x

Create popup that has a trigger and `on="hover"`
Use`open` and `onOpenChange` to control the state
Add button to the popup content to close the popup by `onClick={() => setOpen(false)}`

#### Actual behavior:
Popup stays open after clicking on the close button

#### Expected behavior:

Popup closes

https://codesandbox.io/s/epic-kirch-yk2kp?file=/example.js

assuncaocharles:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14080

#### Description of changes

Removing the addition of `click` event to set popup open, this forced the popup to be always open even if a close button that changes the state of the `Popup` is inside the `PopupContent`. Keeping the popup open on clicking the content doesn't need to be enforced since it's the common behavior ( It's gonna close only `onMouseLeave` ). 

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/test/specs/components/Popup/Popup-test.tsx
```diff
@@ -150,6 +150,34 @@ describe('Popup', () => {
     });
   });
 
+  describe('controlled popup', () => {
```
layershifter, line 153: I suggest to move this to e2e tests

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
```diff
@@ -365,10 +365,6 @@ export const Popup: React.FC<PopupProps> &
         setPopupOpen(false, e);
         predefinedProps && _.invoke(predefinedProps, 'onMouseLeave', e, contentProps);
       };
-      contentHandlerProps.onClick = (e, contentProps) => {
-        setPopupOpen(true, e);
-        predefinedProps && _.invoke(predefinedProps, 'onClick', e, contentProps);
-      };
```
layershifter, line 371: I checked the original PR (https://github.com/microsoft/fluent-ui-react/pull/622) and this looks a bug as I don't see any scenario in which this is required as this handlers will go to `PopupContent`... Good catch 👍

packages/fluentui/react-northstar/test/specs/components/Popup/Popup-test.tsx
assuncaocharles: Added e2e

PR #14181, opened 2020-07-23
[Callout] Migrate Callout to a function component

MLoughry:
#### Pull request checklist

#### Description of changes

Migrates `Callout` and `CalloutContentBase` to function components

#### Focus areas to test

(optional)


packages/react-next/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -50,136 +51,353 @@ const OFF_SCREEN_STYLE = { opacity: 0, filter: 'opacity(0)' };
...
+  preventDismissOnScroll: false,
+  preventDismissOnResize: false,
+  isBeakVisible: true,
+  beakWidth: 16,
+  gapSpace: 0,
+  minPagePadding: 8,
+  directionalHint: DirectionalHint.bottomAutoEdge,
+} as const;
+
+function useTargets({ target }: ICalloutProps, calloutElement: React.RefObject<HTMLDivElement | null>) {
```
dzearing, line 65: pretty neat hook! it would be nice to add some jsdoc comments above; also curious if it could just take in the target - maybe this is generalized enough to be used whenever we want to point something at an element (whether it be an element, string, or ref.)

dzearing: trying not to ask too much here :) good job though.

packages/react-next/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -50,136 +51,353 @@ const OFF_SCREEN_STYLE = { opacity: 0, filter: 'opacity(0)' };
...
+        targetWindowRef.current = getWindow(currentElement)!;
+        targetRef.current = target as Point;
+      }
+    } else {
+      targetWindowRef.current = getWindow(currentElement)!;
+    }
+    previousTargetProp.current = target;
+  }
+
+  return [targetRef, targetWindowRef] as const;
```
dzearing, line 106: curious why you need windowref.

packages/react-next/src/components/Callout/CalloutContent.base.tsx
dzearing: Same with the other hooks, just a doc comment above them would help with understanding purpose and readability.

packages/react-next/src/components/Callout/CalloutContent.base.tsx
MLoughry: To cache the window reference used by other functions, rather than re-fetching it every render.

PR #14184, opened 2020-07-23
fix(dropdown): event propagation

assuncaocharles:
#### Pull request checklist

#### Description of changes

`Dropdown` should only prevents propagation of `ESC` keydown if the list is open, otherwise propagation should still happen. 

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
+++ b/packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
@@ -1037,7 +1037,9 @@ export const Dropdown: ComponentWithAs<'div', DropdownProps> &
           case keyboardKey.Escape:
             // If dropdown list is open ESC should close it and not propagate to the parent
             // otherwise event should propagate
-            open && e.stopPropagation();
+            if (open) {
+              e.stopPropagation();
+            }
           default:
             break;
         }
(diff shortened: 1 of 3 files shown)


packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1035,7 +1035,9 @@ export const Dropdown: ComponentWithAs<'div', DropdownProps> &
             tryRemoveItemFromValue();
             break;
           case keyboardKey.Escape:
-            e.stopPropagation();
+            // If dropdown list is open ESC should close it and not propagate to the parent
+            // otherwise event should propagate
+            open && e.stopPropagation();
```
pompomon, line 1040: Please could you add a test for that? e2e or unit, up to you ;)
Just so we have a tool to test it, in case the behavior changes ;)

layershifter, line 1040:
```suggestion
            if (open) {
              e.stopPropagation();
            }
```

nit

assuncaocharles: Done

assuncaocharles: added

PR #14187, opened 2020-07-23
Move ThemeProvider from react-next to react-theme-provider's compat folder

### LINKED ISSUE #14174: Pivot from react-tabs package has a dependency on ThemingProvider from react-next
Attempting to use a `<Pivot>` from the new `@fluentui/react-tabs` package doesn't have the correct styling unless it's wrapped in a `<ThemeProvider>` from the `@fluentui/react-next` package. This makes it not possible to use the new Pivot control without taking a dependency on react-next.

We'll either need to move ThemeProviderout of react-next, or add a temporary workaround to the styling in react-tabs.

Example:
```tsx
import { Fabric } from "office-ui-fabric-react";
import { Pivot, PivotItem } from "@fluentui/react-tabs";

export const TabsPackageExample = () => {
  return (
    <Fabric>
      <Pivot>
          <PivotItem headerText="Tab 1">Foo</PivotItem>
          <PivotItem headerText="Tab 2">Bar</PivotItem>
          <PivotItem headerText="Tab 3">Baz</PivotItem>
      </Pivot>
    </Fabric>
  );
};
```

Result:
image

Expected:
image

behowell:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14174

#### Description of changes

Move `<ThemeProvider>` from `react-next` to `react-theme-provider`'s `compat` folder


packages/react-next/src/Styling.ts
```diff
@@ -1 +1,2 @@
-export * from './Styling/index';
+export * from 'office-ui-fabric-react/lib/Styling';
+export * from '@fluentui/react-theme-provider/lib/Compat';
```
xugao, line 2: we should probably export this from `react-tab` too

packages/react-theme-provider/src/Compat/ThemeProvider.tsx
```diff
@@ -1,10 +1,10 @@
...
-import { ThemeProvider as ReactThemeProvider, ThemeProviderProps } from '@fluentui/react-theme-provider';
-import { useCustomizationSettings } from '../Utilities';
+import { ThemeProvider as ReactThemeProvider, ThemeProviderProps } from '../index';
+import { useCustomizationSettings } from '@uifabric/utilities';
 import { convertLegacyTheme } from './convertLegacyTheme';
-import { ITheme } from '../Styling';
+import { ITheme } from '@uifabric/styling';
 
-export { ThemeProviderProps, useTheme } from '@fluentui/react-theme-provider';
+export { ThemeProviderProps, useTheme } from '../index';
```
khmakoto, line 7: nit: You can now put these and the imports on line 2 in the same line.

behowell: How? This is an export statement. Also I don't want to get too far down the rabbit hole of cleaning up the code while moving it.

khmakoto: Oh, my bad, thought it was an input as well. Resolving.

behowell: Np :)

PR #14203, opened 2020-07-24
feat: add web components to fluent ui as a new package and fix text field styling issue

### LINKED ISSUE #14071: RFC: Intent to contribute FAST web components for FluentUI
## Overview

[FAST](https://fast.design) is a collection of JavaScript packages centered around web standards, designed to help you efficiently tackle some of the most common challenges in website and application design and development. For over a year now, we've been syncing regularly with several of the folks involved with the FluentUI effort in an effort to continually find opportunities to align and converge. We're also an active part of the W3C Open UI effort. While we've always had a desire to converge from a code standpoint, previous requirements and deltas in implementation made it difficult to do so with our *now deprecated/archived* React implementation. As we've made the shift to focus solely on web components (HTML Custom Elements), we believe now is a great time to partner on the FluentUI effort and provide a Web Component offering built on FAST.

Currently there is a MSFT package of components which lives in the FAST repository, we'd like to bring that here and make it a FluentUI offering.

## Goals & Convergence

Our long term goals for this would be to have alignment across design, as well as having unified API and terminology that aligns with the patterns established in Open UI. Considering that there are teams both inside and outside Microsoft leveraging these controls, and knowing that certain design requirements feed certain technical implementations, we do see some of this as a longer-term goal. Some things may be easier to align on than others, but we don't believe that the current divergence is a reason to block this move.

### Proposed model for pursuing convergence

In the short term, we'd like to move the package over and continue our development openly here as a first step. As we identify gaps in current conventions, potentially new requirements, or opportunities to find alignment, we'd like to drive and track those through Github issues. The design issues and things like how we calculate and apply color will likely be larger areas and require a bit more investment across our own design partners and teams. Ultimately though, we are committed to doing so and believe that moving forward will create the momentum needed to make this happen. Anything that relates to component naming, parts, or anatomy we are committed to resolving through Open UI.

#### What this means in regards to FAST

In FAST, we primarily focus on primitives, common UI patterns, utilities, and some composed components or experiences for our own FAST design system implementation (shown on our [component explorer](https://explore.fast.design)). We have several layers that make up FAST. I won't go into a ton of detail on these as there is plenty of documentation on the site, but I do want to highlight a couple of important things about the packages and how it may impact convergence.

**fast-element:** 
The `fast-element` library is a lightweight means to easily building performant, memory-efficient, standards-compliant Web Components. FAST Elements work in every major browser and can be used in combination with any front-end framework or even without a framework. 

**fast-foundation:** 
The fast-foundation package is a library of Web Component classes, templates, and other utilities intended to be composed into registered Web Components by design systems (e.g. Fluent Design, Material Design, etc.). The exports of this package can generally be thought of as un-styled base components that implement semantic and accessible markup and behavior.

This package does not export Web Components registered as custom elements - it exports parts and pieces intended to be composed into Web Components, allowing you to implement your own design language by simply applying CSS styles and behaviors without having to write all the JavaScript that's involved in building production-quality component implementations.

The Web Component classes defined in `fast-foundation` should provide the base functionality of a given component without the prescriptive semantics specific to a give design language. While a Fluent button may have defined appearances, these would be defined when the button is composed, rather than on the primitive. In the event I wanted to create a "navigation-item" component, I may not have a need for those appearances. This should help enable convergence as we can focus our efforts in FAST on supporting much of the base functionality for these components, while leaving more prescriptive attributes and choices up to the implementor - in this case, our work here in FluentUI. We follow this pattern for our FAST components and would be committed to preserving that here. From our conversations with the Fluent UI team, we believe we're aligned here 🎉 .

**fast-components:** This is the last layer we export, in addition to the `fast-components-msft` which we are proposing to move over here. This is a library of Web Components that composes the exports of `fast-foundation` with stylesheets aligning to the FAST design language, whereas the MSFT package is more closely aligned with Microsoft's design language.

_NOTE: The `fast-components` package would not be a dependency brought in, it is simply our own composed components with our FAST design system and denoted here for awareness of how we build components._

## Existing Package

As mentioned above, to start, I'd like to raise a PR to move the existing `fast-components-msft` package into the repo and have our team continue our work here. This will create some awareness, transparency, and help us start working more closely towards convergence across our parts, tokens, anatomy, etc. The current package takes a dependency on `fast-element` and `fast-foundation` and provides the following primitives:

- accordion
- anchor
- badge
- button
- card
- checkbox
- design-system-provider ("Fluent" design system)
- dialog
- divider
- flipper
- menu
- menu-item
- progress
- radio
- radio-group
- slider
- slider-label
- switch
- tabs
- text-area
- text-field
- tree-view
- tree-item

We are currently tracking several others and actively working on select. That work would continue here once we migrate the package into the FluentUI repository. As part of our work on convergence, we're happy to explore and land a new thoughtful name for the package that aligns with the FluentUI repo. For the time being, to prevent regressions and churn for existing partners, we would request that no changes be made as part of the move.

**What about non-primitives which currently exist in FluentUI:** 
My personal hope is that we begin to build these out as well over time. Moving to FluentUI will give us the opportunity to create more bespoke *Fluent* controls which don't necessarily fit within the model we have in FAST. Once we land the final primitives I hope we can move towards more complex and composed patterns that exist elsewhere in Fluent.

### Ownership

As our team remains SME's on FAST and the patterns for building web components, and since we have a deep knowledge of the requirements of those currently leveraging this package, we would request to be the sole owners of this package for the time being. We have an existing model for bringing on new codeowners and would be happy to evaluate that process with the team here. My hope is that folks excited about the effort would contribute and we would have the opportunity to include those making valuable contributions to the package over time as Code Owners, even replacing folks from our side as owners.

## Final Thoughts

We are excited about the possibility of moving the `fast-components-msft` package here and providing a FluentUI Web Component offering for those inside and outside Microsoft. We believe this benefits FAST, FluentUI, and ultimately the web development community. I'm excited to open this RFC. Please do feel free to add any questions or comments you have!

Thanks,

The FAST Team

chrisdholt:
#### Pull request checklist
closes #14071 

#### Description of changes
This PR Closes the RFC to add web components to the repository. A few fixes do accompany this change, but as far as this repo is concerned it is net-new.


packages/web-components/package.json
```diff
@@ -0,0 +1,100 @@
...
+    "test-chrome:verbose": "karma start karma.conf.js --browsers=ChromeHeadlessOpt --single-run --coverage --reporter=mocha",
+    "test-chrome:watch": "karma start karma.conf.js --browsers=ChromeHeadlessOpt --coverage --watch-extensions js",
+    "test-chrome:debugger": "karma start karma.conf.js --browsers=ChromeDebugging",
+    "test-chrome:verbose:watch": "karma start karma.conf.js --browsers=ChromeHeadlessOpt --coverage --watch-extensions js --reporter=mocha ",
+    "test-chrome:verbose:debugger": "karma start karma.conf.js --browsers=ChromeDebugging --reporter=mocha",
+    "test-firefox": "karma start karma.conf.js --browsers=FirefoxHeadless --single-run --coverage",
+    "test-firefox:verbose": "karma start karma.conf.js --browsers=FirefoxHeadless --single-run --coverage --reporter=mocha"
+  },
+  "devDependencies": {
+    "@microsoft/api-extractor": "^7.7.13",
```
ecraig12345, line 48: Having different versions of build and types dependencies between this package and the rest of the repo is going to cause some headaches:
- We use [syncpack](https://www.npmjs.com/package/syncpack) to validate that all packages have matching dependency versions (otherwise we're relying on people noticing duplicates in yarn.lock, and things regularly used to sneak through that way and cause problems later). Unfortunately, syncpack currently has no option to ignore a particular package or allow multiple versions for certain deps, so we'll need to either find a different tool, or add this feature to syncpack (original or fork).
- Largely due to the common practice of type packages using `*` as the version spec when depending on other type packages, if multiple versions of the same types are installed, it can be unpredictable which one will be picked up
- I think having multiple versions of typescript or other dependencies around can also cause problems (don't remember specifics as I haven't had to deal with this for awhile)

In some cases, we may be able to resolve the dependency mismatches without too much trouble, but in other cases it will require work in either web-components or the entire repo to handle a nontrivial upgrade or downgrade.

packages/web-components/package.json
```diff
@@ -0,0 +1,100 @@
...
+    "name": "Microsoft",
+    "url": "https://discord.gg/FcSNfg4"
+  },
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Microsoft/fast.git"
+  },
+  "bugs": {
+    "url": "https://github.com/Microsoft/fast/issues/new/choose"
```
ecraig12345, line 16: Update these?

packages/web-components/package.json
```diff
@@ -0,0 +1,100 @@
...
+  },
+  "main": "dist/esm/index.js",
+  "types": "dist/fast-components-msft.d.ts",
+  "unpkg": "dist/fast-components-msft.min.js",
+  "scripts": {
```
ecraig12345, line 21: For this package to build properly with the rest of the repo, there may be some modifications needed to the scripts and/or to the CI job. 

We use `lerna run` to run the following scripts in each package:
- `clean`
- `build`: core build stuff, usually running TS but not bundling (though this could vary by package)
- `bundle`: runs webpack if appropriate (not all packages have this)
- `lint`: runs eslint
- `prettier`: runs prettier (*without* `--write` option by default)
- `test`: runs tests (mostly unit tests)

If there are any other scripts that you need run as part of the PR/CI job, `azure-pipelines.yml` will need to be modified to handle that.

packages/web-components/package.json
```diff
@@ -0,0 +1,100 @@
...
+    "clean:dist": "node ./build/clean.js dist",
+    "doc": "api-extractor run --local",
+    "doc:ci": "api-extractor run --local",
+    "build": "tsc -p ./tsconfig.json && rollup -c && npm run doc",
+    "dev": "tsc -p ./tsconfig.json -w",
+    "tdd": "npm run dev & npm run test-chrome:watch",
+    "prepare": "yarn clean:dist && yarn build",
+    "prettier": "prettier --config ../../prettier.config.js --write \"**/*.{ts,html}\"",
```
ecraig12345, line 29: Looks like there are some differences between the prettier config of the incoming files and the one for the whole repo (2 vs 4 space most obviously). If you're okay with reformatting it would be nice to have consistency, or if not can you add an override to the root `prettier.config.js`?

lerna.json
```diff
@@ -1,5 +1,11 @@
 {
-  "packages": ["apps/*", "packages/*", "scripts", "packages/fluentui/*"],
+  "packages": [
```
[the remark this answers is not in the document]
ecraig12345: Or if we isolate the package for now, the glob can change to `packages/!(web-components)/*`.

Will need to be updated in the root package.json under `workspaces.packages` too.

packages/web-components/.prettierignore
```diff
@@ -0,0 +1,3 @@
+coverage/*
```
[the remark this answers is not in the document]
chrisdholt: I'll update the root file if need be.

Regarding ignoring the spec files, I'll need to go check those once again but part of the issue was that prettier formatting was changing the expected output and causing tests to fail.

packages/web-components/package.json
```diff
@@ -0,0 +1,100 @@
+{
+  "name": "@microsoft/fast-components-msft",
```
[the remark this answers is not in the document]
chrisdholt: Currently, per the issue in reference this package has been published already and there are projects which depend on this both internally at Microsoft and externally, so we can't change the name without consequence. The long term plan is to have this live under the fluent namespace, but that will require planning and some coordination. While I agree in principle of having the name match the package, that's really only possible if we want to leverage the `fast` namespace here; though, I think that would cause more confusion. I'll ping @levithomason as FYI here as we discussed the organization under "web-components" the other day.

packages/web-components/package.json
chrisdholt: I'll try to reconcile these, but one we won't be able to reconcile is Typescript. There are ways to prevent hoisting of certain packages to the root which in the case of Typescript may be beneficial for the short term.

packages/web-components/package.json
chrisdholt: I ran prettier and this references the Fluent project config. It could be a different version of prettier that's causing the issues. I'll take a look.

packages/web-components/package.json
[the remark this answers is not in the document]
chrisdholt: @ecraig12345 I think we need to work on integrating releases. We will likely need to prioritize that discussion, but in the meantime - are we able to release manually within the scope of the package? I have permissions to publish to this package at that namespace.

packages/web-components/README.md
```diff
@@ -0,0 +1,57 @@
+# FAST Components MSFT
```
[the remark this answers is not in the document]
chrisdholt: As mentioned above and in the RFC, we cannot change the package name. As this pertains to the use and installation of this via npm or CDN, I believe this is all still accurate.

packages/web-components/.eslintrc.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
```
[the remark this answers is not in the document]
chrisdholt: I'd tend to generally agree, I'm happy to create a follow-up issue to reconcile this, though some of it may be contentious depending on the rules 😆

packages/web-components/package.json
chrisdholt: *most* of these have been reconciled. I did add a flag to prevent hoisting of the TS related dependencies here, so that should leave everything else undisturbed and install those locally for use in this package: https://classic.yarnpkg.com/blog/2018/02/15/nohoist/

packages/web-components/package.json
chrisdholt: Updated the task to `clean` from `clean:dist` and `eslint` to `lint`. I've dropped prettier as it doesn't seem to be defined for any of the other packages and running at the root accomplishes what's needed. While we have several tests denoted in here, the primary task is labelled `test`. Hopefully that reconciles where need be!

packages/web-components/package.json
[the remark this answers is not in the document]
chrisdholt: Another question here - if we need to publish separately from the current system, do I need to opt out of the change requests to pass the build? If so, is that acceptable to do as part of this PR?

EisenbergEffect:
While I appreciate the desire to fully align all deps, linting, etc. I don't think that should be a requirement in the short-term (and for some things, perhaps never). It's important to note that we're building a tech stack from the ground up, with no dependency on React. As such, we have different patterns than would be expressed in React components.

A few examples:

* Some of our lint rules are configured to facilitate FASTElement patterns, because the way we implement components is significantly different from React.
* The web components community uses different libraries and patterns for testing than React. For example, Jest is not typically used.
* I'm working with tc39 and TypeScript on new JS language features, some of which are based on web components patterns. So, I want to be free to leverage a particular version of TS and particular settings of the compiler based on that for our work on FASTElement, which will trickle down to the Fluent web components.

yarn.lock
```diff
@@ -5118,14 +4960,8 @@ ansi-align@^2.0.0:
...
-ansi-align@^3.0.0:
-  version "3.0.0"
-  resolved "https://registry.yarnpkg.com/ansi-align/-/ansi-align-3.0.0.tgz#b536b371cf687caaef236c18d3e21fe3797467cb"
-  integrity sha512-ZpClVKqXN3RGBmKibdfWzqCY4lnjEuoNzU5T0oEFpfd/z5qJHVarukridD4juLO2FXMiwUQxr9WqQtaYa8XRYw==
-  dependencies:
-    string-width "^3.0.0"
-
-ansi-colors@3.2.3:
+ansi-align@^3.0.0, ansi-colors@3.2.3:
+  name ansi-align
```
chrisdholt: @ecraig12345 If this was a manual change I believe it's broken the repo from being able to be built. I'll need to try and revert (at least) this change. Were there any other manual changes to the lockfile being made as part of this? Is manually modifying the yarn.lock file a common practice? My concern would be the introduction of breaks fairly easily be messing with how yarn resolves and manages dependencies.

ecraig12345: Yeah...that part was a mistake, sorry. Yarn probably warned about it but it got lost in the sea of [spurious missing peer dependency warnings](https://github.com/yarnpkg/yarn/issues/5810). I just pushed a fix. (The fix process is to delete the invalid entry and run `yarn` to let it be regenerated.)

To the other part of your question, there were quite a few manual changes because yarn tends to resolve new dependencies in a way that results in often-unnecessary duplicates (like we saw with babel and storybook), and I'm not sure how to completely work around it without manual changes.

For example, say you have some existing (probably indirect) dependency on `foo@^1.1.0` which yarn resolved to `foo@1.1.0` and put in the lockfile. Then something else adds a dependency (again, probably indirect) on `foo@^1.2.0` . Instead of changing both to resolve to `foo@1.2.0`, which satisfies both semver specs, it adds a *separate* entry/resolution just for the new spec. (IIRC this can even happen with *older* semver specs: a new indirect dep on `foo@^1.0.0` will also result in a separate resolution being added to the *newer* `foo@1.2.0`.)

I'm pretty sure this is an intentional behavior done for safety purposes (at least avoiding the implicit upgrade of the original `foo@^1.1.0` to `foo@1.2.0`...adding a separate resolution to a newer version for `foo@^1.0.0` just seems weird) but is less than ideal from an install time/size perspective--especially in cases like babel or storybook stuff that has a bunch of dependencies and you get a cascading effect.

Regarding how to validate manual changes, yarn does warn on blatantly invalid things like this, so I should have looked more closely through the sea of peer dep errors to check for any real issues. Beyond that, generally as long as the repo builds and tests pass afterwards, it's probably fine. Also I try and avoid messing with direct dependencies where version upgrades may be particularly impactful, but we usually try and lock those down to a pretty specific version (`~` or exact) to begin with.

ecraig12345: It's probably possible to resolve some of this without manual changes using [`yarn upgrade`](https://classic.yarnpkg.com/en/docs/cli/upgrade) or [`yarn upgrade-interactive`](https://classic.yarnpkg.com/en/docs/cli/upgrade-interactive), and maybe I should try that first in the future instead. Both take an optional package name or scope, so it may be possible to do something similar to what I was doing manually with `yarn upgrade-interactive`.

ecraig12345:
> While I appreciate the desire to fully align all deps, linting, etc. I don't think that should be a requirement in the short-term (and for some things, perhaps never). It's important to note that we're building a tech stack from the ground up, with no dependency on React. As such, we have different patterns than would be expressed in React components.

Chris and I have been talking about this and resolved or worked around most of the issues for now.

There are a few different reasons for wanting dependency alignment where reasonable, primarily revolving around maintainability. (You have some good points about exceptions, which I responded to below.)

- When something breaks and blocks everyone's PRs, who's responsible for fixing it? At the moment in this repo, that often falls to me. Adding yet another build stack (especially one which the primary repo owners are less familiar with) increases the possibility of breaks and complexity of fixing them. We need to at least have a clearly communicated plan for addressing this situation, with backups in case the primary person is OOF.

- I've seen instances in the past where having different tool versions in the same repo causes hard-to-diagnose issues. One place I know this can come up is with type packages, since they typically use a `*` version spec when depending on other type packages. Pretty sure I've seen other issues like this too (including maybe with typescript), but unfortunately it's been long enough that I don't remember specifics--hopefully the `nohoist` settings Chris added for particular dependencies will help avoid this.

- We use [syncpack](https://www.npmjs.com/package/syncpack) to avoid accidentally having multiple versions of the same dependency, since that's been a problem in the past. However it's not very configurable currently: either on or off per package; can't vary only some of the deps. For now we fixed some of the mismatches but temporarily turned off `syncpack` checks for `web-components`. (I also made a [feature request](https://github.com/JamieMason/syncpack/issues/41) to enable varying individual deps, and might help implement since it would be useful to us in various scenarios.

- Trying to keep install time (and size) from getting even worse

Regarding the specific cases you mentioned where not aligning dependencies may make sense:

> * Some of our lint rules are configured to facilitate FASTElement patterns, because the way we implement components is significantly different from React.

This is a good point, but I do still think that long-term, it would be nice to align our non-stack-specific rules if possible (*definitely* not blocking on that now). Having significantly different rules within the same repo could be confusing to contributors, unless there's no overlap between who's contributing to which packages. Though I guess you could also argue that having different rules in the web components package and the main FAST repo would confuse developers moving between the two of those, so maybe the question is where there will be more overlap.

> * The web components community uses different libraries and patterns for testing than React. For example, Jest is not typically used.

Agreed, this is an area where aligning is not really necessary.

> * I'm working with tc39 and TypeScript on new JS language features, some of which are based on web components patterns. So, I want to be free to leverage a particular version of TS and particular settings of the compiler based on that for our work on FASTElement, which will trickle down to the Fluent web components.

I definitely understand wanting to be able to pick up new features in a way that suits your requirements, as well as sometimes wanting to experiment with new versions in selective packages (could see us wanting that with react components sometimes too). But if we do start sharing code between web component and React packages, any package with shared code will always have to use the "lowest common denominator" TS version to avoid breaks. (There's also the potential for other breaks due to multiple TS versions, which I *really* wish I remembered the details of better...)

chrisdholt:
> I definitely understand wanting to be able to pick up new features in a way that suits your requirements, as well as sometimes wanting to experiment with new versions in selective packages (could see us wanting that with react components sometimes too). But if we do start sharing code between web component and React packages, any package with shared code will always have to use the "lowest common denominator" TS version to avoid breaks. (There's also the potential for other breaks due to multiple TS versions, which I _really_ wish I remembered the details of better...)

@ecraig12345 Thanks for the response above. I won't speak to everything there, but I do want to zero in on this specific final paragraph. 

While I understand the goals of keeping dependencies shared in an effort to minimize conflicts, keep build times down, etc - we can't move forward under the premise of using the "lowest common denominator". Web components have a specific browser matrix which doesn't include many of the down-level browsers which need to be supported by Fluent React. The requirements are fundamentally different and in order to offer the best possible web component solution, we can't build for the lowest common denominator. In many cases, these are browsers which will never have support for custom elements, decorators, css custom properties, etc. Ultimately, to provide the best possible web component experience, I don't think we can have a requirement that we share code with other packages in this repo. I'm totally on board with a goal of shared code, but that can't eclipse the goals of the Web Component package. Again, I totally understand the goals and intent with sharing dependencies and code across the mono repo - but from a core requirement standpoint, the WC's need to be able to stand alone. Ultimately, the web component work can't be held back for the lowest common denominator.

ecraig12345:
One quick clarification is that the "lowest common denominator" comment only applies to the TS version of any hypothetical future packages which contain code shared between different tech stacks (if/when we reach that point), not really anything about browser support, and wouldn't be a concern if there are never shared-code packages.

chrisdholt:
@ecraig12345 super odd - but the test failures I saw in master on my mac don't repro when pushing :). Looks like all checks passed if you want to take another look. + @dzearing as FYI.

dzearing:
Woohoo!

BrunoBlanes:
Now that the web components have been moved here, I was wondering if there is going to be an in-depth documentation for how to properly use it as with all the other Fluent UI packages. The FAST documentation isn't really as explanatory as Fluent's.

chrisdholt:
Hey @BrunoBlanes - thanks for this. We're working with the team to coordinate a shared documentation story for these. I'll also write up an issue here as I think we'll need a temporary solution as well. Are there specific things you would like to see prioritized while we work to coordinate the docs story with the existing Fluent solutions?

BrunoBlanes:
Yes, it is a bit hard to use JavaScript with Blazor, so it makes it even harder to apply custom styling to components. For example, Fluent UI React allows me to set a custom theme, using the Theme Designer and the `createTheme()` function, to all my components:

```typescript
const myTheme = createTheme({
  palette: {
    themePrimary: '#0078d4',
    [...]
  }});
```
With FAST Design I haven't yet figured out how to do customizations on a general level.

Blazor also doesn't support NPM so we have to rely on CDN scripts and I couldn't get it to work with FAST Design using TypeScript. The compiled JS version of Fluent UI's tutorials is much less complex then FAST Design.

Also I've been using codepen.io to better understand what I have to do to use Fluent UI in Blazor and the direct demo links just saves a lot of time, those aren't available on FAST Design's documentation.

I think improving the Blazor specific documentation of both FAST Design and Fluent UI is very important given how much Blazor is expected to grow and also taking into consideration that there aren't many UI libraries available to use with Blazor right now let alone consistent ones like Fluent.

PR #14206, opened 2020-07-25
Card: Initial implementation

khmakoto:
#### Pull request checklist

#### Description of changes

This PR contains the first implementation of a `Card` component using our latest stack of technologies.

image

#### Notes

- It's using `office-ui-fabric-react` components for the examples.
- There are a couple of TODOs regarding some props that I still need to address.


khmakoto:
There's an open question about what to do with the `react-cards` package and with the old `Card` before we're able to ship this. Hence why I'm not exporting the `Card` here yet and why this is under a `Card2` folder.

packages/react-cards/src/components/Card2/CardBase.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const CardBase = compose<'div', CardProps, CardProps, {}, {}>(
+  (props, ref, options) => {
+    const { state } = options;
+    const { slots, slotProps } = mergeProps<CardProps, CardProps, CardSlots, CardSlotProps>(state, options);
+
+    const { children } = state;
+
+    return (
+      <slots.root ref={useMergedRefs(ref)} {...slotProps.root}>
```
layershifter, line 15:
```suggestion
      <slots.root ref={ref} {...slotProps.root}>
```

?

packages/react-cards/src/components/Card2/useCardBehavior.ts
```diff
@@ -0,0 +1,36 @@
+import { getCode, EnterKey, SpacebarKey } from '@fluentui/keyboard-key';
+import { CardProps, CardState } from './Card.types';
+
+export const useCardBehavior = (props: CardProps): Partial<CardState> => {
+  const { disabled, onClick: _onClick, onKeyDown: _onKeyDown } = props;
+
+  let onClick: ((ev: React.MouseEvent<HTMLDivElement>) => void) | undefined = undefined;
+  if (_onClick) {
```
layershifter, line 8: Is it a premature optimization?...

packages/react-cards/src/components/Card2/useCardBehavior.ts
```diff
@@ -0,0 +1,36 @@
...
+        const eventCode = getCode(ev);
+        if (eventCode === EnterKey || eventCode === SpacebarKey) {
+          onClick(ev as any);
+        }
+      }
+    }
+  };
+
+  return {
+    'aria-disabled': disabled,
```
layershifter, line 30:
```suggestion
    'aria-disabled': disabled,
    role: 'group',
```

In Northstar there is `role` attr, see `cardBehavior`. Should we reuse it?

packages/react-cards/src/components/Card2/useCardBehavior.ts
```diff
@@ -0,0 +1,36 @@
...
+    tabIndex: disabled ? undefined : 0,
+    ...props,
+    onClick,
+    onKeyDown,
```
layershifter, line 34: `Card`s in Northstar are not focusable/actionable by default, only with `cardFocusableBehavior`. Is it intentional?

khmakoto: We can make them focusable only when `onClick` has been provided. I'll make the change.

packages/react-cards/src/components/Card2/useCardBehavior.ts
khmakoto: Added

packages/react-cards/src/next/Card.types.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export interface CardState extends CardProps {
+  cardRef: React.RefObject<HTMLDivElement>;
+}
+
+export interface CardSlots extends BaseSlots {}
+
+export type CardSlotProps = SlotProps<CardSlots, CardProps, React.HTMLAttributes<HTMLDivElement>>;
+
+export type CardTokens = ColorTokenSet & {};
```
dzearing, line 55: in my variants PR i am renaming `ColorTokenSet` to just `ColorTokens`. Just a heads up. If you merge first here, i'll get this updated.

dzearing (approved):

packages/react-cards/package.json
```diff
@@ -48,10 +49,20 @@
     "react-test-renderer": "^16.3.0"
   },
   "dependencies": {
+    "@fluentui/keyboard-key": "^0.2.12",
+    "@fluentui/react-compose": "^0.19.3",
+    "@fluentui/react-theme-provider": "^0.13.3",
+    "@fluentui/theme": "^0.3.2",
     "@microsoft/load-themed-styles": "^1.10.26",
+    "@uifabric/azure-themes": "^7.5.13",
```
dzearing, line 57: wait, this should not be a dependency.

dzearing: check the others as well.

khmakoto: Spoke about this offline with David. A lot of this cleanup is already being done in #15132.

PR #14213, opened 2020-07-27
Grouped components in left-side menu into a Tree

vyhnalekl:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/fluentui/react-builder/src/config.tsx
```diff
@@ -12,6 +12,63 @@ type FiberNavigator = FUI.FiberNavigator;
...
+  'Content',
+  'Layout',
+  'Forms',
+  'Actionable',
+  'Containers',
+  'Utilities',
+  'Other',
+];
+
+export const GET_COMPONENT_GROUP_INDEX = {
```
miroslavstastny, line 26: This is fragile and hard to maintain, I would prefer single object:

```jsx
{
  Surfaces: [Popup, Dialog, ...],
  ...
}
```

packages/fluentui/react-builder/src/components/List.tsx
```diff
@@ -37,6 +37,74 @@ export const List: React.FunctionComponent<ListProps> = ({ onDragStart, style })
...
   );
 
+  const titleRenderer = (Component, { content, expanded, open, hasSubtree, ...restProps }) => (
+    <Component expanded={expanded} hasSubtree={hasSubtree} {...restProps}>
+      {hasSubtree ? expanded ? <TriangleDownIcon /> : <TriangleEndIcon /> : null}
+      {content}
+    </Component>
+  );
+
+  function addComponentToGroup(info, property) {
```
assuncaocharles, line 47:
```suggestion
 const  addComponentToGroup(info, property) => {
```

packages/fluentui/react-builder/src/components/List.tsx
```diff
@@ -37,6 +37,74 @@ export const List: React.FunctionComponent<ListProps> = ({ onDragStart, style })
...
+            </div>
+          )}
+
+          {displayMode === 'Display Name' && info.displayName}
+        </Box>
+      ),
+    });
+  }
+
+  const componentGroupItems = {};
```
assuncaocharles, line 80: Can you move it above the `addComponentToGroup` function since it's being used there?

packages/fluentui/react-builder/src/components/List.tsx
```diff
@@ -37,6 +37,74 @@ export const List: React.FunctionComponent<ListProps> = ({ onDragStart, style })
     },
   );
 
+  const titleRenderer = (Component, { content, expanded, open, hasSubtree, ...restProps }) => (
```
assuncaocharles, line 40: I would probably avoid using `titleRenderer` I think in the future we want to deprecate it, you can achieve the same using `as` prop in the `addComponentToGroup` method:

```javascript
  const addComponentToGroup = (info, property) => {
    componentGroupItems[property].push({
      id: info.displayName,
      as: ({ content, expanded, open, hasSubtree, children, ...restProps }) => {
        return (
          <>
            {hasSubtree ? expanded ? <TriangleDownIcon /> : <TriangleEndIcon /> : null}
            {children}
          </>
        );
      },
```

packages/fluentui/react-builder/src/components/List.tsx
```diff
@@ -37,6 +37,74 @@ export const List: React.FunctionComponent<ListProps> = ({ onDragStart, style })
...
+    <Component expanded={expanded} hasSubtree={hasSubtree} {...restProps}>
+      {hasSubtree ? expanded ? <TriangleDownIcon /> : <TriangleEndIcon /> : null}
+      {content}
+    </Component>
+  );
+
+  const componentGroupItems = {};
```
assuncaocharles, line 47:
```suggestion
  const componentGroupItems = Object.keys(COMPONENT_GROUP).reduce((acc, property) => ({...acc, property: []}), {});
```

packages/fluentui/react-builder/src/components/List.tsx
```diff
@@ -37,6 +37,74 @@ export const List: React.FunctionComponent<ListProps> = ({ onDragStart, style })
...
+    });
+  };
+
+  const componentTreeItems = [];
+  for (const property in COMPONENT_GROUP) {
+    componentGroupItems[property] = [];
+    COMPONENT_GROUP[property].forEach(item =>
+      supportedComponents.forEach((o, index) => {
+        if (o.displayName === item || (o.isChild && o.parentDisplayName === item)) {
+          delete supportedComponents[index];
```
assuncaocharles, line 87: Do we really need to delete it?

packages/fluentui/react-builder/src/components/List.tsx
```diff
@@ -37,6 +37,74 @@ export const List: React.FunctionComponent<ListProps> = ({ onDragStart, style })
...
+        </Box>
+      ),
+    });
+  };
+
+  const componentTreeItems = [];
+  for (const property in COMPONENT_GROUP) {
+    componentGroupItems[property] = [];
+    COMPONENT_GROUP[property].forEach(item =>
+      supportedComponents.forEach((o, index) => {
```
assuncaocharles, line 85: Is `o` the component info or the component itself here?

packages/fluentui/react-builder/src/components/List.tsx
```diff
@@ -37,6 +37,74 @@ export const List: React.FunctionComponent<ListProps> = ({ onDragStart, style })
...
+        }
+      }),
+    );
+
+    componentTreeItems.push({
+      id: property,
+      title: property,
+      items: componentGroupItems[property],
+    });
+  }
```
assuncaocharles, line 98: _NIT_

```suggestion
    Object.entries(COMPONENT_GROUP).forEach(([property, items]) => {
    componentGroupItems[property] = [];

    items.forEach(item =>
      supportedComponents.forEach((o, index) => {
        if (o.displayName === item || (o.isChild && o.parentDisplayName === item)) {
          delete supportedComponents[index];
          addComponentToGroup(o, property);
        }
      }),
    );

    componentTreeItems.push({
      id: property,
      title: property,
      items: componentGroupItems[property],
    });
  });
```

packages/fluentui/react-builder/src/components/List.tsx
vyhnalekl: I delete the items because latter on I add all the remaining items into the "Others" group. I could set it to some value like undefined, that might be better?

packages/fluentui/react-builder/src/components/List.tsx
vyhnalekl: I will change the name to info

assuncaocharles:
Just double check with @miroslavstastny

PR #14216, opened 2020-07-27
Adding useResponsiveMode hook inside of office-ui-fabric-react utilities

czearing:
#### Pull request checklist

#### Description of changes
1. Added a new hook called `useResponsiveMode` inside of  office-ui-fabric-react's utilities.
2. Added tests for hook.
3. Moved `withResponsiveMode` decorator into separate folder called `responsiveMode` within utilities.
4. Updating `withResponsiveMode` imports.
5. Added a index file within `responsiveMode`.
6. Added a types file within `responsiveMode`.

#### Focus areas to test
useResponsiveMode


ecraig12345:
I'm not sure if we want to do this--if I remember right there was some discussion in the past of `withResponsiveMode` not being recommended for external use (it's intentionally never been exported from OUFR) and maybe even being semi-deprecated, but I don't remember why. @dzearing do you remember? Was the issue just that it's a decorator and we've moved away from that pattern, or something more fundamental?

dzearing:
@ecraig12345 It's because it was written as a decorator. @czearing was blocked on converting a few components which rely on it (Modal, Dropdown, others), so I recommended making a PR to convert it to a hook before completing those. I think it's a reasonable thing. I think the breakpoints really should be configurable through theme though, but to unblock the function component conversion I think this is important.

ecraig12345:
@dzearing Sounds good then; I'll look over it.

packages/office-ui-fabric-react/src/utilities/responsiveMode/withResponsiveMode.tsx
```diff
@@ -1,21 +1,8 @@
 import * as React from 'react';
 import { findDOMNode } from 'react-dom';
-import { BaseDecorator } from './BaseDecorator';
+import { BaseDecorator } from '../decorators/BaseDecorator';
```
ecraig12345, line 3: I don't think this file should be moved. Because this wasn't exported from the root, people **are** importing it by deep path (@dzearing if you search ODSP code you can find *many* specific examples in active projects), and this change would break them.

packages/office-ui-fabric-react/src/utilities/responsiveMode/useResponsiveMode.tsx
```diff
@@ -0,0 +1,72 @@
...
+ */
+let _defaultMode: ResponsiveMode | undefined;
+
+/**
+ * Tracking the last mode we successfully rendered, which allows us to
+ * paint initial renders with the correct size.
+ */
+let _lastMode: ResponsiveMode | undefined;
+
+const getResponsiveMode = (currentWindow: Window | undefined): ResponsiveMode => {
```
ecraig12345, line 19: Can any of this be shared with `withResponsiveMode` for now? The logic is complex enough that we should avoid duplicating it.

packages/office-ui-fabric-react/src/utilities/responsiveMode/useResponsiveMode.tsx
```diff
@@ -0,0 +1,72 @@
...
+      setLastResponsiveMode(newResponsiveMode);
+    }
+  }, [elementRef, lastResponsiveMode]);
+
+  useOnEvent(window, 'resize', onResize as (ev: Event) => void);
+
+  // Call resize function initially on mount.
+  React.useEffect(() => {
+    onResize();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ecraig12345, line 68: It's good to include the reason for disabling the rule inline. Note that the `--` surrounded by spaces is specifically required by ESLint.
```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps -- only meant to run on mount
```

packages/office-ui-fabric-react/src/utilities/responsiveMode/useResponsiveMode.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const onResize = React.useCallback(() => {
+    // Setting the same value should not cause a re-render.
+    const newResponsiveMode = getResponsiveMode(getWindow(elementRef.current));
+
+    if (lastResponsiveMode !== newResponsiveMode) {
+      setLastResponsiveMode(newResponsiveMode);
+    }
+  }, [elementRef, lastResponsiveMode]);
+
+  useOnEvent(window, 'resize', onResize as (ev: Event) => void);
```
ecraig12345, line 63: Can you just add an `ev: Event` parameter to the callback instead, even though it will be unused?

PR #14222, opened 2020-07-28
Hackathon: Make UI Builder a bit more accessible

jurokapsiar:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/fluentui/react-builder/src/components/Canvas.tsx
```diff
@@ -114,6 +117,8 @@ export const Canvas: React.FunctionComponent<CanvasProps> = ({
       return () => null;
     }
 
+    role && iframe.setAttribute('role', role);
```
assuncaocharles, line 120: _nit_

```suggestion
    if(role){
        iframe.setAttribute('role', role);
    } 
```

packages/fluentui/react-builder/src/components/Designer.tsx
```diff
@@ -185,8 +196,48 @@ const stateReducer: Reducer<DesignerState, DesignerAction> = (draftState, action
...
+          where === 'after' && index++;
+        }
+      }
 
+      resolveDrop(element, parent, index);
+
+      draftState.selectedJSONTreeElementUuid = element.uuid;
+      draftState.selectedComponentInfo = componentInfoContext.byDisplayName[element.displayName];
+      treeChanged = true;
+      setTimeout(() => focusTreeTitle(element.uuid));
```
assuncaocharles, line 238: In theory reducer should be free of site effect and only manipulate the state, I wonder if there would be better place for this mutations.

packages/fluentui/react-builder/src/components/Designer.tsx
```diff
@@ -392,6 +463,13 @@ export const Designer: React.FunctionComponent = () => {
         overflow: 'hidden',
       }}
     >
+      {insertComponent && (
+        <InsertComponent
+          onDismiss={handleCloseAddComponentDialog}
+          onComponentAdded={handleAddComponent}
+          // setTimeout(() => focusTreeTitle(addedComponent.uuid));
```
assuncaocharles, line 470:
```suggestion
```

packages/fluentui/react-builder/src/components/ReaderText.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const ReaderText: React.FunctionComponent<ReaderTextProps> = ({ selector }) => {
+  const ref = React.createRef<HTMLElement>();
+  const [text, setText] = React.useState('');
+
+  React.useEffect(() => {
+    if (ref.current) {
+      const t = ref.current.ownerDocument.querySelector(selector)?.textContent;
+      setText(t);
+    }
+  }, [setText, ref, selector]);
```
assuncaocharles, line 17:
```suggestion
  }, [setText, ref.current, selector]);
```

packages/fluentui/react-builder/src/components/ReaderText.tsx
```diff
@@ -0,0 +1,28 @@
...
+export type ReaderTextProps = {
+  selector: string;
+};
+
+export const ReaderText: React.FunctionComponent<ReaderTextProps> = ({ selector }) => {
+  const ref = React.createRef<HTMLElement>();
+  const [text, setText] = React.useState('');
+
+  React.useEffect(() => {
+    if (ref.current) {
```
assuncaocharles, line 13: Should `selector` be part of the condition?

packages/fluentui/react-builder/src/components/Designer.tsx
jurokapsiar: is dispatching another event from the handler possible? if yes, we can just have another handler to focus

packages/fluentui/react-builder/src/components/ReaderText.tsx
jurokapsiar: it could, but since this is typescript, it should be checked during build time, I hope

PR #14223, opened 2020-07-28
feat(builder): download/upload

assuncaocharles:
#### Pull request checklist

#### Description of changes

This PR adds the options to `download` or `upload` json tree for UI builder

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-builder/src/components/Designer.tsx
+++ b/packages/fluentui/react-builder/src/components/Designer.tsx
@@ -31,6 +31,7 @@ import { ComponentTree } from './ComponentTree';
 import { GetShareableLink } from './GetShareableLink';
 import { codeToTree } from '../utils/codeToTree';
 import { ErrorBoundary } from './ErrorBoundary';
+import { DownloadIcon } from '@fluentui/react-icons-northstar';
 
 const HEADER_HEIGHT = '3rem';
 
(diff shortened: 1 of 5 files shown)


packages/fluentui/react-builder/src/components/Designer.tsx
```diff
@@ -166,6 +168,12 @@ const stateReducer: Reducer<DesignerState, DesignerAction> = (draftState, action
       treeChanged = true;
       break;
 
+    case 'UPLOAD_TREE':
+      draftState.jsonTree = action.jsonTree;
```
assuncaocharles, line 172: @jurokapsiar @miroslavstastny Question for all the state change: Shouldn't we be returning a new state instead of changing it directly?

```
return {...draftState, jsonTre: action.jsonTree}
```

jurokapsiar: looking at https://www.smashingmagazine.com/2020/06/better-reducers-with-immer/, they do both - sometines construct new state, sometimes mutate the old one...

miroslavstastny: always mutate the draft state and return nothing. that's the purpose of immer.

packages/fluentui/react-builder/src/components/Designer.tsx
```diff
@@ -270,6 +278,14 @@ export const Designer: React.FunctionComponent = () => {
     }
   }, [dispatch]);
 
+  const handleUpload = React.useCallback(
+    jsonTree => {
+      console.log('jsonTree: ', jsonTree);
```
miroslavstastny, line 283:
```suggestion

```

dispatch logs the action

PR #14224, opened 2020-07-28
fix(UIBuilder) maximize view

miroslavstastny:
#### Description of changes

UIBuilder Maximize view failed to load because of missing babel.
Babel is not required for the maximize view but was a hard dependency in `Designer` - it is now lazily loaded only when code editor is opened.

#### Focus areas to test

- Code editor in UI Builder
- Maximize view


miroslavstastny:
It fails to load from time to time, investigating...

PR #14228, opened 2020-07-29
Adding compose updates

dzearing:
The current `compose` helper has a lot of complexities that make it less simple than it should be. This PR adds a new version of `compose` to the `/next` folder. I did this explicitly to not change any of the existing code, so that we can play test these changes first in other components before committing. If it works well, and we want to move with it, then I will add more tests and replace the existing one. Otherwise, I'll delete it.

We all agree that components should be built from smaller building blocks; atomic hooks, behavior/state hooks, styling hooks, render functions, etc. None of this is contentious. None of this code prevents anyone from creating components from scratch, or reusing the parts of a component. However it's a little unclear about the best way to glue these together in a way that lets us augment existing things without rewriting them.

The scenario that started the ideas was to provide both unstyled and styled components; e.g. `ButtonBase` vs `Button`. The question we ask is: how can we "extend" the base to provide the styled version? Styling is more than just css; it's the slot defaults we use as well, as an unstyled component must not depend on a styled subcomponent or it inadvertently becomes partially styled.

We also know that wrapper components add performance overhead when done in bulk. While wrapper components can solve the concern, the add overhead when done in bulk. Having a utility which can re-scaffold new components while reusing the same essence of the original component can avoid this.

The existing compose had a number of problems:

* Typings complicated and hard to use
* Unclear on what it's doing and how much its doing
* No way to configure default props
* Slots support baked in
* Handled props baked in
* classes a separate concern when really could just be a hook
* only one state hook allowed
* Why is this even better than just rebuilding things from scratch?
* Naming is weird for recomposition; calling `compose` with an existing component is unclear on purpose.

This new update to `compose` attempts to address these concerns. It simplifies a lot. Lots has been removed. Here's what's changed.

First, the generics have been simplified to `<TProps, TState>`:

```jsx
const ButtonBase = compose<ButtonProps, ButtonState>( ... );
```

Second, hooks now can be provided in an array, each taking in "draft state" which can be directly mutated to adjust the state for the component:

```jsx
const ButtonBase = compose<...>(
  state => JSX,
  {
     useHooks: [
        useButtonState,
     ]
}
```

Hooks will be guaranteed to be called on every render in the same order. So in keeping them as an array, we can ensure less mistakes, symmetry on hooks definition, and with draft state, better perf and less immutability bugs. The hooks can be used to provide state, accessibility, and styling in a layered way.

Third, there is no more classes, slots, handledProps. There is only `defaultProps`, which take special consideration to shorthand.

Default slot values should be defined as defaultProps, just as any other default which may be overridable:

```jsx
const ButtonBase = compose<...>(
  state => JSX,
  {
     defaultProps: {
        as: 'button',
        icon: { as: 'span', children: <AddIcon /> }
     },
     shorthandProps: ['icon']
  });
```

Getting slots and slot props are done via a `getSlots` helper called in the render function, optionally, if your component supports slots.

And, lastly, compose provides an `extend` static, which helps with no extra utility imports. This means creating styled versions are very simple and really have no magic:

```tsx
const Button = ButtonBase.extend({
  useHooks: [
    makeStyles(styles, 'Button')
  ]
});
```

This creates a component copy (not a wrapper), which re-uses the state, accessibility, and render structure of ButtonBase, and adds styling opinions as overrides. We could use this, for example, to support multiple styling approaches without tying the component to a particular style implementation.

### DIFF
--- a/packages/react-compose/src/next/README.md
+++ b/packages/react-compose/src/next/README.md
@@ -1,0 +1,315 @@
+# @fluentui/react-compose
+
+Each Fluent UI React component is built from small building blocks, composed together in a way that is easy to customize and extend.
+
+Components are comprised of:
+
+- Default props
+- State/accessibility hooks
+- Styling hooks which provide class names for the parts of a component.
+- Render function: function which consumes the final state and styling and returns markup.
+
+The `compose` helper is a simple utility which lets us glue these parts together in a way that allows them to be easily extended.
+
+## How compose works
+
+Example usage:
+
+```jsx
+// compose a Foo component.
+const Foo = compose(
+  (state) => {
+    return <foo/>;
+  }, {
+  // Component has a name.
+  displayName: 'Foo',
+
+  // Component manages some state, behaviors, styling.
+  useHooks: [ (draftState) => newState | undefined ],
+
+  // Component has some default props.
+  defaultProps: { ... }
+});
+```
+
+Compose handles these concerns:
+
+1. draftState is created based on defaultProps and user input. Draft state allows you to directly manipulate a copy of the state without worrying about immutability mistakes or performance overhead of object spreading.
+2. Each hook is executed with draft state.
+3. Final state is passed to render function and result is returned.
+
+Additionally it allows the result to be extended through an attached `extend` helper:
+
+```jsx
+const AnotherComponent = MyComponent.extend({
+  useHooks: [ ...more hooks ],
+  defaultProps: { ...new defaults }
+})
+```
+
+The extend call creates a new component with new settings merged on top of the base settings. Doing this avoids extra wrapper component layers, while still reusing the component.
+
+### But why is extend important?
+
+Fluent UI components should have 2 separate layers; base components which are unstyled, and styled versions of those components. This gives developers 2 layers to use; use the styled component when building M365 experiences. Extend the unstyled component (e.g. `ButtonBase`) to use a different styling approach, or build a new, slightly augmented component extending the styled version (e.g. `Button`).
+
+## Getting started writing components
+
+### Write and review a spec
+
+Know what you're building. Do the research on the following:
+
+- What is the right component name? Consult guidance in Open UI.
+- What props are accepted? Are there standards in Open UI
+- What slots should be configurable?
+- How does customized styling work?
+
+### Build a base component
+
+A base component defines the render dom heirarchy, state management, and accessible behaviors. It does not provide styling. It does not attach dependencies. These are done in a separate layer, so that customers can always fall back to extending the base, should they wish to provide their own styling or subcomponent dependencies.
+
+Use `compose` to create the extendable base component. It takes in a render function and options:
+
+```jsx
+const MyComponentBase = compose<TProps, TState>(
+  (state, options) => {
+    return <div />;
+  }, {
+    displayName: 'MyComponentBase',
+    // ...
+  });
+```
+
+## Supporting component slots
+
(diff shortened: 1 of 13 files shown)


packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,315 @@
...
+
+Some hooks should be created by a factory function so that inputs can be cached. For example, say the prop name and default incrementing step should be provided to a factory creating the hook. We use the `make` prefix to indicate a hook factory:
+
+```jsx
+const MyComponent = MyComponentBase.extend({
+  useHooks: [makeIncrementingValue({ propName: 'value', step: 1 })],
+});
+```
+
+## Create a styled component from the base
```
xugao, line 191: i think this section belongs better to `react-theme-provider` README

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,315 @@
...
+
+    return (
+      <slots.root {...slotProps.root}>
+        <slots.icon {...slotProps.icon} />
+      </slots.root>
+    );
+  },
+  {
+    // Instruct merging on which props are shorthand.
+    shorthandProps: ['icon'],
```
xugao, line 120:
```suggestion
    shorthandPropNames: ['icon'],
```

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,315 @@
...
+    const { slots, slotProps } = getSlots(state, options);
+
+    return (
+      <slots.root {...slotProps.root}>
+        <slots.icon {...slotProps.icon} />
+      </slots.root>
+    );
+  },
+  {
+    // Instruct merging on which props are shorthand.
```
xugao, line 119: I personally don't like the name "shorthandProp" because i had no idea what it means when i first saw it.  since we have "slot" everywhere else, why not calling this "slotProp"

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,315 @@
...
+
+    defaultProps: {
+      // Define the default props for the icon.
+      icon: { as: 'span' },
+    },
+  },
+);
+```
+
+Note: the `shorthandProps` array option is required for compose to apply shorthand normalization logic for proper default props merging. (Because a shorthand prop can be a string, object, or JSX, you need to apply custom merge logic for these.)
```
xugao, line 130: we should have a section in README for shorthand/slot prop with more details on how to use

packages/react-compose/src/next/getSlots.tsx
```diff
@@ -0,0 +1,30 @@
...
+    const slotDefinition = state[name];
+
+    const slot = (slots[name] = slotDefinition.as && slotDefinition.children ? slotDefinition.as : nullRender);
+
+    if (slots[name] !== nullRender) {
+      const nativeElementProps = (slotProps[name] = getNativeElementProps(slot, slotDefinition));
+
+      if (typeof nativeElementProps.children === 'function') {
+        slots[name] = React.Fragment;
+        nativeElementProps.children = nativeElementProps.children(slots[name], nativeElementProps);
```
xugao, line 24: this doesn't seem quite right.
1. we should set ` slots[name] = React.Fragment;` after this line so `slots[name]` is passed with right/original value
2. we should be passing original slotProps instead of nativeElementProps

dzearing: you're right!

dzearing:
Closing temporarily; will re-open with some new changes.

PR #14229, opened 2020-07-29
Adding useResponsiveMode hook inside of office-ui-fabric-react utilities

czearing:
#### Pull request checklist

#### Description of changes

Added a new hook called `useResponsiveMode`  inside of office-ui-fabric-react's utilities.

#### Focus areas to test
`useResponsiveMode `


packages/office-ui-fabric-react/src/utilities/hooks/useResponsiveMode.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { getWindow } from '@uifabric/utilities';
+import { useOnEvent } from '@uifabric/react-hooks';
+import { ResponsiveMode, getResponsiveMode } from '../decorators/withResponsiveMode';
+
+/**
+ * User specified mode to default to, useful for server side rendering scenarios.
+ */
+let _defaultMode: ResponsiveMode | undefined;
```
ecraig12345, line 9: This isn't doing anything right now. Ideally you should find a way to share this (and maybe also _lastMode) between the decorator and the hook.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -22,13 +22,13 @@ const RESPONSIVE_MAX_CONSTRAINT = [479, 639, 1023, 1365, 1919, 99999999];
 /**
  * User specified mode to default to, useful for server side rendering scenarios.
  */
-let _defaultMode: ResponsiveMode | undefined;
+export let _defaultMode: ResponsiveMode | undefined;
```
ecraig12345, line 25: Should have said this before but I'm not sure that exporting these directly from the file is actually correct/desirable. For one thing, this is very much an internal detail that shouldn't really be exposed. Also, I'm pretty sure that when you import it the value will be *copied* and not pick up or propagate any updates.

Probably the easiest solution would be to add a new function `getInitialResponsiveMode` which returns `_defaultMode || _lastMode || ResponsiveMode.large`, and use it in both the new and old verisons.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -50,6 +50,10 @@ export function initializeResponsiveMode(element?: HTMLElement): void {
   }
 }
 
+export const getInitialResponsiveMode = () => {
+  return _defaultMode || _lastMode || ResponsiveMode.large;
+};
+
```
ecraig12345, line 56: Small thing but can you switch this to `function` syntax to match the rest of the file? 
```suggestion
export function getInitialResponsiveMode() {
  return _defaultMode || _lastMode || ResponsiveMode.large;
}

```

I still think `function` is preferable to lambdas in general for exported utilities since it makes debugging easier (IIRC lambdas show up as anonymous functions in callstacks, whereas functions show up with names), but that's probably something to discuss separately.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -102,7 +106,7 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
   return hoistStatics(ComposedComponent, resultClass);
 }
 
-function getResponsiveMode(currentWindow: Window | undefined): ResponsiveMode {
+export const getResponsiveMode = (currentWindow: Window | undefined): ResponsiveMode => {
```
ecraig12345, line 109: Please switch this back to `function` syntax too. Or if you must use lambdas, also switch `withResponsiveMode` so it's consistent within the file

packages/office-ui-fabric-react/src/utilities/hooks/useResponsiveMode.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { getWindow } from '@uifabric/utilities';
+import { useOnEvent } from '@uifabric/react-hooks';
+import { ResponsiveMode, getResponsiveMode, getInitialResponsiveMode } from '../decorators/withResponsiveMode';
+
+export const useResponsiveMode = (elementRef: React.RefObject<HTMLElement | null>) => {
+  const [lastResponsiveMode, setLastResponsiveMode] = React.useState<ResponsiveMode>(getInitialResponsiveMode());
```
dzearing, line 7:
```suggestion
  const [lastResponsiveMode, setLastResponsiveMode] = React.useState<ResponsiveMode>(getInitialResponsiveMode);
```

If you pass in the function without calling it, you can avoid calling it on every render. `useState` default can accept a function which will only be called on initialization.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { findDOMNode } from 'react-dom';
-import { BaseDecorator } from './BaseDecorator';
+import { BaseDecorator } from '../decorators/BaseDecorator';
```
ecraig12345, line 3: This line can be reverted now
```suggestion
import { BaseDecorator } from './BaseDecorator';
```

PR #14232, opened 2020-07-29
feat(builder): simulate virtual cursor in the use mode

assuncaocharles:
#### Pull request checklist

#### Description of changes

This PR allows to navigate through components using arrow down/ arrow up and to focus element based on other key press. 

lU7YXllOSH

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-builder/src/components/Canvas.tsx
+++ b/packages/fluentui/react-builder/src/components/Canvas.tsx
@@ -24,6 +24,7 @@ export type CanvasProps = {
   onDeleteComponent?: () => void;
   onGoToParentComponent?: () => void;
   renderJSONTreeElement?: (jsonTreeElement: JSONTreeElement) => JSONTreeElement;
+  enabledVirtualCursor?: boolean;
   style?: React.CSSProperties;
   mode?: 'build' | 'design' | 'use';
   role?: string;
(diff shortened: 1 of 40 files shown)


packages/fluentui/react-builder/src/components/Canvas.tsx
```diff
@@ -126,6 +131,47 @@ export const Canvas: React.FunctionComponent<CanvasProps> = ({
...
+            [
+              'a',
+              'button',
+              'checkbox',
+              'dialog',
+              'gridcell',
+              'link',
+              'log',
+              'marquee',
+              'menuitem',
```
jurokapsiar, line 146: for these it should be something like [role='menuitem']

packages/fluentui/react-builder/src/components/Canvas.tsx
```diff
@@ -293,10 +383,27 @@ export const Canvas: React.FunctionComponent<CanvasProps> = ({
               />
             )}
 
-            <Provider theme={teamsTheme} target={document}>
+            <Provider
+              theme={teamsTheme}
+              target={document}
+              styles={{
+                '::after': {
+                  content: 'asdijhaoijdoi',
```
jurokapsiar, line 391: what is this? :)

assuncaocharles: Upsi

PR #14238, opened 2020-07-29
Updating TeachingBubble to a function component inside React-Next

czearing:
#### Pull request checklist

#### Description of changes
1. Converted `TeachingBubble` and `TeachingBubbleContent` to a function component inside React-Next.
2. Added perf-tests and vr-tests for `TeachingBubble`.
3. Added a document type to the element parameter in `useOnEvent` and `on.ts` utility.
4. Removed unused defaultProps from `TeachingBubbleContent`.
5. Added useDocumentRef and useWindowRef hooks to allow `TeachingBubbleContent`'s "keydown" event to function properly.
6. Removed rootElementRef from public api.

#### Focus areas to test
`TeachingBubble`


packages/react-next/src/components/TeachingBubble/TeachingBubble.base.tsx
```diff
@@ -1,78 +1,59 @@
...
+      focus: () => rootElementRef.current?.focus(),
+    }),
+    [rootElementRef],
+  );
+};
 
-  public render(): JSX.Element {
+export const TeachingBubbleBase = React.forwardRef(
+  (props: ITeachingBubbleProps, forwardedRef: React.Ref<HTMLDivElement>) => {
+    const rootElementRef = React.useRef<HTMLDivElement>(null);
```
ecraig12345, line 40: In the previous implementation, this was exposed as a public member for some reason, which if you're directly translating means it should go on the `useComponentRef`. It's probably okay to leave that out but it might be worth noting in RELEASE_NOTES.md breaking changes.

packages/react-next/src/components/TeachingBubble/TeachingBubble.types.ts
```diff
@@ -21,9 +19,7 @@ export interface ITeachingBubble {
  * TeachingBubble component props.
  * {@docCategory TeachingBubble}
  */
-export interface ITeachingBubbleProps
-  extends React.ClassAttributes<TeachingBubbleBase | TeachingBubbleContentBase>,
-    IAccessiblePopupProps {
+export interface ITeachingBubbleProps extends React.ClassAttributes<HTMLElement>, IAccessiblePopupProps {
```
ecraig12345, line 22: This isn't a class component, so having it continue extending `React.ClassAttributes` seems wrong. You should probably switch it to either `React.RefAttributes` or `React.Attributes`.

packages/react-next/src/components/TeachingBubble/TeachingBubbleContent.base.tsx
```diff
@@ -86,6 +64,26 @@ export class TeachingBubbleContentBase extends React.Component<ITeachingBubblePr
...
+    const onKeyDown = React.useCallback(
+      (ev: React.KeyboardEvent<HTMLElement> | KeyboardEvent): void => {
+        if (props.onDismiss) {
+          // eslint-disable-next-line deprecation/deprecation
+          if (ev.which === KeyCodes.escape) {
+            props.onDismiss(ev);
+          }
+        }
+      },
+      [props.onDismiss],
```
ecraig12345, line 76: You should reference `onDismiss` directly rather than from props. That way the callback won't be re-created as often. (The identity of the `props` object will change every time props are updated, even if the identity of `onDismiss` stays constant.)
```suggestion
    const onKeyDown = React.useCallback(
      (ev: React.KeyboardEvent<HTMLElement> | KeyboardEvent): void => {
        if (onDismiss) {
          // eslint-disable-next-line deprecation/deprecation
          if (ev.which === KeyCodes.escape) {
           onDismiss(ev);
          }
        }
      },
      [onDismiss],
```

packages/react-next/src/components/TeachingBubble/TeachingBubbleContent.base.tsx
```diff
@@ -86,6 +64,26 @@ export class TeachingBubbleContentBase extends React.Component<ITeachingBubblePr
...
+    );
+
+    useOnEvent(getDocument(rootElementRef.current), 'keydown', onKeyDown as (ev: Event) => void);
```
ecraig12345, line 79: I'm not sure this will work--`rootElementRef.current` won't be defined yet on initial render, so finding the relevant document won't work correctly. Instead I think this will need to be wrapped in `useEffect`.

packages/react-next/src/components/TeachingBubble/TeachingBubbleContent.base.tsx
```diff
@@ -86,6 +64,26 @@ export class TeachingBubbleContentBase extends React.Component<ITeachingBubblePr
...
+
+    let imageContent;
+    let headerContent;
+    let bodyContent;
+    let footerContent;
+    let closeButton;
```
ecraig12345, line 85:
```suggestion
    let imageContent: JSX.Element | undefined;
    let headerContent: JSX.Element | undefined;
    let bodyContent: JSX.Element | undefined;
    let footerContent: JSX.Element | undefined;
    let closeButton: JSX.Element | undefined;
```

PR #14240, opened 2020-07-29
Add @fluentui/storybook package and dedup withThemeProvider decorator

### LINKED ISSUE #14144: Move common storybook addons into shared package
Need to reduce duplication of storybook files between packages by putting them in a common place. See discussion here: https://github.com/microsoft/fluentui/pull/14125#discussion_r458250367

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14144

#### Description of changes
Moved `withThemeProvider` and related code to new package `@fluentui/storybook` to avoid duplication.
As we scaffolding more packages and moving code out of `react-next`, we need to proper share the code instead of duplicating more.

image

#### Focus areas to test

(optional)


packages/storybook/.eslintrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": ["plugin:@fluentui/eslint-plugin/react"],
```
ecraig12345, line 2: The `node` config might be more appropriate here since this package is used for dev-time stuff. (`node` might not be the best name for the config.)
```suggestion
  "extends": ["plugin:@fluentui/eslint-plugin/node"],
```

packages/storybook/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@fluentui/storybook",
+  "version": "0.1.0",
+  "description": "Storybook addons",
```
ecraig12345, line 4:
```suggestion
  "description": "Storybook addons for Fluent UI React",
```

packages/storybook/src/themes/V7/AzureDarkTheme.ts
```diff
@@ -0,0 +1,7 @@
+import { AzureCustomizationsDark } from '@uifabric/azure-themes';
```
ecraig12345, line 1: Probably the v7 folder name should be lowercase to more closely match the convention elsewhere.

packages/storybook/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "declaration": true,
+    "sourceMap": true,
+    "experimentalDecorators": true,
+    "importHelpers": true,
+    "noUnusedLocals": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "moduleResolution": "node",
+    "preserveConstEnums": true,
+    "types": ["jest", "react", "react-dom"],
```
ecraig12345, line 18: This setting is just for which types should be automatically included globally (types such as `react` will be automatically pulled in when the package is imported). Here the only thing you might need is `webpack-env`, since this package doesn't use jest.
```suggestion
    "types": ["webpack-env"],
```

packages/storybook/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "outDir": "lib",
+    "target": "es5",
+    "module": "commonjs",
+    "lib": ["es5", "esnext", "dom"],
```
ecraig12345, line 7: Does this package have an IE 11 support requirement? If so you need to remove the `esnext` lib. (If not, you can remove `es5` since it's automatically included by `esnext`.)

packages/storybook/.eslintrc.json
xugao: since so far the package only contains shared addon code (which is React). i think we should keep `react`. perhaps we will need to also extend `node` later

packages/storybook/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "declaration": true,
+    "sourceMap": true,
+    "experimentalDecorators": true,
+    "importHelpers": true,
+    "noUnusedLocals": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "moduleResolution": "node",
+    "preserveConstEnums": true,
+    "types": ["react"],
```
ecraig12345, line 18: I still don't think this should be here. Also if you're not using webpack-env it should be removed from dependencies.

packages/storybook/tsconfig.json
xugao: I do need `react` here. so i updated to only that. I don't see a need for `webpack-env` for now or future? please correct me if i'm wrong here.

packages/storybook/tsconfig.json
ecraig12345: If you do remove `react` and don't add `webpack-env`, it's important to use an empty array instead to prevent automatic inclusion of all `@types` in the compilation, which is very slow.

ecraig12345: Why do you need `react`? Even without this the types will be automatically picked up in any file where you import `react`. The only time you need to reference it like this is if you want to pull React into the ambient global namespace.

xugao: yes i also just realized the same. fixed!

PR #14244, opened 2020-07-30
[Hackathon] simplified narration for ui builder elements

jurokapsiar:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/fluentui/react-builder/src/components/ReaderText.tsx
```diff
@@ -11,9 +12,15 @@ export const ReaderText: React.FunctionComponent<ReaderTextProps> = ({ selector
 
   React.useEffect(() => {
     if (ref.current) {
-      const t = ref.current.ownerDocument.querySelector(selector)?.textContent;
-      setText(t);
+      const element = ref.current.ownerDocument.querySelector(selector);
```
miroslavstastny, line 15: The `ref` handling is messy. You don't want to use `createRef` in a functional component, that created a new ref on each render. You want `useRef`.
With that you cannot have `ref.current` as a dependency.
When do you want the effect to be called?
Also run `yarn lint` in `packages/fluentui/react-builder`.

PR #14252, opened 2020-07-30
Added prototype calls

vyhnalekl:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/fluentui/docs/src/prototypes/Calls/recentContact.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { CallIcon, CallVideoIcon } from '@fluentui/react-icons-northstar';
+
+const Contact: React.FC<{
+  image?: string;
+  name: string;
+  status: object;
+}> = ({ image, name, status }) => {
+  return (
+    <Flex fill gap="gap.small" space="between">
+      <div>
```
mnajdova, line 12: I agree that using `<div />` is the correct thing, but this is not reflecting 100% the test case we want to create. Let's use `<Box />` or `<Flex />` here. Later if we want to show how big improvements changes like this can make, it will be more obvious.

packages/fluentui/docs/src/prototypes/Calls/recentContact.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
```
mnajdova, line 1: Rename the file to `Contact.tsx` as it is a component file.

packages/fluentui/docs/src/prototypes/Calls/recentContact.tsx
mnajdova, line 12: Can we use `Flex` with some `gap` here instead of adding div and overrides on the children for the margins?

packages/fluentui/docs/src/prototypes/Calls/recentContact.tsx
```diff
@@ -0,0 +1,34 @@
...
+          {
+            icon: <CallIcon outline />,
+            key: 'star',
+          },
+          {
+            icon: <CallVideoIcon outline />,
+            key: 'search',
+          },
+        ]}
+        variables={{ isContactMenu: true }}
```
mnajdova, line 28: This variable does not have any styles associated with it. Please remove it.

packages/fluentui/docs/src/prototypes/Calls/recentContactsList.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
```
mnajdova, line 1: Rename to `RecentContactsList.tsx`

packages/fluentui/docs/src/prototypes/Calls/recentContact.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <Flex fill gap="gap.small" space="between">
+      <div>
+        <Avatar image={`data:image/jpeg;base64,${image}`} name={name} status={status} />
+        <Text content={name} variables={{ isNameText: true }} />
+      </div>
+      <Menu
+        iconOnly
+        items={[
+          {
+            icon: <CallIcon outline />,
```
mnajdova, line 20: Should we add some actions on these elements?

packages/fluentui/docs/src/prototypes/Calls/participant.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <Text weight={talking && 'bold'} content={name} variables={{ isNameText: true }} />
+        <Text variables={{ isRoleText: true }} content={role} />
+      </Flex>
+
+      {muted ? (
+        <>
+          <FlexItem push>
+            <MicOffIcon />
+          </FlexItem>
+          <MenuButton
```
mnajdova, line 51: Seems like the `MenuButton` is rendered anyway. I would suggest rendering it after the condition. Something like:

{muted && <FlexItem>...</FlexItem>}
<MenuButton ... />

packages/fluentui/docs/src/prototypes/Calls/recentContact.tsx
vyhnalekl: What actions do you have in mind?

ecraig12345:
@vyhnalekl Is this still needed?

PR #14259, opened 2020-07-30
Adding raw version of acc spec generator 

kolaps33:
#### Description of changes

Adding "Show Acc Spec" checkbox. 
Then adding button "generate Acc Spec".  (added button, otherwise I had react failure complain about too many re rendering )
When user presses button then component are given into the table. 
Then next information based on the component should be filled in the table. 

(give an overview)
image

### DIFF
--- a/packages/fluentui/react-builder/src/components/Designer.tsx
+++ b/packages/fluentui/react-builder/src/components/Designer.tsx
@@ -317,7 +317,7 @@ export const Designer: React.FunctionComponent = () => {
   const [components, setComponents] = React.useState([]);
 
   const componentsExcludedFromAccSpec = ['Flex', 'Box', 'FlexItem'];
-  let componentsForAccSpec = [];
+  const componentsForAccSpec = [];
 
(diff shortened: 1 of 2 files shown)


packages/fluentui/react-builder/src/components/Designer.tsx
```diff
@@ -704,6 +741,13 @@ export const Designer: React.FunctionComponent = () => {
                 )}
               </div>
             )}
+            {showAccSpec && (
+              <div style={{ flex: 1, padding: '1rem', color: '#543', background: '#ddd' }}>
+                <h3 style={{ margin: 0 }}>Accessibility specification</h3>
+                <Button content="generate Acc Spec" onClick={generateAccSpec} />
```
jurokapsiar, line 747:
```suggestion
                <Button content="Generate" onClick={generateAccSpec} />
```

kolaps33: changed :)

PR #14268, opened 2020-07-30
Replace compose with a simpler set of helpers (not export impacted, included under /next folder)

dzearing:
Note: this PR changes no exported code; I've staged a `/next` folder in `react-compose` to host these changes for us to play test in use cases. Feedback is welcome, any ways we can improve, let's get it right and keep on going.

After much discussion, experimentation, and pros and cons, @levithomason and I collaborated on some simplifications which emphasize hooks and functional composition over config. Here's essentially where we landed:

1. No compose or extend helper, no complex typings, no config!
2. No "base" components; instead, component factory functions which return the final state and the final render function for the component. This essentially provides the exact same re-composability, but it's functional.
3. Only provide simple utilities which you can use when you need, and can be easily replaced without redoing everything.

So! This PR creates 3 simple utilities:

`mergeProps` - a simple deep merge that is props safe (e.g. classnames get appended)
`getSlots` - a helper for extracting `slots` and `slotProps` from the props.
`resolveShorthandProps` - a helper for making shorthand props deep merge-able.

I've written a more thorough writeup in the README; please read that for further details!

I have converted Button, ToggleButton, and MenuButton all to this pattern. This will come in a separate PR.


packages/react-compose/src/next/types.tsx
```diff
@@ -0,0 +1,33 @@
+export const OPTIONS_NAME = '__options';
```
xugao, line 1: not used, remove?

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,221 @@
...
+const createButton = (userProps, ref, defaultProps) => {
+  const state = mergeProps(
+    {
+      // default props
+      as: 'button',
+      ref,
+      icon: { as: 'span' },
+    },
+    defaultProps, // optional default props from the caller
+    simplifyShorthand(userProps, ['icon']), // simplify the user's props
```
xugao, line 155: why don't we call `simplifyShorthand ` internally in `mergeProps`?  instead, pass `userProps` and `shorthandPropNames` to `mergeProps`

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,221 @@
...
+<Button icon="X" />
+
+// The icon can be JSX
+<Button icon={ <FooIcon/> }/>
+
+// The icon can be an object
+<Button icon={{ as: 'i', children: getCode('Add') } } />
+
+// The icon can be a children function (which receives the original slot and props)
+<Button icon={{ children: (C, p) => <C {...p} otherThings /> }} />
```
xugao, line 133:
```suggestion
<Button icon={{
  children: (Component, props) => 
      <>
         <Component {...props} />
         //  other things
      </> 
   }}
/>
```

packages/react-compose/src/next/types.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export interface ComponentProps extends GenericDictionary {
+  as?: React.ElementType;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ref?: React.Ref<any>;
+  className?: string;
+}
+
+//
+// Slot Prop / Shorthand types
```
xugao, line 17:
```suggestion
// Shorthand types
```

packages/react-compose/src/next/getSlots.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * collections. Ensures that each slot is a renderable component, and each slotProp is
+ * a list of acceptable native properties for the given slot. Native properties are only
+ * filtered if the slot is rendered "as" a string type, and will use `getNativeElementProps`
+ * to pass along only valid props. If the slot is a non-string value, no filtering is done.
+ *
+ * @param state - State including slot definitions
+ * @param slotNames - Name of which props are slots
+ * @returns An object containing the `slots` map and `slotProps` map.
+ */
+export const getSlots = (state: GenericDictionary, slotNames: string[] | undefined) => {
```
xugao, line 17: where did the tests go?

packages/react-compose/src/next/README.md
dzearing: I was thinking that not every component has slots, yet you'd still want to use mergeProps for merging user input on defaults.

dzearing: Also, right now mergeProps takes in ...rest as the additional sets. I am not sure how to do that without making the shorthandPropNames the first or second argument, which seems off to emphasize

packages/react-compose/src/next/getSlots.tsx
dzearing: I haven't added them yet, I had some tests for compose. Happy to add them if we're good with this, but i was hoping to get more eyes on the general approach before i go too deep with tests. They will get done before moving this out of /next folder.

dzearing: I did add some getSlots tests; i will add more as well for the others, but I'm eager to get the button PR going to see numbers.

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+`@fluentui/react-compose` is a library of utilities for composing Fluent UI components.
+
+## A basic component walkthrough
+
+Building a recomposable component requires that we build legos; we put them together, but we let others put them together how they may need. Here's what's needed:
+
+- **State hooks** - all processing of the component should be done within one or more reusable hooks. The hooks should be allowed to manipulate the state in some reusable way, so that recomposing the component can be easily done.
+- **Render function** - a function which takes in final state and returns JSX. (e.g. `renderButton`)
+- **Component factory function** - a create\* function which returns `{ state, render }` parts for creating the component. This is used to scaffold a new version of teh component as needed.
```
khmakoto, line 11:
```suggestion
- **Component factory function** - a create function which returns `{ state, render }` parts for creating the component. This is used to scaffold a new version of the component as needed.
```

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+  return render(state);
+});
+```
+
+### Details
+
+#### Creating mutable state with `mergeProps`
+
+In the previous example, `_.merge` was used to deep clone the props into a state object. Creating a single clone and using that to construct state simplifies hook development and usage; rather than trying to re-clone objects unnecessarily on every small mutation, hooks can assume operating against a draft state. This creates more self contained hooks, which can ensure they are used correctly, avoid accidents like stomping on existing event handlers by blind object assigning the results.
```
khmakoto, line 71:
```suggestion
In the previous example, `_.merge` was used to deep clone the props into a state object. Creating a single clone and using that to construct state simplifies hook development and usage; rather than trying to re-clone objects unnecessarily on every small mutation, hooks can assume operating against a draft state. This creates more self contained hooks, which can ensure they are used correctly, avoiding accidents like stomping on existing event handlers by blind object assigning the results.
```

packages/react-compose/src/next/getSlots.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
+import { getNativeElementProps } from '@uifabric/utilities';
+import { GenericDictionary } from './types';
+import { nullRender } from './nullRender';
+
+/**
+ * Given the state and an array of slot names, will break out `slots` and `slotProps`
+ * collections. Ensures that each slot is a renderable component, and each slotProp is
+ * a list of acceptable native properties for the given slot. Native properties are only
+ * filtered if the slot is rendered "as" a string type, and will use `getNativeElementProps`
```
khmakoto, line 10: `Native properties are only filtered if the slot is rendered "as" a string type`

You're doing this check for the slot `as` prop but not for the `root`.

packages/react-compose/src/next/getSlots.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const getSlots = (state: GenericDictionary, slotNames?: string[] | undefined) => {
+  const slots: GenericDictionary = {
+    root: state.as || nullRender,
+  };
+  const slotProps: GenericDictionary = {
+    root: getNativeElementProps(state.as, state),
+  };
+
+  if (slotNames) {
+    for (const name of slotNames!) {
```
khmakoto, line 26: Do you have to do the `!`? Doesn't the condition above take care of that?

packages/react-compose/src/next/getSlots.tsx
```diff
@@ -0,0 +1,48 @@
...
+      const slotDefinition = state[name];
+      const { as: slotAs, children, ...rest } = slotDefinition;
+
+      const slot = (slots[name] = slotDefinition.children || typeof slotAs !== 'string' ? slotAs : nullRender);
```
khmakoto, line 30:
```suggestion
      const slot = (slots[name] = children || typeof slotAs !== 'string' ? slotAs : nullRender);
```

packages/react-compose/src/next/mergeProps.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+import { css } from '@uifabric/utilities';
+import { GenericDictionary } from './types';
+
+/**
+ * Helper which deep clones props, but respecting JSX and classnames.
+ * @param target
+ * @param propSets
```
khmakoto, line 8: A brief description of the params here could be very helpful for people onboarding on this.

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,228 @@
...
+    state.tabIndex = 0;
+  }
+};
+```
+
+A factory function sets up the state and render function:
+
+```jsx
+const createButton = (userProps, ref, defaultProps) => {
+  const state = _.merge({}, defaultProps, userProps);
```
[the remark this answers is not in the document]
dzearing: i only used it as a deep merge example people might be familiar with.

packages/react-compose/src/next/simplifyShorthand.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * @param props - The incoming props
+ * @param shorthandPropNames - An array of prop names to apply simplification to
+ */
+export const simplifyShorthand = (props: GenericDictionary, shorthandPropNames?: string[]) => {
+  let newProps = props;
+
+  if (shorthandPropNames && shorthandPropNames.length) {
+    newProps = {
+      ...props,
+    };
```
khmakoto, line 16: Isn't this exactly the same as the `newProps = props` you're doing above?

packages/react-compose/src/next/types.tsx
```diff
@@ -0,0 +1,29 @@
+/**
```
khmakoto, line 1: Two things:
1. This should be a .ts file.
2. You probably want to import `React` here.

packages/react-compose/src/next/types.tsx
```diff
@@ -0,0 +1,29 @@
...
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type GenericDictionary = Record<string, any>;
+
+export interface ComponentProps extends GenericDictionary {
+  as?: React.ElementType;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ref?: React.Ref<any>;
+  className?: string;
+}
```
khmakoto, line 12: Are these duplicates from `src/types.ts` purposeful?

packages/react-compose/src/next/getSlots.tsx
dzearing: Fixed.

packages/react-compose/src/next/getSlots.tsx
dzearing: fixed.

packages/react-compose/src/next/getSlots.tsx
dzearing: fixed already.

packages/react-compose/src/next/mergeProps.tsx
dzearing: done :)

packages/react-compose/src/next/types.tsx
dzearing: awesome catch

packages/react-compose/src/next/types.tsx
dzearing: My plan was to replace the files in the src folder completely, so i started with the types.ts file content and started purging as needed.

dzearing: The whole folder is temporary, so i'll leave for now.

packages/react-compose/src/next/simplifyShorthand.tsx
dzearing: spreading will ensure we don't make modifications to the original props. if I could get rid of spreading here, I would, but i don't see a clean way to avoid the duplicate.

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,226 @@
...
+  return <button {...state} />;
+};
+```
+
+A hook which can manipulate the state (defining accessibility and behaviors):
+
+```jsx
+const useButton = state => {
+  // If the button is rendered as something besides a button or anchor, make it focusable.
+  if (state.as !== 'button' && state.as !== 'a') {
```
layershifter, line 31: This should be computed based on `ref.current.tagName`

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+#### Creating mutable state with `mergeProps`
+
+In the previous example, `_.merge` was used to deep clone the props into a state object. Creating a single clone and using that to construct state simplifies hook development and usage; rather than trying to re-clone objects unnecessarily on every small mutation, hooks can assume operating against a draft state. This creates more self contained hooks, which can ensure they are used correctly, avoiding accidents like stomping on existing event handlers by blind object assigning the results.
+
+However, deep merge overlooks many edge cases for component props:
+
+- Deep merging classnames should append them, not replace
+- Deep merging JSX, ref objects, or arrays should replace, not recurse/clone
+- Deep merging an object on a string should replace
```
layershifter, line 77: And how callbacks will be handled? I.e. `mergeProps({ onClick: props.onClick }, { onClick: () => {} })`?

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,226 @@
...
+  return <root {...state} />;
+};
+```
+
+Additionally, you will need to filter out native properties which apply to the root; otherwise you will end up mixing any unexpected props into the element. To do this, we have a `getNativeElementProps` helper, which can be used for this purpose:
+
+```jsx
+const renderButton = state => {
+  const root = state.as;
+  const rootProps = getNativeElementProps(root, state);
```
layershifter, line 100:
```jsx
const Link = React.forwardRef((props, ref) => {
  return <a {...props} ref={ref} />;
});

//

const root = state.as; // { as: Link }
const rootProps = getNativeElementProps(root, state);
```

What will be `rootProps` in this case?

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+Example:
+
+```jsx
+mergeProps(props, { ...etc }, { ...etc });
+```
+
+### getSlots(state: Record<string, any>, slotNames: string[])
+
+The `getSlots` function returns
```
layershifter, line 201: Missing description?

packages/react-compose/src/next/simplifyShorthand.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { GenericDictionary } from './types';
+
+/**
+ * Ensures that the given slots are represented using object syntax. This ensures that
+ * the object can be merged along with other objects.
+ * @param props - The incoming props
+ * @param shorthandPropNames - An array of prop names to apply simplification to
+ */
+export const simplifyShorthand = (props: GenericDictionary, shorthandPropNames?: string[]) => {
```
layershifter, line 10: I suggest to improve typings: `shorthandPropNames` should be present in `props`

packages/react-compose/src/next/simplifyShorthand.tsx
```diff
@@ -0,0 +1,27 @@
...
+  let newProps = props;
+
+  if (shorthandPropNames && shorthandPropNames.length) {
```
layershifter, line 13: Can you please clarify why `shorthandPropNames` should be optional?

packages/react-compose/src/next/simplifyShorthand.tsx
layershifter, line 10: `resolveShorthands`?

packages/react-compose/src/next/simplifyShorthand.tsx
```diff
@@ -0,0 +1,27 @@
...
+      const propValue = props[propName];
+
+      if (propValue !== undefined && (typeof propValue !== 'object' || React.isValidElement(propValue))) {
+        newProps[propName] = { children: propValue };
+      }
+    }
+  }
+
+  return newProps;
+};
```
layershifter, line 27:
```jsx
console.log(simplifyShorthand({ image: false }, ["image"]));
console.log(simplifyShorthand({ image: true }, ["image"]));
```

image

Is it expected?

packages/react-compose/src/next/getSlots.test.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react';
+import { getSlots } from './getSlots';
+import { nullRender } from './nullRender';
+
+describe('getSlots', () => {
+  const Foo = () => <div />;
+
+  it('renders nullRender for root if the as prop is not provided', () => {
```
mnajdova, line 8: This seems a bit dangerous, especially for the root element. Should we default to `div` if it is nothing is provided?

packages/react-compose/src/next/getSlots.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  it('renders root as a span with no props', () => {
+    expect(getSlots({ as: 'span' })).toEqual({
+      slots: { root: 'span' },
+      slotProps: { root: {} },
+    });
+  });
+
+  it('renders root as a button, omitting the href prop', () => {
```
mnajdova, line 22:
```suggestion
  it('omits props which are not valid for the element rendered', () => {
```
I think this better explain what we are trying to test

packages/react-compose/src/next/getSlots.test.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react';
+import { getSlots } from './getSlots';
+import { nullRender } from './nullRender';
+
+describe('getSlots', () => {
```
mnajdova, line 5: Generally would be better in my opinion if we omit the `renders` prefix in the tests as this is just a utility that is not rendering anything. Then for the tests which are actually rendering something, we can test whether the result rendered is really what we would expect.

packages/react-compose/src/next/getSlots.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  it('renders a component slot with no children', () => {
+    expect(getSlots({ as: 'div', icon: { as: Foo } }, ['icon'])).toEqual({
+      slots: { root: 'div', icon: Foo },
+      slotProps: { root: {}, icon: {} },
+    });
+  });
+
+  it('renders slot as button and omits unsupported props (href)', () => {
```
mnajdova, line 57: I like this description on the unsupported props more 👍

packages/react-compose/src/next/getSlots.tsx
```diff
@@ -0,0 +1,52 @@
...
+    root: state.as || nullRender,
+  };
+  const slotProps: GenericDictionary = {
+    root: typeof state.as === 'string' ? getNativeElementProps(state.as, state) : omit(state, { as: 1 }),
+  };
+
+  if (slotNames) {
+    for (const name of slotNames) {
+      const slotDefinition = state[name];
+      const { as: slotAs, children } = slotDefinition;
```
layershifter, line 33: This thing implements different behavior than current shorthand API, I mentioned this multiple times. It relies only on `as` prop value which may lead to broken scenarios.

```jsx
  mergeProps(
    {
      image: { as: Image, src: "foo.png", fit: 'center' } // Image supports `fit` prop
    },
    {
      image: { as: 'img' } // Image does not support `fit` prop
    }
  )
```

`mergeProps()` & `getSlots()` will produce ``
`createShorthand()` will produce `<Image as='img' src="foo.png" fit="center" />`

packages/react-compose/src/next/getSlots.tsx
```diff
@@ -0,0 +1,52 @@
...
+      const isSlotPrimitive = typeof slotAs === 'string';
+
+      slots[name] = slotDefinition.children !== undefined || !isSlotPrimitive ? slotAs : nullRender;
```
layershifter, line 36: nit: The condition looks cryptic. IMHO, introducing is `isSlotNull` will be more obvious

packages/react-compose/src/next/getSlots.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * @param state - State including slot definitions
+ * @param slotNames - Name of which props are slots
+ * @returns An object containing the `slots` map and `slotProps` map.
+ */
+export const getSlots = (state: GenericDictionary, slotNames?: string[] | undefined) => {
+  const slots: GenericDictionary = {
+    root: state.as || nullRender,
+  };
+  const slotProps: GenericDictionary = {
+    root: typeof state.as === 'string' ? getNativeElementProps(state.as, state) : omit(state, { as: 1 }),
```
mnajdova, line 27: This API looks a bit weird to me:
```
 omit(state, { as: 1 })
```

Can we refactor to something like
```
omit(state, ['as'])
```

packages/react-compose/src/next/getSlots.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+      slots[name] = slotDefinition.children !== undefined || !isSlotPrimitive ? slotAs : nullRender;
+
+      if (slots[name] !== nullRender) {
+        slotProps[name] = isSlotPrimitive
+          ? getNativeElementProps(slotAs, slotDefinition)
+          : omit(slotDefinition, { as: 1 });
+      }
+
+      if (children === 'function') {
```
layershifter, line 44:
```suggestion
      if (typeof children === 'function') {
```

packages/react-compose/src/next/getSlots.tsx
mnajdova, line 36: Maybe I do not understand the logic, but there can be component without children, but still valid for rendering, no?

```
slots: { divider: Divider },
slotProps: { divider: { vertical : true } },
```

How would this work?

packages/react-compose/src/next/getSlots.tsx
layershifter:
```tsx
console.log(
  getSlots(
    {
      image: {
        as: "img",
        src: "bar.jpg",
        children: (Component, props) => <Component {...props} />
      }
    },
    ["image"]
  )
);
```

image

`as` & `src` will go to `React.Fragment`, this will produce a React warning

packages/utilities/etc/utilities.api.md
```diff
@@ -330,7 +330,7 @@ export function getLastFocusable(rootElement: HTMLElement, currentElement: HTMLE
 export function getLastTabbable(rootElement: HTMLElement, currentElement: HTMLElement, includeElementsInFocusZones?: boolean, checkNode?: boolean): HTMLElement | null;
 
 // @public
-export function getNativeElementProps<TAttributes extends React.HTMLAttributes<any>>(tagName: keyof React.ReactHTML, props: {}, excludedPropNames?: string[]): TAttributes;
+export function getNativeElementProps<TAttributes extends React.HTMLAttributes<any>>(tagName: string, props: {}, excludedPropNames?: string[]): TAttributes;
```
mnajdova, line 333: Why was this change?

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,226 @@
...
+A factory function sets up the state and render function:
+
+```jsx
+const createButton = (userProps, ref, defaultProps) => {
+  const state = _.merge({}, defaultProps, userProps);
+
+  // Apply button behaviors.
+  useButton(state);
+
+  return { state, render };
```
miroslavstastny, line 46:
```suggestion
  return { state, render: renderButton };
```

packages/react-compose/src/next/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+  return render(state);
+});
+```
+
+### Details
+
+#### Creating mutable state with `mergeProps`
+
+In the previous example, `_.merge` was used to deep clone the props into a state object. Creating a single clone and using that to construct state simplifies hook development and usage; rather than trying to re-clone objects unnecessarily on every small mutation, hooks can assume operating against a draft state. This creates more self contained hooks, which can ensure they are used correctly, avoiding accidents like stomping on existing event handlers by blind object assigning the results.
```
layershifter, line 71: ~~Then we will need a custom utility to deep merge JSX objects (i.e. `children` or slots) because JSX objects are deep nested may be circular: https://github.com/FormidableLabs/react-fast-compare~~

layershifter, line 71: > Creating a single clone and using that to construct state simplifies hook development and usage;

Why? Do you have examples in other libraries?
For me state mutations are confusing and non-intuitive.

> rather than trying to re-clone objects unnecessarily on every small mutation

Is it a performance bottleneck?

packages/react-compose/src/next/README.md
layershifter: How many cases are you going to handle?

```js
const today = new Date();
const result = mergeProps({
  date: { as: "span", today }
});

today.setFullYear(2100);

console.log(today.toDateString(), result.date.today.toDateString()); // Tue Aug 03 2100 Tue Aug 03 2100 
```

Lodash handles this scenario.

packages/react-compose/src/next/README.md
layershifter: We have components like `Tree`, `Table`, etc. that accept nested shorthands:

```tsx
[
  {
    id: "tree-item-1",
    title: "House Lannister",
    items: [
      {
        id: "tree-item-11",
        title: "Tywin",
        items: [
          {
            id: "tree-item-111",
            title: "Jaime"
          },
          {
            id: "tree-item-112",
            title: "Cersei"
          },
          {
            id: "tree-item-113",
            title: "Tyrion"
          }
        ]
      },
      {
        id: "tree-item-12",
        title: "Kevan",
        items: [
          {
            id: "tree-item-121",
            title: "Lancel"
          },
          {
            id: "tree-item-122",
            title: "Willem"
          },
          {
            id: "tree-item-123",
            title: "Martyn"
          }
        ]
      }
    ]
  },
  {
    id: "tree-item-2",
    title: "House Targaryen",
    items: [
      {
        id: "tree-item-21",
        title: "Aerys",
        items: [
          {
            id: "tree-item-211",
            title: "Rhaegar"
          },
          {
            id: "tree-item-212",
            title: "Viserys"
          },
          {
            id: "tree-item-213",
            title: "Daenerys"
          }
        ]
      }
    ]
  }
];
```

https://fluentsite.z22.web.core.windows.net/components/tree/definition?showCode=true&showRtl=false&showTransparent=false&showVariables=false#types-tree

Will be deep clone more effective in this case?

packages/react-compose/src/next/simplifyShorthand.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  if (shorthandPropNames && shorthandPropNames.length) {
+    newProps = {
+      ...props,
+    };
+    for (const propName of shorthandPropNames) {
+      const propValue = props[propName];
+
+      if (propValue !== undefined && (typeof propValue !== 'object' || React.isValidElement(propValue))) {
+        newProps[propName] = { children: propValue };
```
miroslavstastny, line 21: v0 shorthands can map to a different `prop` based on config:
https://github.com/microsoft/fluentui/blob/10a4c280bd1feb94a149e35555d0539d55ff24a8/packages/fluentui/react-northstar/src/components/Menu/MenuItem.tsx#L579-L581
and also supports an array collections:
https://github.com/microsoft/fluentui/blob/10a4c280bd1feb94a149e35555d0539d55ff24a8/packages/fluentui/react-northstar/src/components/Menu/Menu.tsx#L389-L391

Do you plan to support these as well?

packages/react-compose/src/next/README.md
dzearing: Isn't that a circular dependency? We need to know, before render, what type we expect to render, in order to determine props. We need this for filtering (getNativeElementProps), we need it here for determining the appropriate behaviors.

Refs are resolved at mount. So assuming we did what you say, we'd need to first render, then re-render when we get our assumption wrong, so that we can render the correct attributes.

Could we simply lock typings to assume "as" is a string tag?

packages/react-compose/src/next/README.md
dzearing: functions and arrays are pass through. it does not try to merge 2 callbacks together; it will replace.

we could also apply the deep merge only for Object prototypes and leave class instances as replace rather than merge.

I'm not following what you're referring to with lodash.

packages/react-compose/src/next/README.md
dzearing: Assuming that as is a keyof html element names. I might have the typings wrong, but that's the intent.

dzearing: `as={ Link }` doesn't really make sense. `as={ AnythingWithTotallyDifferentProps }` would not make sense in typings land. If `Link` took in `foo` and `bar` as props, you would get errors.

What does make sense is to keep `as` restricted to intrinsic elements. We can always extend props to take in AllHTMLAttributes and that will always work.

packages/react-compose/src/next/README.md
dzearing: whoops, fixed :)

packages/react-compose/src/next/simplifyShorthand.tsx
dzearing: I'm a little confused. The array of keys within props are not a prop in props.

dzearing: good suggestion!

dzearing: updated typings.

packages/react-compose/src/next/simplifyShorthand.tsx
dzearing: I had a reason, but I don't think it's valid any more. Removed, but if I hit something I'll add a comment here.

packages/react-compose/src/next/simplifyShorthand.tsx
dzearing: resolveShorthandProps

packages/react-compose/src/next/simplifyShorthand.tsx
dzearing: yep. It's analogous to: `{ label: 'value' }, ['label']` returning `{ label: { children: 'value' }}`. Render the value as children of the slot.

The `mappedProp` slot feature is unaccounted for here. I could add it as needed, just not sure on the best solution other than to change the argument to an options bag and then have something like:

```
{
  props,
  shorthandPropNames,
  propMappings: {
    image: 'src'
  }
}
```

packages/react-compose/src/next/getSlots.test.tsx
dzearing: much better :)

packages/react-compose/src/next/getSlots.test.tsx
dzearing: I could do that especially for root.

packages/react-compose/src/next/getSlots.tsx
dzearing: Effectively, don't use `as` for the default slot container; rather separate it out into something which the user doesn't stomp on (which is what `createShorthand` is doing.) That way you can use slot containers other than primitives.

I think the analogy here would be to retain a defaultSlots collection which does not get merged. Later when we call `getSlots`, pass that along:

```tsx
const defaultSlots = { image: Image };
const { slots, slotProps } = getSlots(state, defaultSlots);
```

dzearing: @khmakoto @xugao

packages/react-compose/src/next/getSlots.tsx
dzearing: No; this is what lodash does, and it means that every prop must call indexOf on the array, which is slot. I agree it looks weird, but it is necessary for perf, which was the whole purpose of adding omit.

dzearing: https://jsperf.com/omit-vs-rest-vs-reduce/6 see "indexOf" solution perf.

packages/react-compose/src/next/getSlots.test.tsx
dzearing: not sure what i was thinking there. Good suggestion. fixed.

packages/react-compose/src/next/README.md
dzearing: root "as", at least.

packages/react-compose/src/next/simplifyShorthand.tsx
dzearing: renamed

packages/react-compose/src/next/getSlots.tsx
dzearing: Turns out I was wrong earlier on perf. For short lists of exclusions, indexOf seems faster than object maps. Who'd have thunk. Here's the perf results: https://jsperf.com/omit-vs-rest-vs-reduce/10

And here's the PR to fix: https://github.com/microsoft/fluentui/pull/14313

dzearing: merged and updated.

packages/react-compose/src/next/getSlots.tsx
dzearing: I added a test for this just to make sure we're good!

```jsx
it('can use slot children functions to replace default slot rendering', () => {
    expect(getSlots({ as: 'div', icon: { as: Foo, href: 'bar', children: (C, p) => <C {...p} /> } }, ['icon'])).toEqual(
      {
        slots: { root: 'div', icon: React.Fragment },
        slotProps: { root: {}, icon: { children: <Foo href="bar" /> } },
      },
    );
  });
```

packages/react-compose/src/next/getSlots.tsx
dzearing: Only primitive slots are excluded; this is an optiization.

I found that most slots have some logic which says: put a slot here, contain it in a DIV, but if the user doesn't specify anything, don't render it. Something like this:

```jsx
const Foo = (props) => {
  const { slots, slotProps } = getSlots(props, [ 'icon' ]);
  return (
    <slots.root { ... }>
      { props.icon && <slots.icon /> }
    </slots.root>
  );
};
```

And actually in practice, hardly any dev avoids prepends the condition, so you end up with Buttons which have containers rendered containing nothing.

So I thought it would be wise to assume this default behavior: a slot rendered without content shouldn't be rendered.

Of course there are exceptions like `input` or `hr` or whatnot, which don't render children. For exceptions, you'd render null children:

```jsx
getSlots({ as: 'div', { divider: { as: 'hr', children: null }})
```

This will render the `hr` even if the user has not specified the `divider` slot.

dzearing: And just to reiterate; this exclusion behavior is only for primitives. For component containers, the slot is rendered no matter what, because the container would own how to manage the slot.

This is a perf optimization. If this just is too confusing, we can avoid. But trying to cull unneeded elements by default.

packages/utilities/etc/utilities.api.md
dzearing: In usage the restriction of only allowing ReactHTML keys fought with the incoming typings, so I'd find myself casting tagname to `any` or `as unknown as React.ReactHTML` just to coerce things. 

The function can support any string. You can pass "foobar" to it which in the web components world could be a valid element, and it will still give you basic baseline filtered attributes.

packages/react-compose/src/next/README.md
dzearing: This for inspiration: https://immerjs.github.io/immer/docs/introduction

arrays will not be traversed, so your example would have created an object clone and the items would have been left unchanged.

I can see this being confusing. I think blind deep clones aren't good enough, and shallow clones are error prone as well. That's why I felt the need to have something besides a dumb deep merge that would do something a little more specific for draft state.

packages/react-compose/src/next/getSlots.tsx
dzearing: resolved.

packages/react-compose/src/next/README.md
dzearing: And yes; immutability does add overhead. This is what led me here. Trying to kill extra _rest, _merge, etc calls showing up in flamegraphs.

packages/react-compose/src/next/simplifyShorthand.tsx
dzearing: Yep! I haven't yet added this;

packages/react-compose/src/next/getSlots.test.tsx
dzearing: fixed

packages/react-compose/src/next/getSlots.test.tsx
dzearing: addressed most of the descriptions :)

packages/react-compose/src/next/simplifyShorthand.tsx
dzearing: Will leave this in a separate PR. I've made a note for pending change.

dzearing:
To summarize the issues, there are 4 pending that I will track as separate RFCs that we can follow up on. I want to unblock merging this and the button updates to consume it, so that I can be unblocked integrating the code into ribbon work asap. I'm happy to make changes though pending followup discussions, so rest assured checking this in does not imply a final form :)

* mergeProps - it deep clones objects and properties, but not refs, callbacks or arrays.
* `mappedProp` and `mappedArrayProp` are missing
* The `as` prop is currently used to define the container component. This makes it hard to pass `as` into that container.
* The `getSlots` optimization to not render primitive slots with undefined `children` is confusing.

packages/react-compose/src/next/README.md
dzearing: I believe this is ok; but happy to discuss further. I've opened a related RFC here https://github.com/microsoft/fluentui/issues/14365 and we can go through some alternatives if you want tweaks or see a better direction!

packages/react-compose/src/next/simplifyShorthand.tsx
dzearing: If it should do something different, we can certainly tweak.

packages/react-compose/src/next/getSlots.tsx
dzearing: I've opened an RFC to track this discussion here: https://github.com/microsoft/fluentui/issues/14373

We could certainly tweak things to address this case, but I think the current solution is most efficient. We're avoiding hard edges to container components, we don't have both "slots" and "as" to replace the container, and if you really need to render an `Image` with an extra `as` prop, you can do that through the children function.

In your particular example, you started with a default of using the `Image` component. Then the customer or parent component overrode that by using an `img` primitive. Because it's a primitive, getSlots would have slimmed down the props using `getNativeElementProps` and `fit` would have been removed. This is precisely the expected behavior.

packages/react-compose/src/next/getSlots.tsx
dzearing: I've opened an RFC here to track: https://github.com/microsoft/fluentui/issues/14374

Also scheduled some time next week. If this is too much magic, happy to revert.

packages/react-compose/src/next/README.md
dzearing: And also, one size doesn't fit all. If the deep clone does or does not do something you need done differently, we can either tweak the utility or use an alternative approach.

Created an RFC to discuss mutable approach vs immutable approach:
https://github.com/microsoft/fluentui/issues/14365

Also have scheduled time on friday to discuss :)

dzearing:
All issues have been resolved, or tracked in an RFC. I'd like to playtest this all in app scenarios, and because none of it should affect any current code, I'd like to merge this to keep progress unblocked.

We have time on Friday to discuss the remaining 4 RFCs and I'm more than happy to apply any changes deemed needed assuming they don't regress perf or any fundamental behaviors :)

PR #14270, opened 2020-07-30
Link examples: Fix accessibility and contrast issues in dark mode 

marygans:
#### Pull request checklist

#### Description of changes
Link and Message Bar: 
Remove UHF overrides to link component. Removed overrides to messageBar since the background has changed from transparent, there is no longer a need to make the links darker. Also added Text component to Link example. 

### Before:
(Hover and Rest state)
link-before
mssgbar

### After:
(Hover and Rest state)
linke-after
mssgbar-after

### Focus Areas to test: 
HIG and FluentUI resources website


packages/office-ui-fabric-react/src/components/Link/Link.doc.tsx
```diff
@@ -21,15 +21,11 @@ export const LinkPageProps: IDocPageProps = {
...
-                color: theme!.palette.themePrimary,
                 margin: 0,
                 padding: 0,
                 overflow: 'inherit',
                 textOverflow: 'inherit',
                 selectors: {
-                  ':active, :hover, :active:hover': {
-                    color: theme!.palette.themeDarker,
-                  },
                   ':focus': {
```
ecraig12345, line 29: Shoudn't this be removed too if you're removing the other state overrides? Potentially same for the disabled state.

packages/office-ui-fabric-react/src/components/Link/Link.doc.tsx
```diff
@@ -47,7 +43,9 @@ export const LinkPageProps: IDocPageProps = {
     },
   ],
   overview: require<string>('!raw-loader!office-ui-fabric-react/src/components/Link/docs/LinkOverview.md'),
-  bestPractices: require<string>('!raw-loader!office-ui-fabric-react/src/components/Link/docs/LinkBestPractices.md'),
+  bestPractices: '',
+  dos: require<string>('!raw-loader!office-ui-fabric-react/src/components/Link/docs/LinkDos.md'),
+  donts: require<string>('!raw-loader!office-ui-fabric-react/src/components/Link/docs/LinkDonts.md'),
```
ecraig12345, line 48: Revert this part

PR #14272, opened 2020-07-31
fix(Popup): share layers logic between Fabric + Fluent N*

ThomasMichon:
#### Description of changes

> Introduced a new `@fluentui/dom-utilities` package to contain the `dom` sub-group from `@uifabric/utilities`. These functions handle the 'virtual' parent hierarchy functions which account for logic that uses React Portals, such as `Layer` and `Popup`. Updated `@uifabric/utilities` to depend on this new package and re-export all migrated functions from their new locations.

Done in #15039.

Updated `react-northstar` (legacy Fluent) so make its `Popup` use `setVirtualParent` and `elementContains`, which ensure that a Fluent `Popup` is not closed automatically when interacting with a Fabric `Layer`.

Updated the `local-sandbox` example to showcase usage of a Fabric `ContextualMenu` within a Fluent `Popup`.

#### Focus areas to test

Validate existing `Popup` flows from `react-northstar`. There does not appear to be much in the way of example projects in this repository using this framework.

### DIFF
--- a/packages/fluentui/e2e/tests/fabricLayerInPopup-example.tsx
+++ b/packages/fluentui/e2e/tests/fabricLayerInPopup-example.tsx
@@ -1,0 +1,65 @@
+import { Button, Popup, Flex, Ref } from '@fluentui/react-northstar';
+import { ContextualMenu } from 'office-ui-fabric-react';
+import * as React from 'react';
+
+const selectors = {
+  buttonInPopup: 'button-in-popup',
+  menuTrigger: 'menu-trigger',
+  popupTrigger: 'popup-trigger',
+};
+
+const FabricLayerInPopupExample = () => {
+  const [menuOpen, setMenuOpen] = React.useState(false);
+  const [popupOpen, setPopupOpen] = React.useState(false);
+  const buttonRef = React.useRef(null);
+
+  return (
+    <>
+      <Popup
+        content={
+          <Flex vAlign="center" column>
+            <Ref innerRef={buttonRef}>
+              <Button
+                content="Open Fabric Menu"
+                id={selectors.menuTrigger}
+                onClick={() => {
+                  setMenuOpen(state => !state);
+                }}
+              />
+            </Ref>
+            <Button id={selectors.buttonInPopup} content="A button" />
+            <ContextualMenu
+              items={[
+                {
+                  key: 'item-1',
+                  text: 'item-1',
+                  onClick: () => console.log('Item 1: click'),
+                },
+                {
+                  key: 'item-2',
+                  text: 'item-2',
+                  onClick: () => console.log('Item 2: click'),
+                },
+              ]}
+              hidden={!menuOpen}
+              onDismiss={() => {
+                setMenuOpen(false);
+                console.log('Menu: dismiss');
+              }}
+              target={buttonRef}
+            />
+          </Flex>
+        }
+        onOpenChange={(e, data) => setPopupOpen(data.open)}
+        open={popupOpen}
+        position="below"
+        trigger={<Button content="Open N* Popup" id={selectors.popupTrigger} />}
+      />
+      <div id="outside" style={{ border: '1px solid green', position: 'fixed', top: 5, right: 5 }}>
+        An element outside
+      </div>
+    </>
+  );
+};
+
+export default FabricLayerInPopupExample;
--- a/packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
+++ b/packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
@@ -17,6 +17,7 @@ import { getCode, keyboardKey, SpacebarKey } from '@fluentui/keyboard-key';
 import * as _ from 'lodash';
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
+import { elementContains, setVirtualParent } from '@fluentui/dom-utilities';
 
 import {
   ChildrenComponentProps,
@@ -370,8 +371,8 @@ export const Popup: React.FC<PopupProps> &
 
   const shouldBlurClose = (e: React.FocusEvent) => {
     const relatedTarget = e.relatedTarget as Node;
-    const isInsideContent = popupContentRef.current?.contains(relatedTarget);
-    const isInsideTarget = e.currentTarget?.contains(relatedTarget);
+    const isInsideContent = elementContains(popupContentRef.current, relatedTarget as HTMLElement);
+    const isInsideTarget = elementContains(e.currentTarget as HTMLElement, relatedTarget as HTMLElement);
     // When clicking in the popup content that has no tabIndex focus goes to body
     // We shouldn't close the popup in this case
     return relatedTarget && !(isInsideContent || isInsideTarget);
@@ -473,7 +474,7 @@ export const Popup: React.FC<PopupProps> &
     const activeElement = activeDocument.activeElement;
 
     triggerFocusableRef.current =
-      triggerRef.current && triggerRef.current.contains(activeElement)
+      triggerRef.current && elementContains(triggerRef.current, activeElement as HTMLElement)
         ? (activeElement as HTMLElement)
         : triggerRef.current;
   };
@@ -509,6 +510,18 @@ export const Popup: React.FC<PopupProps> &
 
   const triggerNode = useTriggerElement(props);
   const triggerProps = getTriggerProps(triggerNode);
+
+  React.useEffect(() => {
+    if (open) {
+      setVirtualParent(popupContentRef.current, triggerRef.current);
+    }
+
+    return () => {
+      if (open && popupContentRef.current) {
+        setVirtualParent(popupContentRef.current, null);
+      }
+    };
+  }, [open]);
 
   const contentElement = (
     <Animation mountOnEnter unmountOnExit visible={open} name={open ? 'popup-show' : 'popup-hide'}>
--- a/packages/fluentui/react-northstar/src/utils/doesNodeContainClick.tsx
+++ b/packages/fluentui/react-northstar/src/utils/doesNodeContainClick.tsx
@@ -1,4 +1,5 @@
 import * as _ from 'lodash';
+import { elementContains } from '@fluentui/dom-utilities';
 
 /**
  * Determines if a click's coordinates are within the bounds of a node.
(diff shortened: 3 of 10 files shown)


packages/dom-utilities/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@fluentui/dom-utilities",
+  "version": "7.0.1",
+  "description": "DOM utilities for use within Fluent web components",
+  "private": true,
```
dzearing, line 5: if you want this published, you probably should remove this.

packages/dom-utilities/jest.config.js
```diff
@@ -0,0 +1,3 @@
+let path = require('path');
```
ecraig12345, line 1:
```suggestion
```

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
```diff
@@ -505,6 +506,8 @@ export const Popup: React.FC<PopupProps> &
     if (open) {
       // when new state 'open' === 'true', save the last focused element
       updateTriggerFocusableRef();
+
+      setVirtualParent(popupContentRef.current, triggerRef.current);
```
ThomasMichon, line 510: I wish I could add this to `PortalInner` somehow, but that component never automatically created an inner `div` as `Layer` does, so we can't guarantee existence of an element to mark.

packages/dom-utilities/.eslintrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": ["plugin:@fluentui/eslint-plugin/react"],
```
ecraig12345, line 2: Should probably use `react--legacy` here since this is moving old code which uses `I` prefixes on interface names.
```suggestion
  "extends": ["plugin:@fluentui/eslint-plugin/react--legacy"],
```

packages/dom-utilities/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "dependencies": {
+    "@uifabric/set-version": "^7.0.18",
+    "tslib": "^1.10.0"
+  },
+  "peerDependencies": {
+    "@types/react": ">=16.8.0 <17.0.0",
+    "@types/react-dom": ">=16.8.0 <17.0.0",
+    "react": ">=16.8.0 <17.0.0",
+    "react-dom": ">=16.8.0 <17.0.0"
+  }
```
ecraig12345, line 48: Doesn't look like this package uses React, so unless you expect it to in the future you can remove all React stuff. You can probably also remove `@types/webpack-env`.
```suggestion
    "@uifabric/build": "^7.0.0"
  },
  "dependencies": {
    "@uifabric/set-version": "^7.0.18",
    "tslib": "^1.10.0"
  }
```

packages/dom-utilities/README.md
```diff
@@ -0,0 +1,11 @@
+# @fluentui/dom-utilities
+
+**Dom Utilities components for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
```
ecraig12345, line 3: Can you update this to reflect the actual content of the package?

packages/dom-utilities/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "strict": true,
+    "moduleResolution": "node",
+    "preserveConstEnums": true,
+    "lib": ["es5", "dom"],
+    "typeRoots": ["../../node_modules/@types", "../../typings"],
+    "types": ["jest", "webpack-env", "custom-global"],
+    "paths": {
+      "@fluentui/dom-utilities/lib/*": ["./src/*"],
+      "@fluentui/dom-utilities": ["./src"]
+    }
```
ecraig12345, line 23: You can probably simplify this as follows:
```suggestion
    "lib": ["es5", "dom"]
```

packages/fluentui/local-sandbox/package.json
```diff
@@ -10,7 +10,8 @@
     "serve": "node ./server.js"
   },
   "dependencies": {
-    "@fluentui/react-northstar": "^0.51.0"
+    "@fluentui/react-northstar": "^0.51.0",
+    "office-ui-fabric-react": "^7.124.0"
```
dzearing, line 14: I'm curious if this will cause any issues. (I'm not familiar with the local-sandbox package, @mnajdova @miroslavstastny or @layershifter should double check)

packages/dom-utilities/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+These are not production-ready components and **should never be used in product**. This space is useful for testing new components whose APIs might change before final release.
```
jurokapsiar, line 5: is this still true?

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
jurokapsiar, line 510: do we need cleanup? If popupContent or trigger gets unmounted, won't the references stay and cause leaks?

layershifter: > do we need cleanup?

Yes, it should be done.

packages/fluentui/local-sandbox/package.json
layershifter: The purpose of `local-sandbox` is to be to able to test some code locally with your current changes like in 3rd app.

I don't think that changes there and in `packages/fluentui/local-sandbox/src/index.tsx` are needed, I suggest to move them to E2E tests.

packages/fluentui/react-northstar/src/utils/doesNodeContainClick.tsx
```diff
@@ -23,7 +24,7 @@ export const doesNodeContainClick = (
 
     if (target.querySelector('[data-suir-click-target=true]')) {
       _.invoke(e.target, 'removeAttribute', 'data-suir-click-target');
-      return node.contains(e.target as HTMLElement);
+      return elementContains(node, e.target as HTMLElement);
```
layershifter, line 27: Change there looks a bit scary as `doesNodeContainClick()` is used widely.

ThomasMichon: However, this is the core purpose of this PR. We want to ensure that clicking into a Fabric Layer does not cause Fluent Popups to close, so clicking into a 'virtual' child needs to be treated as clicking 'within' the element.

packages/dom-utilities/README.md
dzearing: (no i don't believe this is true, good catch. this is the default commentary from the create-project template)

packages/dom-utilities/tsconfig.json
ThomasMichon: Sounds like someone should update the `create-package` tool, then. Or provide a base `tsconfig.json` for which we can use `extends`.

ecraig12345: Not all the suggestions here apply universally, and some things (`paths` and `baseUrl`, maybe others) will always need to be specified in each package separately even if we have config extension. (Basically the reason we don't have config extension is that no one has had/made time to set it up.)

In the create-package template, we include the paths by default because they're needed in projects that have examples (which use the full package names for importing). This project will never have examples, so it doesn't need the paths. It would potentially be possible to add a question to create-package about examples to control whether this stuff is added.

For the types and typeRoots, I realize now that removing `jest` types was incorrect (not sure what I was thinking there, sorry). `webpack-env` and `custom-global` probably will not be strictly needed, but it doesn't hurt having them there, so the only reason for removing them was to make the config a little simpler.

layershifter:
This PR breaks our `yarn start` experience:

image

Required changes in `packages/fluentui/docs/tsconfig.json`:

```diff
      "@fluentui/date-time-utilities": ["packages/date-time-utilities/src"],
+      "@fluentui/dom-utilities": ["packages/dom-utilities/src"],
      "src/*": ["packages/fluentui/react-northstar/src/*"],
```

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
layershifter: Is this call valid? 😕  In `packages/office-ui-fabric-react/src/components/Layer/Layer.base.tsx` it will be called with `setVirtualParent(layerElement, this._rootRef.current!);` where:
- `layerElement` is an element where portal's content will be rendered
- some `span` element that will be always rendered inline 

In our case `triggerRef` points to some element, but it's optional and can be missing as `trigger` is not required...

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
```diff
@@ -374,7 +375,7 @@ export const Popup: React.FC<PopupProps> &
     return (
       !e.currentTarget ||
       !popupContentRef.current ||
-      (!e.currentTarget.contains(e.relatedTarget) && !popupContentRef.current.contains(e.relatedTarget))
+      (!elementContains(e.currentTarget, e.relatedTarget) && !elementContains(popupContentRef.current, e.relatedTarget))
```
layershifter, line 378: Is there a clear statement when we should use `elementContains()` vs `.contains()`?

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
ThomasMichon: It was extremely difficult looking at this code and even debugging to figure out which element was the popup element and which was the host. Some guidance would be appreciated.

packages/fluentui/react-northstar/src/components/Popup/Popup.tsx
ThomasMichon: We should now always use `elementContains` so that the virtual-parent thing works nicely. I feel like React should have provided this function now that `createPortal` is a thing.

packages/fluentui/react-northstar/src/utils/doesNodeContainClick.tsx
layershifter: ⚠️ This change causes unit test failure in `Portal-test.tsx`:

```jsx
node.contains(e.target as HTMLElement); // true
elementContains(node, e.target as HTMLElement); // false
```

dzearing (approved):

packages/dom-utilities/src/elementContains.ts
```diff
@@ -14,7 +14,9 @@ export function elementContains(
   let isContained = false;
 
   if (parent && child) {
-    if (allowVirtualParents) {
+    if (parent === child) {
+      isContained = true;
+    } else if (allowVirtualParents) {
```
layershifter, line 19: https://github.com/microsoft/fluentui/pull/15089

packages/fluentui/e2e/tests/fabricLayerInPopup-example.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const menuItems = [
+  {
+    key: 'newItem',
+    text: 'New',
+    onClick: () => console.log('New clicked'),
+  },
+];
+
+export default PopupParent;
```
layershifter, line 82: ⚠️ Should be fixed before merge, E2E test should be added

PR #14274, opened 2020-07-31
@fluentui/react-button: Removing compose in favor of hooks and factories.

dzearing:
Note that this PR is a branch of the other PR up for review which adds helpers in react-compose/lib/next. Once that's merged, this one should be reduced.

This change removes all button base components. It removes all `compose` usage.
All things are converted into create* factory functions, which can create components, their appropriate reusable hooks, and scaffolded components which use the factories to build components glued to styling (also provided by hooks.)

Let's see how this pans out in perf run.


packages/react-theme-provider/src/useInlineTokens.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Hook which given draftState, will ensure that tokens are spit out to inline styles.
+ * @param draftState - state to read and manipulate. Expected to have `tokens` prop, will
+ * transform into inline
+ */
+export const useInlineTokens = (draftState: { style?: React.CSSProperties; tokens?: TokenSetType }) => {
+  const { tokens, style } = draftState;
+
+  if (tokens) {
+    draftState.style = style || {};
+    tokensToStyleObject(tokens, '--button', draftState.style);
```
xugao, line 14: prefix `--button` needs to be passed

packages/react-theme-provider/src/tokensToStyleObject.ts
```diff
@@ -1,26 +1,23 @@
...
+  style: React.CSSProperties | undefined = {},
 ): React.CSSProperties => {
-  for (const name in tokens) {
-    if (tokens.hasOwnProperty(name)) {
+  if (tokens) {
+    for (const name of Object.keys(tokens)) {
       const varName = prefix ? `${prefix}${name === 'default' ? '' : '-' + name}` : `--${name}`;
-      const varValue = tokens[name];
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const varValue = (tokens as any)[name];
```
xugao, line 12: why is `as any` needed here?

packages/react-button/src/components/Button/createButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+      as: props.href ? 'a' : 'button',
+      icon: { as: 'span' },
+      children: { as: 'span' },
+      loader: { as: 'span' },
+    },
+    defaultProps,
+    simplifyShorthand(props, buttonShorthandProps),
+  );
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
xugao, line 44: which rule are we violating here?

packages/react-button/src/components/Button/createButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+      {!iconOnly && <slots.children {...slotProps.children} />}
+      {iconPosition === 'after' && <slots.icon {...slotProps.icon} />}
+    </slots.root>
+  );
+};
+
+/**
+ * Given user props, returns state and render function for a Button.
+ */
+export const createButton = (props: ButtonProps, ref: React.Ref<HTMLElement>, defaultProps?: ButtonProps) => {
```
xugao, line 31: this is a hook, we should call it `useButton`

packages/react-button/src/components/Button/useButton.ts
```diff
@@ -1,28 +1,34 @@
...
  */
-export const useButton = (
-  props: ButtonProps,
-  ref: React.Ref<HTMLElement>,
-  options: ComposePreparedOptions,
-): ButtonState => {
-  const buttonRef = React.useRef<HTMLButtonElement | null>(null);
+export const useButton = (draftState: ButtonState) => {
+  // Update the button's tab-index, keyboard handling, and aria attributes.
+  if (draftState.as !== 'button') {
```
xugao, line 11: nit: can we invert these if conditions so we can return early without nesting code for readability

packages/react-button/src/components/MenuButton/MenuButton.tsx
```diff
@@ -1,15 +1,30 @@
...
+    menu: { as: ContextualMenu },
+  });
+
+  // Styling hooks.
+  useButtonClasses(state);
+  useMenuButtonClasses(state);
+  useFocusRects(state.ref);
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  useInlineTokens(state as any);
```
xugao, line 25: why is `as any` needed here?

packages/react-button/src/components/MenuButton/MenuButton.tsx
```diff
@@ -1,15 +1,30 @@
...
+import { useInlineTokens } from '@fluentui/react-theme-provider';
 
-export const MenuButton = compose<'button', MenuButtonProps, MenuButtonProps, {}, {}>(MenuButtonBase, {
-  classes: createClassResolver(classes),
-  displayName: 'MenuButton',
-  slots: {
-    menu: ContextualMenu,
-    menuIcon: ChevronDownIcon,
-  },
+export const useMenuButtonClasses = makeClasses(classes);
```
xugao, line 11: i think we should let `useMenuButtonClasses` to call `useButtonClasses`
```
const useOnlyMenuButtonClasses = makeClasses(classes);
const useMenuButtonClasses = state => {
  useButtonClasses(state);
  useOnlyMenuButtonClasses(state);
}
```

packages/react-button/src/components/Button/Button.stories.tsx
```diff
@@ -35,6 +35,22 @@ const ButtonVariants = (props: ButtonProps) => (
   </Stack>
 );
 
+export const ButtonFocus = () => {
+  const buttonRef = React.useRef<HTMLElement | null>(null);
+  return (
+    <Stack>
+      <Button
+        onClick={() => {
+          // ok
```
khmakoto, line 44: Remove this comment?

packages/react-button/src/components/Button/createButton.tsx
dzearing: A function `create*` can not contain a hook `use*`.

There is no way to allow this scenario without disabling the rule.

packages/react-button/src/components/Button/createButton.tsx
dzearing: This is a factory function, but yes, maybe you're right.

`useButtonFactory`?

dzearing: then we wouldn't have to disable the rule :)

packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -1,87 +1,94 @@
...
-  tokens?: RecursivePartial<ButtonTokens>;
-}
+export type ButtonProps = ComponentProps &
+  React.HTMLAttributes<HTMLElement> & {
+    /**
+     * Shorthand icon. A shorthand prop can be a literal, object, or
+     * JSX. The `children` prop of the object can be a render function,
+     * taking in the original slot component and props.
+     */
+    icon?: ShorthandProps;
```
mnajdova, line 16: Ideally this would be generic to accept the `IconProps` or whatever is the shorthand

packages/react-button/src/components/Button/createButton.tsx
mnajdova: I would say that this is the `useButton` hook. The other one should either be called: `useButtonAria` if it handles only a11y, or `useButtonState`.

packages/react-button/src/components/Button/createButton.tsx
mnajdova: This is a clear indicator then that this should be hook :)

packages/react-button/src/components/Button/useButton.ts
```diff
@@ -1,28 +1,34 @@
...
+ * The useButton hook processes the Button draft state.
+ * @param draftState - Button draft state to mutate.
  */
-export const useButton = (
-  props: ButtonProps,
-  ref: React.Ref<HTMLElement>,
-  options: ComposePreparedOptions,
-): ButtonState => {
-  const buttonRef = React.useRef<HTMLButtonElement | null>(null);
+export const useButton = (draftState: ButtonState) => {
```
mnajdova, line 9: This definitely look like `useButtonAria`.

packages/react-button/src/components/MenuButton/createMenuButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+    },
+    defaultProps,
+    simplifyShorthand(props, menuButtonShorthandProps),
+  ) as MenuButtonState;
+
+  // It behaves like a button.
+  useButton(state);
+
+  // It can be expanded.
+  useExpanded(state);
```
mnajdova, line 55: I think ideally we should have a hook `useMenuButtonState`, that would internally use the `useButtonState` hook. Then we can expose this hook so this component can be extended

packages/react-button/src/components/Button/createButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+      icon: { as: 'span' },
+      children: { as: 'span' },
+      loader: { as: 'span' },
+    },
+    defaultProps,
+    simplifyShorthand(props, buttonShorthandProps),
+  );
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  useButton(state);
```
mnajdova, line 45: I personally don't like hooks that mutates input. If a client does not need state, these hooks can potentially be invoked with `props`.

packages/react-button/src/components/MenuButton/createMenuButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+    {
+      ref,
+      as: props.href ? 'a' : 'button',
+      icon: { as: 'span' },
+      children: { as: 'span' },
+      menuIcon: { as: 'span' },
+      menu: { as: 'span' },
+    },
+    defaultProps,
+    simplifyShorthand(props, menuButtonShorthandProps),
```
mnajdova, line 48: Feel risky to expect people always to know to send this as last param... Can we extract this to a separate hook, something like:

```
const state = resolveState(
    menuButtonShorthandProps, // slots
    {
      ref,
      as: props.href ? 'a' : 'button',
      icon: { as: 'span' },
      children: { as: 'span' },
      menuIcon: { as: 'span' },
      menu: { as: 'span' },
    },
    defaultProps,
)

```

packages/react-button/src/components/Button/Button.types.tsx
```diff
@@ -1,87 +1,94 @@
...
+     * and a non-anchor as the render type will prevent the Button from behaving like a
+     * hyperlink and opening the href on click.
+     */
+    href?: string;
+
+    /**
+     * Defines the target window to open the href in. Only is applied if the button renders
+     * as an anchor tag, which is the default behavior if href is provided.
+     */
+    target?: string;
```
layershifter, line 40: I don't think that it's the valid approach to add HTML props for specific element types. This will not scale as it will too easy to miss some property.

packages/react-button/src/components/Button/Button.tsx
```diff
@@ -1,9 +1,27 @@
...
+
+/**
+ * Define a styled Button, using the `createButton` factory.
+ */
+export const Button = React.forwardRef<HTMLElement, ButtonProps>((props, ref) => {
+  const { render, state } = createButton(props, ref);
+
+  // Apply styling.
+  useButtonClasses(state);
+  useFocusRects(state.ref);
```
layershifter, line 20: It looks that `state` returns `any` or something close to it.

- `ref` that comes from `forwardRef` defaults to `null` ⏭️ 
- after merging via `mergeProps` `ref` will be still `null` ⏭️ 
- `useFocusRects` will point to invalid `window`

https://github.com/microsoft/fluentui/issues/14062

packages/react-button/src/components/Button/useButton.ts
layershifter, line 11: Using `as` is not enough to determine `tagName`

packages/react-button/src/components/Button/useButton.ts
```diff
@@ -1,28 +1,34 @@
...
-  };
+      draftState['data-isFocusable'] = true;
+
+      draftState.onKeyDown = (ev: React.KeyboardEvent<HTMLElement>) => {
+        if (onKeyDown) {
+          onKeyDown(ev);
+        }
+
+        if (!ev.defaultPrevented && onClick && (ev.which === 20 || ev.which === 13)) {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
layershifter, line 26: It means that typings are not correct and it can be surprise for a client to receive `React.KeyboardEvent` instead of `React.MouseEvent`

packages/react-button/src/components/Button/createButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+};
+
+/**
+ * Given user props, returns state and render function for a Button.
+ */
+export const createButton = (props: ButtonProps, ref: React.Ref<HTMLElement>, defaultProps?: ButtonProps) => {
+  const state = mergeProps(
+    {
+      ref,
+      as: props.href ? 'a' : 'button',
```
layershifter, line 35: I would like to keep this less magic and require from a client to pass `as='a' href='#'`

packages/react-button/src/components/ToggleButton/useChecked.ts
```diff
@@ -0,0 +1,44 @@
...
+import { useControllableValue } from '@uifabric/react-hooks';
+
+export interface CheckedState {
+  checked?: boolean;
+  defaultChecked?: boolean;
+  onClick?: React.DOMAttributes<HTMLElement>['onClick'];
+
+  role?: string;
+  'aria-checked'?: boolean | 'false' | 'mixed' | 'true';
+  'aria-pressed'?: boolean | 'false' | 'mixed' | 'true';
```
layershifter, line 11: I suggest to reuse `React.AriaAttributes`

```suggestion
  'aria-pressed'?: React.AriaAttributes['aria-pressed'];
```

packages/react-button/src/components/Button/createButton.tsx
dzearing: Agreed. Renaming this to `useButtonFactory`. Other suggestions welcome

packages/react-button/src/components/Button/createButton.tsx
dzearing: alright i'll go with that.

packages/react-button/src/components/Button/useButton.ts
dzearing: It's a little more than just adding aria attributes. I'll go with `useButtonState`.

packages/react-button/src/components/Button/Button.types.tsx
dzearing: The problem I hit, which open to ideas, is that if I extend HTMLButtonAttributes and HTMLAnchorAttributes, their onClick definitions conflict and typescript disallows it. Maybe I can omit that conflicts.

packages/react-button/src/components/Button/useButton.ts
dzearing: Better suggestion?

packages/react-theme-provider/src/useInlineTokens.ts
dzearing: fixed.

packages/react-button/src/components/Button/createButton.tsx
dzearing: This has been renamed to `useButton`.

packages/react-button/src/components/Button/useButton.ts
layershifter: https://github.com/reakit/reakit/blob/9bd6882fb50948c616e757d750b436decc36ddfc/packages/reakit/src/Button/Button.ts#L40
https://github.com/microsoft/fluentui/pull/13717#discussion_r454935803

packages/react-button/src/components/MenuButton/MenuButton.tsx
dzearing: I've added a comment on the error but i don't get why. will see if I can resolve with you offline.

packages/react-theme-provider/src/tokensToStyleObject.ts
dzearing: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'TokenSetType'.
  No index signature with a parameter of type 'string' was found on type 'TokenSetType'.

dzearing: export type TokenSetType = string | { [key: string]: TokenSetType | undefined };

dzearing: oh nevermind; I see the issue. typings fixed.

packages/react-button/src/components/MenuButton/MenuButton.tsx
dzearing: State in `useInlineTokens` is expected to be:

```jsx
{ style?: React.CSSProperties; tokens?: TokenSetType }
```
The object we're passing in of type `MenuButtonState` should indeed have these.

Typescript errors with:
> Type 'MenuButtonState' has no properties in common with type '{ style?: CSSProperties | undefined; tokens?: TokenSetType | undefined; }'.

Now, the state type should indeed have these well defined:

```
export type MenuButtonProps = Omit<ButtonProps, 'iconPosition' | 'loader'> & {
```

Where `ButtonProps` is defined as :

```jsx
export type ButtonProps = ComponentProps &
  React.HTMLAttributes<HTMLElement> & {
    // ...
    tokens?: RecursivePartial<ButtonTokenSet>;
};
```
TypeScript things that neither style or tokens are valid in this type. I'm stumped here.

packages/react-button/src/components/Button/createButton.tsx
dzearing: I don't disagree. But I see pros and cons of both approaches.

I was thinking about immer.js in the draft state idea. Make one copy of props, layering default props, extended defaults, and then user props on top. Then use the draft to construct the final state for rendering usage.

Without a draft state concept, the mergeProps deep clone must be done on every individual hook layer, and the caller must be responsible for chaining.

```tsx
let state = useButtonState(props); // deep clone
state = useExpanded(state);  // deep clone
state = useChecked(state); // deep clone
state = useStyles(state); // deep clone
state = useFocusRect(state);

return { render, state };
```

I don't think mutations are great. The concept fights against much of what react state, redux and other immutable approaches out there preach. I tried calling out "draftState" as the input rather than "state" to use a naming pattern to opt into mutability in hopes that might make things more clear.

I don't think every hook needs to assume mutatibility. Atomic hooks should stay atomic. Primarily integrating behavioral hooks is where I see benefit. Less mistakes, better performance, less runtime penalties for splitting logic out into smaller units.

The perf gains are pretty clear. Before, I was seeind "_spread" and "_rest" coming up here and there adding a few samples in the flamegraph on each usage. They pile up fast. Soon in a single render pass your finding 10-20% overhead in just creating and recreating objects. In large apps you'll see huge chunks of time spent in garbage collection caused by lots of little allocations and deallocations. Mutability reduces some of that GC pressure. 

So I dunno. I'm not liking the lack of purity in concept; but the practicality of it has merit.

packages/react-button/src/components/MenuButton/createMenuButton.tsx
dzearing: well, in the spirit of smaller functional lego blocks, I was trying to keep it separate because not every component has shorthand props.

One idea might be:

```jsx
mergeProps( // or resolveState if that's more clear
  { /* options, including shorthandProps */ },
  ...propSets
);
```

packages/react-button/src/components/Button/createButton.tsx
dzearing: Assuming this, it means that `<Button href="..."/>` would not work as expected by the caller. Why should the caller know that `as` should be an `a` tag to make href work correctly?

I'm trying to understand why this is better for the customer. Historically we've always had this logic even without `as`; if a user provides an href, we render the button as an `a` tag to ensure the browser behaviors work correctly.

packages/react-button/src/components/ToggleButton/useChecked.ts
dzearing: good call!

packages/react-button/src/components/MenuButton/createMenuButton.tsx
mnajdova: Anything that would not require from the clients to invoke `simplifyShorthand` would be better in my opinion..

packages/react-button/src/components/Button/createButton.tsx
mnajdova: I think this deserves an RFC so we will all have a chance to experiment with it and go deeper, as in my opinion is important decision overall.

packages/react-button/src/components/MenuButton/useExpanded.ts
```diff
@@ -0,0 +1,86 @@
...
+  );
+
+  // Assign extra props to the menu slot.
+  draftState.menu = {
+    ...draftState.menu,
+    target: rootRef,
+    onDismiss: React.useCallback(() => {
+      if (onMenuDismiss) {
+        onMenuDismiss();
+      }
```
khmakoto, line 77: You should be able to do

```suggestion
      onMenuDismiss?.();
```

packages/react-button/src/components/MenuButton/useMenuButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+const renderMenuButton = (state: MenuButtonState) => {
+  const { slots, slotProps } = getSlots(state, menuButtonShorthandProps);
+  const { iconOnly, expanded } = state;
+
+  return (
+    <slots.root {...slotProps.root}>
+      <slots.icon {...slotProps.icon} />
+      {!iconOnly && <slots.children {...slotProps.children} />}
+      {!iconOnly && <slots.menuIcon {...slotProps.menuIcon} />}
```
khmakoto, line 21: Should we render the `spans` even when these are not present?

khmakoto: Similar for the other render functions.

packages/react-button/src/components/MenuButton/useMenuButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+/**
+ * Redefine the component factory, reusing button factory.
+ */
+export const useMenuButton = (props: MenuButtonProps, ref: React.Ref<HTMLElement>, defaultProps?: MenuButtonProps) => {
+  const state = mergeProps(
+    {
+      ref,
+      as: props.href ? 'a' : 'button',
```
khmakoto, line 34: I understand this logic for a regular button. However, since we are going to the trouble of separating these components and their states, should we remove this along with support for `hrefs` in menu and split buttons?

packages/react-button/src/components/Button/useButton.ts
dzearing: I've resolved this by calling `click` to force a `MouseEvent`.

packages/react-button/src/components/Button/useButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+      loader: { as: 'span' },
+    },
+    defaultProps,
+    resolveShorthandProps(props, buttonShorthandProps),
+  );
+
+  useButtonState(state);
+
+  return { state, render: renderButton };
+};
```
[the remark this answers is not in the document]
dzearing: Yeah i think that's very reasonable.

packages/react-button/src/components/MenuButton/useMenuButton.tsx
dzearing: sure thing

packages/react-button/src/components/Button/createButton.tsx
dzearing: nevermind, after thinking about this, i am good with this

packages/react-button/src/components/MenuButton/useMenuButton.tsx
dzearing: Ah, now you are seeing why I made the `getSlots` optimization to say: If the slot is a primitive, do not render it if children is `undefined`. It eliminates all of the redundant condition checks and makes things consistent.

I am starting to think it should be reverted because it causes magic.

dzearing: https://github.com/microsoft/fluentui/issues/14374

packages/react-button/src/components/MenuButton/useMenuButton.tsx
dzearing: Resolved

packages/react-button/src/components/MenuButton/useExpanded.ts
dzearing: Fixed.

packages/react-button/src/components/Button/useButton.tsx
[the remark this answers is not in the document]
dzearing: moved!

packages/react-button/src/components/Button/createButton.tsx
dzearing: fixed!

packages/react-button/src/components/Button/useButton.ts
dzearing: I think this is good enough; I would hate to add perf overhead calling `useEffect` to evaluate if we need to re-evaluate attributes and then re-rendering in this case.

In the worse case scenario, we find `{as: Foo}` and we treat it like a `span`. This technically is more work than needed but shouldn't be breaking.

packages/react-button/src/components/Button/Button.types.tsx
dzearing: I agree with you here though. I have an idea for fixing the typings to extend the right attributes based on the `as` prop, and it works but will need to be a separate PR.

dzearing: Tracking: https://github.com/microsoft/fluentui/issues/14414

dzearing:
Merging so that we can try this changes in product scenarios and iterate on them as needed.

PR #14283, opened 2020-07-31
DetailsList/GroupedList: Passing focusZone props down to GroupedList when groups used

### LINKED ISSUE #14275: DetialsList: onActiveItemChanged does not invoke when groups is set
### Environment Information

- **Package version(s)**: 7.124.0
- **Browser and OS versions**: chrome

https://codepen.io/ngyukman/pen/JjGQgVe?editors=1010

in 7.121.11 it is working as expected

#### Actual behavior:

onActiveItemChanged is no longer invoked once groups is set

#### Expected behavior:

it should work the same with or without groupings

Are you willing to submit a PR to fix? (Yes, No) Yes, depends if I could find a solution

micahgodbolt:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #14275

#### Description of changes
Seems nested FocusZones don't work with `onActiveItemChanged` when applied to the outer focuszone, causing an issue in DetailsList when that callback is called. 

https://codepen.io/ngyukman/pen/JjGQgVe?editors=1011

In the past DetailsList wrapped the same FocusZone around the List and GroupedList. As we're moving to GroupedList being its own focus managed control, I moved the DetailsList FocusZone to only wrap the list, and passed the FocusZone props to a new focusZoneProps in GroupedList.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/GroupedList/GroupedList.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/GroupedList/GroupedList.base.tsx
@@ -111,12 +111,12 @@ export class GroupedListBase extends React.Component<IGroupedListProps, IGrouped
 
     const { version } = listProps;
 
-    const { isInnerZoneKeystroke = this._isInnerZoneKeystroke, shouldEnterInnerZone } = focusZoneProps;
+    const { shouldEnterInnerZone = this._isInnerZoneKeystroke } = focusZoneProps;
 
     return (
       <FocusZone
         {...focusZoneProps}
-        shouldEnterInnerZone={shouldEnterInnerZone || isInnerZoneKeystroke}
+        shouldEnterInnerZone={shouldEnterInnerZone}
         direction={FocusZoneDirection.vertical}
         className={css(this._classNames.root, focusZoneProps.className)}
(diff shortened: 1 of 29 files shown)


packages/office-ui-fabric-react/src/components/GroupedList/GroupedList.base.tsx
```diff
@@ -102,11 +111,14 @@ export class GroupedListBase extends React.Component<IGroupedListProps, IGrouped
 
     const { version } = listProps;
 
+    const { shouldEnterInnerZone = this._isInnerZoneKeystroke } = focusZoneProps;
+
     return (
       <FocusZone
-        isInnerZoneKeystroke={this._isInnerZoneKeystroke}
+        {...focusZoneProps}
```
xugao, line 118: i think we should only override `focusZoneProps` with `direction` and `className` here

micahgodbolt:
@ThomasMichon We'd like to merge this change as soon as possible, is there someone who'd be able to review this and ensure it doesn't break any of your scenarios?

micahgodbolt:
Thomas gave a thumbs up to the changes. Going to merge.

PR #14302, opened 2020-08-03
Fix BasePicker's onDismiss behavior from changing if getTextFromItem provided

### LINKED ISSUE #10892: Picker `getTextFromItem` changes dismiss behavior
### Environment Information
- 7.52.0
- Version 77.0.3865.120 (Official Build) (64-bit)

Tried to reproduce here, but it doesn't work properly
https://codesandbox.io/s/picker-calling-onchange-incorrectly-dpwr0

Created video to demonstrate and walk through bug
https://microsoft.sharepoint.com/:v:/t/convsysresearch/ERMRSnt0UNRFuJ0h5uKdS8cBhXbKyC5ATXYvVWj28kOfSA?e=V1aiQf

#### Actual behavior:
First item of suggestions list is highlighted and selected by default when Picker is dismissed

#### Expected behavior:
Should not add tag. When dismissing the picker the user intention is not do do anything with picker anymore so it should not do anything. Especially when the user clicks a cancel button.

As stated in video there are two bugs:
1. `getTextFromItems` should not change dismiss / onChange behaviors.
It seems like it would just be called turing some `map` operation to translate items into text, the fact that it can change component behavior is very strange and scary. Makes you wonder what other behaviors / bugs exist only because you set one seemingly innocent attribute differently.

2. `onDismiss` should still be called when dismissing the picker directly to another Office Fabric component. It shouldn't matter what component you click on. This is also very weird.

Requested priority: (Blocking, High, **Normal**, Low)

Products/sites affected: Conversation Learner

17jlee2:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10892

#### Description of changes

As suggested in #10892, this changes the return type of the BasePicker's onDismiss prop to optionally return a boolean. If it returns false, then the first suggested item is not selected when the suggestions callout is dismissed. If it returns anything else, then the default behavior of it selecting the first suggestion is preserved.

#### Focus areas to test

The BasePicker and its derivatives (like TagPicker)


17jlee2:
> ## **Asset size changes**
> ⚠️ Insufficient baseline data to detect size changes
> 
> Unable to find bundle size details for Baseline commit: [bc90b07](https://github.com/microsoft/fluentui/commit/bc90b07e1da53cd189082f7b7bbf4d1118967c95)
> 
> **Possible causes**
> 
> * The baseline build [bc90b07](https://github.com/microsoft/fluentui/commit/bc90b07e1da53cd189082f7b7bbf4d1118967c95) is broken
> * The Size Auditor run for the baseline build [bc90b07](https://github.com/microsoft/fluentui/commit/bc90b07e1da53cd189082f7b7bbf4d1118967c95) was not triggered
> 
> **Recommendations**
> 
> * Please merge your branch for this Pull request with the latest master build and commit your changes once again

@xugao @joschect  Hey! I'm not very familiar with the workflow here. Is the recommendation saying to update my fork with the main repo, to create another PR, or something else entirely?

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -199,16 +199,22 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
 
   public dismissSuggestions = (ev?: any): void => {
     const selectItemFunction = () => {
+      let addItemOnDismimss: boolean | void = true;
       if (this.props.onDismiss) {
-        this.props.onDismiss(
+        addItemOnDismimss = this.props.onDismiss(
```
xugao, line 204: instead of making `onDismiss` to return a `boolean` (un-intuitive interface to have for `onDismiss`), why can't you just call `ev.preventDefault();` in your `onDismiss` handler?

xugao:
if you pull master again , the size auditor error should go away
>Recommendations
>Please merge your branch for this Pull request with the latest master build and commit your changes once again

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
17jlee2: It's definitely weird, but it looks like from #10892 and my own personal experimentation, preventDefault doesn't actually prevent all events from performing their default behavior. I think @joschect suggested modifying onDismiss to return a boolean to work around that.

17jlee2:
> if you pull master again , the size auditor error should go away
> 
> > Recommendations
> > Please merge your branch for this Pull request with the latest master build and commit your changes once again

I think I understand, thanks!

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
xugao: i see. okay

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -1636,7 +1636,7 @@ export interface IBasePickerProps<T> extends React.Props<any> {
     itemLimit?: number;
     onBlur?: React.FocusEventHandler<HTMLInputElement | Autofill>;
     onChange?: (items?: T[]) => void;
-    onDismiss?: (ev?: any, selectedItem?: T) => void;
+    onDismiss?: (ev?: any, selectedItem?: T) => boolean | void;
```
joschect, line 1639: ITT that this return should be `boolean | undefined` since void isn't really a return value.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -199,16 +199,22 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
...
           this.suggestionStore.currentSuggestion ? this.suggestionStore.currentSuggestion.item : undefined,
         );
       }
 
       if (!ev || (ev && !ev.defaultPrevented)) {
-        // Select the first suggestion if one is available when user leaves.
-        if (this.canAddItems() && this.suggestionStore.hasSelectedSuggestion() && this.state.suggestedDisplayValue) {
+        // Select the first suggestion if one is available and permitted by onDismiss when user leaves.
+        if (
+          addItemOnDismimss !== false &&
```
joschect, line 213: If the return type is `boolean | undefined` you can keep this like this here and remove line 202 that sets it to true. Javascript does some weird stuff but luckily `undefined !== false`.

packages/office-ui-fabric-react/src/components/pickers/BasePicker.types.ts
```diff
@@ -191,9 +191,11 @@ export interface IBasePickerProps<T> extends React.Props<any> {
   onInputChange?: (input: string) => string;
 
   /**
-   * A callback to override the default behavior of adding the selected suggestion on dismiss.
+   * A callback to override the default behavior of adding the selected suggestion on dismiss. If it returns true or
```
joschect, line 194: Given the earlier comments I might change this to `if it returns true or undefined` rather than `nothing` to be more explicit. But this is not required.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
17jlee2: I wanted to keep this backwards compatible, and it looks like changing void to undefined would require for old onDismiss's to explicitly return undefined (the compiler complains when assigning a function that doesn't return anything to () => void).

PR #14313, opened 2020-08-03
omit: perf improvements, better api, export from index

dzearing:
Changes omit to use `for in` loop and array/indexOf for matching exclusions. This seems fastest for small sets of exclusions (and is also the most ergonomic api.)

Updated jsperf test here: https://jsperf.com/omit-vs-rest-vs-reduce/10

image

I'm really confused why my initial tests were not reporting indexOf faster, but now i'm seeing that being faster.

Also realized that the omit I recently checked in was not being exported. Changing the API to take in an array of exclusions first.

### DIFF
--- a/packages/utilities/src/object.ts
+++ b/packages/utilities/src/object.ts
@@ -115,15 +115,15 @@ export function values<T>(obj: any): T[] {
  * https://jsperf.com/omit-vs-rest-vs-reduce/1
  *
  * @param obj - The object to clone
- * @param exclusions - The object of keys to exclude
+ * @param exclusions - The array of keys to exclude
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function omit<TObj extends Record<string, any>>(obj: TObj, exclusions: Record<string, any>): TObj {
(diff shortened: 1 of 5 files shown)


packages/utilities/src/object.ts
```diff
@@ -115,15 +115,15 @@ export function values<T>(obj: any): T[] {
...
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function omit<TObj extends Record<string, any>>(obj: TObj, exclusions: Record<string, any>): TObj {
+export function omit<TObj extends Record<string, any>>(obj: TObj, exclusions: (keyof TObj)[]): TObj {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const result: Record<string, any> = {};
 
-  for (const key of Object.keys(obj)) {
-    if (!exclusions[key]) {
+  for (const key in obj) {
```
xugao, line 125: why this change?

dzearing: see perf numbers.

PR #14316, opened 2020-08-03
Add hc styling to fix Toggle control

### LINKED ISSUE #12298: Fabric toggle control does not support high contrast
### Environment Information

### Describe the issue:

Currently the toggle does not support high contrast.

#### Actual behavior:

The background of the toggle should be cyan when on (or whatever the user setting is for high contrast). 

#### Expected behavior:

Instead, the background of the toggle is black. 

High

tomi-msft:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12298 

#### Description of changes

Toggle component does not support HC styling in Edge Chromium. This fix adds styling that will allow Toggle to adopt HC styling.

**Before**
image

**After**
image

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-next-2020-08-04-09-58-27-a11y-toggle.json
+++ b/change/@fluentui-react-next-2020-08-04-09-58-27-a11y-toggle.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "update Toggle in react-next package",
+  "packageName": "@fluentui/react-next",
+  "email": "ololubek@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-08-04T16:58:27.560Z"
+}
--- a/change/@fluentui-react-next-2020-08-04-09-58-27-a11y-toggle.json
+++ b/change/@fluentui-react-next-2020-08-04-09-58-27-a11y-toggle.json
@@ -1,6 +1,6 @@
 {
   "type": "prerelease",
-  "comment": "update Toggle in react-next package",
+  "comment": "Toggle: fix pill HC style when selected",
   "packageName": "@fluentui/react-next",
   "email": "ololubek@microsoft.com",
   "dependentChangeType": "patch",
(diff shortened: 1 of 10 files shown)


dzearing:
It seems like the focus rectangle is off. It may not have been caused by this PR, but there is a tiny dot in the thumb of the toggle which I think is the focus rect.

packages/office-ui-fabric-react/src/components/Toggle/Toggle.styles.ts
```diff
@@ -129,8 +134,9 @@ export const getStyles = (props: IToggleStyleProps): IToggleStyles => {
                 },
               ],
               [HighContrastSelector]: {
-                backgroundColor: 'WindowText',
+                backgroundColor: 'Highlight',
```
xugao, line 137: could you make the similar fix to `toggle.scss` in `react-next`?
we should add the styles in `.pill &.checked` section and `@include high-contrast-edge-chromium-no-adjust;` is how you achieve the same as `getEdgeChromiumNoHighContrastAdjustSelector`

change/@fluentui-react-next-2020-08-04-09-58-27-a11y-toggle.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "update Toggle in react-next package",
```
xugao, line 3:
```suggestion
  "comment": "Toggle: fix pill HC style when selected",
```

change/office-ui-fabric-react-2020-08-03-14-14-33-a11y-toggle.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Add hc styling to fix toggle",
```
xugao, line 3:
```suggestion
  "comment": "Toggle: fix pill HC style when selected",
```

xugao:
to David's comment, i think the issue is with this line [here](https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/Toggle/Toggle.styles.ts#L179). I think it should be `borderWidth: DEFAULT_THUMB_SIZE / 2` instead.

@tomi-msft - could you confirm my theory and issue a separate PR to fix? thanks!

PR #14319, opened 2020-08-03
Updating ChoiceGroup and ChoiceGroupOption to a function component inside React-Next.

czearing:
#### Pull request checklist

#### Description of changes

1. Converted `ChoiceGroup` and `ChoiceGroupOption` to a function component inside React-Next.
2. Added perf-tests and vr-tests for `ChoiceGroup`.
3. Moved the root class to the acutal root and removed deprecated `appplicationRole` class. 
4. Removed deprecated `onChanged` prop.

#### Focus areas to test
ChoiceGroup, ChoiceGroupOption


apps/vr-tests/src/stories/ChoiceGroupNext.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+    text: 'Unselected',
+  },
+  {
+    key: 'C',
+    text: 'Disabled',
+    disabled: true,
+  },
+];
+
+storiesOf('ChoiceGroup', module)
```
xugao, line 25:
```suggestion
storiesOf('ChoiceGroup Next', module)
```

packages/react-next/src/components/ChoiceGroup/ChoiceGroup.base.tsx
```diff
@@ -1,128 +1,141 @@
...
-    if (prevProps !== this.props) {
-      const newKeyChecked = this._getKeyChecked(this.props);
-      const oldKeyChecked = this._getKeyChecked(prevProps);
-
-      if (newKeyChecked !== oldKeyChecked) {
-        this.setState({
-          keyChecked: newKeyChecked,
-        });
+    const onFocus = (key: string) => {
+      // This extra mess is necessary because React won't pass the `key` prop through to ChoiceGroupOption
```
xugao, line 93: Ah this is indeed a mess. i think we can and should just fix this.

instead of having `key: string` in `IChoiceGroupOptionProps` , we should change it to `optionKey: string`.
that way we can read/set the key not using the preserved prop `key`. 

this way, we can clean up the code here to simply be:

```
const onFocus = (ev: React.FocusEvent<HTMLElement | HTMLInputElement>, option: IChoiceGroupOption) => setKeyFocused(option.optionKey)
```
similarly for `onChange` / `changeCallbacks`

packages/react-next/src/components/ChoiceGroup/ChoiceGroup.base.tsx
```diff
@@ -1,128 +1,141 @@
...
+
+    if (process.env.NODE_ENV !== 'production') {
+      warnMutuallyExclusive('ChoiceGroup', props, {
+        selectedKey: 'defaultSelectedKey',
+      });
+    }
 
-    const labelId = this._id + '-label';
-    const ariaLabelledBy = this.props.ariaLabelledBy || (label ? labelId : this.props['aria-labelledby']);
+    useComponentRef(props, keyChecked, id);
```
xugao, line 134: nit: move this to above, with other hooks calls

packages/react-next/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
```diff
@@ -17,135 +18,124 @@ const LARGE_IMAGE_SIZE = 71;
...
-    return (
-      <span id={props.labelId} className="ms-ChoiceFieldLabel">
-        {props.text}
-      </span>
-    );
-  };
-}
+  const { onRenderField = renderField } = props;
+
+  const onChange = useConstCallback((evt: React.FormEvent<HTMLInputElement>): void => {
```
xugao, line 102: we should remove `useConstCallback` from `onChange` `onBlur` and `onFocus` handlers. I don't see benefits for using them

packages/react-next/src/components/ChoiceGroup/ChoiceGroup.base.tsx
```diff
@@ -1,128 +1,141 @@
...
     return (
-      // eslint-disable-next-line deprecation/deprecation
-      <div className={classNames.applicationRole} {...divProps}>
-        <div
-          className={classNames.root}
-          role="radiogroup"
-          {...(ariaLabelledBy && { 'aria-labelledby': ariaLabelledBy })}
-        >
+      <div className={classNames.root} {...divProps} ref={forwardedRef}>
+        <div role="radiogroup" ref={forwardedRef} {...(ariaLabelledBy && { 'aria-labelledby': ariaLabelledBy })}>
```
xugao, line 138:
```suggestion
        <div role="radiogroup" {...(ariaLabelledBy && { 'aria-labelledby': ariaLabelledBy })}>
```

packages/react-next/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
ecraig12345: This one is a bit tricky--I do think if we're passing callbacks to our own components, it's a good idea to memoize, but it's probably not necessary when passing to native elements. The other tricky part is that `jsx-no-bind` lint rule can't distinguish between the two scenarios, so the options are either to disable it everywhere callbacks are passed to native elements (which could lead to people thinking it's okay to disable in other places where it's potentially more important), or memoize when technically not necessary.

ecraig12345:
@xugao I'm not so sure about renaming `IChoiceGroupOption.key` to `IChoiceGroupOption.optionKey`--that's a big breaking change from a consumer perspective, and it makes ChoiceGroup options inconsistent with the options of other components like Dropdown, ComboBox, Nav, and maybe others.

Instead I think we should look at how those other components get around the related issues, and see if we can use a similar approach here.

packages/react-next/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
```diff
@@ -17,135 +18,124 @@ const LARGE_IMAGE_SIZE = 71;
...
-    }
   };
 
-  private _onRenderField = (props: IChoiceGroupOptionProps): JSX.Element => {
-    const { id, imageSrc, imageAlt = '', selectedImageSrc, iconProps } = props;
+  const renderField = (): JSX.Element => {
+    const { imageAlt = '', selectedImageSrc } = props;
 
-    const imageSize = props.imageSize ? props.imageSize : { width: 32, height: 32 };
+    const currentImageSize = props.imageSize ? props.imageSize : { width: 32, height: 32 };
```
ecraig12345, line 63: You already specified the default above, so I don't think you need a separate `currentImageSize` variable.

packages/react-next/src/components/ChoiceGroup/ChoiceGroup.base.tsx
```diff
@@ -132,21 +104,24 @@ export class ChoiceGroupBase extends React.Component<IChoiceGroupProps, IChoiceG
...
               };
 
               return (
                 <ChoiceGroupOption
-                  key={option.key}
-                  onBlur={this._onBlur}
-                  onFocus={this._onFocus(option.key)}
-                  onChange={this._onChange(option.key)}
+                  key={option.optionKey}
+                  // eslint-disable-next-line react/jsx-no-bind
```
ecraig12345, line 119: This is a case where I think the callbacks should be memoized, since they're being passed to another custom component.

packages/react-next/src/components/ChoiceGroup/ChoiceGroup.base.tsx
```diff
@@ -110,19 +57,44 @@ export class ChoiceGroupBase extends React.Component<IChoiceGroupProps, IChoiceG
...
-    // `applicationRole` class (but the div structure will stay the same by necessity)
     return (
-      // eslint-disable-next-line deprecation/deprecation
-      <div className={classNames.applicationRole} {...divProps}>
-        <div
-          className={classNames.root}
-          role="radiogroup"
-          {...(ariaLabelledBy && { 'aria-labelledby': ariaLabelledBy })}
-        >
+      <div className={classNames.root} {...divProps} ref={forwardedRef}>
```
ecraig12345, line 96: We need to document the move of the `root` class in RELEASE_NOTES.md breaking changes.

Also, since the `applicationRole` class is no longer being used, we should remove it from `IChoiceGroupStyles` (and also document that breaking change).

packages/react-next/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
```diff
@@ -17,135 +18,124 @@ const LARGE_IMAGE_SIZE = 71;
...
 
-    const onRenderLabel = props.onRenderLabel
-      ? composeRenderFunction(props.onRenderLabel, this._onRenderLabel)
-      : this._onRenderLabel;
+    const renderedLabel = props.onRenderLabel
```
ecraig12345, line 65: `renderedLabel` isn't really a good name for this since it's a function, not the label itself.

Also I wonder if we could change `composeRenderFunction` to accept undefined for the custom renderer (and if it's undefined, just return the default renderer) and get rid of this extra conditional everywhere.

packages/react-next/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
```diff
@@ -17,135 +18,124 @@ const LARGE_IMAGE_SIZE = 71;
...
     return (
-      <label htmlFor={id} className={this._classNames.field}>
+      <label htmlFor={id} className={classNames.field}>
         {imageSrc && (
-          <div className={this._classNames.innerField}>
-            <div className={this._classNames.imageWrapper}>
-              <Image src={imageSrc} alt={imageAlt} width={imageSize.width} height={imageSize.height} />
+          <div className={classNames.innerField}>
+            <div className={classNames.imageWrapper}>
+              <Image src={imageSrc} alt={imageAlt} width={currentImageSize.width} height={currentImageSize.height} />
```
ecraig12345, line 76: I think you can just do this:
```suggestion
              <Image src={imageSrc} alt={imageAlt} width={currentImageSize.width} {...imageSize} />
```

packages/react-next/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
```diff
@@ -17,135 +18,124 @@ const LARGE_IMAGE_SIZE = 71;
...
             </div>
-            <div className={this._classNames.selectedImageWrapper}>
-              <Image src={selectedImageSrc} alt={imageAlt} width={imageSize.width} height={imageSize.height} />
+            <div className={classNames.selectedImageWrapper}>
+              <Image
+                src={selectedImageSrc}
+                alt={imageAlt}
+                width={currentImageSize.width}
+                height={currentImageSize.height}
+              />
```
ecraig12345, line 84:
```suggestion
              <Image src={selectedImageSrc} alt={imageAlt} {...imageSize} />
```

apps/perf-test/src/scenarios/ChoiceGroupNext.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
xugao: The main goal of doing this now is to ensure we didn't regress perf by FC conversion (given hook calls are not cheap).
Ideally we have perf tests for ALL components (i believe that has been always something we want to do but never got to). Couple things to also take into consideration: 
1) perf tests today do not blocking build
2) given 1), it doesn't mean it can be slow. if we need to scale the perf test, do that that, without limiting our ability to write new tests

All this being said, I am open to discussion. For this particular case, I agree it doesn't seem to be very useful. Perhaps given the goal here, what we can do instead is to do a perf run before we ship (v7 vs. v8 component)

packages/react-next/src/components/ChoiceGroup/ChoiceGroup.styles.ts
```diff
@@ -7,14 +7,11 @@ const GlobalClassNames = {
...
-  const { className, optionsContainIconOrImage, theme } = props;
+  const { optionsContainIconOrImage, theme } = props;
 
   const classNames = getGlobalClassNames(GlobalClassNames, theme);
 
   return {
-    // TODO (Fabric 8?) - merge className back into `root` and apply root style to
-    // the actual root role=application element
-    applicationRole: className,
     root: [
```
xugao, line 15: `className` should be part of the `root`

packages/react-next/RELEASE_NOTES.md
```diff
@@ -13,6 +13,11 @@ TODO: Diff of OUFR vs date-time Calendar
 - Deprecated `onRenderLabel`.
 - Added `label`/`checkmark` slot props.
 
+### ChoiceGroup
+
+- Moved `root` class to the actual root element by replacing `applicationRole`.
+- Removed `applicationRole` from IChoiceGroupStyles.
+
```
xugao, line 20:
```suggestion
- Removed deprecated `onChanged` prop.

```

PR #14323, opened 2020-08-03
CommandBar does not render tooltipHostProps.content with iconOnly mode as true

### LINKED ISSUE #14310: CommandBar does not render tooltipHostProps.content with iconOnly mode as true
CommandBar does not render tooltipHostProps.content with iconOnly=true, when text/name is not available in props

### Environment Information

- **Package version(s)**: 7.123.3

https://codepen.io/pranithasurya/pen/MWKLgqJ

#### Actual behavior:

When you hover on the share icon, the tooltip content is not shown even when the iconOnly mode is true. The reason is because the text/name prop is not available. 

#### Expected behavior:

The tooltip content rendering should only depend on iconOnly prop and not on text/name prop.

Requested priority: Normal

Products/sites affected: All fluentui dependent websites

pranitha-surya:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14310

#### Description of changes

CommandBar does not render tooltipHostProps.content with iconOnly=true, when text/name is not available in ICommandBarItemProps

#### Focus areas to test

(optional)


packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.base.tsx
```diff
@@ -169,9 +169,10 @@ export class CommandBarBase extends React.Component<ICommandBarProps, {}> implem
       onClick: this._onButtonClick(item),
     };
 
-    if (item.iconOnly && itemText !== undefined) {
+    if (item.iconOnly && (itemText !== undefined || (item.tooltipHostProps && item.tooltipHostProps.content))) {
```
brandonthomas, line 172: I feel like we could simplify this just to check for tooltipHostProps instead of tooltipHostProps and tooltipHostProps.content. Thinking about scenarios where someone passes in only onRenderContent.

pranitha-surya: True, but it could be possible to not send anything too and may end up showing a blank content. Instead we should check on onRenderContent as well. What do you think?

brandonthomas: It would be weird for someone to pass tooltipHostProps and not want to display it no? I just don't think we need to box it in like that. It doesn't seem an unsafe assumption to make that if someone passes this optional prop in that they want a tooltip.

packages/office-ui-fabric-react/src/components/CommandBar/CommandBar.base.tsx
```diff
@@ -169,9 +169,10 @@ export class CommandBarBase extends React.Component<ICommandBarProps, {}> implem
...
+      const content = itemText ? itemText : item.tooltipHostProps!.content;
```
brandonthomas, line 173: We shouldn't actually need this since the spread below will take care of this logic for us.

brandonthomas:
ping @khmakoto can we get a review?

change/office-ui-fabric-react-2020-08-03-14-39-05-master.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "show tooltipHost.content when iconMode = true and even when text/name is provided in commandBarItemProps",
```
khmakoto, line 3:
```suggestion
  "comment": "CommandBar: Show tooltipHost.content when iconMode = true even when text/name is provided in commandBarItemProps",
```

micahgodbolt:
good catch, lgtm :shipit:

PR #14327, opened 2020-08-04
Add ways to try code with React.StrictMode

### LINKED ISSUE #14325: Enable testing strict mode in demo app/website
We need a way to test components with strict mode in the demo apps and website.

1. Remove UNSAFE_ method in `packages/example-app-base/src/components/LoadingComponent/LoadingComponent.tsx` (there's also an UNSAFE_ in `apps/fabric-website/src/components/Site/Site.tsx` but it would be harder to remove without undesired side effects, and it's easy to work around for now by just enabling strict mode immediately below that level)

2. Add support for a query param `strict=all` to enable strict mode for the whole page.

3. Add support for a query param `strict=ex` to enable strict mode just for examples.

4. Add a way to turn on strict mode for storybook examples

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14325

#### Description of changes

Add ways to try React.Strict mode in different inner loop and demo experiences:
- In the website and legacy demo app, add a query param `?strict=all` to run the whole app in strict mode, or `?strict=ex` to run examples in strict mode.
- In storybook, add a decorator `withStrictMode` for wrapping a story in strict mode (and update the storybook setup in some of the new packages to use the decorator)

While looking for storybook preview.js files to update, I noticed that todo-app had storybook files for some reason but no way to use them (and a webpack config that didn't make sense), so I deleted those.


packages/react-next/.storybook/preview.js
```diff
@@ -4,13 +4,14 @@ import { configure, addParameters, addDecorator } from '@storybook/react';
...
 import { withKnobs } from '@storybook/addon-knobs';
-import { withThemeProvider } from '@fluentui/storybook';
+import { withThemeProvider, withStrictMode } from '@fluentui/storybook';
 import { withKeytipLayer } from './decorators';
 
 addDecorator(withA11y());
 addDecorator(withPerformance);
 addDecorator(withKnobs({ escapeHTML: false }));
 addDecorator(withThemeProvider);
+addDecorator(withStrictMode);
```
xugao, line 14: actually, we should default to be enabled for `react-next` package

packages/react-button/.storybook/preview.js
```diff
@@ -5,13 +5,14 @@ import { withA11y } from '@storybook/addon-a11y';
...
 import { withPerformance } from 'storybook-addon-performance';
-import { withThemeProvider } from '@fluentui/storybook';
+import { withThemeProvider, withStrictMode } from '@fluentui/storybook';
 
 addDecorator(withPerformance);
 addDecorator(withInfo());
 addDecorator(withA11y());
 addDecorator(withKnobs({ escapeHTML: false }));
 addDecorator(withThemeProvider);
+addDecorator(withStrictMode);
```
xugao, line 15: same here, should default as enabled

packages/react-next/.storybook/preview.js
ecraig12345: I think it should be opt-in to avoid confusing people. One of the things strict mode does is [run certain things twice](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects), and if someone isn't expecting that, it could confuse them while trying to debug something.

ecraig12345:
> Do we want to surface the StrictMode errors more prominently, such as with a `window.alert()`? Outlook Web has code to wrap the `console` methods to report any of these warnings to our telemetry in dev builds.

Interesting idea, though I'd probably prefer an overlay/banner (not sure how often the errors might show up, and an alert would be super disruptive). The only existing related solution I can find is [`@storybook/addon-console`](https://github.com/storybookjs/storybook-addon-console), but the docs for how to set it up are confusing and the UX for showing the logs is not great. So I think in the interest of time I'm going to stick with just console logs for now, and maybe we can do something better in the future.

packages/react-next/.storybook/preview.js
xugao: I don't feel too strongly about this. But i still think it's overall better to turn this on by default. That way dev can spot issues with strict mode on as early as possible during development phase with less chance of missing. After all, if we have code that doesn't work well with strict mode, we need to get those fixed.

ecraig12345: For obvious new issues, I think a better way to prevent them from being introduced is with lint rules. [`eslint-plugin-react`](https://github.com/yannickcr/eslint-plugin-react) already has rules for preventing UNSAFE_ and findDOMNode, so we just need to enable them. (We already have a rule about no string refs enabled.) 

For the non-obvious issues with side effects, which the double rendering is meant to catch, I'm still concerned that it would cause confusion when debugging more often than it would be helpful.

xugao: ok, i'm fine starting with this. can always update when we find fit

PR #14329, opened 2020-08-04
Fix: Top's component debug frame is above iframe

vyhnalekl:
#### Pull request checklist

#### Description of changes

Top's component Debug Frame was outside of iframe, This PR is fixing the issue by moving the Debug frame to the bottom of the components that are at top of the iframe.


packages/fluentui/react-builder/src/components/DebugFrame.tsx
```diff
@@ -71,14 +72,32 @@ export const DebugFrame: React.FunctionComponent<DebugFrameProps> = ({
...
 
     const el = target.querySelectorAll(selector);
-
-    animationFrameId.current =
-      el.length === 1
-        ? requestAnimationFrame(() => setFramePosition(frameRef.current, el[0]))
-        : requestAnimationFrame(() => hideFrame(frameRef.current));
+    if (el.length === 1) {
+      const rect = el[0].getBoundingClientRect();
+      rect.top < 20 ? !isTopElement && setIsTopElement(true) : isTopElement && setIsTopElement(false);
```
mnajdova, line 77: Can we add const for the 20?

packages/fluentui/react-builder/src/components/DebugFrame.tsx
```diff
@@ -71,14 +72,32 @@ export const DebugFrame: React.FunctionComponent<DebugFrameProps> = ({
...
+      animationFrameId.current = requestAnimationFrame(() => setFramePosition(frameRef.current, el[0]));
+    } else {
+      animationFrameId.current = requestAnimationFrame(() => hideFrame(frameRef.current));
+    }
 
     return () => cancelAnimationFrame(animationFrameId.current);
-  }, [target, selector, setFramePosition]);
+  }, [target, selector, setFramePosition, isTopElement]);
```
mnajdova, line 84: Should we add `animationFrameId` too?

PR #14342, opened 2020-08-04
Remove all usage of string refs

### LINKED ISSUE #14290: Remove string ref usage from list components
List, GroupedList, and GroupedListSection use string refs (`this.refs`) to store refs of rendered items. For strict mode compliance, they need to be converted to use function refs or ref objects.

MLoughry:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14290 

#### Description of changes

Towards the goal of making all of OUFR compliant with React StrictMode, this change replaces all usage of string-based `refs` in OUFR with compliant alternatives.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/List/List.tsx
+++ b/packages/office-ui-fabric-react/src/components/List/List.tsx
@@ -183,7 +183,7 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
     this._pageCache = {};
   }
 
-  public get PageRefs(): Readonly<Record<string, unknown>> {
+  public get pageRefs(): Readonly<Record<string, unknown>> {
     return this._pageRefs;
   }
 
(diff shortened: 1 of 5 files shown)


packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -186,6 +183,10 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
     this._pageCache = {};
   }
 
+  public get PageRefs(): Readonly<Record<string, unknown>> {
```
ecraig12345, line 186: What's the reasoning for having this pascal case?

MLoughry: Nothing specific; was just think it should be as a public property. I'll change it.

MLoughry:
@ecraig12345 @xugao Could either of you approve the Screener additions?

ecraig12345:
@MLoughry Not seeing anything to approve on screener?

MLoughry:
@ecraig12345 @xugao That would be because I made this comment on the wrong PR. 😳

PR #14344, opened 2020-08-04
React conformance: Updating doctrine and modifying merge function to use lodash.

### LINKED ISSUE #14205: react-conformance: dep fixes and publishing
Various small fixes needed for react-conformance:

1. Publish the react-conformance package (remove `"private": true` from package.json)

2. Per https://github.com/microsoft/fluentui/pull/13939#discussion_r460300974, remove the custom merge function and replace it with https://lodash.com/docs/4.17.15#mergeWith with custom array merging (see example in link)

3. Per https://github.com/microsoft/fluentui/pull/13939#discussion_r460296403, try either replacing `doctrine` with [`comment-parser`](https://www.npmjs.com/package/comment-parser) (only in react-conformance, not scripts, to be safe) or at least upgrade `doctrine` to the latest version (both in react-conformance and scripts)

sopranopillow:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14205

#### Description of changes

This PR updates doctrine in `@fluentui/react-conformance` and `@uifabric/build` and modifies `@fluentui/react-conformance`'s merge to use lodash's `mergeWith()`.


packages/react-conformance/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@fluentui/react-conformance",
   "version": "0.1.0",
   "description": "Customizable conformance testing utility for Fluent UI React components.",
-  "private": true,
+  "private": false,
```
ecraig12345, line 5: Just remove it instead
```suggestion
```

packages/react-conformance/src/utils/merge.tsx
```diff
@@ -1,31 +1,26 @@
...
-  const merged = { ...objs[0] };
-  objs = objs.slice(1);
+  const [merged, ...rest] = objs;
 
-  objs.forEach(option => {
-    _merge(merged, option);
-  });
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const customMergeFunction = (dest: any, src: any): any => {
+    if (isObject(dest) && isObject(dest)) {
```
ecraig12345, line 15: I feel like this could be further simplified--the only case where we need special handling is if one or more of `src` and `dest` is an array. Otherwise we can just use plain `_.merge`.

PR #14345, opened 2020-08-04
feat(web-components): update package name to align with @fluentui and update component namespace

### LINKED ISSUE #14343: Proposal: Update web-components package to use the @fluentui/ namespace
Currently the web components package is under the global `@microsoft/` namespace with a package name of `@microsoft/fast-components-msft`. Aside from the obvious of not sharing the existing `@fluentui/` namespace, there are also publishing considerations that updating the package name will resolve. This package was previously published using lerna, which would rev the package version, create a changelog, fire off tags and force push a publish commit to master. I don't think that is a wise publishing method to inject into the workflow here. Additionally, there are permissions issues with that approach.

It seems that the fastest and safest way to address this issue would be to update the namespace and include the Web Components package as part of the publish process.

Proposed package name: `@fluentui/web-components`. Ideally, I like the concept of simply calling these "components" rather than specifying web components, but I understand with existing conventions that could cause confusion. :)

chrisdholt:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14343

#### Description of changes

This PR updates the package name to align with the FluentUI namespace.


packages/web-components/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "@microsoft/fast-components-msft",
+  "name": "@fluentui/web-components",
   "description": "A library of Fluent Web Components",
   "sideEffects": false,
-  "version": "1.6.0",
+  "version": "0.0.1",
```
EisenbergEffect, line 5: We should start this as `0.1.0` or we'll have some issues representing breaking vs. non-breaking changes.

chrisdholt: @dzearing I'm making an assumption here on how beachball works, but perhaps you or @ecraig12345 can confirm for me so I can resolve this :). 

I reset this to 0.0.1 and targeted my change request as a minor version change. I'm assuming that should update the package.json here to be 0.1.0 - is that correct?

dzearing: yup!

EisenbergEffect:
Some follow-up notes for reviewers based on a conversation that @chrisdholt and I had.

We're changing the element names to align with fluent branding and also to prevent a potential breaking change to default names in the future. Once we implement the new provider design outlined [here](https://github.com/microsoft/fast/issues/3361) we'll be able to give control over the prefix to end users. So, if someone was using `@microsoft/fast-components` and wanted to switch to `@fluentui/web-components` they could configure a prefix and not have to change any of their HTML. We also believe we can enable this through a single script file build, which might look like this when used:

```html
<script type="module" src="..." data-prefix="fast"></script>
<script type="module" src="..." data-prefix="fluent"></script>
<script type="module" src="..." data-prefix="material"></script>
<script type="module" src="..." data-prefix="ui"></script>
...
```

The single file script build would set up a default provider configuration, checking its own script tag to see if a prefix was configured for the provider.

Additionally, based on some conversations with Graph, we'll be looking into whether the new provider model could support lazily creating a provider if one is not defined. In this case, the prefix would default to `fluent` and would enable using the components without a wrapping `<fluent-design-system-provider>` element. This last idea is "theoretical" at present. We'll be investigating as we build out the new provider system.

packages/web-components/package.json
dzearing, line 5: yup, the minor will update this to 0.1.0.

dzearing:
@layershifter @miroslavstastny @levithomason 

Wanted to give you a heads up here. The build errors in this PR are all in `react-northstar` and seem to have confusing TypeScript typing problems in the `design` prop.

I believe something here is causing react-northstar to build with a newer typescript build, but I'm not sure. I know that if I clone a fresh enlistment and open VSCode (defaulting to TS 3.9.x) to `Table.tsx`, I am seeing the red squiggles on the `PropTypes` definition, and lowering the TypeScript version to 3.7.x makes it go away.

Still parsing why TypeScript version has changed here; we thought it was an incorrectly defined `nohoist` in the root package.json, but updating that seems to not have affected this.

```
@fluentui/react-northstar:   Type '{ activeIndex: PropTypes.Requireable<AnimationIterationCountProperty>; iconOnly: PropTypes.Requireable<boolean>; thumbnails: PropTypes.Requireable<...>; ... 13 more ...; accessibility: PropTypes.Requireable<...>; }' is not assignable to type 'WeakValidationMap<CarouselNavigationProps>'.
@fluentui/react-northstar:     Types of property 'design' are incompatible.
@fluentui/react-northstar:       Type 'Requireable<InferProps<{ position: Requireable<string>; display: Requireable<string>; top: Requireable<string>; right: Requireable<string>; bottom: Requireable<...>; ... 16 more ...; maxHeight: Requireable<...>; }>>' is not assignable to type 'Validator<ComponentDesignProp>'.
@fluentui/react-northstar: src/components/Carousel/CarouselItem.tsx(129,1): error TS2322: Type '{ active: PropTypes.Requireable<boolean>; navigation: PropTypes.Requireable<boolean>; itemPositionText: PropTypes.Requireable<string>; styles: PropTypes.Requireable<...>; ... 7 more ...; accessibility: PropTypes.Requireable<...>; }' is not assignable to type 'WeakValidationMap<CarouselItemProps> & { as: Requireable<string | ((props: any, context?: any) => any) | (new (props: any, context?: any) => any)>; }'.
@fluentui/react-northstar:   Type '{ active: PropTypes.Requireable<boolean>; navigation: PropTypes.Requireable<boolean>; itemPositionText: PropTypes.Requireable<string>; styles: PropTypes.Requireable<...>; ... 7 more ...; accessibility: PropTypes.Requireable<...>; }' is not assignable to type 'WeakValidationMap<CarouselItemProps>'.
@fluentui/react-northstar:     Types of property 'design' are incompatible.
```

PR #14347, opened 2020-08-04
Improve query params for enabling strict/dev mode

ecraig12345:
Quick follow-up for #14327: right after merging that, I realized it doesn't make sense to require people to type *both* `?strict=all&isDev=1`, and the inconsistent naming was annoying. So...
- Rename `isDev` to `dev`
- Make `strict` imply `dev` (`React.StrictMode` only works in React dev builds)

### DIFF
--- a/apps/fabric-website/index.html
+++ b/apps/fabric-website/index.html
@@ -156,7 +156,7 @@
     <script type="text/javascript">
       // @ts-check
       var isLocal = location.hostname === 'localhost' || location.hostname.indexOf('ngrok.io') > -1;
-      var isDev = isLocal || getParameterByName('isDev') === '1';
+      var isDev = isLocal || getParameterByName('dev') === '1' || !!getParameterByName('strict');
       var isProduction = location.hostname === 'developer.microsoft.com' || getParameterByName('isProduction') === '1';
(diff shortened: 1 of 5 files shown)


apps/fabric-website/index.html
```diff
@@ -156,7 +156,7 @@
     <script type="text/javascript">
       // @ts-check
       var isLocal = location.hostname === 'localhost' || location.hostname.indexOf('ngrok.io') > -1;
-      var isDev = isLocal || getParameterByName('isDev') === '1';
+      var isDev = isLocal || getParameterByName('dev') === '1' || !!getParameterByName('strict');
```
xugao, line 159: where is `var isDev` being used?

xugao: also, looks odd to have both `isDev` and `isProduction`.

ecraig12345: Oops, there's some logic I forgot to copy from homepage.htm.

ecraig12345: `isDev` determines whether to use dev mode (non-minified scripts and dev mode react). `isProduction` determines whether to use the production manifest or dogfood/staging manifest.

PR #14348, opened 2020-08-04
Remove usage of `findDOMNode` from dragDropHelper

### LINKED ISSUE #14321: DragDropHelper: strict mode compliance (findDOMNode)
DragDropHelper uses [`ReactDOM.findDOMNode`](https://reactjs.org/docs/react-dom.html#finddomnode) which is now discouraged for various reasons and [deprecated in strict mode](https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage).

It will continue to work as long as the parent and child for every drag/drop operation are class components (despite causing warnings in strict mode), but once something is converted to a function component, per the docs it sounds like the `findDOMNode` call will no longer work. So long-term we'll need to find another way to handle this scenario, even if we choose not to fix the warning right now.

MLoughry:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14321

#### Description of changes

There was no reason at all to use `findDOMNode` in the first place. Fixing up one interface signature obviates the need for it.

#### Focus areas to test

(optional)

### DIFF
--- a/change/office-ui-fabric-react-2020-08-04-22-49-31-v8-drag-drop-helper.json
+++ b/change/office-ui-fabric-react-2020-08-04-22-49-31-v8-drag-drop-helper.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Remove findDOMNode usage from DragDropHelper",
+  "packageName": "office-ui-fabric-react",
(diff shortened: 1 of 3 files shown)


MLoughry:
Verified with the DetailsList drag story

PR #14350, opened 2020-08-05
Fix nohoist settings

ecraig12345:
People were intermittently hitting typescript build issues in `react-northstar` which @dzearing and I figured out were due to the additional TS version introduced by web-components getting installed at the repo root. That was supposed to be prevented by `nohoist` settings, but it wasn't configured quite correctly (probably thanks to [unclear documentation](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/)). So this PR fixes the settings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,15 +79,20 @@
       "packages/*",
       "scripts",
       "packages/fluentui/*"
+    ],
+    "nohoist": [
+      "packages/web-components/typescript",
+      "packages/web-components/typescript/**",
+      "packages/web-components/ts-loader",
+      "packages/web-components/ts-loader/**",
+      "packages/web-components/ts-node",
+      "packages/web-components/ts-node/**",
+      "packages/web-components/mocha",
+      "packages/web-components/mocha/**",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -79,15 +79,20 @@
...
+      "packages/web-components/typescript/**",
+      "packages/web-components/ts-loader",
+      "packages/web-components/ts-loader/**",
+      "packages/web-components/ts-node",
+      "packages/web-components/ts-node/**",
+      "packages/web-components/mocha",
+      "packages/web-components/mocha/**",
+      "packages/web-components/webpack",
+      "packages/web-components/webpack/**"
     ]
```
chrisdholt, line 94: Should this result in yarn.lock changes?

chrisdholt: just ran this locally - I don't see any. Should be good!

ecraig12345: Apparently not. It seems yarn.lock documents the full tree of packages present, but not which ones are installed where (which is probably how this issue was intermittent).

PR #14360, opened 2020-08-05
Add drag drop to SelectedItemsList (experiment) (and UnifiedPicker as well)

elisabethcvs:
#### Pull request checklist

#### Description of changes

Add drag drop functionality to the SelectedItemsList:
- It is not on by default.
- Added a SelectedPeopleList example that has single item drag drop enabled. 
- Multi item drag drop is also now available. A MarqueeSelection is required.

UnifiedPicker control's SelectedItemsList also now has drag drop enabled:
- Multiple item drag drop is always enabled, as multi select was already the selection system.
- The UnifiedPicker example will show this.

Note: This is within the controls only. Between controls will be done in a future separate work item.


packages/experiments/src/components/UnifiedPicker/hooks/useSelectedItems.ts
```diff
@@ -58,6 +59,28 @@ export const useSelectedItems = <T extends {}>(
     }
   };
 
+  const dropItemsAt = (insertIndex: number, draggedItemsIndices: number[]): void => {
+    const currentItems: T[] = [...items];
+    const updatedItems: T[] = [];
+    console.log(draggedItemsIndices);
```
nebhatna, line 65: remove

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
```diff
@@ -1,6 +1,14 @@
...
 import { IPickerItemProps, ISuggestionModel, ValidationState } from 'office-ui-fabric-react/lib/Pickers';
 import { IRefObject } from 'office-ui-fabric-react/lib/Utilities';
+import {
+  IDragDropEvents,
+  IDragDropContext,
+  IDragDropHelper,
+  IDragDropOptions,
+} from 'office-ui-fabric-react/lib/utilities/dragdrop/index';
+
+export { IDragDropContext, IDragDropEvents, IDragDropHelper, IDragDropOptions };
```
nebhatna, line 11: I don't think we need to do this right?

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/examples/SelectedPeopleList.Basic.DragDrop.Example.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const theme = getTheme();
+  const dragEnterClass = mergeStyles({
+    backgroundColor: theme.palette.neutralLight,
+  });
+
+  const _insertBeforeItem = (item: T): void => {
+    const draggedItems = selection.isIndexSelected(draggedIndex) ? (getSelectedItems() as T[]) : [draggedItem!];
+    const insertIndex = selectedItems.indexOf(item);
+    const items = selectedItems.filter(itm => draggedItems.indexOf(itm) === -1);
```
nebhatna, line 35: nit: item :)

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -27,6 +30,16 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
   const { suggestions, selectedSuggestionIndex, isSuggestionsVisible } = props.floatingSuggestionProps;
+  const [draggedItem, setDraggedItem] = React.useState<T>();
+  const dragDropHelper = new DragDropHelper({
+    selection: selection,
+  });
+
+  const theme = getTheme();
+  const dragEnterClass = mergeStyles({
+    backgroundColor: theme.palette.neutralLight,
+  });
```
nebhatna, line 41: It is good to have a default but i think we should also give the option to the caller ( owa ) to override this if needed. I would make this dependent on some sort of props - basically something like using what comes as part of props from the caller or return this in _OnDragEnter.

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -65,6 +79,44 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
+    return dragEnterClass;
+  };
+
+  const _onDrop = (item?: any, event?: DragEvent): void => {
+    if (draggedItem) {
+      _insertBeforeItem(item);
+    }
+  };
+
+  const _onDragStart = (item?: any, itemIndex?: number, selectedItems?: any[], event?: MouseEvent): void => {
```
nebhatna, line 102: I think for all these functions, we should be having an option to override/achieve behavior of the caller. This way if the caller wants to achieve drag/drop in a different manner than the default, we give them that option. Right now, we aren't giving that option. I would add a check for the props containing any callback on these drag/drop properties, and if they dont exist we can use the default behavior. Some examples where we are already doing this is - getItemsCopyText, _onInputFocus , _onCopy etc.

elisabethcvs: That makes sense, I'll add that

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/examples/SelectedPeopleList.Basic.DragDrop.Example.tsx
elisabethcvs: given that the input name is item, I thought having this be slightly different would be less confusing

nebhatna: got it.. i thought the spelling got missed .. :) maybe some other var name would also work

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
elisabethcvs: So all of these functions are a part of the IDragDropEvents. I'm thinking just make that a prop and use it if there, otherwise use all these functions. I think this makes more sense, because the functionality of all of these functions is intertwined, but am open to other opinions

nebhatna: Yea i think that makes sense, as they are dependent on each other.

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/examples/SelectedPeopleList.Basic.DragDrop.Example.tsx
elisabethcvs: I changed it to be less confusing

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
```diff
@@ -91,4 +110,15 @@ export interface ISelectedItemsListProps<T> extends React.ClassAttributes<any> {
    * A callback on whether this item can be removed
    */
   canRemoveItem?: (item: T) => boolean;
+
+  /** Map of callback functions related to row drag and drop functionality. */
```
nebhatna, line 114: is the comment correct?

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
```diff
@@ -91,4 +110,15 @@ export interface ISelectedItemsListProps<T> extends React.ClassAttributes<any> {
...
+  dragDropEvents?: IDragDropEvents;
+
+  /**
+   * Helper for the drag and drop
+   */
+  dragDropHelper?: IDragDropHelper;
+
+  /** Event names and corresponding callbacks that will be registered to rendered selected items. */
+  selectedItemEventMap?: { eventName: string; callback: (context: IDragDropContext, event?: any) => void }[];
```
nebhatna, line 123: i dont see this being used anywhere? what is the intent of this?

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
```diff
@@ -91,4 +110,15 @@ export interface ISelectedItemsListProps<T> extends React.ClassAttributes<any> {
    * A callback on whether this item can be removed
    */
   canRemoveItem?: (item: T) => boolean;
+
+  /** Map of callback functions related to row drag and drop functionality. */
+  dragDropEvents?: IDragDropEvents;
+
+  /**
```
nebhatna, line 117: I see the dragDropHelper and dragDropEvents exist in both ISelectedItemsProps and ISelectedItemsListProps.. is this to achieve single /multi drag and drop?

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/Items/SelectedPersona.tsx
```diff
@@ -1,11 +1,21 @@
 import * as React from 'react';
-import { getId, styled, classNamesFunction, IStyleFunctionOrObject, css } from 'office-ui-fabric-react/lib/Utilities';
+import * as ReactDOM from 'react-dom';
+import {
+  getId,
+  styled,
+  classNamesFunction,
+  IStyleFunctionOrObject,
```
nebhatna, line 7: IStyleFunctionOrObject doesnt seem to be used in the file. Could you remove it please? Though it existed earlier :)

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/Items/SelectedPersona.tsx
```diff
@@ -45,8 +57,76 @@ const SelectedPersonaInner = React.memo(
...
+    React.useEffect(() => {
+      setDragDropSubscription(createDragDropSubscription());
+      return () => {
+        dragDropSubscription?.dispose();
+        setDragDropSubscription(undefined);
+      };
+    }, [dragDropHelper, dragDropSubscription, createDragDropSubscription]);
+
+    React.useEffect(() => {
+      setIsDraggable(dragDropEvents ? !!(dragDropEvents.canDrag && dragDropEvents.canDrag) : undefined);
```
nebhatna, line 124: for dragDropEvents.canDrag && dragDropEvents.canDrag.. did you mean to just use one

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -65,6 +74,51 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
     onInputChange,
   } = props;
 
+  // All of the drag drop functions are the default behavior. Users can override that by setting the dragDropEvents prop
+
```
nebhatna, line 78: nit: remove extra newline

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -171,6 +226,8 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
       selectedItems: selectedItems,
       focusedItemIndices: focusedItemIndices,
       onItemsRemoved: _onRemoveSelectedItems,
+      dragDropHelper: dragDropHelper,
+      dragDropEvents: props.dragDropEvents ? props.dragDropEvents : defaultDragDropEvents,
```
nebhatna, line 230: i think to begin with this seems okay.. if later on a case arises where we want to only override a specific functionality.. we can think about making this more granular

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
elisabethcvs: Not exactly. Those are needed by both the list and the individual item. Basically, the item registers to say hey I'm doing a drag drop thing and the list is like excellent and keeps track of the overall state.

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
elisabethcvs: It's not actually a map. Fixed in the latest update.

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
elisabethcvs: It isn't. Copy pasted from an example and didn't use this. Removed in the most recent update.

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/Items/SelectedPersona.tsx
elisabethcvs: It is used, for styles in the ISelectedPersonaProps. styled is used for the SelectedPersona on line 219

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/Items/SelectedPersona.tsx
elisabethcvs: Mean to do canDrag and canDrop. Fixed in the latest iteration.

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
```diff
@@ -1,16 +1,19 @@
...
   const [items, setItems] = React.useState(props.selectedItems || props.defaultSelectedItems || []);
+  const { dragDropEvents, dragDropHelper } = props;
+
   const renderedItems = React.useMemo(() => items, [items]);
 
   React.useEffect(() => {
-    setItems(props.selectedItems || []);
-  }, [props.selectedItems]);
+    setItems(props.selectedItems || props.defaultSelectedItems || []);
+  }, [props.selectedItems, props.defaultSelectedItems]);
```
nebhatna, line 16: why did we need to do this? i think selecteditems should just be dependent on defaultSelectedItems only on the first render.

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
nebhatna: makes sense, thanks

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/Items/SelectedPersona.tsx
```diff
@@ -45,8 +57,72 @@ const SelectedPersonaInner = React.memo(
...
+      dragDropHelper,
+      dragDropEvents,
+      eventsToRegister,
     } = props;
     const itemId = getId();
+    const [root, setRoot] = React.useState<HTMLElement | undefined>();
+    const [dragDropSubscription, setDragDropSubscription] = React.useState<IDisposable>();
+    const [isDropping, setIsDropping] = React.useState(false);
+    const [droppingClassNames, setDroppingClassNames] = React.useState('');
+    const [droppingClassName, setDroppingClassName] = React.useState('');
```
nebhatna, line 69: does it make sense to put DEFAULT_DROPPING_CSS_CLASS here as default value .. basically React.useState(DEFAULT_DROPPING_CSS_CLASS); ?

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
elisabethcvs: I did this because a test was failing (SelectedPeopleList.test.tsx, renders personas that are passed in as default). There were no items there.

I googled this. Turns out useEffects do run on initial render.  So looks like the items are set to default then immediately reset to whatever the selectedItems props is, which was 0. So this way they stay set to the default props. This may not be the best way to fix this. There is an alternate way to basically disable the effect for the first time. Maybe I should do that instead? But I figured that the defaultSelectedItems can change, so maybe we should be doing it this way? It's not really what was intended, but it is a possibility.

nebhatna: I see, yea the reason we dont want to do this is - if someone changes the defaultSelectedItems, it would update the items. For the initial render, we do want the defaultSelectedItems to be set, but we dont have the items to be updated/dependent on a change in defaultSelectedItems. Hence we would not want it to be a part of useEffect.
Question-
- Has this unit test been failing? Or is it due to the changes we are doing in this PR?
If the unit test has been failing for sometime, i think it would make sense to fix the unit test. If not then we need to look into alternate ways of doing this.
I would want the selectedItems to be updated only when props.selectedItems is updated.

elisabethcvs: Ok, updated it so that the effect doesn't run the first time.

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -16,6 +17,8 @@ import { useFloatingSuggestionItems } from './hooks/useFloatingSuggestionItems';
 import { useSelectedItems } from './hooks/useSelectedItems';
 import { IFloatingSuggestionItemProps } from '../../FloatingSuggestionsComposite';
 import { copyToClipboard } from '../SelectedItemsList/index';
+import { getTheme, mergeStyles } from 'office-ui-fabric-react/lib/Styling';
+import { IDragDropEvents } from 'office-ui-fabric-react';
```
nebhatna, line 21: nit: could you combine this with other imports from office-ui-fabric-react on line 6

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.types.ts
```diff
@@ -3,6 +3,7 @@ import { IRefObject } from '@uifabric/utilities';
 import { IBaseFloatingSuggestionsProps } from '../FloatingSuggestionsComposite/FloatingSuggestions.types';
 import { ISelectedItemsListProps } from '../SelectedItemsList/SelectedItemsList.types';
 import { IFocusZoneProps, IInputProps, Autofill } from 'office-ui-fabric-react';
+import { IDragDropEvents } from 'office-ui-fabric-react';
```
nebhatna, line 6: nit: same comment here, combine with other imports from office-ui-fabric-react

PR #14362, opened 2020-08-05
chore: add typescript as explicit dev dependency to react-northstar

chrisdholt:
#### Pull request checklist

#### Description of changes
This PR adds typescript as an explicit dev dependency for react-northstar. For some reason, it was building with TS version of ^3.9.x. My working assumption is that it defaulted to an installed version as one was not explicitly defined in the package.json.

#### Focus areas to test
The build passes 😄

### DIFF
--- a/packages/fluentui/ability-attributes/package.json
+++ b/packages/fluentui/ability-attributes/package.json
@@ -8,7 +8,8 @@
     "ability-attributes": "^0.0.8"
   },
   "devDependencies": {
-    "ability-attributes-generator": "^0.0.8"
+    "ability-attributes-generator": "^0.0.8",
+    "typescript": "3.7.2"
   },
   "files": [
     "dist"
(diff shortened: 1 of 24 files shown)


levithomason:
This looks good by me.  @layershifter ?

layershifter:
I suggest to do the same change for other packages under `fluentui` folder, other than that LGTM 👍

chrisdholt:
> I suggest to do the same change for other packages under `fluentui` folder, other than that LGTM 👍

@layershifter my only concern with adding everywhere is that we don't quite know why this is happening. I definitely get the goal of uniformity across the fluentui folder, but I feel like it may be safer to fix the immediate problem and not make the change everywhere. In the case that this break pops up elsewhere, I think that would be the immediate next step to see if it fixes the problem there. Thoughts on that? I'm certainly open to making the update elsewhere, but not exactly knowing *why* this fixed the problem gives me pause on adding the explicit dependency everywhere else.

PR #14363, opened 2020-08-05
Fix: Dismiss callout when blurring to a new window

### LINKED ISSUE #13036: The Callout onDismiss handler is not triggered when clicking a different iframe
### Environment Information

- **Package version(s)**: `"office-ui-fabric-react": ""^7.93.0"`, but it was reproducible on the latest version as well.

- **Browser and OS versions**: `Chrome/81.0.4044.129`

**Steps:**
1. Open codepen: https://codepen.io/rochaudh/pen/yLYvjZv
2. Press the show callout button. A callout should appear.
3. Click the square box (iframe). Notice that the callout does not disappear and onDismiss handler is not called. 

image

#### Actual behavior:

When clicking inside an iframe (but outside of the callout), the callout is not dismissed.

#### Expected behavior:

When clicking outside the callout, the onDismiss handler should be triggered, and the callout should disappear. 

Are you willing to submit a PR to fix? `No`

Requested priority:  `Low`

Products/sites affected: `Dynamics 365`

jodom:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13036 

#### Description of changes

Callouts remain open when focus shifts outside the current window in which they're anchored. This PR uses a blur event handler to dismiss open callouts when the target window is blurred

#### Focus areas to test

Callout dismissal


ecraig12345:
Thanks for doing this! You'll also need to create a change file by running `yarn change`.

jodom:
Trying to figure out a beachball fail when I run yarn change. Here is the error 
image

jodom:
Fixed! I simply had to point remote origin url to microsoft/fluentui instead of my fork

jodom:
@ecraig12345 , @khmakoto , please review PR if you can

jodom:
I'm having trouble running `yarn start` in packages/react-next. Is there a known issue building the package? Here is a partial import error I'm seeing in console
image

khmakoto:
@jodom can you try doing `yarn buildto common-styles` before doing `yarn start`? This is a known issue and should be resolved by building `@fluentui/common-styles` once.

jodom:
`yarn buildto common-styles` completed successfully, but still seeing the import errors. Also tried buildto office-ui-fabric react although I'm not sure that would have helped in any way

jodom:
@khmakoto I've replicated the dismiss logic to react-next, but haven't validated these locally. Is there any way of testing these in CodeSandbox?

khmakoto:
Can you do `yarn buildto next`? I think that might be what allows you to do it.

change/@fluentui-react-next-2020-08-13-13-16-29-dismiss-callout-on-window-change.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "dismiss callout when blurring to a new window",
```
khmakoto, line 3:
```suggestion
  "comment": "Callout: Dismiss when blurring to a new window.",
```

change/office-ui-fabric-react-2020-08-05-12-08-00-dismiss-callout-on-window-change.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "dismiss callout when blurring to a new window",
```
khmakoto, line 3:
```suggestion
  "comment": "Callout: Dismiss when blurring to a new window.",
```

jodom:
Got time to clean install monorepo and then validate these changes. Blocking issue yesterday was a folder rename not autoresolved by git
image

packages/react-next/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -475,6 +475,11 @@ export const CalloutContentBase = React.memo(
       return null;
     }
 
+    // We probably just switched focus context to another window. Don't rerender
```
jodom, line 478: @khmakoto , @ecraig12345 , With the new dismiss on window blur logic, I'm seeing a bug in react-next callout where we dismiss properly, but callout pops up right after. My initial intuition was there's a refocus logic somewhere which triggers such a  behavior, although disabling `shouldRestoreFocus` on Popup didn't prevent it. Do you know what else could be causing CalloutContent.base to rerender after dismiss? Repro by disabling this logic block

packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -346,6 +346,12 @@ export class CalloutContentBase extends React.Component<ICalloutProps, ICalloutS
     }
   }
 
+  private _dismissOnTargetWindowBlur = (ev: FocusEvent) => {
+    if (ev.relatedTarget === null && !document.hasFocus()) {
```
ecraig12345, line 350: Will this be the correct `document`? Wondering if you need to use either the `getDocument` utility or `this._targetWindow.document`.

packages/react-next/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -382,6 +382,12 @@ function useDismissHandlers(
       }
     };
 
+    const dismissOnTargetWindowBlur = (ev: FocusEvent) => {
+      if (ev.relatedTarget === null && !document.hasFocus()) {
```
ecraig12345, line 386: Same comment here about ensuring it's the correct document.

packages/react-next/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -468,6 +475,11 @@ export const CalloutContentBase = React.memo(
       return null;
     }
 
+    // We probably just switched focus context to another window. Don't rerender
+    if (!document.hasFocus()) {
```
ecraig12345, line 479: In this case I'm pretty sure you need to use the `getDocument` utility; otherwise this will fail in server-side rendering.

packages/react-next/src/components/Callout/CalloutContent.base.tsx
ecraig12345: I'm not sure offhand. @MLoughry do you have any idea what might be going on since you just converted Callout?

packages/react-next/src/components/Callout/CalloutContent.base.tsx
jodom: This wont get called in server-side rendering, but I'll still make sure we are always getting the right document

ecraig12345: Ah I missed that this was inside a useEffect, sorry

MLoughry: It's not inside a `useEffect`, it's inside the render

jodom: simply blocked by the check for server-side rendering above it

packages/react-next/src/components/Callout/CalloutContent.base.tsx
MLoughry: I'm unclear on the repro. I don't even see this window-blurring code in master....

jodom: The changes are propose in this branch. To repro the rerender I describe above, you could just apply the changes in [current file](https://github.com/microsoft/fluentui/pull/14363/files/ef69a170f2e975d3e1bd736d433a151aa1323890..7353e4c19545f4d66e9d74034ed481f07c58914e#diff-d43dba3bcdb7a43545edf7be95f49465) on any branch you have locally, then comment out this block that return null on render(). With a callout open, clicking outside the page should trigger a dismiss, then you'll see the callout pop back up again

jodom:
@joschect. I'm looking for your review here

packages/react-next/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -491,6 +498,11 @@ export const CalloutContentBase = React.memo(
       return null;
     }
 
+    // We probably just switched focus context to another window. Don't rerender
+    if (!targetWindowRef.current.document.hasFocus()) {
```
joschect, line 502: I think that it might be unsafe to conditionalize render based on this. The event listeners, refs, and other parts of the menu that interact directly with the dom might not get cleaned up properly. If there is something going with focus returning, you can use the new onFocusRestore to prevent focus from returning in whatever scenario you might want.

jodom: I see no such prop in file / repo. Did you mean `shouldRestoreFocus` prop available?

khmakoto: Hey @jodom! Yeah, I think @joschect referred to `shouldRestoreFocus`.

jodom: Oh yeah, it's actually `onFocusRestore` prop on Callouts :) . Thanks

packages/react-next/src/components/Popup/Popup.types.ts
```diff
@@ -51,5 +51,9 @@ export interface IPopupProps extends React.HTMLAttributes<HTMLDivElement> {
    * contains focus. If this is provided, focus will not be restored automatically,
    * you'll need to call originalElement.focus()
    */
-  onRestoreFocus?: (options: { originalElement?: HTMLElement | Window; containsFocus: boolean }) => void;
+  onRestoreFocus?: (options: {
+    originalElement?: HTMLElement | Window;
+    containsFocus: boolean;
+    documentContainsFocus?: boolean;
```
joschect, line 57: I don't believe that this should be able to be undefined. If we can't determine if a document has focus, let's set it to false.

jodom: Would this be a breaking change for the current api?

jodom: Or back compat is mainly relevant for office-ui-fabric-react? Here, I believe I can make it a required without any such concerns

packages/react-next/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -447,9 +448,10 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
 
   private _tryFocusPreviousActiveElement = (options: {
     containsFocus: boolean;
+    documentContainsFocus: boolean;
     originalElement: HTMLElement | Window | undefined;
   }) => {
-    if (options && options.containsFocus && this._previousActiveElement) {
+    if (options && options.containsFocus && options.documentContainsFocus && this._previousActiveElement) {
```
joschect, line 454: Let's leave this check out as it might break some scenario that we're not sure of. Now users who supply their own onFocusRestore can choose to check it if they like.

packages/react-next/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -425,6 +425,7 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     // Popup uses to determine if focus is contained when dismissal occurs
     this._tryFocusPreviousActiveElement({
       containsFocus: this._focusingPreviousElement,
+      documentContainsFocus: this._targetWindow.document.hasFocus(),
```
joschect, line 428: I believe that the changes in this file should be in `packages/office-ui-fabric-react/...`

joschect, line 428: I don't believe you need to port this to the react-next package yet.

packages/office-ui-fabric-react/src/components/Popup/Popup.tsx
```diff
@@ -168,10 +172,14 @@ export class Popup extends React.Component<IPopupProps, IPopupState> {
...
-  const { originalElement, containsFocus } = options;
+function defaultFocusRestorer(options: {
+  originalElement?: HTMLElement | Window;
+  containsFocus: boolean;
+  documentContainsFocus: boolean;
+}) {
+  const { originalElement, containsFocus, documentContainsFocus } = options;
 
-  if (originalElement && containsFocus && originalElement !== window) {
+  if (originalElement && containsFocus && documentContainsFocus && originalElement !== window) {
```
joschect, line 182: Let's leave the `documentContainsFocus` check out for now as it might break existing scenarios.

joschect: Users can chose to check documentContainsFocus if they want in their own override.

packages/react-next/src/components/ContextualMenu/ContextualMenu.base.tsx
jodom: Should I drop all changes in react-next, or just this api change

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -446,12 +447,13 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
...
   private _tryFocusPreviousActiveElement = (options: {
     containsFocus: boolean;
+    documentContainsFocus: boolean;
     originalElement: HTMLElement | Window | undefined;
   }) => {
     if (this.props.onRestoreFocus) {
       this.props.onRestoreFocus(options);
     } else {
-      if (options && options.containsFocus && this._previousActiveElement) {
+      if (options && options.containsFocus && options.documentContainsFocus && this._previousActiveElement) {
```
joschect, line 456: Can we remove this check here? Because you can pass in your own this check isn't needed.

joschect (approved):

romanisa:
@jodom With this change, is there a way we can still keep Callout open upon changing focus to another window or browser tab? Our scenario requires that and we have a data loss bug caused by this.

ecraig12345:
@romanisa Can you file an issue about that?

romanisa:
@ecraig12345 Sure done: https://github.com/microsoft/fluentui/issues/15156

PR #14387, opened 2020-08-06
Fix input and label IDs not respected by choice group option.

### LINKED ISSUE #14386: ChoiceGroup Component: ChoiceGroupOption does not respect custom 'id' and 'labelId' fields
### Environment Information

- **Package version(s)**: Latest - used in fluent UI documentation examples.

https://codepen.io/frags51/pen/YzqzmWQ

#### Actual behavior:
input and label fields for an option have their default randomly generated 'id'

#### Expected behavior:
The respective input fields should have id as set in the choice group options, by the 'id' prop.
Similarly, label should have id as given by 'labelId'.

Are you willing to submit a PR to fix? Yes,

Requested priority: Normal
Normal because you can apply a unique class name to input/label via 'styles' prop as a workaround.

Products/sites affected: MS Internal.

frags51:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14386

#### Description of changes

Passed the correct input and label id's to ChoiceGroupOption.

#### Focus areas to test

Please refer to the issue linked above for more details.

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroup.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroup.base.tsx
@@ -234,7 +234,7 @@ export class ChoiceGroupBase extends React.Component<IChoiceGroupProps, IChoiceG
   }
 
   private _getOptionId(option: IChoiceGroupOption): string {
-    return option.id ? option.id : `${this._id}-${option.key}`;
+    return option.id || `${this._id}-${option.key}`;
   }
 
   private _getOptionLabelId(option: IChoiceGroupOption): string {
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroup.base.tsx
```diff
@@ -234,7 +234,11 @@ export class ChoiceGroupBase extends React.Component<IChoiceGroupProps, IChoiceG
   }
 
   private _getOptionId(option: IChoiceGroupOption): string {
-    return `${this._id}-${option.key}`;
+    return option.id ? option.id : `${this._id}-${option.key}`;
```
ecraig12345, line 237:
```suggestion
    return option.id || `${this._id}-${option.key}`;
```

packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroup.base.tsx
```diff
@@ -234,7 +234,11 @@ export class ChoiceGroupBase extends React.Component<IChoiceGroupProps, IChoiceG
...
+  }
+
+  private _getOptionLabelId(option: IChoiceGroupOption): string {
+    return option.labelId ? option.labelId : `${this._labelId}-${option.key}`;
```
ecraig12345, line 241:
```suggestion
    return option.labelId || `${this._labelId}-${option.key}`; 
```

frags51:
@ecraig12345 Done!

ecraig12345:
Not sure why the build was failing, but there's no way it was due to your change. I started a new build.

PR #14396, opened 2020-08-07
Export new card-related theming slots for SASS

phkuo:
#### Pull request checklist

#### Description of changes

Fix script that autogenerates SASS theming variable files, and add the new card-related semantic slots to them.


phkuo:
@ecraig12345  can I get a review please?

ecraig12345:
There are several copies of this file now which also ought to be updated, in common-styles and some other places. Really we should either have the script update all of the locations, or better yet see if there are any we can delete. If you don't have time to look into that, maybe @behowell can?

phkuo:
@ecraig12345 I don't know why the other copies exist. I'd like to not block this PR on that, since that can be done later, and this PR is blocking feature work.

PR #14398, opened 2020-08-07
ThemeProvider: support V7 theme

xugao:
#### Pull request checklist

#### Description of changes
- Support passing V7 theme (`ITheme`) to `ThemeProvider`
- Added `components` prop in `Theme` to support passing component-level styles (similar to `<Customizer scopeSettings={ Component: { styles: { // styles prop }}} />`)
- `ThemeProvider` now should be able to do everything theming-related that `Customizer` can do
- `ThemeProvider` respects global theme (from `loadThem`) and theme within `CustomizerContext`. All supported scenarios are covered by new screener tests in `apps/vr-tests/src/stories/ThemeProvider.stories.tsx`.
- Removed `Customizer` shimmer previously added in `react-next`. We no longer need it for V8. For converged components (e.g. button), `ThemeProvider` will be required for be functional. For all other V8 components (scoped to only contains FC changes, without `scss` conversion), current `Customizer` is sufficient.
- Removed `lib/compat` inside `react-theme-provider`. We might need to have `lib/next` later based on work beyond v8.

#### Upcoming
- Make V7 theme works for `react-button`: inside `getTokens`, complete mapping `Theme` to `Tokens` which are used by buttons

#### Focus areas to test
Screener tests inside `apps/vr-tests/src/stories/ThemeProvider.stories.tsx`


apps/vr-tests/src/stories/ThemeProvider.stories.tsx
```diff
@@ -0,0 +1,160 @@
...
+        <PrimaryButton>Customized by ThemeProvider</PrimaryButton>
+      </ThemeProvider>
+    </Customizer>
+  ))
+  .addStory('ThemeProvider wraps Customizer', () => (
+    <ThemeProvider
+      theme={{
+        semanticColors: {
+          primaryButtonBackground: '#FFF',
+          primaryButtonText: 'red',
```
dzearing, line 145: why is `createTheme` used elsewhere but not here?

packages/storybook/src/themes/v8/FluentTheme.ts
```diff
@@ -1,11 +1,7 @@
-import { Theme } from '@fluentui/react-theme-provider';
+import { PartialTheme } from '@fluentui/react-theme-provider';
 
-export const FluentTheme: Theme = {
+export const FluentTheme: PartialTheme = {
```
dzearing, line 3: we should really consider renaming this package to `@fluentui/storybook-addons` or something that doesn't seem like a fork of storybook.

packages/react-theme-provider/src/useTheme.ts
```diff
@@ -1,8 +1,30 @@
...
+    return legacyTheme;
+  }
+
+  return createDefaultTheme();
+};
+
+/**
+ * Get theme from CustomizerContext or Customizations singleton.
+ */
+function useLegacyTheme(): ITheme {
```
dzearing, line 28: useCompatTheme?

packages/storybook/src/themes/v8/FluentTheme.ts
xugao: I agree, will take that as a follow-up!

apps/vr-tests/src/stories/ThemeProvider.stories.tsx
xugao: so `ThemeProvider` supports partial theme, but `Customizer` actually doesn't. If we see places i use `createTheme`, it's for creating a full theme to pass to `Customizer`. (This is an improvement for  `ThemeProvider` compared to `Customizer`)

PR #14404, opened 2020-08-07
feat(Icon): factory patch for supporting jsx icons in the shorthands

mnajdova:
This PR augments the Icon's create factory to be able to convert new jsx icons to the current icon shorthand format.

image

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Icon/Icon.tsx
+++ b/packages/fluentui/react-northstar/src/components/Icon/Icon.tsx
@@ -192,7 +192,9 @@ function isFluentJSXIcon(value: ShorthandValue<IconProps>): value is React.React
 }
 
 function toIconNameFromDisplayName(displayName: string) {
-  return inconsistentIconNames[displayName] ? inconsistentIconNames : _.kebabCase(displayName.replace('Icon', ''));
+  return inconsistentIconNames[displayName]
+    ? inconsistentIconNames[displayName]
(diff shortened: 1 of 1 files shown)


packages/fluentui/react-northstar/src/components/Icon/Icon.tsx
```diff
@@ -147,7 +161,73 @@ Icon.propTypes = {
...
+      if ((value as IconProps).name) {
+        console.warn(
+          `Fluent UI: Deprecation notice, please use JSX icon as a value instead of an object, for example replace <Button icon={{ name: 'bell', outline: true }} /> with <Button icon={<BellIcon outline />} />.`,
+        );
+      }
+    }
+  }
+
+  if (isFluentJSXIcon(value)) {
+    const iconName = toIconNameFromDisplayName(value.type.displayName.replace('Icon', ''));
```
layershifter, line 222: If `value.type.displayName` is `MoreIcon` it will cut it to `More` and then `inconsistentIconNames` will not work

```suggestion
    const iconName = toIconNameFromDisplayName(value.type.displayName);
```

mnajdova: Good catch, we have the replace again here https://github.com/microsoft/fluentui/pull/14404/files#diff-b0039d331230fd1622db7d2eea352483R195

PR #14407, opened 2020-08-07
Fixed lage with scoping issues with fluentui:test:e2e and deploy sites

kenotron:
The previous attempt of `lage` merge broke the "fluentui:test:e2e". It was due to the fact that the tests expect a full build but lage --since is SKIPPING those builds (v7 changes don't trigger v0 code yet the v0 tests are being run).


azure-pipelines.yml
```diff
@@ -122,6 +84,13 @@ jobs:
         env:
           DANGER_GITHUB_API_TOKEN: $(DANGER_GITHUB_API_TOKEN)
 
+      - script: |
+          yarn lage build --scope @fluentui/e2e --no-deps --verbose --no-cache --grouped
```
ecraig12345, line 88: I'm confused, won't skipping deps cause these to fail? Or does lage caching handle that somehow?

kenotron: no, --no-deps is a terrible name - it means no dependents (downstream)
so it's a "build up to"

kenotron: I could change the scope to @fluentui/* but that will include web-components as well. I think if you have time, you could make e2e's test script actually be the test (with e2e project taking a dep of everything else v0). That way, even that can participate the lage dep graph based builds.

ecraig12345: `@fluentui/*` would build a bunch of other random stuff for both v0 and v7 now. Also if you get a chance it would probably be good to change `--no-deps` to have a better name.

ecraig12345: I'm pretty sure lerna has a similarly-named option which does the exact opposite...

kenotron: that one is called --include-dependents or --include-dependencies

kenotron: jason gave me the same feedback, btw - i want to switch to dependent & dependencies

PR #14427, opened 2020-08-09
No scope on master builds

kenotron:
This will disable scoped builds for CI master builds because there's no scope at all. Build the whole thing for master.


kenotron:
Merging this one because screener is hitting that infamous timeout-even-though-it-is-successful bug.

PR #14428, opened 2020-08-09
changing vr-tests to reference static images rather than 3rd party service

dzearing:
Adding a few test images to vr-tests public folder and removing dependency on placehold.it

There are a couple of Tile and FolderCover tests that still have references to placehold.it but the urls are dynamic. Need to figure those out separately.


ecraig12345:
Looks like it's not quite working...
image

dzearing:
Still weeding this out;

xugao:
Thanks! (I volunteered to make this change at one point but i forgot to followup) 
I think you will need to configure an image loader in webpack for the images to work. you will need to import them then use in code.

layershifter:
@dzearing is it alive or should be closed? 🤔 

In #15097 @ecraig12345 added an ability to upload images from `packages/fluentui/docs/src/public` to CDN. Should be this reused?

ecraig12345:
IIRC this got stuck on how to get the images to load, which we've more recently solved for n* by uploading the images to the fabric-website CDN. If we want to go forward with this we should add these images to the CDN upload too and reference them from there.

ecraig12345:
Oops now that I read @layershifter's full comment I see he mentioned that as well...reading too fast. :P

dzearing:
Closing as this has become stale.

PR #14441, opened 2020-08-10
 feat(react-image): align with design spec

layershifter:
#### Pull request checklist

#### Description of changes

This PR implements the proposed design spec.

#### Focus areas to test

(optional)


packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroup.base.tsx
```diff
@@ -238,7 +238,7 @@ export class ChoiceGroupBase extends React.Component<IChoiceGroupProps, IChoiceG
   }
 
   private _getOptionLabelId(option: IChoiceGroupOption): string {
-    return option.labelId || `${this._labelId}-${option.key}`; 
+    return option.labelId || `${this._labelId}-${option.key}`;
```
layershifter, line 241: 🤦

packages/fluentui/accessibility/src/behaviors/Image/imageBehavior.ts
```diff
@@ -15,6 +15,12 @@ export const imageBehavior: Accessibility<ImageBehaviorProps> = props => ({
   },
 });
 
+export function useImageAria(props: ImageBehaviorProps): AccessibilityAttributes {
+  return {
+    'aria-hidden': props.alt || props['aria-label'] ? undefined : 'true',
+  };
+}
```
layershifter, line 22: I don't think that it's an expected way to go. @mnajdova do you have some input on this?

packages/react-image/src/components/Image/useImage.ts
```diff
@@ -23,6 +24,26 @@ export const useImage = (
...
+      const rect = imageRef.current.getBoundingClientRect();
+      let actualImageSize: 'small' | 'medium' | 'large' = 'small';
+
+      if (rect.width > 80) {
+        actualImageSize = 'large';
+      } else if (rect.width > 32) {
+        actualImageSize = 'medium';
+      }
+
+      imageRef.current.setAttribute('data-image-size', actualImageSize);
```
layershifter, line 44: _This is the most controversial part 🤔_  The design spec requires to have different border sizes (see `Image.scss`) for different image sizes.

- Is it a valid requirement at all?
- Should be this a part of `BaseImage` component or only of `Image`?
- Better solutions? 😺

layershifter: /cc @khmakoto @dzearing

packages/react-image/src/components/Image/Image.scss
```diff
@@ -6,12 +6,15 @@
...
 **/
 :root {
-  --image-borderRadius: 0;
+  --image-borderColor: rgb(225, 223, 221) /* TODO: replace with a proper value */;
+  --image-borderRadius: none;
+  --imageBoxShadhow: 0px 2px 4px -0.75px rgba(0, 0, 0, 0.1) /* --effects-elevation4 */;
+
+  --image-rounded-small-borderWidth: 2px /* --effects-roundedCorner2 */;
+  --image-rounded-medium-borderWidth: 4px /* --effects-roundedCorner4 */;
+  --image-rounded-large-borderWidth: 6px /* --effects-roundedCorner6 */;
```
layershifter, line 15: @dzearing is there some guidance how tokens should be named and how they are connected with themes?

packages/fluentui/accessibility/src/behaviors/Image/imageBehavior.ts
mnajdova: Ideally I think we like to re-use the structure we had before, as it's clear by what we had so far we need to define aria props for all the slots inside. 

For the key-handlers, still not sure, as I am still looking over those. Not sure if we should make them part of the accessibility or just add them to the `useImage` hook directly.. 

Anyway, I would at least propose adding these inside `root`, and then if we extend the APIs we can add attributes as a first key.

packages/react-image/src/components/Image/useImage.ts
```diff
@@ -23,6 +24,26 @@ export const useImage = (
   options: ComposePreparedOptions,
 ): ImageState => {
   const imageRef = React.useRef<HTMLElement>(null);
+  const ariaProps = useImageAria(props);
```
mnajdova, line 27: I believe this would get more complicated and we will have to have some common hook to resolve these (maybe something similar to the `useAccessibility`, but I am not sure yet.

packages/react-image/src/components/Image/useImage.ts
dzearing: This seems like a bad requirement unless it were handled universally on every component and likely should be handled in the parent styling. I am not understanding the use case. It seems like it would depend on the use case to define.

I'd suggest we ask the designer for use cases they had in mind and understand if this is more a higher level concern or really belongs in the `Image`. I thought of the `Image` more as a utility.

packages/react-image/src/components/Image/Image.scss
dzearing: Does it make sense to make an image specific set of border tokens? I could understand a bordered modifier assigning values to general component tokens. This is probably more of a designer question; i don't understand the use case well.

Component tokens will generally follow the form:

```
--{componentName}-{state if applicable}-{property}
```

For flags, what you're doing here seems logical; you're defining what the enum `rounded` when set to `small` translates to for `borderWidth`. But I would question if that is valuable for a component token. Are there many scenarios where we would only want to round the border, and not apply some other effect? Would a variant make more sense?

How they fit into the theme - this would work right now:

```
const theme = { 
  tokens: {
   color: {},
   effects: {
      roundedCorner2: '2px'
   },
   etc: {},

    image: { // component tokens
      rounded: {
         small: {
           // we should have this string in a consts in a base-theme package
           borderWidth: 'var(--effects-roundedCorner2)',
        }
    }
 }
};
```

On type safety, the image object could be case as an `ImageTokens` object, or, we standardize on a factory to produce the image `createImageTheme({ partial image token })`.

packages/office-ui-fabric-react/src/components/ChoiceGroup/ChoiceGroup.base.tsx
ecraig12345: Fixed in master now.

packages/fluentui/accessibility/src/behaviors/Image/imageBehavior.ts
mnajdova: After some thoughts, I think we should add these to the `useButtonState` hook. I really don't see a value for us learning to write yet another hook, when already some of the props & key handlers are inside this hook. We can work more towards testing the a11y better, but from implementation point of view, I think it's easier to just include these things inside the `useButtonState` hook. What are your thoughts?

packages/react-image/src/components/Image/useImage.ts
layershifter: I have an approval from design, we will use `medium` as a default. We will not use this logic.

packages/fluentui/accessibility/src/behaviors/Image/imageBehavior.ts
layershifter: Changes to accessibility will be done in #14506.

packages/react-image/src/components/Image/useImage.ts
layershifter: Changes to accessibility will be done in #14506.

packages/react-image/src/components/Image/Image.scss
```diff
@@ -6,12 +6,13 @@
  * they will be resolved on the :root element only; not at usage time.
 **/
 :root {
-  --image-borderRadius: 0;
+  --image-borderColor: rgb(225, 223, 221) /* TODO: replace with a proper value */;
+  --image-borderRadius: none;
+  --imageBoxShadhow: 0px 2px 4px -0.75px rgba(0, 0, 0, 0.1) /* --effects-elevation4 */;
```
layershifter, line 11:
```suggestion
  --image--boxShadow: 0px 2px 4px -0.75px rgba(0, 0, 0, 0.1) /* --effects-elevation4 */;
```

apps/pr-deploy-site/pr-deploy-site.js
```diff
@@ -26,6 +26,12 @@ const siteInfo = [
     icon: 'LikeSolid',
     title: 'Button',
   },
+  {
+    package: '@fluentui/react-image',
+    link: './react-button/dist-storybook/index.html',
```
layershifter, line 31:
```suggestion
    link: './react-image/dist-storybook/index.html',
```

mnajdova (approved):
LGTM 👍

PR #14454, opened 2020-08-10
Change question issue template to a link

micahgodbolt:
This creates an 'issue template' that is simply a link 

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,6 +1,6 @@
 --- 
 contact_links: 
   - 
-    about: "Please ask and answer usage questions on Stack Overflow."
+    about: "Please ask and answer usage questions on Stack Overflow. (If you are a Microsoft employee you can also join the \"Fluent UI Community\" Teams channel.)"
     name: Question
     url: "https://stackoverflow.com/questions/tagged/fluentui"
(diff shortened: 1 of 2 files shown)


micahgodbolt:
delete the other 'question' entry after this is in

ecraig12345:
Can you just delete the question template in this PR too?

micahgodbolt:
i just created the file via the github gui :D not sure how i can add that deletion onto this

micahgodbolt:
ok...i'll stop being lazy and I cloned the repo, deleted the file, pushed

micahgodbolt:
feel free to change this to whatever you like...i just saw this link example and thought it was a MUCH better approach to what we're doing here

ecraig12345:
I updated the link to also mention that we have Teams for MS employees (I think that's important to call out).

One remaining question, what tag should people use? The old question template suggests 2 or 3 different ones, and I'm not sure which (if any) are being actively monitored.

micahgodbolt:
https://stackoverflow.com/questions/tagged/office-ui-fabric-react is at 91 


https://stackoverflow.com/questions/tagged/fluent-ui is at 63

Seems fluent-ui would be safe moving forward

micahgodbolt:
closing this down. feel free to re-open merge/ or use this approach in the future.

PR #14463, opened 2020-08-11
focusing legends based on props condition

### LINKED ISSUE #14466: legends tab stops
we should expose a prop such as `allowFocusOnLegends` if this prop is set true then we should be able to focus the legends in the charts else not

shivasai09:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14466 

#### Description of changes

focusing the legends based on the the prop 'allowFocusOnLegends' and exposing the new prop 'legendProps' on all the charts which are using legends form which they can manipulate the respective lengends with this prop

#### Focus areas to test

(optional)


packages/charting/src/components/DonutChart/DonutChart.types.ts
```diff
@@ -84,6 +85,11 @@ export interface IDonutChartProps {
    * @default false
    */
   hideTooltip?: boolean;
+
+  /**
+   * props for the legneds in the chart
```
Raghurk, line 90: typo

shivasai09: done

PR #14470, opened 2020-08-11
feat: add useCSS hook

layershifter:
An alternative for #13641.

---

This PR adds `useCSS()` hook for Fluent UI Northstar. It allows to use CSS-in-JS to generate a `className` that can be applied to any component. To merge styles I used an approach inspired by Emotion's [`cx()` function](https://emotion.sh/docs/emotion#cx) which allows merging styles and/or class names.

```tsx
const red = useCSS({ color: 'red' })

const greenWins = useCSS(red, { color: 'green' })
const redWins = useCSS({ color: 'green' }, red)
```

All features like nested & preudo selectors are supported.

---

I've added 3 perf examples to this PR.  The first is representative of usage in Teams today.  The second two are for comparison to that baseline, both using `useCSS`.  This was done to ensure `useCSS` would not be slower than what is in react-northstar usage today.

| Example                      | min   | avg   | median | max   |
| ---------------------------- | ----- | ----- | ------ | ----- |
| **ButtonOverridesMiss.perf.tsx** - Applies boolean variable overrides | 25.34 | 30.05 | 28.85  | 47.01 |
| **ButtonUseCss.perf.tsx** - Applies style object overrides        | 13.67 | 19.42 | 19.4   | 27.64 |
| **ButtonUseCssNesting.perf.tsx** - Merges generated class names | 22.41 | 26.5  | 25.11  | 37.53 |


mnajdova:
image

packages/fluentui/react-bindings/src/hooks/useCSS.ts
```diff
@@ -0,0 +1,94 @@
+import { serializeStyles } from '@emotion/serialize';
+import { ICSSInJSStyle, ThemePrepared } from '@fluentui/styles';
+import cx from 'classnames';
+// @ts-ignore
+import _Stylis from 'stylis';
+// @ts-ignore No typings :(
+import focusVisiblePlugin from '@quid/stylis-plugin-focus-visible';
+// @ts-ignore No typings :(
```
mnajdova, line 8: :(

packages/fluentui/react-bindings/src/hooks/useCSS.ts
```diff
@@ -0,0 +1,94 @@
...
+import rtlPlugin from 'stylis-plugin-rtl';
+
+import { useFluentContext } from '../context';
+
+//
+// Types
+//
+
+export type UseCSSStyle = Omit<ICSSInJSStyle, 'animationName'>;
```
mnajdova, line 17: Do we support our animation definition in `useCss`? If not we do not have to omit this prop.

packages/fluentui/react-bindings/src/hooks/useCSS.ts
```diff
@@ -0,0 +1,94 @@
...
+    } else {
+      acc.push(style);
+    }
+
+    return acc;
+  }, []);
+
+  // serializeStyles() will concat all passed styles and will resolve functions
+  const serializedStyles = serializeStyles(resolvedStyles, stylesCache, theme);
+  // ".name" is not a valid CSS classname
```
mnajdova, line 78: Why? And why adding an "f" before it makes it valid?

packages/fluentui/react-bindings/test/hooks/useCSS-test.tsx
```diff
@@ -0,0 +1,143 @@
...
+        renderer: { renderGlobal },
+        rtl,
+      },
+    },
+    wrappingComponent: Unstable_FluentContextProvider,
+  };
+}
+
+describe('useCSS', () => {
+  it('returns a className that contains', () => {
```
mnajdova, line 38:
```suggestion
  it('returns a className', () => {
```

packages/fluentui/react-bindings/test/hooks/useCSS-test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  });
+
+  it('handles RTL without collisions', () => {
+    const styles = [{ left: '20px', paddingLeft: '20px' }];
+    const renderGlobal = jest.fn();
+
+    mount(<TestComponent styles={styles} />, getMountOptions(renderGlobal, false));
+    mount(<TestComponent styles={styles} />, getMountOptions(renderGlobal, true));
+
+    expect(renderGlobal).toMatchInlineSnapshot(`
```
mnajdova, line 115: Wouldn't it be better if we test with `window.getComputerStyle`? Feels like more stable that snapshot

packages/fluentui/react-bindings/test/hooks/useCSS-test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    wrappingComponent: Unstable_FluentContextProvider,
+  };
+}
+
+describe('useCSS', () => {
+  it('returns a className that contains', () => {
+    const styles = [{ color: 'red' }];
+    const wrapper = mount(<TestComponent styles={styles} />);
+
+    expect(wrapper.find('div').prop('className')).toBeDefined();
```
mnajdova, line 42: On the other hand shouldn't we also test what that `className` is, or what styles the component has.

packages/fluentui/react-bindings/src/hooks/useCSS.ts
layershifter: No we don't support, changed to:

```tsx
// Inline keyframe definitions are not supported by useCSS() hook
export type UseCSSStyle = Omit<ICSSInJSStyle, 'animationName'> & { animationName?: string };
```

packages/fluentui/react-bindings/src/hooks/useCSS.ts
layershifter: Changed to `// ".name" is not a valid CSS classname as it can start from a digit`, now it should be cleaner 🐱

packages/fluentui/react-bindings/test/hooks/useCSS-test.tsx
layershifter: It means that we will need to actually invoke Fela/Emotion to the process that can make results depend on a renderer...

packages/fluentui/react-bindings/test/hooks/useCSS-test.tsx
layershifter: Changed it as `className` is stable

packages/fluentui/react-bindings/src/hooks/useCSS.ts
```diff
@@ -0,0 +1,95 @@
...
+
+// Inline keyframe definitions are not supported by useCSS() hook
+export type UseCSSStyle = Omit<ICSSInJSStyle, 'animationName'> & { animationName?: string };
+export type UseCSSStyleInput = string | UseCSSStyle | ((theme: ThemePrepared) => UseCSSStyle);
+
+//
+// Definitions
+//
+
+const SPECIFICITY_CLASSNAME = 'fcss';
```
levithomason, line 25: I find the previous `.increase-specificity` more informative than `fcss`.  We could also consider `.use-css` or something as well.

Having been told to consider production size of the CSS, we could use `.css` which would still correlate to the `useCSS` call.

dzearing:
Overall, this is great, it helps to abstract the css in js calls to use any centralized renderer, and extracts the important parts (rtl processing, merging, classname to style destructuring, etc).

I'm wondering if there is a way to avoid all of the work on every render when it's not necessary, like deps array or string key.

packages/fluentui/react-bindings/src/hooks/useCSS.ts
```diff
@@ -0,0 +1,97 @@
...
+// @ts-ignore No typings :(
+import rtlPlugin from 'stylis-plugin-rtl';
+
+import { useFluentContext } from '../context';
+
+//
+// Types
+//
+
+// Inline keyframe definitions are not supported by useCSS() hook
```
pompomon, line 17: Should we put that note in the docs as well?

levithomason:
Perf definitely needs to be tested and proven before we can publicly ship this.  For now, let's not let premature optimization hold up testing it.  We need to merge and test against a couple component pages in the docs and in some production app code.

Once we do this, it may fail entirely, in which case there's no need to spend time optimizing.  If it passes those pragmatic checks, let's hit the perf hard.

packages/fluentui/docs/src/examples/components/Button/Performance/index.tsx
```diff
@@ -11,6 +11,9 @@ const Performance = () => (
       description="A default test."
       examplePath="components/Button/Performance/ButtonMinimal.perf"
     />
+    <ComponentPerfExample title="Overrides" examplePath="components/Button/Performance/ButtonOverridesMiss.perf" />
+    <ComponentPerfExample title="useCSS" examplePath="components/Button/Performance/ButtonOverridesUseCss.perf" />
```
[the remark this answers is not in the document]
layershifter: Yep, good catch. I fixed it previously, now it is:

```tsx
<ComponentPerfExample title="Overrides" examplePath="components/Button/Performance/ButtonOverridesMiss.perf" />
<ComponentPerfExample title="useCSS" examplePath="components/Button/Performance/ButtonUseCss.perf" />
<ComponentPerfExample title="useCSS nesting" examplePath="components/Button/Performance/ButtonUseCssNesting.perf" />
```

Issue #14449, opened 2020-08-10
FluentUI build system optimizations


kenotron:
#### Describe the feature that you would like added

I've been looking to optimize the build times in PR / CI with the integration of `lage` which provides caching, scoping, and pipelining. To completely get to an optimized build, we should have these bits of work done:

- [x] optimize the speed of component info generation (#14590)
- [x] make all the build steps run from a package, rather than assuming they're run globally through a root level gulpfile.js (#14472 #14547)
- [ ] speed up v7 ts only typecheck once
- [x] @fluentui/docs build & @fluentui/react-northstar test both generate component info - and they both take 4-5 minutes uncached, CI is suffering from that (#14590)

#### What component or utility would this be added to

/packages/fluentui/*

#### Have you discussed this feature with our team, and if so, who

Discussing this with @layershifter, @dzearing, @ecraig12345, @jurokapsiar, @miroslavstastny

layershifter:
@kenotron it seems that all things from v0 side are done there. Is there something that we need to change?

ecraig12345:
Closing this since most elements are done and any issues moving forward will be tracked elsewhere.

PR #14472, opened 2020-08-11
chore: use componentInfo per package

layershifter:
Related #14449.

---

This PR moves `componentInfo` from `@fluentui/docs` package to matching packages. This fixes the issue with unclear build dependencies between packages.

Also removes UMD build, it's broken even from Stardust UI times and is nowhere used.


.gitignore
```diff
@@ -3,7 +3,6 @@
 *.info.json
 *.tsbuildinfo
 packages/fluentui/docs/src/componentInfo
```
layershifter, line 5: This should be updated

packages/fluentui/react-northstar/gulpfile.ts
```diff
@@ -4,5 +4,5 @@ import { parallel, series, task } from 'gulp';
 import '../../../gulpfile';
 
 // Redefine test and test:watch to build info json files beforehand
-task('test', series('build:docs:component-info', 'test:jest'));
-task('test:watch', series('build:docs:component-info', parallel('test:jest:watch', 'watch:docs:component-info')));
+task('test', series('build:component-info', 'test:jest'));
```
dzearing, line 7: does this still need to happen here if it's happening in `build` ?

layershifter: Yes, as in Northstar there is no requirement to build before test run

packages/fluentui/docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -22,11 +22,13 @@ import { NavLink, withRouter, RouteComponentProps } from 'react-router-dom';
...
 import { renderSidebarTitle } from './SidebarTitle';
 import config from '../../config';
+import componentInfoContext from '../../utils/componentInfoContext';
 
 type ComponentMenuItem = { displayName: string; type: string };
 
 const pkg = require('@fluentui/react-northstar/package.json');
-const componentMenu: ComponentMenuItem[] = require('../../componentMenu');
+
+const componentMenu: ComponentMenuItem[] = _.sortBy(componentInfoContext.parents, 'displayName');
```
layershifter, line 31: This change allows to get rid of additional JSON file 🚀

scripts/gulp/tasks/docs.ts
```diff
@@ -212,14 +203,26 @@ task('serve:docs:stop', () => forceClose(server));
...
+          )
+          .pipe(dest('componentInfo', { cwd: pkg.packagePath }));
+      // @ts-ignore
+      internalTask.displayName = 'build:docs:component-info';
+
+      const watcher = watch(pkg.packageJson.gulp?.componentInfo, { cwd: pkg.packagePath }, internalTask);
+
+      watcher.on('add', logWatchAdd);
+      watcher.on('change', logWatchChange);
+    }
```
layershifter, line 224: It will be great to remove this, but running three copies of `watch:component-info` explodes my machine 😭

ecraig12345:
Went ahead and merged this to see if it fixes another issue. Hopefully that's okay (or if it's really bad we can revert).

PR #14474, opened 2020-08-11
Target new pool for CI

jdhuntington:
## Description of changes

Use new build pool for CI.


ecraig12345:
By "somewhere" I probably mean a private onenote or wiki (if you go with a wiki, be sure to link from the team onenote).

jdhuntington:
Absolutely; assuming this works the plan is to update our current internal infrastructure README/automation with references to this set of machines, then we can discuss merging for real and rollback plans if something goes wrong down the road.

PR #14478, opened 2020-08-12
Enable eslint-plugin-react-hooks for react-next

ecraig12345:
#### Pull request checklist

#### Description of changes

Enable `eslint-plugin-react-hooks` for react-next to help catch hook-related mistakes, and fix existing issues. Also add a rule to our eslint config to help catch accidental usage of DOM globals (such as `focus()` which was accidentally used in SearchBox).

In the interest of getting this merged more quickly, I temporarily disabled the rule in a bunch of places for CalloutContent and SpinButton and will make follow-up issues to fix this.

Also started making some updates to the release notes breaking changes to alphabetize components and add a couple missing things (or notes about components to look over later).

### DIFF
--- a/packages/react-next/RELEASE_NOTES.md
+++ b/packages/react-next/RELEASE_NOTES.md
@@ -2,17 +2,9 @@
 
 ## Breaking changes
 
-### Beak
+### Calendar
 
-- Removed empty `IBeak` interface
-- Removed `componentRef` prop
-
-### SpinButton
-
-- Simplified props to `ISpinButtonStyles` to include only the parts of the component to bring inline with
-  other components.
-- Replaced `getClassNames` legacy prop with `styles` prop to bring component consistent to other components
-  and improve cachability of internal styles.
+TODO: Diff of OUFR vs date-time Calendar
 
 ### Checkbox
 
@@ -24,10 +16,27 @@
 ### Coachmark
 
 - Removed `isBeaconAnimating` and `isMeasured` style props
+- Beak:
+  - Removed empty `IBeak` interface
+  - Removed `componentRef` prop
+
+### DatePicker
+
+TODO: Diff of OUFR vs date-time DatePicker
 
 ### Pivot
 
 - Removed deprecated and redundant props from v7, including: `initialSelectedKey` and `defaultSelectedIndex`. Use `selectedKey` or `defaultSelectedKey` to define the selected tab, and provide `itemKey` on pivot item children.
+  - TODO: enumerate all removed props
+
+### Slider
+
+TODO: document any API or functionality changes
+
+### SpinButton
+
+- Simplified props to `ISpinButtonStyles` to include only the parts of the component to bring inline with other components.
+- Replaced `getClassNames` legacy prop with `styles` prop to bring component consistent to other components and improve cachability of internal styles.
 
 ### Others
 
@@ -45,7 +54,7 @@
 
 - Updated enums to string union type: `PivotLinkFormat`, `PivotLinkSize`. (#13370)
 
-## Changes worth callout
+## Other notable changes
 
 - `styles` prop backward compat solution.
 - css variables and IE 11 solution.
--- a/packages/react-next/src/components/Coachmark/Coachmark.base.tsx
+++ b/packages/react-next/src/components/Coachmark/Coachmark.base.tsx
@@ -118,7 +118,7 @@ function useBeakPosition(
   targetAlignment: RectangleEdge | undefined,
   targetPosition: RectangleEdge | undefined,
 ) {
-  const { theme } = props;
+  const isRTL = getRTL(props.theme);
 
   return React.useMemo(() => {
     const beakDirection = targetPosition === undefined ? RectangleEdge.bottom : getOppositeEdge(targetPosition);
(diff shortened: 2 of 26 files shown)


packages/react-next/src/components/Coachmark/Coachmark.base.tsx
```diff
@@ -116,36 +118,15 @@ function useBeakPosition(
...
   targetPosition: RectangleEdge | undefined,
 ) {
-  const beakDirection = targetPosition === undefined ? RectangleEdge.bottom : getOppositeEdge(targetPosition);
+  const { theme } = props;
 
-  /**
-   * The left position of the beak
-   */
-  const [beakLeft, setBeakLeft] = React.useState<string>();
+  return React.useMemo(() => {
```
ecraig12345, line 123: The parts of the beak position will never change independently, so this is a much more efficient way to calculate and update them.

packages/react-next/src/components/Coachmark/Coachmark.base.tsx
```diff
@@ -270,58 +246,50 @@ function useProximityHandlers(
...
-        // of setTimeout.
+        // When the window resizes we want to async get the bounding client rectangle.
+        // Every time the event is triggered we want to setTimeout and then clear any previous
+        // instances of setTimeout.
         events.on(window, 'resize', (): void => {
           timeoutIds.forEach((value: number): void => {
-            clearInterval(value);
+            clearTimeout(value);
           });
+          timeoutIds.splice(0, timeoutIds.length); // clear array
```
ecraig12345, line 281: Without this, the array of timeouts keeps unnecessarily growing (and each one is unnecessarily cleared again) with every resize event

packages/react-next/src/components/SearchBox/SearchBox.base.tsx
```diff
@@ -26,27 +28,29 @@ const useComponentRef = (
...
     underlined,
     styles,
     // eslint-disable-next-line deprecation/deprecation
     labelText,
+    // eslint-disable-next-line deprecation/deprecation
+    placeholder = labelText,
     theme,
-    clearButtonProps = { ariaLabel: 'Clear text' },
+    clearButtonProps = defaultClearButtonProps,
     disableAnimation = false,
```
xugao, line 50: let's move these to `SearchBoxBase.defaultProps` instead

packages/react-next/RELEASE_NOTES.md
```diff
@@ -2,17 +2,9 @@
...
-- Removed empty `IBeak` interface
-- Removed `componentRef` prop
-
-### SpinButton
-
-- Simplified props to `ISpinButtonStyles` to include only the parts of the component to bring inline with
-  other components.
-- Replaced `getClassNames` legacy prop with `styles` prop to bring component consistent to other components
-  and improve cachability of internal styles.
+TODO: Diff of OUFR vs date-time Calendar
```
xugao, line 7: good call; this also reminded me, I have chatted with @lorejoh12 at one point related to why can't we just remove OUFR calendar and export and maintain only `date-time`. He did mention to me it's fully backward compatible (we should definitely double check), but the blocker is around untangling the dependencies (I'm not exactly sure how much work is needed, but we will run into this problem soon).

packages/react-next/src/components/Coachmark/Coachmark.base.tsx
```diff
@@ -182,45 +163,40 @@ function useBeakPosition(
...
 
-    setTransformOrigin(`${transformOriginX} ${transformOriginY}`);
-  }, [targetAlignment, targetPosition]);
-
-  return [
-    { direction: beakDirection, top: beakTop, bottom: beakBottom, left: beakLeft, right: beakRight } as const,
-    transformOrigin,
-  ] as const;
+    return [beakPosition as Readonly<BeakPosition>, `${transformOriginX} ${transformOriginY}`] as const;
+  }, [targetAlignment, targetPosition, theme]);
```
xugao, line 199: define `isRTL = getRTL(theme)` outside and pass `isRTL` here

packages/react-next/RELEASE_NOTES.md
ecraig12345: I thought the plan for v8 was to remove the OUFR Calendar and DatePicker and re-export those from date-time. We need to clarify that.

ecraig12345: Hmm though I see what you mean by dependency issues... If we can't find a better approach, we could potentially move the date-time components into the main package and have date-time re-export them for now. Then move them back and fix the dependencies properly once individual-component packages are available.

ecraig12345: I made a note in the scope doc about looking into this, but in any case the decision shouldn't block this PR (since release notes aren't final).

packages/react-next/src/components/SearchBox/SearchBox.base.tsx
ecraig12345: We don't really have a standard pattern around default props for non-`composed` function components (or even our class components) right now, so I think it's better to leave this as-is.

xugao: i just learned defaultProps is getting deprecated https://github.com/yannickcr/eslint-plugin-react/issues/2396

PR #14486, opened 2020-08-12
Feat/consolidating date time utilities

rymeskar:
#### Pull request checklist

#### Description of changes

* Added more default localization english strings
* Added localizations strings to the date formatting functions. The reason is that the users can only change the localization strings and suddenly the whole formatting would be in the selected language.

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -349,11 +349,6 @@ Datepicker.defaultProps = {
   required: false,
 
   ...DEFAULT_CALENDAR_STRINGS,
-
```
pompomon, line 352: Please could you also move "Open calendar" string from the calendarIcon button?

PR #14491, opened 2020-08-12
Pickers: Update to remove deprecated lifecycle methods and enable concurrent mode

### LINKED ISSUE #14315: BasePicker: strict mode compliance (UNSAFE_)
BasePicker has the following methods which prevent strict mode compliance:
- `componentWillReceiveProps` which maps `props.selectedItems` to state and handles focusing.
- `componentWillUpdate` which synchronizes the selection's items with the items from state. This could potentially be moved to `componentDidUpdate` (with a `forceUpdate` call to force re-rendering).

Converting to a function component is not an option here (without larger architectural changes) since pickers currently rely on inheritance.

joschect:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14315

#### Description of changes

Pickers: Update to remove deprecated lifecycle methods and enable concurrent mode

#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
+++ b/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
@@ -106,6 +106,13 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
   private _id: string;
   private _async: Async;
 
+  public static getDerivedStateFromProps(newProps: IBasePickerProps<any>) {
+    if (newProps.selectedItems) {
+      return { items: newProps.selectedItems };
+    }
+    return null;
+  }
+
   constructor(basePickerProps: P) {
     super(basePickerProps);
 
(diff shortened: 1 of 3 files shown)


packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx
```diff
@@ -106,6 +106,13 @@ export class BasePicker<T, P extends IBasePickerProps<T>> extends React.Componen
   private _id: string;
   private _async: Async;
 
+  public static getDerivedStateFromProps(newProps: IBasePickerProps<any>) {
```
ecraig12345, line 109: This unconditionally overwrites `state.items` with `props.items` **on every render**, not just on prop updates--which means setting `state.items` in `_updateSelectedItems` will no longer do anything. So you'll probably need a different approach here.

ecraig12345: For this reason, in some cases moving logic to `componentDidUpdate` under a `this.props !== prevProps` check provides a closer approximation of old `componentWillReceiveProps` behavior. This means technically you'd have slightly-mismatched data for one render, but IIRC the render with mismatched data will never actually be shown because React flushes any additional renders caused by `componentDidUpdate` before actually updating the DOM. So it's a bit of a perf hit, but likely not a practical problem unless it causes timing issues with side effects in other lifecycle methods.

joschect: This is correct behavior I believe. If selected items are provided, then this is a controlled picker and the incoming `selectedItems` should be the _only_ source of truth.

PR #14500, opened 2020-08-13
react-buttons: Support V7 theme and update default theme to Fluent theme

xugao:
#### Pull request checklist

#### Description of changes
- Changed buttons to have default theme to Fluent (previously was in Teams theme).
- Add Teams and (default) Fluent themes in `react-theme-provider`
- Support v7 theme to work with buttons (conversion logic from v7 theme to tokens, see `getTokens.ts`)
- Various styles fixes for buttons.
- Removed css-var declarations from `.scss` files. This also means from now on `react-buttons` has to be wrapped with `ThemeProvider`

#### Focus areas to test
- [PR deploy site](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/14500/merge/react-button/dist-storybook/index.html) to try different themes using addon
- There are some expected VR test baseline changes.
- Buttons should now work with all the themes I configured in storybook dev:
image


packages/react-theme-provider/etc/react-theme-provider.api.md
```diff
@@ -49,14 +55,11 @@ export const getStyleFromPropsAndOptions: <TProps extends StyleProps<import("./t
 export function mergeThemes<TResult = PartialTheme>(...themes: (undefined | PartialTheme | Theme)[]): TResult;
 
 // @public
-export interface PartialTheme extends RecursivePartial<Theme> {
+export interface PartialTheme extends Omit<Theme, 'tokens'> {
```
xugao, line 58: I removed using `RecursivePartial` here because
1. it's not working correctly when `tokens` is optional: `T[P] extends object` becomes `false`
2. even if i fix `RecursivePartial` to support `T[P] extends object | undefined`, i run into another issue with `stylesheet` typing being incorrect (changed from `string[] | undefined` to `(string | undefined)[] | undefined`.
then i decided to just not use `RecursivePartial`

packages/react-button/src/components/Button/Button.scss
```diff
@@ -164,6 +64,8 @@
   * 3. Avoid defining styles directly.
  **/
   &._primary {
+    // TODO: Check with David
```
xugao, line 67: @dzearing - would need to pick your brain on this one. this would mean another v8 breaking related to theming buttons, but i guess we have to just accept that and document this well?

packages/react-theme-provider/etc/react-theme-provider.api.md
dzearing: yup. I'm running into the same issue.

dzearing:
Perf is hit pretty hard...

xugao:
> Perf is hit pretty hard...

@dzearing - I made some improvements but there are still some added cost in `getTokens`. Let me know if you have ideas to cut that down even more. One thing to note here is - right now it's still passing through many parts of the `V7` theme to tokens (see `passThroughTokens ` in `getTokens`. that will be removed before we ship since we no longer need to use it for other components which converted to `scss`)

packages/react-theme-provider/etc/react-theme-provider.api.md
dzearing: Fixed #14562

packages/react-theme-provider/src/createDefaultTheme.ts
```diff
@@ -1,7 +1,102 @@
...
+import { Theme, Tokens } from './types';
 import { createTheme } from '@uifabric/styling';
+import { getTokens } from './getTokens';
 
 /**
- * Creates a blank initial theme.
+ * Creates default theme (Fluent theme).
  */
-export const createDefaultTheme = (): Theme => createTheme({});
+export const createDefaultTheme = (): Theme => {
```
dzearing, line 8: @fluentui/base-theme

xugao: will address in next pr

packages/react-button/src/components/Button/Button.scss
```diff
@@ -249,8 +149,8 @@
   font-size: var(--button-iconSize, inherit);
   font-weight: normal;
 
-  &._disabled {
-    color: var(--button-disabled-iconColor, var(--button-disabled-contentColor));
+  ._disabled & {
```
khmakoto, line 152: Why are we doing this change?

xugao: it was wrong previously and am fixing it to be correct here;
icon slot doesn't have `_disabled` modifier class name. only root does

PR #14506, opened 2020-08-13
feat(react-image): move to hooks

mnajdova:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/react-image/src/components/Image/Image.tsx
```diff
@@ -1,10 +1,22 @@
...
+
+// Create a hook to resolve classnames.
+export const useImageClasses = makeClasses(classes);
+
+export const Image = React.forwardRef<HTMLElement, ImageProps>((props, ref) => {
+  const { render, state } = useImage(props, ref);
+
+  // Apply styling.
+  useImageClasses(state);
+  useFocusRects(state.imageRef);
```
layershifter, line 17:
```suggestion
```

I don't think that `Image` should have focus behavior. In the spec it's also not defined

packages/react-image/src/components/Image/Image.tsx
```diff
@@ -1,10 +1,22 @@
-import { compose, createClassResolver } from '@fluentui/react-compose';
-
-import { ImageBase } from './ImageBase';
+import * as React from 'react';
 import { ImageProps } from './Image.types';
+import { useFocusRects } from '@uifabric/utilities';
+import { makeClasses } from '@fluentui/react-compose/lib/next/index';
```
layershifter, line 4:
```suggestion
import { makeClasses } from '@fluentui/react-compose/lib/next';
```

?

packages/react-image/src/components/Image/useImage.tsx
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Consts listing which props are shorthand props.
+ */
+export const iconShorthandProps = [];
+
+/**
+ * Define the render function. Given the state of a button, renders it.
+ */
+export const renderImage = (state: ImageState) => {
+  const { slots, slotProps } = getSlots(state, iconShorthandProps);
```
layershifter, line 16: Do we really need this for a component that does not have any slots? 🤔

packages/react-image/src/components/Image/Image.tsx
mnajdova: This is how the import was used in all other components, I would suggest if we decide to change the imports to do it on all components at the same time..

packages/react-image/src/components/Image/useImage.tsx
mnajdova: I believe we still need it as we need the resolved `slots.root`, as well as the `slotProps.root`

packages/react-image/src/components/Image/Image.tsx
dzearing: Background: if we use file path imports without resolving to a physical file, it breaks consumers who use AMD bundlers like r.js which don't automatically infer index.js when refering to a path.

packages/react-image/src/components/Image/useImage.tsx
dzearing: Yep! It boils down to:

* slot.root resolution
* slotProps.root resolution (if primitive, getNativeProps. Otherwise pass through props and assume the component will do the filter.)

PR #14508, opened 2020-08-13
fix modal getting visible shortly when re-rendering while closing

phryneas:
#### Pull request checklist

#### Description of changes

If the parent component triggers a re-render while a Dialog or Modal is in the process of closing, the Modal resets it's internal state to `isVisible: true` in `componentDidUpdate`, which leads to the Dialog getting visible again before finally disappearing (flickering).

This should prevent this behavior.

#### Focus areas to test

This PR includes a storybook story that demonstrates this behavior. Once you have validated the bug & the fix, please ping me and I will remove that story.


joschect (approved):

khmakoto (changes requested):

PR #14513, opened 2020-08-13
Factor out useTarget hook, and make ContextualMenu in react-next StrictMode compliant

### LINKED ISSUE #14338: ContextualMenu: strict mode compliance (UNSAFE_)
ContextualMenu has the following UNSAFE_ methods which prevent strict mode compliance:
- `componentWillUpdate`: handles updating target, hidden/open/closed state, and focus
- `componentWillMount`: handles setting target and previous active element

MLoughry:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14338 

#### Description of changes

This change accomplishes two goals:
1. It refactors out the `Target` type and logic from Callout into a `useTarget` link in `@uifabric/react-hooks`. Code in OUFR and react-next has been modified to point to this export for the `Target` type, as well as the hook in the case of functional components.
2. `ContextualMenu` in `@fluentui/react-next` has been partially migrated to a functional component in order to strip out all the `UNSAFE_` methods that violate StrictMode compliance.

#### Focus areas to test

(optional)


packages/office-ui-fabric-react/src/components/Callout/Callout.types.ts
```diff
@@ -1,14 +1,10 @@
...
 import { DirectionalHint } from '../../common/DirectionalHint';
-import { Point, IRectangle, IStyleFunctionOrObject } from '../../Utilities';
+import { IRectangle, IStyleFunctionOrObject } from '../../Utilities';
 import { ICalloutPositionedInfo } from '../../utilities/positioning';
 import { ILayerProps } from '../../Layer';
-
-/**
- * {@docCategory Callout}
- */
-export type Target = Element | string | MouseEvent | Point | null | React.RefObject<Element>;
```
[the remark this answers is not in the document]
ecraig12345: Yes, we definitely need to continue re-exporting this for now. Re-exporting the version from react-hooks would be fine.

joschect:
Reopening to trigger a build.

packages/react-hooks/src/useTarget.ts
```diff
@@ -0,0 +1,58 @@
...
+  const previousTargetProp = React.useRef<
+    Element | string | MouseEvent | Point | React.RefObject<Element> | null | undefined
+  >();
+
+  const targetRef = React.useRef<Element | MouseEvent | Point | null>(null);
+  /**
+   * Stores an instance of Window, used to check
+   * for server side rendering and if focus was lost.
+   */
+  const targetWindowRef = React.useRef<Window>();
```
ecraig12345, line 26: Should we switch this to use the new `useWindow` hook from `react-window-provider`? I think it was added after this PR started.

packages/react-next/etc/react-next.api.md
```diff
@@ -146,24 +147,26 @@ export class ComboBox extends React.Component<IComboBoxProps, IComboBoxState> {
...
-    constructor(props: IContextualMenuProps);
+export const ContextualMenuBase: {
+    (propsWithoutDefaults: IContextualMenuProps): JSX.Element;
+    displayName: string;
+};
+
+// Warning: (ae-forgotten-export) The symbol "IContextualMenuInternalProps" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export class ContextualMenuInternal extends React.Component<IContextualMenuInternalProps, IContextualMenuState> {
```
ecraig12345, line 158: If this is internal, can/should we stop it from being exported from the root?

packages/react-next/src/components/Callout/Callout.types.ts
```diff
@@ -1,14 +1,10 @@
...
-import { Point, IRectangle, IStyleFunctionOrObject } from '../../Utilities';
+import { IRectangle, IStyleFunctionOrObject } from '../../Utilities';
 import { ICalloutPositionedInfo } from 'office-ui-fabric-react/lib/utilities/positioning';
 import { ILayerProps } from '../../Layer';
-
-/**
- * {@docCategory Callout}
- */
-export type Target = Element | string | MouseEvent | Point | null | React.RefObject<Element>;
+import { Target } from '@uifabric/react-hooks';
```
ecraig12345, line 7: Even for react-next I do wonder if we should continue exporting Target, especially since we don't re-export react-hooks.

packages/react-next/src/components/ContextualMenu/ContextualMenu.base.tsx
```diff
@@ -297,14 +374,15 @@ export class ContextualMenuBase extends React.Component<IContextualMenuProps, IC
     this._adjustedFocusZoneProps = { ...focusZoneProps, direction: this._getFocusZoneDirection() };
 
     const hasCheckmarks = canAnyMenuItemsCheck(items);
-    const submenuProps = this.state.expandedMenuItemKey && this.props.hidden !== true ? this._getSubmenuProps() : null;
+    const submenuProps =
+      this.props.hoisted.expandedMenuItemKey && this.props.hidden !== true ? this._getSubmenuProps() : null;
```
ecraig12345, line 378: Not hugely important, but in methods with multiple references to `this.props.hoisted` it can be beneficial for final bundle size to put the value in a local const (since dotted names can't be minified).

PR #14517, opened 2020-08-13
Adding createMergedRef for class-based components to merge refs

dzearing:
We have a number of cases which require merging refs, but we can't use `useMergedRefs` in a class component.

Adding a utility to help with this scenario to unblock function component conversion edge cases.


packages/utilities/src/createMergedRef.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+};
+
+/**
+ * Helper to merge refs from within class components.
+ */
+export const createMergedRef = <TType, TValue = null>(value?: TValue) => {
+  const local: LocalState<TType, TValue> = ({
+    refs: [],
+    resolver: undefined,
```
khmakoto, line 33: Since you're already casting as `unknown` I think you can remove this line that is setting `resolver` to `undefined`.

PR #14518, opened 2020-08-14
Staging the react-avatar package;

dzearing:
Getting this up and running to unblock @behowell .

This will include the converged `Avatar`. We may need to separate out a separate react-badge package as well, separately of this PR.

Note; this has a converted Image, but really we should change this asap to `react-image`.


yarn.lock
```diff
@@ -116,6 +116,28 @@
     invariant "^2.2.4"
     semver "^5.5.0"
 
+"@babel/core@7.10.5":
```
behowell, line 119: Are all of these yarn.lock changes intended? Or do you need to update dependency versions in package.json?

Issue #9169, opened 2019-05-22
Cards should be using themable elevation rather than depth constants


gggjtgxing:
### Environment Information 
 
- __Package version(s)__: (fill this out) 
- __Browser and OS versions__: (fill this out if relevant)

#### Actual behavior:
 

![image](https://user-images.githubusercontent.com/17696563/58138497-3ba96900-7beb-11e9-8884-0e8eb34fc41c.png)

 
#### Expected behavior:
 

![image](https://user-images.githubusercontent.com/17696563/58138542-85924f00-7beb-11e9-9996-c147e5416daa.png)

#### Priorities and help requested:
 
Are you willing to submit a PR to fix? (Yes, No)
 
Requested priority: (Blocking, High, Normal, Low) 
 
Products/sites affected: (if applicable)
 
I think it's caused by boxShadow color in fluent style 
https://github.com/OfficeDev/office-ui-fabric-react/blob/40487942eabddbb5a7daee45acd7f90a909ad07a/packages/react-cards/src/components/Card/Card.styles.ts

kenotron:
@khmakoto for FYI
@JasonGore gave me a great tip about `<Customizer>` that will allow you to override border:

https://codesandbox.io/s/card-tokens-4cg9n

To note, these are the magical lines for how to make border change for Cards:

```tsx
      <Customizer
        scopedSettings={{
          Card: {
            tokens: {
              boxShadow: "none"
            }
          }
        }}
      >
        <Card>hello</Card>
      </Customizer>
```

kenotron:
Please close this bug if this addresses the issue for you, thanks!

gggjtgxing:
Thanks for your great tip, but is that possible to make Depths.depth4 and other Depths variables themable, otherwise people need to override all boxShadow with Depths variables(as it's not used only for cards), and not sure if override border breaks fluent style.

kenotron:
Unfortunately not at this time - but these are just variables and they're used as such. There's not much of an opportunity to override them other than one component at a time. This is a good feature request though. For reference: 

```tsx
export declare namespace Depths {
    const depth0 = "0 0 0 0 transparent";
    const depth4 = "0 1.6px 3.6px 0 rgba(0, 0, 0, 0.132), 0 0.3px 0.9px 0 rgba(0, 0, 0, 0.108)";
[...]

PR #14519, opened 2020-08-14
Fix Elevation Implementation to use DefaultEffects

dggsax:
The documentation before would reference using a `Depth` class. This is no longer the case, apparently you can access the depths (now referred to as "elevations") in DefaultEffects. So I just put an example and changed the import. Hope this was fine!

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000 (doesn't address an issue, but rather this closed issue #9169 ([link](https://github.com/microsoft/fluentui/issues/9169#issuecomment-494903484)) where I was able to learn that I needed to use `DefaultEffects` as opposed to trying to import/find `Depth`)
- [ ] Include a change request file using `$ yarn change` (ran this command but it didn't generate a change file from my fork)

#### Description of changes

Updated the markdown for the documentations page for "Elevation" ([link](https://developer.microsoft.com/en-us/fluentui#/styles/web/elevation)) to use functionality available from `@fluentui/react`.

#### Focus areas to test

None.

### DIFF
--- a/apps/fabric-website/src/pages/Styles/ElevationPage/docs/web/ElevationImplementation.md
+++ b/apps/fabric-website/src/pages/Styles/ElevationPage/docs/web/ElevationImplementation.md
@@ -1,9 +1,9 @@
 ### Fluent UI React (JavaScript variables)
 
 ```jsx
-import { Depths } from '@uifabric/fluent-theme/lib/fluent/FluentDepths';
+import { DefaultEffects } from '@fluentui/react';
 
-<div style={{ boxShadow: Depths.depth8 }} />;
+<div style={{ boxShadow: DefaultEffects.elevation }} />;
 ```
 
 ### Fabric Core (SCSS variables)


ecraig12345:
Opened #14545 with a more complete fix. Feel free to make suggestions there if you still think it's unclear.

dggsax:
Amazing, thanks!! Will take a look at the new PR. Please let me know if there is a more helpful/useful avenue to utilize in the future for Docs related issues/suggestions/comments! We're starting a new Web UI project for our team so we'll be depending on the docs heavily for the coming months.

ecraig12345:
Creating PRs is fine (and much appreciated!), it's just in this case there happened to be sooo much wrong across multiple similar pages that it seemed better to fix it all-up rather than just fixing this one instance. The other weird thing about this page was the mismatch between "depth" (used on the rest of the page) and "elevation" (used in your suggested change as well as in theming), so switching to use "elevation" here without modifying the rest of the page could have been confusing.

You're also welcome to file issues regarding problems with or suggestions for the docs. I can't guarantee we'll get to it right away, especially requests for new content or other large changes, but we would much rather that someone put those things on our radar (to fix at some point) rather than having them just silently continue to be unhelpful.

dggsax:
@ecraig12345 thank you for the notice! I'm glad I read this, because a co-worker and I got something really cool working with the DetailsList, when I bring the idea and what we did here, I'll be sure to do so as an issue. Looking forward to interacting with y'all more as we're using FluentUI heavily for our current project 😀

PR #14528, opened 2020-08-14
RFC(FocusZone): multiple windows handling for FocusZone

assuncaocharles:
#### Pull request checklist

#### Description of changes

there is a global object used to track all instances of FocusZone - this global object is shared across window instances and can cause unexpected behavior (listener being removed on other window instance than it was added) which can cause issues.

One solution would be adding some `context` for FocusZone and have one `outersFocusZone` per target. Similar to https://github.com/microsoft/fluentui/pull/14010/files

Second solution shown in this PR is to add nested `window => set<FocusZone>`

#### Focus areas to test

(optional)


assuncaocharles:
@dzearing @layershifter @miroslavstastny @mnajdova @pompomon Would love to hear what you have to say on this

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -58,9 +58,9 @@ const ALLOW_VIRTUAL_ELEMENTS = false;
 /**
  * Handle global tab presses so that we can patch tabindexes on the fly.
  */
-function _onKeyDownCapture(ev: KeyboardEvent) {
+function _onKeyDownCapture(this: Window, ev: KeyboardEvent) {
```
layershifter, line 61:
```suggestion
function _onKeyDownCapture(window: Window, ev: KeyboardEvent) {
```
```suggestion
function _onKeyDownCapture(currentWindow: Window, ev: KeyboardEvent) {
```

??

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -156,6 +156,8 @@ export class FocusZone extends React.Component<FocusZoneProps> implements IFocus
 
     // @ts-ignore
     this.windowElement = getWindow(this._root.current);
+    // console.log(this.windowElement);
+    // console.log(_outerZones);
```
layershifter, line 160: Cruft?

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -222,9 +228,9 @@ export class FocusZone extends React.Component<FocusZoneProps> implements IFocus
     delete _allInstances[this._id];
 
     if (!this._isInnerZone) {
-      _outerZones.delete(this);
+      _outerZones.get(this.windowElement!)?.delete(this);
```
layershifter, line 231: These should be an additional check to remove entries from `_outerZones`

layershifter:
If we don't want to have a React context - this is a single way. I suggest to extract this code to a better abstraction that will allow to avoid checks inside the component itself:

```tsx
      if (_outerZones.get(this.windowElement!)) {
        _outerZones.get(this.windowElement!)?.add(this);
      } else {
        _outerZones.set(this.windowElement!, new Set([this]));
      }
```
**vs.**
```tsx
outZones.register(this.windowElement, this)
```

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
assuncaocharles: Removed

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
assuncaocharles: Actually if we rename that we would have to pass it explicit when add it as listener

assuncaocharles:
> If we don't want to have a React context - this is a single way. I suggest to extract this code to a better abstraction that will allow to avoid checks inside the component itself:
> 
> ```tsx
>       if (_outerZones.get(this.windowElement!)) {
>         _outerZones.get(this.windowElement!)?.add(this);
>       } else {
>         _outerZones.set(this.windowElement!, new Set([this]));
>       }
> ```
> 
> **vs.**
> 
> ```tsx
> outZones.register(this.windowElement, this)
> ```

Created an object with `register`, `unregister` and `getOutzones` check if it's what you meant.

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
assuncaocharles: Which check?

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -58,9 +76,9 @@ const ALLOW_VIRTUAL_ELEMENTS = false;
 /**
  * Handle global tab presses so that we can patch tabindexes on the fly.
  */
-function _onKeyDownCapture(ev: KeyboardEvent) {
+function _onKeyDownCapture(this: Window, ev: KeyboardEvent) {
```
dzearing, line 79: Please don't call a function argument `this`, `window`, or any other keyword that means something special in javascript. It can throw people off.

assuncaocharles: In this case is needed as it is part of the signature for the listener 

```typescript
listener: (this: Window, ev: WindowEventMap[K]) => any
```

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
dzearing: not sure i followed that.

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -169,9 +184,9 @@ export class FocusZone extends React.Component<FocusZoneProps> implements IFocus
     }
 
     if (!this._isInnerZone) {
```
dzearing, line 186: probably should move the windowElement null check to 186 and prevent registration.

dzearing: Another thought; if `register` returned the count of active outZone windows, you could avoid having a "getOutZone" method altogether I think.

dzearing:
Overall this makes sense, but I'm wondering when we're going to just have 1 focus zone so that bugs like this don't have to be fixed in 2 places. There is another one in `@fluentui/react-focus`; why isn't v0 using that one? And if there is some good reason, we should fix it there too.

@khmakoto @jurokapsiar

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
assuncaocharles: Done

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
dzearing: right but having the variables called "foo" doesn't make this behave differently.

dzearing:
```jsx
const outZones = {
  _outerZones: new Map<Window, Set<FocusZone>>(),
  register(foo: Window, FZ: FocusZone) {
    if (this._outerZones.get(foo)) {
      this._outerZones.get(foo)?.add(FZ);
    } else {
      this._outerZones.set(foo, new Set([FZ]));
    }
  },
  unregister(foo: Window, FZ: FocusZone) {
    this._outerZones.get(foo)?.delete(FZ);
  },
  deleteOutZone(foo: Window) {
    this._outerZones.delete(foo);
  },
  getOutZone(foo: Window) {
    return this._outerZones.get(foo);
  },
};
```

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -45,7 +45,25 @@ const _allInstances: {
   [key: string]: FocusZone;
 } = {};
 
-const _outerZones: Set<FocusZone> = new Set();
+const outZones = {
+  _outerZones: new Map<Window, Set<FocusZone>>(),
```
dzearing, line 49: nit on naming: rename `_outerZones` to `_windowToOuterZoneMap` or something

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -45,7 +45,25 @@ const _allInstances: {
   [key: string]: FocusZone;
 } = {};
 
-const _outerZones: Set<FocusZone> = new Set();
+const outZones = {
```
dzearing, line 48: Suggestion: rename to `outerZones` so that we don't have outzones and outerzones.

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
assuncaocharles: Actually it does, if we "rename" the `this` we get some errors when adding the listener, trying with `currentWindow` as Shift suggested for example:

Screenshot 2020-08-18 at 16 54 38

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -220,12 +235,12 @@ export class FocusZone extends React.Component<FocusZoneProps> implements IFocus
...
     delete _allInstances[this._id];
+    outZones.unregister(this.windowElement!, this);
 
     if (!this._isInnerZone) {
-      _outerZones.delete(this);
-
-      if (this.windowElement && _outerZones.size === 0) {
+      if (this.windowElement && outZones.getOutZone(this.windowElement)?.size === 0) {
         this.windowElement.removeEventListener('keydown', _onKeyDownCapture, true);
+        outZones.deleteOutZone(this.windowElement!);
```
dzearing, line 243: I'm wondering why the outerZone unregister doesn't do this.

dzearing: Recommend moving this to last unregistration and removing the api. (Because it does this work on first registration, but on last unregistration it doesn't. This is asymmetrical.)

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
assuncaocharles: Done

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -45,7 +45,25 @@ const _allInstances: {
...
+  },
+  unregister(window: Window, FZ: FocusZone) {
+    this._outerZones.get(window)?.delete(FZ);
+  },
+  deleteOutZone(window: Window) {
+    this._outerZones.delete(window);
+  },
+  getOutZone(window: Window) {
+    return this._outerZones.get(window);
+  },
```
dzearing, line 65: you can remove these 2 apis. And make sure unregister deletes the entry in the map when the last entry for the window is unregistered.

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
assuncaocharles: Done

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -45,7 +45,25 @@ const _allInstances: {
   [key: string]: FocusZone;
 } = {};
 
-const _outerZones: Set<FocusZone> = new Set();
+const outZones = {
+  _outerZones: new Map<Window, Set<FocusZone>>(),
+  register(window: Window, FZ: FocusZone) {
```
dzearing, line 50: Have register return a count of active outer zones. This can be used by the caller to know if this is the first outer zone registered.

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
assuncaocharles: removed `deleteOutZone` but `getOutZone` is only for tests usage

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
assuncaocharles: Done

packages/fluentui/react-bindings/src/FocusZone/FocusZone.tsx
dzearing: oh I see what youre saying, but let me suggest a better alternative. Don't take this argument in. Leave the signature intact because that's the expected event hander arg (ev first).

To get the window, use the `getWindow` helper based on the ev.target element. I'm not sure if there is one in react-bindings, but there is one in the utilities package that you can look at. You basically need to derive the correct window based on the target that the event was triggered from.

dzearing: As the code stands right now, I believe it's compiling by accident. If you break into this code, you'll probably find the first argument (`this`) is actually the event args.

dzearing: I might be missing something though!

assuncaocharles: Great idea, thanks @dzearing 

Solved with `getWindow`

```javascript
outerZones.getOutZone(getWindow(ev.target as Element)!)?.forEach(zone => zone.updateTabIndexes());
```

PR #14529, opened 2020-08-14
Make Dropdown React.StrictMode compliant in react-next

### LINKED ISSUE #14289: Dropdown: strict mode compliance (UNSAFE_)
Dropdown does the following in `componentWillReceiveProps` which prevents strict mode compliance:
- Updates `state.selectedIndices` if `props.options` or the selected key prop changes. This is also be updated elsewhere in the class (a variant of the incorrect semi-controlled pattern).
  - It also incorrectly respects updates to `props.defaultSelectedKey(s)`--updates to defaults should be ignored
- Updates `this._sizePosCache` (a cached list of visible item positions and set size for aria) based on `props.options`

It's likely that this would be easier to fix in-place (likely by moving to `componentDidUpdate`) rather than doing a function component conversion due to the overall complexity of the component.

MLoughry:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14289 

#### Description of changes

Partially migrate `Dropdown` to a functional component in order to remove the `UNSAFE_` method and make the component React.StrictMode compliant.

#### Focus areas to test

(optional)


MLoughry:
When I try to update the API, I'm getting strange warnings that cause the update to fail:
```
[4:08:19 PM] ■ Extracting Public API surface from '/home/miclo/git/fluentui/packages/react-next/lib/index.d.ts'
Warning: /home/miclo/git/fluentui/packages/office-ui-fabric-react/src/components/Pivot/Pivot.base.tsx:356:64 - (TS2339) [object Object]
Warning: /home/miclo/git/fluentui/packages/office-ui-fabric-react/src/components/Pivot/Pivot.base.tsx:356:83 - (TS2339) Property 'name' does not exist on type 'typeof PivotItem'.
```

@ecraig12345, it looks like you added that Pivot code. Do you have any idea what's going on?

ecraig12345:
I'm not that familiar with Pivot--since I've run prettier, switched linters, and made other extensive changes, my name being in the file history may not mean much. In general though, my first guess is that this somehow has to do with the fact that we currently have two versions of Pivot referenced in react-next: the OUFR one, and the one from `react-tabs`. But TS itself should be able to clearly distinguish them, so erroring on name overlap would have to be some kind of weird bug in API Extractor? Second guess is some kind of React typing mismatch. If you're still stuck on this I can try and repro it locally to debug.

MLoughry:
Ugh... I worked around the API extractor issue, but now the `codesandbox-react-next-template` is failing with a bunch of spurious errors.

apps/codesandbox-react-next-template/tsconfig.json
```diff
@@ -11,10 +11,13 @@
     "forceConsistentCasingInFileNames": true,
     "moduleResolution": "node",
     "preserveConstEnums": true,
-    "strict": true,
+    "strictNullChecks": true,
```
ecraig12345, line 14: I'm curious, what issues were you hitting with `strict`?

packages/office-ui-fabric-react/src/components/Pivot/Pivot.base.tsx
```diff
@@ -353,6 +353,6 @@ function _isPivotItem(item: React.ReactNode): item is PivotItem {
     !!item &&
     typeof item === 'object' &&
     !!(item as React.ReactElement).type &&
-    ((item as React.ReactElement).type as React.ComponentType).name === PivotItem.name
+    ((item as React.ReactElement).type as any).name === (PivotItem as any).name
```
ecraig12345, line 356: It's so strange that this is where the weird typing issue was popping up, when there's absolutely no connection to Pivot as far as I can tell... Also kind of surprised the issue didn't just move somewhere else. But if this works around it... 🤷‍♀️

apps/codesandbox-react-next-template/tsconfig.json
MLoughry: https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=112673&view=logs&j=258ec178-2d8b-5611-7b9b-60c5c95dae55&t=e99f753c-3f64-5289-3575-c1a650af1617

ecraig12345: Ah okay, I forgot that `strict` implies `strictFunctionTypes`

MLoughry: The underlying issue is that this tsconfig's strictness should be in line with the underlying projects it's building, rather than stricter. Otherwise, the source project builds fine, and you get these spurious errors downstream.

dzearing:
This is passing, any concerns blocking this from merging @ecraig12345?

ecraig12345:
Jon and I still need to review (this is too big of a change to just rely on our limited test coverage). I'll try and take a look today.

packages/react-next/src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -92,18 +93,21 @@ describe('Dropdown', () => {
...
+      safeCreate(
         <Dropdown
           label="testgroup"
           options={[
             { key: '1', text: '1', selected: true },
             { key: '2', text: '2' },
           ]}
         />,
+        container => {
+          const dropdownOptionText = container.root.find(node => {
```
ecraig12345, line 105: I know you mentioned this in the past but remind me why doing this with enzyme didn't work? It seems kind of unfortunate to switch the tests over to react-test-renderer since the syntax is so much more verbose and doing updates is more error-prone (since you have to manually ensure all the same props are passed except for the one you meant to change).

packages/react-next/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -39,30 +39,155 @@ import { IProcessedStyleSet } from '../../Styling';
...
-export interface IDropdownInternalProps extends IDropdownProps, IWithResponsiveModeState {}
+export interface IDropdownInternalProps extends IDropdownProps, IWithResponsiveModeState {
+  hoisted: {
+    rootRef: React.Ref<HTMLDivElement>;
+    selectedIndices: number[];
+    setSelectedIndices: React.Dispatch<React.SetStateAction<number[]>>;
+  };
+}
 
 export interface IDropdownState {
```
ecraig12345, line 65: Should probably stop exporting this too if possible, since I don't think it's used by anything that needs to be exported.

packages/react-next/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -39,30 +39,155 @@ import { IProcessedStyleSet } from '../../Styling';
...
+    } else {
+      if (didOptionsChange && defaultSelectedKey !== undefined) {
+        selectedKeyPropToUse = defaultSelectedKey;
+      } else {
+        selectedKeyPropToUse = selectedKey;
+      }
+    }
+
+    if (
+      (selectedKeyPropToUse !== undefined || !oldOptions) &&
```
ecraig12345, line 150: Just to clarify, the new `|| !oldOptions` is to handle the old constructor logic, right?

packages/react-next/src/components/Dropdown/Dropdown.base.tsx
```diff
@@ -404,14 +490,8 @@ export class DropdownBase extends React.Component<IDropdownInternalProps, IDropd
...
-    this.setState(
-      {
-        selectedIndices: newIndexes,
-      },
-      () => {
-        this._onChange(event, options, index, checked, multiSelect);
-      },
-    );
+    this.props.hoisted.setSelectedIndices(newIndexes);
+    this._onChange(event, options, index, checked, multiSelect);
```
ecraig12345, line 494: Calling `onChange` immediately after `setSelectedIndices` (instead of in the post-`setState` callback) is a subtle behavior change, but I'm not sure how to tell if it will matter for consumers.

ecraig12345 (commented):
Looked over the logic and it mostly looks good, but I have a few minor questions and suggestions.

packages/react-next/src/components/Dropdown/Dropdown.base.tsx
MLoughry: Yes

packages/react-next/src/components/Dropdown/Dropdown.test.tsx
MLoughry: In this case, I'm not sure why I changed it. I've reverted it since it works the old way.

The other cases below were using internal state, which isn't exposed on functional component. Additionally, I think hooks are not run (or run consistently) when using enzyme.

ecraig12345: You should be able to get hooks to work by wrapping certain operations in `act()` (from `react-dom/test-utils`), though if I remember right it can still be a bit challenging to get certain things working.

packages/react-next/src/components/Dropdown/Dropdown.base.tsx
joschect: I believe that this is okay. onChange doesn't really rely on anything that would happen from the newIndexes so it should be alright.

joschect (approved):

ecraig12345 (approved):

ecraig12345:
Random build timeout! 🙄 Trying again...

ecraig12345:
Size auditor is showing a 9k increase to bundle size of Dropdown, which I think is compared to when it was re-exporting from OUFR. Any idea why that would be?

> ## **Asset size changes**
> Project	Bundle	Baseline Size	New Size	Difference
> office-ui-fabric-react	[fluentui-react-next-Dropdown](https://microsoft.github.io/sizeAuditor-website/?baseline=123044&pr=123201&artifactFileName=fluentui-react-next-Dropdown.stats.json)	224.289 kB	234.131 kB	ExceedsTolerance     9.842 kB
> ExceedsTolerance Over Tolerance (1024 B) ExceedsBaseline Over Baseline BelowBaseline Below Baseline New New Deleted  Removed **1 kB** = 1000 B
> 
> Baseline commit: [764d61e447535825edd612dfa8927b66fddd1d3f](https://github.com/OfficeDev/office-ui-fabric-react/commit/764d61e447535825edd612dfa8927b66fddd1d3f) [(build)](https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=123201)

MLoughry:
> Size auditor is showing a 9k increase to bundle size of Dropdown, which I think is compared to when it was re-exporting from OUFR. Any idea why that would be?

It's importing the React hooks, without really saving code space on getting rid of class templating. Without a more detailed breakdown, I can't speculate beyond that.

ecraig12345:
That's unfortunate but makes sense. There's a [link to a report](https://microsoft.github.io/sizeAuditor-website/?baseline=123044&pr=123201&artifactFileName=fluentui-react-next-Dropdown.stats.json) which should theoretically show more detail, but the report page has a script error right now.

If you're correct that the increase is from the hooks, I'm guessing in practice that will only be a one-time impact for the whole app (not per component) since the hooks can be shared across components, so I think it's okay.

I am a bit concerned about the build timeout though...it's now timed out twice on the same step with no logs shown. https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=123202&view=logs&j=258ec178-2d8b-5611-7b9b-60c5c95dae55&t=e99f753c-3f64-5289-3575-c1a650af1617

ecraig12345:
I finally remembered to watch the log as the build was going. It's hard to tell for sure with `lage` grouped build output, but I think it's getting stuck on `build` for `react-next`. @MLoughry what happens if you merge with master and try a full build locally? (You may need to do `yarn scrub` first to ensure a clean state like the build machines have.)

ecraig12345:
Never mind, the react-next build succeeded, so it must be stuck on something else.

MLoughry:
Calling `yarn buildci` locally builds without errors until it hits an error about missing the Chrome headless browser (which I don't think is the issue on the PR build)

MLoughry:
The last lines in the PR build log as it's building are below, which is ~5m after the start of the build step. The build step has been running for >40m now.

```
info ----------------------------------------------
info ▶️ start perf-test lint
verb |  Running /mnt/work/_tool/node/10.22.0/x64/bin/npm run lint
verb |   The node binary used for scripts is /tmp/yarn--1599842461957-0.007489414011746076/node but npm is using /mnt/work/_tool/node/10.22.0/x64/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
verb |  > perf-test@7.0.0 lint /mnt/work/4/s/apps/perf-test
verb |  > just-scripts lint
verb |  [4:47:17 PM] ■ started 'lint'
verb |  [4:47:17 PM] ■ started 'lint-imports'
verb |  [4:47:17 PM] ■ started 'eslint'
verb |  [4:47:17 PM] ■ "/mnt/work/4/s/node_modules/eslint/bin/eslint.js" /mnt/work/4/s/apps/perf-test/src --ext ".ts,.tsx,.js,.jsx" --config "/mnt/work/4/s/apps/perf-test/.eslintrc.json" --cache --color
verb |  [4:47:17 PM] ■ finished 'lint-imports' in 0.07s
verb |  [4:47:27 PM] ■ finished 'eslint' in 9.63s
verb |  [4:47:27 PM] ■ finished 'lint' in 9.64s
info ✔️ done perf-test lint - 12.37s
info ----------------------------------------------
```

PR #14536, opened 2020-08-14
Codemods: Add support for repath mod from json

tmaster628:
Added a simple repath mod to the json general config codemod.

This brought about a few other changes:

-Introduced total typing into the json object to remove those pesky `any` types floating around.
-Instead of using `if` casing to determine what mod to run, I created a dictionary that maps codemod types to functions.
-Changed the repathImport utility _very slightly_ to support regexes that could come from json (because strings aren't very flexible).
-Added some more error throwing in propHelpers because codemods can and should be able to throw detailed errors when mods fail instead of silently doing nothing.

### DIFF
--- a/packages/codemods/src/codeMods/utilities/imports.ts
+++ b/packages/codemods/src/codeMods/utilities/imports.ts
@@ -22,20 +22,15 @@ export function renameImport(file: SourceFile, originalImport: string, renamedIm
  */
 export function getImportsByPath(file: SourceFile, pathOrRegex: string | RegExp): ImportDeclaration[] {
   let imps: ImportDeclaration[] = [];
-  if (typeof pathOrRegex === 'string' && !stringIsRegex(pathOrRegex)) {
+  if (typeof pathOrRegex === 'string') {
     imps = file.getImportDeclarations().filter(cond => {
       return cond.getModuleSpecifierValue() === pathOrRegex;
     });
   } else {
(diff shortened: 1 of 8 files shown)


packages/codemods/src/codeMods/utilities/imports.ts
```diff
@@ -22,13 +22,17 @@ export function renameImport(file: SourceFile, originalImport: string, renamedIm
  */
 export function getImportsByPath(file: SourceFile, pathOrRegex: string | RegExp): ImportDeclaration[] {
   let imps: ImportDeclaration[] = [];
-  if (typeof pathOrRegex === 'string') {
+  if (typeof pathOrRegex === 'string' && !stringIsRegex(pathOrRegex)) {
     imps = file.getImportDeclarations().filter(cond => {
       return cond.getModuleSpecifierValue() === pathOrRegex;
     });
   } else {
+    if (typeof pathOrRegex === 'string' && stringIsRegex(pathOrRegex)) {
```
joschect, line 30: After thinking about this a bit, I think that we shouldn't convert the string to regex. At least not yet. Instead there should be another option in the json that is just regex so that we can be sure it's always parsed in the right way.

tmaster628: That makes sense -- I didn't like how clunky this felt when I was implementing it.

PR #14542, opened 2020-08-14
Add back UHF link style overrides via Customizer

ecraig12345:
#### Pull request checklist

#### Description of changes

#14270 removed some global link style overrides because they were messing up examples in dark mode, but thanks to the horrible UHF global styles, this caused links to be rendered as text color (not blue). Adding back the overrides in Page but doing it via Customizer will hopefully fix this.

### DIFF
--- a/packages/example-app-base/src/utilities/getLinkColors.ts
+++ b/packages/example-app-base/src/utilities/getLinkColors.ts
@@ -1,0 +1,30 @@
+import { IRawStyle } from 'office-ui-fabric-react';
+
+/**
+ * Gets a style which should be specific enough to override the UHF link styles.
+ * This should be applied to the root of a Link.
+ * @param rest - Color to use in default/rest state
+ * @param hover - Color to use in hover/active state
+ * @param focus - Color to use in focus state (defaults to `rest`)
+ */
+export function getLinkColors(rest: string, hover: string, focus?: string): IRawStyle {
+  return {
+    selectors: {
+      // Be specific to override UHF styles
+      '& a.ms-Link': {
+        color: rest,
+        selectors: {
+          '&:link': {
+            color: rest,
+          },
+          '&:active, &:hover, &:active:hover': {
+            color: hover,
+          },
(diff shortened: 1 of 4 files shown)


packages/example-app-base/src/utilities/getLinkColors.ts
```diff
@@ -0,0 +1,30 @@
+import { IRawStyle } from 'office-ui-fabric-react';
+
+/**
+ * Gets a style which should be specific enough to override the UHF link styles.
+ * This should be applied to the root of a Link.
+ * @param rest - Color to use in default/rest state
+ * @param hover - Color to use in hover/active state
+ * @param focus - Color to use in focus state (defaults to `rest`)
+ */
+export function getLinkColors(rest: string, hover: string, focus?: string): IRawStyle {
```
dzearing, line 10: is this really the right place to fix this? Shouldn't this be done on uhf overrides on the website?

dzearing: The purpose of example-app-base is to create a base set of components for examples; it really has nothing to do with UHF. In addition to the website, M365 and Ribbon examples use it to render example pages. They don't care about UHF overrides, as we may not if we ever move away from the platform.

ecraig12345: This is effectively the same place the UHF overrides were until Mary removed them in #14270, so putting them back here is no worse than it was before. The only difference is previously they were in Page.module.scss (and therefore globally applied regardless of the theme--the problem Mary was trying to solve) and now they're being applied here via Customizer (therefore overrideable by theme).

I don't want to put them at a higher scope such as fabric-website (or waste time adding a finer-grained customization mechanism) because they should only apply to `Link`s within the `Page` component--otherwise you have to do even more weird re-overrides like the one in SideRail.

dzearing: ah ok.

PR #14544, opened 2020-08-14
Updating SwatchColorPicker to a function component inside React-Next

### LINKED ISSUE #14318: SwatchColorPicker: strict mode compliance
SwatchColorPicker has a `componentWillReceiveProps` implementation which maps `props.selectedId` to `state.selectedIndex` (an instance of the incorrect semi-controlled pattern). This could be fixed in place by moving the logic to `componentDidUpdate`, or by conversion to a function component.

If converting to a function component, this would also be a good opportunity to fix the controlled behavior (see #10516).

czearing:
#### Pull request checklist
- [x] Fixes #14318

#### Description of changes
1. Updating `SwatchColorPIcker` to a function component. 
2. Added an onChange prop and deprecated onColorChanged.
3. Added defaultSelectedId  and deprecated isControlled.
4. Removed selectedIndex and replaced it with selectedId.
5. Updated `ColorPickerGridCell` to a function component.
6. Changed `ColorPickerGridCell` color prop to not handle undefined type.
6. Added perf and vr tests for `SwatchColorPIcker` and `SwatchColorPIckerNext`
7. Added breaking changes to RELEASE_NOTES.
8. Updated `GridCell` and `GridBase` to a function component to allow ref to function correctly in `SwatchColorPicker`.
9. Implemented createMergedRef inside `FocusZone` to allow ref to function correctly in `Grid`.
10. Added elementRef type to FocusZone.types.

### Focus areas to test
`SwatchColorPIcker`, `Grid`


packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.types.ts
```diff
@@ -45,24 +45,30 @@ export interface ISwatchColorPickerProps {
...
+  onColorChanged?: (id?: string, color?: string) => void;
 
   /**
    * Callback for when the user changes the color.
    * If `id` and `color` are unspecified, there is no selected cell.
    * (e.g. the user executed the currently selected cell to unselect it)
    */
-  onColorChanged?: (id?: string, color?: string) => void;
-
+  onChange?: (event: React.MouseEvent<HTMLElement>, color?: string) => void;
```
dzearing, line 71: This might need to be just `React.Event`, as it could be mouse or keyboard

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
+  const { setTimeout, clearTimeout } = useSetTimeout();
+
+  useDebugWarnings(props);
+
+  const {
+    colorCells,
+    cellShape = 'circle',
+    columnCount,
+    // eslint-disable-next-line deprecation/deprecation
+    ariaPosInSet = props.positionInSet,
```
ecraig12345, line 57: We can probably just remove `positionInSet` and `setSize` from props. They were deprecated 8 months ago, which was after v7 released, but usage should be pretty rare.

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
-    } = this.props;
-
-    const classNames = getClassNames(styles!, {
-      theme: this.props.theme!,
-      className,
-      cellMargin,
+  /**
+   *  Add an index to each color cells. Memoizes this so that color cells do not re-render on every update.
+   */
+  const getItemsWithIndex = React.useMemo(() => {
```
ecraig12345, line 74: This should just be called `itemsWithIndex` since it's a value not a function.

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
   /**
    * When the whole swatchColorPicker is blurred,
    * make sure to clear the pending focused stated
    */
-  private _onSwatchColorPickerBlur = (): void => {
-    if (this.props.onCellFocused) {
-      this._cellFocused = false;
-      this.props.onCellFocused();
+  const onSwatchColorPickerBlur = React.useCallback((): void => {
+    if (props.onCellFocused) {
```
ecraig12345, line 99: Add this to the initial props destructuring so you only need a dep on `onCellFocused` itself, not the whole props object

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
+export const SwatchColorPickerBase = React.forwardRef<HTMLElement, ISwatchColorPickerProps>((props, ref) => {
+  const defaultId = useId('swatchColorPicker');
+  const id = props.id || defaultId;
+
+  const internalState = useConst<ISwatchColorPickerState>(() => ({
+    isNavigationIdle: true,
+    cellFocused: false,
+    navigationIdleTimeoutId: undefined,
+    navigationIdleDelay: 250,
+  }));
```
ecraig12345, line 46: Small thing but you don't need a wrapping function here.
```suggestion
  const internalState = useConst<ISwatchColorPickerState>({
    isNavigationIdle: true,
    cellFocused: false,
    navigationIdleTimeoutId: undefined,
    navigationIdleDelay: 250,
  });
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
-
-    if (this.isNavigationIdle && !this.props.disabled) {
-      ev.currentTarget.focus();
-    }
-
-    return true;
-  };
+  const onMouseEnter = React.useCallback(
+    (ev: React.MouseEvent<HTMLButtonElement>): boolean => {
+      if (!props.focusOnHover) {
```
ecraig12345, line 110: Same, destructure this and `disabled` earlier

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
+        return !internalState.isNavigationIdle || !!props.disabled;
+      }
+      if (internalState.isNavigationIdle && !props.disabled) {
+        ev.currentTarget.focus();
+      }
+      return true;
+    },
+    [props.focusOnHover, internalState.isNavigationIdle, props.disabled],
```
ecraig12345, line 118: Here you should just depend on `internalState` itself, not the navigation idle value. The final dep list once you fix the destructuring will be like this.
```suggestion
    [focusOnHover, internalState, disabled],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
   /**
    * Callback passed to the GridCell that will manage Hover/Focus updates
    */
-  private _onMouseMove = (ev: React.MouseEvent<HTMLButtonElement>): boolean => {
-    if (!this.props.focusOnHover) {
-      return !this.isNavigationIdle || !!this.props.disabled;
-    }
+  const onMouseMove = React.useCallback(
+    (ev: React.MouseEvent<HTMLButtonElement>): boolean => {
+      if (!props.focusOnHover) {
```
ecraig12345, line 126: Same comment on desctructuring from props

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
+      // if (!this.isNavigationIdle || (document && targetElement === (document.activeElement as HTMLElement))) {
+      if (internalState.isNavigationIdle && !(document && targetElement === (document.activeElement as HTMLElement))) {
+        targetElement.focus();
+      }
 
-    return true;
-  };
+      return true;
+    },
+    [props.focusOnHover, internalState.isNavigationIdle, props.disabled],
```
ecraig12345, line 140: Similar comment about deps
```suggestion
    [focusOnHover, internalState, disabled],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
-  private _onMouseLeave = (ev: React.MouseEvent<HTMLButtonElement>): void => {
-    const parentSelector = this.props.mouseLeaveParentSelector;
+  const onMouseLeave = React.useCallback(
+    (ev: React.MouseEvent<HTMLButtonElement>): void => {
+      const parentSelector = props.mouseLeaveParentSelector;
 
-    if (!this.props.focusOnHover || !parentSelector || !this.isNavigationIdle || this.props.disabled) {
-      return;
-    }
+      if (!props.focusOnHover || !parentSelector || !internalState.isNavigationIdle || props.disabled) {
```
ecraig12345, line 150: same

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
+            (elements[index] as HTMLElement).focus();
           }
-        } else {
-          (elements[index] as HTMLElement).focus();
+
+          break;
         }
+      }
+    },
+    [internalState.isNavigationIdle, props.disabled, props.focusOnHover, props.mouseLeaveParentSelector],
```
ecraig12345, line 181:
```suggestion
    [internalState, disabled, focusOnHover, mouseLeaveParentSelector],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
   /**
-   * Callback to make sure we don't update the hovered element during mouse wheel
+   * Callback passed to the GridCell class that will trigger the onCellFocus callback of the SwatchColorPicker
    */
-  private _onWheel = (): void => {
-    this._setNavigationTimeout();
-  };
+  const onGridCellFocused = React.useCallback(
+    (item?: IColorCellProps): void => {
+      const { onCellFocused } = props;
```
ecraig12345, line 202: same

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
-      ev.which === KeyCodes.down ||
-      ev.which === KeyCodes.left ||
-      ev.which === KeyCodes.right
-    ) {
-      this._setNavigationTimeout();
-    }
-  };
+  const onCellClick = React.useCallback(
+    (item: IColorCellProps): void => {
+      if (props.disabled) {
```
ecraig12345, line 221: same

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
+   * Callback to make sure we don't update the hovered element during mouse wheel
    */
-  private _onGridCellHovered = (item?: IColorCellProps): void => {
-    const { onCellHovered } = this.props;
-
-    if (onCellHovered) {
-      return item ? onCellHovered(item.id, item.color) : onCellHovered();
-    }
-  };
+  const onWheel = React.useCallback((): void => {
```
ecraig12345, line 256: You could probably just pass `setNavigationTimeout` directly as `onWheel`

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
...
-        this.props.onColorChanged(item.id, item.color);
-      }
+  const renderOption = React.useCallback(
+    (item: IColorCellProps): JSX.Element => {
+      return (
+        <ColorPickerGridCell
+          item={item}
+          idPrefix={id}
+          color={item.color}
+          styles={props.getColorGridCellStyles}
```
ecraig12345, line 290: Same about destructuring. Though in this case it might be better to just *not* use `useCallback` and let it re-generate every time, since that will likely happen anyway due to the number of deps... Also since looking at the code of Grid, nothing else is memoized based on this function.

packages/react-next/src/utilities/grid/Grid.base.tsx
```diff
@@ -1,96 +1,79 @@
...
-                })}
-              </tr>
-            );
-          })}
-        </tbody>
-      </table>
-    );
+  const content = (
+    <table
+      ref={doNotContainWithinFocusZone ? (ref as React.Ref<HTMLTableElement>) : undefined}
```
ecraig12345, line 40: Not seeing how/why this ref is used?

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,341 @@ import {
 import { Grid } from '../../utilities/grid/Grid';
 import { IColorCellProps } from './ColorPickerGridCell.types';
 import { ColorPickerGridCell } from './ColorPickerGridCell';
-import { memoizeFunction, warnDeprecations } from '@uifabric/utilities';
+import { useId, useConst, useSetTimeout, useControllableValue, useWarnings } from '@uifabric/react-hooks';
+import { IGridProps } from '../../utilities/grid/Grid.types';
 
 export interface ISwatchColorPickerState {
```
ecraig12345, line 14: This shouldn't be exported now.
```suggestion
interface ISwatchColorPickerState {
```

packages/react-focus/src/components/FocusZone/FocusZone.types.ts
```diff
@@ -49,6 +49,11 @@ export interface IFocusZoneProps extends React.HTMLAttributes<HTMLElement | Focu
    */
   componentRef?: IRefObject<IFocusZone>;
 
+  /**
+   * Optional callback to access the root DOM element.
+   */
+  elementRef?: React.Ref<HTMLElement>;
```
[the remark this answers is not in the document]
MLoughry: Ultimately, I'd prefer to refactor into a ref-forwarding functional component and simply use `ref`. In the intermediate term, you can export a ref-forwarding functional component *wrapper* around the current `FocusZone`, and pass the ref to the current (now internal) class-based component with an added, undocumented `elementRef` prop.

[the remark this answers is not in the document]
dzearing: Agreed that the correct option would be `ref`; I think not documenting, or even immediately deprecating, `elementRef` to indicate it's a temporary internal mechanism makes a lot of sense.

Let's see if we can do the wrapper. Here's a codesandbox to mock it up:

https://codesandbox.io/s/misty-cookies-29731?file=/src/App.tsx:668-671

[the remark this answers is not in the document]
czearing: @dzearing and I tried to implement this but it was decided it would be to breaking to address in this PR.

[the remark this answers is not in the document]
ecraig12345: Do we want to go ahead and mark `elementRef` is deprecated since it seems to be a temporary solution?

packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
```diff
@@ -1,157 +1,135 @@
 import * as React from 'react';
 import { ITheme, mergeStyleSets } from '../../Styling';
-import { classNamesFunction, memoizeFunction } from '../../Utilities';
-import { getColorFromString } from 'office-ui-fabric-react/lib/utilities/color/getColorFromString';
-import { GridCell, IGridCellProps } from '../../utilities/grid/index';
+import { classNamesFunction } from '../../Utilities';
+import { getColorFromString } from 'office-ui-fabric-react/lib/Color';
+import { GridCell } from '../../utilities/grid/GridCell';
```
ecraig12345, line 5: Guessing you're planning to do the ButtonGrid rename here in a subsequent PR? (that's fine)

packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
```diff
@@ -1,157 +1,135 @@
...
 import { getStyles as getActionButtonStyles } from 'office-ui-fabric-react/lib/components/Button/ActionButton/ActionButton.styles';
 import { IButtonClassNames } from 'office-ui-fabric-react/lib/components/Button/BaseButton.classNames';
+import { IProcessedStyleSet } from 'office-ui-fabric-react/lib/Styling';
```
ecraig12345, line 8: Merge this with the ../../Styling import above

packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
```diff
@@ -1,157 +1,134 @@
...
-      <svg
-        className={this._classNames.svg}
-        viewBox="0 0 20 20"
-        fill={getColorFromString(colorOption.color as string)!.str}
-      >
-        {this.props.circle ? <circle cx="50%" cy="50%" r="50%" /> : <rect width="100%" height="100%" />}
-      </svg>
-    );
-  };
+  const getColorPickerGridCellButtonClassNames = React.useCallback(
```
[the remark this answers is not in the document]
ecraig12345: This still needs to be addressed.

packages/react-next/src/utilities/grid/Grid.base.tsx
```diff
@@ -1,96 +1,77 @@
...
+      {...htmlProps}
+      className={classNames.root}
+    >
+      <tbody>
+        {rowsOfItems.map((rows: [], rowIndex: number) => {
+          return (
+            <tr role={'row'} key={rowIndex + '-row'}>
+              {rows.map((cell, cellIndex: number) => {
+                return (
+                  <td role={'presentation'} key={cellIndex + '-cell'} className={classNames.tableCell}>
```
ecraig12345, line 52: While you're here this can be a little simpler:
```suggestion
        {rowsOfItems.map((rows, rowIndex) => {
          return (
            <tr key={rowIndex}>
              {rows.map((cell, cellIndex) => {
                return (
                  <td role="presentation" key={cellIndex} className={classNames.tableCell}>
```

packages/react-next/src/utilities/grid/Grid.base.tsx
```diff
@@ -1,96 +1,77 @@
...
-    const htmlProps = getNativeProps<React.HTMLAttributes<HTMLTableElement>>(
-      this.props,
-      htmlElementProperties,
-      // avoid applying onBlur on the table if it's being used in the FocusZone
-      doNotContainWithinFocusZone ? [] : ['onBlur'],
-    );
+  const classNames = getClassNames(styles!, { theme: props.theme! });
 
-    const classNames = getClassNames(styles!, { theme: this.props.theme! });
+  // Array to store the cells in the correct row index
```
ecraig12345, line 33: Not very important but best to leave the spacing as-is here
```suggestion
```

change/@fluentui-react-focus-2020-08-14-15-13-09-feat-updating-swatchColorPickerNext.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Implementing createMergedRef inside FocusZone.",
```
ecraig12345, line 3:
```suggestion
  "type": "minor",
  "comment": "FocusZone: add elementRef prop.",
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.types.ts
```diff
@@ -45,24 +46,31 @@ export interface ISwatchColorPickerProps {
    */
   colorCells: IColorCellProps[];
 
+  /**
+* @deprecated Provide `selectedId` if controlled or `defaultSelectedId` if uncontrolled
```
ecraig12345, line 50: Some spaces got lost
```suggestion
   * @deprecated Provide `selectedId` if controlled or `defaultSelectedId` if uncontrolled
```

packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
```diff
@@ -1,157 +1,135 @@
...
-  private _onRenderColorOption = (colorOption: IColorCellProps): JSX.Element => {
     // Build an SVG for the cell with the given shape and color properties
     return (
-      <svg
-        className={this._classNames.svg}
-        viewBox="0 0 20 20"
-        fill={getColorFromString(colorOption.color as string)!.str}
-      >
-        {this.props.circle ? <circle cx="50%" cy="50%" r="50%" /> : <rect width="100%" height="100%" />}
+      <svg className={svgClassName} viewBox="0 0 20 20" fill={getColorFromString(colorOption.color!)!.str}>
```
ecraig12345, line 105:
```suggestion
      <svg className={svgClassName} viewBox="0 0 20 20" fill={getColorFromString(colorOption.color)?.str}>
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,325 @@ import {
...
-        onWheel={this._onWheel}
-        onKeyDown={this._onKeyDown}
-        height={props.cellHeight}
-        width={props.cellWidth}
-        borderWidth={props.cellBorderWidth}
-      />
-    );
-  };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [onCellFocused]);
```
ecraig12345, line 107: In this case it's probably better to just add `internalState` to the dep list (even though we know it won't change) so that if anybody edits this in the future and adds references to more things which *might* change, the rule will catch it.
```suggestion
  }, [onCellFocused, internalState]);
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,325 @@ import {
...
+      if (!focusOnHover) {
+        return !internalState.isNavigationIdle || !!disabled;
+      }
+      if (internalState.isNavigationIdle && !disabled) {
+        ev.currentTarget.focus();
+      }
+      return true;
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [focusOnHover, disabled],
```
ecraig12345, line 123: same
```suggestion
    [focusOnHover, disabled, internalState],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,325 @@ import {
...
+      if (internalState.isNavigationIdle && !(document && targetElement === (document.activeElement as HTMLElement))) {
+        targetElement.focus();
+      }
 
-    return true;
-  };
+      return true;
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [focusOnHover, disabled],
```
ecraig12345, line 145:
```suggestion
    [focusOnHover, disabled, internalState],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,325 @@ import {
...
           }
-        } else {
-          (elements[index] as HTMLElement).focus();
+
+          break;
         }
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [disabled, focusOnHover, mouseLeaveParentSelector],
```
ecraig12345, line 187:
```suggestion
    [disabled, focusOnHover, mouseLeaveParentSelector, internalState],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,325 @@ import {
...
+          internalState.cellFocused = true;
+          return onCellFocused(item.id, item.color);
+        } else {
+          internalState.cellFocused = false;
+          return onCellFocused();
+        }
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [onCellFocused],
```
ecraig12345, line 219:
```suggestion
    [onCellFocused, internalState],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,325 @@ import {
...
+      if (item.id !== selectedId) {
+        if (onCellFocused && internalState.cellFocused) {
+          internalState.cellFocused = false;
+          onCellFocused();
+        }
+        setSelectedId(item.id);
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [disabled, onCellFocused, selectedId],
```
ecraig12345, line 240:
```suggestion
    [disabled, onCellFocused, selectedId, internalState],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,325 @@ import {
...
 
-    this.navigationIdleTimeoutId = this.async.setTimeout(() => {
-      this.isNavigationIdle = true;
-    }, this.navigationIdleDelay);
-  };
+    internalState.navigationIdleTimeoutId = setTimeout(() => {
+      internalState.isNavigationIdle = true;
+    }, internalState.navigationIdleDelay);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
ecraig12345, line 259: Even though in this case we know none of the things currently referenced will change, it's still safer to just add them to deps so the rule can stay enabled to catch potential future mistakes.
```suggestion
  }, [internalState, setTimeout, clearTimeout]);
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,325 @@ import {
...
+        ev.which === KeyCodes.left ||
+        ev.which === KeyCodes.right
+      ) {
+        setNavigationTimeout();
       }
-    }
-  };
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [],
```
ecraig12345, line 276:
```suggestion
    [setNavigationTimeout],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,325 @@ import {
...
+   * Render a color cell
+   * @param item - The item to render
+   * @returns - Element representing the item
    */
-  private _onCellClick = (item: IColorCellProps): void => {
-    if (this.props.disabled) {
-      return;
-    }
-
-    const index = item.index as number;
```
ecraig12345, line 284:
```suggestion
```

packages/react-next/RELEASE_NOTES.md
```diff
@@ -70,6 +70,11 @@ TODO: document any API or functionality changes
 
 - Updated enums to string union type: `PivotLinkFormat`, `PivotLinkSize`. (#13370)
 
+### SwatchColorPicker
+
+- Added an `onChange` prop and deprecated `onColorChanged`.
+- Deprecated `isControlled`.
```
ecraig12345, line 76: A couple extra notes. `id` was required before, so I'm not sure why indices were used to track selection, but it's not a bad idea to call out.

```suggestion
- Removed deprecated props `positionInSet` (use `ariaPosInSet`) and `setSize` (use `ariaSetSize`).
- Added an `onChange` prop and deprecated `onColorChanged`.
- Deprecated `isControlled`. Provide `selectedId` for controlled behavior and `defaultSelectedId` for uncontrolled behavior.
- Selection state is now tracked internally based on `IColorCellProps.id`, not item index. Ensure that all color cells have a unique `id` property.
```

packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
czearing: That's what I was planning. This PR is already quite large so for readability it would probably be best to rename it to ButtonGrid after the Grid conversion.

packages/react-next/src/utilities/grid/Grid.base.tsx
czearing: I'm not sure if role={'row'} should be removed since it is used to get the table rows in the tests.

packages/react-next/src/components/SwatchColorPicker/ColorPickerGridCell.base.tsx
```diff
@@ -12,146 +12,121 @@ import {
...
+    disabled,
+    selected,
+    circle,
+    isWhite: isWhiteCell(color!),
+    height,
+    width,
+    borderWidth,
+  });
+
+  const getColorPickerGridCellButtonClassNames = (
```
ecraig12345, line 57: Please move this back to the exact previous implementation, where it's outside the component body and memoized. If I remember right that was an important perf optimization for a partner.

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.types.ts
```diff
@@ -45,24 +46,30 @@ export interface ISwatchColorPickerProps {
...
+
   /**
    * Indicates whether the SwatchColorPicker is fully controlled.
    * When true, the component will not set its internal state to track the selected color.
    * Instead, the parent component will be responsible for handling state in the callbacks like
    * `onColorChanged`.
-   *
-   * NOTE: This property is a temporary workaround to force the component to be fully controllable
-   * without breaking existing behavior
    */
```
ecraig12345, line 59: This comment needs to be updated to something like the following:
```
  /**
   * ID of the default selected color swatch. Only provide this if the SwatchColorPicker is an
   * uncontrolled component; otherwise, use the `selectedId` property.
   */
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.types.ts
```diff
@@ -45,24 +46,30 @@ export interface ISwatchColorPickerProps {
...
   colorCells: IColorCellProps[];
 
+  /**
```
ecraig12345, line 49: Can't directly comment on it, but the docs for `selectedId` need to be updated like this:
```
  /**
   * ID of the current selected color swatch. Only provide this if the SwatchColorPicker is a
   * controlled component where you are maintaining its current state; otherwise, use the
   * `defaultSelectedId` property.
   */
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,316 @@ import {
...
 import { IColorCellProps } from './ColorPickerGridCell.types';
 import { ColorPickerGridCell } from './ColorPickerGridCell';
-import { memoizeFunction, warnDeprecations } from '@uifabric/utilities';
-
-export interface ISwatchColorPickerState {
-  selectedIndex?: number;
+import { useId, useConst, useSetTimeout, useControllableValue, useWarnings } from '@uifabric/react-hooks';
+import { IGridProps } from '../../utilities/grid/Grid.types';
+
+interface ISwatchColorPickerState {
```
ecraig12345, line 14: Nitpicky thing but it would be better to call this `ISwatchColorPickerInternalState` to reduce potential confusion with the old version of React state.

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,316 @@ import {
...
+    (ev: React.MouseEvent<HTMLButtonElement>): boolean => {
+      if (!focusOnHover) {
+        return !internalState.isNavigationIdle || !!disabled;
+      }
+      if (internalState.isNavigationIdle && !disabled) {
+        ev.currentTarget.focus();
+      }
+      return true;
+    },
+    [focusOnHover, internalState.isNavigationIdle, disabled],
```
ecraig12345, line 121: This should just depend on the state object, not the actual idle value. That's part of the point of internal state objects: not needing to re-generate callbacks when values change.
```suggestion
    [focusOnHover, internalState, disabled],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,316 @@ import {
...
+      // If navigation is idle and the targetElement is the focused element bail out
+      if (internalState.isNavigationIdle && !(document && targetElement === (document.activeElement as HTMLElement))) {
+        targetElement.focus();
+      }
 
-    return true;
-  };
+      return true;
+    },
+    [focusOnHover, internalState.isNavigationIdle, disabled],
```
ecraig12345, line 142:
```suggestion
    [focusOnHover, internalState, disabled],
```

packages/react-next/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -16,361 +8,316 @@ import {
...
+            (elements[index] as HTMLElement).focus();
           }
-        } else {
-          (elements[index] as HTMLElement).focus();
+
+          break;
         }
+      }
+    },
+    [disabled, focusOnHover, internalState.isNavigationIdle, mouseLeaveParentSelector],
```
ecraig12345, line 183:
```suggestion
    [disabled, focusOnHover, internalState, mouseLeaveParentSelector],
```

ecraig12345:
Since this has already been through so many iterations and delays (sorry) I pushed an update fixing a couple more issues I noticed (plus some minor things):
1. Even though `onColorChanged` is deprecated, we still need to call it
2. Passing `onChange` with 3 params directly to `useControllableValue` will result in it being called incorrectly, so a wrapper is needed (even without needing to handle the deprecated `onColorChanged`, but a wrapper is convenient for both)
3. We should probably keep the old param order for `onChange` (`id` then `color`)

ecraig12345:
I approved the screener results since they're the same as the old versions as far as I can tell (from side by side visual comparison). There's an issue where a selected square seems to get shifted down by 1-2px, which I suspect is wrong, but it's not new in this PR.

PR #14545, opened 2020-08-14
Fix styles implementation instructions

### LINKED ISSUE #13380: Add note about grid vs Stack to layout styles page on website
On the Layout page of the website's Styles section (https://developer.microsoft.com/en-us/fluentui#/styles/web/layout), we should add a note that grid styles are not available in Fabric React (only with Fabric Core), and that while you can use Fabric Core with Fabric React, we recommend trying Stack first.

ecraig12345:
#### Pull request checklist

- [x] Fixes #13380

#### Description of changes

The implementation instructions on the website styles pages were kind of incomplete and/or incoherent. (This was pointed out by #14519 but a more comprehensive fix was needed.) This PR fixes the worst of the issues, as well as clarifying that the layout grid is only available via Fabric Core and suggesting alternatives.


apps/fabric-website/src/pages/Styles/ElevationPage/docs/web/ElevationImplementation.md
```diff
@@ -1,35 +1,53 @@
...
+There are multiple ways to reference the depth variables in JavaScript.
+
+The recommended way is via the theme:
+
 ```jsx
-import { Depths } from '@uifabric/fluent-theme/lib/fluent/FluentDepths';
+import { getTheme } from '@fluentui/react';
 
-<div style={{ boxShadow: Depths.depth8 }} />;
+const theme = getTheme();
```
micahgodbolt, line 10: this gets the global theme, right? Might need to add that caviat vs contextual theme.

ecraig12345: I thought about that, but I couldn't find a good existing explanation for that on the website and didn't want to spend time trying to write one (because it would take awhile)... I guess I could add a purposefully vague comment like "or other method to get the theme" to avoid having to explain it?

micahgodbolt: this is where @xugao's useTheme hook might come in.

xugao: we don't have a `useTheme` for v7 but do have one for v8 (in `react-theme-provider`).  
I would say leave this way and I will update this when post ship V8

ecraig12345:
Noticed some issues with markdown parsing on the grid page--will figure that out then update the PR and remove "do not merge."

PR #14547, opened 2020-08-15
increase speed of v0 builds by making it participate in the per package test script

kenotron:
#### Description of changes

In this change, I created a 2 projects: @fluentui/projects-test and @fluentui/circulars-test

These two projects are shims on top of the existing `gulp test:projects` and `gulp test:circulars`. I also made `@fluentui/e2e` work by package. By moving these scripts into their own package, we can fully realize scoped builds for those packages as well as taking advantage of the multiple core build system we have in our local machines and CI build agents. These tests can fully be running in parallel with other build tasks now!

Because of this, we can remove an entire build job in the pipeline because these tests are going to be participating in the same granularity level of package-task. Which will allow for scoping to be properly used in CI build situations as well! 

#### Future (not in scope in this PR)
When we properly scope the output of these tests / jobs to within package folders, we can even achieve build caching as well! Another thing to do is to break up the screener tests more so we can scope bits and pieces of those so we don't run the entire suite when we are certain that the downstream isn't affected. (e.g. web component vs v0 vs v7 vs v8).

#### Focus areas to test

(optional)


packages/fluentui/circulars-test/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "dependencies": {
+    "@fluentui/react-northstar": "^0.51.0"
+  },
+  "devDependencies": {
+    "@uifabric/build": "^7.0.0",
+    "gulp": "^4.0.2",
+    "typescript": "3.7.2"
+  },
+  "publishConfig": {
+    "access": "public"
```
ecraig12345, line 15: Coming in late here but I don't think these should be published?

PR #14550, opened 2020-08-16
Only run screener that is needed - lage run graph

kenotron:
#### Description of changes

Previously, I've combined other validation steps to be run in one job (no longer dividing v0, v8, v7, web-components). With a standardized npm run script naming convention, we can now allow `lage` to do proper scoping for PR builds. This PR will add the capability of doing screener only for affected changes. No longer will the screener for v0 be run for v7 changes and vice versa UNLESS the change is in a shared utility, etc.

#### Focus areas to test

(optional)


packages/fluentui/visual-test/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@fluentui/visual-test",
+  "version": "0.51.0",
+  "private": true,
+  "license": "MIT",
+  "dependencies": {
+    "@fluentui/react-northstar": "^0.51.0"
```
layershifter, line 7: Should it have a dependency on `@fluentui/docs` instead?

layershifter:
@kenotron is it alive or should be closed? 🤔

ecraig12345:
Closing the PR since I think this was solved a different way--thanks @layershifter for the ping.

ecraig12345:
Actually I take that back...we seem to still be running screener unconditionally. Let me try updating this from master and re-running the build since I don't remember what was the issue.

ecraig12345:
Hmm on second thought, re-closing and opening a new version that's based off the latest master. Too many changes to merge this nicely.

PR #14562, opened 2020-08-17
ThemeProvider: Fixing RecursivePartial

dzearing:
The typings for RecursivePartial were not working correctly, and subsequently typing errors would occur from partial themes containing tokens without `body`. This should resolve the issue.


packages/react-theme-provider/src/types.ts
```diff
@@ -64,11 +64,7 @@ export type TokenSetType = { [key: string]: string | TokenSetType | undefined };
  * Recursive partial type.
  */
 export type RecursivePartial<T> = {
-  [P in keyof T]?: T[P] extends (infer U)[]
-    ? RecursivePartial<U>[]
-    : T[P] extends object
-    ? RecursivePartial<T[P]>
-    : T[P];
+  [P in keyof T]?: T[P] extends Array<infer I> ? Array<RecursivePartial<I>> : RecursivePartial<T[P]>;
```
xugao, line 67: this is not quite we want either - `stylesheet` in `Theme` will be type `(string | undefined)[] | undefined` instead of `string[] | undefined`

dzearing: I'm not sure that's right; I do expect `(string|undefined)[] | undefined` to be allowed from something which is recursively partial.

xugao: it's totally right for the concept of recursive partial. but if you purely thinking about defining partial theme, is that what you want?

PR #14567, opened 2020-08-17
Add ie11 polyfill package and test with react-button package

### LINKED ISSUE #14668: IE 11 polyfill for CSS variables
This is to add a polyfill for ie11 since it does not support css variables. Main work items are:

- [x] Make a package to put this polyfill script and future ie11 polyfills
- [x] Check in polyfill script to the package
- [x] Test polyfill on component that uses css variables(Button in @fluentui/react-button)
- [x] Write documentation for using the polyfill
- [ ] Repackage (soft fork)

For websites that use hot reloading, pages that need the polyfill script would need to be refreshed, or the polyfill script needs to be put in these pages.

tomi-msft:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14668

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/ie11-custom-properties/LICENSE
```diff
@@ -0,0 +1,15 @@
+@fluentui/ie11-custom-properties
+
+Copyright (c) Microsoft Corporation
```
ecraig12345, line 3: @dzearing What should we do about the copyright and authorship notice since this is forked from another package?

packages/ie11-custom-properties/README.md
```diff
@@ -0,0 +1,7 @@
+# @fluentui/ie11-custom-properties
+
+**Ie Eleven Custom Properties for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
```
ecraig12345, line 3: Update this so it makes sense :)

packages/ie11-custom-properties/ie11CustomProperties.js
```diff
@@ -0,0 +1,675 @@
+/*! ie11CustomProperties.js v4.1.0 | MIT License | https://git.io/fjXMN */
```
ecraig12345, line 1: Should probably remove this line, or replace it with something indicating the original source.

packages/example-app-base/src/components/CodeSnippet/CodeSnippet.tsx
```diff
@@ -38,7 +38,7 @@ style['hljs-comment'] = style['hljs-quote'] = {
 };
 
 style['hljs-built_in'] = style['builtin-name'] = {
-  color: FluentTheme.palette.themeDarker,
+  color: FluentTheme?.palette.themeDarker,
```
ecraig12345, line 41: Oh weird, this is not where I thought the error came from--this part is a bit older (1 month) and it really seems like the default FluentTheme ought to be defined...

khmakoto:
Waiting until we figure out licensing to merge this PR.

packages/ie11-custom-properties/ie11CustomProperties.js
tomi-msft: I have removed this line

packages/ie11-custom-properties/README.md
```diff
@@ -0,0 +1,13 @@
+# @fluentui/ie11-custom-properties
+
+**Ie11 Polyfill for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
```
ecraig12345, line 3:
```suggestion
**IE11 CSS variables polyfill**
```

packages/ie11-custom-properties/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Package to hold ie11 polyfill
```
ecraig12345, line 5: Would be good to specify what this is a polyfill for. There are lots of other types of polyfills besides for CSS variables (like for adding Promise, newer array methods, etc).

tomi-msft: Added more info to the README

change/@fluentui-ie11-custom-properties-2020-08-27-12-13-14-ie11-polyfill.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
```
ecraig12345, line 2: Delete this file

packages/ie11-polyfills/README.md
```diff
@@ -0,0 +1,13 @@
+# @fluentui/ie11-polyfills
+
+**IE11 CSS variables polyfill**
```
ecraig12345, line 3:
```suggestion
**IE11 polyfills for Fluent UI React**
```

packages/ie11-polyfills/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Package to hold ie11 polyfill that allows webpages that use css variables for styling to render correctly in ie11. It will also hold any future ie11 polyfills
+
+To import ie11-polyfill
+
+```js
+import '@fluentui/ie11-polyfills';
+```
+
+It just works. If you are using any websites with hot reloading, you will have to either put the polyfill on each page that uses css variables, or refresh the page for the script to take effect.
```
ecraig12345, line 13: Some nitpicky suggestions
```suggestion
Package to hold IE11 polyfill that allows webpages that use CSS variables for styling to render correctly in IE11. It will also hold any future IE11 polyfills

To use the polyfills, just import the package:

```js
import '@fluentui/ie11-polyfills';
```

If you are using any websites with hot reloading, you will have to either put the polyfill on each page that uses CSS variables, or refresh the page for the script to take effect.
```

File: packages/ie11-polyfills/index.js [hunk 10]
```diff
@@ -0,0 +1 @@
+export * from 'ie11-custom-properties';
```
ecraig12345, line 1: If we're going to use ES6 export syntax like this, the package will need to use TS to transpile (not everything can handle ES6 exports). Also, it would be better to just do `import 'ie11-custom-properties'` since that package relies on side effects rather than exporting anything.

The other option would be to keep it as JS and change this line to `require('ie11-custom-properties')`.

I can see arguments for either keeping as JS (mainly that there's no build cost) or converting to TS (which would probably be needed if we add more polyfills later). Unless someone else has strong feelings I think we should stick with JS for now.

packages/ie11-polyfills/just.config.ts
```diff
@@ -0,0 +1,3 @@
+const { preset } = require('@uifabric/build');
```
ecraig12345, line 1: Assuming we stick with JS (not TS) delete this file

packages/ie11-polyfills/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/fluentui"
+  },
+  "devDependencies": {
+    "@uifabric/build": "^7.0.0",
+    "ie11-custom-properties": "4.1.0"
+  },
+  "scripts": {
+    "build": "just-scripts build"
```
ecraig12345, line 16: Delete this if we're using JS

packages/ie11-polyfills/index.js
tomi-msft: Will stick with JS for now

change/@uifabric-experiments-2020-08-27-14-11-31-ie11-polyfill.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
ecraig12345, line 2: Remove this file (I think it was fixed in master)

ecraig12345 (approved):
Looks good once you delete the extra change file.

PR #14571, opened 2020-08-17
Scaffolding react-utilities

dzearing:
Scaffolding `@fluentui/react-utilities` package to start separating React specifics from general utilities.


packages/react-utilities/README.md
```diff
@@ -0,0 +1,17 @@
+# @fluentui/react-utilities
+
+**React Utilities components for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
```
levithomason, line 3:
```suggestion
**React Utilities for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui) components**
```

packages/react-utilities/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+Tests if children are nil in React and Preact.
+
+Example:
+
+```jsx
+if (childrenExist(props.children)) {
+  // render children.
+}
+```
```
levithomason, line 17: Would be great to say something like:

```mdx
React `children` can be one many types. See the [tests](/test/...) for all the covered cases.
```

levithomason:
Boy, perf results are strange.

packages/react-utilities/README.md
dzearing: done and done

dzearing:
perf results...strange for sure...

packages/react-utilities/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "strict": true,
+    "moduleResolution": "node",
+    "preserveConstEnums": true,
+    "lib": ["es2015", "dom", "es2015.promise"],
+    "typeRoots": ["../../node_modules/@types", "../../typings"],
+    "types": ["jest", "webpack-env", "custom-global"],
+    "paths": {
+      "@fluentui/react-utilities/lib/*": ["./src/*"],
+      "@fluentui/react-utilities": ["./src"]
+    }
```
ecraig12345, line 23: Since the package doesn't have examples:
```suggestion
    "types": ["jest", "webpack-env", "custom-global"]
```
(we really ought to add an option to create-package about whether the package will have examples, so I don't have to keep making this suggestion)

dzearing: I'm afraid to reset the CI as this is the 3rd attempt, and I don't think having these paths is going to cause confusion or problems, but happy to merge in a separate pr later when I'm in the area.

PR #14572, opened 2020-08-17
Drag drop between SelectedItemLists (and UnifiedPickers as well)

elisabethcvs:
#### Pull request checklist

#### Description of changes

- Integrated changes to the UnifiedPicker as well
- Modified the double wells drag drop example
- Added a SelectedItemsList drag drop between lists example

Note: You can only drop in the unified picker if there are already items (so SelectedItemsList is visible). Planning on fixing that in a follow up change, but some of these changes are needed for copy paste so I am getting the bulk of the change in now.


packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
```diff
@@ -114,4 +114,8 @@ export interface ISelectedItemsListProps<T> extends React.ClassAttributes<any> {
    * Helper for the drag and drop
    */
   dragDropHelper?: IDragDropHelper;
+
+  getSerializedItems?: (items: T[]) => string;
```
elisabethcvs, line 118: These need comments

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.types.ts
```diff
@@ -87,4 +87,6 @@ export interface IUnifiedPickerProps<T> {
    * Drag drop events callback interface
    */
   dragDropEvents?: IDragDropEvents;
+
+  customClipboardType?: string;
```
elisabethcvs, line 91: This also needs a comment

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -94,18 +86,56 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
+  };
+
   const _onDrop = (item?: any, event?: DragEvent): void => {
-    if (draggedItem) {
-      _insertBeforeItem(item);
+    if (event?.dataTransfer) {
+      event.preventDefault();
+      const data = event.dataTransfer.items;
+      for (let i = 0; i < data.length; i++) {
+        if (data[i].kind === 'string' && data[i].type === props.customClipboardType) {
```
elisabethcvs, line 109: With these changes, the UPP will only do drag drop if the serialized/deserialized functions are defined. Should I add secondary logic to say if there isn't data transfer data, but is a dragged item index, move that?

I think probably yes thinking about it

packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
```diff
@@ -114,4 +114,19 @@ export interface ISelectedItemsListProps<T> extends React.ClassAttributes<any> {
    * Helper for the drag and drop
    */
   dragDropHelper?: IDragDropHelper;
+
+  /**
+   * Callback for when items need to be converted to a string for a drag action
+   */
+  getSerializedItems?: (items: T[]) => string;
```
nebhatna, line 121: might want to make the callback name more specific to drag/drop ? Same for the deserailized one?

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/examples/SelectedPeopleList.DragDropBetweenWells.Example.tsx
```diff
@@ -0,0 +1,124 @@
...
+    return dragEnterClass;
+  };
+
+  const _onDrop = (item?: any, event?: DragEvent): void => {
+    if (event?.dataTransfer) {
+      event.preventDefault();
+      const data = event.dataTransfer.items;
+      for (let i = 0; i < data.length; i++) {
+        if (data[i].kind === 'string' && data[i].type === 'persona') {
+          data[i].getAsString((s: string) => {
```
nebhatna, line 47: consider naming 's' here to be something more relevant to the context.

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/examples/SelectedPeopleList.DragDropBetweenWells.Example.tsx
```diff
@@ -0,0 +1,124 @@
...
+          });
+        }
+      }
+    }
+  };
+
+  const _onDragStart = (item?: any, itemIndex?: number, tempSelectedItems?: any[], event?: DragEvent): void => {
+    if (event) {
+      const dataList = event?.dataTransfer?.items;
+      const str = (item as IPersonaProps).text;
```
nebhatna, line 62: nit: const personaText = ....

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/examples/SelectedPeopleList.DragDropBetweenWells.Example.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const _onAddItemButtonClicked = React.useCallback(() => {
+    setCurrentSelectedItems([...currentSelectedItems, people[index]]);
+    setIndex(index + 1);
+  }, [currentSelectedItems, index]);
+
+  const _onItemsRemoved = React.useCallback(
+    (items: IPersona[]) => {
+      const currentSelectedItemsCopy = [...currentSelectedItems];
+      items.forEach(item => {
+        const indexToRemove = currentSelectedItemsCopy.indexOf(item);
```
nebhatna, line 94: let indexToRemove

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/examples/SelectedPeopleList.DragDropBetweenWells.Example.tsx
```diff
@@ -0,0 +1,124 @@
...
+        currentSelectedItemsCopy.splice(indexToRemove, 1);
+        setCurrentSelectedItems([...currentSelectedItemsCopy]);
+      });
+    },
+    [currentSelectedItems],
```
nebhatna, line 99: does this need to re-render on items change as well?

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -94,18 +86,64 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
-    if (draggedItem) {
-      _insertBeforeItem(item);
+    const insertIndex = selectedItems.indexOf(item);
+    let isDropHandled = false;
+    if (event?.dataTransfer) {
+      event.preventDefault();
+      const data = event.dataTransfer.items;
+      for (let i = 0; i < data.length; i++) {
+        if (data[i].kind === 'string' && data[i].type === props.customClipboardType) {
+          data[i].getAsString((s: string) => {
```
nebhatna, line 109: nit: consider renaming s to be contextual

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/examples/SelectedPeopleList.DragDropBetweenWells.Example.tsx
elisabethcvs: ? Isn't that what I'm doing with the [currentSelectedItems] bit? (hooks are hard)

packages/experiments/src/components/UnifiedPicker/UnifiedPeoplePicker/examples/DoubleUnifiedPeoplePicker.Example.tsx
```diff
@@ -1,5 +1,239 @@
...
+          });
+        }
+      });
+    }
+
+    setPeopleSelectedItems(prevPeopleSelectedItems => [...prevPeopleSelectedItems, ...newList]);
+  };
+
+  const _getSerializedItems = (items: IPersonaProps[]): string => {
+    return _getItemsCopyText(items); // Do we want to combine these or have them be separate?
```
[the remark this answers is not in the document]
elisabethcvs: Makes sense. I think what we're probably going to end up wanting to do is calling both of them for copy - so we put both a plain text version and the serialized version on the clipboard.

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/examples/SelectedPeopleList.DragDropBetweenWells.Example.tsx
elisabethcvs: Lint wants this to be const. It doesn't get reassigned so I think that's correct.

PR #14574, opened 2020-08-17
CodeMods: Sreamline CodeMod creation process with createCodeMod

tmaster628:
Implemented `createCodeMod()`, a helper function that takes in a string codemod name and a function that mirrors a "run" body in an actual code mod. This function will return a nicely packaged codemod that is ready-to-run. This is great because I previously wrote some transforms that devs can use to change prop values that are only accessible if they can invoke `renameProp()` directly. Hopefully this should give devs the flexibility they need when writing codemods!

Here's how it could be used to rename a prop.

```
    // Rename 'checked' to 'toggled' in Button
    const renamePropCodeMod = (file : SourceFile) => void { 
        // Example body of a code mod. Easily customizable by a developer!
        const tags = findJsxTag(file, 'Button');
        renameProp(tags, 'checked', 'toggled');         
    };
    const shinyNewCodeMod = createCodeMod('Rename checked to toggled in Button', renamePropCodeMod);
    // run shinyNewCodeMod with runMods() on some source files!
```

`createCodeMod()` is also used by `getCodeModsFromJson()` since it's such a generic function. Woo hoo!

In other news, the codeMod generator no longer packages all mods in the json into a single mod. Although this could hurt performance (more mods = more runs), now that mods are individual, one could group them for better logging or potentially less redundancy with file visitation. More metadata might be needed there, though.

### DIFF
--- a/packages/codemods/src/codeMods/mods/configMod/configMod.ts
+++ b/packages/codemods/src/codeMods/mods/configMod/configMod.ts
@@ -51,9 +51,9 @@ export function createCodeMod(modName: string, mod: (file: SourceFile) => void):
         /* Codemod body. */
         mod(file);
       } catch (e) {
-        return { success: false };
+        return Err({ reason: `Mod failed: ${e}` });
       }
-      return { success: true };
+      return Ok({ logs: ['Upgrade completed'] });
     },
     version: '100000',
     name: modName,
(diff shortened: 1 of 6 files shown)


packages/codemods/src/codeMods/mods/configMod/configMod.ts
```diff
@@ -12,61 +12,67 @@ import { Ok, Err } from '../../../helpers/result';
...
+    run: (file: SourceFile) => {
+      try {
+        /* Codemod body. */
+        mod(file);
+      } catch (e) {
+        return Err({ reason: `Mod failed: ${e}` });
+      }
+      return Ok({ logs: ['Upgrade completed'] });
+    },
+    version: '100000',
```
joschect, line 58: Lets pass version in as well, maybe instead of modName it should take a list of modOptions?

tmaster628: Sounds good!

PR #14587, opened 2020-08-18
chore: fix double info generation

layershifter:
#### Description of changes

In #14472 I moved `componentInfo` to related packages. The issue is that on `yarn build` it will run `gulp build:component-info` twice 🤦


dzearing:
thank you!

PR #14590, opened 2020-08-18
chore: improve docgen perf & use lazy mode to load imports

layershifter:
#### Description of changes

This is heavily based on work that @ecraig12345 did in #12897. However, the actual implementation is inspired by https://github.com/strothj/react-docgen-typescript-loader/pull/40. The idea is still the same - to reuse the same `ts.Program` as it's quite heavy. Direct requires of component's files were also removes to parse `defaultProps` we reuse existing AST produced by Babel.

This allows to speed up `build:component-info` without caching it takes around 100-110 seconds on machine.

**Previous PRs:** `Finished 'build:docs:assets:component:info' after 5.35 min`
**This PR:** `Finished 'build:docs:assets:component:info' after 1.57 min`

---

~~Another change is the lazy evaluation for [`@babel/plugin-transform-modules-commonjs`](https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs). This improves initial load time of modules because dependencies will be evaluated based on their usage.~~


scripts/gulp/plugins/util/getComponentInfo.ts
```diff
@@ -130,8 +131,8 @@ const getComponentInfo = (tsConfigPath: string, filepath: string, ignoredParentI
     const visibleInTags = !_.find(tags, { title: 'docSiteIgnore' });
 
     if (visibleInDefinition && visibleInTags) {
-      const types = parseType(componentFile, info.displayName, propName, propDef);
-      const defaultValue = parseDefaultValue(Component, propDef, types);
+      const types = parseType(componentAst, info.displayName, propName, propDef);
+      const defaultValue = undefined; // parseDefaultValue(Component, propDef, types);
```
layershifter, line 135: Oops, should be fixed before merge...

scripts/gulp/plugins/util/getComponentInfo.ts
```diff
@@ -148,7 +149,7 @@ const getComponentInfo = (tsConfigPath: string, filepath: string, ignoredParentI
   if (info.props.as) {
     props.push({
       description: 'An element type to render as (string or component).',
-      defaultValue: parseDefaultValue(Component, info.props.as, []) || 'div',
+      defaultValue: undefined, // parseDefaultValue(Component, info.props.as, []) || 'div',
```
dzearing, line 152: same

dzearing:
I think @ecraig12345 should probably take a look here as well, seems reasonable.

layershifter:
> This allows to speed up `build:component-info` without caching it takes around 100-110 seconds on machine.

This is definitely better, but it still awful 😨 I don't have the explanation why it slow downs on some files. @ecraig12345 any ideas?

| File                                 | Time in ms         |
| ------------------------------------ | ------------------ |
| Accordion\Accordion.tsx              | 3376.5990000367165 |
| Accordion\AccordionContent.tsx       | 1234.237399995327  |
| Accordion\AccordionTitle.tsx         | 39.98910003900528  |
| Alert\Alert.tsx                      | 1233.830999970436  |
| Alert\AlertDismissAction.tsx         | 1197.1260999441147 |
| Attachment\Attachment.tsx            | 1207.950500011444  |
| Attachment\AttachmentAction.tsx      | 1249.627600967884  |
| Attachment\AttachmentBody.tsx        | 44.709100008010864 |
| Attachment\AttachmentDescription.tsx | 30.8121999502182   |
| Attachment\AttachmentHeader.tsx      | 29.475099980831146 |
| Attachment\AttachmentIcon.tsx        | 28.09469997882843  |
| Animation\Animation.tsx              | 778.9541000127792  |

scripts/gulp/plugins/util/getComponentInfo.ts
```diff
@@ -6,10 +6,19 @@ import fs from 'fs';
...
-import parseDefaultValue from './parseDefaultValue';
 import parseDocblock from './parseDocblock';
 import parseType from './parseType';
 import getShorthandInfo from './getShorthandInfo';
+import { getProgram, loadFiles } from './tsLanguageService';
+
+export type GetComponentInfoOptions = {
+  /** Path to the file containing a single component. */
+  filePath: string;
+  /** Path to the tsconfig to use for processing the component file. Not used if `program` is provided. */
```
ecraig12345, line 17:
```suggestion
  /** Path to the tsconfig to use for processing the component file. */
```

scripts/gulp/plugins/util/tsLanguageService.ts
```diff
@@ -0,0 +1,82 @@
+import * as fs from 'fs';
+import path from 'path';
+import * as ts from 'typescript';
+
+export type TSFile = {
+  text?: string;
+  version: number;
+};
+
+function getTSConfigFile(tsconfigPath: string): ts.ParsedCommandLine {
```
ecraig12345, line 10: I'm curious, what's the reasoning for adding this more complicated logic rather than just parsing the tsconfig like in react-conformance?

layershifter: It works with `watch` without issues as files can be added/changed on the fly. Is it possible in your implementation?

ecraig12345: Ah okay, hadn't realized that was the purpose. Might be good to add a comment about that somewhere.

layershifter: Good point 👍 I will add comments tomorrow and will fix remaining things

layershifter:
> Another change is the lazy evaluation for [`@babel/plugin-transform-modules-commonjs`](https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs). This improves initial load time of modules because dependencies will be evaluated based on their usage.

This change was reverted as we have circular references in our code 😭

scripts/gulp/plugins/util/getComponentInfo.ts
layershifter: Was fixed, now we are using AST instead of direct imports

scripts/gulp/plugins/util/tsLanguageService.ts
layershifter: Done 👍

pompomon (approved):

PR #14592, opened 2020-08-18
docs(Datepicker): Add cell customization

pompomon:
#### Pull request checklist

#### Description of changes

Added examples for `calendarCell` and `calendarHeaderCell` customization using slots and children API

HeaderCell with Tooltip
image

Cell with Tooltip
image

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Datepicker/Slots/DatepickerCellExample.shorthand.tsx
+++ b/packages/fluentui/docs/src/examples/components/Datepicker/Slots/DatepickerCellExample.shorthand.tsx
@@ -1,0 +1,22 @@
+import { Datepicker, Tooltip } from '@fluentui/react-northstar';
+import * as React from 'react';
+
+const DatepickerCellExample = () => {
+  return (
+    <Datepicker
+      today={new Date(2020, 8, 12, 0, 0, 0, 0)}
+      calendar={{
+        calendarCell: {
+          children: (ComponentType, props) =>
+            props.isToday ? (
+              <Tooltip content="TODAY!" trigger={<ComponentType {...props} />} />
+            ) : (
+              <ComponentType {...props} />
+            ),
+        },
+      }}
+    />
+  );
+};
+
+export default DatepickerCellExample;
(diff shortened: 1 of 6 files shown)


packages/fluentui/docs/src/examples/components/Datepicker/Slots/DatepickerHeaderCellExample.shorthand.tsx
```diff
@@ -0,0 +1,19 @@
...
+import * as React from 'react';
+
+const DatepickerHeaderCellExample = () => {
+  return (
+    <Datepicker
+      today={new Date(2020, 8, 12, 0, 0, 0, 0)}
+      calendar={{
+        calendarHeaderCell: {
+          children: (ComponentType, props) => (
+            <Tooltip content={props['aria-label']} trigger={<ComponentType {...props} />} />
```
layershifter, line 11: It looks that we will have `aria-label` on a `trigger` still, is it expected?

packages/fluentui/docs/src/examples/components/Datepicker/Slots/DatepickerCellExample.shorthand.tsx
```diff
@@ -0,0 +1,22 @@
...
+import * as React from 'react';
+
+const DatepickerCellExample = () => {
+  return (
+    <Datepicker
+      today={new Date(2020, 8, 12, 0, 0, 0, 0)}
+      calendar={{
+        calendarCell: {
+          children: (ComponentType, props) =>
+            props.isToday ? (
```
layershifter, line 11: Is it really `isToday`? We don't use `is*` prefix in prop names...

layershifter (approved):

packages/fluentui/docs/src/examples/components/Datepicker/Slots/DatepickerHeaderCellExample.shorthand.tsx
pompomon: Yep :)

packages/fluentui/docs/src/examples/components/Datepicker/Slots/DatepickerCellExample.shorthand.tsx
pompomon: Ooops, will fix it with @rymeskar in a separate PR :D

PR #14596, opened 2020-08-18
Avatar visual updates

behowell:
#### Pull request checklist

#### Description of changes

Update Avatar with icon support, token support, and visual alignment with the design spec.
Rename Status to Badge, and add tokens support and visual updates


packages/react-avatar/src/components/Avatar/Avatar.scss
```diff
@@ -24,86 +14,65 @@
...
+  clip-path: var(--avatar-clipPath);
 }
 
 // Label slot
 .label {
   display: flex;
   align-items: center;
   justify-content: center;
-  background-color: var(--avatarFill);
+  background-color: var(--avatar-fill);
```
dzearing, line 39: i think find/replace `--avatar-fill` with `--avatar-background`.

packages/react-avatar/src/components/Avatar/Avatar.scss
```diff
@@ -24,86 +14,65 @@
...
-  --avatarSize: var(--avatarSizeLargest);
+  // Size the badge appropriately for this avatar size
+  --badge-size: calc(var(--avatar-size) / 4);
 }
 
 // Modifiers
 ._square {
   .image {
-    border-radius: var(--squareAvatarBorderRadius);
+    border-radius: var(--avatar-square-borderRadius, calc(var(--avatar-size) / 12));
```
dzearing, line 65: We probably should rework this a little: at first I thought `avatar-square-borderRadius` was correct, but I don't think this approach scales. What about when you want square to also change the border width? or the border color? Or maybe square needs box shadow.

Instead, we should look at class modifiers as a different "variant" of the control, where a variant simply defines new set of token values:

```css
._square .image {
  --avatar-borderRadius: calc(var(--avatar-size) / 12);
}
```

packages/react-avatar/src/components/Avatar/Avatar.scss
```diff
@@ -24,86 +14,65 @@
...
   .label {
-    border-radius: var(--squareAvatarBorderRadius);
+    border-radius: var(--avatar-square-borderRadius, calc(var(--avatar-size) / 12));
+  }
+}
+
+._display_icon {
+  .label {
+    // The icon should be half the size of the avatar
+    font-size: calc(var(--avatar-size) / 2);
```
dzearing, line 76: Same here:

```css
._display_icon .label {
  --avatar-fontSize: calc(var(--avatar-size) / 2);
}
```

dzearing: There are others in this css file; see if you can look for these and snap to this pattern and if it makes sense.

packages/react-avatar/src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import { Badge } from '../Badge/Badge';
+import { StoryExample } from '../utils/StoryExample';
+import { SkypeCheckIcon, SkypeArrowIcon, SkypeMinusIcon, SkypeClockIcon } from '@fluentui/react-icons';
+
+export const BadgeExamples = () => (
+  <>
+    <StoryExample title="Badge">
+      <Badge size="smallest" state="error" />
```
dzearing, line 9: Just some thoughts:

I hadn't bothered questioning this earlier, but v0 refers to the Status mode as `state` prop. First, we should probably question that. The term is a bit overloaded with component state, and we really should consider a name more relevant to the purpose. Maybe `status` or `variant` even, where additional variants could be provided through theme.

So, what is a generic `Badge` component and when should it be used? Right now it represents generic status enum values which map to colors. However the spec assumes these are more about online presence. Would we have something like `PresenceBadge` with status enum values like `online` or `outOfOffice` or `offline`? Because I'm seeing things like the "no" icon used for offline which doesn't map to any implication in the Badge component.

For other types of Avatars, will they have different Badges? We've seen LinkedIn badges be represented as a LinkedIn image. Maybe for email clients, the badge might represent if the user is in your contacts list or not.

dzearing: (I totally do agree that `Avatar` should have a `badge` slot. Makes sense. Just wondering about what goes in that and what the default Avatar should do here.

packages/react-avatar/src/components/Badge/Badge.types.tsx
```diff
@@ -21,6 +21,21 @@ export interface StatusProps extends ComponentProps, React.HTMLAttributes<HTMLEl
 
   /** The pre-defined state values which can be consumed directly. */
   state?: 'success' | 'info' | 'warning' | 'error' | 'unknown';
+
+  /** Style tokens */
+  tokens?: BadgeTokens;
```
dzearing, line 26: Question: I think I teetered between `ButtonTokens` and `ButtonTokenSet`, and landed on the latter because the base type is `TokenSet` I think. What's the right name here?

behowell: I don't have a strong preference either way, but we definitely should be consistent between components. If button uses `ButtonTokenSet` then I'll rename these to match: `AvatarTokenSet` and `BadgeTokenSet`

packages/react-avatar/src/components/Badge/Badge.stories.tsx
behowell: I like the idea of having a very basic/generic `Badge` component (or `useBadge` component hook) that can render any given color, icon, and shape (border radius), and uses the size given by the `--badge-size` token (which the Avatar's styles can set the value of).

Then we can have specialized badges like `PresenceBadge` or `StatusBadge` that has a domain-specific enum that maps to specific colors and icons: `badge={{ as: PresenceBadge, status: 'offline' }}`

Or, plot twist: maybe `Badge` is just a smaller `Avatar`? Perhaps that's going too far, but they would basically be the same thing: they can render an image or a colored background with an optional icon...

behowell: That said, I'd rather address this in a follow-up PR and get this one checked in sooner than later.

dzearing (approved):

PR #14598, opened 2020-08-18
Make the floating suggestions picker have an optional set width

sjadelson:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/experiments/src/components/UnifiedPicker/UnifiedPeoplePicker/examples/UnifiedPeoplePicker.Example.tsx
```diff
@@ -159,6 +159,7 @@ export const UnifiedPeoplePickerExample = (): JSX.Element => {
     noResultsFoundText: 'No suggestions',
     onFloatingSuggestionsDismiss: undefined,
     showSuggestionRemoveButton: true,
+    pickerWidth: '300px',
```
nebhatna, line 162: looks like 300px is not the default then :) Sounds okay to me, just wanted to make sure you intended it to be this way.

sjadelson: Yes, it was intended.  I went back and forth on this.... Making the default _on the floating suggestions_ 300px means that there's no need to change the UPP example, so that anyone who reuses it may miss that it's configurable.  I can change it if you like it better the other way.

nebhatna: Yea, we can add a function in the UnifiedPicker, something similar like-
     targetElement: input.current?.inputElement,

so `const _renderFloatingPicker = () =>
    onRenderFloatingSuggestions({
      ...floatingSuggestionProps,
pickerWidth: props.floatingSuggestionProps.pickerWidth? props.floatingSuggestionProps.pickerWidth : '300px`

sjadelson: Yes!

change/@uifabric-experiments-2020-08-18-14-59-13-InputWidth.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
ecraig12345, line 2: This should be minor since it's adding an API.
```suggestion
  "type": "minor",
```

PR #14601, opened 2020-08-18
Scaffold @fluentui/themes package

xugao:
#### Pull request checklist

#### Description of changes
- added `@fluentui/themes` package for specific theme definitions to live.
- move theme-related typings, `createDefaultTheme`, `FluentTheme`, `TeamsTheme` out from `react-theme-provider` into `@fluentui/themes`.
- moved Button, Flex tokens typing and default tokens to `@fluentui/themes`. 
- updated `ButtonTokenSet` typing to match what we actually have
- removed `--button-focused-*` tokens as they are not being used

#### Focus areas to test

(optional)


packages/themes/src/tokens/button/tokens.ts
```diff
@@ -0,0 +1,109 @@
+import { Theme } from '../../types';
+import { ButtonTokenSet } from './types';
+
+export const defaultButtonTokens: ButtonTokenSet = {
```
xugao, line 4: I don't feel it is ideal to define default tokens for a component in a separate package instead of in the component package itself. so i'm open to suggestion/discussion here

xugao: will move the component tokens back to `react-{component}` packages

packages/react-theme-provider/src/useTheme.ts
```diff
@@ -1,9 +1,9 @@
 import { useContext } from 'react';
 import { useCustomizationSettings } from '@uifabric/utilities';
 import { ITheme } from '@uifabric/styling';
+import { createDefaultTheme } from '@fluentui/themes/lib/createDefaultTheme';
```
khmakoto, line 4: nit: Should we do

```suggestion
import { createDefaultTheme } from '@fluentui/themes';
```

packages/themes/README.md
```diff
@@ -0,0 +1,16 @@
+# @fluentui/themes
+
+**Themes for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
+
+To import theme and provide the theme down to components:
```
khmakoto, line 5:
```suggestion
**Themes for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**

This package contains a set of different themes provided for consumption by the Fluent UI components.

To import a theme and provide it down to the components you can do:
```

packages/themes/src/tokens/button/tokens.ts
khmakoto: Will that be done as part of this PR or as part of a separate one?

xugao: I am going to do it right in this one. let me temporarily close the PR to avoid wasting people's review time!

xugao:
close for now since this needs some more work

PR #14603, opened 2020-08-19
Fixing dragdrop useEffect dependencies to not cause cascading re-renders

nebhatna:
#### Pull request checklist

#### Description of changes
Removing the not needed dependencies for dragdropsubscription useEffect. This is causing infinite loop in the consumer of this package( OWA in this case ).

#### Focus areas to test

(optional)


nebhatna:
@elisabethcvs could you please take a look at this? I tested this using the UPP example in the code, was working as expected. Just want to make sure I didnt miss anything. Thank you!

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/Items/SelectedPersona.tsx
```diff
@@ -114,7 +114,7 @@ const SelectedPersonaInner = React.memo(
         dragDropSubscription?.dispose();
         setDragDropSubscription(undefined);
       };
-    }, [dragDropHelper, dragDropSubscription, root, events, dragDropOptions]);
+    }, [dragDropHelper]);
```
elisabethcvs, line 117: Can you run lint? This is actually what I had in the first place, but lint didn't like it. If not, is there some way to disable it? (Or rewrite this?)

ecraig12345:
Happened to see this PR and started looking over the code, and there are quite a few issues (and fixing the re-render issue properly may involve addressing several of them).

#### Immediate issue

Since the subscription isn't used outside the effect, a better solution is to just keep a local reference rather than storing it in state.

You also shouldn't be creating a new EventGroup every render, and should be disposing it at some point. Easiest way is to create (and dispose) it inside the effect.

`dragDropOptions` definition should move inside the effect as well.

At this point, you can look at the dependencies ESLint wants you to add for the effect, and determine which ones you *actually* want to trigger a re-run. Then list only those ones and disable the lint rule for the line.

#### Other issues

- `getId` will return a new ID each render. Instead you should use `useId` from `@uifabric/react-hooks`.

- Getting `root`: Rather than custom ref logic, findDOMNode, and storing the root element in state, you should just use `React.useRef` like this. (In general you shouldn't be using findDOMNode because it doesn't work on function components and is deprecated in strict mode. And here it's just completely unnecessary.)
```diff
- const [root, setRoot] = React.useState<HTMLElement | undefined>();
+ const rootRef = React.useRef<HTMLElement>(null);

- const _onRootRef = React.useCallback((div: HTMLDivElement) => { ... }

      <div
-        ref={_onRootRef}
+        ref={rootRef}
```

- `isDraggable` should be calculated using `useMemo`, rather than storing in state and calculating with an effect. `const isDraggable = React.useMemo(() => dragDropEvents ? !!(dragDropEvents.canDrag && dragDropEvents.canDrop) : undefined)`

- `droppingClassName` should also be calculated with `useMemo`: `const droppingClassName = React.useMemo(() => isDropping ? droppingClassNames || DEFAULT_DROPPING_CSS_CLASS : '')`

packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/Items/SelectedPersona.tsx
```diff
@@ -61,13 +60,10 @@ const SelectedPersonaInner = React.memo(
       dragDropEvents,
       eventsToRegister,
     } = props;
-    const itemId = getId();
+    const itemId = useId();
     const [root, setRoot] = React.useState<HTMLElement | undefined>();
```
elisabethcvs, line 64: I think according to Elizabeth's comment this needed to get modified as well?

nebhatna: yea i am working on that :) did not seem straightforward enough as it is being used in multiple places below..

PR #14606, opened 2020-08-19
Make FocusTrapZone useful for focusing the first item in a virtual list

ThomasMichon:
#### Description of changes

Made some minor updates to `FocusTrapZone` so it can be leveraged to implement a common accessibility requirement: "Select the first row in the List". Unlike previous attempts at solving this issue, such as using `initialFocusedIndex` on `DetailsList`, this approach allows a host app to specify when and how it wants to focus an element, simply by providing the correct selector for `firstFocusableElement` on a `FocusTrapZone` and calling `.focus()` on its component ref after the content loads. The ideal selector to use is `[data-selection-index="0"]`, which is the first content row according to the `SelectionZone`. This works well with both `DetailsList` *and* `TilesList` without requiring either component to implement something custom.

#### Focus areas to test

Example pages for `TilesList` and `DetailsList` have been updated to include the behavior and showcase its non-invasive impact.


packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -138,7 +138,7 @@ export class FocusTrapZone extends React.Component<IFocusTrapZoneProps, {}> impl
   }
 
   public focus() {
-    const { focusPreviouslyFocusedInnerElement, firstFocusableSelector } = this.props;
+    const { focusPreviouslyFocusedInnerElement, firstFocusableSelector, firstFocusableElement } = this.props;
```
dzearing, line 141: cc @czearing

dzearing: There's a separate PR (#14804) to convert this to a function component, so this change needs to be brought over there. @czearing can help with this once this is merged. (cc @ecraig12345 )

packages/utilities/src/focus.ts
```diff
@@ -467,10 +467,23 @@ export function focusAsync(element: HTMLElement | { focus: () => void } | undefi
...
       // element.focus() is a no-op if the element is no longer in the DOM, meaning this is always safe
       win.requestAnimationFrame(() => {
-        targetToFocusOnNextRepaint && targetToFocusOnNextRepaint.focus();
+        const focusableElement = targetToFocusOnNextRepaint as HTMLElement | null;
 
         // We are done focusing for this frame, so reset the queued focus element
         targetToFocusOnNextRepaint = undefined;
+
+        if (focusableElement) {
+          if (focusableElement.getAttribute && focusableElement.getAttribute(IS_FOCUSABLE_ATTRIBUTE) === 'true') {
```
dzearing, line 476: isn't there a helper for checking if an element is focusable? isTabbable or something.

change/office-ui-fabric-react-2020-08-18-19-00-44-details-list-focus.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
dzearing, line 2:
```suggestion
  "type": "minor",
```

Also mention the new prop being added to FocusTrapZone.These go into release notes.

packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.types.ts
```diff
@@ -64,6 +64,12 @@ export interface IFocusTrapZoneProps extends React.HTMLAttributes<HTMLDivElement
    */
   firstFocusableSelector?: string | (() => string);
 
+  /**
+   * Either a full query selector for the first focusable element, or a function to select the focusable element
+   * within the area directly.
+   */
+  firstFocusableElement?: string | ((element: HTMLElement) => HTMLElement | null);
```
dzearing, line 71: now there are 2 "firstfocusable" props. And wow. That other one is terrible (called a selector (rather than firstFocusableClassName), but not actually a selector...)

If we're going to make adjustments here, let's do this properly and deprecate the old one, and support the "right" approach. I'd recommend we call this `firstFocusableTarget`, which can take an actual query selector or a React.Ref, which we could do the right thing with. (analogous to `target` in ContextualMenu I think.)

packages/utilities/src/focus.ts
ThomasMichon: The browser doesn't seem to have any issue applying focus to natively-tabbable elements, just the custom-tabbable ones. Hence why I scoped this to only `IS_FOCUSABLE_ATTRIBUTE`.

khmakoto (dismissed):

azure-pipelines.release.yml
```diff
@@ -10,14 +10,15 @@ variables:
...
 
-schedules:
-  # minute 0, hour 12 in UTC (5am in UTC+7), any day of month, any month, days 1-5 of week (M-F)
-  # https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?tabs=yaml&view=azure-devops#supported-cron-syntax
-  - cron: '0 12 * * 1-5'
-    displayName: 'Daily release (M-F at 5am)' # will be 4am when DST ends unless trigger is updated
-    branches:
-      include:
-        - master
+# DISABLED DURING BETA PREP
```
ecraig12345, line 13: Can you revert this part? Looks like a bad merge (this was a change I made in master but shouldn't be in 7).

packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -158,7 +159,11 @@ export class FocusTrapZone extends React.Component<IFocusTrapZoneProps, {}> impl
     let _firstFocusableChild: HTMLElement | null = null;
 
     if (this._root.current) {
-      if (focusSelector) {
+      if (typeof firstFocusableTarget === 'string') {
```
dzearing, line 162: consider adding a test to make sure this doesn't regress.

packages/utilities/src/focus.ts
dzearing: I meant `isElementTabbable` in utilities.

dzearing (approved):
see 2 comments

khmakoto (approved):
I agree on adding a test but approved for now.

khmakoto:

ecraig12345:
Looks like some API updates are missing

PR #14613, opened 2020-08-19
Remove references to global React namespace and add lint rule against it

ecraig12345:
#### Pull request checklist

#### Description of changes

Accidental references to the `React` global namespace were causing problems for API Extractor and possibly for other tools. I think this might have been the cause of the weird issues @MLoughry was having running `update-api` in #14529, which I spent several hours debugging (and it's been a big time sink in other cases as well).

Fix is to write a new lint rule `no-global-react` to detect and ban implicit references, and fix all the existing ones by adding an explicit import.


packages/experiments/src/components/SelectedItemsList/SelectedPeopleList/Items/SelectedPersona.tsx
```diff
@@ -63,6 +63,7 @@ const SelectedPersonaInner = React.memo(
     } = props;
     const itemId = getId();
     const [root, setRoot] = React.useState<HTMLElement | undefined>();
+    const rootRef = React.useRef<HTMLElement>(null);
```
dzearing, line 66: undo this

PR #14616, opened 2020-08-20
Floating/Extended pickers: Remove deprecated react lifecycles 

### LINKED ISSUE #14293: ExtendedPicker: strict mode compliance (UNSAFE_, easy)
BaseExtendedPicker has a simple `componentWillReceiveProps` implementation which copies things straight from props to state or private members. This should be an easy fix: none of the things copied appear to be updated elsewhere, so it's likely the state/private copies could be removed and everywhere could be updated to just read from props.

Note that converting to a function component is probably not an option here (without larger architectural changes) since pickers currently rely on inheritance.

### LINKED ISSUE #14294: FloatingPicker: strict mode compliance (UNSAFE_, easy/medium)
FloatingPicker's related components have two obvious issues preventing strict mode compliance:

- BaseFloatingPicker's `componentWillReceiveProps` updates its suggestions store based on `props.suggestionItems`. This should be fairly easy to move elsewhere.
- SuggestionsControl copies `props.suggestions` to `state.suggestions` and then resets selected items. The update is a straight copy from props and `state.suggestions` does not appear to be updated elsewhere, so it should be fairly straightforward to fix.

Note that converting to a function component is probably not an option here (without larger architectural changes) since pickers currently rely on inheritance.

joschect:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #14294 
Fixes #14293

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

### DIFF
--- a/change/office-ui-fabric-react-2020-08-19-18-19-31-compliance-floatingpickers.json
+++ b/change/office-ui-fabric-react-2020-08-19-18-19-31-compliance-floatingpickers.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "packageName": "office-ui-fabric-react",
+  "email": "joschect@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-08-20T01:19:31.670Z"
+}
--- a/change/office-ui-fabric-react-2020-08-19-18-19-31-compliance-floatingpickers.json
+++ b/change/office-ui-fabric-react-2020-08-19-18-19-31-compliance-floatingpickers.json
@@ -1,5 +1,6 @@
 {
   "type": "patch",
+  "comment": "FloatingPickers: Remove deprecated methods",
   "packageName": "office-ui-fabric-react",
   "email": "joschect@microsoft.com",
   "dependentChangeType": "patch",
--- a/packages/office-ui-fabric-react/src/components/ExtendedPicker/BaseExtendedPicker.tsx
+++ b/packages/office-ui-fabric-react/src/components/ExtendedPicker/BaseExtendedPicker.tsx
@@ -14,6 +14,8 @@ export interface IBaseExtendedPickerState<T> {
   queryString: string | null;
   selectedItems: T[] | null;
   suggestionItems: T[] | null;
+  floatingPickerProps: IBaseFloatingPickerProps<T>;
+  selectedItemsListProps: IBaseSelectedItemsListProps<T>;
 }
 
 export class BaseExtendedPicker<T, P extends IBaseExtendedPickerProps<T>>
@@ -25,8 +27,28 @@ export class BaseExtendedPicker<T, P extends IBaseExtendedPickerProps<T>>
   protected root = React.createRef<HTMLDivElement>();
   protected input = React.createRef<Autofill>();
   protected selection: Selection;
-  protected floatingPickerProps: IBaseFloatingPickerProps<T>;
-  protected selectedItemsListProps: IBaseSelectedItemsListProps<T>;
+
+  public static getDerivedStateFromProps(newProps: IBaseExtendedPickerProps<any>) {
+    const updatedState: Partial<IBaseExtendedPickerProps<any>> = {};
+    let hasUpdate: boolean = false;
+
+    if (newProps.floatingPickerProps) {
+      updatedState.floatingPickerProps = newProps.floatingPickerProps;
+      hasUpdate = true;
+    }
+
+    if (newProps.selectedItemsListProps) {
+      updatedState.selectedItemsListProps = newProps.selectedItemsListProps;
+      hasUpdate = true;
+    }
+
+    if (newProps.selectedItems) {
+      updatedState.selectedItems = newProps.selectedItems;
+      hasUpdate = true;
+    }
+
+    return hasUpdate ? updatedState : null;
+  }
 
   constructor(basePickerProps: P) {
     super(basePickerProps);
@@ -44,10 +66,9 @@ export class BaseExtendedPicker<T, P extends IBaseExtendedPickerProps<T>>
         : this.props.selectedItems
         ? (this.props.selectedItems as T[])
         : null,
+      floatingPickerProps: this.props.floatingPickerProps,
+      selectedItemsListProps: this.props.selectedItemsListProps,
     };
-
-    this.floatingPickerProps = this.props.floatingPickerProps;
-    this.selectedItemsListProps = this.props.selectedItemsListProps;
   }
 
   public get items(): any {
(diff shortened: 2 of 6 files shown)


ecraig12345:
Can you add "Fixes ####" for the issues that this fixes? Especially right now it's important for tracking.

packages/office-ui-fabric-react/src/components/ExtendedPicker/BaseExtendedPicker.tsx
```diff
@@ -14,6 +14,8 @@ export interface IBaseExtendedPickerState<T> {
   queryString: string | null;
   selectedItems: T[] | null;
   suggestionItems: T[] | null;
+  floatingPickerProps: IBaseFloatingPickerProps<T>;
```
ecraig12345, line 17: Since you have the getters, I don't think you really need these in state.

packages/office-ui-fabric-react/src/components/ExtendedPicker/BaseExtendedPicker.tsx
```diff
@@ -25,8 +27,28 @@ export class BaseExtendedPicker<T, P extends IBaseExtendedPickerProps<T>>
...
+      updatedState.floatingPickerProps = newProps.floatingPickerProps;
+      hasUpdate = true;
+    }
+
+    if (newProps.selectedItemsListProps) {
+      updatedState.selectedItemsListProps = newProps.selectedItemsListProps;
+      hasUpdate = true;
+    }
+
+    if (newProps.selectedItems) {
```
ecraig12345, line 45: As far as I can tell, `selectedItems` is never updated in state--so could you just get rid of the state member (and read directly from props), or would that break child classes?

packages/office-ui-fabric-react/src/components/SelectedItemsList/BaseSelectedItemsList.tsx
```diff
@@ -6,27 +6,44 @@ import { initializeComponentRef } from '../../Utilities';
 
 export interface IBaseSelectedItemsListState<T = any> {
   items: T[];
+  selection: Selection;
```
ecraig12345, line 9: Similar here, I don't see why this needs to be in state since it never changes.

joschect: I think it needs to be in state because  it was potentially changed in componentWillReceiveProps. I agree that it doesn't seem like it needs to be there but I wanted to try to preserve it as is to help reduce friction for this checkin.

packages/office-ui-fabric-react/src/components/ExtendedPicker/BaseExtendedPicker.tsx
joschect: I believe that it might break child classes which is part of why I'm doing the weird getter hybrid with props. It's not ideal but I don't know how to go forward without it being a breaking change in some other way.

joschect:
Any last minute comments?

ecraig12345:
Sorry to be so stubborn about this, but I really don't think you should add things to state (earlier comments). Copying things directly from props to state is well-established as an anti-pattern, so we shouldn't be adding new instances of it. Especially since there are other fairly straightforward ways of handling this situation.

packages/office-ui-fabric-react/src/components/ExtendedPicker/BaseExtendedPicker.tsx
```diff
@@ -39,39 +48,23 @@ export class BaseExtendedPicker<T, P extends IBaseExtendedPickerProps<T>>
...
 
   public get items(): any {
-    return this.state.selectedItems ?? this.selectedItemsList.current?.items ?? null;
+    if (this.props.selectedItems !== null && this.props.selectedItems !== undefined) {
+      return this.props.selectedItems;
+    }
+    if (this.selectedItemsList.current) {
+      return this.selectedItemsList.current.items;
+    }
+    return this.props.defaultSelectedItems;
```
ecraig12345, line 61: Nitpicky thing but you could optionally change this back to the nullish coalescing syntax (it does the same thing as these checks). `this.props.selectedItems ?? this.selectedItemsList.current?.items ?? this.props.defaultSelectedItems`

packages/office-ui-fabric-react/src/components/ExtendedPicker/BaseExtendedPicker.tsx
```diff
@@ -25,8 +24,18 @@ export class BaseExtendedPicker<T, P extends IBaseExtendedPickerProps<T>>
   protected root = React.createRef<HTMLDivElement>();
   protected input = React.createRef<Autofill>();
   protected selection: Selection;
-  protected floatingPickerProps: IBaseFloatingPickerProps<T>;
-  protected selectedItemsListProps: IBaseSelectedItemsListProps<T>;
+
+  public static getDerivedStateFromProps(newProps: IBaseExtendedPickerProps<any>) {
```
ecraig12345, line 28: I think this should be removed now that you've removed `selectedItems` from state?

packages/office-ui-fabric-react/src/components/FloatingPicker/BaseFloatingPicker.tsx
```diff
@@ -157,6 +151,9 @@ export class BaseFloatingPicker<T, P extends IBaseFloatingPickerProps<T>>
 
   protected renderSuggestions(): JSX.Element | null {
     const TypedSuggestionsControl = this.SuggestionsControlOfProperType;
+    if (this.props.suggestionItems) {
+      this.suggestionStore.updateSuggestions(this.props.suggestionItems!);
```
ecraig12345, line 155: Depending on how expensive this is, you might not want to do it every render. A potential alternative would be either memoizing the `updateSuggestions` call, or adding something like this to `componentDidUpdate`. (Any re-renders done here will be flushed before the displayed DOM is updated.)
```ts
if (this.props.suggestionItems && prevProps.suggestionItems !== this.props.suggestionItems) {
  this.suggestionStore.updateSuggestions(this.props.suggestionItems);
  this.forceUpdate();
}
```

packages/office-ui-fabric-react/src/components/SelectedItemsList/BaseSelectedItemsList.tsx
```diff
@@ -140,32 +145,27 @@ export class BaseSelectedItemsList<T, P extends IBaseSelectedItemsListProps<T>>
...
-      this.selection = newProps.selection;
-    }
-  }
-
   public render(): any {
     return this.renderItems();
   }
 
+  protected get selection(): Selection {
+    if (this.props.selection !== null && this.props.selection !== undefined) {
```
ecraig12345, line 163: You could do this with nullish coalescing (or just ||) too. `return this.props.selection || this._defaultSelection`

change/office-ui-fabric-react-2020-08-19-18-19-31-compliance-floatingpickers.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
ecraig12345, line 2: This is missing a comment somehow?

packages/office-ui-fabric-react/src/components/ExtendedPicker/BaseExtendedPicker.tsx
joschect: Good call! I had mean to remove that and missed it.

packages/office-ui-fabric-react/src/components/FloatingPicker/BaseFloatingPicker.tsx
joschect: This really isn't too expensive and just makes sure everything is in sync. I think that this should stay here.

joschect:
Closing this. I'll reopen with all the changes made in react next since changes to this type of component could lead to breaks if inheritance is used down stream.

PR #14634, opened 2020-08-20
chore: remove codepen script

layershifter:
The goal of this PR is to remove `ngrok` dependency as it often causes build failures on `yarn install`. `codepen` script is a single place where this dependency is used.


ecraig12345:
The purpose of the `codepen` script is to let you preview local changes to v7 in codepen. I don't think people use it super often but it is nice to have sometimes (like if you're fixing a bug which came with a codepen repro). So I'd rather it not be removed without an alternative.

PR #14639, opened 2020-08-20
react-cards: Updating imports to remove duplicated dependencies

### LINKED ISSUE #14605: Consider recategorizing packages for @uifabric/react-cards

I'm asking the Fabric team to consider the following changes/questions for dependencies in the @uifabric/react-cards package.

- Move `office-ui-fabric-react` to a peerDependency.  This can help us have just one version of Fabric in our project
- If we're already importing office-ui-fabric-react in @uifabric/react-cards then can some of the @uifabric/* imports be replaced with imports from the fabric directly, so we can further reduce the dependencies?
- Can @uifabric/azure-themes and @uifabric/theme-samples be removed as dependencies (maybe put in optional or dev dependencies instead?) As far as I know they are used only a demo capacity

@uifabric/react-cards package

No

N/A

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14605

#### Description of changes

This PR does a couple of things:
- Moves `@uifabric/azure-themes` and `@uifabric/theme-samples` from `dependencies` to `devDependencies`.
- Changes the imports from `@uifabric/styling` and `@uifabric/utilities` to import from `office-ui-fabric-react` and removes those packages from `package.json`.

#### Focus areas to test

(optional)

### DIFF
--- a/change/@uifabric-react-cards-2020-08-20-11-57-01-cardImports.json
+++ b/change/@uifabric-react-cards-2020-08-20-11-57-01-cardImports.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "react-cards: Updating imports to remove duplicated dependencies.",
+  "packageName": "@uifabric/react-cards",
+  "email": "humbertomakotomorimoto@gmail.com",
+  "dependentChangeType": "patch",
+  "date": "2020-08-20T18:57:01.750Z"
+}
--- a/packages/react-cards/package.json
+++ b/packages/react-cards/package.json
@@ -35,9 +35,11 @@
     "@types/react-dom": "16.8.4",
     "@types/react-test-renderer": "^16.0.0",
     "@types/webpack-env": "1.15.1",
+    "@uifabric/azure-themes": "^7.4.1",
     "@uifabric/build": "^7.0.0",
     "@uifabric/example-app-base": "^7.14.0",
     "@uifabric/jest-serializer-merge-styles": "^7.0.36",
+    "@uifabric/theme-samples": "^7.1.1",
     "enzyme": "~3.10.0",
     "enzyme-adapter-react-16": "^1.15.0",
     "react": "16.8.6",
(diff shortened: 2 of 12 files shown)


packages/react-cards/package.json
```diff
@@ -47,19 +49,15 @@
...
-    "@uifabric/styling": "^7.15.1",
-    "@uifabric/theme-samples": "^7.1.1",
-    "@uifabric/utilities": "^7.29.1",
-    "office-ui-fabric-react": "^7.129.1",
     "tslib": "^1.10.0"
   },
   "peerDependencies": {
     "@types/react": ">=16.8.0 <17.0.0",
     "@types/react-dom": ">=16.8.0 <17.0.0",
+    "office-ui-fabric-react": "^7.129.1",
```
ecraig12345, line 60: You should add OUFR to devDependencies if you take this approach. Also the peerDependencies range is typically wider, and I don't think it will be automatically updated by beachball.

change/@uifabric-react-cards-2020-08-20-11-57-01-cardImports.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
ecraig12345, line 2: Maybe make this minor? (I'm not sure what the standard way to handle this case is)
```suggestion
  "type": "minor",
```

khmakoto: I think it shouldn't be minor since we are not changing the public facing API?

packages/react-cards/src/components/Card/Card.styles.ts
```diff
@@ -1,4 +1,4 @@
-import { getGlobalClassNames, HighContrastSelector } from '@uifabric/styling';
+import { getGlobalClassNames, HighContrastSelector } from 'office-ui-fabric-react';
```
ecraig12345, line 1: You might want to import these from `oufr/lib/Styling` to reduce bundle size impact for consumers. (In theory there should be no impact, but with some bundling setups it can make a difference.)

Same in other places.

change/@uifabric-react-cards-2020-08-20-11-57-01-cardImports.json
ecraig12345: I sometimes also use minor for potentially-more-impactful changes even if they don't change the API. Up to you.

packages/react-cards/src/components/Card/Card.styles.ts
ecraig12345: Actually never mind, since you're already importing from the root elsewhere it's probably fine to just leave this as-is.

khmakoto:

PR #14640, opened 2020-08-20
GroupedList: Fixing errors in the examples that are triggered when the tsconfig strictFunctionTypes rule is turned on

### LINKED ISSUE #14604: Example from GroupedList document cannot be compiled
### Environment Information

- **Package version(s)**: "
"@fluentui/react": "^7.128.1
"typescript": "~3.7.2"

[exmpale](https://codepen.io/ytingyeu/pen/rNeMOWN)
Note: this example is taken from [documentation](https://developer.microsoft.com/en-us/fluentui#/controls/web/groupedlist)

The error is not visible on codepen. One has to test this locally where typescript strictFunctionTypes is applied

#### Actual behavior:
Typescript compiler argues a type error. Set tsconfig `strictFunctionTypes` to false can bypass this error

```
Type '(nestingDepth: number, item: IExampleItem, itemIndex: number) => JSX.Element' is not assignable to type '(nestingDepth?: number | undefined, item?: any, index?: number | undefined) => ReactNode'.
  Types of parameters 'nestingDepth' and 'nestingDepth' are incompatible.
    Type 'number | undefined' is not assignable to type 'number'.
      Type 'undefined' is not assignable to type 'number'.ts(2322)
GroupedList.types.d.ts(77, 5): The expected type comes from property 'onRenderCell' which is declared here on type 'IntrinsicAttributes & IGroupedListProps & { children?: ReactNode; }'
```

#### Expected behavior:
Should have no error

Requested priority: Normal

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14604

#### Description of changes

This PR fixes the errors that were arising in the `GroupedList` component examples when the `strictFunctionTypes` rule was turned on in the `tsconfig`.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/src/components/GroupedList/examples/GroupedList.CustomCheckbox.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/GroupedList/examples/GroupedList.CustomCheckbox.Example.tsx
@@ -6,13 +6,12 @@ import {
   IGroupHeaderProps,
   IGroupRenderProps,
 } from 'office-ui-fabric-react/lib/GroupedList';
-import { IColumn, DetailsRow } from 'office-ui-fabric-react/lib/DetailsList';
+import { IColumn, IObjectWithKey, DetailsRow } from 'office-ui-fabric-react/lib/DetailsList';
 import { FocusZone } from 'office-ui-fabric-react/lib/FocusZone';
 import { ISelection, Selection, SelectionMode, SelectionZone } from 'office-ui-fabric-react/lib/Selection';
 import { Toggle } from 'office-ui-fabric-react/lib/Toggle';
 import { useConst } from '@uifabric/react-hooks';
 import { createListItems, createGroups, IExampleItem } from '@uifabric/example-data';
-import { IObjectWithKey } from '../../DetailsList';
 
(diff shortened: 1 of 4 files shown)


packages/office-ui-fabric-react/src/components/GroupedList/examples/GroupedList.CustomCheckbox.Example.tsx
```diff
@@ -39,19 +42,20 @@ export const GroupedListCustomCheckboxExample: React.FunctionComponent = () => {
         }),
       ),
   );
-  const selection = useConst(() => new Selection({ items }));
+  const selection = (useConst(() => new Selection({ items })) as unknown) as ISelection<IObjectWithKey>;
```
ecraig12345, line 45: Curious why this was needed? Seems like it ought to be inferred correctly, or if not maybe there's some way to specify a generic to get the proper result by default.

khmakoto: It is definitely not inferred correctly. I'll fiddle a bit more with it.

PR #14647, opened 2020-08-20
Fix negative date navigation crashing the page

### LINKED ISSUE #14541: Calendar underflow to year -500 and below, engine crash
### Environment Information

- **Package version(s)**: not applicable
- **Browser and OS versions**: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.108 Safari/537.36 (Vivaldi Browser)

See https://developer.microsoft.com/en-us/fluentui#/controls/web/calendar

#### Actual behavior:

1. Scroll to the "calendar with year picker" demo
2. Enter the year selection screen
3. Keep clicking ↑ until you see the years go negative
4. Click any year
5. Month view is filled with "NaN" days and the JS engine crashes
image

#### Expected behavior:

Year selection screen should not allow negative years to be selected.

Requested priority: Low

Products/sites affected: Any React app using the Calendar control and internal products.

lorejoh12:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14541 

#### Description of changes

Apparently there's an issue with default Javascript dates where you can create a date, but if you try to clone it using new Date(date.toString()), for certain dates that will fail (return an Invalid Date object instead). For example, dates with negative years will fail, as the bug notes. This issue was present for both date-time and OUFR Calendar components.

#### Focus areas to test

Open the inline date picker, switch the month picker to be a year picker (to go a little faster), use the back arrow to go backwards all the way back to negative dates, and pick a month. In production, we get stuck in a loop until the page crashes. With the change, everything works as expected.

Wasn't able to figure out a unit test for this- the same bug that causes the issue in the first place, where Dates with negative years get the toString() messed up, also affects unit tests. When creating a date new Date('-000377-06-20T00:00:00.000Z'), which should be Jun 20 -0377 at midnight GMT, this actually ends up printing to the jest snapshot as "Mon Jun 19 -0377 16:07:02 GMT-0752 (Pacific Daylight Time)". Notice the weird GMT-0752, whatever logic is underflowing here is trying to put the hours offset into the time zone rather than the actual time. Changing the machine timezone to UTC causes it to print as "Tue Jun 20 -0377 00:00:00 GMT+0000 (Greenwich Mean Time)", exactly as you'd expect. When machines in different time zones print this information out to compare the snapshots, they end up not matching. So, I won't add a unit test for this case.


packages/date-time-utilities/src/dateGrid/getDayGrid.ts
```diff
@@ -34,7 +34,7 @@ export const getDayGrid = (options: IDayGridOptions): IDay[][] => {
   let date;
   if (weeksToShow && weeksToShow <= 4) {
     // if showing less than a full month, just use date == navigatedDate
-    date = new Date(navigatedDate.toString());
+    date = new Date(navigatedDate.getFullYear(), navigatedDate.getMonth(), navigatedDate.getDate());
```
pompomon, line 37: Please, why don't you do  
```suggestion
    date = new Date(navigatedDate);
```

like in other places?

pompomon (approved):

packages/date-time-utilities/src/dateGrid/getDayGrid.ts
lorejoh12: In this case, it's an actual bug fix. See right below in the else case. This is supposed to create dates for the grid starting at midnight, rather than at the same time as whatever the navigatedDate is.

PR #14648, opened 2020-08-20
do not merge: Variant support in components

### LINKED ISSUE #14669: Variants PR reviewed and checked in

dzearing:

## Variants support

Variants are defined as such:

```jsx
  name: 'Foo',
  prefix: '--foo', 
  styles: {
    root: { ... },
    icons: { ... },
  },
  variants: {
    root: { 
      background: 'red'
    },
    primary: { 
      background: 'green'
    }
  }
});
```

```jsx
const Foo = (props) => {
  const state = { ...props };

  useFooClasses (state);

  return (
  );
```

```jsx
<Foo /> // red foo
<Foo primary /> // green foo
```

```jsx
<ThemeProvider theme={{
  components: {
     Foo: {
       variants: {
          primary: {
             background: 'purple'
          }
        }
      }
    }
  }}>
  <Foo/> // red button
</ThemeProvider>
```


packages/react-theme-provider/src/getTokens.ts
```diff
@@ -68,168 +68,8 @@ export function getTokens(theme: Theme): Tokens {
           },
         },
       },
-
-      body: {
```
dzearing, line 72: dont remove

packages/react-button/src/components/Button/Button.types.ts
```diff
@@ -89,61 +91,62 @@ export type ButtonProps = ComponentProps &
...
 
-export interface ButtonSlots extends BaseSlots {
-  icon: React.ElementType;
-  loader: React.ElementType;
-}
-
-export type ButtonSlotProps = SlotProps<ButtonSlots, ButtonProps, React.ButtonHTMLAttributes<HTMLButtonElement>>;
-
-export type ButtonTokenSet = ColorTokenSet & {
+export type ButtonTokens = ColorTokens & {
```
xugao, line 101: typing doesn't look right; if you compare this with what we have in `useButtonVariants`

packages/react-button/src/components/Button/useButtonClasses.ts
```diff
@@ -137,79 +137,6 @@ export const useButtonClasses = makeClasses<ButtonState>({
...
-  _circular: {
-    '--button-borderRadius': '50000px',
-  },
-
-  _fluid: {
-    '--button-width': '100%',
-    '--button-maxWidth': '100%',
-  },
-
   _size_smallest: {
```
xugao, line 140: we should move these too?

packages/react-button/src/components/Button/useButtonVariants.ts
```diff
@@ -0,0 +1,175 @@
...
+        smaller: '24px',
+        small: '24px',
+        regular: '32px',
+        large: '40px',
+        larger: '48px',
+        largest: '64px',
+      },
+
+      // Sizing tokens
+      iconSize: fonts?.mediumPlus?.fontSize,
```
xugao, line 20:
```suggestion
      iconSize: fonts.mediumPlus.fontSize,
```

we should not need `?` everywhere

packages/react-theme-provider/src/makeVariants.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+            .first()
+            .getDOMNode()
+            .getAttribute('class'),
+        ).toEqual('Button-0');
+        expect(_stylesheet.getRules()).toEqual('.Button-0{--button-background:red;}');
+      },
+    );
+  });
+
+  it('can render primary variant', () => {
```
xugao, line 61:
```suggestion
  it('can render modifier variant', () => {
```

packages/react-theme-provider/src/makeVariants.ts
```diff
@@ -0,0 +1,79 @@
...
+  state: GenericDictionary,
+  variantKeys: string[],
+  variantObjects: TokenSetType[],
+) => {
+  if (variants) {
+    for (const variantName of Object.keys(variants)) {
+      const isBase = variantName === 'base';
+      if (variants[variantName] && (isBase || state[variantName] || state.variant === variantName)) {
+        variantKeys.push(isBase ? componentName : variantName);
+        variantObjects.push(variants[variantName]!);
```
xugao, line 28:
```suggestion
        variantObjects.push(variants[variantName]);
```

packages/react-theme-provider/etc/react-theme-provider.api.md
```diff
@@ -1,151 +1,157 @@
...
+export const createDefaultTheme: () => Theme;
+
+// @public (undocumented)
+export const FluentTheme: Theme;
+
+// @public (undocumented)
+export type FontFace = IFontFace;
+
+// @public (undocumented)
+export type GenericDictionary = Record<string, any>;
```
xugao, line 30: i don't think we should export this

packages/react-button/src/components/Button/useButtonVariants.ts
xugao: pls sync code with latest code from `getTokens`

packages/theme/src/types/Theme.ts
```diff
@@ -4,18 +4,18 @@ import { IStyleFunctionOrObject } from '@uifabric/utilities';
...
-  secondaryContentColor: string;
-  linkColor: string;
-  iconColor: string;
-  borderColor: string;
-  dividerColor: string;
-  focusColor: string;
-  focusInnerColor: string;
-  opacity: string;
-}>;
+export type ColorTokenSet = {
```
xugao, line 7: why not using `Partial`?

packages/theme/src/types/Theme.ts
```diff
@@ -37,16 +37,14 @@ export type ColorTokens = Partial<{
...
-}>;
-
-export type ColorTokenSet = ColorTokens & ColorTokenStates;
+export type ColorTokens = ColorTokenSet & {
+  hovered?: ColorTokens;
+  pressed?: ColorTokens;
+  disabled?: ColorTokens;
+  checked?: ColorTokens;
+  checkedHovered?: ColorTokens;
+  checkedPressed?: ColorTokens;
```
xugao, line 46: why not using `Partial` (what we had before)?

packages/theme/etc/theme.api.md
```diff
@@ -729,6 +730,9 @@ export type TokenSetType = {
     [key: string]: TokenSetType | string | number | undefined;
 };
 
+// @public (undocumented)
+export type Variants = Record<string, any>;
```
xugao, line 734: Variants should always contain base, something like:

```
interface Variants {
  base?: RecursivePartial<Tokens>;
  [key: string]: RecursivePartial<Tokens>;
} 
```

packages/react-theme-provider/src/makeVariants.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export const makeVariants = <TTokenSetType extends TokenSetType>(
+  componentName: string,
+  prefix: string,
+  variants: Variants | ((theme: Theme) => Variants),
+) => {
+  const variantToClassName: Record<string, string> = {}; // useThemeSettings(id);
+
+  return (state: GenericDictionary) => {
+    // Grab the theme.
```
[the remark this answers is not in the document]
dzearing: done

packages/react-button/src/components/Button/Button.types.ts
```diff
@@ -89,61 +91,62 @@ export type ButtonProps = ComponentProps &
...
+  borderWidth?: string;
+  boxShadow?: string;
+  width?: string;
+
+  transform?: string;
+  transition?: string;
+
+  fontFamily?: string;
+  fontSize?: string;
+  fontWeight?: string;
```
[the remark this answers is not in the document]
dzearing: done

packages/react-button/src/components/Button/useButtonClasses.ts
dzearing: done.

packages/theme/src/types/Theme.ts
dzearing: It seemed more ergonomic.

packages/theme/etc/theme.api.md
dzearing: I think this we should leave as is. I don't think base (changed to root) is required.

packages/react-theme-provider/src/makeVariants.ts
dzearing: no longer applicable.

.vscode/settings.json
```diff
@@ -15,7 +15,7 @@
     "**/.git": true,
     "**/.DS_Store": true,
     "**/coverage": true,
-    "**/*.scss.ts": true,
+    "**/*.scss.ts": false,
```
dzearing, line 18: revert

.vscode/launch.json
```diff
@@ -14,6 +14,7 @@
       "env": {
         "NODE_ENV": "development"
       },
+      "outFiles": ["${workspaceFolder}/lib-commonjs/**/*"],
```
dzearing, line 17: revert

PR #14650, opened 2020-08-20
Scaffold @fluentui/theme package

xugao:
#### Pull request checklist

#### Description of changes
`@fluentui/theme` package is meant to contain basic building blocks for themes. 
- moved some constants and Theme related typings from `@uifabric/styling` and `@uifabric/fluent-theme` to this new package. (there are more theme related code in `styling` we can move out. but would be better to stage that in next PR)
- moved _new_ Theme related typings and `mergeThemes` from `react-theme-provider` to this new package

#### Focus areas to test

(optional)


packages/fluent-theme/package.json
```diff
@@ -26,6 +26,7 @@
     "@uifabric/build": "^7.0.0"
   },
   "dependencies": {
+    "@fluentui/theme": "^0.1.0",
```
dzearing, line 29: it seems like the directory should just be `theme` if we call it `@fluentui/theme`.

packages/theme/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@fluentui/theme",
+  "version": "0.1.0",
+  "description": "Basic building blocks for Fluent UI React Component themes",
```
dzearing, line 4:
```suggestion
  "description": "Basic building blocks for Fluent UI themes.",
```

JasonGore:
I've been troubleshooting a runtime exception unit tests in our repo:

image

This seems to be related to `@uifabric/fluent-theme` ver `7.2.3` and does not appear to happen with `7.1.130`. I wonder if due to recent changes `FluentTheme` may be used internally before it is initialized statically.

PR #14656, opened 2020-08-21
Feat/adding more datepicker examples

rymeskar:
#### Pull request checklist

#### Description of changes

Adding more datepicker examples.

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -268,6 +274,18 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+        const keyCode = getCode(e);
+
+        if (keyCode === keyboardKey.Enter) {
+          setPreventClosing(true);
+          setOpenState(true);
+        }
+
+        _.invoke(predefinedProps, 'onKeyPress', e, predefinedProps);
+      }
+    },
```
pompomon, line 288: Let's move all keyPress handlers to appropriate behaviors, e.g. in https://github.com/microsoft/fluentui/blob/a4898636dfef66f59f63740ab2646c7497410137/packages/fluentui/accessibility/src/behaviors/Menu/menuItemBehavior.ts#L46

packages/fluentui/accessibility/src/behaviors/Datepicker/datepickerCalendarCellBehavior.ts
```diff
@@ -10,11 +10,15 @@ export const datepickerCalendarCellBehavior: Accessibility<DatepickerCalendarCel
   attributes: {
     root: {
       'aria-selected': props.selected || false,
```
mnajdova, line 12:
```suggestion
      'aria-selected': !!props.selected,
```

I know it's not part of these changes, but let's change it

packages/fluentui/docs/src/examples/components/Datepicker/Types/DatepickerExampleLocalizationStrings.shorthand.tsx
```diff
@@ -1,11 +1,11 @@
...
 import * as React from 'react';
 
-const DatepickerExampleLocalizationStrings = () => (
-  <Datepicker
-    shortDays={['أحد', 'اثنين', 'ثلاثاء', 'أربعاء', 'خميس', 'جمعة', 'سبت']}
-    days={['أحد', 'اثنين', 'ثلاثاء', 'أربعاء', 'خميس', 'جمعة', 'سبت']}
-    months={[
+const DatepickerExampleLocalizationStrings = () => {
+  const localizationStrings = {
+    shortDays: ['أحد', 'اثنين', 'ثلاثاء', 'أربعاء', 'خميس', 'جمعة', 'سبت'],
```
mnajdova, line 6: Just a thought, would be nice if all these props in the `Datepicker` are ending with `String` or `Label` suffix, so it would be obvious they are just strings. to be honest `goToToday` for example is not associating me that it is a string.

packages/fluentui/docs/src/examples/components/Datepicker/Usage/DatepickerExampleStandaloneCalendarInput.shorthand.tsx
```diff
@@ -0,0 +1,18 @@
+import { Datepicker } from '@fluentui/react-northstar';
+import * as React from 'react';
+
+const DatepickerExampleStandaloneCalendarInput = () => {
+  const inputRef = React.useRef(null);
+  return (
+    <>
```
mnajdova, line 7:
```suggestion
```

packages/fluentui/docs/src/examples/components/Datepicker/Usage/DatepickerExampleStandaloneCalendarInput.shorthand.tsx
```diff
@@ -0,0 +1,18 @@
...
+      <Datepicker
+        input={{ ref: inputRef }}
+        popup={{
+          trigger: null,
+          target: inputRef.current,
+        }}
+      />
+    </>
```
mnajdova, line 15:
```suggestion
```

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -170,13 +144,59 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
+    ...restrictedDatesOptions,
+  };
+
   const ElementType = getElementType(props);
   const unhandledProps = useUnhandledProps(Datepicker.handledProps, props);
   const getA11yProps = useAccessibility(props.accessibility, {
     debugName: Datepicker.displayName,
-    actionHandlers: {},
+    actionHandlers: {
+      open: e => {
```
mnajdova, line 193: Looks strange that on click we are alternating between opening and closing it, but with keyboard we are always opening it. Is that really by design? Is it because the focused is moved when opened?

packages/fluentui/react-northstar/src/components/Datepicker/navigateToNewDate.tsx
```diff
@@ -0,0 +1,76 @@
...
+  if (isAfterMaxDate(targetDate, restrictedDatesOptions)) {
+    newNavigatedDate = restrictedDatesOptions.maxDate;
+  } else if (isBeforeMinDate(targetDate, restrictedDatesOptions)) {
+    newNavigatedDate = restrictedDatesOptions.minDate;
+  }
+
+  return newNavigatedDate;
+};
+
+export const navigateToNewDate = (
```
mnajdova, line 49: Can we add tests for this?

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Actually, the behavior should be consistent. We keep the calendar open if the manualInput is disabled for both on click and on enter.

If the manualInput is enabled, then we don't handle the enter action in any way.

packages/fluentui/docs/src/examples/components/Datepicker/Types/DatepickerExampleLocalizationStrings.shorthand.tsx
rymeskar: That's a great point! 

We are actually having further discussions about whether the `localizationStrings` should be flattened in the DatepickerProps or not. Currently, they are flattened, so within the ICalendarStrings, it might make sense to not have the suffix but in DatepickerProps, it actually becomes confusing.

I propose to bring up this issue again in the planned v1 task to revisit the approach of localizationStrings. 

WDYT? 

FYI: @pompomon @layershifter

packages/fluentui/react-northstar/src/components/Datepicker/navigateToNewDate.tsx
rymeskar: Great point, added :)

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -82,6 +83,12 @@ export interface DatepickerProps extends UIComponentProps, Partial<ICalendarStri
 
   /** Controls the calendar's open state. */
   calendarOpenState?: boolean;
+
+  /** Controls whether there is a date that is selected by default. */
```
mnajdova, line 87:
```suggestion
  /** Initial selectedDate value. */
```

This would be in my opinion closer to the description of the `default`s that we have in other components.

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -170,13 +144,59 @@ export const Datepicker: ComponentWithAs<'div', DatepickerProps> &
...
   const ElementType = getElementType(props);
   const unhandledProps = useUnhandledProps(Datepicker.handledProps, props);
   const getA11yProps = useAccessibility(props.accessibility, {
     debugName: Datepicker.displayName,
-    actionHandlers: {},
+    actionHandlers: {
+      open: e => {
+        if (!allowManualInput) {
+          e.preventDefault();
+          setOpenState(true);
```
mnajdova, line 196: The difference I see is that here we are always setting it to true ` setOpenState(true);`, which is ok as the name suggests that, but on ther other hand onClick we have `setOpenState(!openState);` This was what my question was about

packages/fluentui/react-northstar/test/specs/components/Datepicker/navigateToNewDate-test.tsx
```diff
@@ -0,0 +1,32 @@
+import { navigateToNewDate } from 'src/components/Datepicker/navigateToNewDate';
+import { addDays, addMonths, addWeeks, compareDatePart } from '@fluentui/date-time-utilities';
+
+describe('navigateToNewDate', () => {
+  const referenceDate = new Date(2020, 0, 10);
```
mnajdova, line 5: Is this a valid date with `0`?

packages/fluentui/react-northstar/test/specs/components/Datepicker/navigateToNewDate-test.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  it('Navigation works for day', () => {
+    const result = navigateToNewDate(referenceDate, 'Day', 1, {});
+    expect(compareDatePart(result, addDays(referenceDate, 1))).toBeFalsy();
```
mnajdova, line 9: Let's use dates here instead of calling other function, that way we can be sure it won't regress in the future.

packages/fluentui/react-northstar/test/specs/components/Datepicker/navigateToNewDate-test.tsx
rymeskar: Yeah, counter-inuitevely, the months are 0-indexed.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
rymeskar: Both of them are guided by the allowManualInput property.

This leads to the same [behavior](https://github.com/microsoft/fluentui/blob/54e53325fd3bba8381b7210aa96842e13d886a68/packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx#L239).

packages/fluentui/accessibility/src/behaviors/Datepicker/datepickerCalendarCellBehavior.ts
rymeskar: @mnajdova 
After the changes, I got the following error in pipeline. Should I then return the previous code?
```
datepickerCalendarCellBehavior.ts › Adds attribute 'aria-selected' based on the property 'selected' to 'root' slot.
ERR! 
ERR!     expect(received).toEqual(expected) // deep equality
ERR! 
ERR!     Expected: "value of property"
ERR!     Received: true
ERR! 
ERR!       108 |     property[propertyDependingOn] = propertyDependingOnValue;
ERR!       109 |     const expectedResult = parameters.behavior(property).attributes[elementWhereToBeAdded][attributeToBeAdded];
ERR!     > 110 |     expect(testHelper.convertToMatchingTypeIfApplicable(expectedResult)).toEqual(
ERR!           |                                                                          ^
ERR!       111 |       testHelper.convertToMatchingTypeIfApplicable(propertyDependingOnValue),
ERR!       112 |     );
ERR!       113 |   },
ERR! 
ERR!       at Object.testMethod (test/behaviors/testDefinitions.ts:110:74)
ERR!       at Object.test (test/behaviors/testHelper.tsx:60:24)
```

mnajdova: Let's not tackle it now, it was not part of this PR anyway

mnajdova (approved):
See latest comments before merging

PR #14657, opened 2020-08-21
Charting: Duplicating 1k in Y axis in all charts resolved

jameelakowsar:
#### Pull request checklist

#### Description of changes

Added tickFormat('.2s') to Y axis scale in all charts.

#### Focus areas to test
Line chart
Area chart
Vertical bar chart
Vertical stacked bar chart
Grouped vertical bar chart

#### After fix
image

#### Before fix
image


mbest:
This doesn't seem like a good change. Why should 0 always be formatted as `0.0`?

image

ecraig12345:
@mbest Guessing it was an unintended side effect. Would you mind [filing an issue](https://github.com/microsoft/fluentui/issues/new?template=bug_report.md) about that?

mbest:
#14811

PR #14687, opened 2020-08-21
Removing MsHighContrastAdjust:none from MessageBar and CompoundButton

### LINKED ISSUE #14655: Components using MSHighContrastAdjust:none dont override their children's color style
If any component which uses the HighContrastSelector has a child with a self-set color, that self-set color won't be overridden in HighContrast mode, **even if in the parent component it sets the color within the HighContrastSelector**

### Environment Information

- **Package version(s)**: @fluentui/react ^7.117.1, @uifabric/styling ^7.12.15
- **Browser and OS versions**: Version 81.0.410.1 (Official build) Dev (64-bit) (repros on other browsers)

https://codepen.io/ibrosen/pen/BaKQWrg

image

```
<MessageBar>
  <span style={{color:"red"}}> text </span>
</MessageBar>
```

Even though MessageBar.styles.ts sets the color to "windowtext", it's being overridden 
image

Even in HighContrastMode, the "text" will still be red, due to MessageBar's usage of the HighContrastSelector

#### Examples
- [MessageBar](https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts)
    - It sets `color:"windowtext"`, but that's overridden by children's self-set color
- Same issue in [CompoundButton](https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/Button/CompoundButton/CompoundButton.styles.ts)

#### Actual behavior:

Any child of a component which uses @uifabric/styling's HighContrastSelector won't have its color overridden, if it sets its own color.

#### Expected behavior:

If a component sets a HighContrastMode color on text, it'll override the colour of any regular-mode text within that component, or its children.

Are you willing to submit a PR to fix? **yes, working on one atm**
Requested priority: **Normal**
Products/sites affected: **account.microsoft.com**

ibrosen:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14655
    - Ran `yarn change`, it said no change file was needed

#### Description of changes

Removed MsHighContrastAdjust:none from MessageBar and CompoundButton

#### Focus areas to test

The main open question is : **why were they set to none in the first place?**

See attached issue for more information, but it means that children of these components weren't having their text color overridden in HighContrastMode

### DIFF
--- a/packages/office-ui-fabric-react/src/components/Button/CompoundButton/CompoundButton.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/Button/CompoundButton/CompoundButton.styles.ts
@@ -81,7 +81,6 @@ export const getStyles = memoizeFunction(
           [HighContrastSelector]: {
             backgroundColor: 'WindowText',
             color: 'Window',
-            MsHighContrastAdjust: 'none',
           },
         },
       },
@@ -103,7 +102,6 @@ export const getStyles = memoizeFunction(
           [HighContrastSelector]: {
             color: 'Window',
             backgroundColor: 'WindowText',
-            MsHighContrastAdjust: 'none',
           },
         },
       },
@@ -115,7 +113,6 @@ export const getStyles = memoizeFunction(
           [HighContrastSelector]: {
             color: 'Window',
             backgroundColor: 'WindowText',
-            MsHighContrastAdjust: 'none',
           },
         },
       },
--- a/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
+++ b/packages/office-ui-fabric-react/src/components/MessageBar/MessageBar.styles.ts
@@ -144,7 +144,6 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
             },
           },
           [HighContrastSelector]: {
-            MsHighContrastAdjust: 'none',
             background: highContrastBackgroundColor[messageBarType],
             border: '1px solid WindowText',
             color: 'WindowText',
@@ -179,7 +178,6 @@ export const getStyles = (props: IMessageBarStyleProps): IMessageBarStyles => {
       color: semanticColors[iconColor[messageBarType]],
       selectors: {
         [HighContrastSelector]: {
-          MsHighContrastAdjust: 'none',
           color: 'WindowText',
         },
       },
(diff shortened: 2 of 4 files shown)


dggsax:
> I can't speak to CompoundButton, but in MessageBar we intentionally disable high contrast adjustment because the background colors have meaning and are set up to meet high contrast requirements. If you want individual children to auto-adjust with high contrast, you should set `MsHighContrastAdjust: 'auto'` specifically for those children.

@ecraig12345 Do you think that it would be worthwhile to start using https://github.com/microsoft/fluentui/wiki/Accessibility#high-contrast as a place to publish guidance for high contrast adoption in FluentUI? For example, your suggestion here about applying `MsHighContrastAdjust: 'auto'` to children of MessageBar fixed the issue that we were trying to resolve. We thought that there was a bug with `<Text>`, so we started re-writing some code to use more semantics like `<strong>` and `<p>` which put us in a weird place between just using `<Text>` everywhere and mixing in the native tags.

Another way we fixed the problem for us was that in our theme definitions, we just did something like the following:

```ts
// Apply high contrast conformance to all Text by default
const TextHighContrastStyles: ITextProps["styles"] = {
  root: {
    MsHighContrastAdjust: "auto",
  },
};

export const defaultTheme: Theme = {
  name: DEFAULT_THEME_NAME,
  ...createTheme({
    components: {
      // globally apply to all Text elements
      Text: {
        styles: TextHighContrastStyles,
      },
    },
    palette: {
      themePrimary: "#0078d4",
      themeLighterAlt: "#eff6fc",
      themeLighter: "#deecf9",
      themeLight: "#c7e0f4",
      themeTertiary: "#71afe5",
      themeSecondary: "#2b88d8",
      themeDarkAlt: "#106ebe",
      themeDark: "#005a9e",
      themeDarker: "#004578",
      neutralLighterAlt: "#faf9f8",
      neutralLighter: "#f3f2f1",
      neutralLight: "#edebe9",
      neutralQuaternaryAlt: "#e1dfdd",
      neutralQuaternary: "#d0d0d0",
      neutralTertiaryAlt: "#c8c6c4",
      neutralTertiary: "#a19f9d",
      neutralSecondary: "#605e5c",
      neutralPrimaryAlt: "#3b3a39",
      neutralPrimary: "#323130",
      neutralDark: "#201f1e",
      black: "#000000",
      white: "#ffffff",
      // Accessible colors
      blue: "#0078D4",
    },
  }),
};
```

Whichever solution is the preferred, or if there are multiple ways to solve the problem, it would be cool to be able to publish the tricks that people use for this kind of work so that others can reference it in the future.

ecraig12345:
I just switched teams. @smhigley might be able to answer this.

PR #14696, opened 2020-08-21
ThemeProvider: support applyTo prop

xugao:
#### Pull request checklist

#### Description of changes
- Support `applyTo: 'none' | 'element' | 'body'` prop
- Align default ('none') behavior with `Fabric` component.

#### Focus areas to test

(optional)


packages/react-theme-provider/src/ThemeProvider.styles.ts
```diff
@@ -0,0 +1,27 @@
+import { memoizeFunction } from '@uifabric/utilities';
+import { mergeStyles } from '@uifabric/styling';
+import { TokenSetType } from './types';
+
+export const getBodyStyles = memoizeFunction((bodyTheme: TokenSetType | undefined): string | undefined => {
```
dzearing, line 5: This could be `makeStyles`:

```tsx
export const useBodyStyles = makeStyles(theme => {
  ...
});
```

Would that work?

dzearing: useThemeProviderStyles

dzearing: I'm wondering if we actually need to spit out the entire set of theme tokens in this class. Basically we are now providing some tokens as inline styles and some as a class. Maybe we should simply try moving them all into makeStyles.

Also, the other thing I'm doing: in a separately PR I'll change makeStyles to use theme.renderRules to render rules in any way we want, which will allow us to flip between mergeStyles, jss, emotion, or whatever tool we need. this will help with convergence and make us less dependent on one css solution.

packages/react-theme-provider/src/useThemeProviderClasses.tsx
```diff
@@ -13,7 +20,15 @@ export const useThemeProviderClasses = makeClasses(theme => {
         fontFamily: 'var(--body-fontFamily)',
         fontWeight: 'var(--body-fontWeight)',
         fontSize: 'var(--body-fontSize)',
+        MozOsxFontSmoothing: 'var(--body-mozOsxFontSmoothing)',
+        WebkitFontSmoothing: 'var(--body-webkitFontSmoothing)',
+      },
+      {
+        '& input': inheritFont,
```
xugao, line 27: in [Fabric.styles](https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/Fabric/Fabric.styles.ts#L26), we also have `& button: inheritFont`. I don't think that's correct because our button defines the font styles and `button` is the root element. if setting `inherit`, the button styling will be incorrectly overwritten by `Fabric`. see [this pen](https://codepen.io/xugao/pen/OJNweZV)

packages/react-theme-provider/src/useThemeProviderClasses.tsx
```diff
@@ -1,9 +1,16 @@
...
 import { makeClasses } from './makeClasses';
 import { tokensToStyleObject } from './tokensToStyleObject';
 
+const inheritFont = { fontFamily: 'inherit' };
+
 export const useThemeProviderClasses = makeClasses(theme => {
   const { tokens } = theme;
 
+  const bodyStyles = {
+    background: 'var(--body-background)',
```
dzearing, line 11: color should be applied?

dzearing: actually shouldn't the font etc all be defined here?

dzearing: Then:

```js
{
  _applyTo_element: bodyStyles,
  _applyTo_body: {
    'global(body)': bodyStyles
  }
}
```

dzearing: (and maybe rename it to `themeStyles` or something more general.

dzearing: oh i don't think that would work

xugao: So I have styles which should be by default apply to `ThemeProvider` root defined in `root` slot. if `applyTo` is set to `element` or `body`, `bodyStyles` will be applied to `ThemeProvider`.

Applying styles to body is handled separately in `useApplyThemeToBody` hook, which handles applying the applicable classNames (`root-*` and `_applyTo_body-*`) to body element

PR #14703, opened 2020-08-21
Updating Rating to a function component inside React-Next

czearing:
#### Pull request checklist

#### Description of changes
1. Updating `Rating` to a function component.
2. Adding perf tests for `Rating` and `RatingNext`.
3. Adding a vr test for `RatingNext`.

#### Focus areas to test
`Rating`


packages/react-next/etc/react-next.api.md
```diff
@@ -1457,24 +1457,23 @@ export type IPositioningContainerTypes = IPositioningContainerProps;
 
 // @public (undocumented)
 export interface IRating {
+    rating: number;
 }
 
 // @public
-export interface IRatingProps extends React.AllHTMLAttributes<HTMLElement> {
+export interface IRatingProps extends React.HTMLAttributes<HTMLDivElement> {
```
dzearing, line 1464: This seems a little arbitrary; can we add back the "all" html attributes here unless there's a good reason?

ecraig12345: "All" is misleading, at least for this component: unknown attributes are applied to the root div (not an input or something), and there are a lot of extra things in AllHTMLAttributes that aren't valid for divs. It was also preventing deprecation warnings from showing up for `min` since *some* HTML element also happens to have a `min` prop.

dzearing: hmm ok.

dzearing (approved):

PR #14716, opened 2020-08-24
fix(BasePicker): suggestion list is opened event if the inputvalue is…

tdetugny:
#### Pull request checklist

#### Description of changes
currently:
When user enter value then remove it, the suggestion is closed
Expected:
When user enter value then remove it, the suggestion **should always be open**

#### Focus areas to test

(optional)

### DIFF
--- a/change/office-ui-fabric-react-2020-08-24-18-29-52-fix-BasePicker.json
+++ b/change/office-ui-fabric-react-2020-08-24-18-29-52-fix-BasePicker.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "fix(BasePicker): suggestion list is opened event if the inputvalue is empty",
+  "packageName": "office-ui-fabric-react",
+  "email": "tdetugny@clever-age.com",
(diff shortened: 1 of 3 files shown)


joschect:
@tdetugny Thanks for submitting this pr! Was there a bug/regression that was recently introduced? There were some changes made within the past weeks. I think that it closing when there are no options might be expected behavior but  I'd like to know more about your use case!

tdetugny:
@joschect I think it's an old issue. Here's an example: I enter the letter "A" and then I delete it. The suggestion list disappears.
https://www.loom.com/share/ea51ea249b49428aba98ec3ecfec3eaf
after my patch, it doesn't disappear anymore:
https://www.loom.com/share/74d643c46dc04691bddce6b2f04cd3fe

joschect:
@tdetugny I believe that this is actually intended behavior. The issue could be that we don't re-display the MRU if that kind of function is provided. You can retrigger it as a user by pressing enter or clicking on the input again.

khmakoto (changes requested):

PR #14718, opened 2020-08-24
Codemods: Adding docs for how to write / test codemods

tmaster628:
Added more documentation to the CodeMods directory on how to write and test mods under the documentation directory. Yay!
- Also added special docs for how to write code transforms -- this part is usually pretty tricky.
- Cleared out the json test files so that devs who follow the docs can write and test their own codemods!
- In the process of writing, found some places where I could clean some code up.
    - Realized that I needed to change my transforms to do nothing when neither map nor replacement value was provided.
    - Added the renameImport type into the json map's domain.


packages/codemods/documentation/howTo.md
```diff
@@ -0,0 +1,119 @@
...
+
+- To create a codemod using the config file, all you'd need to do is go into the `upgrades.json` file and replace what's there with the following template as defined in `src/codemods/types.ts`:
+
+```jsonld=
+{
+  "name": // Classify this collection of codemods
+  "upgrades": [
+    {
+      "name": // What upgrade are you running?
+      "type": // Currently renameProp | repathImport
```
joschect, line 35: I might add something that type corresponds to the function that will be run and what options are available.

packages/codemods/documentation/howTo.md
```diff
@@ -0,0 +1,119 @@
...
+    }
+  ]
+}
+```
+
+- Once you fill out this file, you can then go to [src/codemods/tests/configMod](../src/codemods/tests/configMod/configMod.test.ts) and test your codemod! Invoke the function `createCodeModFromJson()`, which will turn whatever exists in `upgrades.json` into a list of codemods. Check out ##Testing Your Codemods## for more info on how to verify it works!
+
+### Custom CodeMods
+
+- For developers who want a little more flexibility in their mod writing, you can create a codemod manually using codemod _utilities_. Details on these utilities can be found in `src/codemods/utilities`. To rename a prop, we'll need to use two utility functions:
```
joschect, line 55: There's 1 other option, you can make a codemod by creating a const with type Codemod and then exporting that const object as default from the file. It will then get picked up and run automagically.

packages/codemods/documentation/howTo.md
```diff
@@ -0,0 +1,119 @@
...
+  - `findJsxTag(file: SourceFile, tag: string)`
+    - This utility returns an array of JSX elements containing the given tag (in our case, the 'tag' will be the name of the component we're looking for).
+    - To create the codemod, you won't need to know much about AST's and how we parse them, but if you want to read up on it, check out [ts-morph!](https://ts-morph.com/)
+  - `renameProp( instances: (JsxOpeningElement | JsxSelfClosingElement)[], toRename: string, replacementName: string, replacementValue?: string, transform?: PropTransform)` \* This function accepts the return value of `findJSXTag`, as well as names for the deprecated prop and its replacement name. The function also takes in an optional replacement value, if you plan on updating the value of this prop whenever you want to rename it. For developers wanting even more granularity with value changes, there's an optional transform function you can pass in -- read[renamePropTransforms](./renamePropTransforms) for more.
+- To create the codemod, wrap these two function calls in a void function that takes in a single parameter of type `SourceFile`. This example suffices:
```
joschect, line 60: We might want to add a section about good practices, like returning a meaningful codemodresult so it can be processed and logged nicely.

joschect: So you're specific codemod function should return an `Ok({logs: ["useful description of changes that took place"]})` or an `Err({reason: "No operation taken because <Error> or nothing needs to get changed"})`

packages/codemods/documentation/howTo.md
```diff
@@ -0,0 +1,153 @@
...
+- Before you run your codemod, you'll need to specify files to run on! The following code will do:
+
+```typescript=
+project = new Project();
+project.addSourceFileAtPaths(`${process.cwd()}addSomePathHere!`);
+/* If you want to add many paths, you can do that too. */
+project.addSourceFilesAtPaths(`${process.cwd()}someRegexPath`);
+```
+
+- Once you have your codemod(s), to run them, you can use the function `runMods()`, which accepts an array of codemods, an array of source files, and a callback function that reports on the success of the mod. Here's what an invocation might look like:
```
joschect, line 117: This all happens automatically, you shouldn't need to worry about how runMods works.

tmaster628: That makes sense!

PR #14719, opened 2020-08-24
ComboBox: Fix React.StrictMode issues

### LINKED ISSUE #14288: ComboBox: strict mode compliance (UNSAFE_)
ComboBox updates the following state in `componentWillReceiveProps`:
- `selectedIndices` is reset if `props.selectedKeys` changes. This can also be set elsewhere in the class (incorrect semi-controlled pattern).
- `currentOptions` is reset if `props.options` changes. This can also be set elsewhere in the class if the user types new options.
- `suggestedDisplayValue` is cleared if `props.selectedKey` is newly null, but can be reset elsewhere in the class

It's likely that this would be easier to fix in-place rather than doing a function component conversion due to the overall complexity of the component.

MLoughry:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14288 

#### Description of changes

Partially migrate `ComboBox` to a functional component in order to mitigate React.StrictMode issues.

#### Focus areas to test

(optional)


packages/react-next/src/components/ComboBox/ComboBox.tsx
```diff
@@ -104,34 +97,114 @@ interface IComboBoxOptionWrapperProps extends IComboBoxOption {
...
-@customizable('ComboBox', ['theme', 'styles'], true)
-export class ComboBox extends React.Component<IComboBoxProps, IComboBoxState> {
-  public static defaultProps: IComboBoxProps = {
-    options: [],
-    allowFreeform: false,
-    autoComplete: 'on',
-    buttonIconProps: { iconName: 'ChevronDown' },
-  };
+  React.useEffect(() => {
+    if (defaultSelectedKey === undefined) {
```
ecraig12345, line 133: What's the intent of this check? If it's to check for controlled or uncontrolled, it would be better to check `selectedKeys !== undefined` since the controlled prop should always take precedence (even if they incorrectly provide both).

packages/react-next/src/components/ComboBox/ComboBox.tsx
```diff
@@ -104,34 +97,114 @@ interface IComboBoxOptionWrapperProps extends IComboBoxOption {
...
+      setSelectedIndices(indices);
+    }
+    setCurrentOptions(options);
+  }, [defaultSelectedKey, options, selectedKey]);
+
+  React.useEffect(() => {
+    if (selectedKey === null) {
+      setSuggestedDisplayValue(undefined);
+    }
+  }, [selectedKey]);
```
ecraig12345, line 146: To replicate the previous conditions where this was called, I think you'd need to add `text` and `options` to the dependencies. Though I'm not sure if it will matter.

packages/react-next/src/components/ComboBox/ComboBox.tsx
```diff
@@ -1012,23 +1064,18 @@ export class ComboBox extends React.Component<IComboBoxProps, IComboBoxState> {
...
-            if (onChange) {
-              onChange(submitPendingValueEvent, option, index, undefined);
-            }
-          },
-        );
+        this.props.hoisted.setSelectedIndices(selectedIndices);
+        this.props.hoisted.setCurrentOptions(changedOptions);
+
+        // If ComboBox value is changed, revert preview first
+        if (this._hasPendingValue && onPendingValueChanged) {
```
ecraig12345, line 1071: Similar to Dropdown this is potentially a behavior change since you're now calling onPendingValueChanged and onChange immediately after the setters, instead of in the post-`setState` callback.

joschect (approved):

Issue #14660, opened 2020-08-21
Component package scaffolding for v8 release


JustSlone:
(copied from v8 scope doc--modify as appropriate)

Just package scaffolding and ship FC conversations, not new versions of these components. 
- [x] scaffold `react-checkbox` 
- [x] default export V8 Checkbox and move latest code to `/next`
- [x] scaffold `react-link` 
- [x] default export V8 Link and move latest code to `/next`
- [x] scaffold `react-toggle` 
- [x] default export V8 Toggle and move latest code to `/next`
- [x] scaffold `react-slider` 
- [x] default export V8 Slider and move latest code to `/next`
- [x] scaffold `react-tabs` 
- [x] default export V8 Pivot and move latest code to `/next`

PR #14725, opened 2020-08-24
react-tabs: use merge-styles instead of scss and move scss code to next folder

xugao:
#### Pull request checklist

#### Description of changes
In preping for V8 release, we are reverting `Pivot` back to use `merge-styles`. 
`Pivot.styles.ts` is mostly the same as the one in OUFR, but includes some new styles added for overflow pivots
RTL is fixed in the change.

Related issue: #14660

#### Focus areas to test

(optional)


xugao:
can we remove themeprovider wrapper from storybook

packages/react-tabs/src/next/Pivot.tsx
```diff
@@ -0,0 +1,78 @@
...
+  linkIsSelected: 'is-selected',
+  linkContent: 'ms-Pivot-linkContent',
+  overflowMenuButton: 'ms-Pivot-overflowMenuButton',
+  root: 'ms-Pivot',
+  rootIsLarge: 'ms-Pivot--large',
+  rootIsTabs: 'ms-Pivot--tabs',
+  text: 'ms-Pivot-text',
+};
+
+const getStaticStylesMemoized = memoizeFunction(
```
dzearing, line 28: this can be automated and reduced to makeStyles.

PR #14733, opened 2020-08-25
feat(compose): replace with reusable styling hooks

mnajdova:
This PR showcase how we may replace compose's styling with re-usable styling hooks that can be extended. Currently I am using wrapper components that uses the base component, in this case `<Button />`, but once we have the hooks implementation, we can easily replace this with the usage of the `useButton` hook, which would allow us to not have wrappers. The important thing is that even the `TertieryCompactButton` component uses the same base component, which means that we at least are not exploding the render tree for each new composed components.

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleComposition.tsx
+++ b/packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleComposition.tsx
@@ -1,0 +1,79 @@
+import * as React from 'react';
+import { Button, useCSS, ButtonProps } from '@fluentui/react-northstar';
+
+type TertieryButtonProps = ButtonProps & {
+  tertiery?: boolean;
+};
+
+const useTertieryButtonStyles = (props, ...rest) => {
+  const { className, tertiery } = props;
+
+  return useCSS(
+    theme => {
+      const { siteVariables } = theme;
+
+      return tertiery
+        ? {
+            color: 'black',
+            backgroundColor: siteVariables.colorScheme.default.background2,
+          }
+        : {};
+    },
+    ...rest,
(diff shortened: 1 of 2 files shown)


packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleComposition.tsx
```diff
@@ -0,0 +1,79 @@
...
+    className,
+  );
+};
+
+const TertieryButton: React.FC<TertieryButtonProps> = props => {
+  const { className, tertiery, ...rest } = props;
+
+  const resolvedClasses = useTertieryButtonStyles(props);
+
+  return <Button className={resolvedClasses} {...rest} />;
```
mnajdova, line 32: This may be replaced with `useButton`, so we won't have wrapper components.

packages/fluentui/docs/src/examples/components/Button/Visual/ButtonExampleComposition.tsx
```diff
@@ -0,0 +1,79 @@
+import * as React from 'react';
+import { Button, useCSS, ButtonProps } from '@fluentui/react-northstar';
+
+type TertieryButtonProps = ButtonProps & {
+  tertiery?: boolean;
+};
+
+const useTertieryButtonStyles = (props, ...rest) => {
```
levithomason, line 8: To confirm, is the signature of the `use*Styles()` functions then
```tsx
(props, ...styles) => className
```
Where styles is an array of style objects, style functions taking in theme, or class name strings previously created with `useCSS`?

layershifter:
Closing for housekeeping 🏡

PR #14737, opened 2020-08-25
RFC - popper position fix

pompomon:
#### Pull request checklist

#### Description of changes

Fixing the issue with Popper incorrectly positioning content with own focus management or content with `autoFocus` prop.

The first issue we are addressing happens when content has some custom focus handling. In that case first Popper is added to the DOM and initial modifier effects are run and Popper is added to DOM positioning itself absolutely in top left corner, then the focus is called and window scroll jumps, and after that Popper's `forceUpdate` is executed to compute the rectangles of the Popper content and then modifiers are run and Popper instance is positioned correctly, but window scroll is not updated.
**The fix is two-fold:**
`modifier` part is overriding Popper styles to `position: fixed`, so when Popper instance is added to DOM, the popup is added to the top-left corner in the viewport, instead of top-left corner of the page, so no window scroll happens when focus is called.

`patch` part is removing `position: fixed` style override, before the first `forceUpdate` is run, thus keeping the computation of the offsets correct.

The issue with `autoFocus` is more complex, since `autoFocus` is handled by React internally and is triggered before any modifiers or their effects are run, so we've added a check for the Popper children and if any of them have `autoFocus` prop defined, we would output a warning to the console.

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Popup/Types/PopupExample.tsx
+++ b/packages/fluentui/docs/src/examples/components/Popup/Types/PopupExample.tsx
@@ -70,7 +70,6 @@ const LoremParagraph = () => (
 const PopupExample = () => (
   <Popup
     trigger={<Button content="Show popup" />}
-    positionFixed
     content={{
       // eslint-disable-next-line jsx-a11y/no-autofocus
       content: <input autoFocus value="This popup is rendered next to the trigger." />,
--- a/packages/fluentui/docs/src/examples/components/Popup/Types/PopupExample.tsx
+++ b/packages/fluentui/docs/src/examples/components/Popup/Types/PopupExample.tsx
@@ -1,11 +1,92 @@
 import * as React from 'react';
 import { Button, Popup } from '@fluentui/react-northstar';
-import { MoreIcon } from '@fluentui/react-icons-northstar';
+// import { MoreIcon } from '@fluentui/react-icons-northstar';
 
-const PopupExample = () => (
-  <Popup content="Hello from popup!">
-    <Button icon={<MoreIcon />} title="Show popup" />
-  </Popup>
+const LoremParagraph = () => (
+  <p>
+    This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our
+    docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent
+    UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is
+    powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub.
+    This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our
+    docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent
+    UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is
+    powered by Fluent UI, check our docs and GitHub. v This example is powered by Fluent UI, check our docs and GitHub.
+    This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our
+    docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent
+    UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is
+    powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub.
+    This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our
+    docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent
+    UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is
+    powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub.
+    This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our
+    docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent
+    UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is
+    powered by Fluent UI, check our docs and GitHub. v This example is powered by Fluent UI, check our docs and GitHub.
+    This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our
+    docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent
+    UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is
+    powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub.
+    This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our
+    docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent
+    UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is
+    powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub.
+    This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our
+    docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent
+    UI, check our docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is
+    powered by Fluent UI, check our docs and GitHub. v This example is powered by Fluent UI, check our docs and GitHub.
+    This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent UI, check our
+    docs and GitHub. This example is powered by Fluent UI, check our docs and GitHub. This example is powered by Fluent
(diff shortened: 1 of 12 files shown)


packages/fluentui/react-northstar/src/themes/teams/components/Popup/popupContentStyles.ts
```diff
@@ -8,6 +8,7 @@ export const popupContentStyles: ComponentSlotStylesPrepared<PopupContentStylesP
   root: ({ props: p, variables: v }): ICSSInJSStyle => ({
     display: 'block',
     zIndex: v.zIndex,
+    position: 'fixed',
```
pompomon, line 11: @layershifter This is needed only for the autoFocus input example, Datepicker works fine with added modifiers

packages/fluentui/docs/src/examples/components/Popup/Types/PopupExample.tsx
```diff
@@ -1,11 +1,95 @@
...
+    <PopupExample />
+    <DatepickerExample />
+    <LoremParagraph />
+    <LoremParagraph />
+    <LoremParagraph />
+  </>
 );
 
-export default PopupExample;
+export default scrollPopupExample;
```
layershifter, line 95: This definitely should become a separate E2E test

packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx
```diff
@@ -10,6 +10,29 @@ import { getScrollParent } from './getScrollParent';
...
+    (node.nodeName === 'BUTTON' ||
+      node.nodeName === 'INPUT' ||
+      node.nodeName === 'SELECT' ||
+      node.nodeName === 'TEXTAREA') &&
+    node[reactInstanceKey].pendingProps.autoFocus
+  );
+};
+
+const getReactInstanceKey = (elm: HTMLElement): string => {
+  if (!elm) return null;
```
layershifter, line 25:
```suggestion
```

`elm` should not be even `null`

packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx
```diff
@@ -10,6 +10,29 @@ import { getScrollParent } from './getScrollParent';
...
+
+  for (const k in elm) {
+    if (k.startsWith('__reactInternalInstance$')) {
```
layershifter, line 28: `.startsWith()` is not in IE11 yet. We can replace it with `indexOf()` or the Lodash one.

`ReactInstanceKey` is also constant for the whole app lifecycle, we can cache into some variable:

```tsx
let reactInstanceKey: string

const getReactInstanceKey = (elm: HTMLElement): string => {
  if (!reactInstanceKey) {
     reactInstanceKey = ...
  }

  return reactInstanceKey
}
```

packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx
```diff
@@ -10,6 +10,29 @@ import { getScrollParent } from './getScrollParent';
 import { getPlacement, applyRtlToOffset } from './positioningHelper';
 import { PopperModifiers, PopperProps } from './types';
 
+// https://github.com/facebook/react/blob/848bb2426e44606e0a55dfe44c7b3ece33772485/packages/react-dom/src/client/ReactDOMHostConfig.js#L157-L166
+const isAutofocusEnabled = (node: Node, reactInstanceKey: string): boolean => {
```
layershifter, line 14:
```suggestion
const hasAutofocusProp = (node: Node): boolean => {
```

layershifter: BTW, it can be also `undefined`

```suggestion
const hasAutofocusProp = (node: Node): boolean | undefined => {
```

layershifter: If we will cache `getReactInstanceKey()`, we can call it inside `hasAutofocusProp` and we will not need an additional param 👍

packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx
```diff
@@ -226,6 +280,26 @@ export const Popper: React.FunctionComponent<PopperProps> = props => {
     return destroyInstance;
   }, [createInstance]);
 
+  if (process.env.NODE_ENV !== 'production') {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    React.useEffect(() => {
+      if (contentRef.current) {
+        const contentNode = contentRef.current;
+        const reactInstanceKey = getReactInstanceKey(contentNode);
```
layershifter, line 288:
```suggestion
```

packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx
```diff
@@ -226,6 +280,26 @@ export const Popper: React.FunctionComponent<PopperProps> = props => {
...
+        const treeWalker = contentNode.ownerDocument?.createTreeWalker(contentNode, NodeFilter.SHOW_ELEMENT, {
```
layershifter, line 289: Can we also apply this change to `useTriggerElement`?

packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx
```diff
@@ -226,6 +280,26 @@ export const Popper: React.FunctionComponent<PopperProps> = props => {
...
+          acceptNode: node =>
+            isAutofocusEnabled(node, reactInstanceKey) ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_SKIP,
```
layershifter, line 291: Once `hasAutofocusProp` will accept a single param we can move this condition inside of it

```suggestion
          acceptNode: hasAutofocusProp,
```

packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx
```diff
@@ -226,6 +280,26 @@ export const Popper: React.FunctionComponent<PopperProps> = props => {
...
+        });
+
+        while (treeWalker.nextNode()) {
+          const node = treeWalker.currentNode;
+          // eslint-disable-next-line no-console
+          console.warn(node, ['Node with autoFocus', 'in Popper would cause window', 'scroll to jump'].join(' '));
```
layershifter, line 297:
```
          console.warn('<Popper>:', node);
          console.warn(
            '<Popper>',
            [
              '^ this node contains the "autoFocus" prop on a React element. This can break the initial positioning',
              'of an element and cause a window jump effect. This issue occurs because React polyfills "autoFocus"',
              'behavior to solve inconsistencies between different browsers:',
              'https://github.com/facebook/react/issues/11851#issuecomment-351787078',
              '\n',
              'However, ".focus()" in this case occurs before any other React effects will be executed',
              '(React.useEffect(), componentDidMount(), etc.) and we can not prevent this behavior. If you really',
              'want to use "autoFocus" please add "position: fixed" to styles of the element that is wrapped by',
              '"Popper".',
              `In general, it's not recommended to use "autoFocus" as it may break accessibility aspects:`,
              'https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-autofocus.md',
              '\n',
              'We suggest to use the "trapFocus" prop on Fluent components or a catch "ref" and then use',
              '"ref.current.focus" in React.useEffect():',
              'https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element',
            ].join(' '),
          );
```

packages/fluentui/docs/src/examples/components/Popup/Visual/PopupScrollExample.tsx
```diff
@@ -0,0 +1,123 @@
...
+      trigger={<Button content="Show popup" title="PopupWithTrapFocus" />}
+      content={<Button content="Click me!" />}
+      trapFocus={{
+        firstFocusableSelector: `.${buttonClassName}`,
+      }}
+    />
+  );
+};
+
+const ControlledPopupWithFocus = () => {
```
layershifter, line 81: Let's add the same case for `inline` popup

packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx
```diff
@@ -8,7 +8,105 @@ import { isBrowser } from '../isBrowser';
...
+};
+
+function hasAutofocusFilter(node: Node) {
+  return hasAutofocusProp(node) ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_SKIP;
+}
+
+// We are setting the position to `fixed` in the first calculation
+// To prevent scroll jumps in case of the content with managed focus
+// setInitPositionToFix sets the position to `fixed` before applyStyles modifier
+// unsetInitPositionToFix restores the original position after applyStyles modifier
```
layershifter, line 46: Let's update this comment to be aligned with an actual implementation

packages/fluentui/react-northstar/src/utils/positioner/Popper.tsx
```diff
@@ -182,11 +281,21 @@ export const Popper: React.FunctionComponent<PopperProps> = props => {
           phase: 'afterWrite' as PopperJs.ModifierPhases,
           fn: handleUpdate,
         },
+        popperInitialPositionFix.modifier,
       ].filter(Boolean),
       onFirstUpdate: state => handleUpdate({ state }),
     };
 
     popperInstanceRef.current = PopperJs.createPopper(reference, contentRef.current, options);
+    const originalForceUpdate = popperInstanceRef.current.forceUpdate;
```
layershifter, line 290:
```suggestion
    positionPatch(popperInstanceRef.current);
```

My original proposal was to move the whole logic there to our hook

layershifter (approved):
👍 

Please update PR description to reflect the work that was done

PR #14747, opened 2020-08-25
Moving MessageBar to react-next, in preparation for converting to a function component

czearing:
#### Pull request checklist

#### Description of changes
This change does not modify `MessageBar` ; it simply moves it to react-next for better diffing in the actual conversion PR.


dzearing:
@JasonGore I have no idea how the perf numbers could have changed so much here. I think this is a red herring?

PR #14750, opened 2020-08-25
Slider: move from react-next to react-slider

xugao:
#### Pull request checklist

#### Description of changes
Prepping `Slider` for V8 release, with following steps:
- [x] scaffold new `react-slider` and move `Slider` from `react-next` to it (current PR)
- [ ] Move current `Slider` code to `next` folder and revert default `Slider` to only have FC conversion (no scss)

Related issue: #14660

#### Focus areas to test

(optional)


packages/react-next/etc/react-next.api.md
```diff
@@ -2555,6 +2486,7 @@ export * from "@fluentui/react-button/lib/Button";
 export * from "@fluentui/react-button/lib/MenuButton";
 export * from "@fluentui/react-button/lib/SplitButton";
 export * from "@fluentui/react-button/lib/ToggleButton";
+export * from "@fluentui/react-slider/lib/Slider";
```
dzearing, line 2489: I'm wondering why for subpackages we don't export * from the package? Is there any reason why we would be using the top level physical files?

I think it could make downstream testing more difficult when testing with commonjs. In commonjs, this import won't work without some form of pre-bundling or redirect.

dzearing: I know for AMD partners they might need to do extra work to resolve the packages, possibly. But that's the nature of AMD with the lacking in node-compatible resolution. I think out of the box we should prioritize esm and node (commonjs) or environments.

packages/react-slider/src/components/Slider/Slider.base.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
-import { FocusRects } from '../../Utilities';
+import { FocusRects } from '@uifabric/utilities';
```
khmakoto, line 2: You could import from `office-ui-fabric-react/lib/Utilities` across the package to remove the dependency from `@uifabric/utilities` in the `package.json` file.

packages/react-slider/src/components/Slider/Slider.doc.tsx
```diff
@@ -2,16 +2,15 @@ import * as React from 'react';
...
-const SliderVerticalExampleCode = require('!raw-loader!office-ui-fabric-react/src/components/Slider/examples/Slider.Vertical.Example.tsx') as string;
+const SliderBasicExampleCode = require('!raw-loader!@fluentui/react-slider/src/components/Slider/examples/Slider.Basic.Example.tsx') as string;
+const SliderVerticalExampleCode = require('!raw-loader!@fluentui/react-slider/src/components/Slider/examples/Slider.Vertical.Example.tsx') as string;
 
 export const SliderPageProps: IDocPageProps = {
   title: 'Slider',
   componentName: 'Slider',
-  componentUrl:
-    'https://github.com/microsoft/fluentui/tree/master/packages/office-ui-fabric-react/src/components/undefined',
+  componentUrl: 'https://github.com/microsoft/fluentui/tree/master/packages/react-slider/src/components/undefined',
```
khmakoto, line 13:
```suggestion
  componentUrl: 'https://github.com/microsoft/fluentui/tree/master/packages/react-slider/src/components/Slider',
```

packages/react-slider/src/components/Slider/Slider.types.ts
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+
 import * as React from 'react';
-import { IStyle, ITheme } from '../../Styling';
-import { IStyleFunctionOrObject, IRefObject } from '../../Utilities';
+import { IStyle, ITheme } from '@uifabric/styling';
```
khmakoto, line 4: Same comment here for `@uifabric/styling`.

packages/react-slider/src/components/Slider/examples/Slider.Basic.Example.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
-import { Slider } from '@fluentui/react-next/lib/Slider';
-import { IStackTokens, Stack, IStackStyles } from '@fluentui/react-next/lib/Stack';
+import { Slider } from '@fluentui/react-slider';
+import { IStackTokens, Stack, IStackStyles } from 'office-ui-fabric-react';
```
khmakoto, line 3:
```suggestion
import { IStackTokens, Stack, IStackStyles } from 'office-ui-fabric-react/lib/Stack';
```

packages/react-slider/src/components/Slider/Slider.base.tsx
ecraig12345: I think importing from `@uifabric/utilities` is correct since this package actually should NOT depend on OUFR, and the OUFR dep will be removed once we figure out what to do with examples.

packages/react-slider/src/components/Slider/Slider.types.ts
ecraig12345: No this is right too

packages/react-slider/src/components/Slider/Slider.base.tsx
xugao: yeah we want to avoid referencing OUFR as much as possible because that's temporary solution, see [this](https://hackmd.io/@fluentui/Hyu6U-XXP)

packages/react-next/etc/react-next.api.md
xugao: this is not an issue since we have the correct build/output in `lib-commonjs` today

ecraig12345: The only reason we have the correct lib-commonjs output is because of this postprocess script https://github.com/microsoft/fluentui/blob/master/scripts/tasks/postprocess-commonjs.js

ecraig12345: And we have a similar one for AMD build postprocessing

PR #14753, opened 2020-08-26
Fixed incorrect margin of Attachment icon after code editing (UI builder)

vyhnalekl:
#### Pull request checklist

#### Description of changes

When the props of `Attachment ` are changed in the code editor of UI builder. The Icon of attachment is moved due to added margin. This PR fix this issue.


vyhnalekl:
This fix is causing problems with other use case of the code editor - when you want to create an element in code editor from scratch and don't provide the data-builder-id to it. Due to this fix it would be no longer added automatically.

PR #14754, opened 2020-08-26
Removes url parameters after switch to store - UI builder

vyhnalekl:
#### Pull request checklist

#### Description of changes

Removes url parameters when user switches to store view.


miroslavstastny (approved):
please add changelog entry before merging

PR #14755, opened 2020-08-26
feat(Menu): dismiss menu on scroll

assuncaocharles:
#### Pull request checklist

## BREAKING CHANGE

This PR will make that opened menus close at page scroll.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Menu/MenuItem.tsx
+++ b/packages/fluentui/react-northstar/src/components/Menu/MenuItem.tsx
@@ -1,4 +1,5 @@
 import { Accessibility, submenuBehavior, menuItemBehavior, MenuItemBehaviorProps } from '@fluentui/accessibility';
+import { useEventListener, EventListener } from '@fluentui/react-component-event-listener';
 import {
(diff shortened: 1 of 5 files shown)


packages/fluentui/react-northstar/src/components/Menu/MenuItem.tsx
```diff
@@ -303,6 +304,15 @@ export const MenuItem = compose<'a', MenuItemProps, MenuItemStylesProps, {}, {}>
       }
     };
 
+    useEventListener({
+      target: context.target,
+      type: 'scroll',
+      listener: e => {
+        if (!isSubmenuOpen()) return;
+        trySetMenuOpen(false, (e as unknown) as React.MouseEvent);
```
layershifter, line 312:
```suggestion
        trySetMenuOpen(false, e);
```

This is not `React.MouseEvent`, this is native event. Let's fix `trySetMenuOpen` to accept `MouseEvent` instead

layershifter (approved):

PR #14756, opened 2020-08-26
fix(Toolbar): make Popup visible on disabled elements

layershifter:
This fixes an issue in `Toolbar` when `Tooltip`s were not visible on `ToolbarItem`s even if they were wrapped with an additional element. Adds few more validations to catch possible cases when an invalid element can be passed as a `trigger` to `Popup`/`Tooltip`.


packages/fluentui/react-bindings/src/hooks/useTrigger.tsx
```diff
@@ -0,0 +1,86 @@
...
+    if (!React.isValidElement(element)) {
+      throw new Error('useTrigger(): An invalid value was passed, please pass a valid React element as a trigger');
+    }
+
+    if (ReactIs.isFragment(element)) {
+      throw new Error(
+        'useTrigger(): A "React.Fragment" cannot be used as a "trigger" as it will be impossible to spread props on it',
+      );
+    }
+  }
```
dzearing, line 41: Could consider wrapping this if block in a dev mode check, we probably don't need to throw in production builds.

packages/fluentui/react-bindings/src/hooks/useTrigger.tsx
```diff
@@ -0,0 +1,86 @@
...
+                'useTrigger(): A disabled element should have explicit "pointer-events: "none" in its styles due a bug',
+                'in Chrome that breaks "onMouseLeave" event in React: https://github.com/facebook/react/issues/19692',
+              ].join(' '),
+            );
+          }
+        }
+      }
+    }, []);
+
+    return element ? <Ref innerRef={ref}>{element as React.ReactElement}</Ref> : null;
```
dzearing, line 82: It is weird to me that this behavior is in the `process.env` check. I don't really understand why you'd only use `<Ref/>` in a dev build.

packages/fluentui/react-bindings/src/hooks/useTrigger.tsx
```diff
@@ -0,0 +1,86 @@
...
+      node.nodeName === 'TEXTAREA') &&
+    (node as HTMLButtonElement).disabled
+  );
+}
+
+function isInteractiveFilter(node: Node) {
+  return isDisabledInteractive(node) ? NodeFilter.FILTER_ACCEPT : NodeFilter.FILTER_SKIP;
+}
+
+export function useTrigger(props: UseTriggerOptions): React.ReactElement | null {
```
dzearing, line 27: Comments would be appreciated about what the use case is for this.

packages/fluentui/react-bindings/src/hooks/useTrigger.tsx
layershifter: `Ref` is required only for `dev` to capture a `ref` that will be used for `dev` only checks. This `ref` is not used anywhere outside

packages/fluentui/react-bindings/src/hooks/useTrigger.tsx
layershifter:
```suggestion
/**
 * Performs consistent checks for components that are using `trigger` pattern (`Tooltip`, `Popup`). Ensures that
 * `children` or `trigger` props will always pass a valid React element to what additional props and handlers can 
 * be applied.  
 */
```

Does it makes sense?

packages/fluentui/react-bindings/src/hooks/useTrigger.tsx
layershifter: Moved 👍

packages/fluentui/react-bindings/src/hooks/useTrigger.tsx
dzearing: Should this be `useTriggerElement` then? Pass in children/trigger element, get back the element which should be used as the trigger element.

packages/fluentui/react-bindings/src/hooks/useTrigger.tsx
dzearing: ah i see.

packages/fluentui/react-bindings/src/hooks/useTrigger.tsx
dzearing: Minor modification, adjust to indicate what the ins and outs are, like:

> Given the `children` and expected element to `trigger` on, returns the correct trigger element. Performance consistent checks for components that are using the `trigger` pattern (etc)

That way the user knows generally how to use it. (They can look at the examples you mentioned for more details 👍 )

Otherwise looks good

miroslavstastny (approved):

PR #14757, opened 2020-08-26
fix(UI Builder): Add basic accessibility and narration stub

jurokapsiar:
### Description of changes

Adds basic accessibility to UI Builder
Adds stub for displaying narration (disabled by default)

#### Focus areas to test

(optional)


package.json
```diff
@@ -48,6 +48,7 @@
     "scrub": "node ./scripts/scrub.js",
     "start:legacy": "yarn workspace @uifabric/fabric-website-resources start",
     "start": "node scripts/start.js",
+    "start:docs": "yarn workspace @fluentui/docs start",
```
jurokapsiar, line 51: since prompt menu is not accessible, this will allow blind users to start the experience

packages/fluentui/react-builder/src/components/ComponentTree.tsx
```diff
@@ -11,15 +11,41 @@ export type ComponentTreeProps = {
...
+  if (e.shiftKey && keyCode === 121) {
+    const activeElement = document.activeElement;
+    if (activeElement) {
+      const event = new MouseEvent('contextmenu', { bubbles: true });
+      activeElement.dispatchEvent(event);
+    }
+  }
+};
+
+const treeKeyDown = isMac ? macKeyDown : undefined;
```
vyhnalekl, line 31: If I understand it correctly this undefined will be changed latter, so that it works on other OS as well? Should I add it to the list of task to do on UI builder?

jurokapsiar: no - in other systems this is a default shortcut that is handled by the system. Mac does not have such shortcut unless you start VoiceOver.

vyhnalekl: I see, thanks!

package.json
miroslavstastny: don't you want to use `northstar` in the name?

packages/fluentui/react-builder/src/components/ComponentTree.tsx
```diff
@@ -11,15 +11,41 @@ export type ComponentTreeProps = {
...
   onDeleteComponent?: () => void;
+  onAddComponent?: (uuid: string, where: string) => void;
 };
 
+const isMac = navigator.userAgent.indexOf('Mac OS X') !== -1;
+const behavior = isMac ? treeAsListBehavior : treeBehavior;
+
+const macKeyDown = (e: React.KeyboardEvent<HTMLElement>) => {
+  const keyCode = e.keyCode || e.which;
+  if (e.shiftKey && keyCode === 121) {
```
miroslavstastny, line 22: add constant or comment for `121`

packages/fluentui/react-builder/src/components/ComponentTree.tsx
```diff
@@ -160,6 +198,23 @@ export const ComponentTree: React.FunctionComponent<ComponentTreeProps> = ({
...
+    (uuid, where) => {
+      onAddComponent?.(uuid, where);
+    },
+    [onAddComponent],
+  );
+
+  const handleDeleteComponent = React.useCallback(
+    uuid => {
+      onSelectComponent(jsonTreeFindElement(tree, uuid));
+      setTimeout(() => {
```
miroslavstastny, line 211: I would rather rename `onDeleteComponent` to `onDeleteSelectedComponent` and add `onDeleteComponent(uuid)`.

miroslavstastny (approved):

package.json
jurokapsiar: maybe we should then rename @fluentui/docs to @fluentui/northstar?

jurokapsiar: but I agree - will rename the command here, let's rename remaining things in separate PR

PR #14762, opened 2020-08-26
withResponsiveMode: Remove finddomnode usage

### LINKED ISSUE #14320: withResponsiveMode: strict mode compliance (findDOMNode)
`withResponsiveMode` uses [`ReactDOM.findDOMNode`](https://reactjs.org/docs/react-dom.html#finddomnode) which is now discouraged for various reasons and [deprecated in strict mode](https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage).

While this is less likely to present a practical problem in this case, since by definition a decorator like `withResponsiveMode` must wrap a class component, it will probably still trigger warnings in strict mode and we should try to remove it.

### LINKED ISSUE #14761: Remove finddomnode from withResponsiveMode

joschect:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14320 

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -59,48 +59,42 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
...
-          ref={this._updateComposedComponentRef}
-          responsiveMode={responsiveMode}
-          {...(this.props as any)}
-        />
+      return (
+        <WindowContext.Consumer>
+          {({ window }) => (
+            <ComposedComponent
+              ref={this._updateComposedComponentRef}
+              responsiveMode={getResponsiveMode(window)}
```
ecraig12345, line 88: I like this in general, but I wonder if calculating the responsive mode on every render could somehow cause perf problems? Mainly a concern if `window.innerWidth` is one of the things that's expensive to access. (I don't remember the list of problematic properties offhand and I'm not sure what to call it in a search...)

KevinTCoughlin: `window.innerWidth` access may cause forced browser reflow https://gist.github.com/paulirish/5d52fb081b3570c81e3a. Not sure if that's what you are referring to above.

ecraig12345: Yes it is, thanks! So we should probably find a way to ensure the calculation only runs on resize.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -59,48 +59,42 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
 ): any {
   const resultClass = class WithResponsiveMode extends BaseDecorator<TProps, IWithResponsiveModeState> {
     private _events: EventGroup;
+    private _async: Async;
 
     constructor(props: TProps) {
       super(props);
       this._events = new EventGroup(this);
+      this._async = new Async(this);
```
KevinTCoughlin, line 67: I believe you need a corresponding `this._async.dispose()` call in `componentWillUnmount` to avoid leaking memory.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -59,48 +59,42 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
...
-      const element = findDOMNode(this) as Element;
-      const currentWindow = (element && getWindow(element)) || window;
-      const responsiveMode = getResponsiveMode(currentWindow);
-
-      if (responsiveMode !== this.state.responsiveMode) {
-        this.setState({
-          responsiveMode,
-        });
-      }
+      this.forceUpdate();
```
KevinTCoughlin, line 97: SharePoint Online's top-level wrapper component is `withResponsiveMode` so we need to be cautious of introducing unnecessary forced renders. We've had performance problems with the component previously (see https://github.com/microsoft/fluentui/pull/12258).

How does this change look in the profiler before & after?

cc: @pingj

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -59,48 +59,42 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
...
+    private _async: Async;
 
     constructor(props: TProps) {
       super(props);
       this._events = new EventGroup(this);
+      this._async = new Async(this);
       this._updateComposedComponentRef = this._updateComposedComponentRef.bind(this);
 
-      this.state = {
-        responsiveMode: getInitialResponsiveMode(),
```
KevinTCoughlin, line 69: Will this change regress the performance optimization made for SharePoint Online in https://github.com/microsoft/fluentui/pull/12258?

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
pingj: could we keep the existing logic that only re-render when the mode changes. As @KevinTCoughlin  said, it is on the top level wrapper, and many sub components don't have shouldComponentUpdate, so they will just re-render.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
pingj: what is the initial state now? we had an issue where the default initial mode was not correct, and it quickly triggers a re-render after detecting the new window size. Kevin added the getIntialResponsiveMode to fix this issue before.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
joschect: As it turns out you can access class contexts via `static context` I've moved to that and it has cleared all the problems.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -92,9 +95,7 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
     }
 
     private _onResize = () => {
-      const element = findDOMNode(this) as Element;
-      const currentWindow = (element && getWindow(element)) || window;
-      const responsiveMode = getResponsiveMode(currentWindow);
+      const responsiveMode = getResponsiveMode(this.context.window);
```
KevinTCoughlin, line 98: @joschect check with @ThomasMichon that this `window` is correct for embedded scenarios

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
```diff
@@ -58,6 +58,9 @@ export function withResponsiveMode<TProps extends { responsiveMode?: ResponsiveM
   ComposedComponent: new (props: TProps, ...args: any[]) => React.Component<TProps, TState>,
 ): any {
   const resultClass = class WithResponsiveMode extends BaseDecorator<TProps, IWithResponsiveModeState> {
+    public static contextType = WindowContext;
+    public context: React.ContextType<typeof WindowContext>;
```
KevinTCoughlin, line 62: Must these be `public`?

ecraig12345: Yes, React expects it to be public (like the component lifecycles) https://reactjs.org/docs/context.html#classcontexttype

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
ecraig12345: I'm not sure it would be correct by default--I think the new WindowProvider approach relies on the user to correctly set the window for alternative contexts. @dzearing correct me if I'm wrong but that's what I remember from discussions and what's mentioned at the end of [the package docs](https://www.npmjs.com/package/@fluentui/react-window-provider) (though we could stand to make that more explicit).

ecraig12345: If that's the case, we would need to move the responsive utilities to react-next (and update references) so that we can include specific guidance for consumers to add WindowProviders as part of the release notes.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
joschect: :( Yeah I found that out. I was saddened by it too.

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
joschect: It already wasn't going to be quite right. Because the onResize only triggered based off the global `window`, it would only update the resize when that window changed, rather than when the embedded window changed :/

packages/office-ui-fabric-react/src/utilities/decorators/withResponsiveMode.tsx
dzearing: Possible to use `Context.Consumer` as an alternative. Maybe there are reasons why you didn't though.

```jsx
  <WindowContext.Consumer>{window => (
            <.../>
   )}</UserContext.Consumer>
```

joschect: I had it that way initially, but there were concerns about it rerendering too often. The alternative was to make another child component but that would increase the dom tree size and complexity.

PR #14763, opened 2020-08-26
Updating Shimmer to a function component inside React-Next

czearing:
#### Pull request checklist

#### Description of changes
1. Updating Shimmer to a function component inside React-Next.
2. Removing unused componentRef from `Shimmer` react-next shimmer types as there are no public methods.
3.  Deprecating unused componentRef from `Shimmer` OUFR types.
4. Adding RELEASE_NOTES comments.

#### Focus areas to test
`Shimmer`


packages/react-next/src/components/Shimmer/Shimmer.base.tsx
```diff
@@ -1,126 +1,94 @@
...
+    isDataLoaded,
+    className,
+    transitionAnimationInterval: TRANSITION_ANIMATION_INTERVAL,
+    shimmerColor: shimmerColors && shimmerColors.shimmer,
+    shimmerWaveColor: shimmerColors && shimmerColors.shimmerWave,
+  });
+
+  const internalState = useConst<{ lastTimeoutId: number }>(() => ({
+    lastTimeoutId: 0,
+  }));
```
ecraig12345, line 42: You can remove the wrapper here.
```suggestion
  const internalState = useConst({
    lastTimeoutId: 0,
  });
```

packages/react-next/src/components/Shimmer/Shimmer.base.tsx
```diff
@@ -1,126 +1,94 @@
...
-            <DelayedRender>
-              <div className={this._classNames.screenReaderText}>{ariaLabel}</div>
-            </DelayedRender>
-          </div>
-        )}
-      </div>
-    );
-  }
-}
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ecraig12345, line 65: After thinking through this I'm pretty sure it's correct that you don't need to depend on `contentLoaded`, but can you please add a comment about why?

PR #14764, opened 2020-08-26
fix date picker controlled scenario

lorejoh12:
#### Pull request checklist

#### Description of changes

adding fix for scenario where DatePicker is fully controlled and the internal selected date updates, but the formatted date does not. Also adding example to verify in the future.

Comments seem to suggest that updating the passed value prop to DatePicker was expected to call setSelectedDate somewhere, but value isn't a dependency of any useEffects and that function never gets called. Might have been left over from an earlier commit.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/date-time/src/components/DatePicker/DatePicker.base.tsx
+++ b/packages/date-time/src/components/DatePicker/DatePicker.base.tsx
@@ -132,9 +132,7 @@ function useSelectedDate({ formatDate, value, onSelectDate }: IDatePickerProps) 
 
   React.useEffect(() => {
     setFormattedDate(value && formatDate ? formatDate(value) : '');
-    // setSelectedDate already updates the formatetd date if value changes
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [formatDate]);
+  }, [formatDate, value]);
 
   return [selectedDate, formattedDate, setSelectedDate, setFormattedDate] as const;
 }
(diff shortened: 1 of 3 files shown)


packages/date-time/src/components/DatePicker/DatePicker.base.tsx
```diff
@@ -487,7 +489,7 @@ export const DatePickerBase = React.forwardRef(
             {...calloutProps}
             className={css(classNames.callout, calloutProps && calloutProps.className)}
             // eslint-disable-next-line react/jsx-no-bind
-            onDismiss={calendarDismissed}
+            onDismiss={calloutDismissed}
```
lorejoh12, line 492: unrelated bug fix that was annoying me- onDismiss accepts a function where the first parameter is a mouseEvent, but we were passing a function where the first parameter is an option Date. so the calendarDismissed function would get called with a MouseEvent instead of a Date and we'd get an error in the console

packages/date-time/src/components/DatePicker/DatePicker.base.tsx
```diff
@@ -418,6 +416,10 @@ export const DatePickerBase = React.forwardRef(
       // don't need to focus the text box, if necessary the focusTrapZone will do it
     };
 
+    const calloutDismissed = (ev: React.MouseEvent<HTMLElement>): void => {
```
MLoughry, line 419: What's the purpose for this added indirection?

lorejoh12: looks like my earlier comment didn't get added in time, I'll copy/paste it here:

`unrelated bug fix that was annoying me- onDismiss accepts a function where the first parameter is a mouseEvent, but we were passing a function where the first parameter is an option Date. so the calendarDismissed function would get called with a MouseEvent instead of a Date and we'd get an error in the console`

I'm actually not sure why Typescript didn't catch this- perhaps because both parameters were optional, typescript says both being undefined means it's valid?

pompomon (approved):

PR #14765, opened 2020-08-26
Codemods: Add config to streamline mod running process

tmaster628:
Added a config option for devs running their codemods. Now, devs can fill out a simple config file, and let that file determine what mods to run instead of dealing with the command line!

Here's the file devs would need to add to their repo:
`modConfig.json`
  ```
  {
  "stringFilters": [],
  "regexFilters": [],
  "includeMods": true
  }
  ```
where stringFilters corresponds to the -n flag and regexFilters corresponds to the -r flag when normally running mods. includeMods acts as the inverse of the -e flag.

I also had to change a couple import paths that were crashing the modrunner.

### DIFF
--- a/packages/codemods/src/command.ts
+++ b/packages/codemods/src/command.ts
@@ -57,7 +57,7 @@ export class CommandParser {
     let configObj: ModRunnerConfigType = { stringFilters: [], regexFilters: [], includeMods: false };
     if (parsed.config) {
       /* Attempt to locate the modConfig file in the user's repo. */
-      const configResult = runConfig();
+      const configResult = getModRunnerConfig();
       if (configResult.ok) {
         configObj = configResult.value;
       } else {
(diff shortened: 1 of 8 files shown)


packages/codemods/src/command.ts
```diff
@@ -35,25 +37,56 @@ export const yargsParse = (passedArgs: string[]) => {
...
+/* Class responsible for parsing the npx command that runs codemods on a repo.
+   Can either take info from the command line or from a config file. */
 export class CommandParser {
   public parseArgs(passedIn: string[]): CommandParserResult {
     const parsed = yargsParse(passedIn);
     if (parsed.help) {
       return { shouldExit: true, modsFilter: () => true };
     }
+    /* Attempt to locate the modConfig file in the user's repo. */
+    const foundJsonFile = getModRunnerConfig();
```
joschect, line 57: Don't you only want to get the json config if parsed.config is true

packages/codemods/src/command.ts
```diff
@@ -35,25 +37,56 @@ export const yargsParse = (passedArgs: string[]) => {
...
+    let configObj: ModRunnerConfigType = { stringFilters: [], regexFilters: [], includeMods: false };
+    if (parsed.config) {
```
joschect, line 59: I might suggest moving the contents of this if statement into a separate function call. That way the logic of dealing with a config file is kept contained.

packages/codemods/src/command.ts
```diff
@@ -35,25 +38,67 @@ export const yargsParse = (passedArgs: string[]) => {
...
 
     return {
       shouldExit: false,
       modsFilter: filter,
     };
   }
 }
+
+function runConfig(): Result<ModRunnerConfigType, NoOp> {
+  const foundJsonFile = getModRunnerConfig();
```
joschect, line 85: IMO the whole part of this runConfig could be returned from getModRunnerConfig()

PR #14766, opened 2020-08-26
Remove ngrok dependency

ecraig12345:
Remove `ngrok` dependency which was only used for the `local-codepen` script. The script now attempts to use a global install of `ngrok` instead, and instructs the user to install it if not found.

This doesn't fix the main issue which was blocking VMSS use since that was with `screener-ngrok` (not original `ngrok`), but it does speed up install a little since regular `ngrok` also has a slow postinstall script.


ecraig12345:
Merged this anyway because the build failures are from new VMs and clearly unrelated to this change...

PR #14767, opened 2020-08-26
DoNotMerge: Button: how expensive is it to use makeStyles instead of scss?

dzearing:
Want to evaluate a hybrid model of styling. This PR does this:

1. Uses makeStyles, which uses merge-styles under the hood.
2. Still uses flattened styles we introduced with the sass conversion, where the only potential input is theme. This is far better than how we use merge-styles today, which involves style props caching, which are easy to break.

Why this is good:

* SSR will work
* Child windows will work
* RTL flipping will work
* Specificity is normalized
* Opens up the door to compare/contrast different cssinjs renderers

Why this is less good:

* Still runtime evaluation of styles, though done only once per theme/window.


dzearing:
eek builds!

packages/react-theme-provider/src/makeClasses.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * Modifier classnames are added to the root className when the state contains a truthy value
+ * of the same name. For example, when the primary flag is present, the "_primary" modifier
+ * class will be appended to the root className prop.
+ *
+ * Enum classnames are also added to the root className when teh state contains an enum value
+ * which matches the value in the className. for example, when the `size` enum value is `small`,
+ * the "_size_small" enum class will be appended to the root className prop.
+ */
+export const makeClasses = <TStyleSet extends { [key: string]: IStyle }>(
```
xugao, line 35: i think this belongs better in `react-compose`

packages/react-theme-provider/src/getTokens.ts
```diff
@@ -63,12 +63,12 @@ export function getTokens(theme: Theme): Tokens {
...
 
       button: {
-        fontWeight: fonts?.medium?.fontWeight,
+        fontWeight: 600, // fonts?.medium?.fontWeight,
         fontSize: fonts?.medium?.fontSize,
         fontFamily: fonts?.medium?.fontFamily,
         iconSize: fonts?.mediumPlus?.fontSize,
         borderRadius: effects?.roundedCorner2,
-        focusColor: palette?.neutralSecondary,
+        focusColor: palette?.black,
```
xugao, line 71: are you sure? in V7 we have are using `neutralSecondary` for the outline color. however the focus border style has changed in v8, so maybe this is correct

packages/react-theme-provider/src/makeClasses.ts
```diff
@@ -0,0 +1,88 @@
...
+  styleOrFunction: TStyleSet | ((theme: ITheme) => TStyleSet),
+) => {
+  const useStyles = makeStyles(styleOrFunction);
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return (state: any) => {
+    const classes = useStyles();
+    const classNames = Object.keys(classes);
```
xugao, line 43: i think these are slots, not classNames

packages/react-button/src/components/Button/Button.styles.ts
```diff
@@ -0,0 +1,192 @@
...
+      paddingBottom: 0, // 'var(--button-padding)',
+
+      transition: 'var(--button-transition)',
+      whiteSpace: 'var(--button-whiteSpace, nowrap)',
+
+      fontFamily: 'var(--button-fontFamily)',
+      fontSize: 'var(--button-fontSize)',
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      fontWeight: 'var(--button-fontWeight)' as any,
```
xugao, line 43: ah yeah i am fixing this [here](https://github.com/microsoft/fluentui/pull/14696/files#diff-2cefdb4f50d42d742f727bda024ecde6). but since this PR is on hold, i should probably split it out

packages/react-button/src/components/Button/Button.styles.ts
```diff
@@ -0,0 +1,192 @@
...
+      boxShadow: 'var(--button-boxShadow)',
+
+      width: 'var(--button-width)',
+      height: 'var(--button-height)',
+      minHeight: 'var(--button-minHeight)',
+
+      paddingLeft: 16, // 'var(--button-padding)',
+      paddingRight: 16, // 'var(--button-padding)',
+      paddingTop: 0, // 'var(--button-padding)',
+      paddingBottom: 0, // 'var(--button-padding)',
```
xugao, line 34: why this change?

packages/react-button/src/components/Button/Button.styles.ts
```diff
@@ -0,0 +1,192 @@
...
+    width: '100%',
+    maxWidth: '100%',
+  },
+
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  _size_smallest: {
+    '--button-minHeight': 'var(--button-size-smallest)',
+  },
+
+  // eslint-disable-next-line @typescript-eslint/naming-convention
```
xugao, line 155: hm can the rule be updated to support underscore prefix?

packages/react-button/src/components/Button/Button.styles.ts
xugao: Here is the fix: https://github.com/microsoft/fluentui/pull/14773

packages/react-button/src/components/Button/Button.styles.ts
```diff
@@ -0,0 +1,167 @@
+import { makeStyles } from '@fluentui/react-theme-provider';
+
+export const useButtonStyles = makeStyles({
+  displayName: 'Button',
```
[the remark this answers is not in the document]
dzearing: removed.

packages/react-theme-provider/src/makeClasses.ts
dzearing: hmm I'm not convinced that's the right place. 

It depends on makeStyles, which depends on theme.

It relies on the renderer, provided by the themeprovider, to make the styles. (Separate PR.)

My thought is that makeClasses which is in react-compose should be removed. Let react-compose be about utilities that help with composition.

packages/react-theme-provider/src/getTokens.ts
dzearing: Pretty sure this is correct. `neutralSecondary` is not dark enough. I can confirm with design.

packages/react-button/src/components/Button/Button.styles.ts
dzearing: tyty

packages/react-button/src/components/Button/Button.styles.ts
dzearing: resolved as discussed.

packages/react-button/src/components/Button/Button.styles.ts
dzearing: possibly, or we disable it for the file.

dzearing:
Will re-open once split button is updated.

PR #14769, opened 2020-08-26
Updating MessageBar to a function component inside React-Next

czearing:
#### Pull request checklist

#### Description of changes
1. Updating `MessageBar` to a function component inside React-Next.

#### Focus areas to test
`MessageBar`


packages/react-next/src/components/MessageBar/MessageBar.base.tsx
```diff
@@ -1,210 +1,138 @@
...
+    actions: actions !== undefined,
+    truncated: truncated,
+    isMultiline: isMultiline,
+    expandSingleLine: expandSingleLine,
+    className,
+  });
+
+  const iconProps = { iconName: expandSingleLine ? 'DoubleChevronUp' : 'DoubleChevronDown' };
+
+  const getActionsDiv = React.useCallback((): JSX.Element | null => {
```
ecraig12345, line 58: Instead of a function you can just do `const actionsDiv = actions ? <div...> : null` and use it where needed.

packages/react-next/src/components/MessageBar/MessageBar.base.tsx
```diff
@@ -1,210 +1,138 @@
...
+    if (actions) {
+      return <div className={classNames.actions}>{actions}</div>;
+    }
+    return null;
+  }, [actions, classNames.actions]);
 
-  constructor(props: IMessageBarProps) {
-    super(props);
+  const getAnnouncementPriority = React.useCallback((): 'assertive' | 'polite' => {
```
ecraig12345, line 65: I'd recommend moving `getAnnouncementPriority` outside the component body and adding a `messageBarType` arg, since this is just a helper not a callback.

packages/react-next/src/components/MessageBar/MessageBar.base.tsx
```diff
@@ -1,210 +1,138 @@
...
+    return 'polite';
+  }, [messageBarType]);
 
-    initializeComponentRef(this);
-    this.state = {
-      labelId: getId('MessageBar'),
-      // eslint-disable-next-line react/no-unused-state
-      showContent: false,
-      expandSingleLine: false,
+  const getRegionProps = React.useCallback(() => {
```
ecraig12345, line 75: Just do `const regionProps = actions || onDismiss ? { /*props*/ } : {}`

packages/react-next/src/components/MessageBar/MessageBar.base.tsx
```diff
@@ -1,210 +1,138 @@
...
-      return <div className={this._classNames.actions}>{this.props.actions}</div>;
-    }
-    return null;
-  }
+    return hasActions ? regionProps : {};
+  }, [actions, onDismiss, labelId]);
 
-  private _getDismissDiv(): JSX.Element | null {
-    const { onDismiss, dismissIconProps } = this.props;
+  const getDismissDiv = React.useMemo((): JSX.Element | null => {
```
ecraig12345, line 84: This one isn't expensive either, so just do `const dismissDiv = onDismiss ? <div....> : null`. Or in this case I'd probably prefer to just put it inline where it's used.

packages/react-next/src/components/MessageBar/MessageBar.base.tsx
```diff
@@ -1,210 +1,130 @@
...
+          )}
+        </div>
+        <div className={classNames.text} id={labelId} role="status" aria-live={getAnnouncementPriority(messageBarType)}>
+          <span className={classNames.innerText} {...nativeProps}>
+            <DelayedRender>
+              <span>{children}</span>
+            </DelayedRender>
+          </span>
+        </div>
+        {!isMultiline &&
```
ecraig12345, line 87: I'm not sure some of the conditions and operator precedence are quite right. Also it would be clearer to do separate conditions for some parts.

```
// earlier
const dismissButton = onDismiss ? <IconButton .../> : null;

// here
{/* expand/collapse button */ !isMultiline && !actions && truncated && <div>......</div>}
{/* actions */ !isMultiline && actionsDiv}
{/* dismiss */ dismissButton && 
  (isMultiline ? dismissButton : <div className={classNames.dismissSingleLine}>{dismissButton}</div>)}
```

ecraig12345 (commented):
A few more comments but mostly looking good now!

packages/react-next/src/components/MessageBar/MessageBar.base.tsx
```diff
@@ -1,210 +1,117 @@
...
+              disabled={false}
+              className={classNames.expand}
+              onClick={toggleExpandSingleLine}
+              iconProps={expandIconProps}
+              ariaLabel={overflowButtonAriaLabel}
+              aria-expanded={expandSingleLine}
+            />
+          </div>
+        )) ||
+          (onDismiss && <div className={classNames.dismissSingleLine}>{dismissButton}</div>)}
```
ecraig12345, line 110: I think this needs to be `onDismiss && (isMultiline ? dismissButton : <div ...>)`. Right now there are two dismiss buttons sometimes. 😆

ecraig12345 (approved):

PR #14770, opened 2020-08-26
Release build should only build published packages

ecraig12345:
The release build tasks should be scoped to only the packages which are published with beachball. That should make the release builds faster since we won't be building react-northstar and v0 docs.

I also realized that probably the reason we were getting the intermittent "Directory not empty" error on CI builds was because the ScreenerFluent job was missing the `workspace: clean: all` setting. So I'm adding that and deleting the recently added DeleteFiles.

### DIFF
--- a/scripts/monorepo/runPublished.js
+++ b/scripts/monorepo/runPublished.js
@@ -1,0 +1,25 @@
+// @ts-check
+const getAllPackageInfo = require('./getAllPackageInfo');
+const runTo = require('./runTo');
+
+const argv = process.argv.slice(2);
+
+if (argv.length < 1) {
+  console.log(`Usage:
+
+  yarn run:published <script> [<args>]
+
+This command runs <script> for all beachball-published packages (and their dependencies).
+`);
+
+  process.exit(0);
+}
+
+const script = argv[0];
+const rest = argv.slice(1);
+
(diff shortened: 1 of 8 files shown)


scripts/monorepo/runPublished.js
```diff
@@ -0,0 +1,25 @@
...
+`);
+
+  process.exit(0);
+}
+
+const script = argv[0];
+const rest = argv.slice(1);
+
+const beachballPackages = Object.entries(getAllPackageInfo())
+  .filter(([, { packageJson, packagePath }]) => !/[\\/]fluentui[\\/]/.test(packagePath) && packageJson.private !== true)
```
xugao, line 22: hmm, this feels less than ideal. we define `--scope` in `beachball:publish` command and we need to keep this in sync with that. what would happen if we didn't build the package but still let beachball to publish it?

ecraig12345: Not sure what you mean--the publish scope is just `!packages/fluentui/*` (and private packages are implicitly excluded), so this check should do the same thing.

xugao: yes they are the same now. my concern is that we need to make sure they are in sync. what if we updated 
`beachball:publish` to have no scope while missing to update this code here?
with this being said, i don't really have a good suggestion other than adding comments

PR #14787, opened 2020-08-27
use forwardRef for the UnifiedPicker

elisabethcvs:
#### Pull request checklist

#### Description of changes

Set up forwardRef and two new function on the UnifiedPicker, clearInput and focus, so callers can call those functions

### DIFF
--- a/packages/experiments/src/components/UnifiedPicker/UnifiedPeoplePicker/examples/UnifiedPeoplePicker.Example.tsx
+++ b/packages/experiments/src/components/UnifiedPicker/UnifiedPeoplePicker/examples/UnifiedPeoplePicker.Example.tsx
@@ -210,7 +210,7 @@ export const UnifiedPeoplePickerExample = (): JSX.Element => {
   return (
     <>
       <UnifiedPeoplePicker
-        ref={ref}
+        componentRef={ref}
         selectedItemsListProps={selectedPeopleListProps}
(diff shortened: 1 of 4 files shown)


packages/experiments/src/components/UnifiedPicker/UnifiedPeoplePicker/examples/UnifiedPeoplePicker.Example.tsx
```diff
@@ -159,6 +161,15 @@ export const UnifiedPeoplePickerExample = (): JSX.Element => {
   };
 
   const _onInputChange = (filterText: string): void => {
+    // Clear the input if the user types a semicolon or comma
+    // This is meant to be an example of using the forward ref,
+    // feel free to comment out if it impacts your testing
+    const lastCharIndex = filterText.length - 1;
+    const lastChar = filterText[lastCharIndex];
+    if (lastChar === ';' || lastChar === ',') {
```
joschect, line 169: Something to consider is how this will interact with other languages and that there might want to end of word indicators. It might be nice to have this passed in via a prop, but I don't think it's necessary right now.

elisabethcvs: Yup! As this was just an example I wasn't too fussed about internationalizing it

PR #14791, opened 2020-08-27
react-link: remove usage of compose and scss, move latest code to next folder

xugao:
#### Pull request checklist

#### Description of changes
In preping for V8 release, we are reverting `Link` back to use `merge-styles` and remove usage of `compose`.
`Link.styles.ts` is exactly the same as the one in OUFR.

Related issue: #14660

#### Focus areas to test

(optional)


packages/react-link/src/components/Link/index.ts
```diff
@@ -1,4 +1,4 @@
 export * from './Link';
-export * from './LinkBase';
+export * from './Link.base';
 export * from './Link.types';
 export * from './useLink';
```
xugao, line 4: remove

apps/vr-tests/src/stories/LinkNext.stories.tsx
```diff
@@ -3,41 +3,31 @@ import * as React from 'react';
...
+        )
+        .executeScript("document.getElementsByClassName('ms-Link')[0].focus()")
+        .snapshot('focus', { cropTo: '.testWrapper' })
+        .executeScript(
+          "document.getElementsByClassName('testWrapper')[0].classList.remove('ms-Fabric--isFocusVisible')",
+        )
+        .hover('.ms-Link')
+        .snapshot('hover', { cropTo: '.testWrapper' })
+        .click('.ms-Link')
+        .hover('.ms-Link')
```
khmakoto, line 25: Maybe we should remove this second `.hover`?

packages/react-link/src/components/Link/index.ts
```diff
@@ -1,4 +1,3 @@
 export * from './Link';
-export * from './LinkBase';
+export * from './Link.base';
 export * from './Link.types';
-export * from './useLink';
```
khmakoto, line 4: Are we not exporting the state hooks then? Just curious 😃

xugao: if we export from `index`, we are officially supporting it (cannot make breaking changes to it). This is not something we want to do now

khmakoto: Sounds good.

apps/vr-tests/src/stories/LinkNext.stories.tsx
xugao: I am resolving this fix in this PR https://github.com/microsoft/fluentui/pull/14809/files#diff-cc28f495c7f67ad88dc920d1100b79f0 to avoid needing re-run the build :p

PR #14797, opened 2020-08-28
feat(a11y-testing): add package

mnajdova:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/react-button/src/components/ToggleButton/checkedBehaviorDefinition.ts
```diff
@@ -0,0 +1,30 @@
...
+        bubbles: true,
+        cancelable: true,
+      }),
+    )
+    .hasAttribute('aria-pressed', 'true')
+    .description('after the element has been clicked'),
+  BehaviorRule.root()
+    .forProps({ checked: undefined, defaultChecked: true })
+    .afterEvent(
+      'onClick',
```
mnajdova, line 21: `"onClick"` as name of the handler is actually required only for the hooks tests. If we decide to go with only components test, this can safely be removed, as in the `ComponentTestFacade` we are just dispatching the event which is currently second argument.

packages/react-avatar/.eslint.json
```diff
@@ -0,0 +1,8 @@
+{
```
mnajdova, line 1: I couldn't commit the merge changes if I didn't replace `tslint.json` with `.eslint.json` 😕

packages/react-button/src/components/ToggleButton/checkedBehaviorDefinition.ts
```diff
@@ -0,0 +1,30 @@
+import { BehaviorRule, Rule } from '@fluentui/a11y-testing';
+
+export const checkedBehaviorDefinition: Rule[] = [
+  BehaviorRule.root()
+    .doesNotHaveAttribute('aria-pressed')
+    .description('by default'),
+  BehaviorRule.root()
+    .afterEvent(
```
jurokapsiar, line 8: could you also add afterClick(), so that we do not have to create the mouse event in every test?

mnajdova: yeah that makes sense, it can be extended later for other handlers we may add 👍

layershifter:
Have been added in #15040. Closing for housekeeping 🏡

PR #14799, opened 2020-08-28
chore: disable test, add iterations to existing

layershifter:
This PR makes few small changes to existing tests.

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Button/Performance/ButtonOverridesMiss.perf.tsx
+++ b/packages/fluentui/docs/src/examples/components/Button/Performance/ButtonOverridesMiss.perf.tsx
@@ -167,7 +167,7 @@ const ButtonOverridesMissPerf = () => (
   </Provider>
 );
 
-ButtonOverridesMissPerf.iterations = 100;
+ButtonOverridesMissPerf.iterations = 50;
 ButtonOverridesMissPerf.filename = 'ButtonOverridesMiss.perf.tsx';
 
 export default ButtonOverridesMissPerf;
--- a/packages/fluentui/docs/src/examples/components/Button/Performance/ButtonOverridesMiss.perf.tsx
+++ b/packages/fluentui/docs/src/examples/components/Button/Performance/ButtonOverridesMiss.perf.tsx
@@ -167,4 +167,7 @@ const ButtonOverridesMissPerf = () => (
   </Provider>
 );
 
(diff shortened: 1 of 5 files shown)


packages/fluentui/docs/src/examples/components/Button/Performance/ButtonOverridesMiss.perf.tsx
```diff
@@ -167,4 +167,7 @@ const ButtonOverridesMissPerf = () => (
   </Provider>
 );
 
+ButtonOverridesMissPerf.iterations = 100;
+ButtonOverridesMissPerf.filename = 'ButtonOverridesMiss.perf.tsx';
```
layershifter, line 171: This should improve stability of these perf tests

packages/fluentui/e2e/tests/popupWithCloseInContent-test.ts
```diff
@@ -9,7 +9,7 @@ describe('Popup With Close in the content and hover trigger', () => {
     await e2e.gotoTestCase(__filename, popupTrigger);
   });
 
-  it('A click on content should not close popup, a click button should close popup', async () => {
+  xit('A click on content should not close popup, a click button should close popup', async () => {
```
layershifter, line 12: This test is unstable due a bug in `Popup`

packages/fluentui/e2e/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "extends": "../../../scripts/typescript/tsconfig.fluentui",
   "compilerOptions": {
-    "module": "esnext",
     "allowSyntheticDefaultImports": true,
-    "types": ["node", "jest", "webpack-env"]
+    "module": "esnext",
+    "types": ["node", "jest", "webpack-env"],
+    "skipLibCheck": true
```
layershifter, line 7: Should improve build speed a bit.

pompomon (approved):

packages/fluentui/e2e/tests/popupWithCloseInContent-test.ts
layershifter: https://github.com/microsoft/fluentui/issues/14800

ecraig12345:
Merging this since the failures are in screener (not the edited tests) and therefore unrelated.

PR #14801, opened 2020-08-28
RFC: add initial spec for Text component

layershifter:
This PR adds the initial spec for `Text` component add mainly concentrated on the approach that we would like to take to go.

---

I analyzed existing UI libraries/kits and actually there two approaches.

### `size`, `weight`, etc.

A set of props (https://v2.grommet.io/text) that allows to configure `Text` component explicitly:

```tsx
<Text size={300} weight='regular' />
```

### `variant`

- `variant` like (https://www.welcome-ui.com/components/text) that defines explicit text usage:

```tsx
<Text variant="body" />
<Text variant="subtitle1" />
```

---

This PR is labeled as RFC, so please share your thoughts about the approach that we should take.

### DIFF
--- a/specs/Text.md
+++ b/specs/Text.md
@@ -1,0 +1,47 @@
+# Text FluentUI spec
+
+The Text component consistently styles and formats occurrences of text.
+
+## Prior Art/Examples <a href="#prior-art" id="prior-art"></a>
+
+- [Ant Design](https://ant.design/components/typography/)
+- [BaseWeb](https://baseweb.design/blog/base-web-v9#typography)
+- [Blueprint](https://blueprintjs.com/docs/#core/components/text)
+- [Chakra UI](https://chakra-ui.com/text)
+- [Elastic UI](https://elastic.github.io/eui/#/display/text)
+- [Evergreen](https://evergreen.segment.com/components/typography#heading_text_component)
+- [Grommet](https://v2.grommet.io/text)
+- [Fabric UI](https://developer.microsoft.com/en-us/fluentui#/controls/web/text)
+- [Fluent UI Northstar](https://fluentsite.z22.web.core.windows.net/components/text/definition)
+- [Fomantic UI](https://fomantic-ui.com/elements/text.html)
+- [Lightning Design System](https://www.lightningdesignsystem.com/utilities/text/#site-main-content)
+- [Primer](https://primer.style/components/Text)
+- [Ring UI](https://jetbrains.github.io/ring-ui/master/index.html?path=/story/components-text--basic)
+- [Vuetify](https://vuetifyjs.com/en/styles/text-and-typography/)
+- [Welcome UI](https://www.welcome-ui.com/components/text)
+
+## API
+
+### Props
+
+| Attribute Name | Type   | Default Value | Description                                      |
+| -------------- | ------ | ------------- | ------------------------------------------------ |
+| `variant`      | `enum` |               | A text can define its look via defined variants. |
+
+## Slots
+
+| Name | Considerations     |
+| ---- | ------------------ |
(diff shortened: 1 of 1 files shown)


levithomason:
The individual props are a lower-level API and more flexible but also require the implementer to configure them correctly for the design language. The variant prop is a higher level abstraction that will pre-configure the lower level values so it is less flexible but it is also easier to use.

My worry with the lower-level props is that engineers will configure them wrong and they will be harder to update with tokens because there is no semantic meaning to update. I lean toward the variant option as it allows future design updates without needing syntax changes (tokens only.)

### Proposal

Allow both? The variant prop could be used to select a variant while the individual props could be used to override it. This could lead to nothing more than a bunch of overridden text components, though. If others agree with this concern then I would vote that we go with the variant prop only. Any overrides that are truly necessary could be accomplished with styles and or tokens.

specs/Text.md
```diff
@@ -0,0 +1,48 @@
...
+- [Vuetify](https://vuetifyjs.com/en/styles/text-and-typography/)
+- [Welcome UI](https://www.welcome-ui.com/components/text)
+
+## API
+
+### Props
+
+| Attribute Name | Type   | Default Value | Description                                                              |
+| -------------- | ------ | ------------- | ------------------------------------------------------------------------ |
+| `size`         | `enum` | `400`         | A text can define the size of a font.                                    |
```
assuncaocharles, line 29: Should we have a weight enum as well?

specs/Text.md
```diff
@@ -0,0 +1,48 @@
...
+| `weight`       | `enum` | `regular`     | A text can set how thick or thin characters in text should be displayed. |
+
```
dzearing, line 31: Here are some questions:

Should we include formatting flags such as italic, bold, etc?

I'd love to see `variants` as well, but maybe it can be added separately. The idea being that a variant can be defined in the theme and opted into in the application code.

Another couple things:

When should we use Text? I remember struggling with this in the past. At first we thought everywhere, but then perf concerns.

I'd expect text to be inline runs by default, like spans. There are cases where they should be block and overflow. Should this be a flag, like `fluid` on button for consistency?

And, just from a usage scenario, this also feels like icon, where there may be many variants. It's nice to be able to say:

```tsx
import { H1, H2, H3, Paragraph, Code } from '@fluentui/react';

<H1>I am themed header text</H1>
<Paragraph>I am themed paragraph content</Paragraph>
```

layershifter:
### `variant`

> The variant prop could be used to select a variant while the individual props could be used to override it.

We had an offline conversation with @levithomason and agreed for N* to go with `variant` only for now. @dzearing this is aligned with the approach in v7. The main benefit that we see that variants are defined by the design team and if there will be any updates it will be easier to update consumers' apps. I updated the spec based on this.

specs/Text.md
layershifter: > Should we include formatting flags such as italic, bold, etc?

We discussed this item with @levithomason. Currently it makes a sense to go `variant`, on next iteration I will add another proposal to solve this. I see two options:

```tsx
<Text bold />
// or
<Text as="b" />
```

_I like the second as it's semantically correct._ But how to archive `bold` and `italic` at the same time? 😼 

---

> When should we use Text? I remember struggling with this in the past. At first we thought everywhere, but then perf concerns.

For N* the answer is simple: use always for any text as it's required for internationalization (as `Text` adds `dir=auto`) as for styling purposes.

---

> And, just from a usage scenario, this also feels like icon, where there may be many variants. It's nice to be able to say:

We will definitely have `Header` to represent `h*` tags.

dzearing (approved):
lgtm!

PR #14803, opened 2020-08-28
ThemeProvider: Export ThemeContext

xugao:
#### Pull request checklist

#### Description of changes
Export `ThemeContext` and add documentation in README

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-theme-provider-2020-08-28-11-34-30-xgao-themecontext.json
+++ b/change/@fluentui-react-theme-provider-2020-08-28-11-34-30-xgao-themecontext.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Export ThemeContext.",
+  "packageName": "@fluentui/react-theme-provider",
+  "email": "xgao@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-08-28T18:34:30.015Z"
+}
--- a/packages/react-theme-provider/README.md
+++ b/packages/react-theme-provider/README.md
@@ -67,9 +67,56 @@ export const App = () => (
 );
 ```
 
-## Create classes for React components based on theme
+## Accessing theme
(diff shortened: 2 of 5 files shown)


kelseyyoung:
As per our conversation from Teams, it would be a really cool feature to programmatically define what the "default" value of ThemeContext is. Right now it's "undefined", which because we're using Typescript is always a pain to handle. We can do this manually by wrapping ThemeContext.Consumer and checking if 'theme' is undefined (aka no Provider) and then providing our own app's "default theme". It would just be a bonus to be able to tell ThemeContext to take our app's "default theme" as the default value

Not even sure this is possible with React's Context, but something to look into :)

xugao:
@kelseyyoung - I actually need to distinguish if the context is default or it's provided using `Provider` (logic in `useTheme`). But I also agree with the pain of having `undefined` typing for users of `ThemeContext.Customer` like you. However I still lean towards having the correct typing / no default context.  What do you think?

kelseyyoung:
@xugao well like I said it's not a huge thing. I'm thinking our solution looks like this:

```
// User of ThemeContext.Consumer
<ThemeContext.Consumer>
  {(contextTheme: Theme | undefined) => (
    Theme theme = getAppTheme(contextTheme);
    getStyles(theme);

// getAppTheme function
const getAppTheme = (theme: Theme | undefined): Theme => { 
  if (!theme) {
    return defaultTheme; // Our app's default theme
  }
  return theme;
}
```

And because we're abstracting this out into a singular "component", if things change we can make that change in one place. That way the signature of getStyles can be:

`const getStyles = (theme: Theme) => {`

and we get rid of the "undefined". So this isn't hard for us. Having a way to programmatically say "if the consumer has no theme, fill in our own" would also be cool. But really just a nice to have

So I agree with what you have now being fine as well

PR #14804, opened 2020-08-28
Updating FocusTrapZone to a function component

czearing:
#### Pull request checklist

#### Description of changes

1. Updated FocusTrapZone to a function component.
2.  Added perf and vr tests.
3. Changed FocusStack to use an id instead of the component object.
4. Rewrote and added tests for FocusStack.

#### Focus areas to test
`FocusTrapZone`


packages/react-next/src/components/Modal/__snapshots__/Modal.test.tsx.snap
```diff
@@ -125,7 +125,6 @@ exports[`Modal renders Draggable Modal correctly 1`] = `
           }
         >
           <div
-            aria-hidden={true}
```
ecraig12345, line 128: Is this expected?

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -139,204 +104,278 @@ export class FocusTrapZone extends React.Component<IFocusTrapZoneProps, {}> impl
...
-  }
-
-  private _focusAsync(element: HTMLElement): void {
-    if (!this._isBumper(element)) {
-      focusAsync(element);
+    if (firstFocusableChild) {
+      focusAsync(firstFocusableChild);
     }
-  }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- internal state will not change
```
ecraig12345, line 129: Probably better to leave the rule enabled and add `internalState` to the deps list--that way the rule will be able to help catch any deps which might be added in the future.

packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.test.tsx
```diff
@@ -627,7 +627,7 @@ describe('FocusTrapZone', () => {
             forceFocusInsideTrap={false}
             focusPreviouslyFocusedInnerElement={focusPreviouslyFocusedInnerElement}
             data-is-focusable={true}
-            ref={focusTrapZoneRef}
+            componentRef={focusTrapZoneRef}
```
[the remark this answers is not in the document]
czearing: It should still work as expected. ref was changed to componentRef to allow access to the public focus method in `FocusTrapZone`.

khmakoto (approved):

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -6,131 +6,97 @@ import {
...
-    const { focusPreviouslyFocusedInnerElement, firstFocusableSelector } = this.props;
-
+      focus,
+    }),
+    [previouslyFocusedElement, focus],
+  );
+};
+
+export const FocusTrapZone: React.FunctionComponent<IFocusTrapZoneProps> & {
+  focusStack: string[];
```
xugao, line 45: this should not be here. we should just define `focusStack` in file scope

dzearing: focusStack is a static. Minimally it needs to be file scoped, but the tests also need access to it, which is why it's attached as a static.

ecraig12345:
Looks like you're missing snapshot updates.

packages/react-next/src/components/FocusTrapZone/FocusTrapZone.tsx
xugao: should we avoid adding this to public api then? we can still access it the same way in our tests, just without the typing:
```
export const FocusTrapZone: React.FunctionComponent<IFocusTrapZoneProps> & {
   focusStack: string[];
};
```

xugao: we can take this as follow up and address in future PR if needed

dzearing: We looked at removing, but I don't think it's worth addressing. Recommend we leave as is.

PR #14810, opened 2020-08-29
DetailsList: Set selectAllVisibility to hidden when there are no items to select

erichdev:
#### Pull request checklist

#### Description of changes

The "Select all" button is currently visible in DetailsList even where there are no items to select. This can be especially confusing for keyboard users using a screenreader.

This change sets `selectAllVisibility` to `hidden` when there are no items and makes the the checkbox not focusable.

#### Focus areas to test

(optional)


JustSlone:
This looks good, @ThomasMichon any concerns taking this?

@erichdev, this will need to be updated based on some changes to DetailsList in master.

khmakoto (changes requested):

PR #14816, opened 2020-08-31
Feat/datepicker adding more accessible behaviors

rymeskar:
#### Pull request checklist

#### Description of changes

Adding even more accessible behaviors

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Datepicker/DatepickerCalendarHeader.tsx
```diff
@@ -192,6 +192,12 @@ DatepickerCalendarHeader.propTypes = {
   weekNumberFormatString: PropTypes.string,
   selectedDateFormatString: PropTypes.string,
   todayDateFormatString: PropTypes.string,
+  calendarCellFormatString: PropTypes.string,
```
pompomon, line 195: Please why do we need this props here, if they are not used anywhere? Or am I missing something?

rymeskar: It's because we have flattened the localization props. 

This 'long-list' of prop types and default values actually really favors the 'composite' approach for localization strings.

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
```diff
@@ -36,6 +36,7 @@ import { DatepickerCalendarHeader } from './DatepickerCalendarHeader';
 import { DatepickerCalendarHeaderAction } from './DatepickerCalendarHeaderAction';
 import { DatepickerCalendarHeaderCell } from './DatepickerCalendarHeaderCell';
 import { validateDate } from './validateDate';
+import { format } from '@uifabric/utilities';
```
ecraig12345, line 39: Adding the dependency on `@uifabric/utilities` is likely to cause build problems at the moment, so you may need to either move the `format` function (from `packages/utilities/src/string.ts`) to an existing shared location if a suitable one exists, or make a local copy for now.

rymeskar: Can you please expand why the dependency would cause problems?

The @uifabric/utilities package is already in the dependency chain within the fluentui/react-bindings. So the overall's react-northstar's package flattened dependencies do not change.

packages/fluentui/react-northstar/src/components/Datepicker/DatepickerCalendarHeader.tsx
rymeskar: Addressed the comment.

pompomon (approved):

packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx
ecraig12345: Oh okay I hadn't realized that. In that case it should be fine. The reason I was cautious in general is we've run into some difficulty importing v7 packages from v0 in that past, but that may be resolved now.

rymeskar: I understand now.

Thanks a lot for explaining the background. :)

PR #14820, opened 2020-08-31
feat(Datepicker): Add Form.Datepicker

pompomon:
#### Pull request checklist

#### Description of changes
Add `Form.Datepicker` component and show the example of errorMessage handling

image


mnajdova (approved):
LGTM, let's add changelog entry please

packages/fluentui/docs/src/examples/components/Form/Usage/FormExampleDatepicker.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import { Form, Button } from '@fluentui/react-northstar';
+
+const FormExampleDatepicker = () => {
+  const [errorMessage, setErrorMessage] = React.useState(null);
+  const errorMessageHandler = (e, v) => {
+    setErrorMessage(v.error);
```
rymeskar, line 7: Related:
Should we call the `onDateChangeError` handler even when setting the initial [errorMessage](https://github.com/microsoft/fluentui/blob/a7bea67257817216ec117661f313dba07625a819/packages/fluentui/react-northstar/src/components/Datepicker/Datepicker.tsx#L169)?  Currently, this errorMessage is actually lost.

assuncaocharles: I would guess so since the message can change when the type of error changes

pompomon: Agreed with @rymeskar to implement this change in a separate PR, since it doesn't affect `Form.Datepicker` usage/logic

PR #14821, opened 2020-08-31
fix/Keeping input width fixed

rymeskar:
#### Pull request checklist

#### Description of changes

* Keep input width fixed when error prop is controlled. 

[Example](https://microsofteur-my.sharepoint.com/:v:/g/personal/karymes_microsoft_com/Eb2ROQR6h3VHj-6x_NF3BmEBr-gEFxJzd0G7neeoFtaw9g?e=UTOp0m) of wrong behavior that got fixed.

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Input/Input.tsx
```diff
@@ -189,7 +189,10 @@ export const Input = compose<'input', InputProps, InputStylesProps, {}, {}>(
       ? ((required && hasValue) || showSuccessIndicator) && !error
       : showSuccessIndicator;
 
-    const hasIcon = !!icon || showSuccessIndicator || (required && isShowSuccessIndicatorUndefined) || !!error;
+    const isErrorUndefined = typeof error === 'undefined';
+
+    const hasIcon =
```
assuncaocharles, line 194: this will make the input always keep the width for icon even when we have empty string for erroMessage, is that the desired?

assuncaocharles: Actually it will be always since in the `FormField` we have `error: !!errorMessage || null,` what will always evaluate for true, so if we want push this probably have to update `FormField` to have `error: !!errorMessage || undefined`

assuncaocharles: Or maybe use `_.isNil(value)` instead `typeof ...`?

rymeskar: The only change should happen in the scenario where you control the error state externally.

You perform the validation outside of the component.

rymeskar: The only change caused by this PR should happen in the scenario where you control the error state externally.

You perform the validation outside of the Input component and pass the result as props.

==> in such error controlled state you always pass either true or false value; you don't leave the prop undefined.

Now, when controlling the error , you want to keep the width of Input consistent so that the UI does not jump for you when writing. 

I agree that  _.isNull is more suitable though due to the FormInput

assuncaocharles: You can see it's not totally true by the screen testers for `FormInlineExample` for example

rymeskar: I completely agree, already changed to isNil :) 


@assuncaocharles can you double-check?

assuncaocharles: @rymeskar LGTM, let's see how screener will behave : )

assuncaocharles (approved):

pompomon (approved):

PR #14823, opened 2020-08-31
chore: update Screener config for Northstar

layershifter:
### 📄 Story behind

In this repo we have to separate Screener projects and previously it was impossible to integrate them both 😭 Only the Screener project for Fabric had fully integration, for Northstar it was required to rerun the whole CI job after regressions were accepted 🤕 

With @miroslavstastny we paired to create a [custom Github App](https://github.com/layershifter/screener-proxy) to have a custom integration with Screener for Fluent UI React Northstar. These checks are visible in each PR now:

image

---

This PR finishes the integration, now CI job will not fail on a Screener failure (now we have a separate check). We tested multiple scenarios with @miroslavstastny and then later I confirmed with @pompomon that the custom check works for forks and will be marked as successful once regressions will be accepted in Screener.

### DIFF
--- a/scripts/screener/screener.config.js
+++ b/scripts/screener/screener.config.js
@@ -64,21 +64,9 @@ module.exports = {
   // screenshot every example in maximized mode
   states: require('./screener.states').default,
 
-  // CircleCI config
-  // ...(process.env.CI && {
-  //   baseBranch: 'master',
-  //   // Disable exit code to fail in Github Actions
-  //   // failureExitCode: 0,
-  //   // GITHUB_REF can be:
-  //   // - refs/heads/feature-branch-1 for "push"
-  //   // - refs/pull/2040/merge for "pull_request"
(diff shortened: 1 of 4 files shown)


scripts/screener/screener.config.js
```diff
@@ -64,21 +64,9 @@ module.exports = {
...
-  // ...(process.env.CI && {
-  //   baseBranch: 'master',
-  //   // Disable exit code to fail in Github Actions
-  //   // failureExitCode: 0,
-  //   // GITHUB_REF can be:
-  //   // - refs/heads/feature-branch-1 for "push"
-  //   // - refs/pull/2040/merge for "pull_request"
-  //   branch: process.env.GITHUB_REF.split('/')[2],
-  //   commit: process.env.GITHUB_SHA
-  // })
```
layershifter, line 77: We are not going to use Github Actions for this so this can be deleted

PR #14824, opened 2020-08-31
Export a few extras for partners who are using deep imports to pull them

dzearing:
We recommend partners only importing from either package imports:

```
import { Button } from '@fluentui/react'
```

This doesn't always work for partners, especially for those who are AMD bundling or are struggling to get tree shaking working correctly with their bundle. For these scenarios we also allow "top-level" imports, which are source files directly under the `lib` folder:

```jsx
import { Button } from '@fluentui/react/lib/button';
```

However we find many partners going far deeper than that and importing things from deeply nested imports:

```jsx
import { IButtonClassNames } from 'office-ui-fabric-react/lib/component/Button/Button.classNames';
```

We never recommend doing this, because we can't expect file names, folder names, and even their casing to become part of the public contract. Just fixing a casing problem would break partners.

But often partners resort to this because the types they need are not being exported by either of the first 2 options.

This PR fixes some of those scenarios as a stopgap solution. The long term solution is really sub-packages. Sub-packages enforce that everything related to a concern can be contained in a subpackage, and that pulling from the suite would effectively make it impossible to pull a non-contractual file, without taking a direct dependency on the sub-package. It makes it slightly harder to do the wrong thing.


packages/office-ui-fabric-react/src/DragDrop.ts
```diff
@@ -0,0 +1 @@
+export * from 'office-ui-fabric-react/lib/DragDrop';
```
ecraig12345, line 1: Can you make similar files in `react` and `react-next`?

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.types.ts
```diff
@@ -402,3 +401,23 @@ export interface IComboBoxOptionStyles extends IButtonStyles {
    */
   optionTextWrapper: IStyle;
 }
+
+export interface IComboBoxClassNames {
```
ecraig12345, line 405: These should probably have docCategory comments so they show up on the website (especially since they're referenced from existing public interfaces).
```
/**
 * {@docCategory ComboBox}
 */
```

packages/office-ui-fabric-react/src/DragDrop.ts
dzearing: done.

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.types.ts
dzearing: I'm not sure really we want to have any of these additional types exposed.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -1238,9 +1274,17 @@ export const FontIcon: React.FunctionComponent<IFontIconProps>;
 // @public (undocumented)
 export function getAllSelectedOptions(options: ISelectableOption[], selectedIndices: number[]): ISelectableOption[];
 
+// @public
+export function getAriaDescribedBy(keySequences: string[]): string;
+
 // @public (undocumented)
 export function getBackgroundShade(color: IColor, shade: Shade, isInverted?: boolean): IColor | null;
 
+// Warning: (ae-forgotten-export) The symbol "IWindowWithSegments" needs to be exported by the entry point index.d.ts
```
ecraig12345, line 1283: Seems like this ought to be exported

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -39,11 +39,23 @@ import { Omit } from '@uifabric/utilities';
...
 import * as PropTypes from 'prop-types';
 import * as React from 'react';
+import { Rectangle } from '@uifabric/utilities';
 import { Selection } from '@uifabric/utilities';
 import { SELECTION_CHANGE } from '@uifabric/utilities';
 import { SelectionDirection } from '@uifabric/utilities';
 import { SelectionMode } from '@uifabric/utilities';
 
+// @public (undocumented)
+export const __positioningTestPackage: {
```
ecraig12345, line 49: What's this?

packages/office-ui-fabric-react/src/DragDrop.ts
ecraig12345, line 1: This is circular as written?

packages/office-ui-fabric-react/src/utilities/positioning/positioning.types.ts
```diff
@@ -100,18 +100,6 @@ export interface IPosition {
   [key: string]: number | undefined;
 }
 
-// eslint-disable-next-line @typescript-eslint/naming-convention
-export interface Point {
```
ecraig12345, line 104: This should probably re-export Point and IPoint from utilities for now

packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.types.ts
dzearing: resolved

packages/office-ui-fabric-react/src/DragDrop.ts
dzearing: hmmm that's odd. I might have copied the file in the wrong direction; will fix!

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
dzearing: This is why I really don't want to export these from the index.

dzearing: I also do not want to export individual things here as I don't know what is or isn't used by the partner.

packages/office-ui-fabric-react/src/DragDrop.ts
dzearing: resolved.

packages/office-ui-fabric-react/src/utilities/positioning/positioning.types.ts
dzearing: fixed.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
dzearing: I'm gonna keep going here though. I've updated this to not export the test exports.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
dzearing: It *is exported*... 

You can find it lower down. API extractor issue.

dzearing: fixed.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
dzearing: Fixed.

packages/react-next/src/index.ts
```diff
@@ -20,6 +20,7 @@ export * from './DetailsList';
 export * from './Dialog';
 export * from './Divider';
 export * from './DocumentCard';
+export * from './DragDrop';
```
xugao, line 23: missing the same change in `@fluentui/react` package

packages/office-ui-fabric-react/src/utilities/positioning/positioning.ts
```diff
@@ -1,19 +1,16 @@
...
-  Point,
   ICalloutPositionedInfo,
   ICalloutBeakPositionedInfo,
   IPositionProps,
   ICalloutPositionProps,
   RectangleEdge,
+  IWindowWithSegments,
 } from './positioning.types';
-
-export class Rectangle extends FullRectangle {
```
xugao, line 14: are you sure you want to just remove this ? we can leave this but not exporting it in `positioning/index.ts`.

change/@fluentui-react-next-2020-09-08-15-51-40-chore-export-imports.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Adding DragDrop and Positioning top level export files.",
```
xugao, line 3:
```suggestion
  "comment": "Exporting some of the interfaces that are being deep imported in product code.",
```

xugao (approved):

packages/office-ui-fabric-react/src/utilities/positioning/positioning.ts
dzearing: I talked with Jon. I think it's the right thing to do. The whole reasoning for this is literally the 2 casts below.

packages/react-next/src/index.ts
dzearing: The `@fluentui/react` package does this:

```
import './version';
export * from 'office-ui-fabric-react';
```

PR #14828, opened 2020-08-31
fix: button and accordion controls not inheriting font-family styles and update fast packages

SethDonohue:
#### Pull request checklist

#### Description of changes

- Fixes a bug where the `button` element on the shadow DOM for the accordion-item and button components does not inherit the font-family as it should.
- Also updates fast dependencies

### DIFF
--- a/packages/web-components/package.json
+++ b/packages/web-components/package.json
@@ -94,8 +94,8 @@
   "dependencies": {
     "@microsoft/fast-colors": "^5.1.0",
     "@microsoft/fast-components-styles-msft": "^4.29.0",
-    "@microsoft/fast-element": "^0.16.0",
-    "@microsoft/fast-foundation": "^1.7.0",
+    "@microsoft/fast-element": "^0.17.0",
+    "@microsoft/fast-foundation": "^1.7.1",
     "tslib": "^1.13.0"
   }
 }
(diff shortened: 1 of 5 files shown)


packages/web-components/package.json
```diff
@@ -94,8 +94,8 @@
   "dependencies": {
     "@microsoft/fast-colors": "^5.1.0",
     "@microsoft/fast-components-styles-msft": "^4.29.0",
-    "@microsoft/fast-element": "^0.16.0",
-    "@microsoft/fast-foundation": "^1.7.0",
+    "@microsoft/fast-element": "^0.17.0",
+    "@microsoft/fast-foundation": "^1.7.1",
```
chrisdholt, line 98: Did you run yarn after this update? I'd expect to see an update to the yarn lockfile here.

packages/web-components/package.json
```diff
@@ -94,8 +94,8 @@
   "dependencies": {
     "@microsoft/fast-colors": "^5.1.0",
     "@microsoft/fast-components-styles-msft": "^4.29.0",
-    "@microsoft/fast-element": "^0.16.0",
-    "@microsoft/fast-foundation": "^1.7.0",
```
chrisdholt, line 98: I think this *could* stay at 1.7.x - this will take 1.7.1 as acceptable.

SethDonohue: I was curious about that, I'll switch it back to ^1.17.0

packages/web-components/package.json
SethDonohue: I did, but looks like I forgot to commit the lockfile!

chrisdholt:
Thanks @SethDonohue!

PR #14829, opened 2020-08-31
Updating MaskedTextField to a function component within react-next

### LINKED ISSUE #14287: MaskedTextField: strict mode compliance (UNSAFE_, easy/medium)
MaskedTextField updates stored mask info and the current display value in `componentWillReceiveProps`. This could likely be moved to `componentDidUpdate` without changes.

czearing:
#### Pull request checklist
- [x] Fixes #14287

#### Description of changes
1. Updating `MaskedTextField` to a function component.
2. Adding temporary element ref to `Textfield` to allow for `MaskedTextField` to use forwardRef.

#### Focus areas to test
`MaskedTextField`


ecraig12345:
This might be a good opportunity to make a new `IMaskedTextFieldProps extends ITextFieldProps` and move `mask`, `maskChar`, and `maskFormat` there (and deprecate them in OUFR `ITextFieldProps`). I'm not sure why those were added to `ITextFieldProps` previously instead of having a separate interface.

packages/office-ui-fabric-react/src/components/TextField/TextField.types.ts
```diff
@@ -375,3 +359,40 @@ export interface ITextFieldStyles {
    */
   subComponentStyles: ITextFieldSubComponentStyles;
 }
+
+/**
+ * MaskedTextField component props.
+ * {@docCategory MaskedTextField}
```
ecraig12345, line 365: This should stay as TextField since we don't have a separate doc page for MaskedTextField
```suggestion
 * {@docCategory TextField}
```

packages/office-ui-fabric-react/src/components/TextField/TextField.types.ts
```diff
@@ -375,3 +359,40 @@ export interface ITextFieldStyles {
...
+ */
+export interface IMaskedTextFieldProps extends ITextFieldProps {
+  /**
+   * Only used by MaskedTextField:
```
ecraig12345, line 369:
```suggestion
```

packages/office-ui-fabric-react/src/components/TextField/TextField.types.ts
```diff
@@ -375,3 +359,40 @@ export interface ITextFieldStyles {
...
+   * '9': [0-9]
+   * 'a': [a-zA-Z]
+   * '*': [a-zA-Z0-9]
+   *
+   * @example `Phone Number: (999) 999-9999`
+   */
+  mask?: string;
+
+  /**
+   * Only used by MaskedTextField:
```
ecraig12345, line 382:
```suggestion
```

packages/office-ui-fabric-react/src/components/TextField/TextField.types.ts
```diff
@@ -375,3 +359,40 @@ export interface ITextFieldStyles {
...
+   * The character to show in place of unfilled characters of the mask.
+   * @defaultvalue '_'
+   */
+  maskChar?: string;
+
+  /**
+   * Only used by MaskedTextField:
```
ecraig12345, line 389:
```suggestion
```

packages/office-ui-fabric-react/src/components/TextField/TextField.types.ts
```diff
@@ -252,33 +252,17 @@ export interface ITextFieldProps extends React.AllHTMLAttributes<HTMLInputElemen
...
-   * Only used by MaskedTextField:
-   * The masking string that defines the mask's behavior.
-   * A backslash will escape any character.
-   * Special format characters are:
-   * '9': [0-9]
-   * 'a': [a-zA-Z]
-   * '*': [a-zA-Z0-9]
-   *
-   * @example `Phone Number: (999) 999-9999`
+   * @deprecated Deprecated, use IMaskTextField's mask.
```
ecraig12345, line 255:
```suggestion
   * @deprecated Only used by `MaskedTextField`, which now has a separate `IMaskedTextFieldProps` interface.
```

packages/office-ui-fabric-react/src/components/TextField/TextField.types.ts
```diff
@@ -252,33 +252,17 @@ export interface ITextFieldProps extends React.AllHTMLAttributes<HTMLInputElemen
...
    */
   mask?: string;
 
   /**
-   * Only used by MaskedTextField:
-   * The character to show in place of unfilled characters of the mask.
-   * @defaultvalue '_'
+   * @deprecated Deprecated, use IMaskTextField's maskChar.
```
ecraig12345, line 260:
```suggestion
   * @deprecated Only used by `MaskedTextField`, which now has a separate `IMaskedTextFieldProps` interface.
```

packages/office-ui-fabric-react/src/components/TextField/TextField.types.ts
```diff
@@ -252,33 +252,17 @@ export interface ITextFieldProps extends React.AllHTMLAttributes<HTMLInputElemen
...
 
   /**
-   * Only used by MaskedTextField:
-   * An object defining the format characters and corresponding regexp values.
-   * Default format characters: \{
-   *  '9': /[0-9]/,
-   *  'a': /[a-zA-Z]/,
-   *  '*': /[a-zA-Z0-9]/
-   * \}
+   * @deprecated Deprecated, use IMaskTextField's maskFormat.
```
ecraig12345, line 265:
```suggestion
   * @deprecated Only used by `MaskedTextField`, which now has a separate `IMaskedTextFieldProps` interface.
```

packages/react-next/src/components/TextField/TextField.types.ts
```diff
@@ -51,6 +51,12 @@ export interface ITextFieldProps extends React.AllHTMLAttributes<HTMLInputElemen
    */
   componentRef?: IRefObject<ITextField>;
 
+  /**
+   *  Optional callback to access the root DOM element.
+   *  @deprecated Temporary solution which will be replaced with ref in the V8 release.
```
ecraig12345, line 56: Especially for `@deprecated` tags it's important to only have one leading space in comments--otherwise the lint rule won't pick them up properly.
```suggestion
   * Optional callback to access the root DOM element.
   * @deprecated Temporary solution which will be replaced with ref once TextField is converted to a function component.
```

packages/react-next/src/components/TextField/TextField.types.ts
```diff
@@ -376,3 +351,40 @@ export interface ITextFieldStyles {
    */
   subComponentStyles: ITextFieldSubComponentStyles;
 }
+
+/**
+ * MaskedTextField component props.
+ * {@docCategory MaskedTextedField}
```
ecraig12345, line 357:
```suggestion
 * {@docCategory TextField}
```

packages/react-next/src/components/TextField/TextField.types.ts
```diff
@@ -376,3 +351,40 @@ export interface ITextFieldStyles {
...
+
+  /**
+   * Only used by MaskedTextField:
+   * The character to show in place of unfilled characters of the mask.
+   * @defaultvalue '_'
+   */
+  maskChar?: string;
+
+  /**
+   * Only used by MaskedTextField:
```
ecraig12345, line 381:
```suggestion
   * The masking string that defines the mask's behavior.
   * A backslash will escape any character.
   * Special format characters are:
   * '9': [0-9]
   * 'a': [a-zA-Z]
   * '*': [a-zA-Z0-9]
   *
   * @example `Phone Number: (999) 999-9999`
   */
  mask?: string;

  /**
   * The character to show in place of unfilled characters of the mask.
   * @defaultvalue '_'
   */
  maskChar?: string;

  /**
```

packages/react-next/src/components/TextField/MaskedTextField/MaskedTextField.tsx
```diff
@@ -15,385 +14,374 @@ import {
...
+    selectionStart: number;
+    selectionEnd: number;
+  } | null;
+}
 
-/**
- * State for the MaskedTextField component.
- */
-export interface IMaskedTextFieldState {
+interface IMaskedTextField extends ITextField {
```
ecraig12345, line 30: This should go in `TextField.types.ts` (and add a comment with `{@docCategory TextField}`). Also realized you'll need to add `componentRef?: IRefObject<IMaskedTextField>` to `IMaskedTextFieldProps`.

packages/react-next/src/components/TextField/MaskedTextField/MaskedTextField.tsx
```diff
@@ -15,385 +14,374 @@ import {
...
-   * This is what is displayed inside the TextField
-   * @example
-   *  `Phone Number: 12_ - 4___`
+   * The value of all filled format characters, or undefined if not all format characters are filled.
    */
-  displayValue: string;
-  /** The index into the rendered value of the first unfilled format character */
-  maskCursorPosition?: number;
+  value: string | undefined;
+  setValue: (newValue: string) => void;
```
ecraig12345, line 35: I think we actually should *not* expose this on the public interface (and only add it back if someone complains)--exposing a setter separate from props is generally not a good pattern for React components.

ecraig12345 (approved):

PR #14831, opened 2020-08-31
Removing unused state from MessageBar inside office-ui-fabric-react

czearing:
#### Pull request checklist

#### Description of changes
Removing unused showContent state value with OUFR's MessageBar.


ecraig12345 (changes requested):
Unless there's a compelling reason to make this change now, I think we probably should just wait for v8 when the state interface will go away entirely and this will no longer be an issue.

PR #14833, opened 2020-08-31
Correcting a typo in MaskedTextField's example

czearing:
#### Pull request checklist

#### Description of changes
Correcting a typo in Office-UI-Fabric-React's `MaskedTextField` example.

### DIFF
--- a/packages/office-ui-fabric-react/src/components/TextField/examples/TextField.Masked.Example.tsx
+++ b/packages/office-ui-fabric-react/src/components/TextField/examples/TextField.Masked.Example.tsx
@@ -10,7 +10,7 @@ export const TextFieldMaskedExample: React.FunctionComponent = () => {
   return (
     <Stack tokens={stackTokens}>
       <p>The mask has been modified here to allow "_"</p>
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/TextField/examples/TextField.Masked.Example.tsx
```diff
@@ -10,7 +10,7 @@ export const TextFieldMaskedExample: React.FunctionComponent = () => {
   return (
     <Stack tokens={stackTokens}>
       <p>The mask has been modified here to allow "_"</p>
-      <MaskedTextField label="With input mask" mask="m\ask: ****" maskFormat={maskFormat} maskChar="?" />
+      <MaskedTextField label="With input mask" mask="mask: ****" maskFormat={maskFormat} maskChar="?" />
```
ecraig12345, line 13: I think this actually was intentional--by default `a` is used as a placeholder for letters, and even though the maskFormat is overridden here, the `a` was probably escaped to make using the example for other cases easier.

PR #14835, opened 2020-09-01
fix(Detailslist): fix grouped detailslist grid semantics

### LINKED ISSUE #11243: Narrator table shortcut Ctr+Alt+Arrow not working on tables
### Environment Information

- **Package version(s):** 7.60.1
- **Browser and OS versions:** Microsoft Edge 44.18362.449.0

### Describe the issue:

Narrator's shortcut "Ctr+Alt+Arrow" keys to navigate through table elements only works for header, and is not working for items cells.

You can see this issue in the fabric DetailsList - Grouped page linked below:
https://developer.microsoft.com/en-us/fabric#/controls/web/detailslist/grouped

#### Actual behavior:

When narrator is used and "Ctr+Alt+Arrow" is used, nothing happens

#### Expected behavior:

When narrator is used and "Ctr+Alt+Arrow" is used, user is able to navigate through the table

smhigley:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #11243

#### Description of changes

- Updates GroupList to use proper `treegrid` semantics (which makes it behave like a tree to most screen readers)
- Updates child components of GroupList to allow enough customization for DetailsList to use them in a grid structure

API changes are to:
- `role` added to GroupSpacer, to allow a role override (sometimes in needs to be a gridcell)
- `role` added to GroupedList, to allow a role override (it makes sense for the default to be `treegrid`, but DetailsList needs it to be `presentation` since it's a child of the DetailsList `grid`)
- `ariaColSpan` added to GroupHeader and GroupFooter, since they need to span all the columns nested in their group for the table structure to hold up

There's still an issue where the aria-rowindex for each row in a Grouped DetailsList fails to take into account the GroupHeader rows, but I think that should have another separate issue. It will also require API changes, since DetailsRow will need some additional information like `groupIndex` in addition to `itemIndex`.


soettl:
```
There's still an issue where the aria-rowindex for each row in a Grouped DetailsList fails to take into account the GroupHeader rows, but I think that should have another separate issue. It will also require API changes, since DetailsRow will need some additional information like groupIndex in addition to itemIndex.
```
Can you please clarify what you expect these changes to be? What is `aria-rowindex` supposed to be? The index in the flattened tree?

smhigley:
@soettl I think there are two valid approaches for rowindex in a collapsible grid:
1. The index matches the position within the total number of rows, regardless of what is collapsed
2. The index matches the position within the currently visible rows, and would change when sections are expanded/collapsed

The main issue right now is that the expand/collapse group headers are completely left out of the row count, which is definitely confusing. Would it make sense to open a separate issue to fix the row indices? That would also be a good chance to fix the much smaller issue that all Detailslist rowindex values are off by one :D

ThomasMichon (approved):

cschlechty (approved):

dzearing:
Hey @smhigley can you make sure the build is passing?

```
 [8:15:30 PM] x src/components/DetailsList/DetailsList.base.tsx:41:76 - error TS6133: 'IGroupHeaderProps' is declared but its value is never read.
ERR! 
ERR! 41 import { IGroupedList, GroupedList, IGroupDividerProps, IGroupRenderProps, IGroupHeaderProps } from '../../GroupedList';
```

Give me a ping in chat when you have green lights, and we can get this merged.

PR #3796, opened 2018-01-24
Controlled TextFields

jordandrako:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1350 
- [x] Include a change request file using `$ npm run change`

#### Description of changes

Adds the _isControlled flag to the TextField component that disabled onChange if the value prop is set, which follows developer expectations with React inputs. 

#### Focus areas to test

This is a breaking change, developers using value instead of defaultValue will need to change to defaultValue if the field needs to be user-editable.

PR #14836, opened 2020-09-01
fix: centered the indeterminate indicator inside the checkbox component

khamudom:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #3796

#### Description of changes

Update to center the indicator by setting top and left to 50% and shifting it up and left by half of its own width and height, because the indicator is off by a pixel.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/web-components/src/checkbox/checkbox.styles.ts
+++ b/packages/web-components/src/checkbox/checkbox.styles.ts
@@ -68,10 +68,11 @@ export const CheckboxStyles = css`
         border-radius: calc(var(--corner-radius) * 1px);
         background: ${neutralForegroundRestBehavior.var};
         position: absolute;
-        top: 25%;
-        right: 25%;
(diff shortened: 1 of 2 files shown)


packages/web-components/src/checkbox/checkbox.styles.ts
```diff
@@ -68,10 +68,11 @@ export const CheckboxStyles = css`
...
         position: absolute;
-        top: 25%;
-        right: 25%;
-        bottom: 25%;
-        left: 25%;
+        top: 50%;
+        left: 50%;
+        width: 50%;
+        height: 50%;
+        transform: translate(-50%, -50%);
```
chrisdholt, line 75: Can you run `yarn change` to generate a change report so this can be published correctly? That will help the build pass as well :).

chrisdholt (approved):

PR #14839, opened 2020-09-01
Expose onRenderColorCell

bhavinkotak07:
#### Pull request checklist

#### Description of changes

(give an overview)
SwatchColorPicker doesn't have any onRender function. With this change we can customize the UI of the Color picker to show a color and font picker combo. Below is the screenshot which shows the proposed change.
image

#### Focus areas to test

(optional)


packages/office-ui-fabric-react/src/components/SwatchColorPicker/SwatchColorPicker.base.tsx
```diff
@@ -148,6 +148,11 @@ export class SwatchColorPickerBase extends React.Component<ISwatchColorPickerPro
     );
   }
 
+  private _onRenderItem = (item: IColorCellProps, index: number): JSX.Element => {
```
RupamSarkar, line 151: Do we need the `index` argument in this function?

packages/office-ui-fabric-react/src/components/SwatchColorPicker/SwatchColorPicker.types.ts
```diff
@@ -74,7 +74,10 @@ export interface ISwatchColorPickerProps {
    * If `id` and `color` are unspecified, cells are no longer being focused.
    */
   onCellFocused?: (id?: string, color?: string) => void;
-
+  /**
+   * Custom render function for the color cell
+   */
+  onRenderColorCell?: IRenderFunction<IColorCellProps>;
```
RupamSarkar, line 80: Do we have any UT for SwatchColorPicker ? If yes, then can we add one UT for with this new prop ?

packages/office-ui-fabric-react/src/utilities/ButtonGrid/ButtonGrid.types.ts
```diff
@@ -23,7 +23,7 @@ export interface IButtonGridProps extends React.TableHTMLAttributes<HTMLTableEle
   /**
    * Custom renderer for the individual items.
    */
-  onRenderItem: (item: any, index: number) => JSX.Element;
+  onRenderItem: (item: any, index: number) => JSX.Element | null;
```
shagra-ms, line 26: Can you please explain why change in return type is required?

bhavinkotak07: The onRenderItem function invokes onRenderColorCell which is of IRenderFunction type and it return JSX.Element | null. 
Not changing the return type will make compiler complain.

shagra-ms (approved):

packages/office-ui-fabric-react/src/components/SwatchColorPicker/SwatchColorPicker.types.ts
bhavinkotak07: Yes we have a UT for SwatchColorPicker and I have added a scenario for it.

ecraig12345 (commented):
Can you please make this change in the react-next version of SwatchColorPicker as well?

dzearing (approved):
While we would like to rework ColorSwatchPicker, this change is reasonable as a stop gap for reconfiguring the swatches.

dzearing (changes requested):
Needs to be updated in react-next as well.

shagra-ms:
Thanks for the inputs @dzearing and @ecraig12345. We will do the needful.

PR #14841, opened 2020-09-01
fix(Menu): remove style conflict between hover/active

assuncaocharles:
#### Pull request checklist

#### Description of changes

There is a requirement that first level menu should be opened by a click only. Tests and examples were updated to reflect this.
The another requirement is that we shouldn't show the active and hover indicator at same time. 

Screen Shot 2020-09-01 at 14 29 22

To solve this styles were updated removing background of active items where `on="hover"` is set

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/docs/src/examples/components/Menu/Usage/MenuExampleWithSubmenuHover.shorthand.tsx
+++ b/packages/fluentui/docs/src/examples/components/Menu/Usage/MenuExampleWithSubmenuHover.shorthand.tsx
@@ -5,7 +5,6 @@ const items: MenuProps['items'] = [
   {
     key: 'editorials',
     content: 'Editorials',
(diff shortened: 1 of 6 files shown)


packages/fluentui/docs/src/examples/components/Menu/Usage/MenuExampleWithSubmenuHover.shorthand.tsx
```diff
@@ -5,7 +5,6 @@ const items: MenuProps['items'] = [
   {
     key: 'editorials',
     content: 'Editorials',
-    on: 'hover',
```
assuncaocharles, line 8: Removed from the first level menu as it's the scenario we have for actual use

layershifter (approved):

PR #14842, opened 2020-09-01
Add preventDismissOnEvent callback in Callout

### LINKED ISSUE #14527: [Callout]  Decide whether to prevent dismiss on lost focus dynamically

Update `preventDismissOnLostFocus` in `Callout` to be a callback instead of a boolean. This would enable deciding per event whether we want to dismiss or not.

`Callout` (in `ICalloutProps` in `Callout.types`)

No

We are working on a functionality where we need to show a context menu on right click on a gallery. The gallery is built over `ContextualMenu` (which I understand is built over `Callout`). Since the gallery is configured to dismiss on losing focus, the moment we display the context menu, the gallery dismisses. The prop `preventDismissOnLostFocus` can be set to prevent dismissing it, but then it would not dismiss on losing focus in any case. What we need is sort of a middle ground where we can choose whether or not to dismiss depending on the event. (Like in this case, we want it to dismiss if user clicks outside the gallery, but not if they right click on the gallery and the focus shifts to context menu).
One possible suggestion that came from within our team was if `preventDismissOnLostFocus` is updated to be a callback instead of a boolean. Then looking at the event in the callback, we could decide whether or not to dismiss the gallery.

lamha-lagoe:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14527

#### Description of changes

Added a prop `preventDismissOnEvent` in `Callout`, and in favor of this deprecated the 3 existing props `preventDismissOnScroll`, `preventDismissOnResize`, and `preventDismissOnLostFocus` in `Callout` Component


packages/office-ui-fabric-react/src/components/Callout/Callout.types.ts
```diff
@@ -85,21 +85,32 @@ export interface ICalloutProps extends React.HTMLAttributes<HTMLDivElement> {
...
    */
   preventDismissOnLostFocus?: boolean;
 
+  /**
+   *If defined, then takes priority over preventDismissOnLostFocus, preventDismissOnResize,
+   * and preventDismissOnScroll.
+   * If it returns true, then callout will not dismiss for this event.
+   * If not defined or returns false, callout can dismiss for this event.
+   */
+  preventDismissOnEvent?: (ev: any) => boolean;
```
joschect, line 112: Instead of any, I think we might want to more strongly type this. All of those handlers seem to use Event, I might suggest typing it like `<EventType extends Event>(ev: EventType) => boolean `

lamha-lagoe: Sorry if this sounds a bit obvious, but will that also cover things like `React.MouseEvent` and others (`React.KeyboardEvent`, `React.FocusEvent`)? Tracing the hierarchy, I could see interfaces like `React.MouseEvent` extending `SyntheticEvent` which extends `BaseSyntheticEvent<Event,..>`, but couldn't find anything of the sort `extends Event`.

joschect (approved):

PR #14844, opened 2020-09-01
docs(SvgIcon): remove page in production

assuncaocharles:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/docs/src/components/Sidebar/Sidebar.tsx
+++ b/packages/fluentui/docs/src/components/Sidebar/Sidebar.tsx
@@ -31,7 +31,9 @@ const pkg = require('@fluentui/react-northstar/package.json');
 const componentMenu: ComponentMenuItem[] = _.sortBy(componentInfoContext.parents, 'displayName');
 const behaviorMenu: ComponentMenuItem[] = require('../../behaviorMenu');
 
-const componentsBlackList = ['Debug', 'Design', 'Datepicker'];
+const PROD = process.env.NODE_ENV === 'production';
+
+const componentsBlackList = ['Debug', 'Design', 'Datepicker', PROD && 'SvgIcon'];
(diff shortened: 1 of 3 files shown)


packages/fluentui/docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -31,7 +31,9 @@ const pkg = require('@fluentui/react-northstar/package.json');
 const componentMenu: ComponentMenuItem[] = _.sortBy(componentInfoContext.parents, 'displayName');
 const behaviorMenu: ComponentMenuItem[] = require('../../behaviorMenu');
 
-const componentsBlackList = ['Debug', 'Design', 'Datepicker'];
+const PROD = process.env.NODE_ENV === 'production';
```
layershifter, line 34: There is no sense to embed this condition. I am usually using search by `process.env.NODE_ENV` if there is something conditional

assuncaocharles: We were reusing this condition in 2 places in this file, I would be introducing the 3rd. Should it always be "in place"?

packages/fluentui/docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -228,7 +228,7 @@ const baseTreeItems: TreeProps['items'] = [
           to: '/component-architecture',
         },
       },
-      ...(process.env.NODE_ENV !== 'production'
+      ...(process.env.NODE_ENV === 'production'
```
mnajdova, line 231: Why did we change this condition??

mnajdova (approved):
LGTM thanks for doing this

layershifter (approved):

PR #14851, opened 2020-09-01
DetailsRow/DetailsHeader: Remove findDOMNode usage

### LINKED ISSUE #14685: Add API to get root element of FocusZone (and fix DetailsRow/Header)
Motivation: DetailsRow and DetailsHeader use `ReactDOM.findDOMNode` to find the actual HTML element corresponding to a FocusZone. This works for the time being, since FocusZone is still a class component, but it will stop working if/when FocusZone is converted to a function component. 

To fix this, we need to add a new API to FocusZone to get the root element. @MLoughry had previously made a PR #11945 adding a `domRef` prop to all components for this purpose, so one option would be to add a similar API but only to FocusZone.

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14685

#### Description of changes

Remove `findDOMNode` usage from DetailsRow and DetailsHeader by using the new `IFocusZoneProps.elementRef` @czearing added.

### DIFF
--- a/change/@fluentui-react-focus-2020-09-02-15-55-10-list-finddomnode.json
+++ b/change/@fluentui-react-focus-2020-09-02-15-55-10-list-finddomnode.json
@@ -1,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "Fix typo in docs  ",
+  "packageName": "@fluentui/react-focus",
+  "email": "elcraig@microsoft.com",
+  "dependentChangeType": "none",
+  "date": "2020-09-02T22:55:02.427Z"
+}
--- a/change/@fluentui-react-focus-2020-09-02-15-55-10-list-finddomnode.json
+++ b/change/@fluentui-react-focus-2020-09-02-15-55-10-list-finddomnode.json
@@ -1,6 +1,6 @@
 {
   "type": "none",
-  "comment": "Fix typo in docs  ",
+  "comment": "Fix typo in docs",
   "packageName": "@fluentui/react-focus",
   "email": "elcraig@microsoft.com",
   "dependentChangeType": "none",
--- a/change/@fluentui-react-focus-2020-09-02-15-55-10-list-finddomnode.json
+++ b/change/@fluentui-react-focus-2020-09-02-15-55-10-list-finddomnode.json
@@ -1,5 +1,5 @@
 {
-  "type": "none",
+  "type": "patch",
   "comment": "Fix typo in docs",
   "packageName": "@fluentui/react-focus",
   "email": "elcraig@microsoft.com",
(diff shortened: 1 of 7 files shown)


change/@fluentui-react-focus-2020-09-02-15-55-10-list-finddomnode.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "Fix typo in docs  ",
```
dzearing, line 3:
```suggestion
  "comment": "Fix typo in docs",
```

change/@uifabric-utilities-2020-09-02-15-55-10-list-finddomnode.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
```
dzearing, line 2: Shouldn't this be a patch, so that the correct typings are published?

change/@fluentui-react-focus-2020-09-02-15-55-10-list-finddomnode.json
ecraig12345: I did that so that beachball change wouldn't force me to use the commit message from a previous PR 😬  (though obviously it can be manually edited later)

packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.base.tsx
```diff
@@ -102,25 +101,28 @@ export class DetailsHeaderBase extends React.Component<IDetailsHeaderBaseProps,
...
+      }
     }
   }
 
   public componentDidUpdate(prevProps: IDetailsHeaderBaseProps): void {
     if (this._getColumnReorderProps()) {
       if (!this._subscriptionObject) {
         this._subscriptionObject = this._dragDropHelper.subscribe(
-          this._rootElement!,
+          this._rootElement.current!,
```
khmakoto, line 125:
```suggestion
          this._rootElement.current,
```

khmakoto (approved):

packages/office-ui-fabric-react/src/components/DetailsList/DetailsHeader.base.tsx
ecraig12345: Unfortunately the ! is needed here since `current` isn't guaranteed by the typings to be defined.

ecraig12345: Though I suppose the proper way to fix that is wrapping in if()

dzearing (approved):

Issue #14494, opened 2020-08-12
Enable new conformance tests everywhere


ecraig12345:
Now that Esteban has added the react-conformance package, we should enable it for all components and fix any issues it reveals (and maybe add more tests).

If there's a failure when enabling the tests on a new component, the options are:
- Fix the issue in the component, if it's a real issue which is easy to fix and doesn't require API changes (API changes may be okay in /next components)
- Fix the test/add more options if the test is giving incorrect results
- Disable the test if it's irrelevant or not feasible to fix

### In v7/v8

- [x] In `office-ui-fabric-react`:
  - [x] Rename the `sharedIsConformant` file and function to just `isConformant`
  - [x] Call from components. There are a couple possible ways we could do this:
    - Call from `packages/office-ui-fabric-react/src/components/ComponentConformance.test.tsx`
      - Downside: more special logic required for per-component overrides, and determining which components to include
    - Call manually from individual component tests
      - Downside: easier to forget, but if it's included in all the tests it's likely to propagate into new ones since they tend to be copy-pasted
      - If we do this, we could skip the tests in components that have been copied to react-next
- [x] In `react-next`:
  - [x] Rename the `sharedIsConformant` file and function to just `isConformant`
  - [x] Call from components. Options are similar to OUFR.
- [x] In new `react-<component>` packages, call manually from each component's tests (current and /next or /compat) for now
    - [x] react-button (all variants)
    - [x] react-checkbox
    - [x] react-link
    - [x] react-toggle
    - [x] react-slider
    - [x] react-tabs

### In `react-northstar`

The main thing which would be beneficial to do right away is combining `sharedIsConformant` (`packages/fluentui/react-northstar/test/utils/sharedIsConformant.tsx`) with the main `isConformant` (`packages/fluentui/react-northstar/test/specs/commonTests/isConformant.tsx`). This is not directly relevant to v8, but it should be fairly quick and helps move us a little closer to full convergence.
- [x] In `interface Conformant`, add extra props that will be needed to run the shared tests:
  - Update the declaration to `interface Conformant<TProps = {}> extends Pick<IsConformantOptions<TProps>, 'disabledTests' | 'testOptions'>` (can add additional things to the `Pick` later as needed)
  - `testPath: string` - each test's `__filename` (required)
- [x] In northstar `isConformant.tsx`, add a call to `react-conformance`'s version of `isConformant` (with appropriate defaults and prop translations copied from `sharedIsConformant` file)
[...]

PR #14853, opened 2020-09-01
Consolidating react-northstar conformance tests and renaming sharedIsConformant files

czearing:
#### Pull request checklist
- [x] Related to #14494

#### Description of changes
1. Renamed `sharedIsConformant` file and function to `isConformant` in `react-next`, `office-ui-fabric-react`, and `react-slider`.
2. Removed `sharedIsConformant` from react-northstar/test/utils and combined it's logic with react-northstar/test/specs/commonTests/isConformant.tsx.
3. Removed repeated conformant test in react-northstar.
4. Added a testPath type to Conformant interface in react-northstar.
5. Added new testPath to all of react-northstar's Conformant tests.
6. Added disabledTests property:
`Carousel`: 'kebab-aria-attributes'
`DropdownSearchInput`: 'as-renders-fc', 'as-passes-as-value', 'as-renders-html', 'as-renders-react-class'
`Animation`: 'as-renders-fc', 'as-passes-as-value', 'as-renders-html', 'as-renders-react-class'


packages/fluentui/react-northstar/test/specs/commonTests/isConformant.tsx
```diff
@@ -56,21 +59,22 @@ export function isConformant(
...
   const {
+    testPath,
     constructorName = Component.prototype.constructor.name,
     eventTargets = {},
     exportedAtTopLevel = true,
     hasAccessibilityProp = true,
     requiredProps = {},
     rendersPortal = false,
     wrapperComponent = null,
-    handlesAsProp = true,
```
ecraig12345, line 69: It might be better to bring back `handlesAsProp`, then within this file add the "as" tests to `disabledTests` if `handlesAsProp` is false.

packages/fluentui/react-northstar/test/specs/components/Dropdown/DropdownSearchInput-test.tsx
```diff
@@ -3,9 +3,11 @@ import { isConformant } from 'test/specs/commonTests';
 
 describe('DropdownSearchInput', () => {
   isConformant(DropdownSearchInput, {
+    testPath: __filename,
     constructorName: 'DropdownSearchInput',
     hasAccessibilityProp: false,
     handlesAsProp: false,
+    disabledTests: ['as-renders-fc', 'as-passes-as-value', 'as-renders-html', 'as-renders-react-class'],
```
ecraig12345, line 10: This part can be reverted if you centralize `handlesAsProp` handling as suggested above

packages/office-ui-fabric-react/src/common/isConformant.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line import/no-extraneous-dependencies -- this file is for testing
```
ecraig12345, line 1: Probably better to remove these individual disables and centrally handle this in `packages/eslint-plugin/src/utils/configHelpers.js`: in the `testFiles` array add a line `'**/common/isConformant.{ts,tsx}'`

packages/fluentui/react-northstar/test/specs/commonTests/isConformant.tsx
czearing: Open for discussion, but I'm not sure if this would be a change in the right direction. handleAsProp is only referenced once which is in the DropdownSearchInput test file. Since the previous "as" tests have been replaced with the new react-conformance tests, It might make more sense just to use the disabledTests prop and remove handleAsProp.

packages/fluentui/react-northstar/test/specs/commonTests/isConformant.tsx
```diff
@@ -14,17 +15,19 @@ import { act } from 'react-dom/test-utils';
 import { isExportedAtTopLevel } from './isExportedAtTopLevel';
```
ecraig12345, line 15: Just realized, I think you can also remove the call to this helper (and delete its file)

packages/fluentui/react-northstar/test/specs/commonTests/isConformant.tsx
layershifter: Makes sense to use `disabledTests` for me

layershifter (approved):

ecraig12345 (approved):

PR #14854, opened 2020-09-01
Renaming special-purpose Grid component to ButtonGrid within react-next

### LINKED ISSUE #13124: Rename special-purpose Grid component to ButtonGrid

czearing:
#### Pull request checklist
- [x] Fixes #13124

#### Description of changes
Renaming Grid utility within react-next to `ButtonGrid`.


ecraig12345 (changes requested):
I think you're missing updates to `react-next/src/Grid.ts`

change/office-ui-fabric-react-2020-09-02-18-45-05-feat-renaming-gridNext.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
ecraig12345, line 2: This can be `none` since there's no impact on the published package.
```suggestion
  "type": "none",
```

packages/office-ui-fabric-react/src/components/Calendar/CalendarDay.tsx
```diff
@@ -838,7 +838,7 @@ export class CalendarDay extends React.Component<ICalendarDayProps, ICalendarDay
       isAllDaysOfWeekOutOfMonth = true;
 
       for (let dayIndex = 0; dayIndex < DAYS_IN_WEEK; dayIndex++) {
-        const originalDate = new Date(date);
+        const originalDate = new Date(date as any);
```
ecraig12345, line 841: Maybe add a comment about why this was needed

packages/office-ui-fabric-react/src/components/Pivot/Pivot.base.tsx
```diff
@@ -353,6 +353,6 @@ function _isPivotItem(item: React.ReactNode): item is PivotItem {
     !!item &&
     typeof item === 'object' &&
     !!(item as React.ReactElement).type &&
-    ((item as React.ReactElement).type as React.ComponentType).name === PivotItem.name
+    ((item as React.ReactElement).type as any).name === (PivotItem as any).name
```
ecraig12345, line 356: same here

ecraig12345 (approved):
Approved with a couple minor suggestions.

PR #14860, opened 2020-09-02
feat(TextArea): Add autoAdjustHeight

assuncaocharles:
#### Pull request checklist

#### Description of changes

This PR adds the ability for `TextArea` to auto adjust the height based in the content

cAUtPGDi1P

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/TextArea/TextArea.tsx
```diff
@@ -44,9 +45,16 @@ export interface TextAreaProps extends UIComponentProps, ChildrenComponentProps
 
   /** A textarea can take the width of its container. */
   fluid?: boolean;
+
+  /** A textarea can be set to automatically adjust the height */
+  autoAdjustHeight?: boolean;
```
assuncaocharles, line 50: I called it `autoAdjustHeight` to be similar to Fabric and it was also as the user named it when requiring it.... But I wonder if `autoResize` wouldn't be a better name

packages/fluentui/react-northstar/src/themes/teams/components/TextArea/textAreaStyles.ts
```diff
@@ -34,6 +35,13 @@ export const textAreaStyles: ComponentSlotStylesPrepared<TextAreaStylesProps, Te
       color: v.disabledColor,
     }),
 
+    ...(p.autoAdjustHeight && {
+      minHeight: v.minHeightAutoAdjust,
+      height: pxToRem(p.autoHeight),
```
pompomon, line 40: Please could you move this `height` to inline style? If I understand correctly - each change in `autoHeight` here would trigger styles change in the document head, which would be more costly performance-wise, than change of the inline style

packages/fluentui/react-northstar/src/themes/teams/components/TextArea/textAreaStyles.ts
```diff
@@ -34,6 +35,13 @@ export const textAreaStyles: ComponentSlotStylesPrepared<TextAreaStylesProps, Te
...
+      overflow: 'hidden',
+      resize: 'vertical',
```
pompomon, line 42: Please would that mean that users could resize the block by themselves? Is that desirable?

packages/fluentui/react-northstar/src/themes/teams/components/TextArea/textAreaStyles.ts
assuncaocharles: done

packages/fluentui/react-northstar/src/components/TextArea/TextArea.tsx
```diff
@@ -106,21 +126,29 @@ export const TextArea: ComponentWithAs<'textarea', TextAreaProps> &
...
   const handleChange = (e: React.ChangeEvent | React.FormEvent) => {
     const newValue = _.get(e, 'target.value');
-
     _.invoke(props, 'onChange', e, { ...props, value: newValue });
     setValue(newValue);
   };
 
+  React.useLayoutEffect(() => {
+    if (autoAdjustHeight) {
+      inputRef.current.style.height = '0';
```
pompomon, line 135:
```suggestion
      inputRef.current?.style.height = '0';
```
?

pompomon (approved):

packages/fluentui/react-northstar/src/components/TextArea/TextArea.tsx
mnajdova: I would vote for `autoResize` it's what I would search for if I need behavior like this.

packages/fluentui/react-northstar/src/components/TextArea/TextArea.tsx
```diff
@@ -106,21 +115,29 @@ export const TextArea: ComponentWithAs<'textarea', TextAreaProps> &
...
     _.invoke(props, 'onChange', e, { ...props, value: newValue });
     setValue(newValue);
   };
 
+  React.useLayoutEffect(() => {
+    if (autoResize) {
+      inputRef.current.style.height = '0';
+      inputRef.current.style.height = `${inputRef.current?.scrollHeight}px`;
+    }
+  }, [inputRef.current?.scrollHeight, autoResize]);
```
layershifter, line 127: During the first render `inputRef.current.scrollHeight` will be null. In general it's incorrect to use such things as effect dependencies

packages/fluentui/CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added icons `FilterIcon`, `MicrophoneDisabledIcon`, `OptionsIcon`, `ReactionsIcon` @TanelVari ([#14745](https://github.com/microsoft/fluentui/pull/14745))
 - Add `foreground5` and `foreground6` in the colorScheme's default definition @assuncaocharles ([#14818](https://github.com/microsoft/fluentui/pull/14818))
 - Add `Form.Datepicker` component @pompomon ([#14820](https://github.com/microsoft/fluentui/pull/14820))
+- Add `autoAdjustHeight` to `TextArea` component @assuncaocharles ([#14860](https://github.com/microsoft/fluentui/pull/14860))
```
layershifter, line 49: PR title and changelog entry have wrong prop name

layershifter:
https://codesandbox.io/s/microsoftfluentui-codesandbox-react-northstar-template-forked-czevk?file=/src/index.tsx

### `border` & `minHeight`

The combination of `border` and `minHeight` breaks the logic:
textarea-bug

---

You can also go through linked issues in https://github.com/Semantic-Org/Semantic-UI-React/pull/3465, there are numerous interesting issues 🔫 I suggest to revert this an use [`react-textarea-autosize`](https://www.npmjs.com/package/react-textarea-autosize) instead.

PR #14862, opened 2020-09-02
feat(Datepicker): Add advanced keyboard navigation

pompomon:
#### Pull request checklist

#### Description of changes

Following key shortcuts are supported:
Home – First cell of current row
End – Last cell of current row
Ctrl + Home – First cell of the grid (first row first cell)
Ctrl + End – Last cell of the grid (last row last cell)
Page up – First cell of current column (basically scroll up/down behaviour but in this case there’s no scrollbar)
Page down – Last cell of current column

### DIFF
--- a/packages/date-time-utilities/src/dateMath/dateMath.ts
+++ b/packages/date-time-utilities/src/dateMath/dateMath.ts
@@ -304,6 +304,22 @@ export function getStartDateOfWeek(date: Date, firstDayOfWeek: DayOfWeek): Date 
 }
 
 /**
+ * Gets the date for the last day of the week based on the given date assuming
+ * the specified first day of the week.
+ * @param date - The date to find the beginning of the week date for.
+ * @returns A new date object representing the first day of the week containing the input date.
+ */
+export function getEndDateOfWeek(date: Date, firstDayOfWeek: DayOfWeek): Date {
+  const lastDayOfWeek = firstDayOfWeek - 1 >= 0 ? firstDayOfWeek - 1 : TimeConstants.DaysInOneWeek - 1;
+  let daysOffset = lastDayOfWeek - date.getDay();
+  if (daysOffset < 0) {
+    // If last day of week is < date, go 1 week forward, to ensure resulting date is in the future.
+    daysOffset += TimeConstants.DaysInOneWeek;
+  }
+  return addDays(date, daysOffset);
+}
+
+/**
  * Gets a new date with the time portion zeroed out, i.e., set to midnight
  * @param date - The origin date
  * @returns A new date with the time set to midnight
--- a/packages/fluentui/docs/src/examples/components/Datepicker/Types/DatepickerExample.shorthand.tsx
+++ b/packages/fluentui/docs/src/examples/components/Datepicker/Types/DatepickerExample.shorthand.tsx
@@ -1,6 +1,6 @@
-import { Datepicker } from '@fluentui/react-northstar';
+import { Datepicker, DayOfWeek } from '@fluentui/react-northstar';
 import * as React from 'react';
 
-const DatepickerExample = () => <Datepicker />;
+const DatepickerExample = () => <Datepicker firstDayOfWeek={DayOfWeek.Tuesday} />;
 
 export default DatepickerExample;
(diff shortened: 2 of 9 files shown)


packages/date-time-utilities/src/dateMath/dateMath.ts
```diff
@@ -303,6 +303,22 @@ export function getStartDateOfWeek(date: Date, firstDayOfWeek: DayOfWeek): Date
...
 }
 
+/**
+ * Gets the date for the last day of the week based on the given date assuming
+ * the specified first day of the week.
+ * @param date - The date to find the beginning of the week date for.
+ * @returns A new date object representing the first day of the week containing the input date.
+ */
+export function getEndDateOfWeek(date: Date, firstDayOfWeek: DayOfWeek): Date {
+  const lastDayOfWeek = firstDayOfWeek - 1 >= 0 ? firstDayOfWeek - 1 : TimeConstants.DaysInOneWeek - 1;
```
rymeskar, line 313: Curious:
wouldn't modulo operation be more compact?

rymeskar: Also applies to calculation below.

packages/fluentui/docs/src/examples/components/Datepicker/Types/DatepickerExample.shorthand.tsx
```diff
@@ -1,6 +1,6 @@
-import { Datepicker } from '@fluentui/react-northstar';
+import { Datepicker, DayOfWeek } from '@fluentui/react-northstar';
 import * as React from 'react';
 
-const DatepickerExample = () => <Datepicker />;
+const DatepickerExample = () => <Datepicker firstDayOfWeek={DayOfWeek.Tuesday} />;
```
rymeskar, line 4: Do we want to change the basic example? We already have a special firstDayOfWeek example.

packages/fluentui/react-northstar/src/components/Datepicker/DatepickerCalendar.tsx
```diff
@@ -147,6 +149,53 @@ export const DatepickerCalendar: ComponentWithAs<'div', DatepickerCalendarProps>
...
+        const newNavigatedDate = contstraintNavigatedDate(gridNavigatedDate, targetDate, -1, restrictedDatesOptions);
+
+        if (!!newNavigatedDate) {
+          setShouldFocusInDayGrid(true);
+          setGridNavigatedDate(newNavigatedDate);
+        }
+      },
+      moveToEndOfWeek: e => {
+        e.preventDefault();
+        const targetDate = getEndDateOfWeek(gridNavigatedDate, firstDayOfWeek);
```
rymeskar, line 169: Does it make sense to add the end/start of week navigation to the `NavigationKind` and then use the `navigateToNewDate` directly?

packages/fluentui/docs/src/examples/components/Datepicker/Types/DatepickerExample.shorthand.tsx
pompomon: No, that's left from my debugging, good catch :)

packages/date-time-utilities/src/dateMath/dateMath.ts
pompomon: Yeah, makes sense :)

pompomon: Aha, Javascript modulo would return negative numbers, so we've agreed with @rymeskar to keep it like it is now :)

packages/fluentui/react-northstar/src/components/Datepicker/DatepickerCalendar.tsx
pompomon: Yeah, but I'll still need to export `contstraintNavigatedDate` for column navigation, since it would be based on the visible grid, and I don't want to overcomplicate `navigateToNewDate` with additional `firstDayOfWeek` param, so I'd rather keep it like that, WDYT?

layershifter (approved):

PR #14864, opened 2020-09-02
fix(dropdown) avoid switch into screen reader virtual cursor mode when navigate through selected items

kolaps33:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/fluentui/accessibility/src/behaviors/Dropdown/dropdownSelectedItemBehavior.ts
```diff
@@ -10,11 +10,11 @@ import { Accessibility } from '../../types';
 export const dropdownSelectedItemBehavior: Accessibility<DropdownSelectedItemBehaviorProps> = props => ({
   attributes: {
     root: {
-      'aria-selected': true,
```
assuncaocharles, line 13: Should we really remove this attr?

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -766,6 +770,11 @@ export const Dropdown: ComponentWithAs<'div', DropdownProps> &
         render: renderSelectedItem,
       }),
     );
+    return (
+      <div role="listbox" aria-label={a11ySelectedItemsListboxMessage && a11ySelectedItemsListboxMessage}>
```
assuncaocharles, line 774: I thing we should use `Box` to create this container rather the element directly and possible pass the `aria-label` to behavior.

layershifter:
```suggestion
      <div role="listbox" aria-label={a11ySelectedItemsListboxMessage}>
```

@assuncaocharles there is no sense to have `Box` there as it's not a real slot.

---

@kolaps33 As I see other props related to A11y are functions, should it be `getA11ySelectedItemsListboxMessage()`?

packages/fluentui/accessibility/src/behaviors/Dropdown/dropdownSelectedItemBehavior.ts
```diff
@@ -10,11 +10,11 @@ import { Accessibility } from '../../types';
...
+      role: 'option',
       tabIndex: props.active ? 0 : -1,
     },
     icon: {
-      'aria-label': `Remove ${props.header} from selection.`,
+      'aria-hidden': 'true',
```
layershifter, line 17: image

Are we hiding this element from narration?

packages/fluentui/accessibility/src/behaviors/Dropdown/dropdownSelectedItemBehavior.ts
kolaps33: I think yes we should remove. Now it has role option and adding there 'aria-selected' doesn't make sense, because we are not really doing selection on this item. Only we focus the item and user can delete it. 
Even the role listbox/option is not the ideal, but we cannot find any better solution, how to switch off screen reader mode for JAWS and NVDA.

packages/fluentui/accessibility/src/behaviors/Dropdown/dropdownSelectedItemBehavior.ts
kolaps33: Yes hiding from narration by purpose, I don't want to narrate this icon at all, because user need to press Delete or Backspace key to remove selected item. And this instruction message I set to the element with role="listbox"

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
kolaps33: @layershifter I was looking to it and I think it will be enough if it is simple "string" prop, because others a11y need be function but this one is only simple one which will add "string"  as "aria-label".
And I renamed it into:
a11ySelectedItemsMessage
as I think `listbox` was not useful in the name.

packages/fluentui/docs/src/examples/components/Dropdown/Types/DropdownExampleSearchMultiple.shorthand.tsx
```diff
@@ -21,6 +21,7 @@ const DropdownExampleSearchMultiple = () => (
     placeholder="Start typing a name"
     getA11ySelectionMessage={getA11ySelectionMessage}
     noResultsMessage="We couldn't find any matches."
+    a11ySelectedItemsMessage="delete user by press Delete key or Backspace"
```
jurokapsiar, line 24:
```suggestion
    a11ySelectedItemsMessage="Press Delete or Backspace to remove the user from selection"
```

jurokapsiar (approved):

PR #14871, opened 2020-09-03
Implementing v8 conformance tests

czearing:
#### Pull request checklist
- [x] Related to #14494

#### Description of changes
1. Adding conformance tests to V8 components.

2. Renamed `sharedIsConformant` to `isConformant`.

4. Adding `isConformant` files to eslint configHelpers `testFiles` array.

3. Adding `isConformant` as divDependencies to existing V8 packages.

4. Adding `isConformant` file to `react-link`, `react-tabs`, `react-toggle`, `react-button`, and `react-checkbox`.

5. Adding `displayNames` to react-next components.

6. Fix to a visual regression in `ChoiceGroupOption`.

For reference here is a HackMD with a list of the conformance tests that will run by default: https://hackmd.io/4i2Xl0d8TiuZLcBAPznnnA

### V8 conformant tests added and disabledTests: 
- ### `react-next`: 'has-docblock', 'kebab-aria-attributes'
    * Breadcrumb
    * Callout
    * ChoiceGroup
    * ComboBox: `component-renders`
    * ContextualMenu
    * Dropdown
    * FocusTrapZone
    * Image
    * MessageBar
    * Modal
    * OverflowSet
    * Persona
    * PersonaCoin
    * PersonaPresence: `exported-top-level, has-top-level-file`
    * Rating
    * ResizeGroup
    * SearchBox
    * BaseSelectedItemsList: `has-top-level-file`
    * SelectedPeopleList: `has-top-level-file`
    * Shimmer
    * SpinButton
    * SwatchColorPicker
    * TeachingBubble
    * MaskedTextField: `has-top-level-file`
    * TextField
    * Grid: ` has-top-level-file`
- ### `react-button`: 'has-docblock', 'kebab-aria-attributes'
    * Button
    * MenuButton
    * SplitButton
    * ToggleButton:`
    
- ### `react-slider`: 'has-docblock', 'kebab-aria-attributes'
    * Slider
- ### `react-toggle`: 'has-docblock', 'kebab-aria-attributes'
    * Toggle
- ### `react-tabs`: 'has-docblock', 'kebab-aria-attributes'
    * Pivot
- ### `react-link`: 'has-docblock', 'kebab-aria-attributes'
    * Link
- ### `react-checkbox`: 'has-docblock', 'kebab-aria-attributes'
    * Checkbox

## V8 conformant tests that need to be addressed in the future:
- ### `react-next`: 'has-docblock', 'kebab-aria-attributes'
    * Layer
    * ContextualMenuButton
    * ContextualMenuSplitButton


packages/react-next/src/components/Callout/Callout.test.tsx
```diff
@@ -47,6 +48,13 @@ describe('Callout', () => {
     });
   });
 
+  isConformant({
+    componentPath: path.join(__dirname, 'Callout.tsx'),
+    Component: Callout,
+    displayName: 'Callout',
+    disabledTests: ['component-has-displayname'],
```
ecraig12345, line 55: This should definitely have a displayName or constructor name. If it doesn't today we should fix it.

Same in other component tests.

packages/react-toggle/src/common/isConformant.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line import/no-extraneous-dependencies -- this file is for testing
```
ecraig12345, line 1: Remove this

packages/react-next/src/components/Persona/PersonaPresence/PersonaPresence.test.tsx
```diff
@@ -93,4 +95,11 @@ describe('PersonaPresence', () => {
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  isConformant({
+    componentPath: path.join(__dirname, 'PersonaPresence.tsx'),
+    Component: PersonaPresence,
+    displayName: 'PersonaPresence',
+    disabledTests: ['exported-top-level', 'has-top-level-file'],
```
ecraig12345, line 103: This one actually is exported top-level (despite not having a top-level file), so the test shouldn't be disabled. If the test isn't handling that scenario properly we should fix it.

packages/react-next/src/components/SelectedItemsList/BaseSelectedItemsList.test.tsx
```diff
@@ -28,6 +30,13 @@ describe('SelectedItemsList', () => {
       expect(tree).toMatchSnapshot();
     });
 
+    isConformant({
+      componentPath: path.join(__dirname, 'BaseSelectedItemsList.tsx'),
+      Component: BaseSelectedItemsList,
+      displayName: 'BaseSelectedItemsList',
+      disabledTests: [`exported-top-level`, `has-top-level-file`],
```
ecraig12345, line 37: This one also shouldn't have `exported-top-level` disabled

packages/react-next/src/components/SelectedItemsList/SelectedPeopleList/SelectedPeopleList.test.tsx
```diff
@@ -1,11 +1,20 @@
...
 
 import { SelectedPeopleList, IExtendedPersonaProps } from './SelectedPeopleList';
 
 describe('SelectedPeopleList', () => {
   describe('Element keying behavior', () => {
+    isConformant({
+      componentPath: path.join(__dirname, 'SelectedPeopleList.tsx'),
+      Component: SelectedPeopleList,
+      displayName: 'SelectedPeopleList',
+      disabledTests: [`exported-top-level`, `has-top-level-file`],
```
ecraig12345, line 15: This one also shouldn't have `exported-top-level` disabled

ecraig12345 (commented):
Mostly looks good! A few suggestions.

packages/react-button/src/common/isConformant.ts
```diff
@@ -0,0 +1,11 @@
+import { isConformant as baseIsConformant, IsConformantOptions } from '@fluentui/react-conformance';
+
+export function isConformant(testInfo: Omit<IsConformantOptions, 'componentPath'>) {
+  const defaultOptions = {
+    disabledTests: ['has-docblock', 'kebab-aria-attributes', 'as-*'],
+  };
+
+  const path = { componentPath: module!.parent!.filename.replace('.test', '') };
```
ecraig12345, line 8: It might be better to just include this in defaultOptions (same in other packages)

packages/react-button/src/components/SplitButton/SplitButton.test.tsx
```diff
@@ -0,0 +1,10 @@
+import { SplitButton } from './SplitButton';
+import { isConformant } from '../../common/isConformant';
+
+describe('SplitButton', () => {
+  isConformant({
+    Component: SplitButton,
+    displayName: 'SplitButton',
+    disabledTests: ['has-docblock', 'as-renders-html', 'as-passes-as-value', 'as-renders-react-class', 'as-renders-fc'],
```
ecraig12345, line 8: Remove these?

packages/react-checkbox/src/common/isConformant.ts
```diff
@@ -0,0 +1,11 @@
+import { isConformant as baseIsConformant, IsConformantOptions } from '@fluentui/react-conformance';
+
+export function isConformant(testInfo: Omit<IsConformantOptions, 'componentPath'>) {
```
ecraig12345, line 3: Can you add a call to this in the checkbox tests?

packages/react-link/src/common/isConformant.ts
```diff
@@ -0,0 +1,11 @@
+import { isConformant as baseIsConformant, IsConformantOptions } from '@fluentui/react-conformance';
+
+export function isConformant(testInfo: Omit<IsConformantOptions, 'componentPath'>) {
```
ecraig12345, line 3: Can you call this in the link tests? (current and /next)

packages/react-next/src/components/ChoiceGroup/ChoiceGroup.base.tsx
```diff
@@ -156,3 +145,17 @@ export const ChoiceGroupBase: React.FunctionComponent = React.forwardRef(
     );
   },
 );
+ChoiceGroupBase.displayName = COMPONENT_NAME;
+
+function useDebugWarnings(props: IChoiceGroupProps) {
+  if (process.env.NODE_ENV !== 'production') {
+    // eslint-disable-next-line react-hooks/rules-of-hooks -- build-time conditional
+    useWarnings({
+      name: 'Checkbox',
```
ecraig12345, line 154:
```suggestion
      name: COMPONENT_NAME.replace('Base', ''),
```

packages/react-next/src/components/OverflowSet/OverflowSet.test.tsx
```diff
@@ -69,6 +70,11 @@ describe('OverflowSet', () => {
     });
   });
 
+  isConformant({
```
ecraig12345, line 73: Probably rather than renaming OverflowSet.ts to tsx, the test should be updated to handle either ts or tsx.

Same with other components.

change/@fluentui-react-button-2020-09-02-14-32-55-feat-implementing-v8-conformance-tests.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Adding conformant tests to v8 component test files.",
```
ecraig12345, line 3: For packages where you just changed tests, the change type should be none. (But it should be patch if you added displayNames or fixed other things.)

packages/react-button/src/components/SplitButton/SplitButton.test.tsx
ecraig12345: Unless I'm misunderstanding and the as tests work for the other buttons but not this one

packages/react-next/src/components/TextField/MaskedTextField/MaskedTextField.test.tsx
```diff
@@ -13,6 +13,12 @@ describe('MaskedTextField', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  isConformant({
+    Component: MaskedTextField,
+    displayName: 'MaskedTextField',
+    disabledTests: ['has-top-level-file', 'exported-top-level'],
```
ecraig12345, line 19: This does appear to be exported (at least in OUFR--if it's not in react-next we should fix that)
```suggestion
    disabledTests: ['has-top-level-file'],
```

packages/react-checkbox/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -4,6 +4,7 @@ import { create } from '@uifabric/utilities/lib/test';
 import { mount, ReactWrapper } from 'enzyme';
 
 import { Checkbox } from './Checkbox';
+import { isConformant } from '../../common/isConformant';
```
ecraig12345, line 7: Can you add this for the /next checkbox too?

Same for link, slider, tabs, and toggle's /next versions.

xugao: i decided to not duplicate the tests (yet) in `/next` folders. so no need here

packages/react-next/src/utilities/ButtonGrid/ButtonGrid.test.tsx
```diff
@@ -43,7 +44,7 @@ describe('ButtonGrid', () => {
   });
   it('Can render a ButtonGrid with width of 2', () => {
     const wrapper = shallow(
-      <ButtonGrid
+      <ButtonGridBase
```
ecraig12345: Just curious, why are these being run with ButtonGridBase? That's not what we do for most of the component tests.

packages/react-checkbox/src/components/Checkbox/Checkbox.test.tsx
ecraig12345: Why not? To me it seems good to have the conformance tests start running early. (Though if we did that we'd probably need to add an option to react-conformance to use a different entry point file besides `src/index` for parsing.)

ecraig12345:
@czearing There are more merge conflicts now :(

packages/react-checkbox/src/components/Checkbox/Checkbox.test.tsx
xugao: i was trying to avoid duplication of the same code in the code base. `/next` is not something we should worry until we get v8 out of the door. we have 2 copies now, didn't want to make it 3.

packages/react-next/src/utilities/ButtonGrid/ButtonGrid.test.tsx
czearing: I investigated this and it is a bit of a strange scenario. It seems after switching `ButtonGridBase` to `ButtonGrid` and shallow to safeMount, the tests fail because the gridcell role on the `DefaultButton` is called 32 times on both the v7 and v8 versions of the test. What is strange is if you console.log(wrapper.html) the gridcell role only exists eight times. After switching `DefaultButton` to button, all of the tests suddenly pass. I switched both versions to use `ButtonGrid` but I’m not entirely sure of what caused the error with `DefaultButton`.

packages/react-slider/src/common/isConformant.ts
```diff
@@ -1,8 +1,9 @@
 import { isConformant as baseIsConformant, IsConformantOptions } from '@fluentui/react-conformance';
 
-export function isConformant(testInfo: IsConformantOptions) {
+export function isConformant(testInfo: Omit<IsConformantOptions, 'componentPath'>) {
```
xugao, line 3: this typing change looks incorrect to me. why is this here?

xugao: same for other places

ecraig12345: Because componentPath is now being calculated automatically based on the test filename (see below)

packages/react-next/src/common/isConformant.ts
```diff
@@ -1,8 +1,9 @@
 import { isConformant as baseIsConformant, IsConformantOptions } from '@fluentui/react-conformance';
 
-export function isConformant(testInfo: IsConformantOptions) {
+export function isConformant(testInfo: Omit<IsConformantOptions, 'componentPath'> & { componentPath?: string }) {
   const defaultOptions = {
     disabledTests: ['has-docblock', 'kebab-aria-attributes'],
```
xugao, line 5: I am confused on why `as-*` tests are included and passing for components in `react-next`. `as` should not be supported on most components here

packages/react-slider/src/common/isConformant.ts
xugao: then shouldn't we update `IsConformantOptions` typing instead?

packages/react-next/src/common/isConformant.ts
ecraig12345: Internally the `as-*` tests check if the component has an `as` prop, and only run if it does.

packages/react-slider/src/common/isConformant.ts
ecraig12345: No, the `componentPath` is a project-specific detail which each project's `isConformant` wrapper may be able to provide for its own components, but the `react-conformance` package can't provide generically.

packages/react-button/src/components/Button/Button.test.tsx
```diff
@@ -1,16 +1,16 @@
...
 import * as renderer from 'react-test-renderer';
 import { mount, ReactWrapper } from 'enzyme';
+import { isConformant } from '../../common/isConformant';
+
 import { MergeStylesProvider } from '@fluentui/react-theme-provider';
 import { ButtonProps } from './Button.types';
 
 /** Use merge-styles provider to ensure styles show up in snapshots. */
-const Button = React.forwardRef<HTMLButtonElement, ButtonProps>((props, ref) => (
+const ButtonComponent = React.forwardRef<HTMLButtonElement, ButtonProps>((props, ref) => (
```
ecraig12345, line 11: Might be better to call this something like ButtonWrapper

packages/react-checkbox/src/common/isConformant.ts
```diff
@@ -0,0 +1,16 @@
...
+  const defaultOptions = {
+    disabledTests: [
+      'has-docblock',
+      'kebab-aria-attributes',
+      'as-renders-fc',
```
ecraig12345, line 8: What happened with the `as` tests?

packages/react-next/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
```diff
@@ -70,7 +70,7 @@ export const ChoiceGroupOptionBase = (props: IChoiceGroupOptionProps) => {
         {imageSrc && (
           <div className={classNames.innerField}>
             <div className={classNames.imageWrapper}>
-              <Image src={selectedImageSrc} alt={imageAlt} {...imageSize} />
+              <Image src={imageSrc} alt={imageAlt} {...imageSize} />
```
ecraig12345, line 73: Why this change? Bug fix?

ecraig12345 (approved):
A couple questions but this generally looks good, thanks!

packages/react-link/src/common/isConformant.ts
```diff
@@ -0,0 +1,17 @@
...
+  const defaultOptions = {
+    disabledTests: [
+      `has-docblock`,
+      `kebab-aria-attributes`,
+      `as-renders-fc`,
+      `as-renders-react-class`,
+      `as-passes-as-value`,
```
xugao, line 10: i dont think we should have these disabled.

czearing: There seems to be a function component in link that is not using forwardRef. It could be possible that the old compose link is using is responsible, but I will dig in more. Here is the error isConformant runs into after link runs the as tests:
```  jsx
    [console.error] Error: Uncaught [Error: [console.error] Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?%s%s

    Check the render method of `LinkBase`.
        in MyComponent (created by LinkBase)
        in LinkBase (created by StyledLinkBase)
        in StyledLinkBase (created by WrapperComponent)
        in WrapperComponent]

      19 |     throw args[0];
      20 |   } else {
    > 21 |     throw new Error(`[console.${type}] ${args.join(' ')}`);
         |           ^
      22 |   }
      23 | }
      24 |
```

packages/react-next/src/components/ChoiceGroup/ChoiceGroupOption/ChoiceGroupOption.base.tsx
czearing: It is an intended bug fix that screener noticed in this pr. `ChoiceGroupOption`’s is supposed to have an Image src for both imageSrc and selectedImageSrc props to allow for a unique selection image. Here's what the fix does:

Without Fix:
image

Fix:
image

packages/react-next/src/components/ComboBox/ComboBox.test.tsx
```diff
@@ -68,6 +69,18 @@ describe('ComboBox', () => {
...
+  it(`renders`, () => {
+    safeCreate(<ComboBox options={DEFAULT_OPTIONS} />, wrapper => {
+      expect(wrapper.root).toBeDefined();
+    });
+  });
+
+  isConformant({
+    Component: ComboBox,
+    displayName: 'ComboBox',
+    disabledTests: ['component-renders'],
```
ecraig12345, line 81: Can you add a comment about why this had to be disabled?

packages/react-link/src/components/Link/Link.test.tsx
```diff
@@ -66,6 +69,12 @@ describe('Link', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  isConformant({
+    Component: Link,
+    displayName: 'Link',
+    disabledTests: [`as-renders-fc`, `as-passes-as-value`],
```
ecraig12345, line 75: Would also be good to comment about why these are disabled.

ecraig12345 (approved):
Approved once the comments on disabled tests have been added. Thanks!

packages/react-link/src/common/isConformant.ts
czearing: I figured out what was causing the remaining errors and the 'as' tests are now enabled in `link`.

ecraig12345:
Looks like you're missing a change file, and I'm seeing some weird failure from isConformant in react-northstar.

packages/react-toggle/src/components/Toggle/Toggle.test.tsx
```diff
@@ -20,6 +21,12 @@ describe('Toggle', () => {
     ).toEqual('Label');
   });
 
+  isConformant({
+    Component: Toggle,
+    displayName: 'Toggle',
+    disabledTests: ['as-passes-as-value'],
```
xugao, line 27: we should not need this one?

xugao (approved):
Thank you  for making this change!

PR #14876, opened 2020-09-03
feat(UI builder): Prevent invalid drop

vyhnalekl:
#### Pull request checklist

#### Description of changes

Added a list of components where user can drop other components, `DropSelector` is highlighting components in red if you cannot insert the element. Also, I added `boxShadow` instead of a `border` for drop selector (when you are dropping near the edge of a component)


packages/fluentui/react-builder/src/components/PropSelector.tsx
```diff
@@ -0,0 +1,114 @@
+import * as React from 'react';
+import { Dialog, Flex, RadioGroup, Divider, Text, Input } from '@fluentui/react-northstar';
+
+export const PropSelector: React.FunctionComponent<{ components: any; onConfirm: any; onCancel: () => void }> = ({
+  components,
+  onConfirm,
+  onCancel,
+}) => {
+  console.log(JSON.parse(JSON.stringify(components)));
```
ling1726, line 9: Why are you doing a deep copy for a console.log ? maybe I am missing some context here of what `component` might be.

Similarly in the line below, it doesn't look like you are doing anything with `components`

packages/fluentui/react-builder/src/components/PropSelector.tsx
```diff
@@ -0,0 +1,114 @@
...
+    },
+    {
+      name: 'options',
+      key: 'replace',
+      label: 'Replace component',
+      value: 'replace',
+    },
+  ];
+  const [selectedProp, setSelectedProp] = React.useState('');
+  const handlePropChange = (e, props) => {
```
ling1726, line 38: nit: you could use `_` for `e` since that argument is never used

packages/fluentui/react-builder/src/config.tsx
```diff
@@ -511,6 +558,10 @@ export const resolveDrop = (newChild: JSONTreeElement, parent: JSONTreeElement,
   console.log('config:resolveDrop RESULT', JSON.parse(JSON.stringify({ source: newChild, target: parent })));
 };
 
+export const isValidDrop = (element: JSONTreeElement) => {
+  return element.type === 'div' || Object.keys(element.props).length === 0 || element.props?.children;
```
ling1726, line 562: if `element.props` is null or undefined won't `Object.keys(element.props)` throw ?

packages/fluentui/react-builder/src/components/PropSelector.tsx
vyhnalekl: Thank you, I forgot to remove it

packages/fluentui/react-builder/src/config.tsx
vyhnalekl: It should never be undefined, in `resolveDraggingElement` we are retruning 
`return {
    uuid: getUUID(),
    $$typeof: 'Symbol(react.element)',
    type: displayName,
    moduleName: module,
    displayName,
    props: { children: [] },
    ...jsonTreeElement,
  };`
Even when you add component via just code writing the props is set to {}. Though it might be a good idea to check whether there are props just in case the code changes.

packages/fluentui/react-builder/src/components/Designer.tsx
```diff
@@ -150,6 +165,37 @@ const stateReducer: Reducer<DesignerState, DesignerAction> = (draftState, action
       treeChanged = true;
       break;
 
+    case 'PROP_SELECT_CANCEL':
+      draftState.droppingElement = null;
```
miroslavstastny, line 169: If `CANCEL` happens during move, the component being moved is lost. Undo works but wouldn't be better if `CANCEL` would rollback the move?

packages/fluentui/react-builder/src/components/DropSelector.tsx
```diff
@@ -105,6 +104,12 @@ export const DropSelector: React.FunctionComponent<DropSelectorProps> = ({
       selectorRef.current.style.width = `${targetRect.width}px`;
       selectorRef.current.style.height = `${targetRect.height}px`;
 
+      if (!isValidDrop(jsonTreeElement)) {
```
miroslavstastny, line 107: When dropping close to the edge, the component will not be dropped inside the hovered component, but next to it.
In that case the background should not be red:
image

miroslavstastny (commented):
I really like the idea, but there are some issues with that:
- The rest of the builder only considers children tree - it is not possible to select a component in other slot in the UI - either to edit its props or to use it as a drop target. 
- In the PropSelector dialog, there are strange slots offered: `as`, `size` for `Button` - dropping to these slots breaks the builder.

Think about what can be fixed and keep in mind that anything out of `config` should be UI library agnostic.

I would also propose to add an override option (`children`) to the PropSelector (selected by default).

packages/charting/src/components/Legends/Legends.styles.ts
```diff
@@ -94,7 +94,7 @@ export const getStyles = (props: ILegendStyleProps): ILegendsStyles => {
         ...fonts.small,
         lineHeight: '16px',
         padding: '8px',
-      }
+      },
```
vyhnalekl, line 97:
```suggestion
      }
```

packages/charting/src/components/VerticalBarChart/VerticalBarChart.base.tsx
```diff
@@ -132,7 +132,7 @@ export class VerticalBarChartBase extends React.Component<IVerticalBarChartProps
         className={this._classNames.root}
       >
         <FocusZone direction={FocusZoneDirection.horizontal}>
-          <svg width={svgDimensions.width} height={svgDimensions.height} style={{display: "block"}}>
+          <svg width={svgDimensions.width} height={svgDimensions.height} style={{ display: 'block' }}>
```
vyhnalekl, line 135:
```suggestion
          <svg width={svgDimensions.width} height={svgDimensions.height} style={{display: "block" }}>
```

packages/charting/src/components/VerticalBarChart/VerticalBarChart.base.tsx
```diff
@@ -132,7 +132,7 @@ export class VerticalBarChartBase extends React.Component<IVerticalBarChartProps
         className={this._classNames.root}
       >
         <FocusZone direction={FocusZoneDirection.horizontal}>
-          <svg width={svgDimensions.width} height={svgDimensions.height} style={{display: "block"}}>
+          <svg width={svgDimensions.width} height={svgDimensions.height} style={{display: "block" }}>
```
vyhnalekl, line 135:
```suggestion
          <svg width={svgDimensions.width} height={svgDimensions.height} style={{display: "block"}}>
```

ecraig12345:
@vyhnalekl Is this still needed?

PR #14880, opened 2020-09-03
chore: run Screener from Azure Blob

layershifter:
This PR implements a custom Screener runner for Fluent UI React Northstar.

Key differences:
- it uses Azure to store artifacts => we will not use `ngrok` that should improve speed and stability of tests
- it will not wait until a Screener check will end, CI job will be stopped once a Screener job will be created => runners will be freed faster

`screener.api.ts` contains almost 1 to 1 copy-paste from [`screener-runner/steps.js`](https://github.com/screener-io/screener-runner/blob/master/src/steps.js) to keep the API same.

**This PR removes an ability to run Screener locally for Fluent UI React Northstar.**

### Before

image

### After

**Task still runs the same time, but CI agent is now free**

image

---

Also adds a status update to pending:

image


azure-pipelines.yml
```diff
@@ -110,8 +112,20 @@ jobs:
...
+        displayName: build FUI N* VR Test
+
+      - task: AzureUpload@1
+        displayName: Upload PR deploy site
+        inputs:
+          SourcePath: 'packages/fluentui/docs/dist'
+          azureSubscription: 'UI Fabric (bac044cf-49e1-4843-8dda-1ce9662606c8)'
+          storage: fabricweb
+          ContainerName: '$web'
+          BlobPrefix: 'pr-deploy-site/$(Build.SourceVersion)/react-northstar'
```
ecraig12345, line 125: Will this conflict with the site being uploaded in the main version of PR deploy? Also can you use the variable here?

ecraig12345 (commented):
Looking good so far! Some minor comments.

azure-pipelines.yml
```diff
@@ -54,7 +54,7 @@ jobs:
     workspace:
       clean: all
     variables:
-      PR_DEPLOY: 1
+      URL_BASE_PATH: "/pr-deploy-site/${{ variables['Build.SourceBranch'] }}"
```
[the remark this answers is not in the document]
layershifter: I added `.devops/templates/variables.yml`, it also related to your next comment 👇

azure-pipelines.yml
layershifter: With added `.devops/templates/variables.yml` it's more visible. With current implementation it will not collide as for PR deploy we use PR's ref while this task uses commits' hashes. _And it may be overkill..._

---

1. There is no way to cancel jobs in Screener via API, it's visible even in `screener-runner` - it waits until the previous job will be finished
2. Screener uses commit SHA to build a connection between builds

If we will decide to go with PR refs instead we can receive weird race conditions when we build and upload a new artifact before the previous job will end.

Please let me know if you have ideas/suggestions.

scripts/config.ts
```diff
@@ -9,10 +9,8 @@ const env = process.env.NODE_ENV || 'development';
 const __DEV__ = env === 'development';
 const __PERF__ = !!process.env.PERF;
 const __PROD__ = env === 'production';
-let __BASENAME__ = process.env.PR_DEPLOY
-  ? // This needs a trailing slash or images won't work
-    `/pr-deploy-site/${process.env.BUILD_SOURCEBRANCH}/react-northstar/`
-  : '/';
+// This needs a trailing slash or images won't work
+let __BASENAME__ = process.env.DEPLOYBASEPATH ? `${process.env.DEPLOYBASEPATH}/react-northstar/` : '/';
```
layershifter, line 13:
```suggestion
const __BASENAME__ = process.env.DEPLOYBASEPATH ? `${process.env.DEPLOYBASEPATH}/react-northstar/` : '/';
```

scripts/screener/screener.runner.ts
```diff
@@ -0,0 +1,102 @@
...
+    method: 'post',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify(payload),
+  });
+}
+
+export async function screenerRunner(screenerConfigPath) {
+  const screenerConfig: ScreenerRunnerConfig = require(screenerConfigPath) as any;
+
+  const commit = process.env.SYSTEM_PULLREQUEST_SOURCECOMMITID;
```
layershifter, line 84:
```suggestion
  // https://github.com/microsoft/azure-pipelines-tasks/issues/9801
  const commit = process.env.SYSTEM_PULLREQUEST_SOURCECOMMITID;
```

ecraig12345 (approved):
Looks good to me--thanks for doing this!

miroslavstastny (approved):

PR #14881, opened 2020-09-03
feat(Datepicker): Add advanced keyboard navigation

pompomon:
#### Pull request checklist

Continuation of https://github.com/microsoft/fluentui/pull/14862

Following key shortcuts are supported:
Home – First cell of current row
End – Last cell of current row
Ctrl + Home – First cell of the grid (first row first cell)
Ctrl + End – Last cell of the grid (last row last cell)
Page up – First cell of current column (basically scroll up/down behaviour but in this case there’s no scrollbar)
Page down – Last cell of current column

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Datepicker/DatepickerCalendar.tsx
+++ b/packages/fluentui/react-northstar/src/components/Datepicker/DatepickerCalendar.tsx
@@ -15,6 +15,8 @@ import {
   compareDatePart,
   getMonthStart,
   getMonthEnd,
+  getStartDateOfWeek,
+  getEndDateOfWeek,
 } from '@fluentui/date-time-utilities';
 import {
   ComponentWithAs,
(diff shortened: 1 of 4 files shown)


packages/fluentui/react-northstar/src/components/Datepicker/DatepickerCalendar.tsx
```diff
@@ -147,6 +149,53 @@ export const DatepickerCalendar: ComponentWithAs<'div', DatepickerCalendarProps>
...
+        const newNavigatedDate = contstraintNavigatedDate(gridNavigatedDate, targetDate, -1, restrictedDatesOptions);
+
+        if (!!newNavigatedDate) {
+          setShouldFocusInDayGrid(true);
+          setGridNavigatedDate(newNavigatedDate);
+        }
+      },
+      moveToEndOfWeek: e => {
+        e.preventDefault();
+        const targetDate = getEndDateOfWeek(gridNavigatedDate, firstDayOfWeek);
```
pompomon, line 169: Continuing discussion from https://github.com/microsoft/fluentui/pull/14862#discussion_r482913630, since PR was split

pompomon: Agreed with @rymeskar offline to keep current implementation like it is

layershifter (approved):

PR #14882, opened 2020-09-03
Feat/datepicker navigate to disabled dates

rymeskar:
#### Pull request checklist

#### Description of changes

* Add the possibility to navigate to disabled dates.

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Datepicker/DatepickerCalendar.tsx
```diff
@@ -74,6 +74,16 @@ export type DatepickerCalendarStylesProps = never;
...
+const getRidOfSecondsMinutesHours = (date: Date): Date => {
+  return new Date(date.getFullYear(), date.getMonth(), date.getDate());
+};
+
+const normalizeDateInGrid = (date: Date): Date => {
+  const result = new Date(date.getTime());
+  result.setDate(1);
+  return getRidOfSecondsMinutesHours(result);
+};
+
```
pompomon, line 86: Should we move that to `date-time-utilities` so it can be shared if needed?

rymeskar: I realized that I managed to implement some things differently and thus the method is no longer needed.

pompomon (approved):

PR #14887, opened 2020-09-03
useMergedRefs: now returns React.RefObject so that the current value can be referenced easily

dzearing:
Before:

```jsx
const Foo = React.forwardRef((props, ref) => {
  const mergedRef = useMergedRefs(ref, React.useRef());

  // Does not work because the ref is a function without a `current` prop available.
  return <ContextualMenu target={mergedRef} />
}
```

After:

```jsx
const Foo = React.forwardRef((props, ref) => {
  const mergedRef = useMergedRefs(ref, React.useRef());

  // Works because mergedRef is now a RefObject with `current` available.
  return <ContextualMenu target={mergedRef} />
}
```

### DIFF
--- a/packages/react-hooks/src/useMergedRefs.ts
+++ b/packages/react-hooks/src/useMergedRefs.ts
@@ -5,7 +5,7 @@ import * as React from 'react';
  * updates all provided refs
  * @param refs- Refs to collectively update with one ref value.
  */
-export function useMergedRefs<T>(...refs: (React.Ref<T> | undefined)[]): React.RefObject<T> {
+export function useMergedRefs<T>(...refs: (React.Ref<T> | undefined)[]): ((value: T) => void) & React.RefObject<T> {
   const mergedCallback = (React.useCallback(
     (value: T) => {
       // Update the "current" prop hanging on the function.
@@ -22,7 +22,7 @@ export function useMergedRefs<T>(...refs: (React.Ref<T> | undefined)[]): React.R
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps -- already exhaustive
     [...refs],
-  ) as unknown) as React.RefObject<T>;
+  ) as unknown) as ((value: T) => void) & React.RefObject<T>;
 
   return mergedCallback;
 }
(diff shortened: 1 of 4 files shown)


packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,22 +1,28 @@
...
+
 /**
  * React hook to merge multiple React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
  * updates all provided refs
  * @param refs- Refs to collectively update with one ref value.
  */
-export function useMergedRefs<T>(...refs: (Ref<T> | undefined)[]): (instance: T) => void {
-  return useCallback(
+export function useMergedRefs<T>(...refs: (React.Ref<T> | undefined)[]): ((value: T) => void) & React.RefObject<T> {
+  const mergedCallback = (React.useCallback(
```
ecraig12345, line 9: I think you can get rid of some of the casts if you do `const mergedCallback: ReturnType<typeof useMergedRefs<T>> = React.useCallback...` (or some other way of getting the type)

packages/react-hooks/src/useMergedRefs.ts
```diff
@@ -1,22 +1,28 @@
...
+import * as React from 'react';
+
 /**
  * React hook to merge multiple React refs (either MutableRefObjects or ref callbacks) into a single ref callback that
  * updates all provided refs
  * @param refs- Refs to collectively update with one ref value.
  */
-export function useMergedRefs<T>(...refs: (Ref<T> | undefined)[]): (instance: T) => void {
-  return useCallback(
+export function useMergedRefs<T>(...refs: (React.Ref<T> | undefined)[]): ((value: T) => void) & React.RefObject<T> {
```
ecraig12345, line 8: Would be good to add a `@returns` comment explaining the return value

packages/react-hooks/src/useMergedRefs.ts
dzearing: resolved

ecraig12345 (approved):

packages/react-hooks/src/useMergedRefs.ts
dzearing: done

dzearing:
@msft-github-bot merge in -1 minutes.

PR #14889, opened 2020-09-03
Buttons: using makeStyles, reworking SplitButton.

dzearing:
This change updates the buttons to use `makeClasses`, which brings inline the styling registration system. We'll see how this affects performance.

SplitButton was also reworked to render 2 buttons side by side with minimal overhead.


packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
```diff
@@ -45,8 +45,8 @@ export interface IVerticalStackedBarChartState {
 
 export class VerticalStackedBarChartBase extends React.Component<
```
dzearing, line 46: not supposed to be here...

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.types.ts
```diff
@@ -48,7 +48,7 @@ export interface IVerticalStackedBarChartProps {
    * Please look at https://github.com/d3/d3-format for all the formats supported
    */
   yAxisTickFormat?: (n: number) => string;
-
```
dzearing, line 51: umm

packages/react-button/src/components/MenuButton/useMenuButton.tsx
```diff
@@ -4,7 +4,7 @@ import { MenuButtonProps, MenuButtonState } from './MenuButton.types';
 import { useMenuButtonState } from './useMenuButtonState';
 import { renderMenuButton } from './renderMenuButton';
 
-export const menuButtonShorthandProps = ['icon', 'children', 'menuIcon', 'menu'];
+export const menuButtonShorthandProps = ['icon', 'content', 'menuIcon', 'menu'];
```
dzearing, line 7: this should be applied everywhere

packages/react-button/etc/react-button.api.md
```diff
@@ -198,7 +198,7 @@ export const useButton: (props: ButtonProps, ref: React.Ref<HTMLElement>, defaul
 };
 
 // @public (undocumented)
-export const useButtonClasses: (state: Record<string, any>) => void;
+export const useButtonClasses: (state: any) => void;
```
dzearing, line 201: see if this can be fixed.

dzearing: fixed.

dzearing:
Looked into bundle changes (parse sizes), observations:

* The cssinjs styles seem to be about 3k larger (minified), wonder if there's ways to reduce (+3k)
* MergeStyles adds about 4k minified, offset by about 2k removed from loadThemedStyles (+2k)
* Adding useTheme, which adds customizer (+2k)

This could be "reduced" by the next PR I have which moves style registration to the theme provider. This would mean the button graph would not have the style registration mechanism. This would reduce things a little more, but artificially...

PR #14907, opened 2020-09-04
chore: do not reload pages between steps in Screener

layershifter:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


layershifter:
It works, but seems useless as gives zero perf improvements.

PR #14916, opened 2020-09-06
fix(Tree): always show selectable indicator

### LINKED ISSUE #14720: TreeItem’s selectionIndicator needs to be unhidden or unhide-able
### Environment Information

- **Package version(s)**: `@fluentui/react-northstar@0.51.0`

### Describe the issue:

TreeItems which are parents hide their selection indicators when they are collapsed. This is not coherent with [WCAG guideline 1.3](https://www.w3.org/TR/WCAG20/#content-structure-separation) which states “Create content that can be presented in different ways (for example simpler layout) without losing information or structure”, since summary information for the item (the checkbox in indeterminate or checked state) is removed when the TreeItem is collapsed. There does not appear to be a way to override this behavior.

The selection of children in the Tree can have a significant impact on other information in the app using the Tree component. Hiding signifiers of state like this is [often discouraged](https://www.nngroup.com/articles/zen-mode/) since users could inadvertently remove information that requires extraordinary effort to restore: e.g., imagine if a user needs to find one selected child among a hundred parent TreeItems all of which are collapsed.

Regardless of the selection state, hiding the checkbox and showing it only on hover or focus removes the interactivity signifier for touch users, who wouldn't be able to distinguish between a selectable item and a non-selectable item without touching it first.

https://codesandbox.io/s/kv9t7

#### Actual behavior:

After selecting some children and collapsing the parent, the visual state of the Tree is identical to when no children are selected. It is not possible to know whether none, some, or all of a parent's children are selected.

#### Expected behavior:

The selection indicator persists regardless of the parent's collapsed state, revealing whether none, some, or all of its children are selected.

assuncaocharles:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14720

#### Description of changes

> TreeItems which are parents hide their selection indicators when they are collapsed. This is not coherent with [WCAG guideline 1.3](https://www.w3.org/TR/WCAG20/#content-structure-separation) which states “Create content that can be presented in different ways (for example simpler layout) without losing information or structure”, since summary information for the item (the checkbox in indeterminate or checked state) is removed when the TreeItem is collapsed. There does not appear to be a way to override this behavior.

This PR will make the selectable indicator always visible.

#### Focus areas to test

(optional)


kolaps33:
I have following questions :)
1. 
image
When I navigate to the parent node and press 'space' key, then selection is not executed. Parent node is only expanded. 
In production 'space' key do selection when parent node is expanded. 

2. 
image
should be as well custom selection indicator "select all" visible ? in the same way as "checkbox" is visible when tree is collapsed...

packages/fluentui/react-northstar/src/components/Tree/TreeTitle.tsx
```diff
@@ -120,11 +120,13 @@ export const TreeTitle: ComponentWithAs<'a', TreeTitleProps> & FluentComponentSt
     debugName: TreeTitle.displayName,
     actionHandlers: {
       performClick: e => {
+        // console.log('performClick');
```
kolaps33, line 123: high probably only leftover

packages/fluentui/react-northstar/src/components/Tree/TreeTitle.tsx
```diff
@@ -120,11 +120,13 @@ export const TreeTitle: ComponentWithAs<'a', TreeTitleProps> & FluentComponentSt
...
         e.preventDefault();
         e.stopPropagation();
         handleClick(e);
       },
       performSelection: e => {
+        // console.log('performSelection');
```
kolaps33, line 129: high probably only leftover

kolaps33:
I did testing of the changes with the deployed container after the commit:
https://github.com/microsoft/fluentui/pull/14916/commits/91ff8ca78ca72e395845c1b92decc2b4ea337221

and changes looks good for me with following behavior:
1. if parent is collapsed and user click on selection indicator, then children items are selected and parent is NOT expanded
2. if parent is collapsed and user press space key, then all children items are selected and parent is NOT expanded
3. if parent is collapsed and user press enter key, then children are expanded
4. if parent is collapsed and user click on the tree item, then children are expanded
5. if parent is expanded and user click on selection indicator(of parent) or press space key, then all children items stay expanded and state of children selection change to selected/unselected(based on initial state)
6. if parent is expanded and user press enter key or click on parent tree item, then children get collapsed and state of the children is not changed (if items were selected stay selected or vise-versa)

all above scenarios works with mouse as well with keyboard. 
Approving PR.

kolaps33 (approved):

layershifter (approved):

PR #14927, opened 2020-09-07
chore(Provider): fix unstable UT

layershifter:
This PR fixes an unstable unit test in `Provider`s test's suite. There is a small chance that `faker.lorem.word()` will return the same result (the test expects that `className` and `newClassName` will be different)  on a next call that's why tests where sometimes failing.


miroslavstastny (approved):

packages/fluentui/react-northstar/test/specs/components/Provider/Provider-test.tsx
```diff
@@ -325,9 +324,9 @@ describe('Provider', () => {
     });
 
     it('reacts on "className" update and keeps node in HTML tree', () => {
-      const className = faker.lorem.word();
+      const className = 'a-sample-classname';
       const wrapper = mount(
-        <Provider className={className}>
+        <Provider className="foo">
```
layershifter, line 329:
```suggestion
        <Provider className={className}>
```

PR #14935, opened 2020-09-08
chore: refactor E2E tests

layershifter:
This PR performs a refactor of existing E2E tests to make them stable in CI.

---

Previously we used [`slowMo`](https://developers.google.com/web/tools/puppeteer/debugging) option to add a delay between Puppeteer actions and it _was enough_ (not really). Our tests were stable locally, but non in CI as there were random failures 💣

I tried to increase `slowMo` to `25`/`100`/`250` and it made tests slower but they were still unstable 😨 Upgrade of Puppeteer, Chromium, dependencies was not helpful.

During investigation I set `slowMo` to `0` and then I was able to reproduce all CI issues locally. Some E2E methods were stable and some of them were continuously failing. There was a connection between unstable tests, methods that are used in them were implemented as getters, for example this scenario was failing:

```js
e2e.clickOn(button); // click occurred
expect(e2e.isFocused(menuItem).toBe()); // focus may be already there or will be there in 50/100/200/random ms
```

From Puppeteer's issues and docs I found out that we can't use there getters ❔  We should use assertion style and wait for expected changes as there is no right time to execute getter, this is illustrated by changes below:

```tsx
// Before
public exists = async (selector: string) => {
  return await this.page.$$(selector)).length > 0;
};

// After
public exists = async (selector: string): Promise<void> => {
  await this.page.waitForSelector(selector, { timeout: PUPPETEER_ACTION_TIMEOUT });
};
```

The previous implementation will be immediately executed and may fail as changes to DOM will not be applied yet.

---

Changes that I did:

- all E2E methods now return `void` and works like getters
- `clickOnPosition()` was removed as it produced unstable results
- `pressKey()` changed to `waitForSelectorAndPressKey()`


layershifter:
1

packages/fluentui/e2e/e2eApi.ts
```diff
@@ -35,47 +35,78 @@ export class E2EApi {
...
+  public hasComputedStyle = async (
+    selector: string,
+    property: keyof CSSStyleDeclaration,
+    value: string,
+  ): Promise<void> => {
+    await this.page.waitForFunction(
+      (selector, property, value) => {
+        return window.getComputedStyle(document.querySelector(selector))[property] === value;
+      },
+      { timeout: 10 * 1000 },
```
ling1726, line 47: you hard code the same timeouts in each of your selectors. Would it make sense to store this value as a const or provide it as some part of config ?

packages/fluentui/e2e/e2eApi.ts
```diff
@@ -35,47 +35,78 @@ export class E2EApi {
...
 
-  public textOf = async (selector: string) => {
-    const element = await this.getElement(selector);
-    return await (await element.getProperty('textContent')).jsonValue();
+  public clickOn = async (selector: string): Promise<void> => {
+    await (await this.getElement(selector)).click();
   };
 
-  public focusOn = async (selector: string) => await (await this.getElement(selector)).focus();
+  public textOf = async (selector: string, text: string): Promise<void> => {
```
ling1726, line 81: Should these helper assert functions return some kind of meaningful error message instead of only throwing a generic puppeteer exception ?

packages/fluentui/e2e/e2eApi.ts
layershifter: extracted to `PUPPETEER_ACTION_TIMEOUT`

packages/fluentui/e2e/e2eApi.ts
layershifter: It will be great, I will think how to implement it in more obvious way

packages/fluentui/e2e/e2eApi.ts
```diff
@@ -28,54 +45,99 @@ export class E2EApi {
...
+      { timeout: PUPPETEER_ACTION_TIMEOUT },
+      selector,
+      property,
+      value,
+    );
   };
 
-  public count = async (selector: string) => {
-    return (await this.page.$$(selector)).length;
+  public count = async (selector: string, count: number): Promise<void> => {
```
miroslavstastny, line 83:
```suggestion
  public expectCount = async (selector: string, count: number): Promise<void> => {
```
or `hasCount` or anything which implies this is an assertion.

The same applies for `hidden` and `textOf`.

miroslavstastny (approved):

PR #14952, opened 2020-09-09
feat(UI builder): Integrated the screen reader message computation script into UI Builder

adamsamec:
#### Description of changes
This is the first integration of the screen reader message computation code into the Ui builder. When user selects a component added to the canvas, this narration appears below the canvas. The narration tells exactly what a screen reader would announce upon focusing the element.

So far not all element types are supported by the computation, only the most important ones, namely:

* &lt;button&gt;
* <textarea>
* <input type="text">
* <input type="checkbox"> and role="checkbox"
* <input type="radio"> and role="radio"
* role="menuitem"
* role="option"

### DIFF
--- a/packages/fluentui/react-builder/src/components/MessageComputer.tsx
+++ b/packages/fluentui/react-builder/src/components/MessageComputer.tsx
@@ -1,0 +1,247 @@
+/*
+ TODO:
+* Add the missing and not so obvious attributes (e.g. "aria-haspopup" or "aria-expanded") for the already defined roles (e.g. "menuitem" or "checkbox") according to the specification.
+* With JAWS, in the case of the element with the "listbox" role, differentiate between having and not having the aria-multiselectable="true" attribute. If this attribute is present, then aria-selected="false" on the child elements with the role "option" behave differently than if it is not present. Specifically, if aria-multiselectable="true" is present, the aria-selected="false" causes the narration of "not selected", but if present, having the aria-selected attribute makes no difference to the narration. 
+ * What about the "disabled" state?
+ */
+import SRMC from './SRMC-Definitions';
+import './SRMC-Rules-Win_JAWS';
+
+export default class MessageComputer {
+  private computedMessageParts: { [key: string]: string } = {
+    value: '',
+    name: '',
+    description: '',
+    type: '',
+    state: '',
+    position: '',
+    usage: '',
+  };
+
+  // Asynchronously computes and returns the complete screen reader message for the given element and platform.
+  async computeMessage(element: HTMLElement, platform: string): Promise<string> {
+    let definitionName = this.getDefinitionName(element, platform, 'stateRules');
+
+    // Determine and save the usage string
+    this.computedMessageParts.usage = '';
+    const usages = SRMC.usageStrings[platform][definitionName];
+    if (usages) {
+      // Begin if 1
+      if (usages['[default]']) {
+        // Begin if 2
+        this.computedMessageParts.usage = usages['[default]'];
+      } // End if 2
+      for (let usageName in usages) {
+        // Begin for 1
+        const split = usageName.split('=');
+        const state = split[0];
+        const stateValue = split[1];
+        const stateAndValueMatch = element.getAttribute(state) === stateValue;
+        const checkedDOMPropAndValueMatch =
+          state === 'checked' && (element as HTMLInputElement).checked.toString() === stateValue;
+        if (stateAndValueMatch || checkedDOMPropAndValueMatch) {
+          // Begin if 2
+          this.computedMessageParts.usage = usages[usageName];
+        } // End if 2
+      } // End for 1
+    } // End if 1
+
+    // compute the element's accessible description
+    let computedDescription = '';
+
+    // First, handle some special case conditions
+    let value;
+    if (definitionName === 'textarea') {
+      // Begin if 1
+      value = (element as HTMLTextAreaElement).value.trim();
+    } // End if 1
+    if (definitionName === 'textarea' && platform === 'Win/JAWS' && value) {
+      // Begin if 1
+      computedDescription = SRMC.stateStrings['Win/JAWS']['textarea']['[extra1]'];
+    } else {
+      // else if 1
+      const describedby = element.getAttribute('aria-describedby');
+      let descElement;
+      if (describedby) {
+        // Begin if 2
+        const computedDescriptionArr = [];
+        const describedbyArr = describedby.split(/\s+/);
+        for (let i = 0; i < describedbyArr.length; i++) {
+          // Begin for 1
+          descElement = document.getElementById(describedbyArr[i]);
+          if (descElement) {
+            // Begin if 3
+            computedDescriptionArr.push(descElement.textContent);
+          } // End if 3
+        } // End for 1
+        computedDescription = computedDescriptionArr.join('');
+      } // End if 2
+    } // End if 1
+
+    // Save the computed accessible description
+    this.computedMessageParts.description = computedDescription;
+
+    // Retrieve the computed accessibility properties
+    const node = await (window as any).getComputedAccessibleNode(element);
+
+    // Compute and save the element's accessible name
+    const computedName = node.name;
+    this.computedMessageParts.name = computedName;
+
+    // If present, set the element's title as the accessible description if conditions are met
+    let title = element.getAttribute('title');
+    if (title && computedName !== title && !this.computedMessageParts.description) {
+      // Begin if 1
+      this.computedMessageParts.description = title;
+    } // End if 1
+
+    // Retrieve the element's value and save it
+    this.computedMessageParts.value = node.valueText;
+
+    // Add an unspecified position in set
+    const positionRoles = ['role=menuitem', 'role=menuitemcheckbox', 'role=menuitemradio', 'role=tab', 'role=option'];
+    if (positionRoles.includes(definitionName)) {
+      // Begin if 1
+      this.computedMessageParts.position = '[X of Y]';
+    } // End if 1
+
+    // Compute the element's type and state
+    this.computedMessageParts.type = '[' + node.role + ']';
+    this.computedMessageParts.state = '';
+    const rules = SRMC.stateRules[platform][definitionName];
+    if (rules) {
+      // Begin if 1
+      rulesLoop: for (let i = 0; i < rules.length; i++) {
+        // Begin for 1
+        const rule = rules[i];
+        const possibleStates = SRMC.possibleStates[definitionName];
+        for (let j = 0; j < possibleStates.length; j++) {
+          // Begin for 2
+          const possibleState = possibleStates[j];
+          const stateValue = element.getAttribute(possibleState);
+          const elementHasState =
+            stateValue !== null && (stateValue !== 'false' || !SRMC.falseMeansOmitted.includes(possibleState));
+          const elementHasCheckedProp = (element as HTMLInputElement).checked !== undefined;
+          const elementHasStateOrCheckedProp = elementHasState || elementHasCheckedProp;
+          const combinationHasState = rule.combination.includes(possibleState);
+          if (elementHasStateOrCheckedProp !== combinationHasState) {
+            // Begin if 2
+            continue rulesLoop;
+          } //End if 2
+        } // End for 2
+
+        // Retrieve and save the element's type
+        this.computedMessageParts.type = rule.elementType;
+
+        // Compute and save the element's state
+        const computedStateArr = [];
+        let order;
+        if (rule.combination.length <= 1 && !rule.order) {
+          // Begin if 2
+          order = rule.combination;
+        } else {
+          // else if 2
+          order = rule.order;
+        } // End if 2
+        for (let j = 0; j < order.length; j++) {
+          // Begin for 2
+          const state = order[j];
+          let stateValue;
+          if (state === 'checked') {
+            // Begin if 2
+            stateValue = (element as HTMLInputElement).checked;
+          } else {
+            // else if 2
+            stateValue = element.getAttribute(state) || 'false';
+          } // End if 2
+          const strings = SRMC.stateStrings[platform][definitionName];
+          const statePart = strings[state + '=' + stateValue] || strings[state + '='];
+          if (statePart) {
+            // Begin if 2
+            computedStateArr.push(statePart);
+          } // End if 2
+        } // End for 2
+        const computedState = computedStateArr.join(SRMC.STATE_PART_SEPARATOR);
+        this.computedMessageParts.state = computedState;
+        break;
+      } // End for 1
+    } // End if 1
+    definitionName = this.getDefinitionName(element, platform, 'readingOrder');
+    const computedMessage = this.composeMessageFromParts(definitionName, platform);
+    return computedMessage;
+  } // End computeMessage
+
+  // Returns the definition name based on the given DOM element, platform and definition type.
+  getDefinitionName(element: HTMLElement, platform: string, definitionType: string): string {
+    // Determine the definitions source by the definition type
+    let definitions;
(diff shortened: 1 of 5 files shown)


packages/fluentui/react-builder/src/components/MessageComputer.tsx
```diff
@@ -0,0 +1,247 @@
...
+ TODO:
+* Add the missing and not so obvious attributes (e.g. "aria-haspopup" or "aria-expanded") for the already defined roles (e.g. "menuitem" or "checkbox") according to the specification.
+* With JAWS, in the case of the element with the "listbox" role, differentiate between having and not having the aria-multiselectable="true" attribute. If this attribute is present, then aria-selected="false" on the child elements with the role "option" behave differently than if it is not present. Specifically, if aria-multiselectable="true" is present, the aria-selected="false" causes the narration of "not selected", but if present, having the aria-selected attribute makes no difference to the narration. 
+ * What about the "disabled" state?
+ */
+import SRMC from './SRMC-Definitions';
+import './SRMC-Rules-Win_JAWS';
+
+export default class MessageComputer {
+  private computedMessageParts: { [key: string]: string } = {
```
ling1726, line 11: `{ [key: string]: string } ` is an incredibly funny type :)

I would suggest you either strongly type this object or at the very least use the `Record<string, string>` utility type for better readability

packages/fluentui/react-builder/src/components/MessageComputer.tsx
```diff
@@ -0,0 +1,247 @@
...
+  };
+
+  // Asynchronously computes and returns the complete screen reader message for the given element and platform.
+  async computeMessage(element: HTMLElement, platform: string): Promise<string> {
+    let definitionName = this.getDefinitionName(element, platform, 'stateRules');
+
+    // Determine and save the usage string
+    this.computedMessageParts.usage = '';
+    const usages = SRMC.usageStrings[platform][definitionName];
+    if (usages) {
```
ling1726, line 28: you can avoid the first `if` clause with `if(usages && usages['[default]'])`

Newer syntax could potentially be `if(usages?['[default]']` but I'm not sure if this repo supports that

packages/fluentui/react-builder/src/components/MessageComputer.tsx
```diff
@@ -0,0 +1,247 @@
...
+      // Begin if 1
+      if (usages['[default]']) {
+        // Begin if 2
+        this.computedMessageParts.usage = usages['[default]'];
```
ling1726, line 32: my bad, you can avoid the if clause if you assign the variable like this: `this.computedMessageParts.usage = usages['[default']] || '`

packages/fluentui/react-builder/src/components/MessageComputer.tsx
```diff
@@ -0,0 +1,247 @@
...
+      } // End if 2
+      for (let usageName in usages) {
+        // Begin for 1
+        const split = usageName.split('=');
+        const state = split[0];
+        const stateValue = split[1];
+        const stateAndValueMatch = element.getAttribute(state) === stateValue;
+        const checkedDOMPropAndValueMatch =
+          state === 'checked' && (element as HTMLInputElement).checked.toString() === stateValue;
```
ling1726, line 41: if `element` is not a `HTMLInputElement` will this line throw because `checked` might not be defined ?

packages/fluentui/react-builder/src/components/MessageComputer.tsx
```diff
@@ -0,0 +1,247 @@
...
+        if (stateAndValueMatch || checkedDOMPropAndValueMatch) {
+          // Begin if 2
+          this.computedMessageParts.usage = usages[usageName];
```
ling1726, line 44: nit: you can use a ternary expression here it might look cleaner than the if clause i.e.

```
this.computedMessageParts.usage = stateAndValueMath || checkedDOMPropAndValueMatch
  ? usages[usageName]
  : this.computedMessageParts.usage
```

packages/fluentui/react-builder/src/components/MessageComputer.tsx
```diff
@@ -0,0 +1,247 @@
...
+      // Begin if 1
+      if (usages['[default]']) {
+        // Begin if 2
+        this.computedMessageParts.usage = usages['[default]'];
+      } // End if 2
+      for (let usageName in usages) {
+        // Begin for 1
+        const split = usageName.split('=');
+        const state = split[0];
+        const stateValue = split[1];
```
ling1726, line 38: not sure if this language feature is configured in this repo but this might be a nicer assignment:

`const [split, stateValue] = usageName.split('=')`

packages/fluentui/react-builder/src/components/MessageComputer.tsx
```diff
@@ -0,0 +1,247 @@
...
+        const describedbyArr = describedby.split(/\s+/);
+        for (let i = 0; i < describedbyArr.length; i++) {
+          // Begin for 1
+          descElement = document.getElementById(describedbyArr[i]);
+          if (descElement) {
+            // Begin if 3
+            computedDescriptionArr.push(descElement.textContent);
+          } // End if 3
+        } // End for 1
+        computedDescription = computedDescriptionArr.join('');
```
ling1726, line 77: this might be more readable:

```
computedDescription = describedby
  .split(/\s+/)
  .map(x => {
    return document.getElementById(describedbyArr[i]) ? descElement.textContent : nul
  })
  .filter(x => x !== null)
  .join('')
```

packages/fluentui/react-builder/src/components/MessageComputer.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+    // Add an unspecified position in set
+    const positionRoles = ['role=menuitem', 'role=menuitemcheckbox', 'role=menuitemradio', 'role=tab', 'role=option'];
+    if (positionRoles.includes(definitionName)) {
+      // Begin if 1
+      this.computedMessageParts.position = '[X of Y]';
+    } // End if 1
+
+    // Compute the element's type and state
+    this.computedMessageParts.type = '[' + node.role + ']';
```
ling1726, line 109: nit: use string interpolation here: ``` `[${node.role}]` ```

packages/fluentui/react-builder/src/components/MessageComputer.tsx
```diff
@@ -0,0 +1,247 @@
...
+    name: '',
+    description: '',
+    type: '',
+    state: '',
+    position: '',
+    usage: '',
+  };
+
+  // Asynchronously computes and returns the complete screen reader message for the given element and platform.
+  async computeMessage(element: HTMLElement, platform: string): Promise<string> {
```
ling1726, line 22: This is an incredibly large function. Is it possible to compose this from smaller functions ?

adamsamec:
#### Description of changes
This is the first integration of the screen reader message computation code into the Ui builder. When user selects a component added to the canvas, this narration appears below the canvas. The narration tells exactly what a screen reader would announce upon focusing the element.

So far not all element types are supported by the computation, only the most important ones, namely:

* &lt;a&gt;
* &lt;button&gt;
* &lt;textarea&gt;
* &lt;input type="text"&gt;
* &lt;input type="checkbox"&gt; and role="checkbox"
* &lt;input type="radio"&gt; and role="radio"
* &lt;select&gt;
* role="menuitem", role="menuitemcheckbox" and role="menuitemradio"
* role="option"
* role="combobox"
* role="switch"
* role="tab"

Also, currently, only  computation for the JAWS screen reader is supported, and the narration reflects the state when the element is not focused using the virtual PC cursor but using the Tab key.

packages/fluentui/react-builder/src/components/MessageComputer.tsx
assuncaocharles:
```suggestion
  private computedMessageParts: Record<string, string> = {
```

Just to make it simpler :)

packages/fluentui/react-builder/src/components/MessageComputer.tsx
assuncaocharles, line 38:
```suggestion
        const [state, stateValue] = usageName.split('=')
```

adding @ling1726  suggestion just to make it simpler to apply :)

packages/fluentui/react-builder/src/components/MessageComputer.tsx
assuncaocharles: Yeah it will throw, because won't read `checked` from undefined 
```suggestion
          state === 'checked' && (element as HTMLInputElement).checked?.toString() === stateValue;
```

packages/fluentui/react-builder/src/components/SRMC-Definitions.tsx
```diff
@@ -0,0 +1,222 @@
+const SRMC: { [key: string]: any } = {
```
ling1726, line 1: you since this is a object of constants, you shouldn't really define any type for this. Let typescript inference do its job :)

```suggestion
const SRMC = {
```

packages/fluentui/react-builder/src/components/SRMC-Definitions.tsx
```diff
@@ -0,0 +1,222 @@
...
+  // Element usage strings for each platform and definition name
+  usageStrings: {
+    'Win/JAWS': {
+      button: {
+        '[default]': 'To activate press Enter.',
+        'aria-pressed=true': 'To toggle the state press spacebar.',
+        'aria-pressed=false': 'To toggle the state press spacebar.',
+      }, // End button
+      'input:text': {
+        '[default]': 'Type in text.',
```
ling1726, line 129: I'm confused as to why the square brackets are needed

packages/fluentui/react-builder/src/components/MessageComputer.tsx
adamsamec: But the state of undefined should never happen, because state === 'checked' is true only for checkbox element type, provided that the SRMC definitions are correct. If they are incorrect, it would fail silently after this modification.

packages/fluentui/react-builder/src/components/MessageComputer.tsx
adamsamec: Shouldn't it be rather like this:

```
computedDescription = element.getAttribute('aria-describedby')
    ?.split(/\s+/)
    .map(x => {
        return element.ownerDocument?.getElementById(x)?.textContent || null;
    })
    .filter(x => x !== null)
    .join('') || '';
```

ling1726: yeah something along those lines. Main point was that the `map` function might make your life easier here :)

packages/fluentui/react-builder/src/components/MessageComputer.tsx
adamsamec: How would the "strongly typed" version look like?

ling1726: you create a type with each of those properties

```
type ComputedMessage = {
  value: string;
  name: string;
}
```

This will mean that you get some handy tools like autocomplete in your editor and some type safety. However, if those aren't important then stick with the other suggestion and use `Record`

adamsamec: Thanks. What is actually bad on `{ [key: string]: string }`? It looks concise and clear to me.

ling1726: So if you do not type this object at all, typescript type inference will kick in a 'automatically' type this object for you when it's imported and you get all the benefits of declaring a type (you just can't reuse the type)

I do not like any type like `{ [key: string]: string }` or `Record<string, string>` since at that point you might as well use the type `any` or `object` to the same effect. 

why I suggest `Record` if you really really want a generic type with no information on the properties is that it's more readable

ling1726: Then again I am still not in the team yet so feel free to ask others what the team's opinions on this are :)

adamsamec: Okay, thanks. But `{[key: string]: string}` also ensures type safety. It's not the same as `:any`.

ling1726: Oh, that's new to me, that's fair. Apologies.

packages/fluentui/react-builder/src/components/SRMC-Definitions.tsx
adamsamec: When I apply your suggested change, that is, remove the `{ [key: string]: any }` typ, I get the following error:

```
TypeScript error in C:/Users/asamec/Dropbox/Work/Microsoft/Programming/sr-message-computation-ts/Windows/sr-message-computation/src/MessageComputer.tsx(47,27):
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ 'Win/JAWS': { '[default]': string[]; 'input:text':
string[]; 'role=combobox': string; a: string[]; 'role=menuitem': string[]; 'role=menuitemcheckbox': string; 'role=menuitemradio': string; select: string;
'role=tab': string; 'role=option': string; }; }'.
  No index signature with a parameter of type 'string' was found on type '{ 'Win/JAWS': { '[default]': string[]; 'input:text': string[]; 'role=combobox':
string; a: string[]; 'role=menuitem': string[]; 'role=menuitemcheckbox': string; 'role=menuitemradio': string; select: string; 'role=tab': string; 'role=option':
string; }; }'.  TS7053

    45 |         if (definitionType === 'readingOrder') {
    46 |             // Begin if 1
  > 47 |             definitions = SRMC.readingOrder[platform];
       |                           ^
    48 |         } else if (definitionType === 'stateRules') {
    49 |             // else if 1
    50 |             definitions = SRMC.stateRules[platform];
This error occurred during the build time and cannot be dismissed.
```

adamsamec:
The brackets are not needed. It is just to make it more apparent that this definition is not a specific state attribute like "checked=true".

packages/fluentui/react-builder/src/components/SRMC-Definitions.tsx
adamsamec: The brackets are not needed. It is just to make it apparent that it is not a state attribute or element definition. it from the state attributes or element definitions which don't have brakets.

adamsamec: Sorry for the confusing comment. It should have been like this:

The brackets are not needed. It is just to make it more apparent that it is not a specific state attribute like "checked=true".

adamsamec: The brackets are not needed. It is just to make it more apparent that this definition is not a specific state attribute like "checked=true".

packages/fluentui/react-builder/src/components/MessageComputer.tsx
assuncaocharles: `Record<string, string>` in this case will be more readable, concise, and consistent with what we have.

packages/fluentui/react-builder/src/components/MessageComputer.tsx
assuncaocharles: The problem here is that in the function signature you are expecting an `HTMLElement` as `element` and here you are doing the cast to `HTMLInputElement` and if for some reason `element` is a different type other `input`, `.toString` will throw an error. So either do the option optional chain or change the parameter type if this function can only be called with `inputs`

adamsamec: I've got your point. But it is intentional. The `state === 'checked'` condition ensures that this problematic line of code will be executed only in case of elements that hav the "checked" DOM property, that is `HTMLInputElement`.

PR #14954, opened 2020-09-09
feat(UI builder): Detect focus on body in use mode

vyhnalekl:
#### Pull request checklist

#### Description of changes

Detecting focus going to body in `canvas`. Displaying an warning message in header of the window. Based on [the hackathon PR](https://github.com/microsoft/fluentui/commit/d0a7ed1e3a630b6c7fde15732fdd8cbe61753c19).


packages/fluentui/react-builder/src/components/Canvas.tsx
```diff
@@ -298,7 +320,7 @@ export const Canvas: React.FunctionComponent<CanvasProps> = ({
               />
             )}
             <EventListener type="keydown" listener={onKeyDown} target={document} />
-            <Provider theme={teamsTheme} target={document}>
+            <Provider theme={teamsTheme} target={document} tabIndex={0} style={{ outline: 'none' }}>
```
ling1726, line 323: why do you need to set the tabindex on the `Provider` ?

packages/fluentui/react-builder/src/components/Canvas.tsx
```diff
@@ -115,6 +119,17 @@ export const Canvas: React.FunctionComponent<CanvasProps> = ({
...
+  const [bodyFocused, setBodyFocused] = React.useState(false);
+  const handleFocus = (ev: FocusEvent) => {
+    if (ev.target && (ev.target as any).getAttribute('data-builder-id') === null) {
+      !bodyFocused && setHeaderMessage('Warning: Focus on body.');
+      !bodyFocused && setBodyFocused(true);
+    } else {
+      bodyFocused && setHeaderMessage('');
+      bodyFocused && setBodyFocused(false);
+    }
+  };
```
ling1726, line 131: I think that this could be cleaner and easier to read

```suggestion
  const handleFocus = (ev: FocusEvent) => {
    // can add a comment here of what this check means in reality
    if (!ev.target || ev.target.getAttribute('data-builder-id')) {
      return;
    }
    
    if (bodyFocused) {
      setHeaderMessage('.');
      setBodyFocused(false);
    } else {
      setHeaderMessage('Warning: Focus on body.');
      setBodyFocused(true);
    }
```

packages/fluentui/react-builder/src/components/Canvas.tsx
```diff
@@ -115,6 +119,17 @@ export const Canvas: React.FunctionComponent<CanvasProps> = ({
     [iframeCoordinatesToWindowCoordinates, onMoveComponent],
   );
 
+  const [bodyFocused, setBodyFocused] = React.useState(false);
+  const handleFocus = (ev: FocusEvent) => {
+    if (ev.target && (ev.target as any).getAttribute('data-builder-id') === null) {
```
ling1726, line 124: nit: consider putting the expression `ev.target && (ev.target as any).getAttribute('data-builder-id') === null` in a well named variable and using `bodyFocused` in the if clause. I think it might be easier to read.

Certainly took me a while to get it :)

packages/fluentui/react-builder/src/components/Canvas.tsx
vyhnalekl: Because `Provider` essentially represents body.

miroslavstastny (approved):

PR #14956, opened 2020-09-09
hotfix(Breadcrumb): content with children and example name

assuncaocharles:
#### Pull request checklist

#### Description of changes

This PR renames the `BreadcrumbExample` to `BreadcrumbExampleBasic` to avoid the warning that is missing some example and fixes the console warning about `children` and `content` conflicting on `BreadcrumbDivider`

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Breadcrumb/BreadcrumbDivider.tsx
```diff
@@ -85,4 +86,9 @@ BreadcrumbDivider.defaultProps = {
   accessibility: breadcrumbDividerBehavior,
 };
 
-BreadcrumbDivider.propTypes = commonPropTypes.createCommon();
+BreadcrumbDivider.propTypes = {
+  ...commonPropTypes.createCommon({
+    content: false,
+  }),
+  content: customPropTypes.shorthandAllowingChildren,
+};
```
layershifter, line 94: This change is not valid. `children` and `content` cannot be used in the same time

From which example a warning comes?

assuncaocharles: It comes from usage `<Breadcrumb.Divider><Icon /></Breadcrumb.Divider>` and having `content: '/'` as `defaultProps`

layershifter: In this case please change it to `Breadcrumb.defaultProps.children`

assuncaocharles: Please, check it now

layershifter (approved):

PR #14964, opened 2020-09-10
adds the ability to set dialog zindex

paucortescosta:
#### Pull request checklist

#### Description of changes

adds the ability to set the dialog's zindex

#### Focus areas to test

(optional)

### DIFF
--- a/packages/web-components/src/dialog/dialog.styles.ts
+++ b/packages/web-components/src/dialog/dialog.styles.ts
@@ -10,6 +10,7 @@ export const DialogStyles = css`
     --elevation: 14;
     --dialog-height: 480px;
     --dialog-width: 640px;
+    --dialog-z-index: 0;
     display: block;
   }
 
@@ -32,6 +33,7 @@ export const DialogStyles = css`
     left: 0;
     right: 0;
     overflow: auto;
+    z-index: var(--dialog-z-index);
   }
 
   .control {


packages/web-components/src/dialog/dialog.styles.ts
```diff
@@ -32,6 +33,7 @@ export const DialogStyles = css`
     left: 0;
     right: 0;
     overflow: auto;
+    z-index: var(--dialog-z-index);
```
chrisdholt, line 36: Thanks @paucortescosta - we'll take a look at this. Perhaps you can provide a bit of a story around this and how it might scale to other controls that similarly have z-index considerations? We've been looking at larger scale z-index story where we might have a semantic ramp that could be either extended or completely replaced to scale across controls. It seems like something that scales would be preferable here - would that benefit your scenario?

With a solution , I'd likely prefer to bring this into a single line here:

```css
z-index: var(--dialog-z-index, 0);
```

Ahmed-Elkholy:
@chrisdholt 
Any idea how to workaround this problem? I am currently facing an issue and I need to increase the z-index for the dialog.

chrisdholt:
> @chrisdholt Any idea how to workaround this problem? I am currently facing an issue and I need to increase the z-index for the dialog.

Do you have a repo you can create using stackblitz or similar? Z-index and stacking context is a bit tricky so I'm not sure the best way to address a specific issue you have without some detail :).

Related: https://philipwalton.com/articles/what-no-one-told-you-about-z-index/

PR #14967, opened 2020-09-10
fix(UI builder) Component tree accessibility

vyhnalekl:
#### Pull request checklist

#### Description of changes

Fixed component tree accessibility.


miroslavstastny (approved):
LGTM.

It is from other change, but I do not understand this padding:

image

image

Wouldn't it be better without it?

PR #14968, opened 2020-09-10
Accessibility behavior tests and documentation generation

kolaps33:
#### Pull request checklist

#### Open questions/Future work
- invoke click and keyboard event with testing libary - https://github.com/testing-library/user-event > can be implement in future after agreement inside the team 
- where should be "[behaviorName]Definition" files stored? so far the definition file is in the same folder where is "behavior" file 
- currently generating documentation from "behavior" file and from "definition" exists together. 
If documentation is not found in behavior file, then is generated from "definition" file. In the future generating doc from behavior file need to be removed. 

#### Description of changes
This PR is coming out from the @mnajdova PR. 
https://github.com/microsoft/fluentui/pull/14797

Purpose of the PR is to test "behavior" files and generate documentation from each behavior file based on the test. 

In this PR following behaviors are tested:
- buttonBehavior
- buttonGroupBehavior
- toggleButtonBehavior

This behaviors are now cover by test which stpes are in the "[behaviorName]Definition" file. 

From "[behaviorName]Definition" file is as well documentation for each behavior file generated. 

It doesn't change the documentation, only content under "description" paragraph is coming from "[behaviorName]Definition" file (pl. see screenshot below):
image

#### Focus areas to test

(optional)


packages/fluentui/react-northstar/src/components/Button/Button.tsx
```diff
@@ -232,7 +232,6 @@ export const Button = compose<'button', ButtonProps, ButtonStylesProps, {}, {}>(
         {...rtlTextContainer.getAttributes({ forElements: [children] })}
         {...getA11yProps('root', {
           onClick: handleClick,
-          disabled,
```
kolaps33, line 235: Removing this line because it break the condition

` disabled: props.disabled && !props.loading ? (props.as === 'button' ? true : undefined) : undefined,`

which is in ButtonBehavior.ts file.

packages/react-button/src/components/ToggleButton/useChecked.test.ts
```diff
@@ -0,0 +1,12 @@
+import { useChecked } from './useChecked';
+import { checkedBehaviorDefinition } from './checkedBehaviorDefinition';
+import { HookTestFacade, validateBehavior } from '@fluentui/a11y-testing';
+
+describe('useChecked', () => {
```
mnajdova, line 5: Let's remove this file, it's not relevant anymore

packages/react-button/src/components/ToggleButton/checkedBehaviorDefinition.ts
```diff
@@ -0,0 +1,16 @@
+import { BehaviorRule, Rule } from '@fluentui/a11y-testing';
```
mnajdova, line 1: Let's remove this file, it's not relevant anymore

packages/react-button/src/components/ToggleButton/ToggleButton.test.tsx
```diff
@@ -1,6 +1,8 @@
 import * as path from 'path';
 import { isConformant } from '@fluentui/react-conformance';
 import { ToggleButton } from './ToggleButton';
+import { checkedBehaviorDefinition } from './checkedBehaviorDefinition';
```
mnajdova, line 4: Let's revert the changes here too, not relevant anymore

packages/react-button/package.json
```diff
@@ -50,6 +50,7 @@
     "react-test-renderer": "^16.3.0"
   },
   "dependencies": {
+    "@fluentui/a11y-testing": "^0.1.0",
```
mnajdova, line 53:
```suggestion
```

packages/react-avatar/.eslint.json
```diff
@@ -0,0 +1,8 @@
+{
```
mnajdova, line 1: Check if this is still required, I was getting some error so I needed to do this change

packages/charting/src/components/VerticalStackedBarChart/examples/VerticalStackedBarChart.Styled.Example.tsx
```diff
@@ -76,15 +76,15 @@ export class VerticalStackedBarChartStyledExample extends React.Component<{}, {}
             allowFocusOnLegends: true,
             styles: {
               rect: {
-                borderRadius: "3px"
-              }
-            }
+                borderRadius: '3px',
```
mnajdova, line 79: Let's revert these changes, they are not relevant anyhow to this PR.

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
```diff
@@ -148,13 +148,11 @@ export class VerticalStackedBarChartBase extends React.Component<
...
           {...this.props.calloutProps}
         >
-          {
-            this.props.onRenderCalloutPerStack
-              ? this.props.onRenderCalloutPerStack(this.state.stackCalloutProps)
-              : this.props.onRenderCalloutPerDataPoint
-                ? this.props.onRenderCalloutPerDataPoint(this.state.dataPointCalloutProps, this._renderCallout)
-                : this._renderCallout(this.state.dataPointCalloutProps)
-          }
+          {this.props.onRenderCalloutPerStack
```
mnajdova, line 151: Let's revert these changes too

packages/a11y-testing/src/validators/validate.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    if (slot.checkSpaceKeyPressed) {
+      testFacade.pressSpaceKey(slot.name);
+    }
+
+    if (slot.checkEnterKeyPressed) {
+      testFacade.pressEnterKey(slot.name);
+    }
+
+    if (slot.wasOnclickExecuted) {
```
mnajdova, line 23: For consistensy, can we rename this to checkOnClickWasExecuted, or something similar

packages/a11y-testing/src/facades/withProvider.tsx
```diff
@@ -0,0 +1,43 @@
+import { Telemetry, Unstable_FluentContextProvider, ProviderContextPrepared } from '@fluentui/react-bindings';
+import { Renderer, noopRenderer } from '@fluentui/react-northstar-styles-renderer';
+import { emptyTheme, ThemePrepared } from '@fluentui/styles';
+import { mount, MountRendererProps, ComponentType } from 'enzyme';
+import * as React from 'react';
+
+export const EmptyThemeProvider: React.FunctionComponent<{
```
mnajdova, line 7: @kolaps33 can we try to test if everything will work if we just don't wrap the component with theme provider. This was initially implemented like this, because the components will throw an error, but I am not sure what is the status at this moment. Would be great if we can remove the styles and other dependencies from this package.

packages/a11y-testing/src/facades/ComponentTestFacade.tsx
```diff
@@ -0,0 +1,95 @@
...
+    // reset body
+    document.body.innerHTML = '';
+    const container = document.createElement('div');
+    document.body.appendChild(container);
+
+    ReactDOM.render(<Component {...props} />, container);
+    this.actual = container.lastChild;
+
+    // we need to render it in this way because using simulate function to fire mouse/keyboard event
+    this.renderedComponent = mountWithProviderAndGetComponent(Component, <Component {...props} />);
```
mnajdova, line 25: I was refering to trying to remove this `mountWithProviderAndGetComponent` and just render the `Component` itself.

packages/a11y-testing/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "test": "just-scripts test",
+    "update-api": "just-scripts update-api",
+    "update-snapshots": "just-scripts jest -u"
+  },
+  "devDependencies": {
+    "@fluentui/common-styles": "^0.2.2",
+    "@fluentui/eslint-plugin": "^0.53.3",
+    "@fluentui/react-conformance": "^0.1.0",
+    "@fluentui/storybook": "^0.2.6",
+    "@testing-library/react": "^10.4.9",
```
mnajdova, line 34: I don't think we are using this anymore, as far as I remember it was only used in the `HooksTestFacade`, but we decided to go only with the `ComponentsTestFacade`

mnajdova (commented):
Let's add comments on the methods, to document what they are used for. It would help people get up to speed faster

packages/a11y-testing/etc/a11y-testing.api.md
```diff
@@ -0,0 +1,146 @@
+## API Report File for "@fluentui/a11y-testing"
```
kolaps33, line 1: This file was generating by running command in the package folder:
yarn update-api

and then manually copied from temp folder to this etc folder

packages/a11y-testing/package.json
kolaps33: removed

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
kolaps33: somehow commit always change formatting this file... I don't know how to fix it.

packages/charting/src/components/VerticalStackedBarChart/examples/VerticalStackedBarChart.Styled.Example.tsx
kolaps33: somehow commit always change formatting this file... I don't know how to fix it.

packages/a11y-testing/etc/a11y-testing.api.md
ecraig12345: `build` will give you the message about manually copying, but `update-api` should copy automatically. Also if this isn't intended to be a published package we can just turn off API Extractor.

packages/a11y-testing/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@fluentui/a11y-testing",
+  "version": "0.1.0",
```
ecraig12345, line 3: Is this intended to be published? If not you can add `"private": true`

packages/a11y-testing/package.json
```diff
@@ -0,0 +1,61 @@
...
+  "description": "Fluent UI tests for a11y patterns.",
+  "main": "lib-commonjs/index.js",
```
ecraig12345, line 5: From what I can tell, this package should have used the `node` template not the `react` template since it's intended to be used for testing only, not run in the browser (though I agree the template name is a little misleading). So now there's a bunch of configuration that's not quite right. 

Is it okay if I push an update to your PR fixing the config issues? (would be easier than trying to explain them all)

packages/react-button/src/components/ToggleButton/useChecked.test.ts
kolaps33: removed :)

packages/react-button/src/components/ToggleButton/checkedBehaviorDefinition.ts
kolaps33: removed :)

packages/react-button/src/components/ToggleButton/ToggleButton.test.tsx
kolaps33: changes reverted

packages/react-button/package.json
kolaps33: removed :)

packages/react-avatar/.eslint.json
kolaps33: removed whole file

packages/a11y-testing/src/validators/validate.ts
kolaps33: renamed :)

packages/a11y-testing/src/facades/withProvider.tsx
kolaps33: removed whole file and replace it with one line in `ComponentTestFacade.tsx`

packages/a11y-testing/src/facades/ComponentTestFacade.tsx
kolaps33: yes thanks, replaced with line:
` this.renderedComponent = mount(<Component {...props} />).find(Component);`

packages/fluentui/react-northstar/test/specs/components/Button/ButtonBehavior-test.tsx
```diff
@@ -0,0 +1,15 @@
+// import * as path from 'path';
```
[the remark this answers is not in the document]
kolaps33: agree, I am wondering what is opinion of @jurokapsiar

packages/a11y-testing/package.json
kolaps33: @ecraig12345 
please feel free to push update to this PR :) 
I appreciate you help. :) as I have no experience with creating the package. 
Even if you have any documentation where I could read more about creating package I would be happy to read it.

packages/a11y-testing/package.json
kolaps33: I am not sure about long term plan. @jurokapsiar please could you clarify? :)

jurokapsiar: not now, just internal. maybe in the future

packages/a11y-testing/package.json
ecraig12345: It's not documented 😬 but that would be a good idea since we're adding more packages now! I'm also planning to update the create-package script with more options (and clearer explanations) so it's easy for people to get the right setup in more cases.

scripts/gulp/plugins/gulp-component-menu-behaviors.ts
```diff
@@ -20,6 +20,11 @@ type BehaviorMenuItem = {
   };
 };
 
+async function importFile(pathDir, name) {
+  const path = `${pathDir}\\${name}.ts`;
```
ecraig12345, line 24: You should probably use `path.join` here since backslashes are Windows-specific

ecraig12345:
I just pushed an update which does the following:
- Set package to `private: true` and remove change file (we can always publish it later if needed)
- Change eslint config to extend the one for node (build/test utility) packages
- Removed storybook since it's only used for component packages
- Removed API Extractor config (locking down API isn't important for an internal package)
- Removed test configs (and most test-related dependencies): although this package has test-related utilities, it doesn't appear to have its own tests. If that changes, the test-related config can be added back.
- Change just.config.ts to use node library-style build (so it will only output commonjs modules to `lib`, rather than the component package build where ES modules go in `lib` and commonjs modules go in `lib-commonjs`) and updated entry points in package.json
- Remove unused scripts and dependencies
- Update tsconfig to use options for a node library: mainly removing unneeded types and using `lib: [es2017, dom]` instead of `es5` (for browser packages we restrict to ES5 types for IE 11 compatibility, but node libraries can use newer types)

packages/charting/src/components/VerticalStackedBarChart/examples/VerticalStackedBarChart.Styled.Example.tsx
kolaps33: with merge latest master it get fixed

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
kolaps33: with merge latest master it get fixed

packages/fluentui/react-northstar/test/specs/components/Button/ButtonBehavior-test.tsx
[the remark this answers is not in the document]
kolaps33: change done, tests were moved into the common component tests to be together in one file

scripts/gulp/plugins/gulp-component-menu-behaviors.ts
kolaps33: thanks, implemented proposal change

packages/a11y-testing/package.json
```diff
@@ -18,6 +18,7 @@
     "lint": "just-scripts lint"
   },
   "dependencies": {
+    "@fluentui/a11y-rules": "0.1.0",
```
layershifter, line 21:
```suggestion
    "@fluentui/a11y-rules": "^0.1.0",
```

packages/fluentui/accessibility/package.json
```diff
@@ -10,6 +10,7 @@
     "lodash": "^4.17.15"
   },
   "devDependencies": {
+    "@fluentui/a11y-rules": "0.1.0",
     "@fluentui/a11y-testing": "0.1.0",
```
layershifter:
```suggestion
    "@fluentui/a11y-rules": "^0.1.0",
    "@fluentui/a11y-testing": "^0.1.0",
```

packages/fluentui/accessibility-definitions/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@fluentui/accessibility-definitions",
+  "description": "A set of acc.... ",
```
ecraig12345, line 3: This looks unfinished? Also I'm curious why another package ended up being needed.

kolaps33: @ecraig12345  thank you for cath. Updated. 

The new package I am adding because of dependency which "definitions" files have inside. 
The "definitions" files depends on "a11y-rules" package. 

"Definitions" files were before together with "behaviors" files in "accessibility" package:
office-ui-fabric-react\packages\fluentui\accessibility\src\behaviors

"Accessibility" package is public.
But we don't want to make "a11y-rules" package public. (as dependency of "definittions" files) 

So this was the reason extract "definitions" files out of the "accessibility" package.

packages/fluentui/accessibility-definitions/src/definitions/Button/buttonBehaviorDefinition.ts
```diff
@@ -0,0 +1,48 @@
+import { Rule, BehaviorRule } from '@fluentui/a11y-rules';
```
layershifter, line 1: I suggest to eliminate there a level of nesting: `fluentui/accessibility-definitions/src/definitions/Button/buttonBehaviorDefinition.ts` => `fluentui/accessibility-definitions/src/Button/buttonBehaviorDefinition.ts`

layershifter: I did this change

scripts/gulp/plugins/gulp-component-menu-behaviors.ts
```diff
@@ -59,12 +74,32 @@ export default () => {
...
+            return definition.stringify();
+          });
+          variation.description = descriptionFromDefinition.join('\r\n');
+          result.push({
+            displayName: behaviorName,
+            type: componentType,
+            variations: variation,
+          });
+          cb();
+        });
```
layershifter, line 87: I simplified this part, please check that it still works

layershifter (approved):

scripts/gulp/plugins/gulp-component-menu-behaviors.ts
kolaps33: thank you very much, verified, tested, works as expected

packages/fluentui/accessibility-definitions/package.json
ecraig12345: I'm still a little confused. If the definitions are only needed as a **devDependency** of `accessibility`, they don't need to be in a published package. But if they're going to be used (directly or indirectly) by consumers of `accessibility`, the packages need to be public. Also, you can't have a private package as (non-dev) dependency of a published package, even indirectly.

kolaps33: I think "definitions" we don't want to be consume/used with accessibility package. 

Because this definitions are used only for testing of "Behaviors" files and user of "accessibility" can use only "behaviors" files. 

`Also, you can't have a private package as (non-dev) dependency of a published package, even indirectly.`
if this issue, should I revert changes and put "definitions" package back to the "rules" package and make "rules" package public?
and if it will be public do I need create separate PR for it? :)

miroslavstastny: We discussed this offline, the distinction among `a11y-testing`, `a11y-rules` and `accessibility-definitions` is hard to explain which probably means it does not have much sense. Based on that we agreed to have the two following packages:
- `@fluentui/accessibility` - PUBLIC package, contains behaviors but not their description.
- `@fluentui/a11y-testing` - private package, contains behavior descriptions, testing rules and testing framework for it.

ecraig12345:
I think the build failure is because you need to add an alias in a tsconfig somewhere for `@fluentui/a11y-testing` since it's not within `packages/fluentui`. Try adding this under `compilerOptions.paths` in `packages/fluentui/react-northstar/tsconfig.json`: `"@fluentui/a11y-testing": ["packages/a11y-testing/src"]` (and if that doesn't work I'd recommend asking Shift or Miro if they can help figure out where to add it)

ecraig12345 (approved):

jurokapsiar (approved):

PR #14970, opened 2020-09-10
fix(Menu+Carousel): prop 'active' not mapped to behavior

### LINKED ISSUE #14537: Selected Menu Item does not have "Selected" information in Northstar Menu
### Environment Information

- **Package version(s)**: @fluentui/react-northstar@0.51.0
- **Browser and OS versions**: Chrome

### Describe the issue:

Selected menu item does not have aria-selected information, or have any narration that the item is selected. Using screenreaders such as NVDA, there is no narration that the selected menu item is selected.

https://codesandbox.io/s/o5ql0?module=/example.js

#### Actual behavior:

1. Turn on NVDA
2. Focus on the selected item
3. Screen reader reads out only the menu item name

#### Expected behavior:

1. Turn on NVDA
2. Focus on the selected item
3. Screen reader should read menu item name and that it is "selected"

YuanboXue-Amber:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14537 

#### Description of changes

The original bug is reported on Menu: when a MenuItem is selected, screen-reader does not read 'selected'.
It's expected with Menu's default accessibility behavior, as there's no `aria-selected` mentioned for Menu in [aria practices](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-1/menubar-1.html)

Menu supports also `accessibility={tabListBehavior}`. And in this case, `aria-selected` should be true for selected item. But it is not.
The same problem happens to `CarouselNavigationItem`, whose default `accessibility` is `tabBehavior`.

This happens, because the `active` prop for `MenuItem/CarouselNavigationItem` is not passed to `tabBehavior`. Therefore aria-selected was not set to true for the active item.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/CHANGELOG.md
+++ b/packages/fluentui/CHANGELOG.md
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Remove the `disabled` property from `Dialog` and `Popup` behaviors @rymeskar ([#14885](https://github.com/microsoft/fluentui/pull/14885))
 
 ### Fixes
+- Fix `MenuItem` and `CarouselNavigationItem` to correctly map `active` prop to behavior @yuanboxue-amber ([#14970](https://github.com/microsoft/fluentui/pull/14970))
 - Fix default focused outline in Safari @yuanboxue-amber ([#14917](https://github.com/microsoft/fluentui/pull/14917))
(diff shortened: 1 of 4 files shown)


YuanboXue-Amber:
[Codesandbox](https://codesandbox.io/s/nifty-lichterman-i3rmm?file=/example.js) with `<Menu accessibility={tabListBehavior}`...

kolaps33 (approved):
image
great, I see issue fixed :)

assuncaocharles (approved):
LGTM

miroslavstastny (approved):

PR #14972, opened 2020-09-10
react-button: fix styles to align with v7 buttons

xugao:
#### Pull request checklist

#### Description of changes
Fit finishing styles to align with v7 button. 
Key changes are:
- Default button height to be `32px`
- Fix spacing between content
- Update menuIcon (chevron) size to be `12px`
~~- Set `flex-grow: 1` in button text content so icons will be left/right aligned when it's fluid.~~
- Fix ghost font weight
- Fix ghost background color
- Fix compound button: Icon size, spacing between text
- Fix split button divider to have the right length by default and made sure `dividerLength` token is working correctly

#### Focus areas to test
Do expect a lot of updates in screener tests


packages/react-button/src/components/Button/__snapshots__/Button.test.tsx.snap
```diff
@@ -97,6 +101,7 @@ exports[`Button renders a default state 1`] = `
     className=
 
         {
+          flex-grow: 1;
```
dzearing, line 104: This seems wrong from a design perspective, maybe check with designers. We did not have a fluid mode for v7, so this is new.

packages/react-button/src/components/CompoundButton/useCompoundButtonClasses.ts
```diff
@@ -43,7 +48,12 @@ export const useCompoundButtonClasses = makeClasses<CompoundButtonState>({
     {
       color: 'var(--button-secondaryContentColor)',
       fontSize: 'var(--button-secondaryContentFontSize)',
+      fontWeight: 'var(--button-secondaryContentFontWeight)',
       lineHeight: '100%',
+
+      '&:nth-child(2)': {
```
dzearing, line 54: i think we do `:not(:first-child)` in other places; i think that might be a little better, as it will apply to additional items if we add later. I'm not sure about perf diff.

packages/react-button/src/components/Button/__snapshots__/Button.test.tsx.snap
xugao: confirmed what we had is what we want. reverted the change.

packages/react-button/src/components/SplitButton/SplitButton.tsx
```diff
@@ -11,7 +11,7 @@ import { MenuButton } from '../MenuButton/index';
 export const SplitButton = React.forwardRef<HTMLElement, SplitButtonProps>((props, ref) => {
   const { state, render } = useSplitButton(props, ref, {
     button: { as: Button },
-    menuButton: { as: MenuButton, iconOnly: true, icon: <ChevronDownIcon /> },
+    menuButton: { as: MenuButton, menuIcon: <ChevronDownIcon /> },
```
dzearing, line 14: hmm. Was this intentional?

dzearing (approved):
check the comment first. I think there might be a snafu with menubutton.

packages/react-button/src/components/SplitButton/SplitButton.tsx
xugao: yes. in fact it should be even simpler: `menuButton: { as: MenuButton }`. updated

xugao: reverted

PR #14980, opened 2020-09-10
Support StrictMode in DetailsList component hierarchy

### LINKED ISSUE #14291: DetailsRow: strict mode compliance (componentWillReceiveProps)
DetailsRow has a `componentWillReceiveProps` implementation which updates `state.selectionState` based on props (it can also be updated elsewhere in the class). This is a case where moving the logic to `componentDidUpdate` is probably **not** a good idea due to performance concerns of the extra render, since there can be many DetailsListRows on a page and they can contain expensive custom renderers.

~~DetailsRow and DetailsHeader use `ReactDOM.findDOMNode` to find the actual HTML element corresponding to a FocusZone. This works for the time being, since FocusZone is still a class component, but it will stop working if/when FocusZone is converted to a function component. So we need to find another way to get the element.~~ Moved to #14685

### LINKED ISSUE #14295: GroupedList and GroupHeader: strict mode compliance (componentWillReceiveProps, medium)
GroupedList has a `componentWillReceiveProps` implementation which does the following:
- Copy `props.groups` to `state.groups`. `state.groups` is never set elsewhere, so the copy could be removed (and references updated) and re-rendering handled by `forceUpdate` if needed.
- Force child list/section updates if groups or certain other things change.

GroupHeader has a `componentWillReceiveProps` implementation which updates `state.isCollapsed` and `state.isLoadingVisible` based on `props.group`.

~~GroupedList and GroupedListSection also use string refs but this is tracked separately in #14290.~~ (fixed)

GroupedList is simpler than some of the other list components and *might* be a candidate for function component conversion, but keeping it as a class component and just fixing the problematic stuff would be fine.

### LINKED ISSUE #14296: List: strict mode compliance (UNSAFE_)
List has a `componentWillReceiveProps` implementation which does various updates related to paging.

~~It also uses string refs but this is tracked separately in #14290.~~ (fixed)

We plan to fix these issues in the class component rather than converting to a function component due to complexity and regression risk.

ThomasMichon:
#### Description of changes

Updated `DetailsList`, `GroupedList`, `GroupHeader`, and `List` so they comply with the requirements of `StrictMode`: namely, they no longer implement `UNSAFE_componentWillReceiveProps` and now use `getDerivedStateFromProps`.
It's important to note that this change only seeks *minimal compliance*. The components aren't guaranteed to work perfectly in concurrent mode, because they still rely on a lot of class fields to maintain state, and the implementations of `getDerivedStateFromProps` still delegate logic back to the class instance, by shuttling a bound method into the component state (this pattern was leveraged by `react-redux` before they fully moved to Hooks).

Where needed, the logic in each component has been converted so helper functions that operate on state will return the new state instead of calling `setState`. Multiple calls to `setState` now simply aggregate a new state, and usage of the `setState` post-update callback has been replaced with `componentDidUpdate`.

#### Focus areas to test

The typical logic which breaks when updating these components:
1. Resizing a column in a grouped list. This appears to be working.
2. Virtual scrolling. Appears to be working.
3. Virtual scrolling within groups. Appears to be working.
4. General rendering of the DetailsList. Appears to be working.

Fixes #14296, fixes #14295, fixes #14291

### DIFF
--- a/packages/office-ui-fabric-react/src/components/GroupedList/GroupedList.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/GroupedList/GroupedList.base.tsx
@@ -23,6 +23,7 @@ const { rowHeight: ROW_HEIGHT, compactRowHeight: COMPACT_ROW_HEIGHT } = DEFAULT_
 export interface IGroupedListState {
   lastSelectionMode?: SelectionMode;
   groups?: IGroup[];
+  version: {};
 }
 
 export class GroupedListBase extends React.Component<IGroupedListProps, IGroupedListState> implements IGroupedList {
@@ -41,6 +42,45 @@ export class GroupedListBase extends React.Component<IGroupedListProps, IGrouped
 
   private _groupRefs: Record<string, GroupedListSection | null> = {};
 
+  public static getDerivedStateFromProps(
+    nextProps: IGroupedListProps,
+    previousState: IGroupedListState,
+  ): IGroupedListState {
+    let nextState = previousState;
+
+    const { listProps: { version = undefined } = {} } = nextProps;
+
+    if (version) {
+      nextState = {
+        ...nextState,
+        version,
+      };
+    }
+
+    const { groups, selectionMode, compact } = nextProps;
+    let shouldForceUpdates = false;
+
+    if (nextProps.groups !== groups) {
+      nextState = {
+        ...nextState,
+        groups: nextProps.groups,
+      };
+    }
+
+    if (nextProps.selectionMode !== selectionMode || nextProps.compact !== compact) {
+      shouldForceUpdates = true;
+    }
+
+    if (shouldForceUpdates) {
+      nextState = {
+        ...nextState,
+        version: {},
+      };
+    }
+
+    return nextState;
+  }
+
   constructor(props: IGroupedListProps) {
     super(props);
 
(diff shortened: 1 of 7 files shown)


packages/office-ui-fabric-react/src/components/GroupedList/GroupedList.base.tsx
```diff
@@ -203,6 +217,11 @@ export class GroupedListBase extends React.Component<IGroupedListProps, IGrouped
       return null;
     }
 
+    const finalListProps: IListProps = {
+      ...(listProps || {}),
```
dzearing, line 221: nit: if you spread an object whose value is null or undefined, still works.

doesn't work for spreading arrays.

packages/office-ui-fabric-react/src/components/List/List.tsx
```diff
@@ -333,31 +342,48 @@ export class List<T = any> extends React.Component<IListProps<T>, IListState<T>>
     }
   }
 
+  public componentDidUpdate(): void {
+    // Multiple updates may have been queued, so the callback will reflect all of them.
+    // Re-fetch the current props and states to avoid using a stale props or state captured in the closure.
+    const finalProps = this.props;
```
dzearing, line 348: unclear why you need to variable-ize these. Is some magic getter logic happening? Not sure why you can't just destructure the values you need here:

```
const { pages } = this.state;
```

dzearing (approved):

packages/office-ui-fabric-react/src/components/List/List.tsx
ThomasMichon: As part of porting over the code, I did my best to remove all direct calls to `this.props` and `this.state` so I could keep track of them. This is because many function now need to take those values purely as function parameters because `this.props` could be the wrong set of values. What you see here is literally what was in the `setState` commit callback, when this code was in `componentWillReceiveProps`.

PR #14981, opened 2020-09-10
Fix markdown for prerelease

ecraig12345:
Add comments for change type `prerelease` to release notes entries.

~~Disallow change type `major` for all packages, to prevent accidental major changes. You can still do a major change by choosing `minor` on the command line then manually editing the change file.~~ (reverted pending beachball fix)

### DIFF
--- a/scripts/beachball/index.ts
+++ b/scripts/beachball/index.ts
@@ -9,6 +9,9 @@ export const config: BeachballConfig = {
       disallowedChangeTypes: ['major'],
     },
   ],
+  // This prevents accidental major changes. You can manually do a major change by selecting
+  // "minor" on the command line and then changing it to "major" by editing the change file.
+  disallowedChangeTypes: ['major'],
   changelog: {
     customRenderers: {
       renderHeader,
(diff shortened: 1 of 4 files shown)


scripts/beachball/index.ts
```diff
@@ -9,6 +9,9 @@ export const config: BeachballConfig = {
       disallowedChangeTypes: ['major'],
     },
   ],
+  // This prevents accidental major changes. You can manually do a major change by selecting
+  // "minor" on the command line and then changing it to "major" by editing the change file.
+  disallowedChangeTypes: ['major'],
```
xugao, line 14: unfortunately this won't work as is. We will either need to update beachball code (I don't think it's going to be straight forward) or add this individually in each package.json
It doesn't work because of `groups` is being defined above.

xugao:
```
Found changes in the following packages:
  office-ui-fabric-react
Cannot have a disallowedChangeType inside a package config (office-ui-fabric-react) when there is a group defined; use the groups.disallowedChangeTypes instead.
ERROR: Groups defined inside the configuration is invalid
[
  {
    name: 'Fluent UI React',
    include: [ 'packages/office-ui-fabric-react', 'packages/react' ],
    disallowedChangeTypes: [ 'major' ]
  }
]
```

xugao: i believe this is the right fix: https://github.com/microsoft/beachball/pull/403/files

PR #432, opened 2016-10-20
Update DocumentCard title font

atneik:
And removing div

PR #14982, opened 2020-09-10
[Nav][Accessibility] Selected items are not announced by narrator

### LINKED ISSUE #13160: Nav Component: Selected Information Not Convey in Edge Chromium  As well same for High light not say it is not selected
Nav Component:  selected / not selected  Information not Convey in Edge Chromium.
**Browser**
 Edge Chromium (any)

Screen Reader
Narrator 

Expected Result
When Focus Jumps to Any Nav item and if it is already selected message should Convey to Narrator in Edge Chromium  that it is selected along with text. Similar case in not selected situation. 

Actual Result
It is not convey message of Selected / not selected nav item.

image

cimoli-dev:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #13160

#### Description of changes

To get narrator to announced the selected nav item, aria-selected attributed must be added to the list item. However, that attribute is only supported on the following roles: gridcell, option, row, tab.

Because of that, change the role of the items from 'listitem' to 'treeitem' and set the value to true/false depending on if the link is selected.

However, there is a known issue in both narrator and JAWS where the selected state is not being announced. I've added a comment in the code about this.
[Aria-selected and aria-checked are not output correctly for trees #432](https://github.com/FreedomScientific/VFO-standards-support/issues/432)

image


ecraig12345 (changes requested):
I don't think `tab` is the right role here. `aria-selected` is also supported on [several other roles](https://www.w3.org/TR/wai-aria/#aria-selected) including `treeitem`, which I think would be more appropriate here. In that case the role of the `<ul>` would need to be changed to `tree`.

cimoli-dev:
@ecraig12345 I'm using tree and treeitem now. However, narrator and JAWS are not announcing the selected property. It appears to be an open issue for both ATs. Because of that i've added a comment about it near the change.

ecraig12345:
> @ecraig12345 I'm using tree and treeitem now. However, narrator and JAWS are not announcing the selected property. It appears to be an open issue for both ATs. Because of that i've added a comment about it near the change.

Thanks for the update! If I remember right from when I started to look into this issue several months ago, the problem may be that `aria-current` is not actually the correct attribute to use, and `aria-selected` should be used instead or in addition. But I'd have to do some more reading to refresh my memory on that. (Would also welcome your input if you're more familiar with the correct way to handle things here.)

Pangamma:
Checking this PR once per day as it would fix one of the bugs currently assigned to me.

khmakoto (changes requested):

Weffe:
This is still an issue and should get fixed because as the user traverses to the "selected" nav item, then the AT doesn't announce it.

PR #14983, opened 2020-09-10
allow consumers of FloatingPeopleSuggestions to use the onRenderSuggestion property

elisabethcvs:
#### Pull request checklist

#### Description of changes

IFloatingPeopleSuggestionsProps was omitting the IBaseFloatingSuggestionsProps onRenderSuggestion property, instead rendering its own default item. Users (us) may want to override how the items are rendered, so I'm removing that omission. Listing this as a patch, because that property was always optional, so now it's using the old default if the property isn't specified, so there should be no change in behavior for users.

I added an example that overrides onRenderSuggestions

### DIFF
--- a/packages/experiments/src/components/FloatingSuggestionsComposite/FloatingPeopleSuggestions/examples/FloatingPeopleSuggestions.CustomRender.Example.tsx
+++ b/packages/experiments/src/components/FloatingSuggestionsComposite/FloatingPeopleSuggestions/examples/FloatingPeopleSuggestions.CustomRender.Example.tsx
@@ -1,0 +1,169 @@
+import * as React from 'react';
+import {
+  IFloatingSuggestionItemProps,
+  FloatingPeopleSuggestions,
+  IFloatingSuggestionItem,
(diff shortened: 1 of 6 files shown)


packages/experiments/src/components/FloatingSuggestionsComposite/FloatingPeopleSuggestions/examples/FloatingPeopleSuggestions.CustomRender.Example.tsx
```diff
@@ -0,0 +1,169 @@
...
+};
+
+const getStyles = (props: ISuggestionItemExampleStylesProps): ISuggestionItemExampleStyles => {
+  const theme = getTheme();
+
+  if (!theme) {
+    throw new Error('theme is undefined or null in Editing item getStyles function.');
+  }
+
+  // const { semanticColors } = theme;
```
nebhatna, line 75: nit : remove if not needed

nebhatna (approved):

Issue #14681, opened 2020-08-21
Address circular dependencies in examples


JustSlone:
(from @ecraig12345 ) 

In a component's examples, it's common to use other components to help control the examples or demonstrate concepts. This worked fine when all the components were in the same package, but it will cause problems once we split the components into separate packages.

Currently the new `react-*` component packages such as `react-button` get around this by depending on OUFR rather than `@fluentui/react-next`, but that hack will no longer work once OUFR and `@fluentui/react-next` merge into `@fluentui/react`.

One possible fix is finishing the move to `packages/examples`, which I (Elizabeth) introduced earlier this year in #11984 to get around this issue. The idea is that each package's examples (and doc .md files) can live under paths like `packages/examples/src/<package-name>/components/<Whatever>`, and each package will have its own separate storybook to ensure start times stay reasonable. I even wrote a script to copy existing examples (still slightly buggy), but ran into some issues with build and had to move on to other things. But it should be possible to finish.

Since merging OUFR and react-next will happen in the beta prep period, we should probably get work on the circular dependency issue underway before then so that we don't get block if it takes longer than expected.

PR #15007, opened 2020-09-12
Setup for moving examples to examples package

ecraig12345:
Related to #14681 

The first commit is changes that are pretty certain and could probably just be checked in. The second one is still a work in progress.


packages/charting/webpack.serve.config.js
```diff
@@ -2,8 +2,6 @@ const path = require('path');
 const resources = require('../../scripts/webpack/webpack-resources');
 
 module.exports = resources.createServeConfig({
-  entry: './src/demo/index.tsx',
```
ecraig12345, line 5: I think this part and the change in webpack-resources were because I tried to also move legacy demo apps (except fabric-website-resources probably) to the examples package, and ran into assorted problems (don't remember the details).

If it proves to be too much of a pain, we might want to just get rid of the legacy demo apps (except fabric-website-resources and the demo app for react-button if we still need that to test polyfills).

packages/examples/.storybook/decorators/index.js
```diff
@@ -0,0 +1 @@
+export * from './withKeytipLayer';
```
ecraig12345, line 1: This and the other keytip things are copied from OUFR

packages/examples/.storybook/main.js
```diff
@@ -1,90 +1,18 @@
...
   webpackFinal: (/** @type {webpack.Configuration} */ config) => {
-    config.resolveLoader = {
-      ...config.resolveLoader,
-      modules: [
-        'node_modules',
-        path.resolve(__dirname, '../../../scripts/node_modules'),
-        path.resolve(__dirname, '../../../node_modules'),
-      ],
-    };
+    config = custom({ config, cwd: path.dirname(__dirname) });
```
ecraig12345, line 9: The custom setup that was previously here seemed mostly redundant with the shared webpack config.

cwd is used to ensure that aliases for all deps of the examples package (not just deps of the package being demo'd) are included.

packages/examples/.storybook/main.js
```diff
@@ -1,90 +1,18 @@
...
-
-    config.resolve.extensions.push('.ts', '.tsx');
-
-    config.resolve.alias = getResolveAlias();
-
-    config.plugins.push(new HardSourceWebpackPlugin(), new IgnoreNotFoundExportWebpackPlugin({ include: [/\.tsx?$/] }));
-
-    config.optimization.minimize = false;
+    config.module.rules.push({
+      // Special loader that only includes stories from the current package
```
ecraig12345, line 12: This is a hack to try and prevent it from building absolutely all files for the demo app of one particular package. I'm not sure if it was working.

packages/examples/.storybook/preview.js
```diff
@@ -3,10 +3,20 @@ import { initializeIcons } from '@uifabric/icons';
 import { configure, addParameters, addDecorator } from '@storybook/react';
 import { withInfo } from '@storybook/addon-info';
 import { withA11y } from '@storybook/addon-a11y';
+import { withPerformance } from 'storybook-addon-performance';
```
ecraig12345, line 6: These changes are copied from updates to the OUFR storybook config.

All individual packages' storybook helpers/config should be deleted as part of the final PR. The generateStoriesFromExamples helper can also be deleted.

packages/examples/.storybook/preview.js
```diff
@@ -3,10 +3,20 @@ import { initializeIcons } from '@uifabric/icons';
...
 import * as React from 'react';
 
 addDecorator(withInfo());
 addDecorator(withA11y());
+addDecorator(withPerformance);
+addDecorator(withKnobs({ escapeHTML: false }));
+addDecorator(withThemeProvider);
+addDecorator(withStrictMode);
+addDecorator(withKeytipLayer);
+
```
ecraig12345, line 19: Later in this file, take note of this part which together with preview-loader.js is supposed to help with loading only stories for the current package.
```
  // This will be updated by preview-loader with the actual current package name
  const req = require.context('../src/PACKAGE_NAME', true, /\.Example\.tsx$/);
```

packages/examples/config/tests.js
```diff
@@ -0,0 +1,15 @@
+/** Jest test setup file. */
```
ecraig12345, line 1: The test config was added because I realized the ComponentExamples snapshot test will need to move into this package.

packages/examples/src/ComponentExamples.test.tsx
```diff
@@ -94,6 +94,9 @@ function setCacheFullWarning(enabled: boolean) {
   };
 }
 
+/** Run tests on these packages' examples */
+const includedPackages = ['date-time', 'office-ui-fabric-react', 'react-cards', 'react-focus', 'react-button'];
```
ecraig12345, line 98: This list is probably out of date. Also some of the changes here may still be buggy.

packages/experiments/src/components/VirtualizedList/VirtualizedList.test.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import * as renderer from 'react-test-renderer';
-import { VirtualizedListBasicExample } from './examples/VirtualizedList.Basic.Example';
+import { VirtualizedList } from './VirtualizedList';
```
ecraig12345, line 3: Since the examples are moving to a separate package, this test within the experiments package can't render an example. So I duplicated the gist of the example below.

packages/mdl2-theme/tsconfig.json
```diff
@@ -16,7 +16,6 @@
     "skipLibCheck": true,
     "moduleResolution": "node",
     "preserveConstEnums": true,
-    "skipLibCheck": true,
```
ecraig12345, line 19: This was incorrectly specified twice (not directly related to this change)

scripts/storybook/webpack.config.js
```diff
@@ -2,22 +2,15 @@ const HardSourceWebpackPlugin = require('hard-source-webpack-plugin');
...
-    ],
-  });
-
+/**
+ * @param {Object} param0
+ * @param {webpack.Configuration} param0.config
+ * @param {string} [param0.cwd] - cwd containing package.json resolve aliases should be based on
+ * (if different from actual cwd)
+ */
+function custom({ config, cwd }) {
```
ecraig12345, line 13: Aside from the addition of the cwd option, the changes to this file are just shuffling things around (not really adding anything)

scripts/webpack/webpack-resources.js
```diff
@@ -231,6 +231,15 @@ module.exports = {
       customConfig,
     );
 
+    if (!config.entry) {
+      // Handle packages which have a legacy demo app in the examples package
```
ecraig12345, line 235: More of the hack for legacy demo apps

ecraig12345 (commented):
Some comments below explaining parts of this change (mostly still in progress).

In general, here's the approach I'd recommend:
1. Run `scripts/move-examples.js`
2. Look through files in a few different packages and see if there's anything it did dramatically wrong. If so, revert all changes, fix the script, and re-run.
3. Try to build (repeat 2 if needed)
4. repeat with start, legacy start, tests (repeat 2 if needed)

If possible it might be good to split some of these changes into separate PRs, but I'm not sure if it will be possible.

If any of the examples came from packages with less-strict lint rules, it might be a good idea to just disable the rules for the entire examples package and then enable them (and fix no-explicit-any violations) in a later PR.

PR #15021, opened 2020-09-14
Reveal Passsword button

### LINKED ISSUE #15333: Add Reveal Password button to TextField

Add a button to reveal the password similar to the one that exists in Edge/Chromium Edge and IE.

Office-UI-Fabric-React TextField

David Zearing

image

hutchcodes:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15333

#### Description of changes

Add reveal password button for browsers that don't have native reveal password (Chrome, Firefox, Safari and Opera)

#### Focus areas to test

I don't have a Mac, so it still needs to be verified on Safari.

### Other notes
Currently defaults to enable the reveal functionality with the ability to turn it off if it's not desired. It also always allows reveal, we may want to only allow revealing if the user has typed everything that is in the textfield (don't allow reveal of auto-entered passwords).

### DIFF
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -8053,6 +8053,7 @@ export interface ITextFieldProps extends React.AllHTMLAttributes<HTMLInputElemen
     autoAdjustHeight?: boolean;
     autoComplete?: string;
     borderless?: boolean;
+    canRevealPassword?: boolean;
     className?: string;
     componentRef?: IRefObject<ITextField>;
     defaultValue?: string;
@@ -8105,6 +8106,7 @@ export interface ITextFieldSnapshot {
 export interface ITextFieldState {
     errorMessage: string | JSX.Element;
     isFocused?: boolean;
+    type: string;
     uncontrolledValue: string | undefined;
 }
 
@@ -8114,6 +8116,7 @@ export type ITextFieldStyleProps = Required<Pick<ITextFieldProps, 'theme'>> & Pi
     hasIcon?: boolean;
     hasLabel?: boolean;
     focused?: boolean;
+    hasRevealButton?: boolean;
 };
 
 // @public (undocumented)
@@ -8124,6 +8127,9 @@ export interface ITextFieldStyles {
     fieldGroup: IStyle;
     icon: IStyle;
     prefix: IStyle;
+    revealButton: IStyle;
+    revealIcon: IStyle;
+    revealSpan: IStyle;
     root: IStyle;
     subComponentStyles: ITextFieldSubComponentStyles;
     suffix: IStyle;
--- a/packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx
@@ -1,8 +1,7 @@
 import * as React from 'react';
 import { IProcessedStyleSet } from '../../Styling';
 import { Label, ILabelStyleProps, ILabelStyles } from '../../Label';
-import { IconButton, IButtonProps } from '../../Button';
-import { Icon } from '../../Icon';
+import { Icon, IIconProps } from '../../Icon';
 import {
   Async,
   DelayedRender,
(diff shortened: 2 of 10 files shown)


packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -8064,6 +8064,7 @@ export interface ITextFieldProps extends React.AllHTMLAttributes<HTMLInputElemen
     prefix?: string;
     readOnly?: boolean;
     resizable?: boolean;
+    showRevealPassword?: boolean;
```
dzearing, line 8067: @ecraig12345 naming-wise, are you ok with this name? Is it consistent with any precedents? Do you have other suggestions?

dzearing: Some other ideas:

allowPasswordVisibilityToggle
canRevealPassword

packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { IProcessedStyleSet } from '../../Styling';
 import { Label, ILabelStyleProps, ILabelStyles } from '../../Label';
+import { IconButton, IButtonProps } from '../../Button';
```
dzearing, line 4: This unfortunately pulls a lot of things in. I think we may want to use a lowercase `<button/>` here with an icon and do minimal styling.

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
hutchcodes: IE/Edge do call it Reveal Password internally. canRevealPassword is better than what I have.

khmakoto (changes requested):

hutchcodes:
I moved this PR over to the V7 branch as it is initially for use in the Azure Portal which is currently using V7, and we're not sure what the timeframe for updating to v8 would be.

packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx
```diff
@@ -567,6 +599,36 @@ export class TextFieldBase extends React.Component<ITextFieldProps, ITextFieldSt
...
     }
   }
+
+  private _shouldShowRevealButton = (): boolean => {
+    if (typeof window === 'undefined' || !window.navigator || !window.navigator.userAgent) {
+      return false;
+    }
+
+    if (this.props.canRevealPassword) {
+      const userAgent = window.navigator.userAgent;
```
dzearing, line 609: hmm. should this be an exception list where this is functional by default?

dzearing: I'm having a hard time understanding why we should conditionally only allow the behavior on a specific list and why we wouldn't allow it otherwise?

dzearing:
It would be good to add an example to use a `TextField` as a password with `canRevealPassword` prop on! Then users can understand how it works.

packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx
hutchcodes: If a new browser comes along that isn't handled, I was thinking it would be a better experience for the end user if the reveal password button was not shown than to potentially have a browser's built in reveal password and this additional reveal password button.

Basically, I'm looking for the least harm way this would fail.

hutchcodes: Switched this to only not show the reveal button on IE and Edge

dzearing (approved):

faizal3199:
Not sure if this is the right place to mention this. (Let me know if I need to create a new issue)

https://github.com/microsoft/fluentui/blob/62b9b2e5daaf345097f6e2382896d64ed6c18e05/packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx#L488-L493

The line 492 line breaks the expected behavior of the component (whose initial `type` was `password`) when the `type` of component is updated by it's parent through props. This should only be considered when `this.props.canRevealPassword` is set to `true`.

CC: @hutchcodes @dzearing

ecraig12345:
@faizal3199 I agree that looks like a bug. Can you file a new issue about it?

faizal3199:
Sure, I will create a new issue. I'd be happy to make PR to fix this, my changes depends on this.😅

PR #15022, opened 2020-09-14
Drag and drop fixes

elisabethcvs:
#### Pull request checklist

#### Description of changes

- Check on drop that we are not dropping over the selected items. Previously we were not checking and this was causing items to be duplicated when dropped over themselves
- Remove items when they were successfully dropped into another well

Tested that the move works with single and multiple items. Also tested that drag and drop within wells is unaffected.

### DIFF
--- a/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
+++ b/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
@@ -155,7 +155,7 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
   };
 
   const _onDragEnd = (item?: any, event?: DragEvent): void => {
-    if (event && event.dataTransfer?.dropEffect == 'move') {
+    if (event && event.dataTransfer?.dropEffect === 'move') {
       const itemstoRemove = focusedItemIndices.includes(draggedIndex)
         ? (getSelectedItems() as T[])
         : [selectedItems[draggedIndex]];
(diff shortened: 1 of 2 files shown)


packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -150,6 +155,12 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
   };
 
   const _onDragEnd = (item?: any, event?: DragEvent): void => {
+    if (event && event.dataTransfer?.dropEffect === 'move') {
+      const itemstoRemove = focusedItemIndices.includes(draggedIndex)
```
nebhatna, line 159: camelCase please :)

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -150,6 +155,12 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
+        ? (getSelectedItems() as T[])
+        : [selectedItems[draggedIndex]];
+      _onRemoveSelectedItems(itemstoRemove);
+    }
```
nebhatna, line 163: can we combine the 2 if statements for clarity

nebhatna (changes requested):
minor feedback

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -150,16 +155,22 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
   };
 
   const _onDragEnd = (item?: any, event?: DragEvent): void => {
-    setDraggedIndex(-1);
     if (event) {
-      const dataList = event?.dataTransfer?.items;
+      if (event.dataTransfer?.dropEffect === 'move') {
```
nebhatna, line 159: do we want to check for event not being null ?

nebhatna (approved):

PR #15027, opened 2020-09-15
Add missing ref prop typing

### LINKED ISSUE #14974: Fix `ref` prop typing
Make sure `ref` typing is correct for all v8 components

xugao:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14974

#### Description of changes
- Add missing ref prop typing for v8 components
- Create and merge refs in Persona, PersonaPresence component
- Add missing `as` prop for `Link`
- Some refactoring related to typing:
     - Always assign `React.FunctionComponent<IComponentProps>` typing to Component and BaseComponent. that way the typing for these components is cleaner for consumer. 
     - Pass generic types to `forwardRef`

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-checkbox-2020-09-14-17-37-05-xgao-fix-ref-typing.json
+++ b/change/@fluentui-react-checkbox-2020-09-14-17-37-05-xgao-fix-ref-typing.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Add missing ref prop typing.",
+  "packageName": "@fluentui/react-checkbox",
+  "email": "xgao@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-09-15T00:36:31.365Z"
+}
--- a/change/@fluentui-react-checkbox-2020-09-14-17-37-05-xgao-fix-ref-typing.json
+++ b/change/@fluentui-react-checkbox-2020-09-14-17-37-05-xgao-fix-ref-typing.json
@@ -1,5 +1,5 @@
 {
-  "type": "patch",
+  "type": "minor",
   "comment": "Add missing ref prop typing.",
   "packageName": "@fluentui/react-checkbox",
   "email": "xgao@microsoft.com",
--- a/packages/react-checkbox/src/components/Checkbox/Checkbox.base.tsx
+++ b/packages/react-checkbox/src/components/Checkbox/Checkbox.base.tsx
@@ -6,130 +6,132 @@ import { Icon } from 'office-ui-fabric-react/lib/Icon';
 
 const getClassNames = classNamesFunction<ICheckboxStyleProps, ICheckboxStyles>();
 
-export const CheckboxBase = React.forwardRef((props: ICheckboxProps, forwardedRef: React.Ref<HTMLDivElement>) => {
-  const {
-    disabled,
-    inputProps,
-    name,
-    ariaLabel,
-    ariaLabelledBy,
-    ariaDescribedBy,
-    ariaPositionInSet,
-    ariaSetSize,
-    title,
-    label,
-    checkmarkIconProps,
-    styles,
-    theme,
-    className,
-    boxSide = 'start',
-  } = props;
-
-  const id = useId('checkbox-', props.id);
-
-  const rootRef = React.useRef<HTMLDivElement | null>(null);
-  const mergedRootRefs: React.Ref<HTMLDivElement> = useMergedRefs(rootRef, forwardedRef);
-  const inputRef = React.useRef<HTMLInputElement>(null);
-
-  const [isChecked, setIsChecked] = useControllableValue(props.checked, props.defaultChecked, props.onChange);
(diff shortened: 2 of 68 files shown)


packages/react-checkbox/src/components/Checkbox/Checkbox.base.tsx
```diff
@@ -6,130 +6,132 @@ import { Icon } from 'office-ui-fabric-react/lib/Icon';
...
-      title,
-      ref: mergedRootRefs,
-    },
-    input: {
-      className: classNames.input,
-      type: 'checkbox' as React.InputHTMLAttributes<HTMLInputElement>['type'],
-      ...inputProps,
-      ref: inputRef,
-      checked: !!isChecked,
+export const CheckboxBase: React.FunctionComponent<ICheckboxProps> = React.forwardRef<HTMLDivElement, ICheckboxProps>(
```
ecraig12345, line 9: It might be worth adding a lint rule for this (as well as for component naming). The trick is how to figure out in a lint rule that something is a component file, but I have ideas for that, and we can always disable the rule in specific files where the heuristic gives wrong results. (We can make an issue to track this rather than doing it within the PR.)

change/@fluentui-react-checkbox-2020-09-14-17-37-05-xgao-fix-ref-typing.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
ecraig12345, line 2: Maybe minor since it's adding an API? (same in other published packages)

ecraig12345 (approved):
This is great! I like that we're getting rid of the ForwardRefExoticComponent stuff from the public API. It also happens that these changes will make upgrading API Extractor easier whenever we want to do that.

packages/react-checkbox/src/components/Checkbox/Checkbox.base.tsx
MLoughry, line 9: I'm wondering why move the parameter typings to the generic arguments, rather than leave them on the parameters? It seems less readable this way. (Same everywhere)

packages/react-checkbox/src/components/Checkbox/Checkbox.tsx
```diff
@@ -4,8 +4,8 @@ import { CheckboxBase } from './Checkbox.base';
 import { getStyles } from './Checkbox.styles';
 import { ICheckboxProps, ICheckboxStyleProps, ICheckboxStyles } from './Checkbox.types';
 
-export const Checkbox = styled<
-  ICheckboxProps & React.RefAttributes<HTMLDivElement>,
+export const Checkbox: React.FunctionComponent<ICheckboxProps> = styled<
```
MLoughry, line 7: I ran into this issue a few times during conversion, but why add explicit typings and generic arguments when they can be inferred? Several styled versions of components I converted to ref-forwarding components lost the `ref` attribute *precisely because* of explicit typings like these.

MLoughry (commented):
> I like that we're getting rid of the ForwardRefExoticComponent stuff from the public API.

I'm not sure why this matters, let alone is a "good" thing? Why not standardize on things?

packages/react-checkbox/src/components/Checkbox/Checkbox.base.tsx
ecraig12345: I think this may be a stylistic preference--it's more redundant with the types specified on both sides. You can still see the type if desired using intellisense.

packages/react-checkbox/src/components/Checkbox/Checkbox.tsx
ecraig12345: I think the cleanliness improvements to the public API make it worthwhile, and we can help prevent `ref` from getting lost in the future by adding a lint rule.

This also happens to make upgrading to the latest API Extractor version (which we'll eventually need to do) easier because the next minor version beyond where we are right now has decided that inferred types which include `import()` types are "unsupported"...even though it seemed to do fine with them in the past.

xugao: The reason is to have cleaner/simpler typing:
image

Also, you should not have any problem going forward because now we have `ref` typing added to `IComponentProps` interfaces

packages/react-checkbox/src/components/Checkbox/Checkbox.base.tsx
xugao: it's less to type when using the generic type. and I don't think there is a readability issue with either way of doing typing. Also, say if later react decides to change the typing for the `React.forwardRef` signature, using the generic types helps the such change to be less disruptive for our code.

xugao: added an issue [here](https://github.com/microsoft/fluentui/issues/15048). pls update if i noted anything incorrectly

PR #15038, opened 2020-09-15
chore: refactor project-tests

layershifter:
This PR refactors `@fluentui/projects-test`.

### What changed?
- Gulp is not used anymore for these tests
  - `packages/fluentui/projects-test/gulpfile.ts` was removed
  - assets in `scripts/gulp/tasks/test-projects` were moved to the package's directory
- if tests will fail a warning that points to `packages/fluentui/projects-test/README.md` will be raised
- `packages/fluentui/projects-test/README.md` was updated to contain FAQ
- only necessary packages now packed, they captured via dependency graph of `@fluentui/react-northstar`


packages/fluentui/projects-test/README.md
```diff
@@ -1,3 +1,14 @@
...
+
+```
+./node_modules/@fluentui/react-northstar/dist/es/components/Datepicker/DatepickerCalendar.js
+Attempted import error: 'getEndDateOfWeek' is not exported from '@fluentui/date-time-utilities'.
+```
+
+Tests in this package consider all packages outside `packages/fluentui` as third party dependencies. If you met such issue please split your changes into two separate PRs: one PR will contain changes to `@fluentui/date-time-utilities`, for example, and another to `@fluentui/react-northstar`. Once your changes to `@fluentui/date-time-utilities` will be published (packages outside of `packages/fluentui` are releasing daily) you will be able to use them in the second PR and pass the test.
+
+As packages inside `packages/fluentui` have a different release cycle these is a high risk to ship broken code. For example, your changes will be shipped in `@fluentui/date-time-utilities@2.0.0` on next day, release will be done by `beachball` and it will bump version to `2.0.0`. However, today it will be still `1.0.0`, this means that if will release `@fluentui/react-northstar` it will have a dependency on `@fluentui/date-time-utilities@1.0.0` and we will get a broken release.
```
pompomon, line 14:
```suggestion
As packages inside `packages/fluentui` have a different release cycle which could lead to shipping a broken code. For example, your changes will be shipped in `@fluentui/date-time-utilities@2.0.0` on next day, release will be done by `beachball` and it will bump version to `2.0.0`. However, today it will be still `1.0.0`, this means that if we release `@fluentui/react-northstar` it will have a dependency on `@fluentui/date-time-utilities@1.0.0` and we will get a broken release.
```

packages/fluentui/projects-test/src/packPackages.ts
```diff
@@ -0,0 +1,71 @@
...
+    requiredPackages.map(async packageName => {
+      const filename = path.join(tmpDirectory, path.basename(packageName)) + '.tgz';
+      const packagePath = projectPackages.find(pkg => pkg.name === packageName).location;
+
+      await sh(`yarn pack --filename ${filename}`, packagePath);
+      packedPackages[packageName] = filename;
+    }),
+  );
+
+  logger(`✔️ Packages were packages to ${tmpDirectory}`);
```
pompomon, line 68:
```suggestion
  logger(`✔️ Packages were packaged to ${tmpDirectory}`);
```

pompomon (approved):

s.js
```diff
@@ -0,0 +1,16 @@
+const { getPackages } = require('@lerna/project');
```
ecraig12345, line 1: What's this file? Maybe a leftover or ended up in the wrong place?

packages/fluentui/projects-test/package.json
```diff
@@ -7,14 +7,26 @@
     "@fluentui/react-northstar": "^0.51.0"
   },
   "devDependencies": {
+    "@lerna/project": "^3.15.0",
```
ecraig12345, line 10: Not super important but if there are any of these deps which are no longer used in `scripts`, would you mind removing them there?

packages/fluentui/projects-test/README.md
```diff
@@ -1,3 +1,14 @@
-For now, all the project tests files reside in `scripts/gulp/tasks/test-projects`. This project serves as a unit of
-modularity, so builds can optimize based on package scripts. In the future, we'll be moving those tests into
-this package.
+The purpose of tests in this package is to ensure that `@fluentui/react-northstar` and its subpackages can be consumed without issues in different environments and bundlers.
```
ecraig12345, line 1: Now that this package isn't using gulp, it could potentially be moved under `/apps` since that's where we've traditionally put test-runner projects like this. (Or leaving it here for now is fine if you prefer.)

ecraig12345 (commented):
Looking good! A few minor comments.

s.js
layershifter: That's cruft, good catch

packages/fluentui/projects-test/package.json
layershifter: Actually `@lerna/project` & `@lerna/package-graph` are new dependencies. They are used to populate required dependencies. I've checked, only `tmp` can be removed from `scripts/package.json`, that's done

packages/fluentui/projects-test/README.md
layershifter: For now I would prefer to keep it in `packages/fluentui` as it also uses a separate Lerna config (`packages/fluentui/lerna.json`)

ecraig12345 (approved):

PR #15039, opened 2020-09-15
Factor out new 'dom-utilities' package to share layers logic between Fabric + Fluent Northstar

layershifter:
#### Description of changes

Introduced a new `@fluentui/dom-utilities` package to contain the `dom` sub-group from `@uifabric/utilities`. These functions handle the 'virtual' parent hierarchy functions which account for logic that uses React Portals, such as `Layer` and `Popup`. Updated `@uifabric/utilities` to depend on this new package and re-export all migrated functions from their new locations.

#### Focus areas to test

Existing unit tests should cover the migration of DOM utilities to the new package, since they are simply re-exported.

---

This PR partially extracts changes done by @ThomasMichon in #14272.


dzearing (approved):

ecraig12345:
The new package is missing an API Extractor config (`config/api-extractor.json`) which would be good to have to help prevent unexpected changes.

PR #15049, opened 2020-09-15
Update Avatar for visual spec

behowell:
#### Pull request checklist

#### Description of changes

Update Avatar for visual spec.
Add support for active ring/shadow/glow.
Improve Avatar.stories to better display all of the avatar properties.


packages/react-avatar/src/components/Avatar/useAvatar.tsx
```diff
@@ -30,33 +76,36 @@ export const useAvatar = (props: AvatarProps, ref: React.Ref<HTMLElement>, defau
...
       ref: useMergedRefs(ref, React.useRef(null)),
-      tokens: props.size && { size: `${props.size}px` },
     },
     defaultProps,
     resolveShorthandProps(props, avatarShorthandProps),
   );
 
-  // Display the initials if the label has no children
+  // Add in props used for styling
+  mergeProps(state, calcStyleProps(state));
```
dzearing, line 89: mergeProps takes in N arguments, so, i'm wondering if we should make mergeProps do something like:

mergeProps(
  { ...defaultStuff },
  { ...parentDefaultStuff },
  { ...userStuff },
  state => ({ ...dyanmically calculated stuff based on the state so far })
);
```

I don't think it would change the perf footprint much, but cases where we want to merge additional state based on state seems like it should be a little more straightforward.

dzearing (approved):

packages/react-avatar/src/components/Avatar/useAvatar.tsx
behowell: Yeah that could be useful. I haven't looked into the internals of mergeProps, but does it merge each of the arguments fully before going onto the next? Or does it merge one **prop** at a time, in which case the state wouldn't be fully merged before being passed into the callback?

PR #15053, opened 2020-09-15
Add new package for spinbutton v8

tomi-msft:
#### Pull request checklist

#### Description of changes

This PR is the first stage in updating the SpinButton component to use button v8. It creates a package that will hold a new version of SpinButton that will use button v8. This is to reduce the number of breaking changes in the button integration work.

#### Focus areas to test

(optional)


packages/react-spinbutton/.eslintrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": ["plugin:@fluentui/eslint-plugin/node"],
```
ecraig12345, line 2: This shouldn't be `node`, it should probably be `react` (and then disable the naming-convention rule for any compat interface names which already start with I)

packages/react-spinbutton/jest.config.js
```diff
@@ -0,0 +1,10 @@
+const { createConfig, resolveMergeStylesSerializer } = require('@uifabric/build/jest/jest-resources');
+const path = require('path');
+
+const config = createConfig({
```
ecraig12345, line 4: You're missing config/tests.js and config/api-extractor.json (both of these can be copied from another project, though I'm about to update the API extractor config--so you might need to wait on that)

packages/react-spinbutton/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@fluentui/react-spinbutton",
+  "version": "0.1.0",
+  "description": "Package to hold SpinButton using button v8",
+  "private": true,
```
ecraig12345, line 5: Probably makes sense for this to remain private until you add files, but once you add the files you'll need to remove this line

packages/react-spinbutton/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "main": "lib/index.js",
```
ecraig12345, line 6: This isn't quite right--it looks like you chose the `node` template in `create-package` rather than `react`. It might be best to just rename the package folder to some temp name, re-run `create-package` with `react` template, copy over any files you manually modified, and delete the old version.

khmakoto:
You should also add the package to the `CODEOWNERS` file with the correct owners.

ecraig12345:
> You should also add the package to the `CODEOWNERS` file with the correct owners.

Good point, we could potentially add a step for this in the create-package script as well. (add a prompt for the github usernames of codeowners)

dzearing (approved):

PR #15056, opened 2020-09-15
Implement dialog boundary clamping so that it doesn't go off-screen

### LINKED ISSUE #12565: Dialogs/Modal can move off-screen via keyboard
### Environment Information

- **Package version(s)**: All

Using the Blocking Dialog example on your test page (https://developer.microsoft.com/en-us/fluentui/#/controls/web/dialog). If you check the IsBlocking box and open the dialog, you can open the keyboard only menu using Ctrl+Alt+Space (you can see this implemented in the underlying Modal code). If you select "move" from those options you can then use the arrow keys to move the dialog. 

#### Actual behavior:

There is nothing stopping a user from moving a dialog completely offscreen. The user could then exit keyboard move and not be able to see any of the content and become lost.

#### Expected behavior:

We restrict moving the dialog off screen.

Requested priority:  Normal

Products/sites affected: Office Online

Dalimil:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #12565

#### Description of changes

Adds a prop `keepInBounds` to IDragOptions that can be optionally set to prevent dragging a dialog off-screen. It will allow the movement but return the dialog to the clamped position on release.

ezgif-4-8ad76729546f


Dalimil:
Demo video: 
[Storybook - Google Chrome 2020-09-15 16-26-40.zip](https://github.com/microsoft/fluentui/files/5228702/Storybook.-.Google.Chrome.2020-09-15.16-26-40.zip)

Dalimil:
@joschect Thoughts?

packages/office-ui-fabric-react/src/components/Modal/Modal.base.tsx
```diff
@@ -307,6 +313,51 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
     }
   }
 
+  private _registerInitialModalPosition = (): void => {
+    if (this.props.dragOptions?.keepInBounds && !this._minClampedPosition && !this._maxClampedPosition) {
+      const dialogMain = document.getElementsByClassName('ms-Dialog-main');
```
joschect, line 318: I think if we can it would be better to use a ref here.

joschect: There could be, although perhaps ill-advised, multiple dialogs in the window at the same time

Dalimil: I have replaced that now with data-id attr, since we cannot use the ref here.

joschect (approved):

PR #15059, opened 2020-09-16
Updating SpinButton to handle onChange

### LINKED ISSUE #5326: SpinButton: Add onChange handler
Make SpinButton useful by adding an OnChange handler.
As it is right now you have to figure out some way to accomplish the same effect using a combination of onIncrement, onDecrement, and onValidate which is really frustrating because it makes useful features of the control, like setting max and min values, unavailable since you have to manage the control yourself.

SpinButton

Right now our team is trying to use a work-around of using TextField set with ```type="number"```, but this doesn't include the nice accessibility aria attribute handling of SpinButton and is missing the useful max/min value clamping.

I noticed the private _onChange method in the component has a comment saying there was a bug in React preventing onChange from being implemented. I assume that issue is no longer a blocker given that the issue is marked as closed and TextField has an onChange handler.

This affects VSTS.

Thanks!

### LINKED ISSUE #14480: SpinButton: fix issues revealed by eslint-plugin-react-hooks
#14478 enables `eslint-plugin-react-hooks` for react-next. There were a bunch of issues in SpinButton which would require a good bit of work to fix. I disabled the rule for the file for now but need to go back and actually fix things before release.

czearing:
#### Pull request checklist
- [x] Fixes #14480, fixes #5326

#### Description of changes
1. Updating `SpinButton` to handle onChange.
2. Resolving issues revealed by eslint-plugin-react-hooks. https://github.com/microsoft/fluentui/issues/14480
3. Updating `SpinButton.stateful.example.tsx` to show the usage of onChange.
4. Adding a forwardRef.

#### Focus areas to test
`SpinButton`


packages/react-next/src/components/SpinButton/examples/SpinButton.Stateful.Example.tsx
```diff
@@ -42,18 +42,26 @@ const onSpinButtonValidate = (value: string) => {
...
-      onValidate={onSpinButtonValidate}
-      onIncrement={onSpinButtonIncrement}
-      onDecrement={onSpinButtonDecrement}
-      incrementButtonAriaLabel={'Increase value by 2'}
-      decrementButtonAriaLabel={'Decrease value by 2'}
-    />
-  </div>
-);
+export const SpinButtonStatefulExample: React.FC = () => {
+  const [spinButtonValue, setSpinButtonValue] = React.useState<string>('7' + suffix);
```
dzearing, line 46: please add a unit test which covers controlled case so we don't regress this feature again.

ecraig12345 (changes requested):
Let's wait on this until the worst of the code moves are over.

PR #15063, opened 2020-09-16
chore: add image preload for Screener in N*

layershifter:
This PR adds `<link preload>` via Webpack plugin to ensure that they will be loaded for Screener in N*.

image


layershifter:
1

scripts/webpack/webpack.config.ts
```diff
@@ -98,6 +99,10 @@ const webpackConfig: webpack.Configuration = {
         reactVis: require('react-vis/package.json').version,
       },
     }),
+    new ResourcePreloadWebpackPlugin({
```
ecraig12345, line 102: Can you add a comment about why this was needed?

layershifter: Sure, I am also thinking to run this only for Screener builds

layershifter:
Closing for housekeeping.

PR #15065, opened 2020-09-16
feat(UI builder): Integrate screen reader narration computation capability into the UI builder

adamsamec:
#### Description of changes
This is an  integration of the screen reader narration computation code into the Ui builder.

When user selects a component added to the canvas, this narration appears below the canvas. The narration tells exactly what a screen reader would read upon focusing the element.

So far not all element types are supported by the computation, only the most important ones, namely:

* &lt;a&gt;
* &lt;button&gt;
* &lt;textarea&gt;
* &lt;input type="text"&gt;
* &lt;input type="checkbox"&gt; and role="checkbox"
* &lt;input type="radio"&gt; and role="radio"
* &lt;select&gt;
* role="menuitem", role="menuitemcheckbox" and role="menuitemradio"
* role="option"
* role="combobox"
* role="switch"
* role="tab"

Also, currently, only  immitation of the computation for the JAWS screen reader is supported, and the narration reflects the state when the element is not focused using the virtual PC cursor but using the Tab key.

##### Note
This PR is a second try to integrate the screen reader narration computation into the UI builder. The [first PR](https://github.com/microsoft/fluentui/pull/14952) has been closed due to the renaming of the original branch. Also, in this PR, some review comments from the first PR has been addressed so this is not a mere renaming of the first PR.


packages/fluentui/react-builder/src/components/ComponentTree.tsx
```diff
@@ -78,7 +78,6 @@ const jsonTreeToTreeItems: (
     onTitleClick: handleSelectedComponent,
     id: tree.uuid as string,
     title: {
-      children: titleRenderer,
```
assuncaocharles, line 81: Is this intentional?

assuncaocharles, line 81: Is this intentional?

adamsamec: It's a fix from Lukas to make the component tree accessible again.

assuncaocharles: So I guess we also need to remove it from the props.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,268 @@
...
+    position: '',
+    usage: '',
+  };
+
+  // Computes and returns the complete screen reader narration for the given element and platform.
+  async computeNarration(element: Element, platform: string): Promise<string> {
+    let definitionName = this.getDefinitionName(element, platform, 'stateRules');
+
+    // Retrieve the computed accessible node
+    const node = await (window as any).getComputedAccessibleNode(element);
```
assuncaocharles, line 26:
```suggestion
    const node = await (window as Window).getComputedAccessibleNode(element);
```
Maybe this is better type?

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,268 @@
...
+  getDefinitionName(element: Element, platform: string, definitionType: string): string {
+    // Determine the definitions source by the definition type
+    let definitions;
+    if (definitionType === 'readingOrder') {
+      // Begin if 1
+      definitions = SRNC.readingOrder[platform];
+    } else if (definitionType === 'stateRules') {
+      // else if 1
+      definitions = SRNC.stateRules[platform];
+    } // End if 1
```
assuncaocharles, line 51:
```suggestion
    const definitionTypes = {
        readingOrder: SRNC.readingOrder[platform],
        stateRules: SRNC.stateRules[platform]
    }
    let definitions = definitionTypes[definitionType];
```

Not sure how big `definitionTypes` can grow so I guess avoiding the if/else would be good thing here.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,268 @@
+/*
+ TODO:
+* Add the missing and not so obvious attributes (e.g. "aria-haspopup" or "aria-expanded") for the already defined roles (e.g. "menuitem" or "checkbox") according to the specification.
+* With JAWS, in the case of the element with the "listbox" role, differentiate between having and not having the aria-multiselectable="true" attribute. If this attribute is present, then aria-selected="false" on the child elements with the role "option" behave differently than if it is not present. Specifically, if aria-multiselectable="true" is present, the aria-selected="false" causes the narration of "not selected", but if present, having the aria-selected attribute makes no difference to the narration. 
+ * Should we also consider the "disabled" state?
+ */
+import SRNC from './SRNC-Definitions';
+import './SRNC-Rules-Win_JAWS';
+
+export default class NarrationComputer {
```
assuncaocharles, line 10:
```suggestion
export class NarrationComputer {
```
I think we have an eslint rule to  prefer named exports

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,268 @@
...
+  private computedParts: Record<string, string> = {
```
assuncaocharles, line 11:
```suggestion
  computedParts: Record<string, string> = {
```
[avoid visibility modifiers (public, protected, private)](https://github.com/microsoft/fluentui/blob/master/packages/eslint-plugin/README.md#no-visibility-modifiers)

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,268 @@
...
+  computeTypeAndState(node: any, definitionName: string, element: Element, platform: string) {
+    // Set the default ttype and state for unknown roles and element types
+    this.computedParts.type = `[${node.role}]`;
+    this.computedParts.state = '';
+
+    // Find the rule which matches the states that are present on the element
+    const rules = SRNC.stateRules[platform][definitionName];
+    if (rules) {
+      // Begin if 1
+      rulesLoop: for (let i = 0; i < rules.length; i++) {
```
assuncaocharles, line 180:
```suggestion
      for (let i = 0; i < rules.length; i++) {
```

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,268 @@
...
+        // Begin for 1
+        const rule = rules[i];
+        const possibleStates = SRNC.possibleStates[definitionName];
+        for (let j = 0; j < possibleStates.length; j++) {
```
assuncaocharles, line 184: I feel like this could be a `possibleStates.forEach`?

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,268 @@
...
+        let order;
+
+        // If there is just one or no state in the combination list, the order does not have to be specified, an therefore the combination can be used as the order. But if the order is specified explicitly, use that order
+        if (rule.combination.length <= 1 && !rule.order) {
+          // Begin if 2
+          order = rule.combination;
+        } else {
+          // else if 2
+          order = rule.order;
+        } // End if 2
```
assuncaocharles, line 222:
```suggestion
        const order = rule.combination.length <= 1 && !rule.order ? rule.combination : rule.order;

        // If there is just one or no state in the combination list, the order does not have to be specified, an therefore the combination can be used as the order. But if the order is specified explicitly, use that order
```

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+        // Determine the state narration for each state in the order list by looking if corresponding state and its value exist in the state strings definitions
+        for (let j = 0; j < order.length; j++) {
```
assuncaocharles, line 225: `order.forEach`?

packages/fluentui/react-builder/src/components/SRNC-Definitions.tsx
```diff
@@ -0,0 +1,221 @@
+const SRNC: { [key: string]: any } = {
```
assuncaocharles, line 1:
```suggestion
const SRNC: Record<string, any> = {
```

assuncaocharles (commented):
Just left few comments, ping me if you need anything :)

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
adamsamec: The `getComputedAccessibleNode()` method is an experimental feature. When i change the type to `Window`, I get the following error: `Property 'getComputedAccessibleNode' does not exist on type 'Window'.`.

packages/fluentui/react-builder/src/components/ComponentTree.tsx
adamsamec: Okay, the change has been merged to the upstream, so nothing to change in this PR>

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
assuncaocharles: @miroslavstastny or @layershifter Can we do some global module augmentation for Window here or just go with `any` anyway?

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
adamsamec: I don't expect it to grow but I can apply this suggestion for the simplicity sake.

adamsamec: So the compiler complains:

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ readingOrder: any; stateRules: any; }'.
  No index signature with a parameter of type 'string' was found on type '{ readingOrder: any; stateRules: any; }'.  TS7053
```
 
I had to change it to:

```
const definitionTypes: Record<string, any> = {
```

assuncaocharles: I am good with it :)

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
adamsamec: okay, thanks, fixed.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,268 @@
...
+          const elementHasCheckedProp =
+            possibleState === 'checked' && (element as HTMLInputElement).checked !== undefined;
+
+          const elementHasStateOrCheckedProp = elementHasState || elementHasCheckedProp;
+          const combinationHasState = rule.combination.includes(possibleState);
+
+          // Check if the presence of the state on the element matches its presence in the combination list. If not, continue with the next rule
+          if (elementHasStateOrCheckedProp !== combinationHasState) {
+            // Begin if 2
+            continue rulesLoop;
```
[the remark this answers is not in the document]
adamsamec: Such a modification will change the behavior of the loop. The continue statement must continue with the outer loop, not the inner one.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
adamsamec: Okay, can I ask why they should be removed?

packages/fluentui/react-builder/src/components/SRNC-Definitions.tsx
adamsamec: Yes, that's more readable.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
adamsamec: See below my comment on the `continue` statement.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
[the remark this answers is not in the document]
adamsamec: Sorry, in case `forEach()` is used, we would get rid of the second loop and could remove the rulesLoop label.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
assuncaocharles: https://github.com/microsoft/fluentui/blob/master/packages/eslint-plugin/README.md#no-visibility-modifiers

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
assuncaocharles: The lint is complaining about the label

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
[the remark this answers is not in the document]
adamsamec: So I was not correct. The `continue` statement will not work inside the `forEach` callback. Is it a big problem to use a labelled loop?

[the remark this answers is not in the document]
assuncaocharles: It looks to e breaking the build on lint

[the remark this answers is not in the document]
adamsamec: So this is resolved. I have replaced the innerr loop that has the problematic `continue rulesLoop` statement with a `Array.some()` loop loop.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
adamsamec: As mentioned above, I have used `Array.some()` instead of `Array.forEach()`

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
layershifter: I don't see any definitions for `getComputedAccessibleNode()` in [lib.dom.d.ts](https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts), so any is a single thing :(

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
adamsamec: Yes, applied, resolving.

assuncaocharles (commented):
Why adding `northstar` file?

packages/fluentui/react-builder/src/components/Canvas.tsx
```diff
@@ -9,7 +9,7 @@ import { DebugFrame } from './DebugFrame';
 import { DropSelector } from './DropSelector';
 import { ReaderText } from './ReaderText';
 
-const showNarration = false;
+const showNarration = true;
```
assuncaocharles, line 12:
```suggestion
```

assuncaocharles: After applying this lines 229 and 337 should be adjusted.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,262 @@
...
+      // The definition name is determined by the "role" attribute and the definition exists
+      if (typeof definition === 'string') {
+        // Begin if 2
+        // The definition is a reference to another definition
+        definitionName = definition;
+      } else {
+        // else if 2
+        // The definition is a regular definition (not a reference to another definition)
+        definitionName = testName;
+      } // End if 2
```
assuncaocharles, line 66: Can we make it a ternary for simplicity? I think it can be better to read.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,262 @@
...
+        // The definition exists
+        if (typeof definition === 'string') {
+          // Begin if 3
+          // The definition is a reference to another definition
+          definitionName = definition;
+        } else {
+          // else if 3
+          // The definition is a regular definition (not a reference to another definition)
+          definitionName = testName;
+        } // End if 3
```
assuncaocharles, line 89: Here also

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,262 @@
...
+    if (definitionName === 'textarea' && platform === 'Win/JAWS' && value) {
+      // Begin if 1
+      this.computedParts.description = SRNC.stateStrings['Win/JAWS']['textarea']['[containsText]'];
+    } else {
+      // else if 1
+      this.computedParts.description =
+        element
+          .getAttribute('aria-describedby')
+          ?.split(/\s+/)
+          .map(x => {
```
assuncaocharles, line 135: What is `x`?

assuncaocharles:
Overall looks good to me. @pompomon and @layershifter can you also have a look?

packages/fluentui/react-builder/src/components/Canvas.tsx
```diff
@@ -226,7 +224,7 @@ export const Canvas: React.FunctionComponent<CanvasProps> = ({
           `
           [data-builder-id="builder-root"] {
             ${isExpanding ? `padding: ${debugSize};` : ''}
-            min-height: ${showNarration ? 'calc(100vh - 1.5rem)' : '100vh'};
+            min-height: calc(100vh - 1.5rem);
```
assuncaocharles, line 227:
```suggestion
            min-height: 'calc(100vh - 1.5rem)';
```

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
adamsamec: yes,yes, applied.

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
adamsamec: Renamed to :id: and added the type.

miroslavstastny (approved):

packages/fluentui/react-builder/src/components/NarrationComputer.tsx
```diff
@@ -0,0 +1,248 @@
+/*
```
layershifter, line 1: nit:
I suggest to move this file to another directory, for example `packages/fluentui/react-builder/src/narration`. Currently it stays in `packages/fluentui/react-builder/src/components/` where I expect too see only React components.

adamsamec: Okay, moved as suggested. The change is included in [this PR](https://github.com/microsoft/fluentui/pull/15345).

PR #15070, opened 2020-09-16
Color tokens are now prefixed with color

dzearing:
* All color tokens now prefixed with color
* "accent" color token set renamed to "brand" per design guidance

Before: 

```jsx
theme = {
  tokens: {
     body: { ... },
     accent: { ... }
  }
};
```

After:
```jsx
theme = {
  tokens: {
    color: {
       body: { ... },
       brand: { ... }
    }
  }
};
```


xugao (approved):
LGTM!

PR #15077, opened 2020-09-16
Add Callout prop to allow dismiss on target click

taenri:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes internal issue

#### Description of changes

An internal issue requires a Callout instance to dismiss on target element click before a new Callout should be rendered. For example, if clicking on the target element displays the Callout then clicking it again would result in an additional Callout element being rendered without dismissing the one that is already displayed. This can result in many Callout elements being rendered on the page as long as the user keeps clicking on the target without clicking elsewhere. Enabling the Callout to dismiss when the target is clicked would solve this issue.

Therefore, we would like to expose a new Callout property to allow the parent to specify whether target element clicks should also dismiss the Callout.

#### Focus areas to test

N/A


packages/office-ui-fabric-react/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -341,7 +341,8 @@ export class CalloutContentBase extends React.Component<ICalloutProps, ICalloutS
         isEventTargetOutsideCallout &&
         ((this._target as MouseEvent).stopPropagation ||
           !this._target ||
-          (target !== this._target && !elementContains(this._target as HTMLElement, target))))
+          (target !== this._target &&
```
soettl, line 344: Does this condition need to be:
`(this.props.allowDismissOnTargetClick || target !== this._target && !elementContains(this._target as HTMLElement, target))` so that it also dismisses when `target === this._target`?

taenri: Ah, that makes sense. Thanks for the catch!

packages/office-ui-fabric-react/src/components/Callout/Callout.types.ts
```diff
@@ -98,6 +98,12 @@ export interface ICalloutProps extends React.HTMLAttributes<HTMLDivElement> {
    */
   preventDismissOnLostFocus?: boolean;
 
+  /**
+   * If true then the callout will dismiss when the target element is clicked
+   * @defaultvalue false
+   */
+  allowDismissOnTargetClick?: boolean;
```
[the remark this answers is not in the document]
taenri: Sounds good to me. :)

soettl (approved):

joschect (approved):

PR #15084, opened 2020-09-17
Avatar tests

behowell:
#### Pull request checklist

#### Description of changes

Add ~~screener,~~ perf, snapshot, and unit tests for the Avatar component.

Refactor some Avatar functionality for testability, and fix a few styling bugs in edge cases found through testing.


ecraig12345:
Instead of checking in the demo images, it would probably be better to add them to our CDN (then add references in TestImages). Otherwise, especially once we switch to running screener from blob storage like Shift just did for v0, we're likely to run into issues the images not loading in time. I'll message you with info about how to do that.

ecraig12345:
See #15097 for uploading images to the CDN

ecraig12345 (approved):

khmakoto (approved):

xugao (approved):

apps/perf-test/package.json
```diff
@@ -24,6 +24,7 @@
     "lodash": "^4.17.15"
   },
   "dependencies": {
+    "@fluentui/react-avatar": "^0.6.0",
```
ecraig12345, line 27: This should be on the latest version (`^0.6.3`) in all packages. Once you fix it in one package you should be able to run `yarn syncpack:fix` to automatically fix the rest.

behowell: Ok. I had reverted the package version to match what was checked into the examples package so I wouldn't be modifying unrelated packages in this PR. But I'll go back and update that other package too.

ecraig12345: Hmm if the examples package is different then @khmakoto or I must have had a bad merge at some point. I'll fix that in master, sorry.

ecraig12345: It's fixed now. Apologies again about requiring the extra changes. :(

behowell: Ok thanks -- just pulled in your change

PR #15087, opened 2020-09-17
feat(Breadcrumb): list behavior with arrow navigation

YuanboXue-Amber:
#### Pull request checklist

#### Description of changes

Breadcrumb component as list behavior.
- Using arrow to navigate between breadcrumbs
- By default the active item is text, and not focusable by screen reader

#### Focus areas to test

(optional)


assuncaocharles:
Tested the [deployed pr](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/15087/merge/react-northstar/components/breadcrumb/definition) and I couldn't navigate by arrows

YuanboXue-Amber:
> Tested the [deployed pr](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/15087/merge/react-northstar/components/breadcrumb/definition) and I couldn't navigate using arrows

Thanks for finding it. I couldn't make it work either. let me test what's wrong.

packages/fluentui/react-northstar/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -74,16 +74,16 @@ export const Breadcrumb = compose<'nav', BreadcrumbProps, BreadcrumbStylesProps,
     const unhandledProps = useUnhandledProps(composeOptions.handledProps, props);
     const ElementType = getElementType(props);
 
-    const result = getA11yProps.unstable_wrapWithFocusZone(
+    const result = (
       <ElementType
```
assuncaocharles: @layershifter if we add the `unstable_wrapWithFocusZone` wrapping `<ElementType />` we get error from `isConformant` for not passing `ref` to root component. For now we moved to inside wrapping the list, is it proper?

layershifter: It can work there, but we need to fix it in `unstable_wrapWithFocusZone`

assuncaocharles: Should I open a issue for that?

YuanboXue-Amber: pinged shift and got reply:

> I don't think so as it's unclear what we would like to fix

kolaps33:
Tried it with JAWS and NVDA on Windows.
VoiceOver on MAC.
Couldn't find anything wrong. Keyboard navigation works through the links. The text which is not link is navigable with the screen reader navigation. (VPC mode, browse mode or VO keys)

kolaps33 (approved):

assuncaocharles (approved):
LGTM

PR #15093, opened 2020-09-17
fix/Do not submit DatepickerForm when clicking Popup button

rymeskar:
#### Pull request checklist

#### Description of changes

* Do not submit DatepickerForm when clicking Popup button.
* Reasoning: button is by defaulf the 'submit' type

#### Focus areas to test

(optional)


assuncaocharles (approved):
LGTM.

🙇🏻‍♂️ Just one heads-up, we are getting this error from Marat's schema:

```Illegal attribute value 'dialog' of attribute 'aria-haspopup'```

I wonder if we should adapt the possible values of `aria-haspopup` in the schema

PR #15094, opened 2020-09-17
Reexport SpinButton from react-spinbutton

tomi-msft:
#### Pull request checklist

#### Description of changes

This PR is the second stage of using v8 button in the SpinButton component. It moves SpinButton from react-next to react-spinbutton and reexports it

#### Focus areas to test

(optional)


packages/react-spinbutton/src/utilities/ButtonGrid/ButtonGridCell.tsx
```diff
@@ -0,0 +1,99 @@
+import * as React from 'react';
```
xugao, line 1: we should not be forking files. we need to make these shared between `react-next` and your new package. i think we will have to find good places for them. this likely means we need to scaffold more things / extract common code into shared packages

ecraig12345:
Closing since I think we decided on a different approach.

PR #15097, opened 2020-09-17
Upload demo images to CDN

ecraig12345:
Upload images used in screener and other places to our CDN that we use for the website (http://fabricweb.azureedge.net/fabric-website/*). This will make them load faster than if we uploaded them to normal blob storages, which can be important for getting correct results in screener tests.

For testing purposes I have the images uploading as part of the PR build, but before check-in I plan to change it so the images upload **only on master builds** (mainly to avoid accidental changes). This means that if you want to use a new image, you'll have to check it in separately first, but that shouldn't be a big issue since it won't happen very often.

Also update other upload tasks to use the new AzureUpload task which gzips things. This should make the PR deploy site load faster.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -101,7 +101,7 @@ jobs:
 
       - task: AzureUpload@2
         displayName: Upload demo images (master)
-        # condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
+        condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
         inputs:
           azureSubscription: 'UI Fabric (bac044cf-49e1-4843-8dda-1ce9662606c8)'
           BlobPrefix: 'assets'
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -99,6 +99,17 @@ jobs:
           ContainerName: '$web'
           BlobPrefix: '$(deployBasePath)'
 
+      - task: AzureUpload@2
+        displayName: Upload demo images (master)
+        # condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
```
behowell, line 104: Should this condition be uncommented? Or deleted?

behowell (approved):
Thanks!

layershifter (approved):
👍

ecraig12345:
It's working! http://fabricweb.azureedge.net/fabric-website/assets/images/avatar/large/chris.jpg

ecraig12345:

PR #15106, opened 2020-09-17
Fix: design system property inheritance

eljefe223:
#### Pull request checklist

#### Description of changes
This fixes an issue where other design system properties like `direction` were not being passed down to the extended DSP.


packages/web-components/src/card/index.ts
```diff
@@ -1,7 +1,7 @@
-import { customElement } from '@microsoft/fast-element';
 import { ColorRGBA64, parseColorHexRGB } from '@microsoft/fast-colors';
-import { designSystemProperty, DesignSystemProvider, CardTemplate as template } from '@microsoft/fast-foundation';
+import { designSystemProperty, designSystemProvider, CardTemplate as template } from '@microsoft/fast-foundation';
 import { createColorPalette, DesignSystem } from '@microsoft/fast-components-styles-msft';
+import { FluentDesignSystemProvider } from '..';
```
chrisdholt, line 4: I think we'll want to point to the index file here for safety, sometimes this can create TS issues

chrisdholt (approved):
Approving based on that one change coming in

packages/web-components/src/card/index.ts
chrisdholt:
```suggestion
import { FluentDesignSystemProvider } from '../index';
```

PR #15108, opened 2020-09-17
FocusZone: Moving custom onFocus call at the end of private _onFocus logic

khmakoto:
#### Pull request checklist

#### Description of changes

The user provided `onFocus` call should be called at the end of the private `_onFocus` logic so that the default logic does not override the user provided one.

#### Focus areas to test

(optional)


packages/react-focus/src/components/FocusZone/FocusZone.tsx
```diff
@@ -469,6 +463,12 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
     if (stopFocusPropagation || doNotAllowFocusEventToPropagate) {
       ev.stopPropagation();
     }
+
+    if (onFocus) {
```
dzearing, line 467: consider making a test case which exercises the issue so that we don't break this.

packages/react-focus/src/components/FocusZone/FocusZone.tsx
```diff
@@ -469,6 +463,12 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
...
+      onFocus(ev);
```
dzearing, line 468: My thought would be that you do the user call first. 

Then, if they have called ev.preventDefault(), don't do anything additional.

```jsx
onFocus?.(ev);

if (!ev.defaultPrevented) {
  ... do stuff
}
```

dzearing (approved):

PR #15113, opened 2020-09-18
Fix useKeytipRef() function adding keytip sequence value multiple times

### LINKED ISSUE #15112: useKeytipRef() function adds the same attribute multiple times
### Environment Information

- **Package version(s)**: 7.132.0
- **Browser and OS versions**: all browsers

In a functional component, call `useKeytipRef()` to add keytip for any UI control. Make sure the functional component has state so that it will render more than once. 

#### Actual behavior:

If we now use the browser `Inspect Element` tool to select the UI control, we'll see that the `aria-describedby` attribute contains multiple values of the keytip sequence. For example, if the keytip sequence is `h-f-f`, the `aria-describedby` attribute will be `h-f-f h-f-f h-f-f`.

#### Expected behavior:

The `aria-describedby` attribute should be exactly `h-f-f`. 

Requested priority: Normal

dotnetjunky:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15112 

#### Description of changes

There's a bug in the `setAttribute()` function which fails to check if the incoming value already exists in the affected attribute.

#### Focus areas to test

Unit test

### DIFF
--- a/change/office-ui-fabric-react-2020-09-17-18-11-24-fixUseKeytipRef.json
+++ b/change/office-ui-fabric-react-2020-09-17-18-11-24-fixUseKeytipRef.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fix bug in useKeytipRef function",
+  "packageName": "office-ui-fabric-react",
+  "email": "kinhln@microsoft.com",
+  "dependentChangeType": "patch",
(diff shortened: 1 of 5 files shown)


dotnetjunky:
@xugao for your review.

xugao (approved):
Thanks for fixing!

JustSlone:
Re-spinning build checks

PR #15122, opened 2020-09-18
Link/Toggle/Pivot/Slider: typing fixes and update release notes.

xugao:
#### Pull request checklist

#### Description of changes
I did a comparison between typings in OUFR and typings in react-checkbox/toggle/link/slider/pivot and fix some un-intended differences I found. Also added few related points in release notes.

#### Focus areas to test

(optional)


packages/react-next/RELEASE_NOTES.md
```diff
@@ -150,10 +150,10 @@ TODO: Diff of OUFR vs date-time DatePicker
 ### Pivot
 
 - Removed deprecated and redundant props from v7, including: `initialSelectedKey` and `defaultSelectedIndex`. Use `selectedKey` or `defaultSelectedKey` to define the selected tab, and provide `itemKey` on pivot item children.
-  - Removed deprecated styles prop `linkIsSelected?: boolean;`.
-  - Removed styles prop `rootIsLarge` and added `linkSize` instead.
-  - Removed styles prop `rootIsTabs` and added `linkFormat` instead.
-  - TODO: enumerate all removed props
+- Style props changes
```
behowell, line 153: (nit) I got confused that this was talking about style-related props on IPivotProps, instead of IPivotStyleProps. Perhaps clarify?
```suggestion
- `IPivotStyleProps` changes
```

packages/react-next/RELEASE_NOTES.md
```diff
@@ -215,6 +215,10 @@ TODO: document any API or functionality changes
 
 - Updated enums to string union type: `PivotLinkFormat`, `PivotLinkSize`. (#13370)
 
+## New features
+
+- Pivot overflow (TODO: Add more detailed documentation)
```
behowell, line 220:
```suggestion
- Pivot supports displaying an overflow menu when there is not enough room to display all of the tabs. This can be enabled by setting `overflowBehavior="menu"` on the Pivot.
```

behowell (approved):

packages/react-link/src/components/Link/Link.types.ts
```diff
@@ -75,6 +73,11 @@ export interface ILinkProps
    */
   theme?: ITheme;
 
+  /**
+   * A component that should be used as the root element of the link returned from the Link component.
```
ecraig12345, line 77: Should this be "component or element" or "component or tag" to make it clear that native elements are okay too?

packages/react-next/RELEASE_NOTES.md
```diff
@@ -165,7 +165,7 @@ TODO: Diff of OUFR vs date-time DatePicker
 
 ### Slider
 
-TODO: document any API or functionality changes
+- Deprecated prop `ariaLabel` (use `aria-label`)
```
ecraig12345, line 168: If this is the overall plan, we should probably make a work item to make a change like this across all the components (and ensure they handle the native prop correctly).

xugao: good catch, i believe this change was made bc of the convergence work. I reverted the change. we can discuss in next sync to see if we should do a round of deprecations.

PR #15131, opened 2020-09-18
Floating Pickers: Make pickers strict mode compliant

### LINKED ISSUE #14293: ExtendedPicker: strict mode compliance (UNSAFE_, easy)

### LINKED ISSUE #14294: FloatingPicker: strict mode compliance (UNSAFE_, easy/medium)

### LINKED ISSUE #14317: BaseSelectedItemsList: strict mode compliance (UNSAFE_)
BaseSelectedItemsList has the following methods which prevent strict mode compliance:
- `componentWillReceiveProps` which maps `props.selectedItems` to state and `props.selection` to a private member
- `componentWillUpdate` which synchronizes the selection's items with the items from state. This could potentially be moved to `componentDidUpdate` (with a `forceUpdate` call to force re-rendering).

Converting to a function component is not an option here (without larger architectural changes) since pickers currently rely on inheritance.

joschect:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14317
Fixes #14293
Fixes #14294

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-next-2020-09-18-16-25-30-pickers-fix.json
+++ b/change/@fluentui-react-next-2020-09-18-16-25-30-pickers-fix.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Floating Pickers: Make compliant",
+  "packageName": "@fluentui/react-next",
+  "email": "joschect@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-09-18T23:25:30.987Z"
+}
--- a/change/@fluentui-react-next-2020-09-18-16-25-30-pickers-fix.json
+++ b/change/@fluentui-react-next-2020-09-18-16-25-30-pickers-fix.json
@@ -1,6 +1,6 @@
 {
   "type": "prerelease",
-  "comment": "Floating Pickers: Make compliant",
+  "comment": "Floating Pickers: Make strict mode compliant",
   "packageName": "@fluentui/react-next",
   "email": "joschect@microsoft.com",
   "dependentChangeType": "patch",
--- a/packages/react-next/src/components/ExtendedPicker/BaseExtendedPicker.tsx
+++ b/packages/react-next/src/components/ExtendedPicker/BaseExtendedPicker.tsx
@@ -14,7 +14,6 @@ const styles: any = stylesImport;
 
 export interface IBaseExtendedPickerState<T> {
   queryString: string | null;
-  suggestionItems: T[] | null;
 }
 
 export class BaseExtendedPicker<T, P extends IBaseExtendedPickerProps<T>>
@@ -35,9 +34,6 @@ export class BaseExtendedPicker<T, P extends IBaseExtendedPickerProps<T>>
 
     this.state = {
       queryString: '',
-      // TODO: determine whether this can be removed
-      // eslint-disable-next-line react/no-unused-state
-      suggestionItems: this.props.suggestionItems ? (this.props.suggestionItems as T[]) : null,
     };
   }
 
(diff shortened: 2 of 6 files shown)


packages/react-next/src/components/ExtendedPicker/BaseExtendedPicker.tsx
```diff
@@ -41,40 +38,18 @@ export class BaseExtendedPicker<T, P extends IBaseExtendedPickerProps<T>>
       // TODO: determine whether this can be removed
       // eslint-disable-next-line react/no-unused-state
```
ecraig12345, line 39: While you're here do you think it would be possible to remove this from state? The no-unused-state rule means the static analysis thinks the state member is never accessed (even if it's sometimes updated).

packages/react-next/src/components/FloatingPicker/BaseFloatingPicker.tsx
```diff
@@ -161,6 +155,11 @@ export class BaseFloatingPicker<T, P extends IBaseFloatingPickerProps<T>>
 
   protected renderSuggestions(): JSX.Element | null {
     const TypedSuggestionsControl = this.SuggestionsControlOfProperType;
+
+    if (this.props.suggestionItems) {
+      this.suggestionStore.updateSuggestions(this.props.suggestionItems!);
```
ecraig12345, line 160: Sorry if I asked this before, but is the cost of doing this every render a concern? Wondering if it would be better to do only in `componentDidUpdate` if `this.props !== prevProps` (any updates done there will be flushed before rendering, assuming you setState or forceUpdate).

joschect: You did! And it shouldn't be an issue.

packages/react-next/src/components/ExtendedPicker/BaseExtendedPicker.tsx
joschect: I want to clean this up when I clean up the rest of the linting rules since it would be nice to batch those together. First get all of these changes and make it compliant, then remove all the anys and fix that so it can be done in stages.

change/@fluentui-react-next-2020-09-18-16-25-30-pickers-fix.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Floating Pickers: Make compliant",
```
ecraig12345, line 3: Probably good to specify that compliant refers to strict mode (there are lots of possible types of "compliant")
```suggestion
  "comment": "Floating Pickers: Make strict mode compliant",
```

ecraig12345 (approved):

PR #15142, opened 2020-09-21
fix(Tree): aria-selected to aria-checked

### LINKED ISSUE #15042: Tree aria-selected is not narrated
Figure out that JAWS has stopped narrated "aria-selected" state for the multi select tree.
https://fluentsite.z22.web.core.windows.net/components/tree/definition?showCode=false&showRtl=false&showTransparent=false&showVariables=false#usage-tree-multiselect-example

Bug was reported to the JAWS:
https://github.com/FreedomScientific/VFO-standards-support/issues/432#issuecomment-686385093

JAWS admit it as bug but recommending use "aria-checked".

Did some testing and figure out that "aria-checked" will fix JAWS narration. 
NVDA doesn't narrate not checked state only once user "select" the option and then navigate back it narrate checked.
JAWS with aria-checked narrates both "checked" and "not checked" state. 

So basically implementing "aria-checked" will break a bit NVDA experience, but in overall it will fix more. 

My proposal of the fix is change current "aria-selected" on the treeitem into "aria-checked".

YuanboXue-Amber:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15042

#### Description of changes

Use `aria-checked` for multi select tree instead of `aria-selected`.

#### Focus areas to test

(optional)


packages/fluentui/accessibility/test/behaviors/treeItemBehavior-test.tsx
```diff
@@ -62,4 +62,25 @@ describe('TreeItemBehavior', () => {
...
+    test(`is added with 'false' value to an item that has hasSubtree, is selectable and not selected`, () => {
+      const expectedResult = treeItemBehavior({ hasSubtree: true, selectable: true });
+      expect(expectedResult.attributes.root['aria-checked']).toEqual(false);
+    });
+    test(`is added with 'mixed' value to an item that has hasSubtree, is selectable and in indeterminate state`, () => {
+      const expectedResult = treeItemBehavior({ hasSubtree: true, selectable: true, indeterminate: true });
+      expect(expectedResult.attributes.root['aria-checked']).toEqual('mixed');
+    });
+
+    test(`is not added to an item that is not selectable`, () => {
```
assuncaocharles, line 81: ❤️

packages/fluentui/accessibility/src/behaviors/Tree/treeItemBehavior.ts
```diff
@@ -35,7 +35,9 @@ export const treeItemBehavior: Accessibility<TreeItemBehaviorProps> = props => {
           'aria-setsize': props.treeSize,
           'aria-posinset': props.index,
           'aria-level': props.level,
-          ...(props.selectable && { 'aria-selected': props.selected }),
+          ...(props.selectable && {
+            'aria-checked': props.selected ? true : props.indeterminate ? ('mixed' as const) : false,
```
assuncaocharles, line 39:
```suggestion
            'aria-checked': props.indeterminate ? ('mixed' as const) : props.selected,
```

?

YuanboXue-Amber:
Below is the behavior of navigating up from top to Tyrion:
- NVDA
NVDA-chrome
- JAWS
JAW-chrome-Up-to-Tyrion

packages/fluentui/accessibility/src/behaviors/Tree/treeItemBehavior.ts
```diff
@@ -35,7 +35,9 @@ export const treeItemBehavior: Accessibility<TreeItemBehaviorProps> = props => {
           'aria-setsize': props.treeSize,
           'aria-posinset': props.index,
           'aria-level': props.level,
-          ...(props.selectable && { 'aria-selected': props.selected }),
+          ...(props.selectable && {
```
jurokapsiar, line 38: can you pls add a specification or description in the comments above?

jurokapsiar (approved):

packages/fluentui/accessibility/src/behaviors/Tree/treeItemBehavior.ts
YuanboXue-Amber: Added `description` comments :) 
Didn't add comments as `specification` because there's unit test targeting this behavior.

kolaps33 (approved):
Look great. 

Tried it with the JAWS 2020.2006.12.
NVDA 2020.2
Chrome: Version 85.0.4183.121 (Official Build) (64-bit)

Confirm it works as on the attached screenshot from Yuanbo.

PR #15143, opened 2020-09-21
docs: serve images from CDN for N* docs

layershifter:
A followup on #15097, tries to solve continuous issues with Screener that serves our images from Azure Blob (#14880):

image

Before this PR images were served from Azure Blob, with these changes they will be served from CDN. This will make them load faster than if we uploaded them to normal blob storages, which can be important for getting correct results in screener tests.

---

**The images will upload only on master builds** (mainly to avoid accidental changes). This means that if you want to use a new image, you'll have to check it in separately first, but that shouldn't be a big issue since it won't happen very often. An FAQ entry to `@fluentui/docs` was added.

This also solves an issue with CodeSandbox where we got broken images:
- before https://codesandbox.io/s/otjqi?module=/example.js
- after https://codesandbox.io/s/jj04y?module=/example.js

### DIFF
--- a/packages/fluentui/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
+++ b/packages/fluentui/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
@@ -355,7 +355,13 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
       },
       {
         disabled: currentCodeLanguage !== 'ts',
-        icon: <Image src="public/images/github.png" width="16px" height="16px" />,
+        icon: (
+          <Image
+            src="https://fabricweb.azureedge.net/fabric-website/assets/images/github.png"
+            width="16px"
+            height="16px"
+          />
+        ),
         content: 'Edit',
         href: ghEditHref,
         rel: 'noopener noreferrer',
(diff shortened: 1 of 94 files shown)


packages/fluentui/docs/src/examples/components/Avatar/Variations/AvatarExampleSize.shorthand.tsx
```diff
@@ -19,15 +19,26 @@ const AvatarExampleSizeShorthand = () => (
       <React.Fragment key={size}>
         <strong>{size}</strong>
         <div>
-          <Avatar size={size} image="public/images/avatar/small/matt.jpg" status={statusProps} />
```
miroslavstastny, line 22: The assets are still accessible using the relative path - what will stop me from using the relative path and mix the two approaches?

layershifter: I fixed this in a19ab2a, assets are no longer accessible via `/public`

miroslavstastny (approved):

packages/fluentui/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -355,7 +355,13 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
       },
       {
         disabled: currentCodeLanguage !== 'ts',
-        icon: <Image src="public/images/github.png" width="16px" height="16px" />,
+        icon: (
+          <Image
+            src="https://fabricweb.azureedge.net/fabric-website/assets/images/github.png"
```
ecraig12345, line 360: Bit late on this but if possible it would be good to put the base URL in a constant somewhere.

PR #15147, opened 2020-09-21
feat(tree): Multiselect tree as list behavior

### LINKED ISSUE #15041: Add multi select support for Tree as a list
Curently Fluent UI has Tree variant "As a list"
https://fluentsite.z22.web.core.windows.net/components/tree/definition#usage-tree-as-list-example
which purpose is to use it on the MAC for the accessibility reason. (as VoiceOver doesn't support tree properly) 

We would need to add similar "As a list" for the "multi selectable" tree. 

I tried it very rafly in this PR:
https://github.com/microsoft/fluentui/pull/15035/files
and tested the concept on MAC with VoiceOver and it was reasonable. 

No discussion with the team.

YuanboXue-Amber:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15041

#### Description of changes

Create a `listbox` behaviour to work with multiselect tree. 
This behavior is mostly for mac Voice Over. VO does not support `tree` role.

This change creates acceptable behaviour for VoiceOver's users.  However there're still some issues with VoiceOver's support over listbox and option role. But these issues won't affect users who use multi-selectable tree with **all items selectable**.
These issues are documented as known issues.

---

Couple things worth noticing for reviewers:

- why I went for `aria-selected` instead of `aria-checked`:
`aria-checked` doesn't work with VO
VO-chrome-checked-not-working

- VoiceOver modifier + space key, i.e. `VO+space`, has interesting behavior on **multi-selectable tree as <ins>listbox</ins>**:
The DOM structure of one tree node with subtree, can be simplified into following:
```html
  <div role="option" id="tree-item-1">
    <span role="none" id="tree-title-1">Item 1</span>
  </div>
```
When `VO+space` is triggered on this kind of tree node, a browser mouse click event will be triggered, only on the `div` element. 
However, for **multi-selectable tree as <ins>list</ins>**, where the parent `div` is of `role=listitem` instead of `option`, a `VO+space` will also triggers the onclick event on the `span` element.
Here's a codepen demo: https://codepen.io/amberpie/pen/YzWyPmd
(tested on: MacOS 10.15.7, Chrome 86)

#### Focus areas to test

(optional)


packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
```diff
@@ -6,14 +6,18 @@ import { treeItemAsListItemBehavior } from './treeItemAsListItemBehavior';
...
+ * Adds attribute 'role=listbox' to 'root' slot if 'selectable' property is true. Sets the attribute to 'list' otherwise.
+ * Adds attribute 'aria-multiselectable=true' to 'root' slot if 'selectable' property is true. Does not set the attribute otherwise.
  */
 export const treeAsListBehavior: Accessibility<TreeBehaviorProps> = props => {
   const behavior = treeBehavior(props);
   return _.merge(behavior, {
     attributes: {
       root: {
-        role: 'list',
+        role: props.selectable ? 'listbox' : 'list',
```
assuncaocharles, line 17: why `listbox`?

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
```diff
@@ -6,14 +6,18 @@ import { treeItemAsListItemBehavior } from './treeItemAsListItemBehavior';
...
+        ...(props.selectable && {
+          'aria-multiselectable': true,
```
assuncaocharles, line 19: I think this might be missing in the `treeBehavior` as well. Should we added @kolaps33 ?

packages/fluentui/accessibility/src/behaviors/Tree/treeItemAsListItemBehavior.ts
```diff
@@ -5,16 +5,21 @@ import { treeTitleAsListItemTitleBehavior } from './treeTitleAsListItemTitleBeha
...
   const behavior = treeItemBehavior(props);
+  delete behavior.attributes.root['aria-checked'];
 
   const definition = _.merge(behavior, {
     attributes: {
       root: {
         ...(props.hasSubtree && {
-          role: 'listitem',
+          role: props.selectable ? 'option' : 'listitem',
+          ...(props.selectable && { 'aria-selected': !!props.selected }),
```
assuncaocharles, line 22: aria-checked?

packages/fluentui/accessibility/src/behaviors/Tree/treeTitleAsListItemTitleBehavior.ts
```diff
@@ -5,16 +5,21 @@ import { treeTitleBehavior, TreeTitleBehaviorProps } from './treeTitleBehavior';
...
   const behavior = treeTitleBehavior(props);
+  delete behavior.attributes.root['aria-checked'];
 
   const definition = _.merge(behavior, {
     attributes: {
       root: {
         ...(!props.hasSubtree && {
-          role: 'listitem',
+          role: props.selectable ? 'option' : 'listitem',
+          ...(props.selectable && { 'aria-selected': !!props.selected }),
```
assuncaocharles, line 22: aria-checked?

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
YuanboXue-Amber: I was checking both:
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role
- https://www.w3.org/WAI/PF/aria/roles#listbox

Seems like this is the one to go for **list** that is **selectable**
Also in the original issue @kolaps33 made a [draft pull request](https://github.com/microsoft/fluentui/pull/15035/files) suggesting listbox

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
YuanboXue-Amber: I'm not so sure if multiselectable is needed for the classic treeBehavior. Because we decided to go with `aria-checkbox` there

packages/fluentui/accessibility/src/behaviors/Tree/treeItemAsListItemBehavior.ts
YuanboXue-Amber: YeaH :D believe me I did not forget. Somehow `aria-checked` refuse to work with Mac VoiceOver. I have a screen shot in the PR description for it. Since `treeAsListBehavior` is mostly for Mac, so I went with `aria-selected`, which narrates properly with VO

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
assuncaocharles: Nice, I wonder if we should change something for Selectable Tree also then?!

packages/fluentui/accessibility/src/behaviors/Tree/treeItemAsListItemBehavior.ts
assuncaocharles: Should we keep only `aria-selected` then @kolaps33 for consistence?

YuanboXue-Amber:
## Test result on Voice Over
multiselect-tree-VO

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
kolaps33: Hi @assuncaocharles & @YuanboXue-Amber agree with both of you :)
based on aria: 
https://www.w3.org/TR/wai-aria-1.1/#aria-multiselectable
looks like it is related only to the 'aria-selected'. So until we have 'aria-selected' it would be good to add. 
But when we are going to change it into aria-checkbox (high probably), then we wouldn't add.

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
YuanboXue-Amber: I'm not sure what do you mean? 
If you meant the normal tree behaviour, it's not list there, it's role='tree'.

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
YuanboXue-Amber: You're right, it should be there for normal tree behaviour as well. I just did a small test on multiselect tree with normal behaviour, the accessibility computed looks like:
```
Name: "default"
aria-labelledby: Not specified
aria-label: "default" 
title: Not specified
Role: tree
Focusable: true
Multi-selectable: false
Required: false
```
And `Multi-selectable: false` will turn `true` if I added 'aria-multiselectable'

Surprisingly NVDA didn't narrate differently with/without 'aria-multiselectable'. But I still think we should put it there.

YuanboXue-Amber: Resolved by [74a6595](https://github.com/microsoft/fluentui/pull/15147/commits/74a6595154110b452669287ef28cf16e4ba108db)

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
assuncaocharles: Sorry, my mistake, I wanted to say Selectable List

packages/fluentui/accessibility/test/behaviors/treeItemAsListItemBehavior-test.tsx
```diff
@@ -7,9 +7,40 @@ describe('TreeItemAsListItemBehavior', () => {
...
+    test(`is not added to an item that is not selectable `, () => {
+      const expectedResult = treeItemAsListItemBehavior({ hasSubtree: true });
+      expect(expectedResult.attributes.root['aria-selected']).toBeUndefined();
+    });
+  });
+
+  describe('aria-checked', () => {
+    test(`is always undefined`, () => {
+      let expectedResult = treeItemAsListItemBehavior({ hasSubtree: true, selectable: true, selected: true });
+      expect(expectedResult.attributes.root['aria-checked']).toBeUndefined();
```
assuncaocharles, line 41: `aria-selected`?

YuanboXue-Amber: For me, it doesn't hurt to do a check to make sure we don't accidentally put both `aria-selected` and `aria-checked`

packages/fluentui/docs/src/examples/components/Tree/Usage/TreeMultiselectAsListExample.shorthand.tsx
```diff
@@ -0,0 +1,136 @@
...
+      },
+    ],
+  },
+];
+
+const TreeMultiselectAsListExample = () => (
+  <Tree
+    defaultSelectedItemIds={['tree-item-122', 'tree-item-123']}
+    selectable
+    aria-label="default"
```
kolaps33, line 130: This is only improving user experience:
Please if you could label the tree a bit more descriptive. 
I mean more related to the example. 
Something like "Multi select" or "Multi select as a list" or anything what is more related to the example. 
Because currently VoiceOver narrate "default, listbox"

packages/fluentui/accessibility/src/behaviors/Tree/treeTitleAsListItemTitleBehavior.ts
```diff
@@ -14,7 +17,8 @@ export const treeTitleAsListItemTitleBehavior: Accessibility<TreeTitleBehaviorPr
...
```
kolaps33, line 21: not sure what is going on, but in the final DOM there is aria-selected as well on the items which are not "selectable"
Please if you could try to fix it :)
image

assuncaocharles: Just saw it also happens in the normal multi selectable tree

packages/fluentui/docs/src/examples/components/Tree/Usage/TreeMultiselectAsListExample.shorthand.tsx
YuanboXue-Amber: Resolved by [1df7b0f](https://github.com/microsoft/fluentui/pull/15147/commits/1df7b0fe7b2c55f88fb05d38f3f816646ed9e583)

packages/fluentui/accessibility/src/behaviors/Tree/treeTitleAsListItemTitleBehavior.ts
YuanboXue-Amber: Okay I will look into this

packages/fluentui/accessibility/src/behaviors/Tree/treeTitleAsListItemTitleBehavior.ts
```diff
@@ -14,7 +17,8 @@ export const treeTitleAsListItemTitleBehavior: Accessibility<TreeTitleBehaviorPr
     attributes: {
       root: {
         ...(!props.hasSubtree && {
-          role: 'listitem',
+          role: props.selectable ? 'option' : 'listitem',
```
kolaps33, line 20: very similar issue as with "aria-selected"
because we need to be consistent and has on whole selectable list just role=option, not mix there 'listitem'
image

YuanboXue-Amber: So if a parent div is `listbox`, all the children should be `option` even if it is not selectable?

packages/fluentui/CHANGELOG.md
```diff
@@ -24,8 +24,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Rename `indicatorBorderColorDefault` variable to `indicatorColorDefault` @assuncaocharles ([#14895](https://github.com/microsoft/fluentui/pull/14895))
 - Remove the `disabled` property from `Dialog` and `Popup` behaviors @rymeskar ([#14885](https://github.com/microsoft/fluentui/pull/14885))
 
-### Fixes
+### Fixes
```
assuncaocharles, line 27:
```suggestion
### Fixes
```

packages/fluentui/accessibility/test/behaviors/treeItemAsListItemBehavior-test.tsx
YuanboXue-Amber: Removed this test in the newer commit

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
YuanboXue-Amber: Sorry for the late reply, somehow I missed this message. Selectable list is already using listbox+option role :)

packages/fluentui/accessibility/src/behaviors/Tree/treeItemAsOptionBehavior.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @description
+ * Adds role 'option' to a non-leaf item.
+ * Adds 'aria-selected' to a non-leaf item if it is selectable, based on the 'selected' property
+ */
+export const treeItemAsOptionBehavior: Accessibility<TreeItemBehaviorProps> = props => {
+  const behavior = treeItemBehavior(props);
+
+  const definition = _.merge(behavior, {
```
kolaps33, line 14: I think we need be careful here which aria attribute we will apply on the option role.
https://www.w3.org/TR/wai-aria-1.1/#option

Because looks like, "aria-expanded", "aria-level" is not applicable on role=option.

packages/fluentui/accessibility/src/behaviors/Tree/treeTitleAsOptionBehavior.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @description
+ * Adds role 'option' if the title is a leaf node inside the tree
+ * Adds 'aria-selected' to a leaf treeTitle if it is selectable, based on the 'selected' property
+ */
+export const treeTitleAsOptionBehavior: Accessibility<TreeTitleBehaviorProps> = props => {
+  const behavior = treeTitleBehavior(props);
+
+  const definition = _.merge(behavior, {
```
kolaps33, line 14: I think we need be careful here which aria attribute we will apply on the option role.
https://www.w3.org/TR/wai-aria-1.1/#option

Because looks like, "aria-expanded", "aria-level" is not applicable on role=option.

packages/fluentui/accessibility/src/behaviors/Tree/treeItemAsOptionBehavior.ts
```diff
@@ -0,0 +1,36 @@
...
+      },
+    },
+    childBehaviors: {
+      title: treeTitleAsOptionBehavior,
+    },
+  });
+
+  if (process.env.NODE_ENV !== 'production' && props.hasSubtree) {
+    // Override the default trigger's accessibility schema class.
+    definition.attributes.root['data-aa-class'] = 'TreeItemList';
```
kolaps33, line 32: maybe we will need a bit different schema for the tree item which is option? :)

assuncaocharles: I guess we will

packages/fluentui/accessibility/src/behaviors/Tree/treeTitleAsOptionBehavior.ts
YuanboXue-Amber: I just pushed a new change. Instead of `_.merge`, using below logic:
```js
  const definition = {
    ...behavior,
    attributes: {
```
And removed "aria-expanded", "aria-level" from this behaviour

mshoho (approved):

mshoho (approved):

YuanboXue-Amber:
## Behavior on MacOS 10.15.7 VoiceOver 

ezgif com-gif-maker

kolaps33:
Was looking to the PR and I see following:
1. looks like some bug, please if you could investigate a bit:
when parent is expanded and selectable, then control+option+space doesn't work on that parent. 
On the child the contol+option+space works and do selection. I think we should be consistent. 

2. (this is just suggestion, doesn't need to be implemented in the PR, because final state will do consumer of the component anyway..)
As there is added "aria-label" on expandable parent `${user name}, expandable` 
I would suggest do it similarly as VoiceOver narrate expandable/collapsable elements. Like "collapsed" or "expanded", depends on the state. 
collapsed-state
expanded-state

kolaps33 (approved):

packages/fluentui/accessibility/src/behaviors/Tree/treeItemAsListItemBehavior.ts
YuanboXue-Amber: Notes upon resolving this discussion:
PR #15142 changes all the `aria-selected` to `aria-checked` in order to provide a better experience for JAW users.
This PR keeps `aria-selected` because `aria-checked` does not work properly with VoiceOver.
Consistency is sacrificed for better screen reader users' experience

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListboxBehavior.ts
```diff
@@ -0,0 +1,31 @@
...
+    attributes: {
+      root: {
+        role: 'listbox',
+      },
+    },
+    childBehaviors: {
+      item: treeItemAsOptionBehavior,
+      title: treeTitleAsOptionBehavior,
+    },
+  });
```
pompomon, line 25: Let's just use spread instead of `_.merge`, like it's done in [toolbarItemBehavior](https://github.com/microsoft/fluentui/blob/a4898636dfef66f59f63740ab2646c7497410137/packages/fluentui/accessibility/src/behaviors/Toolbar/toolbarItemBehavior.ts#L18-L22)
```suggestion
  const definition = {...behavior, ...{
    attributes: {
      root: {
        role: 'listbox',
      },
    },
    childBehaviors: {
      item: treeItemAsOptionBehavior,
      title: treeTitleAsOptionBehavior,
    },
  }};
```

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
```diff
@@ -10,7 +10,7 @@ import { treeItemAsListItemBehavior } from './treeItemAsListItemBehavior';
  */
 export const treeAsListBehavior: Accessibility<TreeBehaviorProps> = props => {
   const behavior = treeBehavior(props);
-  return _.merge(behavior, {
+  const definition = _.merge(behavior, {
```
pompomon, line 13: Let's use spread here as well (see https://github.com/microsoft/fluentui/pull/15147/files#r504468796)

packages/fluentui/accessibility/src/behaviors/Tree/treeBehavior.ts
```diff
@@ -19,6 +19,9 @@ export const treeBehavior: Accessibility<TreeBehaviorProps> = props => {
         role: 'tree',
         'aria-labelledby': props['aria-labelledby'],
         tabIndex: -1,
+        ...(props.selectable && {
+          'aria-multiselectable': true,
```
pompomon, line 23: Please could you update the specification (or add it via new a11y tests created by Milan)?

packages/fluentui/react-northstar/src/components/Tree/TreeItem.tsx
```diff
@@ -238,6 +246,20 @@ export const TreeItem: ComponentWithAs<'div', TreeItemProps> & FluentComponentSt
...
+
+  const handleOnClickOnTreeItemOnly = (e: React.SyntheticEvent) => {
+    // onClick listener for mouse click on treeItem DOM only,
+    // which could be triggered by VO+space on selectable tree parent node
+    e.preventDefault();
+    e.stopPropagation();
+    if (e.target === e.currentTarget) {
+      handleSelection(e);
+    }
+  };
```
pompomon, line 262: Please why do you need a separate function here?

pompomon: And another thing - what if we call    `e.preventDefault();` and ` e.stopPropagation();` inside the `if()` case?

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListboxBehavior.ts
YuanboXue-Amber: The suggestion way of merging will provide different result than `_.merge`. 
For example, `behavior.attributes.root.tabIndex` is not merged in to `definition`. (codepen: https://codepen.io/amberpie/pen/JjKGPEv) 

Is there any specific reason spread operator is prefered? If so I could change to something like:
```ts
 attributes: {
      root: {
        ...behavior.attributes.root,
        role: 'listbox' as AriaRole,
      },
    },
```

YuanboXue-Amber: Communicate thru chat, going to change from merge to spread in each level :)

YuanboXue-Amber: Resolved by this [commit](https://github.com/microsoft/fluentui/pull/15147/commits/f6d948ee7fb131877a7d22cb2b19b28ea484bb34)

packages/fluentui/accessibility/src/behaviors/Tree/treeAsListBehavior.ts
YuanboXue-Amber: Resolved by this [commit](https://github.com/microsoft/fluentui/pull/15147/commits/50290696e06d38de8212e6e56c1823df8fa9aa65)

packages/fluentui/react-northstar/src/components/Tree/TreeItem.tsx
YuanboXue-Amber: Resolved by this [commit](https://github.com/microsoft/fluentui/pull/15147/commits/2cfd658d6462e95d9f67c079cbb76720ff3a7a7e) 

It's not a separate function anymore (so it can't be called by other code, therefore more readable), and `e.preventDefault()` `e.stopPropagation()` is removed

packages/fluentui/accessibility/src/behaviors/Tree/treeBehavior.ts
YuanboXue-Amber: Resolved by this [commit](https://github.com/microsoft/fluentui/pull/15147/commits/da17607de83c312772121a8637f11fbcaa9e3d98)

pompomon (approved):

PR #15160, opened 2020-09-21
Add prop to SelectedItemList as callback for when an item is edited

elisabethcvs:
#### Pull request checklist

#### Description of changes

Editing functionality was already there internally, but there was no callback for the consumer.

I added a UnifiedPicker example to show this functionality


packages/experiments/src/components/SelectedItemsList/SelectedItemsList.types.ts
```diff
@@ -129,4 +129,9 @@ export interface ISelectedItemsListProps<T> extends React.ClassAttributes<any> {
    * Callback for when an item or items needs to be inserted into the list
    */
   dropItemsAt?: (insertIndex: number, itemsToInsert: T[], indicesToRemove: number[]) => void;
+
+  /**
+   * Callback for when an item has been edited
+   */
+  onItemEdited?: (newItem: T | T[], index: number) => void;
```
elisabethcvs, line 136: wondering if replaceItem or onItemChange is a better name

amyngu: Maybe replaceItem is a better concept. Thinking about scenarios like expanding group DLs its really replacing the groupDL with multiple individual group members

nebhatna (approved):

Issue #14675, opened 2020-08-21
Verify all changes to OUFR have been ported to react-next


JustSlone:
- [ ] For converted function components, look at:
  - When it was copied to react-next
  - Whether any changes were made to the component in OUFR since then
  - If so, ensure the changes are included in react-next FC version
- [x] Ensure doc text changes have been ported, OR just delete the react-next copies (this refers to `react-next/src/components/<Component>/docs/*.md`)

czearing:
I created a HackMD with a list of react-next components and the possible changes that will need to be ported from office-ui-fabric-react: https://hackmd.io/LQEZYG2cSoGYgm9u-wvDag?both

JustSlone:
Nice, thanks @czearing!

Are you on point to port these changes over to react-next?

czearing:
Yes, I am currently working on making sure everything is ported over. I will keep the hack md updated to track progress.

ecraig12345:
Thanks for looking through that! I should have mentioned before, it would probably also be good to check for changes made within the couple weeks *before* the copy date of a component. In a few cases the PRs were open for awhile and some changes may have been made after the initial copy the PR was based on, but before merge.

Issue #14992, opened 2020-09-11
Port ContextualMenu/Callout/Popup focus-related PRs to react-next


ecraig12345:
#14351
#14363
#12806

ecraig12345:
Pretty sure this was fixed by #15161 but if not we can reopen.

PR #15161, opened 2020-09-21
React-Next porting a fix to callout dismissal when blurring to a new window

czearing:
#### Pull request checklist
- [X] Addresses #14675

#### Description of changes
Porting a fix to react-next for `Callout` remaining open after focus shift out of window.

porting: https://github.com/microsoft/fluentui/pull/14363


ecraig12345:
@joschect was supposed to handle this port and a few related ones (see #14992) since the changes span several components, and he's most familiar with them. But I'm not sure if he got to start on it yet.

joschect (approved):

packages/react-next/src/components/Callout/CalloutContent.base.tsx
```diff
@@ -345,6 +345,12 @@ function useDismissHandlers(
       }
     };
 
+    const dismissOnTargetWindowBlur = (ev: FocusEvent) => {
+      if (!preventDismissOnLostFocus && !targetWindow?.document.hasFocus() && ev.relatedTarget === null) {
```
joschect, line 349: I just checked in a fix for this, but there was another property added. This can't be the default behavior as it breaks embedded scenarios in things like Iframes.

ecraig12345 (approved):
I'm going to go ahead and merge this in the interest of having as few outstanding PRs as possible when we go into the code moves. @joschect if there's anything else that needs to be changed, please comment and we can make sure it's taken care of later.

PR #15167, opened 2020-09-22
Changing deps to be range.

dzearing:
To help reduce diamond dep problems, changing converged package dependencies to range based.


ecraig12345:
Will beachball properly handle bumping these?

What happens if one of the packages does actually introduce a break in a minor version?

dzearing:
I'm not convinced this is the correct fix. closing for now.

PR #15168, opened 2020-09-22
feat(UI builder): Add support to MaterialUI

vyhnalekl:
#### Pull request checklist

#### Description of changes

Added category `MaterialActionable` containing material components. Changed the way we export the code to code sandbox. Wrote custom `JSONTreeToJSXCode` function. Adding moduleName in codeToTree after code change.


packages/fluentui/react-builder/src/components/ComponentTree.tsx
```diff
@@ -188,18 +190,20 @@ export const ComponentTree: React.FunctionComponent<ComponentTreeProps> = ({
...
-        handleSelectComponent,
-        handleClone,
-        handleMove,
-        handleDeleteSelected,
-        handleAddComponent,
-        handleDeleteComponent,
-      ),
-    ) ?? [];
+    (Array.isArray(tree.props?.children) &&
+      tree.props?.children?.map(item =>
```
assuncaocharles, line 194: Maybe it is a usage for `React.Children.map(children, item => ...)` ?

packages/fluentui/react-builder/src/utils/codeToTree.ts
```diff
@@ -47,14 +48,26 @@ export const codeToTree: (code: string) => JSONTreeElement = code => {
       }
     }
 
+    // eslint-disable-next-line
+    const isElement = 'A' <= name[0] && name[0] <= 'Z';
```
assuncaocharles, line 52:
```suggestion
    const isElement = /^[A-Z]./.test(name);
```

assuncaocharles: Is this enough to say if it's `Element`?

packages/fluentui/react-builder/src/utils/codeToTree.ts
```diff
@@ -47,14 +48,26 @@ export const codeToTree: (code: string) => JSONTreeElement = code => {
...
+    const isComponent = isElement && (!props || (props && props['data-builder-id'] !== 'undefined'));
```
assuncaocharles, line 53: It can happen the `data-builder-id="undefined"` in some element?

packages/fluentui/react-builder/src/utils/codeToTree.ts
```diff
@@ -47,14 +48,26 @@ export const codeToTree: (code: string) => JSONTreeElement = code => {
...
+    let componentName = name;
+    if (isComponent && !name.includes('.')) {
+      componentName = `Fluent.${name}`;
+    }
```
assuncaocharles, line 57: _nit_

```suggestion
    const componentName = isComponent && !name.includes('.') ?  `Fluent.${name}` : name;
```

packages/fluentui/react-builder/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../../scripts/typescript/tsconfig.fluentui",
   "compilerOptions": {
     "module": "esnext",
+    "moduleResolution": "node",
```
assuncaocharles, line 5: Just out of curiosity, why was it needed?

packages/web-components/docs/api-report.md
```diff
@@ -222,6 +228,8 @@ export class FluentDesignSystemProvider extends DesignSystemProvider implements
     accentPalette: string[];
     backgroundColor: string;
     // (undocumented)
+    protected backgroundColorChanged(): void;
```
assuncaocharles, line 231: Not sure if it will apply here but we have some eslint rule to avoid using visibility modifiers. https://github.com/microsoft/fluentui/blob/master/packages/eslint-plugin/README.md#no-visibility-modifiers

packages/fluentui/react-builder/src/components/Canvas.tsx
```diff
@@ -264,82 +268,92 @@ export const Canvas: React.FunctionComponent<CanvasProps> = ({
...
+                  <span style={{ fontSize: '4rem' }} role="img" aria-label="Finger pointing left">
+                    👈
+                  </span>
+                  <div style={{ fontSize: '1.2rem', opacity: 0.5 }}>Drag n' Drop some components</div>
+                </div>
+              )}
+
+              <DebugSelector
+                active={isSelecting}
+                key={`debug-selector-${selectedComponent?.uuid ?? 'unknown'}`}
```
assuncaocharles, line 297: Can it happen it will have more than 1 `key="'unknown'"`? if so we should probably generate something random here.

assuncaocharles, line 297: Can it happen it will have more than 1 key="unknown"? if so we should probably generate something random here.

packages/fluentui/react-builder/src/utils/codeToTree.ts
vyhnalekl: That is a good question. It is not from my head, that is how we were deciding whether to add the `$$typeof` prop previously, I have just put the statement into a variable. The problem with this is that if user writes something like `<Div>` into the code, the UI builder will consider it as an element. The only solution how to fix it that I can think of is to check whether the `name` is in the array of possible objects to drag and drop. But that would cause problems for example for Icons.

packages/fluentui/react-builder/src/utils/codeToTree.ts
vyhnalekl: Yes, it happens to Icons. The problem is that we are adding some margins and paddings to every element with `data-builder-id`. So for icons we want to keep it undefined. But for example if user created new component in code editor. We need to generate new uuid.

packages/fluentui/react-builder/tsconfig.json
vyhnalekl: There were some issues with the editor keep saying `@material-ui/core` is not defined. Even though it was in package.json and added via yarn. The build was also failing because of this. This is fixing the problem.

packages/web-components/docs/api-report.md
vyhnalekl: I don't know why this file was changed, I think it was done automatically by yarn. Is it possible?

packages/fluentui/react-builder/src/components/Canvas.tsx
vyhnalekl: No, I don't think it can happen. Because we either have component selected (there is a uuid value) or we don't.

packages/fluentui/react-builder/src/components/ComponentTree.tsx
vyhnalekl: I am not sure whether I am implementing it correctly
`React.Children.map(tree.props?.children, item => {
      return jsonTreeToTreeItems(
        item,
        selectedComponentId,
        handleSelectComponent,
        handleClone,
        handleMove,
        handleDeleteSelected,
        handleAddComponent,
        handleDeleteComponent,
      );
    })`

But this is not working. In console there is an error with "object is not a valid child.".

packages/fluentui/react-builder/src/utils/codeToTree.ts
assuncaocharles: Would it work if we check with `React.isValidElement`?

packages/fluentui/react-builder/src/utils/codeToTree.ts
assuncaocharles: Maybe use some random id instead of `undefined` to avoid react complain about  2 items with same key ?

packages/fluentui/react-builder/tsconfig.json
assuncaocharles: What about `declare module '@material-ui/core' `? Would it be better solution than changing the module resolution @layershifter ?

packages/fluentui/react-builder/src/utils/codeToTree.ts
vyhnalekl: I don't think so, we only have access to the name, for example: 'div', 'Button'. To my understanding `React.isValidElement` is working with objects?

assuncaocharles: Yeah, only works if we get the instance itself, so let's leave it as it's for now :)

assuncaocharles (approved):
LGTM but I would wait @miroslavstastny have final review since it's a change in the builder :)

crypt096 (approved):

RubenMateus (approved):

PR #15169, opened 2020-09-22
chore: 0.51.x releases

layershifter:
A PR with 0.51.x releases. **Should be never merged to master.**


packages/fluentui/code-sandbox/package.json
```diff
@@ -1,16 +1,16 @@
 {
   "name": "@fluentui/code-sandbox",
   "description": "Fluent UI tools for CodeSandbox.",
-  "version": "0.51.0",
+  "version": "0.51.2",
```
ling1726, line 4: why bumping by 2 patch versions ?

layershifter:
Closing as we are using #16055 for tracking releases.

PR #15174, opened 2020-09-22
Bump theme package version to be 1.0.0 since it's used by other major packages

xugao:
#### Pull request checklist

#### Description of changes
`@fluentui/theme` package is currently in `v0.*`. But it's consumed as dependency from production packages like `@uifabric/styling`, `@uifabirc/fluent-theme`. From [versioning perspective](https://semver.org/spec/v1.0.0.html#how-do-i-know-when-to-release-100), it's more correct to start versioning it from `v1` instead of `v0`.

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-button-2020-09-22-12-43-20-xgao-fix-theme-package-version.json
+++ b/change/@fluentui-react-button-2020-09-22-12-43-20-xgao-fix-theme-package-version.json
@@ -1,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "set theme package version to be 1.0.0.",
+  "packageName": "@fluentui/react-button",
+  "email": "xgao@microsoft.com",
+  "dependentChangeType": "none",
+  "date": "2020-09-22T19:42:23.993Z"
+}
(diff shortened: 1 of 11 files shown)


change/@fluentui-react-button-2020-09-22-12-43-20-xgao-fix-theme-package-version.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "none",
```
ecraig12345, line 2: I think these should be patch?

xugao: these `none` packages will be bumped because it's dependency is bumped. i double checked locally using `beachball bump`

xugao: actually, i should not just think of this from the bumping perspective. this is a rare case but thinking a bit more, having patch feels more correct

ecraig12345 (approved):

kenotron:
This seems okay, but if you wanted to keep the 0.x.y, you could:

image

kenotron:
A more comprehensive fix is here: #15176 which should cover this case as well.

PR #15177, opened 2020-09-23
Updating Modal to a function component within react-next

### LINKED ISSUE #14297: Modal: strict mode compliance
Modal has a moderately complex `componentWillReceiveProps` implementation which does the following:
- starting open/close animation
- setting up dragging
- layout measurement

Given time constraints, fixing this in place rather than converting to a function component might be a better idea.

czearing:
#### Pull request checklist
- [X] Fixes #14297

#### Description of changes

1. Updating `Modal` to a function component within react-next.
2. Adding perf and vr tests for `Modal` to app package.

#### Focus areas to test
`Modal`


packages/react-next/src/components/Modal/Modal.base.tsx
```diff
@@ -209,22 +134,211 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
+      }
+
+      return delta;
+    };
+
+    const handleEnterKeyboardMoveMode = () => {
+      internalState.lastSetXCoordinate = xCoordinate;
+      internalState.lastSetYCoordinate = yCoordinate;
+      setKeyboardMoveModeTrue();
+      setModalMenuClose;
```
xugao, line 313:
```suggestion
      setModalMenuClose();
```

packages/react-next/src/components/Modal/Modal.base.tsx
```diff
@@ -209,22 +134,211 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
       ...DefaultLayerProps,
-      ...this.props.layerProps,
+      ...layerProps,
       onLayerDidMount: layerProps && layerProps.onLayerDidMount ? layerProps.onLayerDidMount : onLayerDidMount,
       insertFirst: isModeless,
       className: classNames.layer,
     };
+
+    // Allow the user to scroll within the modal but not on the body
+    const allowScrollOnModal = (elt: HTMLDivElement | null): void => {
```
xugao, line 144: we should `useCallback` on this

xugao (commented):
I know we didn't plan to convert `Dialog` to FC. But i think we should do it (after this PR).
I looked at the dialog code it's going to be a straight-forward conversion :) Mainly I also want to remove `IWithResponsiveState` prop from `Dialog` to keep it more consistent with `Modal`.

packages/react-next/src/components/Modal/Modal.base.tsx
```diff
@@ -183,12 +90,30 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
+      lastSetXCoordinate: 0,
+      lastSetYCoordinate: 0,
+      events: new EventGroup({}),
+    }));
+
+    const [isModalOpen, { setFalse: setModalClose, setTrue: setModalOpen }] = useBoolean(isOpen);
+
+    const [isModalMenuOpen, { toggle: toggleModalMenuOpen, setFalse: setModalMenuClose }] = useBoolean(false);
+
+    const [hasBeenOpened, { setTrue: setHasOpenedTrue }] = useBoolean(isOpen);
```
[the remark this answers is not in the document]
czearing: It's passed into getClassNames. When Modal is opened for the first time and the topOffsetFixed prop is true it will set root's alignItems to 'flex-start' and set main's top to modalRectangleTop.

xugao (approved):

packages/examples/src/react-next/__snapshots__/Keytips.CommandBar.Example.tsx.shot
```diff
@@ -1,674 +1,855 @@
...
+                  </i>
+                </span>
+              </button>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>,
+  <span
```
ecraig12345, line 675: Did this PR somehow accidentally change Modal so it's by default? That would be my guess from looking at this snapshot.

packages/react-next/src/components/Modal/Modal.base.tsx
```diff
@@ -183,12 +90,30 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
+      onModalCloseTimer: 0,
+      allowTouchBodyScroll: false,
+      hasRegisteredKeyUp: false,
+      scrollableContent: null,
+      lastSetXCoordinate: 0,
+      lastSetYCoordinate: 0,
+      events: new EventGroup({}),
+    }));
+
+    const [isModalOpen, { setFalse: setModalClose, setTrue: setModalOpen }] = useBoolean(isOpen);
```
ecraig12345, line 106: In this case it's probably better to just do `useState`

packages/react-next/src/components/Modal/Modal.base.tsx
```diff
@@ -183,12 +90,30 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
+
+    const [isModalMenuOpen, { toggle: toggleModalMenuOpen, setFalse: setModalMenuClose }] = useBoolean(false);
```
ecraig12345, line 108: Arguably this should also just use `useState`. Toggling can be handled without direct access to the current value by doing `setIsModalMenuOpen(prevValue => !prevValue)`. Though since you're using the toggle callback frequently I could go either way on this one.

packages/react-next/src/components/Modal/Modal.base.tsx
ecraig12345, line 110: This should definitely just do `useState`.

packages/react-next/src/components/Modal/Modal.base.tsx
```diff
@@ -183,12 +90,30 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
+
+    const [hasBeenOpened, { setTrue: setHasOpenedTrue }] = useBoolean(isOpen);
+
+    const [isVisible, { setFalse: setIsVisibleFalse, setTrue: setIsVisibleTrue }] = useBoolean(isOpen);
```
ecraig12345, line 112: Same here about useState

packages/react-next/src/components/Modal/Modal.base.tsx
```diff
@@ -183,12 +90,30 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
+
+    const [isInKeyboardMoveMode, { setFalse: setKeyboardMoveModeFalse, setTrue: setKeyboardMoveModeTrue }] = useBoolean(
```
ecraig12345, line 114: And here

ecraig12345 (commented):
There are several cases here where I think it's more appropriate to do `useState` instead of `useBoolean`. `useBoolean` is mainly meant for when you want to pass a particular setter directly as a callback. Native `useState` will be a bit cheaper and also has the benefit that you don't need to pass it as a dependency everywhere to satisfy the eslint rule (it has a hardcoded exception since it knows the setter from useState will never change).

ecraig12345:
Also a general FYI, if a PR fully fixes a particular issue you should put `Fixes #nnnn` so the issue will be automatically closed.

packages/examples/src/react-next/__snapshots__/Keytips.CommandBar.Example.tsx.shot
czearing: Thanks for mentioning this! There turned out to be a bug in `Modal's` dismissal.

packages/react-next/src/components/Modal/Modal.base.tsx
```diff
@@ -235,31 +344,70 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
             )}
           </div>
         )}
-        <div ref={this._allowScrollOnModal} className={classNames.scrollableContent} data-is-scrollable={true}>
-          {dragOptions && this.state.isModalMenuOpen && (
+        <div ref={allowScrollOnModal} className={classNames.scrollableContent} data-is-scrollable>
+          {dragOptions && isModalMenuOpen && (
             <dragOptions.menu
               items={[
```
ecraig12345, line 350: This may not be a new issue but these items ought to be memoized (and you can move the callback definitions inside the memoization, so they don't need to be memoized separately)

ecraig12345 (changes requested):
I started making changes to fix a couple things I noticed while taking a last pass over this, but then as I got into it things turned out to be more complicated than I thought (like my attempted fixes kept causing or exposing more bugs). I'll add a detailed review later but basically let's wait on this one until after we do the OUFR/react-next merge.

PR #15181, opened 2020-09-23
chore: update version of @fluentui/theme used in @uifabric/react-cards

rymeskar:
#### Pull request checklist

#### Description of changes

* Update theme package in all usages.

#### Focus areas to test

(optional)


pompomon (approved):

layershifter (approved):
This change looks correct to me and it will fix `master` builds.

Currently `@uifabric/react-cards` is a single package that has a dependency for `@fluentui/theme@0.4.0`. According to #15174 all packages should use `^1.0.0` now.

Currently `yarn` also adds an additional entry for `@fluentui/theme@0.4.0` and this is definitely not expected.

image

PR #15192, opened 2020-09-23
Callout: fix issue where it would dismiss on iframe or window focus

### LINKED ISSUE #15156: Need a prop in Callout component to prevent dismiss upon window changed

Commit https://github.com/microsoft/fluentui/pull/14363 in Callout component has introduced a regression in scenarios where the callout needs to be kept open even when window is changed. In scenarios where customer has entered some data on the callout form and is in middle of saving it, if user changes window or browser tab then user loses all the information entered. 

Hence we need a prop on the Callout component something like 'preventDismissOnWindowChanged' to prevent this. 

Callout component

Filing this issue as per suggestion from ecraig12345 in this commit thread: https://github.com/microsoft/fluentui/pull/14363

joschect:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #15156

#### Description of changes

Callouts: Fix a breaking change where new functionality would cause dismissal on giving focus to an iframe or window

#### Focus areas to test

(optional)


dotnetjunky:
This bug makes it difficult to Inspect elements in a Callout. Thanks for fixing it. Would you be able to add some kind of test to prevent this regression from happening again?

packages/office-ui-fabric-react/src/components/Callout/Callout.types.ts
```diff
@@ -102,7 +102,13 @@ export interface ICalloutProps extends React.HTMLAttributes<HTMLDivElement> {
   preventDismissOnLostFocus?: boolean;
 
   /**
-   *If defined, then takes priority over preventDismissOnLostFocus, preventDismissOnResize,
+   * If true then the callout will dismiss when the window gets focus
+   * @defaultvalue false
+   */
+  shouldDismissOnWindowFocus?: boolean;
```
jodom, line 108: @joschect should we add this prop to Callout HOCs? We are using the new behavior on ContextualMenus

joschect: It's already added! You can pass it through to the underlying callout like

```
<ContextualMenu
    calloutProps = {{shouldDismissOnWindowFocus: true }}
/>
```

jodom (approved):

ecraig12345 (approved):

khmakoto (approved):

romanisa:
Thanks @joschect for fixing the issue!

Issue #14740, opened 2020-08-25
ThemeProvider v8 pre-beta work


xugao:
Split #14661 for work that needs to be done post-snap / pre-beta:
- [x] Mark `Customizer`, `Fabric` as deprecated.
- [x] release notes
- [x] Updates to theme designer (update codepen snippet to use ThemeProvider) v7 theme
- [x] Make sure public APIs from `react-theme-provider` are correct
- [x] Export * from `theme` and `react-theme-provider` package from v8 suite package
- [x] Doc page for `ThemeProvider`
- [x] Update existing theme related website doc/wiki
  - [x] https://github.com/microsoft/fluentui/wiki/Theming
  - [x] https://github.com/microsoft/fluentui/wiki/How-to-apply-theme-to-Fluent-UI-React-components
  - [x] https://github.com/microsoft/fluentui/wiki/Fluent-Styles-Moving
  - [x] https://github.com/microsoft/fluentui/wiki/Customizing-fonts
  - [x] https://github.com/microsoft/fluentui/wiki/Server-side-rendering-and-browserless-testing

PR #15194, opened 2020-09-23
ThemeProvider: add instructions to replace Fabric, Customizer, loadTheme in README

xugao:
#### Pull request checklist

#### Description of changes
Add instructions for how to replace `Fabric`, `Customizer`, and `loadTheme` with `ThemeProvider`.

Related issue: #14740

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-theme-provider/README.md
+++ b/packages/react-theme-provider/README.md
@@ -149,13 +149,13 @@ Deprecations remain to be functional as is but they will be removed in Fluent UI
 
 Before:
 
-```
+```jsx
 <Customizer settings={{ theme }} />
 ```
 
 After:
 
-```
+```jsx
 <ThemeProvider theme={theme} />
 ```
 
@@ -163,11 +163,11 @@ After:
 
 Before:
 
-```
+```jsx
 <Customizer
   scopedSettings={{
     Checkbox: {
-      styles: CheckboxStyles
+      styles: CheckboxStyles,
     },
   }}
 />
@@ -175,7 +175,7 @@ Before:
 
 After:
 
-```
+```jsx
 <ThemeProvider
   theme={{
     components: { Checkbox: { styles: CheckboxStyles } },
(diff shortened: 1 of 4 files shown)


packages/react-theme-provider/README.md
```diff
@@ -135,3 +135,99 @@ const Foo = props => {
   return <div className={classes.root} />;
 };
 ```
+
+## How does this change other existing ways of theming Fluent UI components?
```
dzearing, line 139: It might be great to include some of this guidance in the releasenotes as well.

packages/react-theme-provider/README.md
```diff
@@ -135,3 +135,99 @@ const Foo = props => {
...
+
+### Customizer
+
+`Customizer` is now deprecated and you should replace it with `ThemeProvider`.
```
dzearing, line 143: It would be good to have a before/after code example showing how ThemeProvider can take in "components" for scoped settings.

xugao: i have that below under `#### Replace scopedSettings prop`

dzearing (approved):

packages/react-theme-provider/README.md
```diff
@@ -135,3 +135,99 @@ const Foo = props => {
...
+
+To do that, instead of calling `loadTheme(your_theme)`, you will simply wrap the root component of your React application once with `ThemeProvider`:
+
+```
+<ThemeProvider theme={your_theme}>
+  <App />
+</ThemeProvider>
+```
+
+One caveat here is that if you app has styles which relies on `@microsoft/load-themed-styles`, `ThemeProvider` won't be able to replace `loadTheme` in this case.
```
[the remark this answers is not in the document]
xugao: agree this doesn't seem to be the most ideal place. These are mainly purposed as our release notes. 
I am going to do more wiki/doc cleanup next week (tracked in #14740). I am going to start here and re-evaluate as I go.

PR #15196, opened 2020-09-23
Adding an elementRef to v7 BaseButton component

czearing:
#### Pull request checklist

#### Description of changes
Adding an elementRef to office-ui-fabric-react `BaseButton` to allow temporary access to the root DOM element in the compat `Button` components.

#### Focus areas to test
`Button`


packages/office-ui-fabric-react/src/components/Button/Button.test.tsx
```diff
@@ -122,6 +123,13 @@ describe('Button', () => {
       expect(button.tagName).toEqual('A');
     });
 
+    it('can handle elementRef', () => {
+      const ref = React.createRef<HTMLElement>();
+      safeMount(<DefaultButton elementRef={ref}>Content</DefaultButton>, button => {
+        expect(button.find('elementRef.current')).toBeTruthy();
```
dzearing, line 129:
```suggestion
        expect(ref.current).toBeTruthy();
```

khmakoto (approved):
It seems like you need to update the API.

PR #15199, opened 2020-09-23
Add role prop to breadcrumb item

erichdev:
#### Pull request checklist

#### Description of changes

Add optional `role` prop to `IBreadcrumbItem` that gets passed to the `<Link>`. Breadcrumb items render as a button by default, or as a link if an `href` is passed in. This prop will allow consumers to give breadcrumb items a role of `link`, for example, while still rendering it as a button.

### DIFF
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -1830,6 +1830,7 @@ export interface IBreadcrumbItem {
     isCurrentItem?: boolean;
     key: string;
     onClick?: (ev?: React.MouseEvent<HTMLElement>, item?: IBreadcrumbItem) => void;
+    role?: string;
     text: string;
 }
 
--- a/change/@fluentui-react-next-2020-09-24-11-02-17-breadcrumbitem-role.json
+++ b/change/@fluentui-react-next-2020-09-24-11-02-17-breadcrumbitem-role.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Breadcrumb: add role prop for items",
+  "packageName": "@fluentui/react-next",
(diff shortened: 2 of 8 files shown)


packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -1768,6 +1768,7 @@ export interface IBreadcrumbItem {
     isCurrentItem?: boolean;
     key: string;
     onClick?: (ev?: React.MouseEvent<HTMLElement>, item?: IBreadcrumbItem) => void;
+    role?: string;
```
erichdev, line 1771: Looking at this now, I think maybe it'd be better to add this role to `IBreadcrumbProps`? I don't know if there'd ever be a scenario where items within the same breadcrumb would have different roles

xugao: i do think it's more suitable to add it here, as what you have now

xugao:
could you pls make the same fix in `react-next` package? thanks!
https://github.com/microsoft/fluentui/blob/master/packages/react-next/src/components/Breadcrumb/Breadcrumb.types.ts#L110
https://github.com/microsoft/fluentui/blob/master/packages/react-next/src/components/Breadcrumb/Breadcrumb.base.tsx#L258

erichdev:
> could you pls make the same fix in `react-next` package? thanks!
> https://github.com/microsoft/fluentui/blob/master/packages/react-next/src/components/Breadcrumb/Breadcrumb.types.ts#L110
> https://github.com/microsoft/fluentui/blob/master/packages/react-next/src/components/Breadcrumb/Breadcrumb.base.tsx#L258

I've pushed changes to react-next. Thanks!

xugao (approved):
LGTM, thanks!

PR #15201, opened 2020-09-24
Fail build on changed files and do more cleanup

ecraig12345:
At some point the "check for changed files" build step got lost, so add it back where appropriate to help detect issues (like the missing web-components API update).

We've also still had some trouble with "Directory not empty" errors, so add "workspace: clean: all" a couple more places AND add back the DeleteFiles step everywhere for good measure.


azure-pipelines.yml
```diff
@@ -108,6 +110,8 @@ jobs:
           SourcePath: 'packages/fluentui/docs/src/public'
           storage: fabricweb
 
+      - template: .devops/templates/cleanup.yml
```
xugao, line 113: do we need to check change here and below? shouldn't checking in buildtestlint step sufficient?

xugao (approved):

azure-pipelines.yml
ecraig12345: Maybe not totally necessary, but I think I'd rather keep it just to be safe.

PR #15205, opened 2020-09-24
DRAFT - Multiselect tree as list behavior - using role listbox + option

### LINKED ISSUE #15041: Add multi select support for Tree as a list

YuanboXue-Amber:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15041 

#### Description of changes
 
This is an attempt to solve the same problem as PR #15147, but at the same time following aria guidelines.

The **difference** between this PR and PR #15147: in PR #15147, `treeItem` role is used within `listbox` role. This PR removed it.

Try the [example](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/15205/merge/react-northstar/maximize/tree-multiselect-as-list-example-shorthand/false) with screen readers, I discovered the problems below:

- When navigate within the tree (ctrl+option+shift+down), and focus is on the first treeItem, VO still narrates:  'In multi select tree as list, list box'. 
Another ctrl+option+shift+down will make VO narrates: 'House Lannister, text element'. It doesn't mention that 'House Lannister' is expandable.
- When using ctrl+option+right to navigate thru the tree, from 'House Lannister' -> 'Tywin' -> 'Kevan', focus is trapped on 'Tywin'

I have also tried to tweak the roles under aria guidelines, but I could not get rid of the above problems. I put what I have tried so far in the comments below.

#### Focus areas to test

(optional)


YuanboXue-Amber:
Making the following change: use `group` on parent treeItems that are not selectable.
Result in similar problems, so I'm not going to open a separate for it.

Screenshot 2020-09-24 at 11 46 18

YuanboXue-Amber:
Making the following change: use group on parent treeItems that are not selectable, and `option` on all selectable treeItems regardless of if it has subtree or not.
Result in similar problems, so I'm not going to open a separate PR for it.

Screenshot 2020-09-24 at 11 46 18

YuanboXue-Amber:
Close this PR, and continue on the related PR #15147

PR #15216, opened 2020-09-24
Removing findDOMNode usage from Pivot

### LINKED ISSUE #14322: Pivot (next): remove findDOMNode usage
The new Pivot control in `react-tabs` uses [`ReactDOM.findDOMNode`](https://reactjs.org/docs/react-dom.html#finddomnode) to find the overflow button. `findDOMNode` is [deprecated in strict mode](https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage) and does not work with function components, meaning this call will stop working once the component is updated to use the function component version of Button.

czearing:
#### Pull request checklist
- [X] Fixes #14322

#### Description of changes
1. Adding an elementRef to overflow button in Pivot
2. Removing setOverflowMenuButtonRef
3. Passing overflowMenuButtonRef as an elmentRef to `CommandButton`

#### Focus areas to test
`Pivot`


ecraig12345 (approved):

ecraig12345:
It looks like this might actually have caused an issue, based on this screener result:
image

behowell (changes requested):
It looks like overflow is broken by this change, according to the screener result that Elizabeth pasted. Nothing in this PR immediately stands out as wrong; but perhaps Button's elementRef property isn't working as expected?

czearing:
Here's what was done to fix `Pivot's` overflow behavior after removing findDOMNode:
1.	The `BaseButton’s` ref was expected to go to the root element but it was being unnecessarily wrapped by a span. This was removed and replaced with a fragment.
2.	In Pivot.styles the star in the selector was removed as the overflow button had its visibility set to hidden after `BaseButton's` span was removed.

packages/office-ui-fabric-react/src/components/Button/BaseButton.tsx
```diff
@@ -353,10 +353,10 @@ export class BaseButton extends React.Component<IBaseButtonProps, IBaseButtonSta
 
     if (menuProps && menuProps.doNotLayer) {
       return (
-        <span style={{ display: 'inline-block' }}>
+        <>
```
ecraig12345, line 356: This seems pretty risky to do in existing widely-used code. @khmakoto or @dzearing (if you weren't the one to suggest this) what do you think?

khmakoto: Yeah, I'd say we shouldn't change it in case someone is targeting that specific element in the DOM.

khmakoto: Ok, spoke to David offline and the reason for the span wrapper's existence is because this was written before React fragments existed and causes pretty messy bugs since the DOM structure is not inline with normal button. This is indeed a breaking change but it's fundamentally wrong to keep it as is.

My proposal would be to make the change but hold off on merging this PR until we are in the beta phase so that it merges into the compat layer in v8. And since it's going into v8 vs v7 there is less risk in introducing a breaking change.

dzearing: yup let's go with that, thanks all!

ecraig12345: The fragment change should be unblocked now that 7 has forked off, just be sure to add it in the release notes.

Issue #14888, opened 2020-09-03
Convert ComponentConformance and ComponentExamples tests to run with react-conformance


ecraig12345:
OUFR's legacy ComponentConformance and ComponentExamples tests could be re-written to run as `extraTests` via `isConformant` from `@fluentui/react-conformance`. The redundant top-level component file test can also be removed.

Since we want these same tests to run across multiple small component packages, it might make sense to put them in the `react-conformance` package but not run them by default.

(Making an issue for this so it doesn't get lost, but not including it in the v8 project since it's not essential.)

PR #15218, opened 2020-09-24
Implementing component-handles-classname test within react-conformance

czearing:
#### Pull request checklist
- [X] Addresses an existing issue: #14888

#### Description of changes

1. Adding a `component-handles-classname` test to isConformant's  defaultTests.
2. Removing redundant tests in v7 and v8 ComponentConformance.test.tsx
3. Removing redundant tests in react-northstar isConformant tests.
4. Disabling tests that don't handle classNames correctly. (Either they aren't applied to the root element or not passing classNames correctly. )

#### What this test does:
1. Checks if the component handles className prop.
2. Checks if the component handles any existing default provided classNames correctly.
3. Checks if the component puts the className on the root element.

#### List of removed `react-northstar` tests:
- #### ( Basic Tests ) replaced with `name-matches-filename`, `exports-component`:    	* Component exports    	* Component has a constructor name
- #### ( ClassName Tests ) replaced with `component-handles-classname`:
     * ClassName applies to root element
     * ClassName applies to root component
     * User ClassName does not override the default classes

#### List of removed `ComponentConformance` tests:
- #### ( Component File Conformance ) replaced with `component-handles-classname`:
     * Component injects a className prop
- #### ( Top Level Component File Conformance ) replaced with `exported-top-level`, `has-top-level-file`:
     * Component has a top level file for each file in the directory

#### date-time
* `CalendarDayGrid`: Doesn't apply the test className to the root.

#### react-next	
* `Breadcrumb`: Doesn't apply the test className to the root.
* `Callout`: Doesn't apply the test className to the root.
* `ComboBox`: Doesn't work as expected with mount.
* `ContextualMenu`: Requires createPortal.
* `Rating`: Doesn't apply the defaultClassName to the root.
* `SearchBox`: Doesn't apply the test className to the root.
* `Callout`: Doesn't apply the test className to the root.
* `BaseSelectedItemsList`: Doesn't apply the test className to the root.
* `SelectedPeopleList`: Doesn't apply the test className to the root.
* `ButtonGrid`: Doesn't apply the test className to the root.
* `Modal`: Requires createPortal.
* `PersonaPresence`: Doesn't apply the test className to the root.

#### office-ui-fabric-react	
* `Announced`: Doesn't apply the test className to the root.
* `Breadcrumb`: Doesn't apply the test className to the root.
* `Coachmark`: Doesn't apply the test className to the root.
* `Dialog`: Refers to modalProps.className.
* `Nav`: Doesn't apply the test className to the root.
* `Panel`: Doesn't apply the test className to the root.
* `Sticky`: Doesn't apply the test className to the root.
* `SelectionZone`: Doesn't apply the test className to the root.


dzearing:
@layershifter and @miroslavstastny heads up here :)

ecraig12345:
Thanks for doing this! This one should probably wait to merge until next week so Shift or Miro can look at it and verify the northstar changes are okay with them.

ecraig12345 (changes requested):
Let's wait on this one until Shift or Miro can look, and until the worst of the code moves are over.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -66,6 +66,33 @@ export const defaultTests: TestObject = {
...
+        className: 'testComponentClassName',
+      };
+
+      const el = passesUnhandledPropsTo
+        ? customMount(<Component {...mergedProps} />).find(passesUnhandledPropsTo)
+        : customMount(<Component {...mergedProps} />);
+
+      const component = getComponent(el, helperComponents, wrapperComponent);
+
+      expect(component.find('[testComponentClassName]'));
```
layershifter, line 92: This asserts nothing...

layershifter: In v0 we are testing at least one scenario: that passed `className` will not override a default one

```jsx
function Button(props) {
  return <button className='ui-button' {...props} />
}

<Button /> // has ui-button
<Button className="foo" /> // has only foo, that's not correct
```

czearing: @layershifter Thanks for the suggestion! I just updated the pr to handle default classNames on components.

packages/fluentui/react-northstar/test/specs/components/Form/FormButton-test.tsx
```diff
@@ -8,5 +8,6 @@ describe('FormButton', () => {
     constructorName: 'FormButton',
     forwardsRefTo: `Button`,
     passesUnhandledPropsTo: Button,
+    disabledTests: ['component-handles-classname'],
```
layershifter, line 11: Can you please clarify why this is disabled? The same test worked before and now is broken? That looks suspicious 🤔

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -66,6 +66,56 @@ export const defaultTests: TestObject = {
     });
   },
 
+  /** Component file handles classname prop */
+  'component-handles-classname': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    it(`handles className prop`, () => {
```
layershifter, line 71: I suggest to split this to two separate tests (`it`)

packages/fluentui/react-northstar/test/specs/components/Form/FormButton-test.tsx
czearing: I investigated the disabled Form tests and they are now all enabled. Here is what happened:

Reason Why It Failed:
•	The `passessUnhandledPropsTo` type was acting incorrectly in `component-handles-classname` and was behaving as an incorrect workaround for components that didn’t apply classNames to the root.

Solution:
•	I decided to remove usage of `passessUnhandledPropsTo` in the classNames test since it was only added to OverflowSet.

•	After removing the usage of `passessUnhandledPropsTo` I fixed the bug in `OverflowSet` (placement of `getNativeProps`) and enabled the test.

packages/react-next/src/components/OverflowSet/OverflowSet.base.tsx
```diff
@@ -49,13 +49,14 @@ export const OverflowSetBase: React.FunctionComponent<IOverflowSetProps> = React
 
   const showOverflow = !!overflowItems && overflowItems.length > 0;
 
+  // console.log('hello 831289731289732189713289712389713298731298');
```
layershifter, line 52: Oops

```suggestion
```

layershifter (approved):

layershifter:
Impressive 👍 Thanks for working on this ❤️

PR #15219, opened 2020-09-24
Fix incorrect hover behavior during menu animation

### LINKED ISSUE #14017: ContextualMenu submenu and Button submenu hover state flicker on open
### Environment Information

- **Package version(s)**: 7.122.0
- **Browser and OS versions**: Chromium Edge: Version 83.0.478.61 (Official build) (64-bit)

Repros on https://fabricweb.z5.web.core.windows.net/oufr/7.122.0/#/examples/contextualmenu `ContextualMenu with checkable menu items and toggleable split button` example

I've also recorded a video showing this: https://microsoft-my.sharepoint-df.com/:v:/p/owcampbe/EQrBDkNaPDJBr8yHmYOe__cBHV7u54d4NOO9IbbV-1m5LQ?e=HeXj5Z

#### Actual behavior:

Hovering over the right edge of the submenu button causes the newly opened submenu to receive hover states for a short time before returning to the submenu button.

This also repros using an `IconButton` that has a submenu

#### Expected behavior:

Submenu opens with no hover states if the mouse isn't over it.

Requested priority: Normal

behowell:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14017

#### Description of changes

Fix incorrect hover behavior during menu animation. Set `pointer-events: none` in the move animation keyframes to disable mouse events during the animation.

#### Focus areas to test

Slide animations


joschect (approved):

packages/theme/src/motion/AnimationStyles.ts
```diff
@@ -155,15 +155,15 @@ function _createAnimation(
 
 function _createSlideInX(fromX: number): string {
   return keyframes({
-    from: { transform: `translate3d(${fromX}px,0,0)` },
-    to: { transform: `translate3d(0,0,0)` },
+    from: { transform: `translate3d(${fromX}px,0,0)`, pointerEvents: 'none' },
+    to: { transform: `translate3d(0,0,0)`, pointerEvents: 'auto' },
```
behowell, line 159: I think setting it back to `auto` at the end isn't really necessary, since it should get reverted to its previous value once the animation ends. But I'm new to css animations and wasn't sure that was guaranteed, so I added this just in case. I can remove this if it is in fact guaranteed.

xugao (approved):
Nice fix Ben!

ecraig12345 (approved):

packages/theme/src/motion/AnimationStyles.ts
brandonthomas: This has regressed some features within M365 where pointer events none was actually required. This took priority. We should revert this change.

brandonthomas:
Filing an issue as this caused regressions when using isBlocking = false on panel. Basically, it's always blocking now no matter what.

PR #15221, opened 2020-09-24
DetailsList: address onItemInvoked bug by instructing FocusZone to avoid eating the enter keydown event

### LINKED ISSUE #14947: DetailsList onItemInvoked callback should be called when pressing Enter.
In documentation it is said that onItemInvoked is a "Callback for when a given row has been invoked (by pressing enter while it is selected.)" But in reality onItemInvoked is invoked only when the item is double clicked but not when Enter key is pressed. The callback should be invoked when Enter key is pressed (or consider making another callback that is invoked by Enter key).

dzearing:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #14947

#### Description of changes

A previous change to DetailsList re-ordered SelectionZone and FocusZone, which broke SelectionZone properly handling the enter keydown event. Partially reverting this change.

Added a unit test to cover so this doesn't bust again.


packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.Basic.Example.tsx
```diff
@@ -109,6 +109,6 @@ export class DetailsListBasicExample extends React.Component<{}, IDetailsListBas
   };
 
   private _onItemInvoked = (item: IDetailsListBasicExampleItem): void => {
-    alert(`Item invoked: ${item.name}`);
+    console.log(`Item invoked: ${item.name}`);
```
JustSlone, line 112: do we need to update any docs that refer to this being an alert? i.e. is there text that says "double click to see an alert". I'm guessing no, but thought I would ask.

JustSlone (approved):
Looks good to me, thanks for adding tests!

packages/office-ui-fabric-react/src/components/DetailsList/examples/DetailsList.Basic.Example.tsx
dzearing: done. :)

packages/office-ui-fabric-react/src/components/DetailsList/DetailsList.base.tsx
```diff
@@ -505,6 +505,7 @@ const DetailsListInner: React.ComponentType<IDetailsListInnerProps> = (
     direction: FocusZoneDirection.vertical,
     shouldEnterInnerZone: isRightArrow,
     onActiveElementChanged: onActiveRowChanged,
+    shouldRaiseClicks: false,
```
dzearing, line 508: This is the 1 line fix that resolves this.

ThomasMichon (approved):

brandonthomas:
Thanks a million for the quick turn around on this one guys!

PR #15225, opened 2020-09-24
Fix issues with drag item index

elisabethcvs:
#### Pull request checklist

#### Description of changes

- Was attempting to assign the drag index to be -1 if the incoming index was null, accidently was changing it when it's 0 as well. Fixed that.
- For onDrop, getAsString is an async callback, so I couldn't have the insertIndex as an onDrop. So now it's a global variable.

### DIFF
--- a/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
+++ b/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
@@ -143,6 +143,7 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
   };
 
   const _onDragStart = (item?: any, itemIndex?: number, tempSelectedItems?: any[], event?: DragEvent): void => {
+    /* eslint-disable-next-line eqeqeq */
     const draggedItemIndex = itemIndex != null ? itemIndex! : -1;
     setDraggedIndex(draggedItemIndex);
     if (event) {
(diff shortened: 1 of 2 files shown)


packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -138,12 +140,13 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
         : [selectedItems[draggedIndex]];
-      _dropItemsAt(insertIndex, newItems);
+      _dropItemsAt(newItems);
     }
   };
 
   const _onDragStart = (item?: any, itemIndex?: number, tempSelectedItems?: any[], event?: DragEvent): void => {
-    const draggedItemIndex = itemIndex ? itemIndex! : -1;
+    /* eslint-disable-next-line eqeqeq */
+    const draggedItemIndex = itemIndex != null ? itemIndex! : -1;
```
amyngu, line 149: what does the ! for itemIndex**!** do?

elisabethcvs: it says yes really use this I believe. I needed it because it's got the ? on it. 

P.S. I'm great at technical descriptions

amyngu: Interesting! Don't think I've used this before thanks for the explanation :P

amyngu (approved):

PR #15226, opened 2020-09-24
Fixing storybook build task that was preventing PR deploy sites from being bundled correctly

khmakoto:
#### Pull request checklist

#### Description of changes

This PR fixes the storybook build task that was preventing PR deploy sites from being bundled correctly

#### Focus areas to test

(optional)

### DIFF
--- a/scripts/just.config.ts
+++ b/scripts/just.config.ts
@@ -4,6 +4,7 @@ const { task, series, parallel, condition, option, argv, addResolvePath, resolve
 
 const path = require('path');
 const fs = require('fs');
+const findGitRoot = require('./monorepo/findGitRoot');
 
 const { clean } = require('./tasks/clean');
 const { copy } = require('./tasks/copy');
@@ -46,15 +47,14 @@ function basicPreset() {
 
 /** Resolve whereas a storybook config + stories exist for a given path */
 function checkForStorybookExistence() {
-  const dir = process.cwd().replace(/\\/g, '/');
-  const packageName = dir.substring(dir.lastIndexOf('/') + 1);
+  const packageName = path.basename(process.cwd());
 
   // Returns true if the current package has a storybook config or the examples package has a storybook config and
   // contains a folder with the current package's name.
   return (
     !!resolveCwd('./.storybook/main.js') ||
     (!!resolveCwd('../examples/.storybook/main.js') &&
-      fs.existsSync(path.join(process.cwd(), `../examples/${packageName}`)))
+      fs.existsSync(path.join(findGitRoot(), `packages/examples/src/${packageName}`)))
   );
 }
 
--- a/scripts/just.config.ts
+++ b/scripts/just.config.ts
@@ -44,6 +44,20 @@ function basicPreset() {
   option('push', { default: true });
 }
 
+/** Resolve whereas a storybook config + stories exist for a given path */
+function checkForStorybookExistence() {
+  const dir = process.cwd().replace(/\\/g, '/');
+  const packageName = dir.substring(dir.lastIndexOf('/') + 1);
+
+  // Returns true if the current package has a storybook config or the examples package has a storybook config and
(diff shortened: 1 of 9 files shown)


scripts/just.config.ts
```diff
@@ -44,6 +44,20 @@ function basicPreset() {
   option('push', { default: true });
 }
 
+/** Resolve whereas a storybook config + stories exist for a given path */
+function checkForStorybookExistence() {
+  const dir = process.cwd().replace(/\\/g, '/');
+  const packageName = dir.substring(dir.lastIndexOf('/') + 1);
```
ecraig12345, line 50: You can do this with `path.basename(dir)`

scripts/just.config.ts
```diff
@@ -44,6 +44,20 @@ function basicPreset() {
...
+
+  // Returns true if the current package has a storybook config or the examples package has a storybook config and
+  // contains a folder with the current package's name.
+  return (
+    !!resolveCwd('./.storybook/main.js') ||
+    (!!resolveCwd('../examples/.storybook/main.js') &&
+      fs.existsSync(path.join(process.cwd(), `../examples/${packageName}`)))
```
ecraig12345, line 57: Probably better to use the path from the repo root (`findGitRoot()`) here to be safe

scripts/storybook/webpack.config.js
```diff
@@ -74,7 +74,12 @@ module.exports = (/** @type {webpack.Configuration} */ config) => {
 
   config.resolve.extensions.push('.ts', '.tsx');
 
-  config.resolve.alias = getResolveAlias();
+  config.resolve.alias = {
+    ...getResolveAlias(),
+    '@fluentui/examples$': path.join(__dirname, '../../packages/examples/src'),
```
ecraig12345, line 79: Not super important but it's better to use `path.resolve` to get a path without the `../`

scripts/tasks/storybookTask.js
```diff
@@ -41,10 +41,14 @@ module.exports.buildStorybookTask = function buildStorybookTask(options) {
...
 
+    const localConfigDir = path.join(process.cwd(), '.storybook');
+
     await storybook({
       mode: 'static',
       staticDir: [path.join(process.cwd(), 'static')],
-      configDir: path.join(process.cwd(), '.storybook'),
+      configDir: fs.existsSync(localConfigDir)
+        ? localConfigDir
+        : path.resolve(__dirname, '../../packages/examples/.storybook'),
```
ecraig12345, line 51: Again it would be a little cleaner to do this from `findGitRoot()`

ecraig12345 (commented):
Thanks for figuring this out! A few suggestions.

dzearing (commented):
Works locally, A++++++ would recommend

dzearing (approved):

ecraig12345 (approved):
Thanks!

scripts/tasks/storybookTask.js
```diff
@@ -41,14 +42,19 @@ module.exports.buildStorybookTask = function buildStorybookTask(options) {
...
       staticDir: [path.join(process.cwd(), 'static')],
-      configDir: path.join(process.cwd(), '.storybook'),
+      configDir: fs.existsSync(localConfigDir)
+        ? localConfigDir
+        : path.join(findGitRoot(), 'packages/examples/.storybook'),
       outputDir: path.join(process.cwd(), 'dist-storybook'),
       quiet,
       port: port || 3000,
       ci,
+      debugWebpack: true,
```
ecraig12345, line 57: Was this supposed to be checked in?

khmakoto: Nope, I'm still dealing with a bunch of deploy issues and I'm using that to debug. I think it got pushed as I tried to get something working for David, but rest assured that I'll remove it by the time I merge.

PR #15227, opened 2020-09-24
Make autofill compliant

### LINKED ISSUE #14286: Autofill: strict mode compliance (UNSAFE_)
Autofill has 2 main issues preventing strict mode compliance. TBD if these can be addressed without conversion to a function component.

1. It updates `state.displayValue` in `componentWillReceiveProps`. This could probably be moved to `componentDidUpdate`.

2. It has an actual prop called `updateValueInWillReceiveProps`... This appears to have been added specifically for ComboBox and is *highly* unlikely to be used anywhere else. Potentially the logic could be moved to `componentDidUpdate` (and the prop docs updated), but it would be better if we can find a way to remove it entirely.

joschect:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #14286

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


packages/react-next/src/components/Autofill/Autofill.tsx
```diff
@@ -53,7 +65,7 @@ export class Autofill extends React.Component<IAutofillProps, IAutofillState> im
   }
 
   public get value(): string {
-    return this._value;
+    return this.state.inputValue || (this.props.value as string) || '';
```
behowell, line 68: It looks like `InputHTMLAttributes` `value` prop can be `string | string[] | number`. If Autofill only supports `string`, then you should add it to `IAutofillProps` as `value?: string;` to restrict its type.

joschect: I tried doing this and there is a surprising amount (type wise) that relies on this being `string | string[] | number`. I think for this release we should leave it as is to minimize the breaks that it could cause. @behowell do you have thoughts?

behowell: That makes sense. But in that case, then this function should also handle the other types that value might be: `number` or `string[]`.

ecraig12345: I'm not sure how it could actually end up as any of those other types since the autofill input will always just be a text field. I suspect InputHTMLAttributes covers a broader range of components.

ecraig12345: We should probably explicitly add `value?: string` to `IAutofillProps` instead of using the default typing from `InputHTMLAttributes`.

behowell: Yes, that was my initial suggestion :)

ecraig12345: Hmm seems I can't read today, sorry! So there are now two votes for an explicit typing. :)

joschect: Yeah, I had tried that, there are a variety of things that rely on HTMLinputAttrbutes (or whatever that is) and that means that that set of props can't be passed right into the autofill. I think we can do that, but we would need to change it for a variety of thigns.

joschect: @ecraig12345 Ben and I worked on a solution and we're going to keep that signature for now but handle the cases and warn for it.

packages/react-next/src/components/Autofill/Autofill.tsx
```diff
@@ -308,42 +298,64 @@ export class Autofill extends React.Component<IAutofillProps, IAutofillState> im
...
+    const { onInputChange } = this.props;
+    if (onInputChange) {
+      return onInputChange?.(newValue, composing) || '';
+    }
+    return newValue;
+  };
+
+  private _getControlledValue(): string | undefined {
+    const { value } = this.props;
+    if (value === undefined || typeof value === 'string') {
```
ecraig12345, line 331: Can you also update the type of `value` in the types file? That's also good for documentation (people get confused when essential props are implicitly inherited).

ecraig12345: I'm still not sure in what cases it would even be possible/reasonable for the value to be another type, given the expected usage of this component.

joschect: Yeah I'll add it.

packages/react-next/src/components/Autofill/Autofill.types.ts
```diff
@@ -47,6 +47,13 @@ export interface IAutofill {
  * {@docCategory Autofill}
  */
 export interface IAutofillProps extends React.InputHTMLAttributes<HTMLInputElement | Autofill> {
+  /**
+   * The same as an Input's value property.
+   * While it can be any of these types, only a string should be passed in.
+   * The rest of the types (string[], number) are currently provided to keep this inline
+   * with the type provided by InputHTMLAttributes but may be removed at a later time
+   */
+  value?: string | string[] | number | undefined;
```
ecraig12345, line 56: Sorry to keep belaboring this, but I still don't think there's any reason to keep the rest of the types.

ecraig12345: We have no reasonable expectation that anyone should be using them (and especially with this being a major version, I'm fine with introducing a compiler error and making them fix it if they are), and including them just makes the definition unnecessarily confusing.

joschect: I am going to remove this definition then so we can break it easily in a future release but I'm going to keep the native input types. I think that there is already going to be a fair amount of churn and this is a nice interim that doesn't break people much. It preserves the existing behavior since this isn't one of the components that we're refreshing.

ecraig12345 (approved):

Issue #15189, opened 2020-09-23
Area chart is not refreshed when chart data is updated


luckyrass:
### Environment Information

- **Package version(s)**: (4.2.0)
- **Browser and OS versions**: ()

### Please provide a reproduction of the bug in a codepen:

Area chart is not refreshed when chart data is updated

Code to repro it
```
import * as React from 'react';
import { AreaChart } from '@uifabric/charting';
import { ILineChartProps } from '@uifabric/charting';
import { DefaultPalette } from 'office-ui-fabric-react/lib/Styling';

interface IAreaChartBasicState {
  width: number;
  height: number;
  chartData: any;
}

const chart1Points = [
  {
    x: 20,
    y: 7000,
    xAxisCalloutData: '2018/01/01',
    yAxisCalloutData: '10%',
  },
  {
    x: 25,
    y: 9000,
    xAxisCalloutData: '2018/01/15',
    yAxisCalloutData: '18%',
  },
  {
    x: 30,
    y: 13000,
    xAxisCalloutData: '2018/01/28',
    yAxisCalloutData: '24%',
  },
  {
    x: 35,
    y: 15000,
    xAxisCalloutData: '2018/02/01',
    yAxisCalloutData: '25%',
  },
  {
    x: 40,
    y: 11000,
    xAxisCalloutData: '2018/03/01',
    yAxisCalloutData: '15%',
  },
  {
    x: 45,
    y: 8760,
    xAxisCalloutData: '2018/03/15',
    yAxisCalloutData: '30%',
  },
  {
    x: 50,
    y: 3500,
    xAxisCalloutData: '2018/03/28',
    yAxisCalloutData: '18%',
  },
  {
    x: 55,
    y: 20000,
    xAxisCalloutData: '2018/04/04',
    yAxisCalloutData: '32%',
  },
  {
    x: 60,
    y: 17000,
    xAxisCalloutData: '2018/04/15',
    yAxisCalloutData: '29%',
  },
  {
    x: 65,
    y: 1000,
    xAxisCalloutData: '2018/05/05',
    yAxisCalloutData: '43%',
  },
  {
    x: 70,
    y: 12000,
    xAxisCalloutData: '2018/06/01',
    yAxisCalloutData: '45%',
  },
  {
    x: 75,
    y: 6876,
    xAxisCalloutData: '2018/01/15',
    yAxisCalloutData: '18%',
  },
  {
    x: 80,
    y: 12000,
[...]

Issue #15222, opened 2020-09-24
Contribution process changes during version 8 beta


ecraig12345:
## Motivation

We can't [publish the verison 8 beta](https://github.com/microsoft/fluentui/issues/12770) until v7 code in `office-ui-fabric-react` and v8 breaking changes in `@fluentui/react-next` are merged into `@fluentui/react`. But we can't merge them in the same branch as active 7.x development. 

We also need a branch for 7.x fixes while version 8 (beta or release) is being published out of `master`.

## Overview

**Starting on September 28**, move 7.x development into new `7.0` branch (#14684), which will later become the hotfix branch. Then move code around in `master` (#14789, #14679, #13384) and publish the beta from there (#14689) **by October 12**.

~~While code is being moved around, we encourage contributors to **wait to submit PRs which aren't urgently needed** due to the amount of churn that will be happening in the `master` branch. (Exception: PRs related to `@fluentui/react-northstar` and `@fluentui/web-components` are okay since these packages are mostly isolated from the code moves.)~~ *(this is now finished)*

(Since this is our first release in over a year and we have considerably more contributors and packages than last time, we're still working on figuring out a process that works reasonably well for everyone--so your feedback is appreciated!)

### Timelines

| Stage | Date |
| -------- | -------- |
| v8 beta prep ("snap") | September 28 - October 11 (or earlier) |
| v8 beta  | Started ~October 12 |
| v8 official release  | TBD |

### Summary by timeframe and package

**Update: As of January 2021, we're in the "v8 beta release" phase.**

|  | ~~v8 beta prep~~ | v8 beta release | v8 official release |
| -------- | -------- | -------- |-------- |
| `@fluentui/react-northstar` contributors |  ~~`master`~~ | `master`  | `master` |
| `@fluentui/web-components` contributors |  ~~`7.0` (to continue daily release cadence)~~ | `master`  | `master` |
| contributors to **all other packages** | ~~Non-urgent changes should wait. Urgent changes go to `7.0` branch, and maintainers will cherry-pick to `master` when beta prep work is done. (Changes in `master` *will NOT be published* during this time.)~~ | All changes go to `master` first; contributor cherry-picks to `7.0` only if needed. (Changes made in `master` will be published under *beta versions only*, so to publish under the current major version *you must cherry-pick to `7.0`.*) | All changes go to `master` first; contributor cherry-picks to `7.0` only if needed for v7 (same process as v6 fixes) | 

[...]

PR #15235, opened 2020-09-25
Charting: Area chart - Graph update after data change issue resolved and memoize implemented

jameelakowsar:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #https://github.com/microsoft/fluentui/issues/15189

#### Description of changes

After props data updated, need to update stackedValues and CalloutPoints which are using to redraw the graph. These data updated.

Memoize concept implemented - Previously, calculating stacked data, getting keys, colors, creating dataset all these will be calculated for every render which is not required and take time. Now all these will be calculated only when prop changes. 

#### Focus areas to test

Area chart

#### Before Fix, After clicking on update data (Graph is same as previous)
image

#### After fix, Updated graph.
image

Test link: http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/15235/merge/charting/dist/index.html#/examples/areachart


ecraig12345 (changes requested):
Thanks for submitting this change, but due to work we're currently doing to prepare `master` for our version 8 beta release, we're asking contributors to either wait a couple weeks to submit changes (if it's not urgent) or submit to the new `7.0` branch (if it's urgent). See #15222 for more details, and sorry about the inconvenience.

PR #15241, opened 2020-09-25
chore: fix unstable VR tests in N*

layershifter:
This PR tries to make three more tests stable in Screener to avoid phantom regressions.


packages/fluentui/docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.steps.ts
```diff
@@ -15,6 +15,10 @@ const config: ScreenerTestsConfig = {
         .click(selectors.popupTrigger)
         .click(selectors.toggleIndicator)
         .hover(selectors.item(2))
+
+        // A hack to load images properly in Screener
+        .wait(500)
```
miroslavstastny, line 20: 😢

miroslavstastny (approved):

Issue #13384, opened 2020-05-28
Move @uifabric packages under @fluentui scope


ecraig12345:
Part 3/4 in a series of changes that must happen in master *after* v7 development moves to a branch, but *before* beta publishing:
1. #15152 OUFR/react-next merge 
2. #14679 Move date-time components
3. (this) #13384 uifabric/fluentui package renames
4. #14689 set up beta releases

As part of version 8, we'd like to rename our `@uifabric` packages to `@fluentui` (or in a few cases, delete unneeded packages going forward).

~~For packages which are used by the suite, we'll add `@uifabric` aliases (which re-export the `@fluentui` version) so that partners who depend on multiple suite versions won't get multiple versions of all the dependency packages. (Creation of the alias packages will be done at a slightly later point after the renames, but before release--tracked by #15150.)~~

**EDIT:** Trying to handle package renames and getting rid of major-lockstep at the same time turned out to be too complex, so we will be continuing with major-lockstep for now and **NOT** making any alias packages. However, we will likely provide a codemod to help with renames (see #15661).

In some cases, we're changing the names of the `@fluentui` versions of the packages to better match current naming conventions, or to reserve `@fluentui` names for future use.

### Packages to rename (final list)

The table below has been updated with the final list of new package names/scopes.

 | Old name (under `@uifabric`)   | New name (under `@fluentui`) | PR     |
| ------------------------------ | ---------------------------- | ------ |
| `api-docs`                     | same                         | #15617 |
| `azure-themes`                 | same                         | #15454 |
| `charting`                     | `react-charting`             | #15564 |
| `date-time`                    | `react-date-time`            | #15659 |
| `example-app-base`             | `react-docsite-components`   | #15617 |
| `example-data`                 | same                         | #15618 |
| `experiments`                  | `react-experiments`          | #15566 |
| `fabric-website`               | `public-docsite`             | #15439 |
| `fabric-website-resources`     | `public-docsite-resources`   | #15439 |
| `file-type-icons`              | `file-type-icons`            | #15626 |
| `foundation`                   | `foundation-legacy`          | #15595 |
| `icons`                        | `font-icons-mdl2`            | #15603 |
| `@fluentui/react-icons`        | `react-icons-mdl2`           | #15599 |
| `jest-serializer-merge-styles` | same                         | #15615 |
| `mdl2-theme`                   | remove                       | done   |
| `merge-styles`                 | same                         | #15627 |
| `monaco-editor`                | same                         | #15617 |
[...]

PR #15248, opened 2020-09-25
Replace dep `@uifabric/fluent-theme` with `@fluentui/theme`

xugao:
#### Pull request checklist

#### Description of changes
For v8, we'd like to discontinue `@uifabric/fluent-theme` package in favor of `@fluentui/theme` package (related issue #13384). After `7.0` branch is created, we will remove `@uifabric/fluent-theme` package from `master`.

This PR is to update the the dependencies in places we can. Only package which still depend on `@uifabric/fluent-theme` after this PR is `@uifabric/experiments`, which still re-export `FluentStyles` and `FluentCustomizations`. These 2 are not and should not be added to new `@fluentui/theme` package since they are very much legacy code. After the branch cut, I am planning to remove exports of `FluentStyles` and `FluentCustomizations` from `experiments` package.

This PR also removes couple legacy examples from `@uifabric/experiments` which are no longer relevant.

#### Focus areas to test

(optional)


apps/fabric-website/package.json
```diff
@@ -41,13 +41,13 @@
   },
   "dependencies": {
     "@fluentui/examples": "^0.3.23",
+    "@fluentui/theme": "^1.2.0",
```
ecraig12345, line 44: Would you be okay with also updating the markdown files that refer to fluent-theme? (if we're ready to publicly encourage people to use the new theme package)

ecraig12345 (approved):

PR #15249, opened 2020-09-25
Remove sinon dependency and fix requestAnimationFrame mocks

ecraig12345:
#### Pull request checklist

#### Description of changes

Remove dependency on `sinon` since we rarely use it, and the usage can be replaced with jest mocks. This was partly motivated because sinon is one of the things with a postinstall script--though now that I look at it, turns out it just [asks for funding](https://github.com/sinonjs/sinon/blob/v4.5.0/scripts/support-sinon.js) rather than downloading stuff, but still doesn't hurt to remove.

The one potentially breaking part here is that under `@uifabric/utilities/lib/test` there's a helper `setRenderSpy` which relies on sinon. We don't use that anywhere in our repo, and it's not technically part of the public API since it's not exported from the package root.

While testing this, I ran into issues with `safeRequestAnimationFrame.test.tsx` which made me realize that the `requestAnimationFrame` mock all our packages were using was broken (didn't return the timeout ID and didn't have a corresponding `cancelAnimationFrame`). So I added proper mocks in `scripts/jest/jest-setup.js` and deleted the incorrect mocks from all packages.


packages/react-next/src/components/OverflowSet/OverflowSet.test.tsx
```diff
@@ -33,38 +32,26 @@ function getPersistedKeytip(keytipManager: KeytipManager, keySequences: string[]
   return ktp ? ktp.keytip : undefined;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const noOp = (): any => undefined;
```
xugao, line 36: so many `noOp` implemented everywhere :) perhaps move this to our test utils? 
Don't have to be in this PR

packages/utilities/src/test/setRenderSpy.ts
```diff
@@ -1,10 +1,17 @@
...
+// @ts-ignore
 import * as sinon from 'sinon';
 
 /**
  * Replaces the enzyme render function with a spy that can be used to count how many
  * times the render function was called in a test.
  * @param wrapper - The wrapper around the component instance for which you want to spy upon
+ * @deprecated This package no longer uses sinon. Please copy this function locally to continue using it.
  */
 export function setRenderSpy(wrapper: ReactWrapper<{}, {}>): sinon.SinonSpy {
```
xugao, line 16: I'd say let's just remove this. This was never officially supported from `utilities` package

xugao (approved):

khmakoto (approved):

packages/office-ui-fabric-react/src/components/OverflowSet/OverflowSet.test.tsx
```diff
@@ -32,11 +31,13 @@ function getPersistedKeytip(keytipManager: KeytipManager, keySequences: string[]
...
 }
 
+const noOp = (): any => undefined;
+
 describe('OverflowSet', () => {
   describe('snapshot tests', () => {
     test('basicSnapshot', () => {
-      const onRenderItem = sinon.spy();
-      const onRenderOverflowButton = sinon.spy();
+      const onRenderItem = jest.fn();
```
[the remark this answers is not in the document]
ecraig12345: It's a required prop. Though I did miss changing this instance to the general `noOp`.

packages/react-next/src/components/OverflowSet/OverflowSet.test.tsx
ecraig12345: I think it's a simple enough thing that I'd rather just declare it in each file as needed. Kind of like how you probably don't need the `leftpad` npm package. :P

packages/utilities/src/test/setRenderSpy.ts
ecraig12345: Sounds good

layershifter:
@ecraig12345 can be merge conflicts solved there? Do you need any help with it? ⛑️

xugao (approved):

PR #15255, opened 2020-09-25
Avatar animations

behowell:
#### Pull request checklist

#### Description of changes

Add animations for the Avatar's Active<->Inactive state transitions


packages/react-avatar/src/components/Avatar/Avatar.scss
```diff
@@ -267,7 +267,12 @@ $default-inactiveScaleFactor: 0.8;
 }
 
 // Activity modifiers
-._active {
+._active,
+._inactive {
+  // Work-around pixel snapping jankiness at the end of the animation if the scale factor is 1.0
+  transform: scale(0.9999);
```
behowell, line 273: Any opinions on this workaround? Without it the avatar jumps at the end of the scale animation.

dzearing: I'd have to pull to see what you are;

dzearing:
It would be good to add this to the demo site, or add to codesandbox templates. Otherwise can't see the changes you're making.

packages/react-avatar/src/components/Avatar/_tokens.scss
```diff
@@ -158,3 +158,41 @@ $shadow-16-darker: 0px 0px 4px $color-default-shadowAmbientDarker, 0px 6px 16px
...
+$anim-ultraFast: $anim-duration50;
+$anim-faster: $anim-duration100;
+$anim-fast: $anim-duration150;
+$anim-normal: $anim-duration200;
+$anim-slow: $anim-duration300;
+$anim-slower: $anim-duration400;
+$anim-ultraSlow: $anim-duration500;
+
+// Animation curves
+$anim-decelerateMax: cubic-bezier(0.00,0.00,0.00,1.00);
```
dzearing, line 180: Seems like we'd probably want to move this to @fluentui/theme.

dzearing (approved):

packages/react-avatar/src/components/Avatar/_tokens.scss
behowell: Yeah eventually all of these should come from design tokens via the theme. I didn't want that work to block me here, so I just added them here for now until the rest of that is figured out.

behowell:
> It would be good to add this to the demo site, or add to codesandbox templates. Otherwise can't see the changes you're making.

Are there instructions on how to do that?

PR #15269, opened 2020-09-28
Remove northstar-only packages from 7.0

ecraig12345:
Assuming we continue with publishing `@fluentui/react-northstar` and related packages from `master` only, it makes sense to delete them from the `7.0` branch. 

I also deleted the northstar-related scripts and dependencies--look at related items in the commit history for a readable summary of that.

We should definitely do this once the beta releases from `master` and daily beachball publishing is re-enabled, but it's probably okay to go ahead now (unless people would rather wait to be safe).


ecraig12345:
@chrisdholt It looks like web-components has an implicit dependency on `lodash-es` (it's a peer dep of `@microsoft/fast-components-styles-msft`)--it previously appeared to work because some other dependency happened to pull `lodash-es` in locally, but now that's gone and it's causing this PR's build to fail. I can add `lodash-es` as a dependency of the `scripts` package or something to work around the build failure, but seemed like something you should know about.

khmakoto (approved):

PR #15281, opened 2020-09-28
Implementing has-top-level-version-import test within react-conformance

czearing:
#### Pull request checklist
- [X] Fixes: #14888

#### Description of changes
1. Adding a `has-top-level-version-import` test to check if component's import a package version.
2. Adding a version file within `react-link` and importing it in the index file.
3. Adding a new type to `IsConformantOptions` called `packageVersion`.
4. Updating isConformant files to use `packageVersion`.
5. Removing redundant (once the class name test is added: https://github.com/microsoft/fluentui/pull/15218) ComponentConformance test files from `react-next` and `office-ui-fabric-react`.
6. Disabling `is-static-property-of-parent` test for `PeoplePicker`.
7. Renaming react-northstar's isConformant import to `baseIsConformant` instead of `isConformantBase` to keep consistency.
8. Disabling new `has-top-level-version` test in react-northstar.


layershifter:
@czearing is it still valid? Can you please solve merge conflicts in this case?

PR #15284, opened 2020-09-28
Fix expand/collapse for grouping in DetailsList

ThomasMichon:
#### Description of changes

Fixed an issue where expanding and collapsing group headers in `DetailsList` stopped working, even though a previous fix made it work for `GroupedList` directly.

#### Focus areas to test

Validated that the DetailsList Grouped example *and* the GroupedList example both support working group headers.


dzearing (approved):
Tested group expand/collapse, column header expand/collapse.

PR #15293, opened 2020-09-29
Pivot overflow: fix scrollbars appearing when the window is just barely wide enough to show all tabs

### LINKED ISSUE #15292: Pivot/Tabs overflow causes scrollbar to appear just before overflow menu appears
The Pivot/Tabs overflow menu causes scrollbar to appear just before overflow menu appears

Repro steps:
1. Open this example codesandbox of a pivot with overflow: https://codesandbox.io/s/pivot-overflow-menu-demo-grrgg (Standalone: https://grrgg.csb.app/)
2. Start with the window wide enough to show all tabs
3. Resize the window down until just *before* the overflow menu [...] appears.
4. Notice that the scrollbar appears at the bottom of the window

image

behowell:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15292

#### Description of changes

Pivot overflow: fix scrollbars appearing when the window is just barely wide enough to show all tabs

Set the overflow button to have `position: absolute; right: 0;` when it is hidden. This causes it to not take up layout space while hidden, but still have a size that can be measured by the `useOverflow` algorithm.

#### Focus areas to test

Pivot overflow

### DIFF
--- a/packages/react-tabs/src/components/Pivot/Pivot.base.tsx
+++ b/packages/react-tabs/src/components/Pivot/Pivot.base.tsx
@@ -262,7 +262,9 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef<
           {items}
           {overflowBehavior === 'menu' && (
             <CommandButton
-              className={classNames.overflowMenuButton}
+              // TODO when the span wrapper is removed from <CommandButton>, set
+              // className={classNames.link + ' ' + classNames.overflowMenuButton}
+              className={classNames.link}
               ref={setOverflowMenuButtonRef}
               componentRef={overflowMenuButtonComponentRef as React.RefObject<IButton>}
               menuProps={overflowMenuProps}
--- a/packages/react-tabs/src/components/Pivot/Pivot.base.tsx
+++ b/packages/react-tabs/src/components/Pivot/Pivot.base.tsx
@@ -244,8 +244,11 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef<
       (button: React.Component | null) => {
         const node = ReactDOM.findDOMNode(button);
         overflowMenuButtonRef(node instanceof HTMLElement ? node : null);
+        if (overflowMenuButtonRef.current) {
+          overflowMenuButtonRef.current.className = classNames.overflowMenuButton;
+        }
       },
-      [overflowMenuButtonRef],
+      [overflowMenuButtonRef, classNames.overflowMenuButton],
     );
 
     return (
--- a/packages/react-tabs/src/components/Pivot/Pivot.styles.ts
+++ b/packages/react-tabs/src/components/Pivot/Pivot.styles.ts
@@ -199,8 +199,9 @@ export const getStyles = (props: IPivotStyleProps): IPivotStyles => {
         [`&[data-is-overflowing='true']`]: {
           display: 'none',
           selectors: {
-            [`~ * .${classNames.overflowMenuButton}`]: {
+            [`~ .${classNames.overflowMenuButton}`]: {
               visibility: 'visible',
+              position: 'relative',
             },
           },
         },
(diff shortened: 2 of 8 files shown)


packages/react-tabs/src/components/Pivot/Pivot.styles.ts
```diff
@@ -199,18 +199,27 @@ export const getStyles = (props: IPivotStyleProps): IPivotStyles => {
...
+        // TODO after the extra span wrapper around the button is removed,
+        // remove the 'button&' and 'span&' selectors and combine the styles
+        ['button&']: [...getLinkStyles(props, classNames)],
+        ['span&']: [
+          {
+            visibility: 'hidden',
+            position: 'absolute',
+            right: 0,
+          },
+        ],
```
behowell, line 222: FYI @czearing, you will need to do this TODO when merging #15216. I would normally wait until after your PR is merged for this change (and not change it myself), but it looks like your PR is being held back for now. It should look like this after the TODO:
```js
overflowMenuButton: [
  classNames.overflowMenuButton,
  ...getLinkStyles(props, classNames),
  {
    visibility: 'hidden',
    position: 'absolute',
    right: 0,
  },
],
```

xugao:
> Thanks for submitting this change, but due to work currently in progress to prepare `master` for our version 8 beta release, we're asking contributors to either wait a couple weeks to submit changes (if it's not urgent) or submit to the new `7.0` branch (if it's urgent). See #15222 for more details. Thank you for your contribution and understanding!

@behowell - you can ignore this comment :) your change is v8 related so it should be in `master`

packages/react-tabs/src/components/Pivot/Pivot.styles.ts
```diff
@@ -199,18 +199,27 @@ export const getStyles = (props: IPivotStyleProps): IPivotStyles => {
...
       },
     ],
     overflowMenuButton: [
       classNames.overflowMenuButton,
-      ...getLinkStyles(props, classNames),
       {
-        visibility: 'hidden',
+        // TODO after the extra span wrapper around the button is removed,
+        // remove the 'button&' and 'span&' selectors and combine the styles
+        ['button&']: [...getLinkStyles(props, classNames)],
```
xugao, line 215:
```suggestion
        ['button&']: getLinkStyles(props, classNames),
```

packages/react-tabs/src/components/Pivot/Pivot.styles.ts
xugao, line 222:
```suggestion
        ['span&']: 
          {
            visibility: 'hidden',
            position: 'absolute',
            right: 0,
          },
```

xugao (approved):

packages/react-tabs/src/components/Pivot/Pivot.styles.ts
behowell: I had left it like this to make it clearer how to remove this workaround when Button is fixed, but it's probably not necessary :)

packages/react-tabs/src/components/Pivot/Pivot.base.tsx
```diff
@@ -259,7 +262,9 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef<
           {items}
           {overflowBehavior === 'menu' && (
             <CommandButton
-              className={classNames.overflowMenuButton}
+              // TODO when the span wrapper is removed from <CommandButton>, set
+              // className={classNames.link + ' ' + classNames.overflowMenuButton}
+              className={classNames.link}
```
behowell, line 267: It turns out that my original workaround didn't correctly apply hover visuals when when the links were displayed as tabs. This new workaround is a more straightforward fix, which applies different classes to the wrapper `<span>` and the inner `<button>`.

@czearing FYI with this change, you'll now need to add both classes (`link` and `overflowMenuButton`) to the button once  you remove the span.

PR #15303, opened 2020-09-29
fix(Breadcrumb): use a11y slot for breadcrumb items' container

YuanboXue-Amber:
#### Pull request checklist

#### Description of changes

Breadcrumb component wrap all its items in a `div`. Before this change, a 'role=list' is add directly on this div. After this change, the role is obtained from accessibility package.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-northstar/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/packages/fluentui/react-northstar/src/components/Breadcrumb/Breadcrumb.tsx
@@ -91,7 +91,13 @@ export const Breadcrumb = compose<'nav', BreadcrumbProps, BreadcrumbStylesProps,
         })}
       >
         <BreadcrumbContext.Provider value={contextValue}>
-          <div role="list">{childrenExist(children) ? children : content}</div>
+          <div
+            {...getA11yProps('itemsContainer', {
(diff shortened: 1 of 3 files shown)


packages/fluentui/react-northstar/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -91,7 +91,13 @@ export const Breadcrumb = compose<'nav', BreadcrumbProps, BreadcrumbStylesProps,
         })}
       >
         <BreadcrumbContext.Provider value={contextValue}>
-          <div role="list">{childrenExist(children) ? children : content}</div>
+          <div
+            {...getA11yProps('itemsContainer', {
```
layershifter, line 95: nit: let's use just `container`

layershifter (approved):

packages/fluentui/react-northstar/src/components/Breadcrumb/Breadcrumb.tsx
YuanboXue-Amber: LOL lightening review speed. Electricity still off huh? :D 
Changed as requested in the new commit.

pompomon (approved):

PR #15304, opened 2020-09-29
feat: add Text component to `react-text`

layershifter:
This PR adds an initial version of `Text` in `@fluentui/react-text`.


packages/react-theme-provider/src/useTheme.ts
```diff
@@ -17,6 +17,6 @@ function useCompatTheme(): ITheme | undefined {
 export const useTheme = (): Theme => {
   const theme = useContext(ThemeContext);
   const legacyTheme = useCompatTheme();
-
+  console.log(theme);
```
layershifter, line 20:
```suggestion

```

packages/react-text/src/components/Text/useText.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { makeMergeProps, getSlots, resolveShorthandProps } from '@fluentui/react-compose/lib/next/index';
+import { TextProps } from './Text.types';
+
+const mergeProps = makeMergeProps();
+
+/**
+ * Define the render function. Given the state of a text, renders it.
+ */
+export const renderText = (state: TextProps) => {
+  const { slots, slotProps } = getSlots(state, [] /* there is no slots in Text */);
```
khmakoto, line 11:
```suggestion
  const { slots, slotProps } = getSlots(state, [] /* there are no slots in Text */);
```

packages/react-text/src/components/Text/useText.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { makeMergeProps, getSlots, resolveShorthandProps } from '@fluentui/react-compose/lib/next/index';
+import { TextProps } from './Text.types';
+
+const mergeProps = makeMergeProps();
+
+/**
+ * Define the render function. Given the state of a text, renders it.
+ */
+export const renderText = (state: TextProps) => {
```
khmakoto, line 10: Should we move `renderText` into a separate file?

packages/react-text/src/components/Text/Text.types.ts
```diff
@@ -0,0 +1,16 @@
+import { BaseSlots, ComponentProps, SlotProps } from '@fluentui/react-compose';
+import { RecursivePartial } from '@fluentui/theme';
+import * as React from 'react';
+
+export interface TextProps extends ComponentProps, React.ImgHTMLAttributes<HTMLSpanElement> {
```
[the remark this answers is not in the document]
layershifter: Oops, copy-paste issue, good catch 👍

packages/react-text/src/components/Text/useText.tsx
layershifter: Moved 👍

khmakoto (approved):

khmakoto:
Seems like you need to run `yarn update-api` for the build to pass.

packages/react-text/src/components/Text/useTextClasses.ts
```diff
@@ -0,0 +1,66 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { makeClasses } from '@fluentui/react-theme-provider';
+
+const GlobalClassNames = {
+  root: 'ms-Text',
+};
+
+export const useTextClasses = makeClasses({
```
dzearing, line 8: this looks great, and also, I just merged `makeVariantClasses` should we want these to be easily definable and extendable in the theme. Can sync later on this.

dzearing (approved):
looks great!

PR #15306, opened 2020-09-29
fix(button): extra width when loading=false

### LINKED ISSUE #14060: FluentUIv0 - Button with loading=false has slightly more padding than loading=undefined
### Environment Information

- **Package version(s)**: `@fluentui/react-northstar@0.50.0`

https://codesandbox.io/s/fluentui-button-loading-false-issue-rtfb6?file=/example.js

#### Actual behavior:

The following usages of button should yield the same result.

```
<Button loading={false} content="Click here" />
<Button content="Click here" />
```

#### Expected behavior:

There are padding differences between the two implementations - 

image

Requested priority: Normal

Products/sites affected: (if applicable) NA

YuanboXue-Amber:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14060

#### Description of changes

Check if loading is false before applying `loadingMinWidth` to button

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-northstar/src/themes/teams/components/Button/buttonStyles.ts
+++ b/packages/fluentui/react-northstar/src/themes/teams/components/Button/buttonStyles.ts
@@ -1,4 +1,3 @@
-import * as _ from 'lodash';
 import { pxToRem } from '../../../../utils';
 import { ComponentSlotStylesPrepared, ICSSInJSStyle } from '@fluentui/styles';
 import { loaderSlotClassNames } from '../../../../components/Loader/Loader';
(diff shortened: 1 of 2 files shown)


packages/fluentui/react-northstar/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -23,7 +23,7 @@ export const buttonStyles: ComponentSlotStylesPrepared<ButtonStylesProps, Button
 
     return {
       height: v.height,
-      minWidth: _.isNil(p.loading) ? v.minWidth : v.loadingMinWidth,
+      minWidth: _.isNil(p.loading) || !p.loading ? v.minWidth : v.loadingMinWidth,
```
layershifter, line 26:
```suggestion
      minWidth: p.loading ? v.loadingMinWidth : v.minWidth,
```

I am not sure about usage of `_.isNil()`. Can you please check?

YuanboXue-Amber: It returns true when undefined or null. So your change actually makes more sense.

layershifter (approved):

PR #15313, opened 2020-09-30
export design system interface

chrisdholt:
#### Pull request checklist

#### Description of changes
Exports the DesignSystem interface to support better typing of Fluent Design System instances.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/web-components/src/design-system-provider/index.ts
+++ b/packages/web-components/src/design-system-provider/index.ts
@@ -571,3 +571,5 @@ export class FluentDesignSystemProvider extends DesignSystemProvider
   })
   public neutralOutlineFocusDelta: number;
 }
+
+export { DesignSystem };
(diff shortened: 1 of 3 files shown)


packages/web-components/src/design-system-provider/index.ts
```diff
@@ -571,3 +571,5 @@ export class FluentDesignSystemProvider extends DesignSystemProvider
   })
   public neutralOutlineFocusDelta: number;
 }
+
+export { DesignSystem };
```
nicholasrice, line 575: On line 47 you'll see this element actually omits a few properties from the DesignSystem interface. You should probably create an interface that is that omission, implement the new interface into the `FluentDesignSystemProvider` and export the new interface.

nicholasrice (approved):
Approved but left a comment

eljefe223 (approved):

PR #15331, opened 2020-10-01
chore: improve dist size for N*

layershifter:
### warnings about sourcemaps

image

The basic Webpack setup was screaming about missing sourcemaps, now there are inlined so there it's not an issue anymore.

### cleanup build

With @miroslavstastny we mentioned that artifacts published to NPM are too heavy.

- `dist` folder was not cleaned up properly (`/*` does not work at least on Windows) so I was publishing `processedIcons` folder that was removed a decade ago
- `dts` folder was also published however it's not required at all

The result is **31MB => 14MB** on published artifacts for `@fluentui/react-northstar`, this should speed up install of packages on CodeSandbox.

### DIFF
--- a/scripts/gulp/tasks/bundle.ts
+++ b/scripts/gulp/tasks/bundle.ts
@@ -14,16 +14,7 @@ const packageName = config.package;
 // Clean
 // ----------------------------------------
 
-task('bundle:package:clean', () =>
-  del(
-    [
-      `${paths.packageDist(packageName)}/es/*`,
-      `${paths.packageDist(packageName)}/commonjs/*`,
-      `${paths.packageDist(packageName)}/dts`,
-    ],
-    { force: true },
-  ),
-);
+task('bundle:package:clean', () => del([`${paths.packageDist(packageName)}`], { force: true }));
 
 // ----------------------------------------
 // Build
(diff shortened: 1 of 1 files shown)


scripts/gulp/tasks/bundle.ts
```diff
@@ -53,7 +44,14 @@ task('bundle:package:types:tsc', () => {
...
   return src(paths.packageDist(packageName, 'dts/src/**/*.d.ts')).pipe(dest(paths.packageDist(packageName, 'es')));
 });
-task('bundle:package:types', series('bundle:package:types:tsc', 'bundle:package:types:copy'));
+task('bundle:package:types:clean', () => {
+  return del([`${paths.packageDist(packageName)}/dts`], { force: true });
+});
+
+task(
+  'bundle:package:types',
+  series('bundle:package:types:tsc', 'bundle:package:types:copy', 'bundle:package:types:clean'),
```
layershifter, line 53: Types should exist only under `es` directory, this will remove temporary `dts` directory after build.

scripts/gulp/tasks/bundle.ts
```diff
@@ -35,15 +26,15 @@ task('bundle:package:commonjs', () =>
...
+    .pipe(sourcemaps.write())
     .pipe(dest(paths.packageDist(packageName, 'commonjs'))),
 );
 
 task('bundle:package:es', () =>
   src(componentsSrc)
     .pipe(sourcemaps.init())
     .pipe(babel({ caller: { useESModules: true } } as any))
-    .pipe(sourcemaps.write('.'))
+    .pipe(sourcemaps.write())
```
layershifter, line 37: Sourcemaps will be written inline, it saves from sourcemaps warnings.

miroslavstastny (approved):

PR #15336, opened 2020-10-01
Adding variant support for components

### LINKED ISSUE #14669: Variants PR reviewed and checked in

dzearing:

## Variants support

Variants are defined as such:

```jsx
  name: 'Foo',
  prefix: '--foo', 
  styles: {
    root: { ... },
    icons: { ... },
  },
  variants: {
    root: { 
      background: 'red'
    },
    primary: { 
      background: 'green'
    }
  }
});
```

```jsx
const Foo = (props) => {
  const state = { ...props };

  useFooClasses (state);

  return (
  );
```

```jsx
<Foo /> // red foo
<Foo primary /> // green foo
```

```jsx
<ThemeProvider theme={{
  components: {
     Foo: {
       variants: {
          primary: {
             background: 'purple'
          }
        }
      }
    }
  }}>
  <Foo/> // red button
</ThemeProvider>
```


dzearing:
Moved PR to #15337

PR #15337, opened 2020-10-01
Adding variants support

### LINKED ISSUE #14669: Variants PR reviewed and checked in

dzearing:

## Variants support

Variants are defined as such:

```jsx
  name: 'Foo',
  prefix: '--foo', 
  styles: {
    root: { ... },
    icons: { ... },
  },
  variants: {
    root: { 
      background: 'red'
    },
    primary: { 
      background: 'green'
    }
  }
});
```

```jsx
const Foo = (props) => {
  const state = { ...props };

  useFooClasses (state);

  return (
  );
```

```jsx
<Foo /> // red foo
<Foo primary /> // green foo
```

```jsx
<ThemeProvider theme={{
  components: {
     Foo: {
       variants: {
          primary: {
             background: 'purple'
          }
        }
      }
    }
  }}>
  <Foo/> // red button
</ThemeProvider>
```


packages/react-theme-provider/src/ThemeProvider.types.ts
```diff
@@ -40,4 +41,6 @@ export interface ThemeProviderProps extends ComponentProps, React.HTMLAttributes
  */
 export type ThemeProviderState = Omit<ThemeProviderProps, 'theme'> & {
   theme: Theme;
+
+  customizerProps: ICustomizerContext;
```
xugao, line 45:
```suggestion
  customizerContext: ICustomizerContext;
```

packages/react-theme-provider/src/makeVariantClasses.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  // This function will only be called when styles have not been evaluated for this set for
+  // the particular theme/window/direction combo.
+  const styleFunction: (theme: Theme) => IStyle = (theme: Theme) => {
+    const themeVariants = name ? theme?.components?.[name]?.variants : undefined;
+
+    return [
+      callOrReturn(styles, theme),
+      processVariants(variants!, theme, name!, prefix!),
+      processVariants(themeVariants, theme, name!, prefix!),
```
xugao, line 82: please remove/fix the usages of `!` here

packages/react-theme-provider/src/makeStyles.ts
```diff
@@ -54,8 +54,11 @@ export function makeStyles<TStyleSet extends { [key: string]: IStyle }>(
     const win = useWindow();
 
     // Expected: theme and renderer are either always provided or never.
```
xugao, line 56: can remove comment

packages/react-button/src/components/Button/Button.types.ts
```diff
@@ -87,61 +89,59 @@ export type ButtonProps = ComponentProps &
...
+      small?: string;
+      regular?: string;
+      large?: string;
+      larger?: string;
+      largest?: string;
+    };
+
+    pressed?: {
+      transform?: string;
+      transition?: string;
```
xugao, line 137: these fields look outdated

packages/react-button/src/components/Button/Button.test.tsx
```diff
@@ -1,18 +1,18 @@
...
+import { Button as BaseButton } from './Button';
 import * as renderer from 'react-test-renderer';
 import { mount, ReactWrapper } from 'enzyme';
 import { isConformant } from '../../common/isConformant';
+import { withThemeProvider } from '@fluentui/react-theme-provider';
 
-import { MergeStylesProvider } from '@fluentui/react-theme-provider';
-import { ButtonProps } from './Button.types';
+/** Use a ThemeProvider wrapper around the component to ensure styles show up in snapshots. */
+const Button = withThemeProvider(BaseButton);
```
xugao, line 9: i think we should create helper in test-utils to handle this in `renderer.create`

packages/react-button/src/components/Button/Button.types.ts
dzearing: hmmm. look right to me?

packages/react-theme-provider/src/makeVariantClasses.ts
dzearing: done!

packages/react-theme-provider/src/ThemeProvider.types.ts
dzearing: done!

packages/react-button/src/components/Button/Button.types.ts
xugao: if you compare this with useButtonClasses, tokens such as `whitespace` are missing

xugao (approved):

packages/react-theme-provider/src/makeStyles.ts
dzearing: removed

PR #15338, opened 2020-10-01
Implementing defaultErrorMessages file within isConformant to provide more specific error messages

czearing:
#### Pull request checklist

#### Description of changes

1. Adding a new `defaultErrorMessages` file to provide descriptive explanations for isConformant test errors.
2. Implementing `display-failed-tests` in isConformant to provide a list of failing isConformant tests in the component.
3. Taking note of a bug in `consistent-callback-names` test that causes it not to run.


packages/react-conformance/src/defaultErrorMessages.tsx
```diff
@@ -0,0 +1,682 @@
...
+    if (_.words(docblock.description).length === 0) {
+      console.log(
+        defaultErrorMessage(
+          `has-docblock`,
+          displayName,
+          'docblock in the file: ' + paragraph() + chalk.green.italic(componentPath),
+        ) +
+          resolveErrorMessages([
+            `Consider adding a descriptive` +
+              chalk.hex('#e00000')(' 5 ') +
```
ecraig12345, line 35: It might be best to stick with built-in colors, so that they adapt better to different terminal color schemes.

ecraig12345: On a related note, would be good to check the display of these messages in a light theme terminal if you're normally using dark (or vice versa).

packages/react-conformance/src/defaultErrorMessages.tsx
```diff
@@ -0,0 +1,682 @@
...
+}
+
+/** Generates a paragraph.
+ *  @param numberOfParagraphs The number of paragraphs to generate.
+ */
+const paragraph = (numberOfParagraphs?: number) => {
+  if (numberOfParagraphs) {
+    const paragraphs = [];
+    for (let i = -1; i < numberOfParagraphs; i++) {
+      paragraphs.push(`
```
ecraig12345, line 657: Just curious, why this approach instead of just creating a series of `os.EOL`? Also it's okay to use `String.prototype.repeat` here since this runs in Node (`repeat` is in ES6, which is supported by all node versions we support).

ecraig12345: In general for code that runs in Node, we care about supporting Node 8+ (soon to be 10+) which generally aligns with ES2017. That means [everything from ES6](https://kangax.github.io/compat-table/es6/) is allowed, and you can look at [this page for ES2016+](https://kangax.github.io/compat-table/es2016plus/#node8_10).

For code that runs in the browser it's easier to check supported methods--you can just search on https://caniuse.com/ and check the IE 11 column.

packages/react-conformance/src/defaultErrorMessages.tsx
czearing: I investigated this and it appears that most console messages aren't readable in other color schemes. 
image
image

I decided to update the resolve issue text so it will be more readable in light mode but not sure what to do about the other color schemes.
image

packages/react-conformance/src/defaultErrorMessages.tsx
czearing: Thanks for the suggestion! I updated the implementation of the paragraph function.

packages/react-conformance/src/defaultErrorMessages.tsx
ecraig12345: Thanks for checking on that! I think the main concern should be the standard light and dark themes.

ecraig12345: One remaining related question, why did you use #e00000 instead of just red? And if you do want to keep using the custom color, it might be good to extract that part to a helper (under whatever name makes sense). `const otherRed = chalk.hex('#e00000'); ... otherRed(' 5 ')`

packages/react-conformance/src/defaultErrorMessages.tsx
```diff
@@ -0,0 +1,685 @@
...
+        displayName,
+        'export in: ' + paragraph() + chalk.green.italic(componentPath),
+      ) +
+        resolveErrorMessages([
+          `Make sure that your component's ` +
+            chalk.hex('#e00000')(fileName) +
+            ' file contains an export for ' +
+            chalk.hex('#e00000')(displayName) +
+            '.',
+          'Check to see if you component uses export default and consider enabling' +
```
ecraig12345, line 105: Would be good to either use backticks or a color here.
```suggestion
          'Check to see if you component uses `export default` and consider enabling' +
```

packages/react-conformance/src/defaultErrorMessages.tsx
```diff
@@ -0,0 +1,685 @@
...
+    // 2. Make sure that your component's "displayName".base.tsx file contains a valid return statement.
+    // 3. Check to see if your component works as expected with mount and safeMount.
+    if (requiredProps) {
+      const formatRequiredProps = () => {
+        const results = [];
+        for (const libName of Object.keys(requiredProps)) {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          results.push(libName + `: ${(requiredProps as any)[libName] + ', '}`);
+        }
+        return results.join('\n');
```
ecraig12345, line 134: Some suggestions here:
- If you're creating an array and immediately filling it with a 1-1 mapping from another array, it's better to just use `map`
- In Node code you can use `Object.entries()` if you need both the key and value (it's ES6 so not allowed in the browser yet)
- Would be better to just use a template string to build the whole line
- It's probably better to add the commas when joining so you don't get an extra comma after the last prop
```suggestion
        return Object.entries(requiredProps)
          .map(([propName, propValue]) => `${propName}: ${propValue}`)
          .join(',\n');
```

packages/react-conformance/src/defaultErrorMessages.tsx
```diff
@@ -0,0 +1,685 @@
...
+    // 1. Make sure that your docblock meets the required word count of 5 to 25 words.
+    else {
+      console.log(
+        defaultErrorMessage(
+          `has-docblock`,
+          displayName,
+          'docblock between 5 and 25 words. It has a word count of: ' +
+            chalk.green.italic(docBlockLength) +
+            paragraph() +
+            'Here is ' +
```
ecraig12345, line 61: Not super important, but you might consider using template strings throughout the file to join some of these error messages together. (Template strings can also contain paragraph breaks.) For example:
```ts
        defaultErrorMessage(
          `has-docblock`,
          displayName,
          `docblock between 5 and 25 words. It has a word count of: ${chalk.green.italic(docBlockLength)}

Here is ${chalk.white(displayName)}'s docblock:

${chalk.green.italic(docblock.description)}`,
        )
```

packages/react-conformance/src/defaultErrorMessages.tsx
```diff
@@ -0,0 +1,685 @@
...
+    resolveMessage.join('') +
+    paragraph()
+  );
+}
+
+/** Generates the starting default error message: ( "It appears that __displayName__ doesn't have a __errorMessage__." )
+ *  @param testName The conformance test's name.
+ *  @param displayName The component's name.
+ *  @param errorMessage Why the test is failing.
+ */
```
ecraig12345, line 640: Another small thing, but this docblock formatting is a bit non-standard (though most tools will still handle it okay)
```suggestion
/**
 * Generates the starting default error message: ( "It appears that __displayName__ doesn't have a __errorMessage__." )
 * @param testName The conformance test's name.
 * @param displayName The component's name.
 * @param errorMessage Why the test is failing.
 */
```

packages/react-conformance/src/defaultErrorMessages.tsx
czearing: It is a tiny detail. The color is slightly darker and intended to help differentiate between isConformant error messages and others that use the default red chalk. Also, I decided to extract all of colors into one errorMessageColor object to help keep consistency in any future added error messages.

packages/react-conformance/src/defaultErrorMessages.tsx
czearing: I added template strings but would prefer to keep the current paragraph function implementation.

ecraig12345 (approved):
Thanks for all the work on this--should be really helpful for testing! I have some additional suggestions but we can follow up on that later.

ecraig12345:
See #15459 for follow-up feedback. Sorry for not looking at this more thoroughly and providing that feedback sooner.

PR #15271, opened 2020-09-28
Merge OUFR into react and update references

ecraig12345:
Move all OUFR files into react and update references to `office-ui-fabric-react` (to either `@fluentui/react`, just `react`, or occasionally something else).

Current status:
- `/apps`, `/scripts`, and root files are done. 
- `/packages` are in progress (feel free to help push updates)
- Haven't tested building at all

Related to #15152

PR #15341, opened 2020-10-02
Fix size auditor

ecraig12345:
Delete the accidental extra copy of ComponentConformance.test.tsx that was causing size auditor to fail after #15271.

And somewhat related changes:
- Update release notes to reflect changes from #15271 
- Clean up test-bundles package, including deleting stuff which appeared to only be used in a previous implementation of bundle size measurement. None of these things turned out to be the problem, but I made these changes while investigating.


ecraig12345:

packages/react-next/RELEASE_NOTES.md
```diff
@@ -117,6 +125,10 @@ Please see the [`@fluentui/react-theme-provider` package README](https://github.
 - `Button` and `Card` are new components that break from their previous implementation.
 - `WindowProvider` is required for child windows/embeds.
 - `FluentStyles` is removed from `experiments` package.
+- Removed various files which were originally in `office-ui-fabric-react` and not intended to be part of the public API:
+  - `office-ui-fabric-react/src/components/Theme/defaultTheme.ts`
+  - `office-ui-fabric-react/src/customizations/TeamsTheme.ts` (use `@fluentui/theme`)
```
xugao, line 130:
```suggestion
  - `office-ui-fabric-react/src/components/Theme/defaultTheme.ts` (use `@fluentui/theme`)
  - `office-ui-fabric-react/src/customizations/TeamsTheme.ts` (use `@fluentui/theme-samples`)
```

package.json
```diff
@@ -71,7 +71,7 @@
     "danger": "^6.0.5",
     "gulp": "^4.0.2",
     "lerna": "^3.15.0",
-    "lage": "^0.19.0",
+    "lage": "^0.19.8",
```
xugao, line 74: just curious, why are you bumping this?

apps/test-bundles/webpackUtils.js
```diff
@@ -85,7 +86,7 @@ function buildEntries(packageName, entries = {}, includeStats = true) {
 
   fs.readdirSync(packagePath).forEach(itemName => {
     const isJavascriptFile = itemName.match(/.js$/);
-    const isAllowedFile = TopLevelEntryFileExclusions.indexOf(itemName) === -1;
+    const isAllowedFile = TopLevelEntryFileExclusions.includes(itemName);
```
xugao, line 89:
```suggestion
    const isAllowedFile = !TopLevelEntryFileExclusions.includes(itemName);
```

apps/test-bundles/package.json
```diff
@@ -1,45 +1,25 @@
...
-  "module": "lib/index.js",
-  "sideEffects": [
-    "lib/version.js"
-  ],
-  "typings": "lib/index.d.ts",
+  "description": "Testing Fluent UI React bundle sizes.",
   "private": true,
   "license": "MIT",
   "scripts": {
-    "build": "just-scripts build",
```
xugao, line 14: we should keep `build`?

ecraig12345: I'm pretty sure the only things to build in this package were the TS files for the old bundle size thing, which I didn't think were being used at all. Though now I see size auditor is consistently failing with weird lightrail errors, so maybe I was wrong about that?

package.json
ecraig12345: I've been trying out `lage` locally and I wanted support for the `--to` flag (Ken said it was added more recently). Probably going to switch our local build commands to use it soon too.

apps/test-bundles/webpackUtils.js
ecraig12345: ...oops 🤦‍♀️

apps/test-bundles/package.json
ecraig12345: Hmm the lightrail job is failing with similar errors on one of your PRs, so maybe it's unrelated https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=135115&view=results

xugao (approved):

PR #15342, opened 2020-10-02
react-stylesheets: fix error when target is undefined (in case of SSR).

xugao:
#### Pull request checklist

#### Description of changes
When target is `undefined`, we are setting `undefined` as key in `WeakMap` which causes error.

#### Focus areas to test

(optional)


packages/react-stylesheets/src/StylesheetContext.tsx
```diff
@@ -1,10 +1,12 @@
 import * as React from 'react';
 export type HTMLDirection = 'rtl' | 'ltr' | 'auto';
 
+const defaultDocument = { document: 'document' };
```
dzearing, line 4: Hmm; providing a default mock document would break the renderStyles default method; would it be better to just ignore doing work when the `target` is unavailable?

dzearing (approved):

PR #15345, opened 2020-10-02
feat(UI builder): Add screen reader narration computation for descendant elements

adamsamec:
#### Description of changes

This PR adds the possibility to the UI builder to compute screen reader narration also for all the focusable descendant elements of the selected component, and also includes   some refactoring of the code.

IN the UI builder, after selecting a component added to the canvas, the user can select a narration for the desired descendant element using a dropdown.


scripts/config.ts
```diff
@@ -88,7 +88,7 @@ const config = {
   // ----------------------------------
   // Server Configuration
   // ----------------------------------
-  server_host: 'localhost',
+  server_host: '0.0.0.0',
```
miroslavstastny, line 91: Please revert this change before merging

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
```diff
@@ -0,0 +1,82 @@
...
+};
+
+export const ReaderNarration: React.FunctionComponent<ReaderNarrationProps> = ({ selector }) => {
+  const ref = React.useRef<HTMLElement>();
+  const [narrationText, setNarrationText] = React.useState('');
+  const [narrationPath, setNarrationPath] = React.useState('');
+  const [narrationPaths, setNarrationPaths] = React.useState([]);
+
+  React.useEffect(() => {
+    if (ref.current) {
```
ling1726, line 20: Might be personal preference, but when there is an if block that spans all your logic I think it's cleaner to invert and return early

ling1726:
```suggestion
    if (!ref.current) {
      return;
```

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
```diff
@@ -0,0 +1,82 @@
...
+      }); // End compute
+    } // End if 1
+  }, [setNarrationText, setNarrationPath, setNarrationPaths, ref, selector]);
+
+  // Composes and sets the complete screen reader narration text to be displayed
+  const setCompleteText = text => {
+    setNarrationText(`Narration: ${text}`);
+  }; // End setCompleteText
+
+  const handleNarrationPathChange = (event: any, props: DropdownProps) => {
```
ling1726, line 50: You don't seem to use the event, will setting it to an anonymous variable work ? 

```suggestion
  const handleNarrationPathChange = (_: any, props: DropdownProps) => {
```

packages/fluentui/react-builder/src/narration/DescendantsNarrationsComputer.tsx
```diff
@@ -0,0 +1,80 @@
...
+  findActiveDescendantsParents(element: IAriaElement, parents: IAriaElement[]) {
+    if (element.ariaActiveDescendantElement != null) {
+      // Begin if 1
+      parents.push(element);
+    } // End if 1
+    for (let i = 0; i < element.children.length; i++) {
+      // Begin for 1
+      const child = element.children[i] as IAriaElement;
+      this.findActiveDescendantsParents(child, parents);
+    } // End for 1
```
ling1726, line 35: nit: might be cleaner ? 

```suggestion
    element.children.forEach((child: IAriaElement) => this.findActiveDescendantsParents(child, parents))
```

packages/fluentui/react-builder/src/narration/DescendantsNarrationsComputer.tsx
```diff
@@ -0,0 +1,80 @@
...
+      newPath.push(item);
+    }
+
+    // If the element is focusable, stop the traversal and compute and save the narration
+    const isDirectlyFocusable =
+      (element.getAttribute('tabindex') || element.tabIndex >= 0) && element.ariaActiveDescendantElement == null;
+    const isActiveDescendant = activeDescendantsParents.some(parent => {
+      // Begin some 1
+      return parent.tabIndex >= 0 && parent.ariaActiveDescendantElement === element;
+    }); // End some 1
```
ling1726, line 62: Why is this check necessary ? if you are traversing the tree wouldn't you find the directly focusable parent first ?

scripts/config.ts
adamsamec: Done.

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
adamsamec: Agree. Thanks.

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
adamsamec: What are the advantages of using `_` instead? If it is just a matter of code style, I would prefer a variable with a meaningful name.

packages/fluentui/react-builder/src/narration/DescendantsNarrationsComputer.tsx
adamsamec: That won't work because `HTMLCollection` is not an Array object but an array-like object which doesn't have the `forEach()` method, so I get "TS2339: Property 'forEach' does not exist on type 'HTMLCollection'." Neither it can be iterated using `for/of` loop, in which case I get "TS2495: Type 'HTMLCollection' is not an array type or a string type." See more explanation [here](https://stackoverflow.com/questions/53098742/typescript-for-loop-over-htmlcollection). However, I  can use `Array.from()` to transform the `HTMLCollection` into the Array object, and then use `forEach()`.

packages/fluentui/react-builder/src/narration/DescendantsNarrationsComputer.tsx
adamsamec: Sorry, I didn't get it. Can you explain on it more?

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
ling1726: That's fair, it's a personal preference for me to indicate that this variable is not being used. Feel free to ignore or ask team members who have been here longer

packages/fluentui/react-builder/src/narration/DescendantsNarrationsComputer.tsx
ling1726: Your decision. Some people might scream that `Array.from` could iterate the array a second time, but if performance is not an issue, I consider my suggestion to be more easily readable (this is subjective though)

packages/fluentui/react-builder/src/narration/DescendantsNarrationsComputer.tsx
ling1726: Ah sorry, I might not have understood the logic completely so this comment might be unnecessary.

If you were traversing the tree and found an element that was focusable you would stop traversing. Therefore do you need to check if an element has focusable parents ?

ling1726: As discussed offline, I didn't quite understand that you have to keep traversing the if there is an active descendant

packages/fluentui/react-builder/src/narration/DescendantsNarrationsComputer.tsx
adamsamec: Performance is not an issue here, so I will leave it using `forEach()`. If nothing, I at least learned something new :-)

jurokapsiar (approved):

PR #15348, opened 2020-10-02
Expose filetypeIcon src through a utility

arkogupta:
#### Pull request checklist

#### Description of changes

This change is mainly in regard to starting a discussion on the best practice to consume the `src` of FileTypeIcons. [This ](https://onedrive.visualstudio.com/OneDriveWeb/_git/odsp-common?path=%2Futilities%2Fcustom-formatter%2Fsrc%2FCustomFormatter.ts&version=GBarkgupta%2FthumbnailPoc2&line=789&lineEnd=797&lineStartColumn=1&lineEndColumn=6&lineStyle=plain)is a change in a [PR](https://onedrive.visualstudio.com/OneDriveWeb/_git/odsp-common/pullrequest/531880) where I am currently consuming it  using getIcon(iconName).code.props.src. The idea is to compose raw `` tags and set the innerHtml rather than directly using a React Component. Can accessing the src from the React element obtained from getIcon like this cause any issues in the longer run ?

The idea which I wanted to discuss was what if we expose the src through a utility in the way I have done in this PR?

#### Focus areas to test

(optional)

### DIFF
--- a/packages/file-type-icons/src/initializeFileTypeIcons.tsx
+++ b/packages/file-type-icons/src/initializeFileTypeIcons.tsx
@@ -8,10 +8,18 @@ const SVG_SUFFIX = '_svg';
 const DEFAULT_BASE_URL = 'https://spoprod-a.akamaihd.net/files/fabric-cdn-prod_20200921.001/assets/item-types/';
 const ICON_SIZES: number[] = [16, 20, 24, 32, 40, 48, 64, 96];
 
+const _iconSources: { [iconName: string]: string } = {};
+
 export function initializeFileTypeIcons(baseUrl: string = DEFAULT_BASE_URL, options?: Partial<IIconOptions>): void {
   ICON_SIZES.forEach((size: number) => {
     _initializeIcons(baseUrl, size, options);
   });
+}
+
+export function getIconSource(iconName: string): string | undefined {
+  if (_iconSources.hasOwnProperty(iconName)) {
+    return _iconSources[iconName];
+  }
 }
 
 function _initializeIcons(baseUrl: string, size: number, options?: Partial<IIconOptions>): void {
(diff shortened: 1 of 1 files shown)


arkogupta:
Hello! @dzearing @Jahnp @bigbadcapers @miroslavstastny 
Could you please have a look at this ? Just wanted to understand whether consuming `src`  by passing around the JSX element and reading its props is OK, or if we should expose a utility like this.

bigbadcapers:
> Hello! @dzearing @Jahnp @bigbadcapers @miroslavstastny
> Could you please have a look at this ? Just wanted to understand whether consuming `src` by passing around the JSX element and reading its props is OK, or if we should expose a utility like this.

Hi Arko, I sent you a mail with questions, thanks!

dzearing:
The concern in exposing the internals here would be if we wanted to move to another medium, such as SVG or an icon glyph or image map to render the file type icon, this wouldn't work. Rendering a component lets us abstract how the image is rendered completely, and the dom elements which need to come along.

bigbadcapers:
@dzearing I'm not totally sure if I am following the logic of the implementation correctly, but I've seen most usages of this package using [getFileTypeIconProps](https://github.com/microsoft/fluentui/blob/c5a0c759b616ff36e34dc001d9ef6ffb4a17608f/packages/file-type-icons/src/getFileTypeIconProps.ts#L56) directly ([example](https://github.com/microsoft/fluentui/blob/e90e386962218b3b0f1327bd6f9677764367abf1/packages/react-examples/src/experiments/FileTypeIcon/FileTypeIcon.Basic.Example.tsx#L16)).

I had always thought that this method basically just returns a URL to the image, but is it actually returning the whole img tag?

Ridvanovskyy (changes requested):

arkogupta:
Closing this PR. Trying to achieve the same in this PR: #15405

PR #15351, opened 2020-10-02
Remove unneeded Fabric component usages

xugao:
#### Pull request checklist

#### Description of changes
Remove `Fabric` and replace with `div` where it's not needed (in examples and vr-tests only). 

#### Focus areas to test

(optional)

### DIFF
--- a/apps/vr-tests/src/stories/PivotNext.stories.tsx
+++ b/apps/vr-tests/src/stories/PivotNext.stories.tsx
@@ -11,7 +11,7 @@ import {
   IPivotItemProps,
   Label,
   Icon,
-  Fabric,
+  ThemeProvider,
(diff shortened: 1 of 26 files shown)


apps/vr-tests/src/stories/PivotNext.stories.tsx
```diff
@@ -176,7 +176,7 @@ storiesOf('Pivot Next Overflow', module)
     </Pivot>
   ))
   .addStory('Tabs - RTL', () => (
-    <Fabric dir="rtl">
+    <ThemeProvider dir="rtl" theme={{ rtl: true }}>
```
xugao, line 179: can remove `dir="rtl"` once this is merged: https://github.com/microsoft/fluentui/pull/15347

PR #15352, opened 2020-10-02
Avatar visual tweaks and clean up

behowell:
#### Pull request checklist

#### Description of changes

* Design updates and clean up for Avatar styles.
* Update Avatar active animation example.


khmakoto (approved):
This looks good!

One follow up that I think would be good to do in a separate PR would be to add the `react-avatar` package to the list of deployed sites in the PR so we could play with things like the updated example on this one.

PR #15354, opened 2020-10-03
Add theme and react-theme-provider exports and remove foundation exports from @fluentui/react

xugao:
#### Pull request checklist

#### Description of changes
- Export * from `@fluentui/react` and `fluentui/react-theme-provider` from `react` (and `react-internal`)
- Remove `@uifabric/foundation` export from `react` (and `react-internal`)

#### Focus areas to test

(optional)

### DIFF
--- a/packages/experiments/src/components/Button/Button.view.tsx
+++ b/packages/experiments/src/components/Button/Button.view.tsx
@@ -1,7 +1,7 @@
 /** @jsx withSlots */
 import * as React from 'react';
 import { Text, KeytipData } from '@fluentui/react';
-import { withSlots } from '../../Foundation';
+import { withSlots } from '@uifabric/foundation';
 import { getNativeProps, anchorProperties, buttonProperties } from '../../Utilities';
 import { FontIcon } from '../../utilities/factoryComponents';
 
--- a/packages/react-internal/src/Styling.ts
+++ b/packages/react-internal/src/Styling.ts
@@ -1,4 +1,2 @@
 import './version';
 export * from '@uifabric/styling';
-export * from '@fluentui/theme';
-export * from '@fluentui/react-theme-provider';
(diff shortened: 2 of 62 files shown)


packages/react-internal/src/Styling.ts
```diff
@@ -1,2 +1,4 @@
 import './version';
 export * from '@uifabric/styling';
+export * from '@fluentui/theme';
```
ecraig12345, line 3: Do you think it will be obvious enough to people that this is the top-level file where these things should be imported from? Or should they get separate top-level files (probably `Theme` and `ThemeProvider`)?

packages/experiments/src/components/Button/Button.view.tsx
```diff
@@ -1,7 +1,7 @@
 /** @jsx withSlots */
 import * as React from 'react';
 import { Text, KeytipData } from '@fluentui/react';
-import { withSlots } from '../../Foundation';
```
dzearing, line 4: can we just remove things which use foundation from experiments?

packages/react-internal/src/Styling.ts
dzearing: Perhaps we should keep the Styling export separate from a Theme.tsx export, so that we can deprecate Styling?

packages/experiments/src/components/Button/Button.view.tsx
xugao: I don't think so; lots of components in `experiments` are still using foundation:
image

packages/react-internal/src/Styling.ts
xugao: moved the new exports to `Theme.ts` (`lib/Theme`)

packages/react/src/Theme.ts
```diff
@@ -0,0 +1 @@
+export * from '@fluentui/react-internal/lib/Theme';
```
khmakoto, line 1: Should we add an export from the index file for this?

xugao: no need bc index of `react` does: `export * from '@fluentui/react-internal';`

khmakoto (approved):

ecraig12345 (approved):

Issue #15152, opened 2020-09-21
Merge OUFR and react-next into react


ecraig12345:
This is part 1/4 in a series of changes that will happen in master *after* v7 development moves to a branch, but *before* beta publishing:
1. (this) #15152 OUFR/react-next merge 
2. #14679 Move date-time components
3. #13384 uifabric/fluentui package renames
4. #14689 set up beta releases

This will be done across multiple PRs, and some of the work can be split up between multiple people.

- [x] Merge OUFR into react (1 person)
- [ ] Merge react-next into react (last chance to review FC changes). This can be split into groups of components to be done by different people.
- [x] Delete OUFR and update code references to react-next
- [ ] Either delete react-next or make it into a pure re-export package from react (and bump version to 9-alpha)
- [ ] Update gitignore so that removed files don't come back in a bad merge
- [ ] Update code references to react-next
- [ ] Delete "next" versions of vr/perf tests

ecraig12345:
The "merge react-next into react" step ended up being WAY more complicated than expected. Here's the procedure and tracking we used...

----

If the v7 component implementation is in `react-internal` (most components), the `react-next` implementation should be moved to `react-internal` (NOT `react`). In the steps, `react(-internal)` means whichever of the two packages is appropriate for your component.

Full steps:

1. Move the component code and tests from react-next to `react(-internal)`. Steps:
    1. Find where the v7 component is currently located (either `react-internal` or `react`) and delete the whole folder `react(-internal)/src/components/ComponentName`.
    2. Move the whole folder `react-next/src/components/ComponentName` to `react(-internal)`
    3. Update any `react-next` or `react` imports in the moved code to relative imports. If it's in `react-internal` also change those to relative imports. (A few things must still be deep-imported from react-internal; we'll fix this later.)
2. Update the component's root file in react-next (e.g. `react-next/src/Fabric.ts`) to re-export from `@fluentui/react/lib/ComponentName`
3. Delete any `*Next` perf-tests or vr-tests for the component
4. Delete `react-examples/src/react-next/ComponentName/docs` (they should be exact duplicates; even if they're not, the react versions are more likely to be correct)
5. Compare the examples and .doc.tsx in `react-examples/src/react-next/ComponentName` with the `react` versions. If the example names and content appear unchanged you can just delete the react-next versions. Otherwise copy over the moved examples to the react folder (fixing imports to be from react), and delete any remaining react-next example/doc files.
[...]

PR #15367, opened 2020-10-05
Image: move Image component to react and delete from react-next

tomi-msft:
#### Pull request checklist

#### Description of changes

This PR moves the version of `Image` existing in `@fluentui/react-next` to `@fluentui/react-internal`. As part of this API and snapshots were updated, the `@fluentui/react-next` examples of the component were removed, and the versions of these components in `@fluentui/react-next` now re-export the versions in `@fluentui/react-internal`.

(The Image changes in this PR are originally from #12908 where Image was converted to a function component.)

#### Focus areas to test

(optional)


khmakoto (dismissed):
From the looks of it you should be moving `Image` to `react-internal`, not to `react`.

packages/react-next/src/Image.ts
```diff
@@ -1,2 +1 @@
-import './version';
-export * from './components/Image/index';
+export * from '@fluentui/react-internal/lib/Image';
```
ecraig12345, line 1: Generally we want these to re-export from react (which re-exports react-internal)
```suggestion
export * from '@fluentui/react/lib/Image';
```

ecraig12345 (approved):
Mostly looks good, just one comment. Thanks!

ecraig12345:
This is nice
image

change/@fluentui-react-next-2020-10-05-12-10-48-image-migration-react.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "move Image component to react and delete from react-next",
```
khmakoto, line 3:
```suggestion
  "comment": "move Image component from react-next to react-internal",
```

change/@fluentui-react-2020-10-05-12-10-48-image-migration-react.json
```diff
@@ -0,0 +1,8 @@
+{
```
khmakoto, line 1: You should remove this file since you're no longer changing anything in the `react` package.

khmakoto (approved):
Approved with comments.

PR #15368, opened 2020-10-05
Moving FocusTrapZone and Modal from react-next to react-internal

### LINKED ISSUE #15266: Port aria-hidden siblings change to next
#13281 is missing from react-next across multiple components. The prop is present but unused in Modal, and entirely missing from FocusTrapZone.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Part of #15152, fixes #15266

#### Description of changes

This PR moves the versions of `Modal` and `FocusTrapZone` existing in `@fluentui/react-next` to `@fluentui/react-internal`. As part of this API and snapshots were updated, the `@fluentui/react-next` examples, vr tests and perf tests of these components were removed, and the versions of these components in `@fluentui/react-next` now re-export the versions in `@fluentui/react-internal`.

Another change in this PR is that `resetIds` is now being called in those tests that were affected by this change.

#### Focus areas to test

(optional)


packages/react-internal/src/components/Dialog/__snapshots__/Dialog.deprecated.test.tsx.snap
```diff
@@ -371,8 +369,8 @@ exports[`Dialog deprecated props renders Dialog with containerClassName 1`] = `
     onTouchStart={[Function]}
   >
     <div
-      aria-describedby="Dialog2-subText"
-      aria-labelledby="Dialog2-title"
+      aria-describedby="Dialog3-subText"
```
ecraig12345, line 372: Looks like Dialog isn't calling resetIds--if you don't mind, would be good to fix that (in both react-internal and react-next so it doesn't get lost). Should probably be done in beforeEach and afterAll.

ecraig12345:
Can you delete the FocusTrapZoneNext perf-tests and vr-tests?

packages/react-internal/src/components/Modal/Modal.base.tsx
```diff
@@ -464,25 +413,25 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
         }
         case KeyCodes.up: {
           this.setState({
-            y: this._getClampedPositionY(this.state.y - delta),
+            y: this.state.y - delta,
```
ecraig12345, line 416: The clamping is a new change which hasn't been ported. We should definitely not remove that, and if it involved any new props to FocusTrapZone we should add those too.

packages/react-internal/src/components/Modal/Modal.base.tsx
```diff
@@ -470,19 +467,19 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
           this.setState({
-            x: this._getClampedPositionX(this.state.x - delta),
+            x: this._getClampedPositionY(this.state.x - delta),
           });
           break;
         }
         case KeyCodes.right: {
           this.setState({
-            x: this._getClampedPositionX(this.state.x + delta),
+            x: this._getClampedPositionY(this.state.x - delta),
```
ecraig12345, line 482: Several typos here

packages/react-internal/src/components/Modal/Modal.base.tsx
```diff
@@ -152,9 +154,6 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
         isVisible: true,
       });
     }
-    if (!prevProps.isOpen && this.props.isOpen) {
```
ecraig12345, line 155: I'm pretty sure that all the changes to this file should be reverted

packages/react-internal/src/components/FocusTrapZone/FocusTrapZone.types.ts
```diff
@@ -78,11 +78,4 @@ export interface IFocusTrapZoneProps extends React.HTMLAttributes<HTMLDivElement
    * @defaultvalue false
    */
   focusPreviouslyFocusedInnerElement?: boolean;
-
-  /**
-   * Puts aria-hidden=true on all non-ancestors of the current element, for screen readers.
```
ecraig12345, line 83: This is new and needs to come back.

ecraig12345 (commented):
Thanks for doing this, but it looks like we're

packages/react-internal/src/components/Dialog/__snapshots__/Dialog.deprecated.test.tsx.snap
```diff
@@ -91,7 +91,6 @@ exports[`Dialog deprecated props renders Dialog with className 1`] = `
               transition: opacity 0.267s;
               width: 100%;
             }
-        role="document"
```
ecraig12345, line 94: Might be missing a snapshot update? I wouldn't expect this change now that the Modal changes have been reverted.

khmakoto: Updated!

tomi-msft (approved):

packages/react-internal/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -309,9 +313,15 @@ export const FocusTrapZone: React.FunctionComponent<IFocusTrapZoneProps> & {
       if (!disableFirstFocus && !elementContains(root.current, internalState.previouslyFocusedElementOutsideTrapZone)) {
         focus();
       }
+      if (!unmodalize && root.current && enableAriaHiddenSiblings) {
+        unmodalize = modalize(root.current);
+      }
     } else if (!newForceFocusInsideTrap || newDisabled) {
       // Transition from forceFocusInsideTrap / FTZ enabled to disabled.
       returnFocusToInitiator();
+      if (unmodalize) {
```
ecraig12345, line 322: Thanks for adding this part! However I think it needs to be handled a bit differently to replicate the class component behavior: the `unmodalize` callback should be stored in `internalState`, and in addition to this part you'll need to add another useEffect for componentWillUnmount.

I started trying to make this change last night and then noticed potentially a bunch of other issues with lifecycles that may not have been ported properly...so we definitely want to test this one carefully.

khmakoto: I think I did this properly but let me know which other issues you've found.

ecraig12345 (approved):

xugao (approved):

PR #15374, opened 2020-10-06
Cherrypicking variants into 7.0 branch

dzearing:


dzearing:
i did something wrong.......

PR #15379, opened 2020-10-06
chore(Popup): popupBehavior - Add tests and documentation with new approach

kolaps33:
#### Description of changes
This PR contains
1. 
general fix, when doc is generated, fix in 'rules.tsx'
2. 
Added tests for popup behavior and from the tests is new description generated, please see screenshot below the paragraph.
**So far tests doesn't test key Actions. It will be adding in next steps/tickets.** 
3. fix for existing 'inline' unit test, because after adding popup tests, this old one started to fail
4. changed how we render component in "ComponentTestFacade" (comment included in the code)
5. changed how we generate documentation for "props", this affected as well ButtonBehavior documentation... (screenshot included). The purpose is, not write `description` when given props make sense to be generated.
6. popup fixes > suggested and done by @layershifter. Thank you him for this help.

image
#### Focus areas to test

image

(optional)


packages/a11y-testing/src/facades/ComponentTestFacade.tsx
```diff
@@ -1,7 +1,28 @@
 import * as React from 'react';
-import * as ReactDOM from 'react-dom';
 import { Props, PropValue, TestFacade } from '../types';
+import { emptyTheme, ThemePrepared } from '@fluentui/styles';
+import { Renderer, noopRenderer } from '@fluentui/react-northstar-styles-renderer';
 import { mount, ReactWrapper } from 'enzyme';
+import { Telemetry, Unstable_FluentContextProvider, ProviderContextPrepared } from '@fluentui/react-bindings';
+
+const EmptyThemeProvider: React.FunctionComponent<{
```
kolaps33, line 8: This is copied from 
fluentui\packages\fluentui\react-northstar\test\utils\withProvider

Not sure if we should not extract "withProvider" to package and consume it then here and in the other unit tests as well? 

Or is it ok to have it here in this way?

pompomon: I'd vote to export and reuse here :)

packages/a11y-testing/src/definitions/Popup/popupBehaviorDefinition.ts
```diff
@@ -0,0 +1,70 @@
...
+  BehaviorRule.slot('popup')
+    .forProps({ trapFocus: false })
+    .doesNotHaveAttribute('aria-modal')
+    .description(`if 'trapFocus' is 'false'.`),
+];
+
+// remove 'definition' which contains 'dialog' rule because description is already cover in popupBehaviorDefinitionTriggerSlotNotTabbable
+// just want to keep 'definition' in the array, in order to test 'tabbable' and 'notTabbable' trigger
+const modifiedForDocumentation = popupBehaviorDefinitionTriggerSlotTabbable.filter(
+  definition => definition.stringify().search('dialog') === -1,
```
jurokapsiar, line 64: would it be possible to add a `hide()` method to the rule definition and skip it from documentation?

packages/a11y-testing/src/facades/ComponentTestFacade.tsx
jurokapsiar: let's move ComponentTestFacade to react-bindings as it now contains a lot of v0 specific code.

layershifter: Is it required? Our components should not throw when are used without `Provider`

packages/a11y-testing/src/definitions/Popup/popupBehaviorDefinition.ts
```diff
@@ -0,0 +1,70 @@
...
+    .forProps({ tabbableTrigger: false })
+    .doesNotHaveAttribute('tabIndex')
+    .description(`if 'tabbableTrigger' is 'false'.`),
+  BehaviorRule.slot('trigger')
+    .forProps({ tabbableTrigger: false })
+    .doesNotHaveAttribute('aria-haspopup')
+    .description(`if 'tabbableTrigger' is 'false'.`),
+];
+
+export const popupBehaviorDefinitionTriggerSlotTabbable: Rule[] = [
```
layershifter, line 24: I expect that we will have a single definition for the matching behavior. Why we have 5 exports there? 🤔

packages/a11y-testing/src/facades/ComponentTestFacade.tsx
kolaps33: For example Popup.tsx had issue in following function (figured out while debugging):
`updateTriggerFocusableRef`, or any other function, where `context.target`  is required.

Before I had `context.target`  undefined. 

When I user "withProvider" then there is wrapping component `target: document`, what allows popup works. 

@layershifter

kolaps33: > let's move ComponentTestFacade to react-bindings as it now contains a lot of v0 specific code.

moved :) into: packages\fluentui\react-northstar\test\utils

packages/a11y-testing/src/definitions/Popup/popupBehaviorDefinition.ts
kolaps33: it is already there, just I didn't used it properly, now using hide() approach

packages/a11y-testing/src/definitions/Popup/popupBehaviorDefinition.ts
kolaps33: This appears like we have two slots which we need to test like `trigger` and `popup`. 
And then `trigger` is special in the way, that behaves differently when we receive: 
- Tabbable element  
- not tabbable element 
- element with tabIndex already set... 

So in the end I ended with 4 definitions which using in the tests. 

For purpose of doc generation I still exporting only one "popupBehaviorDefinition".

packages/a11y-testing/src/facades/ComponentTestFacade.tsx
layershifter: Our component should render without a `Provider`. I suggest to fix a usage in `Popup` itself:

```diff
  const updateTriggerFocusableRef = () => {
-    const activeDocument: HTMLDocument = context.target;
-    const activeElement = activeDocument.activeElement;
+    const activeElement = context.target?.activeElement;

+ if (activeElement) {
    triggerFocusableRef.current =
      triggerRef.current && elementContains(triggerRef.current, activeElement as HTMLElement)
        ? (activeElement as HTMLElement)
        : triggerRef.current;
+  }
  };
```

- typecast to `HTMLDocument` can be removed as our context now typed properly
- optional chaining in `context.target?.activeElement` is required as definitely `context.target` can be undefined:

https://github.com/microsoft/fluentui/blob/8827b3790eba246d42f34a9b16793afb9cd519ef/packages/fluentui/react-bindings/src/context.ts#L33-L34

kolaps33: After this fix and next fixes pushed by @layershifter the popup works without wrapper/provider. 
Therefore returning back the 'ComponentTestFacade.tsx' into the a11y-testing package.

jurokapsiar (approved):

pompomon (approved):

pompomon (approved):

miroslavstastny (approved):

PR #15381, opened 2020-10-06
chore: rework mergeComponentStyles to avoid additional merging

layershifter:
### Description of changes

This PR updates `mergeComponentStyles()` to accept two params (mainly to reduce the complexity) and avoids the creation additional functions if it's not necessary.

It's done to improve merging as before any merge will multiple the work that should be done in styles functions:

```tsx
const customTheme = mergeThemes(teamsTheme, teamsTheme)
customTheme.componentStyles.Button.root(param); // .root will be executed twice
```

This can be confirmed with measurements:

| case | ops/sec |
| --- | ---: |
| `teamsTheme.componentStyles.Button.root(param); ` | 977153 ops/sec |
| `customTheme.componentStyles.Button.root(param);` (before) | 88273 ops/sec |
| `customTheme.componentStyles.Button.root(param);` (after) | 1067142 ops/sec |

Results in the second case are worst also because we will perform deep merge via `_.merge()`:
https://github.com/microsoft/fluentui/blob/8e1d372b8d5e169d8c3deffb0710402638fff723/packages/fluentui/styles/src/mergeThemes.ts#L81-L84

---

In the real world no one will merge the theme with itself but it can be easily done in `Provider`'s case:

```tsx
<Provider theme={teamsTheme}>
  {/* a lot of nested components */}
  <Provider theme={teamsTheme} rtl />
</Provider>
```


packages/fluentui/styles/src/mergeThemes.ts
```diff
@@ -45,105 +45,124 @@ export const emptyTheme: ThemePrepared = {
...
-    return mergeComponentStyles__PROD(...sources);
+      // if there is no source, merging is a no-op, skip it
+      if (typeof slotA === 'undefined' || slotA === null) {
+        return;
+      }
+
+      // no target means source doesn't need to merge onto anything
+      // just ensure source is callable (prepared format)
+      if (typeof slotB === 'undefined' || slotB === null) {
+        result[partName] = typeof slotA === 'function' ? slotA : () => slotA;
```
layershifter, line 66: I am not using there `callable()` as it always create a new closure

packages/fluentui/styles/src/mergeThemes.ts
```diff
@@ -45,105 +45,124 @@ export const emptyTheme: ThemePrepared = {
...
+  if (!isDebugEnabled) {
+    return mergeComponentStyles__PROD(stylesA, stylesB);
+  }
+
+  const mergedKeys = [...(stylesA ? Object.keys(stylesA) : []), ...(stylesB ? Object.keys(stylesB) : [])];
+  const result = {};
+
+  mergedKeys.forEach(slotName => {
+    const slotA = styleParam => {
+      // originalTarget is always prepared, fn is guaranteed, _debug always exists
```
layershifter, line 124:
```suggestion
```

packages/fluentui/styles/src/mergeThemes.ts
```diff
@@ -45,105 +45,124 @@ export const emptyTheme: ThemePrepared = {
...
+
+      // new object required to prevent circular JSON structure error in <Debug />
+      return {
+        ...styles,
+        _debug: _debug || [{ styles: { ...styles }, debugId: stylesA._debugId }],
+      };
+    };
+
+    const slotB = styleParam => {
+      // originalTarget is always prepared, fn is guaranteed, _debug always exists
```
layershifter, line 135:
```suggestion
```

packages/fluentui/styles/src/mergeThemes.ts
```diff
@@ -45,105 +45,124 @@ export const emptyTheme: ThemePrepared = {
...
+      return {
+        ...styles,
+        _debug: _debug || [{ styles: { ...styles }, debugId: stylesB._debugId }],
+      };
+    };
+
+    if (stylesA?.[slotName] && stylesB?.[slotName]) {
+      // We have both target and source, replace with merge fn
+      result[slotName] = styleParam => {
         // originalTarget is always prepared, fn is guaranteed, _debug always exists
```
layershifter, line 148:
```suggestion
        // slot* are always prepared, fn is guaranteed, _debug always exists
```

packages/fluentui/styles/src/mergeThemes.ts
```diff
@@ -45,105 +45,124 @@ export const emptyTheme: ThemePrepared = {
...
+
+      // new object required to prevent circular JSON structure error in <Debug />
+      return {
+        ...styles,
+        _debug: _debug || [{ styles: { ...styles }, debugId: stylesB._debugId }],
+      };
+    };
+
+    if (stylesA?.[slotName] && stylesB?.[slotName]) {
+      // We have both target and source, replace with merge fn
```
layershifter, line 146:
```suggestion
      // We have both, replace with merge fn
```

packages/fluentui/react-bindings/src/styles/resolveStyles.ts
```diff
@@ -100,7 +100,9 @@ export const resolveStyles = (
     const styles = allDisplayNames.map(displayName => theme.componentStyles[displayName]).filter(Boolean);
 
     if (styles.length > 0) {
-      mergedStyles = mergeComponentStyles(...styles);
+      mergedStyles = styles.reduce<ComponentSlotStylesPrepared>((acc, styles) => {
+        return mergeComponentStyles(acc, styles);
```
ling1726, line 104: nit: you can just arrow return, since there's no other logic in the block

packages/fluentui/styles/test/mergeThemes/mergeComponentStyles-test.ts
```diff
@@ -225,47 +227,25 @@ describe('mergeComponentStyles', () => {
...
-        const source2 = withDebugId(
-          {
-            root: {
-              a: 'tC',
-            },
-          },
-          'source2',
-        );
+        const target = withDebugId({ root: { a: 'tA' } }, 'target');
+        const source2 = withDebugId({ root: { a: 'tB' } }, 'source');
```
miroslavstastny, line 231: nit: Why is this `source2`, not just `source`?

miroslavstastny (approved):

packages/fluentui/styles/test/mergeThemes/mergeComponentStyles-test.ts
layershifter: Copy-paste issue, fixed 👍

Issue #7449, opened 2018-12-20
Search box re-render when value is null leads to TypeError


srinidhiprabhu96:
#### Priorities and help requested (not applicable if asking question):

Are you willing to submit a PR to fix? (No)

Requested priority: (Normal) 

Products/sites affected: (if applicable)

#### Issue:

Use a SearchBox inside another React component.
Do not type anything into the SearchBox.
Update the React component (without updating the SearchBox). This leads to re-render of the SearchBox.

#### Actual behavior:
The re-rendering causes this error:
TypeError: Cannot read property 'length' of null

#### Expected behavior:

There should be no error. The SearchBox component must normalize null or undefined to empty string on component update.

In the code, we see that if 'null' or 'undefined' is passed in the beginning, the constructor handles it and normalizes it to an empty string. However, there is no normalizing done when the component is re-rendered.

ecraig12345:
Hi @srinidhiprabhu96, thanks for reporting this. Would you mind sharing a link to a page where this repros, or [creating a codepen](http://aka.ms/fabricpen) to demonstrate?

srinidhiprabhu96:
You can find the codepen [here](https://codepen.io/srinidhiprabhu96/pen/dwNxEQ).

**To Repro:**
Click on the "Click Me" button.

**Expected:**
 The text on the button changes to "Hello".

**Actual:**
The button and search box are no longer displayed.
We get an error in the console: Uncaught TypeError: Cannot read property 'length' of null

PR #15385, opened 2020-10-06
Moving SearchBox to @fluentui/react-internal package

czearing:
#### Pull request checklist

#### Description of changes
1.	Moving `SearchBox` from react-next to react-internal.
2.	Updating imports for `SearchBox` in react-internal.
3.	Updating the top level files in react-next to reference @fluentui/react/lib/SearchBox.
4.	Deleting perf and vr tests for `SearchBoxNext`.
5.	Deleting react-next examples for `SearchBox`.


packages/react-internal/src/components/SearchBox/SearchBox.test.tsx
```diff
@@ -139,22 +149,20 @@ describe('SearchBox', () => {
     expect(wrapper.find('input').prop('value')).toBe('test');
   });
 
-  it('handles setting null value', () => {
```
ecraig12345, line 142: Remind me why we decided to remove this? It was actually added in response to an issue #7449--I'm not sure now that adding explicit null handling was the right thing to do, but we should at least keep a test to confirm that there's not an exception on either initially setting or updating the value to null.

czearing: Good catch! It looks this feature was not correctly ported with useControlableValue. I re-added the test and updated the base file so that value is set to a String. If the user passes in null for value they will receive “null”.

ecraig12345: I think the desired behavior is that if they pass `null` they get an empty string.

czearing: Technically this is true, but I am not sure if it is worth the extra line of code to support just null given that it is not a valid value. It is also should encourage the use of undefined or a blank string in the unlikely scenario that null is referenced.

ecraig12345: The problem is that changing the behavior now will likely result in the literal string "null" showing up in people's apps. (And if they've been passing that in they've already either disabled the type check by casting, or their app is in JS.)

ecraig12345 (approved):
Approved, and I guess we'll see if anyone complains about the "null" thing?

PR #15386, opened 2020-10-06
Persona: Move persona to react-internal

joschect:
#### Pull request checklist

- [x] Addresses an existing issue: Related to #15152

#### Description of changes

Moving function component conversion changes from #13645 back to the main package.


ecraig12345 (commented):
Mostly looks good but can you also update the examples?

Usually what I've been doing is:
1. Delete `react-examples/src/react/ComponentName`
2. Move `react-examples/src/react-next/ComponentName` to the path in 1
3. Look at the changes and see if there are any real changes which need to be preserved.
    - Any files with just import path changes can be reverted
    - All changes under `docs` should be ignored (the `react` version is authoritative)
    - In any files with real changes, be sure to update the imports to be from `react` not `react-next`.

You'll probably also need to run `update-snapshots` in `react-internal` and `react-examples`.

ecraig12345 (approved):
I went ahead and removed the examples to keep things moving, so this should be good to go now.

PR #15388, opened 2020-10-06
Move Fabric from react-next to react-internal

tomi-msft:
#### Pull request checklist

This PR moves the version of `Fabric` existing in `@fluentui/react-next` to `@fluentui/react-internal`. As part of this API and snapshots were updated, and the versions of these components in @fluentui/react-next now re-export the versions in @fluentui/react-internal.

#### Focus areas to test

(optional)


packages/react-internal/src/components/Fabric/Fabric.test.tsx
```diff
@@ -1,84 +1,78 @@
...
-import * as ReactDOM from 'react-dom';
-import * as renderer from 'react-test-renderer';
-import { mount } from 'enzyme';
+import { create } from '@uifabric/utilities/lib/test';
+import { Customizer } from '@uifabric/utilities';
 import { Fabric } from './Fabric';
-// import { Checkbox } from '../../Checkbox';
-// import { Customizer } from '../../Utilities';
-// import { createTheme } from '../../Styling';
+import { Checkbox } from '@fluentui/react/lib/Checkbox';
```
tomi-msft, line 5: @dzearing This component is not present in react-internal. What component within react-internal would you recommend using instead?

ecraig12345: For now let's just comment out this import and the parts of the test that use it, unless someone has a suggestion right away for another component to use.

ecraig12345:
You'll probably need to update snapshots in `react-internal` to get the build to pass.

ecraig12345 (approved):

PR #15389, opened 2020-10-06
Moving OverflowSet to @fluentui/react-internal package

czearing:
#### Pull request checklist
- [x] Part of #15152

#### Description of changes
1.	Moving `OverflowSet` from react-next to react-internal (including function component conversion from #13077).
2.	Updating imports for `OverflowSet` in react-internal.
3.	Updating the top level files in react-next to reference @fluentui/react/lib/OverflowSet.
4.	Deleting perf and vr tests for `OverflowSetNext`.
5.	Deleting react-next examples for `OverflowSet`.


ecraig12345:
Looks like this will require some updates in the charting package to use new props.

packages/react-examples/src/react/__snapshots__/CommandBar.Basic.Example.tsx.shot
```diff
@@ -827,10 +824,11 @@ exports[`Component Examples renders CommandBar.Basic.Example.tsx correctly 1`] =
             </div>
             <div
               className=
-                  ms-OverflowSet-overflowButton
+                  ms-CommandBar-primaryCommand
```
ecraig12345, line 827: Do you know why this class name change happened?

czearing: Thanks for catching this! getNativeProps was placed incorrectly in the wrapper div and was not applying the className properly.

ecraig12345: Nice to see the snapshot tests being useful! :)

packages/react-examples/src/charting/AreaChart/AreaChart.Multiple.Example.tsx
```diff
@@ -197,11 +197,6 @@ export class AreaChartMultipleExample extends React.Component<{}, IAreaChartBasi
             legendsOverflowText={'Overflow Items'}
             yAxisTickFormat={d3.format('$,')}
             legendProps={{
```
ecraig12345, line 199: Just to confirm, did you check out these examples to verify that they still work as before (even without the FocusZone)? You can compare with the PR deploy site of charting in master. http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/master/

packages/react-examples/src/react/__snapshots__/CommandBar.Basic.Example.tsx.shot
ecraig12345: Hmm original question still stands though--why did the class name change (even after the update to native props location)?

czearing: I updated the pr and it should be good now. The snapshot change was due to `ButtonGrid` applying {...props} after the classNames and then applying and overwriting them.

The remaining snapshot changes for `VerticalBarChart`,`AreaChart`, `OverflowSet`, and `ReizeGroup` are due to the `FocusZone` removal from `OverflowSet` and id updates:

packages/react-examples/src/charting/AreaChart/AreaChart.Multiple.Example.tsx
czearing: I don't see any behavior changes in the Charting components after the `FocusZone` removal.

packages/react-internal/src/components/OverflowSet/OverflowSet.scss
```diff
@@ -1,4 +1,4 @@
-@import '../../common/common';
+@import '~@fluentui/common-styles/dist/sass/common';
```
ecraig12345, line 1: I think we can delete this file--I don't see any references to it

ecraig12345 (approved):
Thanks for all the work on this! I'm going to push an update removing the scss file and then merge it.

Issue #15308, opened 2020-09-29
Make react-monaco-editor and export to codepen work with new package names


ecraig12345:
## Update

Most package names have been updated but `@fluentui/theme` and a couple others are still missing from `defaultSupportedPackages` (which prevents intellisense from working properly in a few cases).
https://github.com/microsoft/fluentui/blob/1cdf7704eb322c7bb621c9dd6188f75928c6532a/packages/react-monaco-editor/src/utilities/defaultSupportedPackages.ts#L64-L71

You can see the full list of deps that should be present (at least their unscoped names) by building then doing `ls packages/react-monaco-editor/dist/types`. As of writing:
```
date-time-utilities.d.ts
dom-utilities.d.ts
example-data.d.ts
font-icons-mdl2.d.ts
foundation-legacy.d.ts
merge-styles.d.ts
react-focus.d.ts
react-hooks.d.ts
react-window-provider.d.ts
react.d.ts
style-utilities.d.ts
theme.d.ts
utilities.d.ts
```

## Previous description

<details>
<summary>click to expand</summary>

tsx-editor currently has a lot of code which is specific to:
- [x] old package names
- [x] a limited set of old packages
- [x] old global names from bundles (like `window.Fabric`) (addressed in #15394)

This means even once it's compiling with the new package names, it will not actually work properly. It needs to be fixed sometime before we update the website so that export to codepen will still work in the beta docs.
</details>

ecraig12345:
There are still one or two packages missing from `defaultSupportedPackages`:
https://github.com/microsoft/fluentui/blob/1cdf7704eb322c7bb621c9dd6188f75928c6532a/packages/react-monaco-editor/src/utilities/defaultSupportedPackages.ts#L64-L71

You can see the full list of deps that should be present (at least their unscoped names) by building then doing `ls packages/react-monaco-editor/dist/types`.
```
date-time-utilities.d.ts
dom-utilities.d.ts
example-data.d.ts
font-icons-mdl2.d.ts
foundation-legacy.d.ts
merge-styles.d.ts
react-focus.d.ts
react-hooks.d.ts
react-window-provider.d.ts
react.d.ts
style-utilities.d.ts
theme.d.ts
utilities.d.ts
```

PR #15394, opened 2020-10-07
Fix codepens after replacing office-ui-fabric-react with @fluentui/react

xugao:
#### Pull request checklist

#### Description of changes

`react` package produces bundle: https://unpkg.com/@fluentui/react@7/dist/fluentui-react.js and lib var changed from `Fabric` to `FluentUIReact`

Related issue: #15308

#### Focus areas to test
Noticed this issue from [theme-designer](https://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/master/theming-designer/index.html), which is actually a `pr-deploy` site from `master`.

### DIFF
--- a/packages/tsx-editor/src/utilities/defaultSupportedPackages.ts
+++ b/packages/tsx-editor/src/utilities/defaultSupportedPackages.ts
@@ -1,7 +1,7 @@
 import { IPackageGroup } from '../interfaces/index';
 
 const fabricGroup: IPackageGroup = {
-  globalName: 'Fabric',
+  globalName: 'FluentUIReact',
   // Theoretically we could use import() here, but that pulls things into bundles when using
   // commonjs modules due to the way import is transpiled for commonjs
   // https://github.com/webpack/webpack/issues/5703#issuecomment-357512412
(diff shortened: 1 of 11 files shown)


packages/tsx-editor/src/utilities/defaultSupportedPackages.ts
```diff
@@ -1,7 +1,7 @@
 import { IPackageGroup } from '../interfaces/index';
 
 const fabricGroup: IPackageGroup = {
-  globalName: 'Fabric',
+  globalName: 'FluentUIReact',
```
ecraig12345, line 4: I think there's a test you'll need to update as well. (search for Fabric in tsx-editor)

ecraig12345 (approved):
Thanks for doing this! I had a tracking issue #15308 for some of this work but hadn't gotten to look into it yet.

I suspect there may still be a few more updates needed in tsx-editor to get things fully working. In particular there are a bunch of new package names that are missing from some hardcoded lists (used when determining if export to codepen is enabled) and/or possibly not included as part of the main bundle. I'm fine with either merging this mostly as-is once the build passes (and fully fixing it later) or doing it all now.

PR #15396, opened 2020-10-07
Delete update-api script which is no longer needed

ecraig12345:
Now that API Extractor updates are run automatically, we don't need the `update-api` command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "test:fluentui:circulars": "gulp test:circulars:run",
     "test:fluentui:e2e": "yarn workspace @fluentui/e2e test",
     "test:fluentui:projects": "yarn workspace @fluentui/projects-test test",
-    "update-api": "cd packages/react && yarn update-api",
+    "update-api": "echo API files are now updated when running yarn build",
     "update-snapshots": "lage update-snapshots --verbose",
     "update-a11y": "cd apps/a11y-tests && yarn update-snapshots",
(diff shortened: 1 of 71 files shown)


behowell (approved):

khmakoto (approved):

package.json
```diff
@@ -61,7 +61,7 @@
     "test:fluentui:circulars": "gulp test:circulars:run",
     "test:fluentui:e2e": "yarn workspace @fluentui/e2e test",
     "test:fluentui:projects": "yarn workspace @fluentui/projects-test test",
-    "update-api": "cd packages/react && yarn update-api",
+    "update-api": "echo API files are now updated when running yarn build",
```
xugao, line 64: i would say just remove this.

xugao (approved):

package.json
ecraig12345: After some discussions I "overheard" (in a giant group chat) with people from outside the core team yesterday, I actually wonder if we should leave both this *and* the individual package update-api commands for awhile (with the same message). Basically the issue is that when we change the workflow, how do we ensure people know about it--both the core team, semi-frequent contributors outside the team, and occasional contributors when they come back to the repo in a couple months? Granted this one is not quite as bad since people only tended to run the command when they encounter an error because they haven't run it, but worth considering in general.

layershifter:
@ecraig12345 is it still a thing? 🤔

PR #15399, opened 2020-10-07
Additional ChoiceGroup v8 cleanup

ecraig12345:
#### Pull request checklist

#### Description of changes

Additional cleanup related to the FC/v8 changes: primarily around removing more deprecated things and cleaning up handling of keys and checked state.

- Remove deprecated `IChoiceGroupOption.checked` and replace it with `IChoiceGroupOptionProps.checked` for custom rendering use only
- Make `IChoiceGroupOptionProps.itemKey` required (and `IChoiceGroupOptionProps.itemKey` optional)
- Update custom render callbacks to take `IChoiceGroupOptionProps` so that they can have the proper `checked` values.
  - In `ChoiceGroupOption`, when passing the option back to custom render callbacks, mix the `key` back in 
- Remove deprecated `OnChangeCallback` and `OnFocusCallback`
- Add explicit FC type to ChoiceGroupOption
- When creating the options, don't let users' custom `onChange`/`onFocus`/`onBlur` overwrite the default ones provided by ChoiceGroup. And in the unlikely event that these custom handlers are provided (I don't know why this would be needed on individual options), call them from the default handlers.
- Update release notes

### DIFF
--- a/change/@fluentui-react-internal-2020-10-22-00-02-17-choicegroup-updates.json
+++ b/change/@fluentui-react-internal-2020-10-22-00-02-17-choicegroup-updates.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "ChoiceGroup option props cleanup",
+  "packageName": "@fluentui/react-internal",
+  "email": "elcraig@microsoft.com",
+  "dependentChangeType": "patch",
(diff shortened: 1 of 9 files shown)


xugao (approved):
LGTM!

dzearing (approved):

PR #15402, opened 2020-10-07
proto: useClasses

layershifter:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/react-bindings/src/hooks/useClasses.ts
+++ b/packages/fluentui/react-bindings/src/hooks/useClasses.ts
@@ -1,0 +1,216 @@
+import hash from '@emotion/hash';
+import { RendererParam } from '@fluentui/react-northstar-styles-renderer';
+import {
+  ComponentSlotStyle,
+  ComponentSlotStylesInput,
+  ComponentSlotStylesPrepared,
+  ComponentStyleFunctionParam,
+  ComponentVariablesInput,
+  ComponentVariablesObject,
+  ICSSInJSStyle,
+  mergeComponentStyles,
+  withDebugId,
+} from '@fluentui/styles';
+import cx from 'classnames';
+import * as _ from 'lodash';
+
+import { useFluentContext } from '../context';
+import { ComponentDesignProp, PrimitiveProps } from '../styles/types';
+
+type UseStylesOptions<StyleProps extends PrimitiveProps> = {
+  /** A classname that will be added by default to all instances of a component on the `root` slot. */
+  className?: string;
+
+  /**
+   * A mapping from component's props to styles functions props. Can be only primitive types as they will be used for
+   * cache keys.
+   */
+  mapPropsToStyles?: () => StyleProps;
+
+  /**
+   * A set props of that contain mapping for props that perform inline styles overrides, for example `styles` or
+   * `variables`.
+   */
+  mapPropsToInlineStyles?: () => InlineStyleProps<StyleProps>;
+};
+
+type InlineStyleProps<StyleProps> = {
+  /** Additional CSS class name(s) to apply.  */
+  className?: string;
+
+  design?: ComponentDesignProp;
+
+  /** Additional CSS styles to apply to the component instance.  */
+  styles?: ComponentSlotStyle<StyleProps, any>; // TODO: see if we can improve it
+
+  /** Override for theme site variables to allow modifications of component styling via themes. */
+  variables?: ComponentVariablesInput;
+};
+
+export const useClasses = <StyleProps extends PrimitiveProps>(
+  displayName: string,
+  options: UseStylesOptions<StyleProps>,
+): string => {
+  const { disableAnimations, performance: performanceFlags, telemetry, theme, renderer, rtl } = useFluentContext();
+  const {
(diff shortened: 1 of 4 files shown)


packages/fluentui/react-bindings/src/hooks/useClasses.ts
```diff
@@ -0,0 +1,218 @@
...
+      resolvedVariablesHash = hash(JSON.stringify(resolvedVariables));
+
+      // @ts-ignore
+      theme.componentVariables[displayName].__resolved = resolvedVariables;
+      // @ts-ignore
+      theme.componentVariables[displayName].__hash = resolvedVariablesHash;
+    }
+  } else if (typeof theme.componentVariables[displayName] !== 'undefined') {
+    throw new Error();
+  }
```
ling1726, line 96: Maybe I'm remembering this wrong... but can't componentVariables be an object too ? 

Also there is no message in this error, it might be really hard to understand what is going when this throws if you are consuming it

packages/fluentui/react-bindings/src/hooks/useClasses.ts
```diff
@@ -0,0 +1,218 @@
...
+  // Merge theme styles with inline overrides if any
+  let mergedStyles: ComponentSlotStylesPrepared;
+
+  if (noInlineStylesOverrides) {
+    mergedStyles = theme.componentStyles[displayName];
+  } else {
+    mergedStyles = mergeComponentStyles(
+      theme.componentStyles[displayName],
+      design && withDebugId({ root: design }, 'props.design'),
+      styles && withDebugId({ root: styles } as ComponentSlotStylesInput, 'props.styles'),
```
ling1726, line 144: didn't you want to work towards a signature that only allowed 2 parameters ?

packages/fluentui/react-bindings/src/hooks/useClasses.ts
```diff
@@ -0,0 +1,218 @@
...
+  // To compute styles we are going through three stages:
+  // - resolve variables (siteVariables => componentVariables + props.variables)
+  // - resolve styles (with resolvedVariables & props.styles & props.design)
+  // - compute classes (with resolvedStyles)
+  // - conditionally add sources for evaluating debug information to component
+
+  const telemetryVariablesStart = telemetry?.enabled ? performance.now() : 0;
+
+  let resolvedVariablesHash: string = '';
+  let resolvedVariables: ComponentVariablesObject = {};
```
ling1726, line 76: maybe you can create a helper function to return these two variables with the logic below that sets them ?

ling1726: i.e. `resolveComponentVariables(theme.componentVariables[displayName])`

packages/fluentui/react-bindings/src/hooks/useClasses.ts
```diff
@@ -0,0 +1,218 @@
...
+  /* istanbul ignore else */
+  if (process.env.NODE_ENV !== 'production') {
+    if (!performanceFlags.enableStylesCaching && performanceFlags.enableBooleanVariablesCaching) {
+      throw new Error(
+        '@fluentui/react-northstar: Please check your "performance" settings on "Provider", to enable "enableBooleanVariablesCaching" you need to enable "enableStylesCaching"',
+      );
+    }
+  }
+
+  if (performanceFlags.enableBooleanVariablesCaching) {
```
ling1726, line 116: nit: declare the variable closer to the usage

```suggestion
  let noVariableOverrides = performanceFlags.enableBooleanVariablesCaching || !variables;
  if (performanceFlags.enableBooleanVariablesCaching) {
```

packages/fluentui/react-bindings/src/hooks/useClasses.ts
```diff
@@ -0,0 +1,218 @@
...
+  if (telemetry?.enabled && telemetry.performance[displayName]) {
+    telemetry.performance[displayName].msResolveStylesTotal += performance.now() - telemetryStylesStart;
+  }
+
+  let classes: string = '';
+  const telemetryClassesStart = telemetry?.enabled ? performance.now() : 0;
+
+  if (cacheEnabled && noInlineStylesOverrides) {
+    // @ts-ignore
+    if (mergedStyles.root[componentCacheKey + 'classes']) {
```
ling1726, line 197: nit: can be moved to variable

```suggestion
    const cacheKey = `${componentCacheKey}classes` 
    if (mergedStyles.root[cacheKey ]) {
```

packages/fluentui/react-bindings/src/hooks/useClasses.ts
```diff
@@ -0,0 +1,218 @@
...
+      // @ts-ignore
+      classes = mergedStyles.root[componentCacheKey + 'classes'];
+
+      if (telemetry?.performance[displayName]) {
+        telemetry.performance[displayName].stylesRootCacheHits++;
+      }
+    } else {
+      classes = renderer.renderRule(resolvedStyles, rendererParam);
```
ling1726, line 205: would it be useful to monitor cache misses here ?

PR #15405, opened 2020-10-07
Expose FileTypeIcons as HTML strings

arkogupta:
#### Pull request checklist

#### Description of changes

Rather than exposing the src of the `img` elements, we can just clone the whole JSX element as a string. If in the future, the implementation changes from `` to something else, then the corresponding DOM element which will be used can again be stored as a string in the same dictionary, so that the consumer can continue to consume it seamlessly.

#### Focus areas to test

(optional)


packages/file-type-icons/src/initializeFileTypeIcons.tsx
```diff
@@ -8,43 +8,69 @@ const SVG_SUFFIX = '_svg';
...
   iconTypes.forEach((type: string) => {
     const baseUrlSizeType = baseUrl + size + '/' + type;
-    fileTypeIcons[type + size + PNG_SUFFIX] = <img src={baseUrlSizeType + '.png'} alt="" />;
-    fileTypeIcons[type + size + SVG_SUFFIX] = <img src={baseUrlSizeType + '.svg'} alt="" />;
+    let src: string;
+    let iconName: string;
+
+    src = `${baseUrlSizeType}.png`;
+    iconName = type + size + PNG_SUFFIX;
+    fileTypeIcons[iconName] = <img src={src} alt="" />;
```
Ridvanovskyy, line 36: minor question. Does this alt do something? Looks like it doesn't play any role here.

dzearing: May be wrong, but I believe this is due to accessibility scanners which will warn when you have an image without considering the alt tag (even providing a blank one is considered "accessible")...

packages/file-type-icons/src/initializeFileTypeIcons.tsx
```diff
@@ -8,43 +8,69 @@ const SVG_SUFFIX = '_svg';
...
+    _fileTypeIconsAsString[iconName] = `<img src="${src}" alt="" />`;
+
+    src = `${baseUrlSizeType}.svg`;
+    iconName = type + size + SVG_SUFFIX;
+    fileTypeIcons[iconName] = <img src={src} alt="" />;
+    _fileTypeIconsAsString[iconName] = `<img src="${src}" alt="" />`;
```
dzearing, line 42: Rather than defining this on every definition, why not just generate it on demand in the exported function`getFileTypeIconAsHTMLString`

KevinTCoughlin: This icon map is quite large in memory so it would be nice if made on-demand rather than bloat the page more.

packages/file-type-icons/src/initializeFileTypeIcons.tsx
bigbadcapers: Might not be a bad idea to populate this alt text with `type + " icon"` ~ the type is not always super descriptive but we try our best to make it so.

packages/file-type-icons/src/initializeFileTypeIcons.tsx
```diff
@@ -5,15 +5,35 @@ import { FileTypeIconMap } from './FileTypeIconMap';
...
+//     if (pixelRatio) {
+//       src = `${baseUrl}${size}_${pixelRatio}/${type}.${suffix}`
+//       return `<img src="${src}" height="100%" width="100%" />`;
+//     } else {
+//       src = `${baseUrl}${size}/${type}.${suffix}`;
+//       return `<img src="${src}" alt="" />`;
+//     }
+//   }
+// }
+
```
arkogupta, line 36: Hi @dzearing @KevinTCoughlin 
Please have a look at this iteration. I had to do some refactoring and export a couple of functions from `getFileTypeIconProps` to be able to implement the function `getFileTypeIconAsHTMLString`  which is basically kind of a sister clone of  `getFileTypeIconProps` in that it accepts the same kind of object but rather than just returning the iconName, it returns the entire DOM element as a string.

The other option I had considered is this commented snippet; in which the consumer had to call `getFileTypeIconProps` to get the iconName and then just pass the iconName to the `getFileTypeIconAsHTMLString` function, but that IMO required some regEx parsing, and hence I avoided this approach.

khmakoto:

arkogupta:
@KevinTCoughlin @dzearing please have a look whenever you get time. Thanks :)

packages/file-type-icons/src/initializeFileTypeIcons.tsx
```diff
@@ -5,15 +5,36 @@ import { FileTypeIconMap } from './FileTypeIconMap';
...
+export const DEFAULT_BASE_URL = 'https://spoprod-a.akamaihd.net/files/fabric-cdn-prod_20201008.001/assets/item-types/';
+export const ICON_SIZES: number[] = [16, 20, 24, 32, 40, 48, 64, 96];
 
 export function initializeFileTypeIcons(baseUrl: string = DEFAULT_BASE_URL, options?: Partial<IIconOptions>): void {
   ICON_SIZES.forEach((size: number) => {
     _initializeIcons(baseUrl, size, options);
   });
 }
 
+// export function getFileTypeIconAsHTMLString(iconName: string, baseUrl: string = DEFAULT_BASE_URL): string |
```
dzearing, line 17: intended to be commented out? (or removed)

packages/file-type-icons/src/initializeFileTypeIcons.tsx
dzearing: can you remove the commented code

arkogupta: Removed.

packages/file-type-icons/src/initializeFileTypeIcons.tsx
arkogupta: That is a good suggestion! Will take that up as a separate PR. Thanks!

change/@uifabric-file-type-icons-2020-10-07-21-27-44-arkgupta-fileTypeIconsHTML.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "fileicons as html strings",
```
dzearing, line 3:
```suggestion
  "comment": "Icons can now be requested using the getFileTypeIconAsHTMLString API, for compatibility with non-React environments.",
```

packages/file-type-icons/src/getFileTypeIconAsHTMLString.ts
```diff
@@ -0,0 +1,28 @@
+import { DEFAULT_BASE_URL } from './initializeFileTypeIcons';
+import {
+  getFileTypeIconNameFromExtensionOrType,
+  getFileTypeIconSuffix,
+  DEFAULT_ICON_SIZE,
+  IFileTypeIconOptions,
+} from './getFileTypeIconProps';
+
+export function getFileTypeIconAsHTMLString(
```
dzearing, line 9: Can you please add a comment to this? Also the function is not being exported through the index file. And you will probably need to run `yarn update-api` to update the api-extractor definition of the library.

packages/file-type-icons/src/initializeFileTypeIcons.tsx
```diff
@@ -5,8 +5,8 @@ import { FileTypeIconMap } from './FileTypeIconMap';
 const PNG_SUFFIX = '_png';
 const SVG_SUFFIX = '_svg';
 
-const DEFAULT_BASE_URL = 'https://spoprod-a.akamaihd.net/files/fabric-cdn-prod_20201008.001/assets/item-types/';
-const ICON_SIZES: number[] = [16, 20, 24, 32, 40, 48, 64, 96];
+export const DEFAULT_BASE_URL = 'https://spoprod-a.akamaihd.net/files/fabric-cdn-prod_20201008.001/assets/item-types/';
```
dzearing, line 8: A little confused why these are being exported.

dzearing: (and the icon_sizes)

arkogupta: I am importing both of these in my test file. Also I need the default url while emitting the HTML as string from the new API.

packages/file-type-icons/src/getFileTypeIconAsHTMLString.ts
arkogupta: Added a comment and exported through the index file. I couldn't find an existing api.md or api-extractor.json file for this package. Should I add one?

arkogupta: @dzearing gentle ping! :)

dzearing:
I am going to sign off to unblock the scenario; however I do think longer term we need to split out a separate `@fluentui/file-type-icons` package which is more the base package that has no React dependencies. This is where we'd do something like export a `getFileTypeIcon` api which returns more agnostic information in a structure.

change/@uifabric-file-type-icons-2020-10-07-21-27-44-arkgupta-fileTypeIconsHTML.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
ecraig12345, line 2: For future reference, changes adding new public APIs should be minor not patch.

packages/file-type-icons/src/getFileTypeIconAsHTMLString.ts
ecraig12345: We should add an API extractor config and API file for this package (same pattern as other packages).

ecraig12345 (commented):
Coming in late here (sorry) but a couple comments to potentially address in the future.

PR #13689, opened 2020-06-18
Updating SpinButton to a function component within React-Next

czearing:
#### Pull request checklist
- [X] Include a change request file using `$ yarn change`

#### Description of changes

1. Updating `SpinButton` to a function component
    - Updated to correctly accept user-entered values when value prop is set
    - Removed deprecated test
    - Updated styling
2. Added `useSetTimeOut` hook 

#### Focus areas to test
`SpinButton`

PR #15408, opened 2020-10-07
Moving SpinButton to @fluentui/react-internal package

czearing:
#### Pull request checklist

#### Description of changes
1.	Moving `SpinButton` from react-next to react-internal (including function component changes from #13689).
2.	Updating imports for `SpinButton` in react-internal.
3.	Updating the top level files in react-next to reference @fluentui/react/lib/SpinButton.
4.	Deleting perf and vr tests for `SpinButtonNext`.
5.	Deleting react-next examples for `SpinButton`.

Also from @ecraig12345: Add list of removed style props in release notes, and an example of how to use conditional styling with a style function as a replacement.


packages/react-internal/src/components/SpinButton/SpinButton.types.ts
```diff
@@ -297,7 +272,7 @@ export interface ISpinButtonStyles {
   /**
    * Style for the icon when the control is disabled.
    */
-  iconDisabled: IStyle;
+  // iconDisabled: IStyle;
```
ecraig12345, line 275: Can you get rid of these commented props?

ecraig12345: Never mind I fixed this. Also updated the release notes about the removed props and added an example of what to do instead.

ecraig12345 (approved):

packages/react-internal/src/components/SpinButton/SpinButton.types.ts
czearing: Thank you!

ecraig12345:
Looks like azure theme and maybe others are using some of the deprecated props. @czearing can you fix this (basically switching to style functions like in the example I added)? 
```
verb |  [10:53:39 PM] x src/azure/styles/SpinButton.styles.ts:12:5 - error TS2322: Type '{ input: { backgroundColor: string; color: string; fontSize: string | number | undefined; }; inputTextSelected: { color: string; fontSize: string | number | undefined; }; }' is not assignable to type 'Partial<ISpinButtonStyles>'.
verb |    Object literal may only specify known properties, and 'inputTextSelected' does not exist in type 'Partial<ISpinButtonStyles>'.
verb |  
verb 12     inputTextSelected: {
verb |         ~~~~~~~~~~~~~~~~~~~~
verb |  13       color: semanticColors.inputText,
verb |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
verb |  14       fontSize: theme.fonts.medium.fontSize,
verb |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
verb |  15     },
```

PR #15409, opened 2020-10-07
Removing unused eslint rules from @fluentui/react-internal TextField.test.tsx

czearing:
#### Pull request checklist

#### Description of changes
Removing unused eslint-disable-next-line no-explicit-any rules from `TextField.test.tsx` within react-internal.


ecraig12345 (approved):
Thanks!

PR #15411, opened 2020-10-07
Moving TeachingBubble to @fluentui/react-internal package

czearing:
#### Pull request checklist

#### Description of changes
1.	Moving `TeachingBubble` from react-next to react-internal.
2.	Updating imports for `TeachingBubble` in react-internal.
3.	Updating the top level files in react-next to reference @fluentui/react/lib/TeachingBubble.
4.	Deleting perf and vr tests for `TeachingBubbleNext`.
5.	Deleting react-next examples for `TeachingBubble`.


packages/react-internal/src/components/TeachingBubble/TeachingBubble.base.tsx
```diff
@@ -1,103 +1,88 @@
...
 
-    this._defaultCalloutProps = {
-      beakWidth: 16,
-      gapSpace: 0,
-      setInitialFocus: true,
-      doNotLayer: false,
-      directionalHint: DirectionalHint.rightCenter,
-    };
-  }
+  const calloutProps = { ...defaultCalloutProps, ...setCalloutProps };
```
ecraig12345, line 57: Just realized, this needs to be memoized to avoid constant style re-calculation. You can do it like this to also preserve decent type safety:
```ts
  const calloutProps: ICalloutProps = React.useMemo(() => ({ ...defaultCalloutProps, ...setCalloutProps }), [
    setCalloutProps,
  ]);

  const stylesProps: ITeachingBubbleStyleProps = {
    theme: theme!,
    isWide,
    calloutProps: calloutProps as ICalloutProps & Required<Pick<ICalloutProps, 'theme'>>,
    hasCloseButton,
  };
```

ecraig12345:
One screener change (and a merge conflict). I think the difference is that the first "Try It" button is now initially focused on open? (it's slightly darker which IIRC means focused)
https://screener.io/v2/states/Skgr32fUL.microsoft-fluentui/feat%2Freact-internal-teachingBubble/1024x768/Chrome/coachmark-next-expanded-default-0

packages/react-internal/src/components/TeachingBubble/TeachingBubble.base.tsx
ecraig12345: Realized my initial suggestion was slightly wrong, you do have to mix in theme manually. I pushed a fix.

ecraig12345 (approved):

PR #15412, opened 2020-10-07
[Do not review] Enabling selection for input changed for UnifiedPicker and updating example for that case

nebhatna:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

### DIFF
--- a/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
+++ b/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
@@ -278,11 +278,19 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
       if (!resultItemsList) {
         resultItemsList = [];
       }
-      onInputChange ? onInputChange(value, composing, resultItemsList) : null;
-
-      if (resultItemsList && resultItemsList.length > 0) {
-        addItems(resultItemsList);
+      if (onInputChange) {
+        onInputChange(value, composing, resultItemsList);
+        if (resultItemsList && resultItemsList.length > 0) {
+          addItems(resultItemsList);
+          showPicker(false);
+          setQueryString('');
+          if (input.current) {
+            input.current._value = '';
+          }
+        }
       }
+    } else {
+      alert(composing);
     }
   };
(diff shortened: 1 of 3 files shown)


packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -270,12 +270,25 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
       props.inputProps.onClick(ev as React.MouseEvent<HTMLInputElement>);
     }
   };
-  const _onInputChange = (value: string, composing?: boolean) => {
+  const _onInputChange = (value: string, composing?: boolean, resultItemsList?: T[]) => {
```
elisabethcvs, line 273: Would this ever have a value?

packages/react-examples/src/experiments/UnifiedPeoplePicker/DoubleUnifiedPeoplePicker.Example.tsx
```diff
@@ -179,14 +161,54 @@ const UnifiedPeoplePickerExample = (): JSX.Element => {
...
     const suggestionList = suggestions.map(item => {
       return { item: item, isSelected: false, key: item.key } as IFloatingSuggestionItem<IPersonaProps>;
     });
-    // We want to show top 5 results
-    setPeopleSuggestions(suggestionList.splice(0, 5));
+
+    const updatedItems: IPersonaProps[] = [];
+    const currentItems: IPersonaProps[] = [...peopleSelectedItems];
+
+    for (let i = 0; i < currentItems.length; i++) {
```
elisabethcvs, line 184: I don't think you need this, you can have the declaration be this right?
const updatedItems: IPersonaProps[] = [...peopleSelectedItems];

elisabethcvs:
There's paste logic in the other two examples, should those get updated as well?

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
nebhatna: yes, it would, this is what is getting updated

PR #15428, opened 2020-10-08
Add drop to the end of the list to UPP, fix minor bugs

elisabethcvs:
#### Pull request checklist

#### Description of changes
This won't be checked in until the 8 release is available

- Add props/functionality so that users can drop over the input (making it look like the item is dropped at the end of the list or in the empty UPP
- add a prop so consumers can disable drag drop in the UPP
- add a compare function to the selected items list so we don't have to rely on findIndex (which does a strict compare)  to find the index of a list item

Modified the examples to show the new options:
- double UPP example has everything enabled
- single UPP has drag drop enabled, but not drop over the input
- UPP with edit has drag drop disabled


elisabethcvs:
I need a change file

packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -66,6 +66,16 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
     onInputChange,
   } = props;
 
+  const defaultDragDropEnabled = React.useMemo(
+    () => (props.defaultDragDropEnabled !== undefined ? props.defaultDragDropEnabled : true),
```
nebhatna, line 70: can props.defaultDragDropEnabled be null too ? i would check on !props.defaultDragDropEnabled

elisabethcvs: I tried that - if the prop is set, it then does a bool compare, so this wasn't working. I should check for null too though, I'll do some comparisons to see what statement will work

elisabethcvs: Though also can you set a boolean value to null?

elisabethcvs: Ok I checked, you can't, this value will be either undefined or a boolean

packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -105,16 +115,49 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
+      event?.preventDefault();
+    }
+  };
+
+  const _onDropAutoFill = (event?: React.DragEvent<HTMLDivElement>) => {
+    if (props.onDropAutoFill) {
+      props.onDropAutoFill(event);
+    } else {
+      insertIndex = selectedItems.length;
+      event?.preventDefault();
```
nebhatna, line 129: do we want to do event?.preventDefault irrespective ? Since we would be leaving that responsibility on props.onDropAutoFill otherwise.

nebhatna (changes requested):
few minor comments

packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
elisabethcvs: I think that's ok? If they choose to override it, they can control this? There are times you may want to call preventDefault selectively

nebhatna: from what i have seen, we usually do end up doing this, since it overrides expected behavior and ends up doing unwanted interactions.. maybe worth while to do so irrespective , than leavng it on the props, to figure it out/

nebhatna (approved):

PR #15430, opened 2020-10-08
Moving SelectedItemsList to @fluentui/react-internal package

czearing:
#### Pull request checklist

#### Description of changes
1.	Moving `SelectedItemsList` from react-next to react-internal.
2.	Updating imports for `SelectedItemsList` in react-internal.
3.	Updating the top level files in react-next to reference @fluentui/react/lib/SelectedItemsList.
4.	There where no perf or vr tests for `SelectedItemsListNext` to be removed.
5.	Deleting react-next examples for `SelectedItemsList`.


packages/react-examples/src/react-next/ExtendedPicker/ExtendedPeoplePicker.Controlled.Example.tsx
```diff
@@ -139,8 +139,8 @@ export class ExtendedPeoplePickerControlledExample extends React.Component<{}, I
       copyMenuItemText: 'Copy name',
       editMenuItemText: 'Edit',
       getEditingItemText: this._getEditingItemText,
-      onRenderFloatingPicker: FloatingPeoplePicker,
-      floatingPickerProps: this._floatingPickerProps,
+      onRenderFloatingPicker: FloatingPeoplePicker as any,
```
ecraig12345, line 142: Do you know why this was needed?

ecraig12345 (approved):
LGTM but probably @joschect should review too.

packages/react-examples/src/react-next/ExtendedPicker/ExtendedPeoplePicker.Controlled.Example.tsx
czearing: It is because the import for SelectedItemsList comes from `react-internal` while FloatingPeoplePicker comes from `react-next`.

PR #15431, opened 2020-10-08
Fix issues with SpinButton revealed by eslint-plugin-react-hooks

### LINKED ISSUE #5326: SpinButton: Add onChange handler

### LINKED ISSUE #14480: SpinButton: fix issues revealed by eslint-plugin-react-hooks

### LINKED ISSUE #15970: SpinButton doesn't work (v8-beta). increment/decrement work only once.
### Environment Information

- **Package version(s)**: 8.0.0-beta.15
- **Browser and OS versions**: Firefox 83 / Chrome 87

https://codepen.io/k_oi/pen/abZxEwq

#### Actual behavior:

increment/decrement only once.
displayed value is 9 or 11 only.

#### Expected behavior:

can incremenet/decrement specific range (min-max).

### Others

* uncontrolled/controlled method doesn't work.
* onIncrement/onDecrement event is fired but argument value is always initlal value.

Requested priority: Normal

czearing:
#### Pull request checklist
- [x] Fixes #14480, fixes #15970 

#### Description of changes
~~1. Updating react-internal `SpinButton` to handle onChange.~~
2. Resolving issues revealed by eslint-plugin-react-hooks. https://github.com/microsoft/fluentui/issues/14480
~~3. Updating react-internal `SpinButton.stateful.example.tsx` to show the usage of onChange.~~
4. Adding a forwardRef to `SpinButton`.

#### Focus areas to test
`SpinButton`


xugao (approved):

ecraig12345 (changes requested):
Please wait on merging this until I can take a look again--I remember from the previous version of this that I had quite a bit of feedback, but I need to get some other things with v8 done ASAP and don't have time right away to review this in detail, sorry. :(

dzearing (approved):

behowell:
Hi @czearing, I wanted to check if you are still planning to go forward with this PR?

It looks like it's waiting for review from @ecraig12345.

packages/react-internal/src/components/SpinButton/SpinButton.types.ts
```diff
@@ -34,7 +34,9 @@ export enum KeyboardSpinDirection {
 /**
  * {@docCategory SpinButton}
  */
-export interface ISpinButtonProps extends React.HTMLAttributes<HTMLDivElement>, React.RefAttributes<HTMLDivElement> {
+export interface ISpinButtonProps
+  extends React.HTMLAttributes<HTMLInputElement | HTMLTextAreaElement>,
```
ecraig12345, line 38: HTMLTextAreaElement shouldn't be included in the props here. If this was copied from TextField, it's only included there because TextField actually can be an `input` or a `textarea` internally (but SpinButton can only be an `input`).

ecraig12345:
After thinking about this and reading back through the great (and long) discussion on #5326, I think we should go ahead with the functionality fixes in this PR but revert the onChange addition for now. (But definitely get it done before the v8 release still!)

Correctly implementing onChange has a lot of nuances which weren't covered here: mainly that we should only call it on "commit" rather than on every keystroke, plus lots of edge cases. See https://hackmd.io/cLn40UjfT_KOM6r4yi2CUQ for a full proposed spec. If the proposal seems good, either I can work on it (hopefully this week) or @czearing you can give it a try if you have time.

Based on all that, I pushed some updates doing the following:
- Revert onChange addition for now
- Add a test for the case of pressing up/down arrow key or button twice
- Revert my previous example changes (see #16040 instead)
- Improve default handling
- Simplify calculation of initial value, last valid value, and precision
- Add helpers in test to reduce duplicate code (makes adding more coverage easier)
- Fix a bug where `lastValidValue` was not being updated after stepping (and add a test)

At this point the functionality issues should be fixed and I'm going to approve, but @dzearing should probably take a look as well so I'm not reviewing (partially) my own code. 😆

ecraig12345 (approved):

ecraig12345:
I should also have said, thank you so much @czearing for all your work on this and other changes, and SO sorry for not reviewing sooner!!

kazuki:
In controlledValue mode, `setValue(elementValue)` in `handleInputChange` doesn't work.
As a result, if input by the keyboard, the value will not be reflected correctly.

ecraig12345:
@kazuki The controlled SpinButton isn't working quite properly yet. I'll be fixing that in another PR this week.

PR #15432, opened 2020-10-08
Add option for selectable headers and footers to FloatingSuggestionsComposite (and therefor the UnifiedPicker)

elisabethcvs:
#### Pull request checklist

#### Description of changes
Changes won't be checked in until the v8 stuff is ready.

Adding a property so that consumers can add selectable footers and headers to the floating picker. It wasn't possible to keyboard navigate to headers or footers previously. This property takes precedence over the old render functions, but I've kept them, so this change is backwards compatible.

I added keyboard navigation to the UnifiedPicker as well.

I added an example for the floating suggestions composite and I modified the UPP single example so that the picker has headers and footers.

#### Focus areas to test

I have checked that keyboard navigation works properly when:
- no headers or footers
- non selectable headers and footers
- non selectable header, selectable footers
- non selectable header, one selectable footer
- selectable header, one selectable footer


packages/react-examples/src/react-experiments/UnifiedPeoplePicker/UnifiedPeoplePicker.Example.tsx
```diff
@@ -62,6 +63,48 @@ export const UnifiedPeoplePickerExample = (): JSX.Element => {
 
   const ref = React.useRef<any>();
 
+  const suggestionProps = useConst(() => {
+    return {
+      /*headerItemsProps: [
+        {
+          renderItem: () => {
+            return <>People Suggestions</>;
```
nebhatna, line 71: please remove this if it is not needed

packages/react-examples/src/react-experiments/UnifiedPeoplePicker/UnifiedPeoplePicker.Example.tsx
```diff
@@ -62,6 +63,48 @@ export const UnifiedPeoplePickerExample = (): JSX.Element => {
...
+        {
+          renderItem: () => {
+            return <>Showing {peopleSuggestions.length} results</>;
+          },
+          shouldShow: () => {
+            return peopleSuggestions.length > 0;
+          },
+          /*onExecute: () => {
+            alert('Showing people suggestions executed');
+          },*/
```
nebhatna, line 91: same

packages/react-examples/src/react-experiments/UnifiedPeoplePicker/UnifiedPeoplePicker.Example.tsx
elisabethcvs: So I was thinking leave them there so users can comment/uncomment to see the various options for headers and footers. I should probably add a comment explaining that though

packages/react-examples/src/react-experiments/FloatingPeopleSuggestions/FloatingPeopleSuggestions.HeaderFooter.Example.tsx
```diff
@@ -0,0 +1,256 @@
...
+        break;
+      case KeyCodes.down:
+        ev.preventDefault();
+        ev.stopPropagation();
+        selectNextItem();
+        break;
+    }
+  };
+
+  const _renderExtendedPicker = () => {
```
nebhatna, line 213: rename _renderExtendedPicker to renderFloatingSuggestions

packages/react-experiments/src/components/FloatingSuggestionsComposite/FloatingSuggestionsList/FloatingSuggestionsList.tsx
```diff
@@ -7,6 +7,7 @@ import {
 } from './FloatingSuggestionsList.types';
 import { FloatingSuggestionsItemMemo } from '../FloatingSuggestionsItem/FloatingSuggestionsItem';
 import { getStyles } from './FloatingSuggestionsList.styles';
+import { ISuggestionsHeaderFooterProps, SuggestionsHeaderFooterItem } from '@fluentui/react/lib/FloatingPicker';
```
nebhatna, line 10: I would not depend on FloatingPicker component, let's create the interfaces in FloatingSuggestionsComposite component, since eventually we would want FloatingPicker component to go away ( since it is inheritance based and we intend to not use it)

packages/react-experiments/src/components/FloatingSuggestionsComposite/FloatingSuggestions.types.ts
```diff
@@ -7,6 +7,7 @@ import {
   IFloatingSuggestionItem,
 } from './FloatingSuggestionsItem/FloatingSuggestionsItem.types';
 import { IRenderFunction, IRefObject } from '@uifabric/utilities';
+import { ISuggestionsControlProps } from '../FloatingSuggestions/Suggestions/Suggestions.types';
```
nebhatna, line 10: I would just create this interface here rather than depending on floatingsuggestions component

packages/react-experiments/src/components/FloatingSuggestionsComposite/FloatingSuggestionsList/FloatingSuggestionsList.tsx
```diff
@@ -20,7 +21,44 @@ export const FloatingSuggestionsList = <T extends {}>(props: IFloatingSuggestion
...
+                  className={headerItemProps.className}
+                />
+              </div>
+            ) : null;
+          })}
+        </div>
+      );
+    }
 
     if (onRenderHeader) {
```
nebhatna, line 63: should headerItemsProps and onRenderHeader checks be in an if-else if, the reason could be that we want to only show onRenderHeader when we dont have headerItemProps. If the consumer decided to provide both, in this case, i think onRenderHeader would take precendence right?

packages/react-experiments/src/components/FloatingSuggestionsComposite/FloatingSuggestionsList/FloatingSuggestionsList.tsx
```diff
@@ -29,7 +67,39 @@ export const FloatingSuggestionsList = <T extends {}>(props: IFloatingSuggestion
...
+                  className={footerItemProps.className}
+                />
+              </div>
+            ) : null;
+          })}
+        </div>
+      );
+    }
+
     if (onRenderFooter) {
```
nebhatna, line 103: same question here for the footer props check. OnRenderFooter will take precedence, if the user decided to provide both footerItemsProps and onRenderFooter

packages/react-experiments/src/components/UnifiedPicker/hooks/useFloatingSuggestionItems.ts
```diff
@@ -1,10 +1,19 @@
 import * as React from 'react';
+import { ISuggestionsHeaderFooterProps } from '@fluentui/react/lib/FloatingPicker';
 
 export interface IUseFloatingSuggestionItems<T> {
   focusItemIndex: number;
   setFocusItemIndex: React.Dispatch<React.SetStateAction<number>>;
   suggestionItems: T[];
   setSuggestionItems: React.Dispatch<React.SetStateAction<T[]>>;
+  footerItemIndex: number;
+  setfooterItemIndex: React.Dispatch<React.SetStateAction<number>>;
```
nebhatna, line 10: why do we need it to be of type  React.Dispatch<React.SetStateAction<number>> ?

packages/react-experiments/src/components/UnifiedPicker/hooks/useFloatingSuggestionItems.ts
```diff
@@ -1,10 +1,19 @@
 import * as React from 'react';
+import { ISuggestionsHeaderFooterProps } from '@fluentui/react/lib/FloatingPicker';
```
nebhatna, line 2: same comment here about not importing from FloatingPicker

nebhatna (changes requested):
Comments

packages/react-experiments/src/components/UnifiedPicker/hooks/useFloatingSuggestionItems.ts
elisabethcvs: absolutely no idea, I just copied the setFocusItemIndex type

nebhatna: no worries.. i will research more on this but i do its usage in the codebase :)

packages/react-experiments/src/components/FloatingSuggestionsComposite/FloatingSuggestionsList/FloatingSuggestionsList.tsx
elisabethcvs: headerItemProps would take precedence as it's first and there's a return. I added a note about how the header or footer props win over onRenderFooter/Header somewhere

nebhatna (approved):

PR #15433, opened 2020-10-08
Button: updating styling hooks for derivative Buttons to abstract depending on button styling

dzearing:
This PR makes `useToggleButtonClasses` call `useButtonClasses`. That way partners don't need to implicitly know to call the base hook.

Same applies to `useMenuButtonClasses` and `useCompoundButtonClasses`.


Ridvanovskyy (approved):

change/office-ui-fabric-react-2020-10-06-08-02-45-feat-variants-cherrypick.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
xugao, line 2: unexpected change file. remove

xugao: most of the change files look wrong; could you remove all and re-generate with accurate comment

xugao (commented):
can we also remove `withThemeProvider` usage in all the button tests?

change/office-ui-fabric-react-2020-10-06-08-02-45-feat-variants-cherrypick.json
dzearing: will do, good catch.

xugao (approved):

PR #15436, opened 2020-10-08
Popup: Replacing version in react-internal with version from react-next

khmakoto:
#### Pull request checklist

#### Description of changes

1. Replacing the version of `Popup` in `react-internal` with the one existing in `react-next` and removing the `react-next` version.
2. Updating the top level `Popup` file in `react-next` to reference `@fluentui/react/lib/Popup`.
3. Updating `Panel` API interface to account for changes in the `Popup` API interface.

#### Focus areas to test

(optional)


packages/react-internal/src/components/Panel/Panel.types.ts
```diff
@@ -98,7 +98,7 @@ export interface IPanelProps extends React.HTMLAttributes<PanelBase> {
    * A callback function for when the panel is closed, before the animation completes.
    * If the panel should NOT be dismissed based on some keyboard event, then simply call ev.preventDefault() on it
    */
-  onDismiss?: (ev?: React.SyntheticEvent<HTMLElement>) => void;
+  onDismiss?: (ev?: React.SyntheticEvent<HTMLElement> | KeyboardEvent) => void;
```
ecraig12345, line 101: Not new to this PR I know, but do we know why this got changed? In most other cases we use the React event typing only even though IIRC that's not always 100% correct.

khmakoto: Not really sure, I could trace down when it was changed and see if they provide any reason in that PR.

khmakoto: So, @MLoughry made the change in #12926 but I don't see a reason to do it as far as I can tell.

@MLoughry do you happen to recall why you added `KeyboardEvent` to `onDismiss` in `Popup.types.ts`?

ecraig12345: I think he's on leave this month. :( Looking at the code, I'm wondering if it's because `onDismiss` can potentially be called from a global `onKeyDown` handler, which receives a native event not a react event. So technically the old typings are incorrect, I'm just a bit worried about the migration implications of this change (we should at least mention it in the release notes).

ecraig12345 (approved):
Thanks!

packages/react-internal/src/components/Panel/Panel.types.ts
ecraig12345: I updated the release notes and am going to merge since the PR was ready otherwise.

khmakoto: Thanks for updating the release notes!

PR #15440, opened 2020-10-09
Add Avatar to the PR deploy site

behowell:
#### Pull request checklist

#### Description of changes

Add Avatar to the PR deploy site


packages/react-avatar/config/pre-copy.json
```diff
@@ -1,5 +1,9 @@
 {
   "copyTo": {
-    "dist": []
+    "dist": [
```
ecraig12345, line 3: I'm pretty sure you can skip this part since it's using a storybook site

packages/react-avatar/webpack.serve.config.js
```diff
@@ -0,0 +1,14 @@
+const resources = require('@uifabric/build/webpack/webpack-resources');
```
ecraig12345, line 1: You should be able to skip this file too, it's only for legacy demo apps

ecraig12345 (commented):
Mostly looks good! But I think you can skip the changes in react-avatar itself since those should only be needed for legacy demo sites (not storybook sites).

ecraig12345 (approved):

khmakoto (approved):

PR #15442, opened 2020-10-09
fix(Tree): onFocusParent not working on leaf node

YuanboXue-Amber:
#### Pull request checklist

#### Description of changes

- Description
Before this change, customised `onFocusParent` on a leaf node won't work. 

For example, our virtualizedTree prototype using customised `onFocusParent`, to make sure parent node can be focused on by `ArrowLeft` key, even when parent node is scrolled out of rendering window. However it does not work:
old
(I'm pressing `ArrowLeft` on the leaf node Tree-Item-0-21 and it does nothing)

- Cause: 
When `ArrowLeft` is pressed, `keydown` event does bubble up from treeTitle to treeItem. However when treeItem has no subtree, it has no listener for this event. 
- Fix: 
Add `onFocusParent` on treeItem, when it has no subtree. And `onFocusParent` listens to ArrowLeft event

- Fixed Behavior:
our virtualizedTree prototype:
new

#### Focus areas to test

(optional)

### DIFF
--- a/packages/fluentui/e2e/tests/treeKeyboardNavigation-test.ts
+++ b/packages/fluentui/e2e/tests/treeKeyboardNavigation-test.ts
@@ -38,7 +38,11 @@ describe('Tree keyboard navigation', () => {
     await e2e.isFocused(selectors.treeItemAt(2));
     await e2e.expectCount(selectors.treeItem, 4);
 
-    await e2e.waitForSelectorAndPressKey(selectors.treeItemAt(2), 'ArrowLeft'); // Focus parent 1nd level and closes 3rd level
+    await e2e.waitForSelectorAndPressKey(selectors.treeItemAt(2), 'ArrowLeft'); // closes 3rd level
+    await e2e.expectCount(selectors.treeItem, 3);
+    await e2e.isFocused(selectors.treeItemAt(2));
+
+    await e2e.waitForSelectorAndPressKey(selectors.treeItemAt(2), 'ArrowLeft'); // Focus parent 1nd level
     await e2e.expectCount(selectors.treeItem, 3);
     await e2e.isFocused(selectors.treeItemAt(1));
   });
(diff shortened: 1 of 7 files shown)


assuncaocharles:
I wonder if it can be a breaking change. The behavior we had previously in the Tree is that pressing `ArrowLeft` in a node with subtree it would collapse its subtree and focus parent and now we would just collapse it.

YuanboXue-Amber:
> I wonder if it can be a breaking change. The behavior we had previously in the Tree is that pressing `ArrowLeft` in a node with subtree it would collapse its subtree and focus parent and now we would just collapse it.

@pompomon @layershifter could you confirm if this is okay?

jurokapsiar (approved):

packages/fluentui/e2e/tests/treeKeyboardNavigation-test.ts
```diff
@@ -38,7 +38,11 @@ describe('Tree keyboard navigation', () => {
     await e2e.isFocused(selectors.treeItemAt(2));
     await e2e.expectCount(selectors.treeItem, 4);
 
-    await e2e.waitForSelectorAndPressKey(selectors.treeItemAt(2), 'ArrowLeft'); // Focus parent 1nd level and closes 3rd level
+    await e2e.waitForSelectorAndPressKey(selectors.treeItemAt(2), 'ArrowLeft'); // closes 3rd level
+    await e2e.expectCount(selectors.treeItem, 3);
+    await e2e.isFocused(selectors.treeItemAt(2));
+
+    await e2e.waitForSelectorAndPressKey(selectors.treeItemAt(2), 'ArrowLeft'); // Focus parent 1nd level
```
assuncaocharles, line 45:
```suggestion
    await e2e.waitForSelectorAndPressKey(selectors.treeItemAt(2), 'ArrowLeft'); // Focus parent 1st level
```

jurokapsiar:
> > I wonder if it can be a breaking change. The behavior we had previously in the Tree is that pressing `ArrowLeft` in a node with subtree it would collapse its subtree and focus parent and now we would just collapse it.
> 
> @pompomon @layershifter could you confirm if this is okay?

we should do the same as https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html - if I understand correctly, this PR makes Tree behave the same way. This is not a breaking change as we do not change API.

Ridvanovskyy (approved):

PR #14912, opened 2020-09-04
Vertical stacked bar chart combined callout and other tweaks

mbest:
#### Pull request checklist

- [x] Include a change request file using `$ yarn change`

#### Description of changes

VerticalStackedBarChart

- props for custom rendering the callout and choosing between combined and per-point callouts
- props for customizing the style or position of the callout
- skip outputting bars for zero-value points
- remove ref array where indexes were not deterministic

## Screenshots

Customize position and display (formatting value)

image

Custom callout for whole stack

image

Design

image

PR #15447, opened 2020-10-09
Charting: Vertical stacked bar chart minor changes including uncaught type error issue fix.

jameelakowsar:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #https://github.com/microsoft/fluentui/issues/15465

#### Description of changes

Vertical stacked bar chart.

1. 'Uncaught type error' issue fixed when click event triggers on bars.
2. Adjusted margin props to the cartesian/specific charts.
3. Updated below conditions as like previous PR:  https://github.com/microsoft/fluentui/pull/14912/
     -  Non zero bars condition added
     - Changed 'data-is-focusable' param to undefined instead of false.
    - Ref's data updated.
4. Margins condition updated. Now 0 value can be accepted.
5. Added description(default prop values) to the margins prop

Note: As Cartesian component consumed by other charts, Due to small change in cartesian, cross checked all other charts and added screenshots.

#### Focus areas to test

Vertical stacked bar chart
Vertical bar chart
Area chart
Line chart

### After fix
 ```
margins: {
left: 45,
right: 30
}
```
#### Vertical stacked bar chart
image

#### Vertical stacked bar chart with RTL
image

```
margins: {{
left: 45,
right: 30
}}
```
#### Vertical bar chart 
image

#### Vertical bar chart with RTL
image

#### Area chart
image

#### Area chart with RTL
image

#### Line chart
image

#### Line chart with RTL
image

### DIFF
--- a/packages/charting/src/components/CommonComponents/CartesianChart.base.tsx
+++ b/packages/charting/src/components/CommonComponents/CartesianChart.base.tsx
@@ -15,7 +15,6 @@ import {
   createDateXAxis,
   createYAxis,
   createStringYAxis,
-  additionalMarginRight,
   IMargins,
   getMinMaxOfYAxis,
   XAxisTypes,
@@ -200,7 +199,7 @@ export class CartesianChartBase extends React.Component<IModifiedCartesianChartP
                 this.xAxisElement = e;
               }}
               id={`xAxisGElement${this.idForGraph}`}
-              transform={`translate(0, ${svgDimensions.height - 35})`}
+              transform={`translate(0, ${svgDimensions.height - this.margins.bottom!})`}
               className={this._classNames.xAxis}
             />
             <g
@@ -208,9 +207,8 @@ export class CartesianChartBase extends React.Component<IModifiedCartesianChartP
                 this.yAxisElement = e;
               }}
               id={`yAxisGElement${this.idForGraph}`}
-              transform={`translate(${
-                this._isRtl ? svgDimensions.width - this.margins.right! - additionalMarginRight : 40
-              }, 0)`}
+              // Removing margins.right 2 times from width for RTL.
+              transform={`translate(${this._isRtl ? svgDimensions.width - 2 * this.margins.right! : 40}, 0)`}
               className={this._classNames.yAxis}
             />
             {children}
--- a/packages/charting/src/components/CommonComponents/CartesianChart.base.tsx
+++ b/packages/charting/src/components/CommonComponents/CartesianChart.base.tsx
@@ -59,11 +59,17 @@ export class CartesianChartBase extends React.Component<IModifiedCartesianChartP
       _height: this.props.height || 350,
     };
     this.idForGraph = getId('chart_');
+    /**
+     * In RTL mode, Only graph will be rendered left/right. We need to provide left and right margins manually.
+     * So that, in RTL, left margins becomes right margins and viceversa.
+     * As graph needs to be drawn perfecty, these values consider as default values.
+     * Same margins using for all other cartesian charts. Can be accessible through 'getMargins' call back method.
+     */
     this.margins = {
       top: this.props.margins?.top || 20,
-      right: this.props.margins?.right || 20,
       bottom: this.props.margins?.bottom || 35,
-      left: this.props.margins?.left || 40,
+      right: this._isRtl ? this.props.margins?.left || 40 : this.props.margins?.right || 20,
+      left: this._isRtl ? this.props.margins?.right || 20 : this.props.margins?.left || 40,
     };
   }
 
@@ -207,8 +213,9 @@ export class CartesianChartBase extends React.Component<IModifiedCartesianChartP
                 this.yAxisElement = e;
               }}
               id={`yAxisGElement${this.idForGraph}`}
-              // Removing margins.right 2 times from width for RTL.
-              transform={`translate(${this._isRtl ? svgDimensions.width - 2 * this.margins.right! : 40}, 0)`}
+              transform={`translate(${
+                this._isRtl ? svgDimensions.width - this.margins.right! : this.margins.left!
+              }, 0)`}
               className={this._classNames.yAxis}
             />
             {children}
--- a/packages/charting/src/components/CommonComponents/CartesianChart.base.tsx
+++ b/packages/charting/src/components/CommonComponents/CartesianChart.base.tsx
@@ -66,10 +66,10 @@ export class CartesianChartBase extends React.Component<IModifiedCartesianChartP
      * Same margins using for all other cartesian charts. Can be accessible through 'getMargins' call back method.
      */
     this.margins = {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by Raghurk — APPROVED** [review]

**Comment by mbest** [comment]
@dzearing  Why is the "Charting" label no longer applied automatically?

**Comment by mbest** [comment]
@jameelakowsar I will test these changes against our code and let know what works or still doesn't.

**Comment by dzearing** [comment]
@ecraig12345 I think the answer to @mbest question above is that the bot is perhaps looking only at target=master PRs? But i'm not sure.

**Comment by ecraig12345** [comment]
Yeah it looks like the bot only considers master PRs for path-based labeling, and I don't see an option to consider other branches unfortunately.

**Comment by ecraig12345** [comment]
Oh actually, I take that back. The way that the configuration works for this feature is a bit weird, and I think we need to update it to  ignore changes to `react-examples` when considering what labels to use. I'll do that now.

**Review by Ridvanovskyy — APPROVED** [review]

**Comment by mbest** [comment]
I've verified that this fixes the following regressions: 

* script error from _redirectToUrl: Uncaught TypeError: Cannot read property 'props' of undefined
* x-axis position ignores bottom margin (it's only in the right place when margin is 35px)

There are still these issues:

* zero-value data isn't skipped
* a zero right margin is ignored (defaults to 20px)
* left margin of x axis is always 40px (ignores provided margin; the length of the axis lines is correct though, just the position is wrong)

**Comment by mbest** [comment]
image

As shown above, the translate value should match the left margin, but is always 40 instead.

**Comment by mbest** [comment]
In RTL mode, the right margin isn't handled correctly. Comparing the `svg` width to the translate value, they should be different by the amount of the right margin (36px in this case), but instead they are 72px different.

image

Here's the same chart before the changes in #15061 
 
image

**Comment by jameelakowsar** [comment]
> I've verified that this fixes the following regressions:
> 
> * script error from _redirectToUrl: Uncaught TypeError: Cannot read property 'props' of undefined
> * x-axis position ignores bottom margin (it's only in the right place when margin is 35px)
> 
> There are still these issues:
> 
> * zero-value data isn't skipped
> * a zero right margin is ignored (defaults to 20px)
> * left margin of x axis is always 40px (ignores provided margin; the length of the axis lines is correct though, just the position is wrong)

What meant by 'zero-value' data?
For remaining, Please check email once.
Test link provided in PR description.

**Comment by jameelakowsar** [comment]
@mbest ,

If all good, I will make this PR to open. Confirm your status.

File: packages/charting/src/components/CommonComponents/CartesianChart.types.ts [hunk 1]
```diff
@@ -157,6 +157,12 @@ export interface ICartesianChartProps {
 
   /**
    * Margins for the chart
+   * @default top: 20
+   * @default bottom: 35
+   * @default left: 40
+   * @default right: 20
```
ecraig12345, line 163: You should only provide a single value for `@default`, and use backtick syntax to wrap the object to prevent some parsers from complaining.
```suggestion
   * @default `{ top: 20, bottom: 35, left: 40, right: 20 }`
```

packages/react-examples/src/charting/VerticalStackedBarChart/VerticalStackedBarChart.Styled.Example.tsx
```diff
@@ -118,8 +128,8 @@ export class VerticalStackedBarChartStyledExample extends React.Component<{}, IV
                 },
               },
             }}
-            // eslint-disable-next-line react/jsx-no-bind
-            onRenderCalloutPerDataPoint={props =>
+            // eslint-disable-next-line react/jsx-no-bind @typescript-eslint/no-explicit-any
```
ecraig12345, line 131: I think for disabling multiple rules, eslint expects you to use a comma between them.
```suggestion
            // eslint-disable-next-line react/jsx-no-bind, @typescript-eslint/no-explicit-any
```

packages/charting/src/components/CommonComponents/CartesianChart.base.tsx
```diff
@@ -59,11 +59,17 @@ export class CartesianChartBase extends React.Component<IModifiedCartesianChartP
...
+     * As graph needs to be drawn perfecty, these values consider as default values.
+     * Same margins using for all other cartesian charts. Can be accessible through 'getMargins' call back method.
+     */
     this.margins = {
       top: this.props.margins?.top || 20,
-      right: this.props.margins?.right || 20,
       bottom: this.props.margins?.bottom || 35,
-      left: this.props.margins?.left || 40,
+      right: this._isRtl ? this.props.margins?.left || 40 : this.props.margins?.right || 20,
+      left: this._isRtl ? this.props.margins?.right || 20 : this.props.margins?.left || 40,
```
mbest, line 72: Use `??` to allow specifying a `0` margin (also do this for top and bottom):

```suggestion
      right: this._isRtl ? this.props.margins?.left ?? 40 : this.props.margins?.right ?? 20,
      left: this._isRtl ? this.props.margins?.right ?? 20 : this.props.margins?.left ?? 40,
```

mbest (changes requested):

mbest:
> What meant by 'zero-value' data?

Before your "refactor" I had added this line to filter our zero-value data:

```
    const nonZeroBars = singleChartData.chartData.filter(point => point.data > 0);
```

I see from looking through the `_createBar` function that you undid most of the improvements I made, including that one. Additionally, it's back to using `refArrayIndexNumber` which is broken if the stacks don't all have the same number of data items. You also undid some code cleanup I did. For example I had:

```
    const xPoint = xBarScale(isNumeric ? (singleChartData.xAxisPoint as number) : indexNumber);
```

which you changed back to (and moved back into the loop even though it doesn't need to be there):

```
        let xPoint;
        if (this._isNumeric) {
          xPoint = xBarScale(singleChartData.xAxisPoint as number);
        } else {
          xPoint = xBarScale(indexNumber);
        }
```

You also undid the normalization of the props that are passed to the callout handlers.

You also undid the code that makes sure that `data-is-focusable` isn't set to a value of `false`, which breaks focus issues.

packages/charting/src/components/CommonComponents/CartesianChart.base.tsx
jameelakowsar: Changes done

packages/charting/src/components/CommonComponents/CartesianChart.types.ts
jameelakowsar: Done, 
Thank you!!!

packages/react-examples/src/charting/VerticalStackedBarChart/VerticalStackedBarChart.Styled.Example.tsx
jameelakowsar: Done,
Thank you!!!

PR #15450, opened 2020-10-09
SearchBox: role=search is removed, but can be optionally defined via role prop.

### LINKED ISSUE #14914: SearchBox role is hardcoded and can not be changed.
### Environment Information

- **Package version(s)**: 7.128.1

### Describe the issue:

In some cases search box is a part of a larger region with a role 'search'. In this case role of the wrapper for a SearchBox need to be change to 'presentation'. However, setting role attribute on a SearchBox control has no effect, it remains "search" for a wrapper and "searchbox" for input.

Reason is that these roles are hard coded and can not be changed. This is the code from SearchBox.base.tsx

      <div **role="search**" ref={this._rootElement} className={classNames.root} onFocusCapture={this._onFocusCapture}>
        <div className={classNames.iconContainer} onClick={this._onClickFocus} aria-hidden={true}>
          <Icon iconName="Search" {...iconProps} className={classNames.icon} />
        </div>
        <input
          {...nativeProps}
          id={id}
          className={classNames.field}
          placeholder={placeholderValue}
          onChange={this._onInputChange}
          onInput={this._onInputChange}
          onBlur={this._onBlur}
          onKeyDown={this._onKeyDown}
          value={value}
          disabled={disabled}
          **role="searchbox"**
          aria-label={ariaLabel}
          ref={this._inputElement}
        />

At least it should be possible to change the role of the outer element.

#### Actual behavior:

Generated HTML for a SearchBox wrapper has a hard coded role of 'searchbox'.

#### Expected behavior:

Generated HTML for a SearchBox wrapper has a role attribute set as specified, e.g. "presentation".

Yes.

Normal.

Word WAC.

dzearing:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14914

## Before:

This JSX:
```
<SearchBox />
```
Renders this HTML: 
```jsx
<div role="search">
  <input ... />
</div>
```

## After:

This JSX:
```
<SearchBox />
<SearchBox role="search" />
```
Renders this HTML: 
```jsx
<div>
  <input ... />
</div>

<div role="search">
  <input ... />
</div>
```


ecraig12345:
I think you have the wrong base branch.

ecraig12345:
I mean it looks like your local branch for this PR was based off of master. It should be based off of 7.0.

dzearing:
I think it's good now?

khmakoto (approved):
Seems like you are missing snapshot updates in `@fluentui/react-examples`.

Ridvanovskyy (approved):

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
```diff
@@ -76,6 +76,7 @@ export class SearchBoxBase extends React.Component<ISearchBoxProps, ISearchBoxSt
       clearButtonProps,
       disableAnimation,
       iconProps,
+      role,
```
ecraig12345, line 79: Can you explicitly add this in ISearchBoxProps too?

Ridvanovskyy (approved):

packages/office-ui-fabric-react/src/components/SearchBox/SearchBox.base.tsx
dzearing: should be fixed now.

packages/react-next/src/components/SearchBox/SearchBox.types.ts
```diff
@@ -108,6 +108,11 @@ export interface ISearchBoxProps
    */
   underlined?: boolean;
 
+  /**
+   * The role assigned to the root DIV element of the SearchBox, useful for defining a landmark role, such as "search".
+   */
+  role?: string;
```
ecraig12345, line 114: This should be in OUFR not react-next

dzearing: whoops, missed this. Fixing now!

PR #15454, opened 2020-10-09
Rename theme packages

ecraig12345:
#### Pull request checklist

#### Description of changes

Move `azure-themes` and `theme-samples` to `@fluentui` scope (same names).

Plus some fixes to the rename script.

### DIFF
--- a/packages/azure-themes/webpack.config.js
+++ b/packages/azure-themes/webpack.config.js
@@ -11,7 +11,7 @@ module.exports = resources.createConfig(BUNDLE_NAME, IS_PRODUCTION, {
 
   output: {
     libraryTarget: 'var',
-    library: 'FabricAzureThemes',
+    library: 'FluentUIAzureThemes',
   },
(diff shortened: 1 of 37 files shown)


packages/azure-themes/webpack.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = resources.createConfig(BUNDLE_NAME, IS_PRODUCTION, {
 
   output: {
     libraryTarget: 'var',
-    library: 'FabricAzureThemes',
+    library: 'FluentUIAzureThemes',
```
xugao, line 14: this makes me realize, we should add/note this change in our release notes

xugao (approved):

packages/azure-themes/webpack.config.js
ecraig12345: Yeah I might update that with all the new package names at the end

PR #15455, opened 2020-10-10
Callout: Move to react-next

joschect:
#### Pull request checklist

#### Description of changes

Move callout v8 changes to react-internal, including function component conversion from #14181.


ecraig12345:
Looks like you're missing some change files, and something weird happened with the tests. Possibly the issue is that tests of other components which use Callout will need stuff wrapped in act()?

ecraig12345:
Might make sense to combine this PR into the ContextualMenu one--it looked like quite a few of the failing tests were in ContextualMenu, which might be fixed by the act() changes in the other PR. (Or just merge that one first?)

joschect:
I'm going to just wait till the contextualmenu changes are in since this could impac t them

ecraig12345:
I fixed the merge conflicts but it looks like there may be a bunch of tests that need stuff wrapped in `act`.

joschect:
@ecraig12345 It looks like the merge fixed the various act breaks, I think that this is ready to go in.

ecraig12345:
Oh interesting, it didn't fix it for me locally but maybe that was because I had a partially stale build?

ecraig12345 (approved):
Yay!

PR #15414, opened 2020-10-07
Dependency, bundling, and PR deploy fixes

ecraig12345:
Dependency fixes:
- Fix some security alerts by updating dependencies.
- Remove `react-text` dependency on `@fluentui/storybook` which was probably causing it to build more packages than necessary.
- Remove unneeded dependencies on `example-app-base` and `prop-types`.

Removing the deps on `example-app-base` caused bundling the demo apps for PR deploy to stop working. The issue turned out to just be that we needed to be calculating resolve aliases for *legacy* demo apps relative to `react-examples`, like we were already doing with storybook demos, but there were a lot of red herrings and rabbit holes on the path to realizing that. Resulting in this overall good (but more time consuming) list of fixes:
- Add more helpers to de-duplicate the webpack configs of component packages: `createBundleConfig` and `createLegacyDemoAppConfig`. (This made it easy to do the fix for the actual issue in one place.)
- Move build output of the demo apps (both legacy and storybook) into sub-directories of `dist`: `dist/storybook` and `dist/demo`
- Stop publishing demo apps and react in our npm packages!! (This was happening in any package with a legacy demo app.) Putting the demo apps under `dist/demo` made this much easier.

Also PR deploy site improvements:
- Sort the packages in the PR deploy site
- Make the message in the header work for branch CI builds, not just PR builds
- Add instructions for how to add new things to the PR deploy site

Other cleanup/improvements:
- Delete `react-spinbutton` for now since we probably won't be using it in the immediate future (can re-create with current template when needed).
- Delete webpack and copy configs which aren't needed
- Remove react-cards legacy demo app which isn't really needed
- Delete empty or incorrectly copy-pasted doc sections for secondary packages (charting, experiments)
- Remove old version of apps/perf-test which isn't being used (the flamegrill-based part is still there)
- Remove the todo app from the PR deploy site since it's not very useful to have there
- Update azure pipelines to use more shared variables

PR #15456, opened 2020-10-10
Remove unneeded package dependencies

xugao:
#### Pull request checklist

#### Description of changes

- Remove unneeded package dependencies
- Add missing set-version for packages.
- Enable ESLint for `date-time-utilites`

#### Focus areas to test

(optional)


apps/codesandbox-react-northstar-template/package.json
```diff
@@ -18,9 +18,7 @@
     "typescript": "3.7.2"
   },
   "dependencies": {
-    "@fluentui/accessibility": "^0.51.0",
```
ecraig12345, line 21: These ones might be needed to get it to substitute the PR versions?

packages/date-time-utilities/package.json
```diff
@@ -24,7 +24,6 @@
     "update-api": "just-scripts update-api"
   },
   "devDependencies": {
-    "@fluentui/eslint-plugin": "^0.54.1",
```
ecraig12345, line 27: This needs to stay too--the fact that this package doesn't have eslintrc is wrong.

packages/fluentui/perf/package.json
```diff
@@ -5,13 +5,11 @@
   "license": "MIT",
   "dependencies": {
     "@babel/polyfill": "^7.10.4",
-    "@fluentui/docs": "^0.51.0",
```
ecraig12345, line 8: This is used in this package.

packages/react-next/package.json
```diff
@@ -35,7 +35,6 @@
   },
   "devDependencies": {
     "@fluentui/common-styles": "^0.2.6",
-    "@fluentui/eslint-plugin": "^0.54.1",
```
ecraig12345, line 38: This one might need to stay even though it's not referenced directly (not sure)

packages/react-spinbutton/package.json
```diff
@@ -26,7 +26,6 @@
   },
   "devDependencies": {
```
ecraig12345, line 27: I thought I deleted this package? Or maybe that PR is still open.

ecraig12345:
Build is failing while running storybook webpack in `date-time` (and probably other packages) with errors like `Module not found: Error: Can't resolve '@fluentui/react' in '/mnt/work/3/s/packages/react-examples/src/date-time/Calendar'`

I hit this in #15414 too, and I'm guessing the issue is that with the `example-app-base` dep removed, it also implicitly prevented `@fluentui/react` from getting added to the resolve alias. I added a console log of the generated aliases in that PR to help debug but haven't gotten to look back at it.

ecraig12345 (approved):
Thanks for doing this! Looks like it still needs a fix for the bundling issue though.

packages/date-time-utilities/src/dateGrid/getDayGrid.ts
```diff
@@ -56,16 +56,14 @@ export const getDayGrid = (options: IDayGridOptions): IDay[][] => {
...
-    selectedDates = getDateRangeArray(
-      selectedDate,
-      selectedDateRangeType,
-      firstDayOfWeek,
-      workWeekDays,
-      daysToSelectInDayView,
-    );
-    selectedDates = getBoundedDateRange(selectedDates, minDate, maxDate);
-  }
+  selectedDates = getDateRangeArray(
```
ecraig12345, line 59: Given the [northstar screener failures](https://screener.io/v2/states/H17FwCf88.microsoft-fluentui-fluentui/xgao%2Fremove-unused-deps/1024x768/Chrome/datepickercellexampleshorthandtsx-shows-datepicker-popup-0) I suspect the typings here are wrong (relative to usage), and the `if (selectedDate)` check is actually needed. Maybe best to add back the if() and update the typings to allow undefined?

xugao: my bad, pushed fix!

ecraig12345 (approved):

PR #15461, opened 2020-10-11
Virtualized Tree with Sticky header Prototype

YuanboXue-Amber:
#### Pull request checklist

#### Description of changes

Created a new prototype of tree: tree rendered with `react-window`, and having the 1st level tree item as sticky headers (i.e. always visible). 

Behavior:
Sticky headers are alway visible.
There's always one 'active' sticky header, that sticks to the top, under all its collapsed left siblings. All its right siblings will be sticked to the bottom and collapsed. Unless there's extra space in the container. In which case, pull up the 'active' sticky header's  right siblings and expand them, one by one, until the extra space is filled.
 
ezgif com-gif-maker (2)

Supports: 
- different height for sticky headers and non-sticky tree items 
- keyboard navigation:
  - left/right arrow key navigates to parent/first child
  - `*` key expands siblings (only works on the 2nd level tree items, in a 3-level tree)
 
Constraints: 
- Only the 1st level tree items are, and has to be sticky
- The tree can be maximum 3 levels.

#### Focus areas to test

(optional)


packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/itemsGenerator.ts
```diff
@@ -0,0 +1,24 @@
+import * as _ from 'lodash';
+
+function getItems(minItems = 5, maxItems = 15, maxLevel = 2, maxSticky = 10) {
```
Ridvanovskyy, line 3: Propose add types in this file. 
 @YuanboXue-Amber , I apologize in advance for such a question, I have not found the answer anywhere, is the `fluentui/packages/fluentui/docs/src/prototypes/` a folder of experimental components?

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/itemsGenerator.ts
```diff
@@ -0,0 +1,24 @@
...
+  function getItemsNumber(min, max) {
```
Ridvanovskyy, line 4: Propose add typings.

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,465 @@
...
+      style={{
+        ...style,
+        height: Math.max(height, style.height),
+      }}
+    >
+      {stickyTopItemsIndex.map((index: number) => {
+        const item = renderedItems[index];
+        return React.cloneElement(item, {
+          style: {
+            position: 'sticky',
```
Ridvanovskyy, line 68: The component uses `position: sticky;` but it is [not supported](https://caniuse.com/css-sticky) in [browsers that we need](https://github.com/microsoft/fluentui/wiki/Browser-Support).
Сan you please explain, is **React Northstar** an experimental and is it acceptable here? Or am I missing something? As i understand **React Northstar** version must support [these browsers too].(https://github.com/microsoft/fluentui/wiki/Browser-Support).
Thanks in advance for your reply!

YuanboXue-Amber: Hi Vlad, so this is a prototype, a showcase for the consumers of React Northstar. The team that is requesting this prototype is not expecting IE 11 support for it (in which case, they could fall back to classic tree).
As for other browsers (edge, firefox, chrome), I tested it and it works well. Let me know if you find anything that looks like a bug

packages/fluentui/docs/package.json
```diff
@@ -44,7 +44,9 @@
     "react-source-render": "4.0.0-1",
     "react-textarea-autosize": "7.0.0",
     "react-virtualized": "^9.21.1",
-    "react-vis": "^1.11.6"
+    "react-vis": "^1.11.6",
+    "react-window": "^1.8.5",
+    "what-input": "^5.2.10"
```
jurokapsiar, line 49: if I remember correctly, we ship our own implementation of what-input, can you use that instead?

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,468 @@
...
+  const ref = React.useRef(null); // ref for entire list
+
+  // When using keyboard, and navigate to non-sticky items, they could be hidden behind sticky headers.
+  // Scroll to make the focused non-sticky items always visible
+  const handleOnFocus = React.useCallback(
+    (e: React.SyntheticEvent<HTMLElement>) => {
+      if (whatInput.ask() !== 'keyboard') return;
+
+      let itemRef = e.target as any;
+      if (itemRef.tagName !== 'DIV') itemRef = itemRef.parentNode; // when treeTitle focused, get its outer treeItem
```
jurokapsiar, line 245: can you check for `treeTitleClassName` class presence instead of the tag name? tag names can be customized.

YuanboXue-Amber: You're right checking tagname is not good. I pushed a new change with this check:
```ts
if (itemRef.className.includes('ui-tree__title'))
```

packages/fluentui/docs/package.json
YuanboXue-Amber: What-input is removed based on the discussion

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/itemsGenerator.ts
YuanboXue-Amber: Thank you for the review, the typings are add in this [commit](https://github.com/microsoft/fluentui/pull/15461/commits/69e840d0ab7ed4db265d4d288b67dcb06aad0b6e)

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
Ridvanovskyy: @YuanboXue-Amber  Thanks for explaining understood!

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,465 @@
...
+
+    return null;
+  }
+}
+
+const InnerElementType = ({ children, style }) => {
+  const context = React.useContext(InnerElementContext);
+  const { height, stickyItemSize, stickyTopIds, stickyBottomIds, onClickSticky } = context;
+
+  if (!children.length) return <></>;
```
ling1726, line 44: nit: why are you returning a fragment here ? I think the recommended return for conditional render is `null`. Also `children` can be null/undefined 

```suggestion
  if (!children?.length) return null;
```

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,465 @@
...
+
+  const data = children[0]?.props?.data;
```
ling1726, line 46: You shouldn't need to validate `children[0]` since you've already done it in the above guard

```suggestion
  const data = children[0].props?.data;
```

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,465 @@
...
+        height: Math.max(height, style.height),
+      }}
+    >
+      {stickyTopItemsIndex.map((index: number) => {
+        const item = renderedItems[index];
+        return React.cloneElement(item, {
+          style: {
+            position: 'sticky',
+            height: stickyItemSize,
+            zIndex: 2,
```
ling1726, line 69: zIndexes can be quite hard to maintain around the place, should a default be provided in some kind of theme ?

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,465 @@
...
+      })}
+      {children}
+      {stickyBottomItemsIndex.map((itemIndex: number, index: number, arr: number[]) => {
+        const item = renderedItems[itemIndex];
+        return React.cloneElement(item, {
+          style: {
+            position: 'sticky',
+            height: stickyItemSize,
+            zIndex: 111,
+            top: height - (arr.length - index) * stickyItemSize,
```
ling1726, line 85: nit: a little comment about this calculation can be nice, the reason behind these things can get lost in the future

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,465 @@
...
+  itemSize: number;
+  height: number;
+  toplvlItemsId: string[];
+  onActiveItemsIdChange: (idsToRemove?: string[], idsToAdd?: string[]) => void;
+}
+
+interface ItemDataType {
+  renderedItems: React.ReactElement<TreeItemProps>[];
+  isNonSticky: (item: React.ReactElement<TreeItemProps>) => boolean;
+  onClick: (event: React.SyntheticEvent<HTMLElement>, itemProps: TreeItemProps) => void;
```
ling1726, line 116: you use `React.SyntheticEvent` a lot, is it possible to be more precise in the typings ? for example a click event can be `React.MouseEvent`

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,465 @@
...
+  handleOnFocus: (e: React.SyntheticEvent<HTMLElement>) => void;
```
ling1726, line 117: nit: might as well be consistent
```suggestion
  handleOnFocus: (event: React.SyntheticEvent<HTMLElement>) => void;
```

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,465 @@
...
+  }, []);
+
+  // when navigate with keyboard, use `itemIdTobeFocused` to remember the focused item id,
+  // then when DOM updates are finished, sticky tree knows which item to focus on
+  const [itemIdTobeFocused, setItemIdTobeFocused] = React.useState<string>(null);
+  const passItemIdTobeFocused = React.useCallback((newItemIdTobeFocused: string) => {
+    setItemIdTobeFocused(newItemIdTobeFocused);
+  }, []);
+
+  React.useLayoutEffect(() => {
```
ling1726, line 148: You might not need `useLayoutEffect` here, since you're only scrolling to a specific item so you don't need to run this logic before the browser paint.... a simply `useEffect` here should do

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
YuanboXue-Amber: My english skill is challenged :D Came up with:
```ts
// stick items to the container bottom, with the smaller-indexed items stacked on bigger-indexed items
```
in this [commit](https://github.com/microsoft/fluentui/pull/15461/commits/c83f3438a3bfdb3ee6c843c983e9d12f2a4d948a)

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
YuanboXue-Amber: Yes you're right. All my onclick here is listeners to native `onClick`, which is mouseEvent, going to make a commit to update the typings here

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
YuanboXue-Amber: Good point. Thank you for the explanation. I tested with useEffect and it works well. Going to make a commit

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
YuanboXue-Amber: Updated in this [commit](https://github.com/microsoft/fluentui/pull/15461/commits/820436b0c2429ebb8e2abf1f3b67c4b505dfdfc5)

All the `onFocus` and `onClick` events are updated to `React.FocusEvent` and `React.MouseEvent`.

I kept the `React.SyntheticEvent` on `handleFocusParent`,`handleFocusFirstChild` and `handleOnSiblingsExpand`, because:
- to be in consistent with `TreeItem`
- I don't know what's the more precise typing for them anyway :D

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
YuanboXue-Amber: I see your point. I'm not sure how we handle the css variables for prototypes. I will look into it and get back to this comment :)

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,472 @@
...
+    </div>
+  );
+};
+
+const calcRenderedHeight = (renderedItems: React.ReactElement[], stickyItemSize: number, itemSize: number) => {
+  let renderedHeight = 0;
+  renderedItems.forEach(item => {
+    renderedHeight = renderedHeight + (item.props.level === 1 ? stickyItemSize : itemSize);
+  });
+  return renderedHeight;
```
ling1726, line 108:
```suggestion
  return renderedItems.reduce((item, renderedHeight) => {
    return renderedHeight + (item.props.level === 1 ? stickyItemSize : itemSize)
  }, 0);
  let renderedHeight = 0;
  renderedItems.forEach(item => {
    renderedHeight = renderedHeight + (item.props.level === 1 ? stickyItemSize : itemSize);
  });
  return renderedHeight;
```

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,472 @@
...
+  // when navigate with keyboard, use `itemIdTobeFocused` to remember the focused item id,
+  // then when DOM updates are finished, sticky tree knows which item to focus on
+  const [itemIdTobeFocused, setItemIdTobeFocused] = React.useState<string>(null);
+  const passItemIdTobeFocused = React.useCallback((newItemIdTobeFocused: string) => {
+    setItemIdTobeFocused(newItemIdTobeFocused);
+  }, []);
+
+  React.useEffect(() => {
+    // retain focus after sticky items gotten clicked and triggers DOM updates
+    if (itemIdTobeFocused === null) return;
```
ling1726, line 157: use either falsy check or `==` you don't handle `undefined here`

```suggestion
    if (!itemIdTobeFocused) return;
```

```suggestion
    if (itemIdTobeFocused ==  null) return;
```

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+    const toFocusIndex = renderedItems.findIndex(item => item.props.id === itemIdTobeFocused);
+    const toFocusItem = renderedItems[toFocusIndex];
+    const toFocus = (toFocusItem?.props.contentRef as any)?.current;
+    if (toFocus) {
+      if (toFocusItem.props.level !== 1) ref.current.scrollToItem(toFocusIndex, 'center');
```
ling1726, line 163:
```suggestion
    if (toFocus&& toFocusItem.props.level !== 1) ref.current.scrollToItem(toFocusIndex, 'center');
```

packages/fluentui/docs/package.json
```diff
@@ -44,7 +44,8 @@
     "react-source-render": "4.0.0-1",
     "react-textarea-autosize": "7.0.0",
     "react-virtualized": "^9.21.1",
-    "react-vis": "^1.11.6"
+    "react-vis": "^1.11.6",
+    "react-window": "^1.8.5"
```
jurokapsiar, line 48: we can remove this :)

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
jurokapsiar: search for `zIndexes` in packages/fluentui/react-northstar/src/themes/teams/siteVariables.ts, maybe we can use some constants from there

YuanboXue-Amber: Updated in this [commit](https://github.com/microsoft/fluentui/pull/15461/commits/142bcb327b17452d9cb4894a258296214fdd558b)

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
YuanboXue-Amber: I have to revert this. 
`if (toFocus)` actually contains 2 lines. I neglected this as well when I commit suggestion :P

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,471 @@
...
+      {children}
+      {stickyBottomItemsIndex.map((itemIndex: number, index: number, arr: number[]) => {
+        const item = renderedItems[itemIndex];
+        return React.cloneElement(item, {
+          style: {
+            position: 'sticky',
+            height: stickyItemSize,
+            zIndex: teamsTheme.siteVariables.zIndexes.overlayPriority,
+            // stick items to the container bottom, with the smaller-indexed items stacked on bigger-indexed items
+            top: height - (arr.length - index) * stickyItemSize,
```
ling1726, line 93: nit: you could generate this obj in a function and pass in the value for top to avoid duplicating the other styles with the top stickies

packages/fluentui/docs/src/prototypes/VirtualizedStickyTree/VirtualizedStickyTree.tsx
```diff
@@ -0,0 +1,486 @@
...
+    const item = data.renderedItems[index];
+    if (item) {
+      return data.isNonSticky(item)
+        ? React.cloneElement(item, {
+            style,
+            onTitleClick: data.onClick,
+            onFocusParent: data.handleFocusParent,
+            onFocusFirstChild: data.handleFocusFirstChild,
+            onSiblingsExpand: (event: React.SyntheticEvent<HTMLElement>, treeItemProps: TreeItemProps) => {
+              if (treeItemProps.level !== 1)
```
miroslavstastny, line 34: Always wrap block statements by curly braces
@layershifter - can we check why lint did not explode? why prettier did not fix that?

YuanboXue-Amber:
closed because better version #16221

PR #15462, opened 2020-10-11
withSlots: refactor out redundant React.Children traversal

KevinTCoughlin:
Visible while profiling `<Stack />`. Avoid traversing React.Children twice for children count, re-use `toArray`'s traversal.

image

Ideally this wrapper could re-use children already being traversed within `<Stack />` itself.

image

### DIFF
--- a/packages/foundation/src/slots.tsx
+++ b/packages/foundation/src/slots.tsx
@@ -40,13 +40,6 @@ export function withSlots<P>(
 ): ReturnType<React.FunctionComponent<P>> {
   const slotType = type as ISlot<P>;
   if (slotType.isSlot) {
-    // TODO: There is something weird going on here with children embedded in props vs. rest args.
-    // Comment out these lines to see. Make sure this function is doing the right things.
-    const numChildren = React.Children.count(children);
-    if (numChildren === 0) {
-      return slotType(props);
-    }
-
     // Since we are bypassing createElement, use React.Children.toArray to make sure children are
     // properly assigned keys.
     // TODO: should this be mutating? does React mutate children subprop with createElement?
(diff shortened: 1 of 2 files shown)


packages/foundation/src/slots.tsx
```diff
@@ -40,13 +40,6 @@ export function withSlots<P>(
 ): ReturnType<React.FunctionComponent<P>> {
   const slotType = type as ISlot<P>;
   if (slotType.isSlot) {
```
Ridvanovskyy, line 42: Hello, @KevinTCoughlin. I have small questions:
1. Can we take out `children = React.Children.toArray (children);` outside of check `if (slotType.isSlot) {}`? As I see this array is needed both `if` and for `else`.
2.  After we can add - **const** children = React.Children.toArray(children);

Ridvanovskyy (approved):

khmakoto (approved):

khmakoto:
Let's wait until version 8 beta release to merge this. Alternatively, we can merge into the 7.0 branch and port into 8.0 later. See #15222 for more details.

khmakoto (changes requested):
Let's wait until version 8 beta release to merge this. Alternatively, we can merge into the 7.0 branch and port into 8.0 later. See #15222 for more details. Otherwise this LGTM.

KevinTCoughlin:
> Let's wait until version 8 beta release to merge this. Alternatively, we can merge into the 7.0 branch and port into 8.0 later. See #15222 for more details. Otherwise this LGTM.

@khmakoto here is the `7.0` cherry-pick, https://github.com/microsoft/fluentui/pull/15471. Can you ensure that is merged so OneDrive & SharePoint benefit soonest. Also please merge this change into `master` once the branch is stable so the perf fix is not lost. Thanks 🙇‍♂️.

cliffkoh (approved):

xugao (approved):
This is very interesting/good finding, thank you Kevin!

I see the gain in flamgrill tests as well:
image

As you pointed out, the `React.children.toArray` call has significant cost as well

khmakoto (approved):

PR #15463, opened 2020-10-12
Button: Adding transparent variant

### LINKED ISSUE #15105: Update ghost variant css once design is finalized
Daisy is working on finalizing the design for the button ghost variant. This issue tracks the followups needed after finalized design

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15105

#### Description of changes

This PR adds the `transparent` variant button that imitates the styling of the `ActionButton` in v7. Examples and visual regression tests have been updated to reflect this.

#### Focus areas to test

(optional)


Ridvanovskyy (approved):

packages/react-button/src/components/Button/__snapshots__/Button.test.tsx.snap
```diff
@@ -251,8 +247,8 @@ exports[`Button renders anchor when href prop is provided 1`] = `
         --color-brand-borderColor: transparent;
         --color-brand-checked-background: #005a9e;
         --color-brand-checked-contentColor: #ffffff;
-        --color-brand-checkedHovered-background: #005a9e;
-        --color-brand-checkedHovered-contentColor: #ffffff;
+        --color-brand-checked-hovered-background: #005a9e;
```
dzearing, line 250: this does not seem right. I think the states should all be at the same level.

packages/react-button/src/components/ToggleButton/useToggleButtonClasses.tsx
```diff
@@ -10,48 +10,85 @@ export const useToggleButtonClasses = makeVariantClasses<ToggleButtonState>({
...
-      checkedPressed: {
-        background: 'var(--ghost-checkedPressed-background)',
-        contentColor: 'var(--ghost-checkedPressed-contentColor)',
+
+      transparent: {
+        checked: {
+          background: 'transparent',
+          contentColor: palette?.themePrimary,
+          iconColor: palette?.themePrimary,
+          hovered: {
```
dzearing, line 85: this should be collapsed as `checkedHovered`.

dzearing: See `ColorTokens` in Theme. This should really be complaining in the typings. Can you see if you can make this typesafe?

packages/react-button/src/components/ToggleButton/useToggleButtonClasses.tsx
```diff
@@ -10,48 +10,85 @@ export const useToggleButtonClasses = makeVariantClasses<ToggleButtonState>({
...
+
+      ghost: {
+        checked: {
+          background: palette?.neutralLight,
+          contentColor: palette?.black,
+          iconColor: palette?.themeDarker,
+
+          hovered: {
+            background: palette?.neutralQuaternaryAlt,
+            contentColor: palette?.themePrimary,
```
xugao, line 74: could you also update the ghost to align with `CommandBarButton` styling? so in this case, we should keep the contentColor as is when hovered. i think this is the only change we need for `ghost`?

apps/vr-tests/src/stories/ButtonNext.stories.tsx
```diff
@@ -131,8 +142,13 @@ storiesOf('Button Next - With icon before content', module)
...
     </Button>
   ))
-  .addStory('Primary Disabled', () => (
-    <Button primary disabled icon="X">
+  .addStory('Transparent', () => (
+    <Button transparent icon="X">
+      Hello, world
+    </Button>
+  ))
+  .addStory('Transparent Disabled', () => (
```
xugao, line 150: let's also add a test for both `transparent` and `ghost` are set (which is what user needs to do if they need `ActionButton`)

packages/react-button/src/components/ToggleButton/useToggleButtonClasses.tsx
khmakoto: Spoke with @dzearing offline, I will address collapsing things in this PR but I'll check the typings in a different PR.

packages/react-button/src/components/Button/__snapshots__/Button.test.tsx.snap
khmakoto: Fixed!

packages/react-button/src/components/ToggleButton/useToggleButtonClasses.tsx
khmakoto: Done!

xugao (approved):

PR #15469, opened 2020-10-12
fix for positioning logic land on fractional values

### LINKED ISSUE #14795: Keyboard focus border right part is not visible in contextual menu item.
Repro steps:
1. Visit https://developer.microsoft.com/en-us/fluentui#/controls/web/contextualmenu
2. Use keyboard to open the contextual menu example.
3. Use keyboard to access each item. Observe the issue. The focus border right part is not visible.

image

dev-a11y:
fractional values in this place cause hard to debug subpixel rendering problems

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14795

#### Description of changes

see discussion in related issue


khmakoto:

joschect (approved):

layershifter:
@dev-a11y can you please generate a changelog entry via `yarn change` command? 🐱

dev-a11y:
@layershifter, sorry - had some difficulties setting up env on my machine, latest build error was sth with electron package installation and I gave it a rest; please reject this pull request and if still valid that code adjustment can be pushed by someone else; I'll remove my branch/fork soon

PR #15478, opened 2020-10-12
Add custom KeyUp support to Group Header

bcoard:
#### Pull request checklist

- [NA ] Addresses an existing issue: Fixes #0000
- [✔ ] Include a change request file using `$ yarn change`

#### Description of changes

Give the user the ability to override what happens with KeyUp. Currently it is only used with Right/Left to expand and collapse the group.

#### Focus areas to test

Ensure that Right/Left to expand and collapse still works.

### DIFF
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -4968,6 +4968,7 @@ export interface IGroupDividerProps {
     isSelected?: boolean;
     loadingText?: string;
     onGroupHeaderClick?: (group: IGroup) => void;
+    onGroupHeaderKeyUp?: (group: IGroup, ev: React.KeyboardEvent<HTMLElement>) => void;
     onRenderTitle?: IRenderFunction<IGroupHeaderProps>;
     onToggleCollapse?: (group: IGroup) => void;
     onToggleSelectGroup?: (group: IGroup) => void;
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -4968,7 +4968,7 @@ export interface IGroupDividerProps {
     isSelected?: boolean;
     loadingText?: string;
     onGroupHeaderClick?: (group: IGroup) => void;
-    onGroupHeaderKeyUp?: (group: IGroup, ev: React.KeyboardEvent<HTMLElement>) => void;
+    onGroupHeaderKeyUp?: (ev: React.KeyboardEvent<HTMLElement>, group: IGroup) => void;
     onRenderTitle?: IRenderFunction<IGroupHeaderProps>;
     onToggleCollapse?: (group: IGroup) => void;
     onToggleSelectGroup?: (group: IGroup) => void;
--- a/packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
+++ b/packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
@@ -185,13 +185,20 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
   };
 
   private _onKeyUp = (ev: React.KeyboardEvent<HTMLElement>): void => {
-    const shouldOpen = this.state.isCollapsed && ev.which === getRTLSafeKeyCode(KeyCodes.right, this.props.theme);
-    const shouldClose = !this.state.isCollapsed && ev.which === getRTLSafeKeyCode(KeyCodes.left, this.props.theme);
-    if (shouldClose || shouldOpen) {
-      this._toggleCollapse();
-      ev.stopPropagation();
-      ev.preventDefault();
+    const { group, onGroupHeaderKeyUp } = this.props;
+
+    if (onGroupHeaderKeyUp) {
+      onGroupHeaderKeyUp(group!, ev);
+    } else {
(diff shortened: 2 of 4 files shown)


bcoard:
> ## **Asset size changes**
> Project	Bundle	Baseline Size	New Size	Difference
> office-ui-fabric-react	[office-ui-fabric-react-DetailsList](https://microsoft.github.io/sizeAuditor-website/?baseline=139377&pr=139428&artifactFileName=office-ui-fabric-react-DetailsList.stats.json)	215.051 kB	215.113 kB	ExceedsBaseline     62 bytes
> office-ui-fabric-react	[office-ui-fabric-react-GroupedList](https://microsoft.github.io/sizeAuditor-website/?baseline=139377&pr=139428&artifactFileName=office-ui-fabric-react-GroupedList.stats.json)	121.499 kB	121.561 kB	ExceedsBaseline     62 bytes
> office-ui-fabric-react	[office-ui-fabric-react-ShimmeredDetailsList](https://microsoft.github.io/sizeAuditor-website/?baseline=139377&pr=139428&artifactFileName=office-ui-fabric-react-ShimmeredDetailsList.stats.json)	225.459 kB	225.521 kB	ExceedsBaseline     62 bytes
> ExceedsTolerance Over Tolerance (1024 B) ExceedsBaseline Over Baseline BelowBaseline Below Baseline New New Deleted  Removed **1 kB** = 1000 B
> 
> Baseline commit: [48eec32af2e31947e90b3324f81e2d8a3dbc003d](https://github.com/OfficeDev/office-ui-fabric-react/commit/48eec32af2e31947e90b3324f81e2d8a3dbc003d) [(build)](https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=139428)

Is there something that I should be doing about this?

packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -4968,6 +4968,7 @@ export interface IGroupDividerProps {
     isSelected?: boolean;
     loadingText?: string;
     onGroupHeaderClick?: (group: IGroup) => void;
+    onGroupHeaderKeyUp?: (group: IGroup, ev: React.KeyboardEvent<HTMLElement>) => void;
```
dzearing, line 4971: These args should be flipped:

```jsx
onGroupHeaderKeyUp: (ev, group)
```
Why: because normal keyup (and click) html event handlers take in "ev" as the first argument, and that's the expectation. It's wrong to put data first. We violate this principle in many places, but we should avoid doing that going forward.

I think `onGroupHeaderClick` is wrong here but fixing it would be a breaking change.

packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
```diff
@@ -185,13 +185,20 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
...
+
+    if (onGroupHeaderKeyUp) {
+      onGroupHeaderKeyUp(group!, ev);
+    } else {
+      const shouldOpen = this.state.isCollapsed && ev.which === getRTLSafeKeyCode(KeyCodes.right, this.props.theme);
+      const shouldClose = !this.state.isCollapsed && ev.which === getRTLSafeKeyCode(KeyCodes.left, this.props.theme);
+      if (shouldClose || shouldOpen) {
+        this._toggleCollapse();
+      }
     }
```
dzearing, line 198: Suggestion - when a user provides a keyup handler, the default should still execute unless `preventDefault` is called by the consumer. That way the consumer can do something like wire up some telemetry to the event, but still have the default behavior work.

```suggestion
    if (onGroupHeaderKeyUp) {
      onGroupHeaderKeyUp(ev, group!);
    }
    
    if (!ev.defaultPrevented) {
      const shouldOpen = this.state.isCollapsed && ev.which === getRTLSafeKeyCode(KeyCodes.right, this.props.theme);
      const shouldClose = !this.state.isCollapsed && ev.which === getRTLSafeKeyCode(KeyCodes.left, this.props.theme);
      if (shouldClose || shouldOpen) {
        this._toggleCollapse();
      }
    }
```

packages/office-ui-fabric-react/src/components/GroupedList/GroupHeader.base.tsx
```diff
@@ -185,6 +185,12 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
   };
 
   private _onKeyUp = (ev: React.KeyboardEvent<HTMLElement>): void => {
+    const { group, onGroupHeaderKeyUp } = this.props;
+
+    if (onGroupHeaderKeyUp) {
+      onGroupHeaderKeyUp(ev, group!);
+    }
+
     const shouldOpen = this.state.isCollapsed && ev.which === getRTLSafeKeyCode(KeyCodes.right, this.props.theme);
```
dzearing, line 194: wrap the default behavior in:

```jsx
if (!ev.isDefaultPrevented) {
  ...stuff
}
```

dzearing (approved):
Thanks!

packages/office-ui-fabric-react/src/components/GroupedList/GroupedList.types.ts
```diff
@@ -314,6 +314,9 @@ export interface IGroupDividerProps {
   /** Callback for when the group header is clicked. */
   onGroupHeaderClick?: (group: IGroup) => void;
 
+  /** Callback for when KeyUp on  the group header is invoked. */
```
dzearing, line 317:
```suggestion
  /** Callback for when the "keyup" event is fired on the group header . */
```

PR #15482, opened 2020-10-12
Update theme instances to contain components (scopedSettings)

xugao:
#### Pull request checklist

#### Description of changes
We recently added new typings for  theme and partial theme (`Theme` and `PartialTheme` without prefix `I`), which support a new property `components`. It's essentially the same as `scopedSettings`. This PR updates the theme instances we have in `azure-theme`, `mdl2-theme` and `theme-samples` packages to use the new typing and populate `components` property with scoped settings.
That way, when user uses `ThemeProvider`, they can simply pass these theme instances to `ThemeProvider` -> `theme` prop.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/azure-themes/README.md
+++ b/packages/azure-themes/README.md
@@ -6,17 +6,34 @@
 The Azure themes require the following import statements:
 
 ```js
-import { Fabric, Customizer } from '@fluentui/react';
-import { AzureCustomizationsLight, AzureCustomizationsDark } from '@uifabric/azure-themes';
+import {
+  AzureThemeDark,
+  AzureThemeLight,
+  AzureCustomizationsLight,
+  AzureCustomizationsDark,
+} from '@uifabric/azure-themes';
 ```
 
-The theme may subsequently be set to either the Azure or Azure-Dark themes
+The theme may subsequently be set to either the Azure or Azure-Dark themes.
+
+In case of applying theme using `Customizer`:
 
 ```jsx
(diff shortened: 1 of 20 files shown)


packages/theme-samples/src/WordCustomizations.ts
```diff
@@ -1,42 +1,27 @@
...
 import { addVariants } from '@uifabric/variants';
+import { Theme } from '@fluentui/theme';
+
+export const WordTheme: Theme = createTheme({
+  palette: {
+    themePrimary: '#2b579a',
+    themeSecondary: '#366ec2',
+  },
+  semanticColors: {
+    buttonBackground: 'white',
```
Ridvanovskyy, line 11: minor: what do you think about writing colors in **hex** format? Like in previous file.

Ridvanovskyy (approved):

packages/azure-themes/README.md
```diff
@@ -6,17 +6,34 @@
...
+In case of applying theme using `ThemeProvider`:
+
+```jsx
+  import { ThemeProvider } from '@fluentui/react-theme-provider';
+  const theme = AzureThemeDark; // or alternatively AzureThemeLight
+
+  <ThemeProvider theme={theme}>
+    <div>{child component}</div>
+  </ThemeProvider>
+```
```
Jacqueline-ms, line 39: Thanks for updating this

Jacqueline-ms (approved):
Azure theme storybook looks good, thank you!

dzearing (approved):

PR #15486, opened 2020-10-13
Component hooks: return only state, render function to be imported separately

dzearing:
Addresses RFC proposal #15497

This change decouples the unnecessarily coupled render function and state hook:

Before:
```jsx
import { useButton } from '@fluentui/react';

const Button = React.forwardRef((props, ref) => {
  const {state, render} = useButton(props, ref);

  return render(state);
});
```

After: 
```jsx
import { useButton, renderButton } from '@fluentui/react';

const Button = React.forwardRef((props, ref) => {
  const state = useButton(props, ref);

  return renderButton(state);
});
```


behowell (approved):
👍

packages/react-button/src/components/Button/index.ts
```diff
@@ -2,4 +2,5 @@ export * from './Button.types';
 export * from './useButtonClasses';
 export * from './useButton';
 export * from './useButtonState';
+export * from './renderButton';
```
khmakoto, line 5: This and `ToggleButton` are the only index files from which you are exporting a render function, is that on purpose?

khmakoto (approved):

packages/react-button/src/components/Button/index.ts
dzearing: no! Thank you! I will make sure they're all exported.

dzearing: done.

PR #15490, opened 2020-10-13
Combobox: move to react-next/react

joschect:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


ecraig12345 (approved):
Thanks! Ideally the examples/docs should be updated too, but I don't expect the ComboBox ones will have changed, so at this point we can just delete the extra examples later in a cleanup pass.

PR #15494, opened 2020-10-13
fix(Video): cleanup src on component unmount

miroslavstastny:
When `<Video autoPlay />` is unmounted, `HTMLVideoElement` is retained causing a memory leak.

This is an [Issue 969049 in Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=969049).

This PR is a workaround which resets `video.src` attribute on unmount.


pompomon (approved):

layershifter (approved):
Changelog entry?

Ridvanovskyy (approved):

PR #15501, opened 2020-10-13
Fix button IE polyfill demo

xugao:
#### Pull request checklist

#### Description of changes

Fix button IE polyfill demo

#### Focus areas to test

(optional)


ecraig12345 (approved):
Thanks! We really need to get some kind of lint enabled to prevent ES6 stuff from sneaking in like this. :(

tomi-msft (approved):

Issue #15473, opened 2020-10-12
@uifabric/foundation withSlots traverses React.Children even if already provided via parent component e.g. StackView


KevinTCoughlin:
Follow-up of https://github.com/microsoft/fluentui/pull/15462. `withSlots` should re-use `children` without having to recursively traverse the component tree (expensive). There are TODO comments mentioning this inefficiency. It appears that the method signature can be reworked to avoid having the re-flatten the `React.Children` to an array. See the "blue" and "red" outlined call stacks below.

![image](https://user-images.githubusercontent.com/706967/95779991-f5412800-0c7f-11eb-80f7-c5baf1e4f635.png)

https://github.com/microsoft/fluentui/blob/e915a1926c9b821f9f0d75e88445d1c431a284fc/packages/foundation/src/slots.tsx#L57

Can we re-use `...children: React.ReactNode[]` already traversed via `map` in `StackView` and passed into `withSlots`? This would eliminate the red call stack entirely.

`React.Children` traversal can be a performance bottleneck for wrapper components if the component sub-tree is large.

### Environment Information

- **Package version(s)**: `@uifabric/foundation` and dependents
- **Browser and OS versions**: ALL

### Please provide a reproduction of the bug in a codepen:

Profile this page using the browser's performance profiler to observe the performance bug: https://cdpn.io/kevintcoughlin/debug/gOMaYYV.

### Priorities and help requested:

Are you willing to submit a PR to fix? No

Requested priority: Normal

Products/sites affected: OneDrive & SharePoint

paulgildea:
@KevinTCoughlin Thanks!

@dzearing - FYI

dzearing:
Yup! Talked with Kevin about it a couple days ago. Great find.

I did a search of Children.count usage in the repo:

![image](https://user-images.githubusercontent.com/1110944/96021356-2480a200-0e04-11eb-80c5-611efccea349.png)

@khmakoto I see some usage in Card that seems unnecessary. That should be fixed.

The Text component is being redone in the `@fluentui/react-text` package, and does not have the reference.

The `LayoutGroup` component in the experimental package should be deleted; in fact we should be deleting all of the dusty components we know aren't used in ODSP, and any we do know should be moved out of experiments into their own package (Tile, TileView)

PR #15503, opened 2020-10-13
[Perf] Avoid React.Children.count traversal for empty state early return

KevinTCoughlin:
#### Pull request checklist

#### Description of changes

Avoid calls to `React.Children.count` which recursively traverses child components. The cost grows with component hierarchy. These cases were for early return `null` so they should be quick. In cases where `children` are not passed via props the value is `undefined`.

`Text` for example:

image

#### Focus areas to test

(optional)

Related: #15462 #15473


KevinTCoughlin:
These need to explicitly check `=== undefined`. I'll push an iteration. The reason is `<div>{0}</div>` is treated as falsey otherwise (thanks unit tests).

packages/react-cards/src/components/Card/CardSection/CardSection.view.tsx
```diff
@@ -1,12 +1,11 @@
 /** @jsx withSlots */
-import * as React from 'react';
 import { withSlots, getSlots } from '@uifabric/foundation';
 import { Stack } from '@fluentui/react/lib/Stack';
 import { ICardSectionComponent, ICardSectionProps, ICardSectionSlots } from './CardSection.types';
 
 export const CardSectionView: ICardSectionComponent['view'] = props => {
   const { children, ...rest } = props;
-  if (React.Children.count(children) < 1) {
+  if (children === undefined) {
```
xugao, line 8:
```suggestion
  if (children == null) {
```

xugao: we should do the check like this so it covers both `undefined` and `null`

KevinTCoughlin: Sure, tho I wouldn't except `null` ever to be passed. React passes `undefined`. `null` would only be passed by library consumers.

xugao (approved):

khmakoto:
Let's wait until after the snap period to merge or spin a v7 version of this PR if needed sooner.

Ridvanovskyy (approved):

change/@fluentui-react-internal-2020-10-13-16-47-27-keco-react-children-for-empty.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
```
ecraig12345, line 2:
```suggestion
  "type": "prerelease",
```

change/@uifabric-react-cards-2020-10-13-16-47-27-keco-react-children-for-empty.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Avoid React.Children traversal for early return cases",
+  "packageName": "@uifabric/react-cards",
```
ecraig12345, line 4:
```suggestion
  "type": "prerelease",
  "comment": "Avoid React.Children traversal for early return cases",
  "packageName": "@fluentui/react-cards",
```

ecraig12345 (commented):
Guessing this can go in now? Needs some minor change file updates and a merge with master.

ecraig12345:
Actually both the change files and the conflicts were easy to fix in the web UI, so I took care of that.

KevinTCoughlin:
> Actually both the change files and the conflicts were easy to fix in the web UI, so I took care of that.

Thank you @ecraig12345 🙇‍♂️. @khmakoto can we merge this?

khmakoto (approved):

ecraig12345 (approved):

Ridvanovskyy (commented):
👍

PR #15513, opened 2020-10-14
Button: Renaming `fluid` prop to `block` so that it is inline with the most used terminology in the web

khmakoto:
#### Pull request checklist

#### Description of changes
This PR renames the `fluid` prop in the `Button` components to `block` so that it is inline with the most used terminology in the web.

#### Focus areas to test

(optional)


dzearing (approved):

khmakoto:

PR #15516, opened 2020-10-14
Implementing useUnmount, useMount, and useMountSync hooks within react-hooks package

czearing:
#### Pull request checklist

#### Description of changes
Adding three new hooks to the react-hooks package that replicate `componentWillMount` and `componentWillUnmount`:

#### `useMount`
Asynchronously fires a callback before mount.
```jsx
export const useMount = (callback: () => void) => {
  const mountRef = React.useRef(callback);
  mountRef.current = callback;
  React.useEffect(() => {
    mountRef.current?.();
  }, []);
};
```

#### `useMountSync`
 Synchronously fires a callback before mount.

```jsx
export const useMountSync = (callback: () => void) => {
  const mountRef = React.useRef(callback);
  mountRef.current = callback;
  React.useLayoutEffect(() => {
    mountRef.current?.();
  }, []);
};

```
#### `useUnmount`
Synchronously fires a callback during unmount.

```jsx
export const useUnmount = (callback: () => void) => {
  const unmountRef = React.useRef(callback);
  unmountRef.current = callback;
  React.useLayoutEffect(
    () => () => {
      unmountRef.current?.();
    },
    [],
  );
};

```


packages/react-hooks/README.md
```diff
@@ -187,6 +187,48 @@ const Example = React.forwardRef(function Example(props: {}, forwardedRef: React
 });
 ```
 
+## useMount
+
+```ts
+const useMount: (callback: () => void) => void;
+```
+
+Hook that asynchronously fires a callback before mount.
```
dzearing, line 196:
```suggestion
Hook which asynchronously executes a callback once the component has been mounted.
```

packages/react-hooks/README.md
```diff
@@ -187,6 +187,48 @@ const Example = React.forwardRef(function Example(props: {}, forwardedRef: React
...
+```
+
+## useMountSync
+
+```ts
+const useMountSync: (callback: () => void) => void;
+```
+
+Hook that synchronously fires a callback before mount. This should only be used if useMount causes problems as it will
+trigger debug warnings in server-rendered scenarios.
```
dzearing, line 218:
```suggestion
Hook which synchronously execute a callback when the component has been mounted. Use `useMount` for most scenarios. You should only use the synchronous version in the rare case you need to perform an action after the component has been mounted and before the browser paints, such as measuring content and adjusting the result. Using this will trigger debug warnings in server-rendered scenarios.
```

packages/react-hooks/README.md
```diff
@@ -313,6 +355,26 @@ function useTarget<TElement extends HTMLElement = HTMLElement>(
 
 Hook which queries the document for the element indicated by a CSS query string (if provided), or returns the element/event/point provided. Also attempts to determine the Window object for the provided target.
 
+## useUnmount
+
+```ts
+const useUnmount: (callback: () => void) => void;
+```
+
+Hook that synchronously fires a callback during unmount.
```
dzearing, line 364:
```suggestion
Hook which synchronously executes a callback when the component is about to unmount.
```

packages/react-hooks/src/useMount.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { useMount } from './useMount';
+
+describe('useMount', () => {
+  it('fires a callback', () => {
+    let value = false;
+
+    const TestComponent: React.FunctionComponent = () => {
+      useMount(() => {
+        value = true;
+      });
```
dzearing, line 12: Slightly better way to test: using `jest.fn()` to mock a mount callback:

```tsx
const onMount = jest.fn();

// ...mount a component using the onMount callback

expect(onMount).toBeCalledTimes(1);
```

dzearing (approved):
Let's get the nit comments addressed, but otherwise this looks great.

packages/react-hooks/README.md
```diff
@@ -187,6 +187,47 @@ const Example = React.forwardRef(function Example(props: {}, forwardedRef: React
 });
 ```
 
+## useMount
+
+```ts
+const useMount: (callback: () => void) => void;
+```
+
+Hook which asynchronously executes a callback once the component has been mounted.
```
ecraig12345, line 196: Might be good to note in the readme which hook is used under the hood (and link to the docs for that hook).

packages/react-hooks/src/useUnmount.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Hook which synchronously executes a callback when the component is about to unmount.
+ *
+ * @param callback - Function to call during unmount.
+ */
+export const useUnmount = (callback: () => void) => {
+  const unmountRef = React.useRef(callback);
+  unmountRef.current = callback;
+  React.useLayoutEffect(
```
ecraig12345, line 11: I don't think we need to do useLayoutEffect here?

ecraig12345 (approved):

PR #15517, opened 2020-10-15
Avoid using css vars for default styles applied by ThemeProvider

xugao:
#### Pull request checklist

#### Description of changes
avoid using css vars for default styles applied by ThemeProvider (ie11 safe)

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-theme-provider/src/__snapshots__/ThemeProvider.test.tsx.snap
+++ b/packages/react-theme-provider/src/__snapshots__/ThemeProvider.test.tsx.snap
@@ -5,11 +5,6 @@ exports[`ThemeProvider can apply body theme to body 1`] = `
   class=
 
       {
-        --body-fontFamily: 'Segoe UI', 'Segoe UI Web (West European)', 'Segoe UI', -apple-system, BlinkMacSystemFont, 'Roboto', 'Helvetica Neue', sans-serif;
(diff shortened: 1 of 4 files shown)


packages/react-theme-provider/src/__snapshots__/ThemeProvider.test.tsx.snap
```diff
@@ -60,11 +55,6 @@ exports[`ThemeProvider can apply body theme to body 2`] = `
   className=
       foo
       {
-        --body-fontFamily: 'Segoe UI', 'Segoe UI Web (West European)', 'Segoe UI', -apple-system, BlinkMacSystemFont, 'Roboto', 'Helvetica Neue', sans-serif;
-        --body-fontSize: 14px;
-        --body-fontWeight: 400;
-        --body-mozOsxFontSmoothing: grayscale;
-        --body-webkitFontSmoothing: antialiased;
```
dzearing, line 67: why aren't these defined?

xugao: we can keep these, but they are not being used anywhere

khmakoto (approved):

xugao:
will make the change in master instead

PR #15518, opened 2020-10-15
Fixing variants typings in makeVariantClasses

### LINKED ISSUE #15483: Fix tokens and variants typings
Recently, we discovered that the `variants` were not taking the correct typings where needed using the `tokens` defined. We need to fix these so that we catch errors at build time.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15483

#### Description of changes

This PR fixes the typings for the `variants` when using `makeVariantClasses` to restrict what can be sent as a value to that parameter in the function.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-button/etc/react-button.api.md
+++ b/packages/react-button/etc/react-button.api.md
@@ -78,17 +78,17 @@ export interface ButtonState extends ButtonProps {
 
 // @public (undocumented)
 export type ButtonTokens = ColorTokens & FontTokens & {
-    paddingLeft?: string;
-    paddingRight?: string;
-    paddingTop?: string;
-    paddingBottom?: string;
+    paddingLeft?: string | number;
+    paddingRight?: string | number;
+    paddingTop?: string | number;
+    paddingBottom?: string | number;
     margin?: string;
     height?: string;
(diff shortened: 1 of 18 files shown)


packages/react-button/etc/react-button.api.md
```diff
@@ -78,25 +78,26 @@ export interface ButtonState extends ButtonProps {
 
 // @public (undocumented)
 export type ButtonTokens = ColorTokens & FontTokens & {
-    paddingLeft?: string;
-    paddingRight?: string;
-    paddingTop?: string;
-    paddingBottom?: string;
+    paddingLeft?: string | number;
```
dzearing, line 81: I don't think we should allow for number. Or at least not yet. 

When you use `useInlineTokens`, it will not auto expand numbers into PX. We could add it, it will cost a bit in perf, and it should be done right. Until that works, i'd suggest we stick with strings.

dzearing (approved):
remove numbers

packages/react-button/etc/react-button.api.md
khmakoto: Removed!

PR #15532, opened 2020-10-15
Enable unit tests for @fluentui/react

xugao:
#### Pull request checklist

#### Description of changes
Enable unit tests for `@fluentui/react`. Previously they are not because `test` script is missing.

Fixing failing combobox tests: `forwardedRef` is not working with `renderer.create`. Initially i tried to switch to using `mount`. then `mount` is running into troubles with us mocking `createPortal`, which is something we do need for combobox. Therefore I updated the tests to not using `ref` to finding the element but using `container.root.find*` instead.
 
version import is not working correctly with `@fluentui/react`, see [this demo](https://codesandbox.io/s/competent-wave-os1rd?file=/src/App.tsx). 
i don't quite have an answer on how to fix it yet, so disabling it for now to unblock re-enabling rest of the tests.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-window-provider/src/version.ts
+++ b/packages/react-window-provider/src/version.ts
@@ -1,4 +1,4 @@
 // Do not modify this file; it is generated as part of publish.
 // The checked in version is a placeholder only and will not be updated.
 import { setVersion } from '@uifabric/set-version';
-setVersion('@fluentui/react-link', '0.0.0');
+setVersion('@fluentui/react-window-provider', '0.0.0');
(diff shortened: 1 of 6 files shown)


packages/react-window-provider/src/version.ts
```diff
@@ -1,4 +1,4 @@
 // Do not modify this file; it is generated as part of publish.
 // The checked in version is a placeholder only and will not be updated.
 import { setVersion } from '@uifabric/set-version';
-setVersion('@fluentui/react-link', '0.0.0');
+setVersion('@fluentui/react-window-provider', '0.0.0');
```
dzearing, line 4: geez. Nice catch.

ecraig12345:
The way the version imports worked before was to include the import in certain critical files which everything else references: `office-ui-fabric-react/src/{Layer,Styling,Utilities}.ts`. That probably won't work in a package that's primarily re-exporting, so @czearing made a change in react-next to do `import './version'` in all the root files. Might need to do the same in react.

khmakoto (approved):

ecraig12345 (approved):

Issue #14679, opened 2020-08-21
Re-export date-time components from react


JustSlone:
(from @ecraig12345)

Part 2/4 in a series of changes that must happen in master *after* v7 development moves to a branch, but *before* beta publishing:
1. #15152 OUFR/react-next merge 
2. (this) #14679 ~~Move~~ Re-export date-time components
3. #13384 uifabric/fluentui package renames
4. #14689 set up beta releases

### Motivation
 
We'd like to re-export the `@uifabric/date-time` versions of Calendar and DatePicker by default from the suite (because they're strict mode compliant and otherwise better). ~~but we can't have the suite re-export components from date-time since they depend on other components which currently still live in the suite.~~ EDIT: Now that we have `react-internal`, this is easier!

### Proposed solution

We were able to break the dependency cycle by putting everything `date-time` depends on in `react-internal`, so now we can just have `react` re-export from `date-time`.
- [ ] Delete Calendar and DatePicker from `react-internal`
- [ ] In `react`, add dep on `date-time` and re-export Calendar and DatePicker from `date-time`
- [x] In `react-examples`:
  - [x] Copy over `react/{Calendar,DatePicker}/{docs,*.doc.tsx}` to `date-time` (the docs from the main package are more up to date)
  - [x] Update example list and import paths in .doc.tsx to reflect the date-time package
  - [x] In the `react-examples/src/date-time`, delete CalendarPage, DatePickerPage, and the old `demo` folder
  - [x] Delete old Calendar and DatePicker examples from `react-examples/src/react`
- [x] Update files in `fabric-website` and `fabric-website-resources` to reference the `date-time` package
- [ ] Ensure release notes reflect this change

<details>
<summary>Click to show OLD plan</summary>
- ~~Move existing Calendar and DatePicker to `/compat` in suite (there are enough API changes that I'm hesitant to delete them entirely)~~ (EDIT: there are not actually that many API changes, so we may skip this)
- Move date-time Calendar and DatePicker to suite and export by default
- Make date-time re-export suite Calendar and DatePicker
- Update release notes about this

This must be done in the final beta prep period since it involves code moves in packages besides react-next. (It's a very straightforward change which anyone could do.)
</details>

ecraig12345:
Did a more complete comparison of the APIs between OUFR and date-time, and actually the only major changes are the following:
- Conversion of Calendar from SCSS (module-style not globals) to CSS-in-JS
- Removed the following from `ICalendarProps`:
[...]

PR #15546, opened 2020-10-16
Update date-time Calendar and DatePicker examples and use in public docs

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Part of #14679

#### Description of changes

Since we'll be changing to re-export Calendar and DatePicker from `date-time` and deleting the old implementations as part of version 8, the website should be switched to show the new implementations. This PR doesn't make the re-export change yet, but it updates which examples and docs are shown on the website, and deletes the old examples/docs.

A couple related updates:
- The examples for the old Calendar and DatePicker had been converted to function components and had scss removed, but the date-time example versions hadn't been updated. So move those updates over to date-time (as well as other minor improvements to the examples).
- Add date-time to the list of packages considered when generating API docs
- Stop publishing the api-docs package, since the packages that consume it are also no longer published

And some fixes to example snapshot tests:
- Add example snapshot tests for all the component packages which are re-exported from the suite. I think it was just an oversight that this hadn't been done already.
- Add a way to update example snapshots ONLY for a particular package: from `packages/react-examples`, run `yarn update-snapshots --package some-package`
- EDIT: After doing all this, it turns out that most/all of the date-time tests either cause infinite loops with date mocking, or spurious diffs due to time zone changes between local machine and build machine. So disable the tests for date-time for now (but at least we have them for the other packages now?).

(Thanks to another recent change, the Calendar and DatePicker examples will automatically show up in the storybook for the suite package.)


packages/react-examples/src/ComponentExamples.test.tsx
```diff
@@ -118,7 +125,18 @@ function getPackageAndExampleName(examplePath: string): [string, string] {
 }
 
 /** Run tests on these packages' examples */
-const includedPackages = ['react', 'react-cards', 'react-focus'];
+const includedPackages = [
+  'react',
+  'react-button',
+  'react-cards',
+  'react-checkbox',
+  'date-time',
```
khmakoto, line 133: nit: Should `date-time` go at the beginning of this list so it is in alphabetical order?

khmakoto (approved):

xugao (approved):

packages/react-examples/src/ComponentExamples.test.tsx
ecraig12345: It will be in alphabetical order once date-time is renamed to `react-date-time` shortly 😆

packages/react-examples/src/date-time/Calendar/Calendar.doc.tsx
```diff
@@ -0,0 +1,108 @@
...
+      code: CalendarInlineMultidayDayViewExampleCode,
+      view: <CalendarInlineMultidayDayViewExample />,
+    },
+    {
+      title: 'Calendar a tooltip for each day and disabling weekends',
+      code: CalendarInlineCustomDayCellRefExampleCode,
+      view: <CalendarInlineCustomDayCellRefExample />,
+    },
+    {
+      title: 'Calendar launched from a button',
```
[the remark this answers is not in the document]
ecraig12345: The examples in the new version are based on the set from the date-time package that was converted to use discrete examples for each concept rather than a mega-example with props (I think @lorejoh12 did those conversions?). I also simplified some of the names. It looks like we have 13 now compared to 15 before, but if there's anything important missing we can add it back later.

PR #15553, opened 2020-10-16
v7: Change ISelectedPersonaProps GetExpandedItems to be async

elisabethcvs:
#### Pull request checklist

#### Description of changes

v8 PR - https://github.com/microsoft/fluentui/pull/15557

- Change the function stated above to be async.
- Modified the example
- Disabled the test - with some googling, turns out it's difficult (or impossible) to test this with the setup we've got. I'm new to this though, so if there's a way to get this working let me know. Can also look at it some more when moving this to the main branch.


elisabethcvs:
Can submit this to 8 myself if that makes it easier.

sjadelson (approved):

PR #15555, opened 2020-10-16
7.0: Add option for selectable headers and footers to FloatingSuggestionsComposite (and therefor the UnifiedPicker)

elisabethcvs:
#### Pull request checklist

#### Description of changes
v8 PR - https://github.com/microsoft/fluentui/pull/15432

Adding a property so that consumers can add selectable footers and headers to the floating picker. It wasn't possible to keyboard navigate to headers or footers previously. This property takes precedence over the old render functions, but I've kept them, so this change is backwards compatible.

I added keyboard navigation to the UnifiedPicker as well.

I added an example for the floating suggestions composite and I modified the UPP single example so that the picker has headers and footers.

#### Focus areas to test

I have checked that keyboard navigation works properly when:
- no headers or footers
- non selectable headers and footers
- non selectable header, selectable footers
- non selectable header, one selectable footer
- selectable header, one selectable footer

### DIFF
--- a/change/@fluentui-react-examples-2020-10-16-11-01-14-headerfooterselectable7.0.json
+++ b/change/@fluentui-react-examples-2020-10-16-11-01-14-headerfooterselectable7.0.json
@@ -1,0 +1,8 @@
+{
+  "type": "none",
+  "comment": "Add example to FloatingSuggestionsComposite with selectable headers and footers",
+  "packageName": "@fluentui/react-examples",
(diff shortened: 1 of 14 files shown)


nebhatna (approved):

nebhatna (commented):
could you please double check on the relative paths in v7 since the folder structure is different..maybe we can consider using absolute path to avoid this breakage.. i would leave it upto you

PR #15557, opened 2020-10-16
Change ISelectedPersonaProps GetExpandedItems to be async

elisabethcvs:
#### Pull request checklist

#### Description of changes

- Change the function stated above to be async.
- Modified the example
- Disabled the test - with some googling, turns out it's difficult (or impossible) to test this with the setup we've got. I'm new to this though, so if there's a way to get this working let me know.


change/@fluentui-react-examples-2020-10-16-11-15-49-expandasync.json
```diff
@@ -0,0 +1,8 @@
+{
```
nebhatna, line 1: might want to remove duplicate changefile

packages/react-experiments/src/components/SelectedItemsList/SelectedPeopleList/SelectedPeopleList.test.tsx
```diff
@@ -68,7 +68,7 @@ describe('SelectedPeopleList', () => {
     expect(onItemsRemoved).toBeCalledTimes(1);
   });
 
-  it('group expansion as one of the items in the list', () => {
+  /*it('group expansion as one of the items in the list', () => {
```
nebhatna, line 71: are you not able to test this because you changed it to async.. or is there any other issue? Did you look into using this - https://www.pluralsight.com/guides/test-asynchronous-code-jest .. if you want to update this test to be simpler , that should be okay to .. something like making sure we are able to edit and see options etc..

elisabethcvs: From what I can tell (and testing) none of that will work. The problem is that we have a UI click which calls a function, which has a promise/then structure. All of those are designed to trust the promise function itself. We could do a test that tests the test GetExpandedItems function to see what it returns, but that wouldn't really test our code

nebhatna: yea makes sense.. i would just remove this unit test then.

nebhatna (approved):

ecraig12345 (approved):

PR #15558, opened 2020-10-16
Removing findDOMNode usage from Pivot

### LINKED ISSUE #14322: Pivot (next): remove findDOMNode usage

czearing:
#### Pull request checklist

- [X] Fixes #14322 

#### Description of changes
Removing findDOMNode usage from the **react-tabs** `Pivot` component.

**Changing BaseButton's wrapper span to a fragment in react-internal**

(*This change was made to allow elementRef to recieve the root element.*)
```jsx
Before:
<span style={{ display: 'inline-block' }}>
   {Content}
   {this._shouldRenderMenu() && onRenderMenu(menuProps, this._onRenderMenu)}
</span>

After:
<>
   {Content}
   {this._shouldRenderMenu() && onRenderMenu(menuProps, this._onRenderMenu)}
</>
```

**Adding an elementRef to Pivot's CommandButton**
```jsx
elementRef={overflowMenuButtonRef}
componentRef={overflowMenuButtonComponentRef}
```

**Updating Pivot's CommandButton className**
```jsx
className={classNames.link + ' ' + classNames.overflowMenuButton}
```

#### Other Minor Changes
* **Fix** Organizing the placement of ref and state in Pivot.base.tsx
* **Fix** Removing `setOverflowMenuButtonRef` and replacing it with `overflowMenuButtonRef`

#### Focus areas to test
`Pivot`, `BaseButton`


packages/react-tabs/src/components/Pivot/Pivot.base.tsx
```diff
@@ -66,13 +65,25 @@ const isPivotItem = (item: React.ReactNode): item is PivotItem => {
...
     const focusZoneRef = React.useRef<IFocusZone>(null);
-    const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(props, divProperties);
+    const overflowMenuButtonComponentRef = React.useRef<IButton>(null);
+    const pivotId: string = useId('Pivot');
+
     const [selectedKey, setSelectedKey] = useControllableValue(props.selectedKey, props.defaultSelectedKey);
+
+    const { componentRef, theme, linkSize, linkFormat, overflowBehavior } = props;
+    // The overflow menu starts empty and items[] is updated as the overflow items change
+    const overflowMenuProps: IContextualMenuProps = {
```
ecraig12345, line 76: Not new but this should probably be memoized (or even useConst) so it doesn't mutate every render. I can actually see that being a source of bugs too, with the way it's used.

packages/react-tabs/src/components/Pivot/Pivot.base.tsx
```diff
@@ -262,11 +253,9 @@ export const PivotBase: React.FunctionComponent<IPivotProps> = React.forwardRef<
           {items}
           {overflowBehavior === 'menu' && (
             <CommandButton
-              // TODO when the span wrapper is removed from <CommandButton>, set
-              // className={classNames.link + ' ' + classNames.overflowMenuButton}
-              className={classNames.link}
-              ref={setOverflowMenuButtonRef}
-              componentRef={overflowMenuButtonComponentRef as React.RefObject<IButton>}
+              className={classNames.link + ' ' + classNames.overflowMenuButton}
```
ecraig12345, line 256:
```suggestion
              className={css(classNames.link, classNames.overflowMenuButton)}
```

ecraig12345 (commented):
You'll probably need to update a bunch of snapshots (probably in all the packages) due to the button change.

ecraig12345:
Hmm actually never mind, the build seems to have passed with no snapshot changes??

packages/react-tabs/src/components/Pivot/Pivot.base.tsx
```diff
@@ -66,13 +65,25 @@ const isPivotItem = (item: React.ReactNode): item is PivotItem => {
...
+      items: [],
+      doNotLayer: true,
+      alignTargetEdge: true,
+      directionalHint: DirectionalHint.bottomRightEdge,
+    };
+
     let classNames: { [key in keyof IPivotStyles]: string };
+    const divProps = getNativeProps<React.HTMLAttributes<HTMLDivElement>>(props, divProperties);
+
+    let linkCollection = getLinkItems(props, pivotId);
```
behowell, line 86: (nit) out of curiosity, why are you moving around so many lines of code in this function? It's fine to clean things up if the code is messy, but this doesn't necessarily seem better than it was before. (E.g. previously, overflowMenuProps was declared closer to where it's used, which makes more sense to me). It's not necessarily a problem, but it seems unnecessary and makes the diff harder to read.

behowell (approved):

packages/react-tabs/src/components/Pivot/Pivot.base.tsx
czearing: I reverted the overflowMenuProps move and the diffing should be improved. The intention is to help keep consistency in Pivot’s useRef and useState/useControllableValue placement to follow patterns in other functional components.

ecraig12345 (approved):

PR #15560, opened 2020-10-16
Fix inferred type error related to ITheme

xugao:
#### Pull request checklist

#### Description of changes
Run into a report of following typing error related to `ITheme`.
```
 error TS2742: The inferred type of 'navGroupDividerStyle' cannot be named without a reference to '../../../../../common/temp/node_modules/.registry.npmjs.org/@uifabric/styling/7.16.12_8fbe4bd393535fec5ef0af7d5e2846e1/node_modules/@fluentui/theme/lib/types'. This is likely not portable. A type annotation is necessary.
35 export const navGroupDividerStyle = (theme: ITheme): IStyle => {
```

the issue can only be fixed in user's code by adding explicit type definition:
```
export const navGroupDividerStyle: (theme: ITheme) => IStyle = (theme: ITheme): IStyle => {
```

Changing `type` to `interface ... extends ...` fixed the issue. However I am still puzzled by why
#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
+++ b/packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
@@ -1322,7 +1322,7 @@ export const getMeasurementCache: () => {
 };
 
 // @public (undocumented)
-export const getMenuItemStyles: (theme: import("@fluentui/theme/lib/types").Theme) => IMenuItemStyles;
+export const getMenuItemStyles: (theme: ITheme) => IMenuItemStyles;
 
 // @public
 export const getNextResizeGroupStateProvider: (measurementCache?: {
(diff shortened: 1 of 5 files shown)


packages/office-ui-fabric-react/etc/office-ui-fabric-react.api.md
```diff
@@ -1322,7 +1322,7 @@ export const getMeasurementCache: () => {
 };
 
 // @public (undocumented)
-export const getMenuItemStyles: (theme: import("@fluentui/theme/lib/types").Theme) => IMenuItemStyles;
+export const getMenuItemStyles: (theme: ITheme) => IMenuItemStyles;
```
xugao, line 1325: this change also indicates there is indeed a different

ecraig12345 (approved):

PR #15561, opened 2020-10-16
Disable marquee selection in UnifiedPeoplePicker

sjadelson:
#### Pull request checklist

#### Description of changes

MarqueeSelection is not the desired behavior for UPP. This change reverts to EPP behavior - no drag selection. A later change will implement correct behavior.

#### Focus areas to test

(optional)


packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -351,45 +343,43 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
       onCopy={_onCopy}
     >
       <FocusZone direction={FocusZoneDirection.bidirectional} {...focusZoneProps}>
-        <MarqueeSelection selection={selection} isEnabled={true}>
```
elisabethcvs, line 354: Is there anything else this is doing for us that we'll lose if we remove it?

elisabethcvs (commented):
Can you run the accessibility checker to make sure this doesn't break anything? I don't think it will, but should probably double check

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
nebhatna: also let us document this as a comment here so we know

sjadelson (commented):
Good suggestion on the acc checker.  No violations with 0, 1, 3, and 10 recipients displayed.

layershifter:
@sjadelson is still a thing? Can you please resolve merge conflicts in this case?

sjadelson:
Thanks for pointing that out.  That change was re-jiggered and submitted for the v8 release.  This one is no longer needed, and has been abandoned.

PR #15567, opened 2020-10-16
Rename @uifabric/styling to @fluentui/style-utilities

ecraig12345:
#### Pull request checklist

#### Description of changes

Rename `@uifabric/styling` to `@fluentui/style-utilities`.


packages/style-utilities/README.md
```diff
@@ -1,30 +1,30 @@
...
 ## Overriding the theme colors
 
-The default palette of colors matches the default Fabric core styling conventions. However, it is possible to override the color slots to match your product requirements:
+The default palette of colors matches the default Fluent UI (and Fabric Core) styling conventions. However, it is possible to override the color slots to match your product requirements:
 
 ```tsx
 import {
   loadTheme({
-    palette: {
+    palette: {q
```
behowell, line 27: Typo?

packages/react-examples/src/react/ThemeGenerator/ThemeGenerator.doc.tsx
```diff
@@ -95,15 +95,16 @@ export class ThemeGeneratorPage extends React.Component<{}, IThemeGeneratorPageS
...
     return (
       <div className="ms-themer">
         <div className="overview">
           <h2 id="Overview">Overview</h2>
           <p>
-            This tool helps you easily create all the shades and slots for a custom theme. The theme can be used by
-            Fabric React's styling package, see the{' '}
+            {/* eslint-disable-next-line @fluentui/max-len */}
+            This tool helps you easily create all the shades and slots for a custom theme. The theme can be used by the{' '}
+            <code>@fluentui/style-utilities</code> package; see the{' '}
```
behowell, line 107: (nit) maybe just reflow the line instead of disabling the line length warning?

behowell (approved):

packages/style-utilities/README.md
ecraig12345: Oops, yes...seems I thought I was focused on a console when I was actually in the editor 😆

packages/react-examples/src/react/ThemeGenerator/ThemeGenerator.doc.tsx
ecraig12345: Unfortunately this is one of the handful of cases where prettier and eslint conflict--prettier thinks the line is "prettiest"(??) this way despite being slightly over the max length, but eslint doesn't have that exception. There's not an easy way to change how prettier formats things, so usually in cases like this we disable the lint rule one-off.

packages/style-utilities/README.md
```diff
@@ -1,25 +1,25 @@
...
 
-```bash
-npm install --save @uifabric/styling
+```
+npm install --save @fluentui/style-utilities
 ```
 
 This gives you access to styling-related constants, utilities, and Fabric Core style classes through JavaScript.
 
 ## Overriding the theme colors
```
xugao, line 20: could you help remove this section? since we are trying to move theme related out of this package. the reason why `loadTheme` is still here is ... we don't have a better place for it. 
User can find information about it in our [wiki](https://github.com/microsoft/fluentui/wiki/How-to-apply-theme-to-Fluent-UI-React-components#loadtheme). So we don't need to call-out here

xugao (approved):

PR #15569, opened 2020-10-16
Updating Modal to a function component within react-internal

### LINKED ISSUE #14297: Modal: strict mode compliance

czearing:
#### Pull request checklist

- [X] Fixes #14297 

#### Description of changes

Updating `Modal` to a function component with **react-internal**.

#### Focus areas to test
`Modal`


packages/react-internal/src/components/Modal/Modal.base.tsx
```diff
@@ -1,202 +1,127 @@
...
+    const focusTrapZone = React.useRef<IFocusTrapZone>(null);
+    const mergedRef = useMergedRefs(rootRef, ref);
 
-    const layerClassName = layerProps === undefined ? '' : layerProps.className;
+    const modalResponsiveMode = useResponsiveMode(mergedRef);
 
+    const FocusTrapZoneId = useId();
+
+    const doc = getDocument();
+    const win = getWindow();
```
xugao, line 98: we should use `useDocument` and `useWindow` from `react-window-provider`

packages/react-internal/src/components/Stack/__snapshots__/Stack.test.tsx.snap
```diff
@@ -392,7 +392,7 @@ exports[`Stack renders horizontal Stack with a gap in rtl context correctly 1`]
           text-overflow: ellipsis;
         }
         & > *:not(:first-child) {
-          margin-right: 10px;
+          margin-left: 10px;
```
xugao, line 395: how did this change happen?

czearing: It seems like some of the packages hadn't been built. It should be good now.

packages/react-internal/src/components/Modal/Modal.base.tsx
```diff
@@ -1,202 +1,131 @@
...
-      return null;
-    }
+    const rootRef = React.useRef<HTMLDivElement>(null);
+    const focusTrapZone = React.useRef<IFocusTrapZone>(null);
+    const mergedRef = useMergedRefs(rootRef, ref);
 
-    const layerClassName = layerProps === undefined ? '' : layerProps.className;
+    const modalResponsiveMode = useResponsiveMode(mergedRef);
+
+    const FocusTrapZoneId = useId();
```
xugao, line 97:
```suggestion
    const FocusTrapZoneId = useId('ModalFocusTrapZone');
```

packages/react-internal/src/components/Modal/Modal.types.ts
```diff
@@ -94,6 +95,11 @@ export interface IModalProps extends React.ClassAttributes<ModalBase>, IWithResp
    */
   onDismissed?: () => any;
 
+  /**
+   * The specified responsiveMode value for Modal to use.
+   * @defaultvalue ResponsiveMode.small
+   */
+  responsiveMode?: ResponsiveMode | undefined;
```
xugao, line 102:
```suggestion
  responsiveMode?: ResponsiveMode;
```

packages/react-internal/src/components/Modal/Modal.styles.ts
```diff
@@ -30,6 +31,9 @@ export const getStyles = (props: IModalStyleProps): IModalStyles => {
 
   const classNames = getGlobalClassNames(globalClassNames, theme);
 
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const win = useWindow();
```
xugao, line 35: instead of calling the hook here, we should add `windowInnerHeight?: number` style prop and pass `win?.innerHeight` from component.

xugao (approved):

packages/react-internal/src/components/Modal/Modal.base.tsx
```diff
@@ -209,29 +146,245 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
+      internalState.lastSetXCoordinate = 0;
+      internalState.lastSetYCoordinate = 0;
+
+      setXCoordinate(0);
+      setYCoordinate(0);
+      setModalMenuClose();
+      setIsModalOpen(false);
+      setKeyboardMoveModeFalse();
+
+      if (dragOptions && internalState.hasRegisteredKeyUp) {
```
ecraig12345, line 238: For cases like this where you only care about the *presence* of dragOptions (and not any of its current values), you could optionally get rid of the dep on the whole object (which might mutate frequently) by doing `const hasDragOptions = !!dragOptions` at the outer scope and then using that as the dependency in the callback. Not sure how important that is though.

ecraig12345 (commented):
Really sorry for the delayed feedback here! This was hard to review because it's a big change and there was a lot of weirdness in the original code... I'd also checked out the branch and started trying to fix some things a couple weeks ago, but it turned into a giant cascade of issues (including what should have been fixes somehow breaking more things) and then got pushed to the back burner by the snap work. 

I've left some comments on specific parts, as well as the section of general comments just below, BUT given time constraints with the release I'd probably be okay with addressing these later if the component seems to work(ish) without them, then follow up shortly after with fixes.

Potentially important issues:
- I'm not sure that the logic of some of the old lifecycles (componentWillReceiveProps/componentDidMount/componentDidUpdate) is completely replicated. At least for mount/unmount, using the new hooks may help.
- Compared to the previous behavior, a different value for is open (props vs state) is being used in some places. It might be a good idea to change the names of `isOpen` and `isModalOpen` to clarify which one is from where and which takes precedence (which would also make it easier to see the changes to which is used when).

Opportunities for optimization (FC-specific) or fixing weirdness with existing code:
- In several places (I commented on one), you have effects or callbacks with dependencies on state values, where the state value is only used when updating. You can get rid of these dependencies by using setter callbacks which get the previous value as an argument.
- Even in the old version, the component was not handling entering/exiting keyboard move mode and cleaning up those event handlers in a symmetrical manner. Keyboard move mode should always be exited/cleaned up when closing modal *and* when opening the menu. One thing that could help with this is adding an `exitKeyboardMoveMode` helper method which does *all* the appropriate things and can be called wherever needed.
- Lots of weirdness with handleKeyDown...
  - If we're properly disposing handleKeyDown (see above), it should ONLY be active while in keyboard move mode (which implies the modal menu is closed), so the cases for if the menu is open or it's not keyboard move mode are unnecessary.
  - First check for escape/enter that stops event is redundant with `if (handledEvent)` later
  - Eliminate switch fallthrough: it appears intentional but was unnecessary and confusing
  - Should call full exitKeyboardMoveMode handler on escape/enter to ensure proper/full cleanup

packages/react-internal/etc/react-internal.api.md
```diff
@@ -3501,6 +3478,7 @@ export type IModalStyleProps = Required<Pick<IModalProps, 'theme'>> & Pick<IModa
     modalRectangleTop?: number;
     layerClassName?: string;
     isDefaultDragHandle?: boolean;
+    windowInnerHeight?: number;
```
ecraig12345, line 3481: Was this supposed to end up in the public interface?

czearing:
@ecraig12345 Thanks for the awesome feedback!

ecraig12345:
Pushed some updates doing the following:
- Combine storage of X and Y coordinates, because it's generally cleaner this way and reduces potential for multiple renders if updates aren't batched (e704b45674d1556abe656cd8f63dfe7bf233c60d)
- Fix keydown disposal and add similar symmetric disposal logic for keyup (e2964ba05e2e0016e4376024dd30be12f76245de)
  - Also remove separate `hasRegisteredKeyUp` because we can now track that based on the presence of `disposeOnKeyUp`
- Use `getPropsWithDefaults` (35c97dbde49dba8392b5316abd2adb5e3e64a22c). This was partly as prep for trying to split the component logic into modular hooks, which I decided not to do here, but the change doesn't hurt.
- Fix clamping and improve logic for accessing the content FocusTrapZone (0373c932fbf401287ee376a6f2bb421a94ddc33f)
  - Clamping was broken in this PR due to math errors (using the `minClampedPosition` for both comparisons)
  - Fix clamping logic to use `DOMRect`'s `left` and `top` (not `x` and `y`) for IE 11 compat (this was a bug in the original clamping implementation)
  - Add a "keep in bounds" toggle to the example to test clamping more easily
  - Ensure logic for setting up `topOffsetFixed` runs after the component is rendered
  - Use `focusTrapZoneId` as the actual `id` not `data-id` (this was only added recently in the clamping change, and I see no reason for using a data attribute here)
  - Instead of querying for the FocusTrapZone from the document, pass a `ref` to the element now that it's a function component which supports that (this is more reliable for multiple reasons; also looks like @joschect requested that change in the original PR but FTZ hadn't been converted yet)
- Remove separate `useMount` call because the general `useEffect` call for handling visibility will also run on mount (532b4a91eee008b1782fc8e8a509b23568d39288)

packages/react-internal/src/components/Modal/Modal.base.tsx
```diff
@@ -177,63 +107,313 @@ export class ModalBase extends React.Component<IModalProps, IDialogState> implem
...
+          registerForKeyUp();
+        }
+
+        internalState.hasBeenOpened = true;
+        setIsVisible(true);
+      }
+
+      // Closing the dialog
+      if (!isOpen && isModalOpen) {
+        internalState.onModalCloseTimer = setTimeout(handleModalClose, parseFloat(animationDuration) * 1000);
```
joschect, line 388: Should this also be on a `requestAnimationFrame`?

joschect (approved):
Approved with 1 comment

packages/react-internal/src/components/Modal/Modal.base.tsx
ecraig12345: Discussed and we decided to leave this as-is for now.

ecraig12345:
Somehow this seems to have caused a change in a Callout screener scenario, which doesn't make sense but appears to be unique to this PR... It's hard to see in the UI but the positioning of the callout moved down slightly (look at the position of the beak relative to the callout edge or the text inside the callout).
image image

PR #15572, opened 2020-10-17
Fix check for browser support of ResizeObserver in Pivot

behowell:
#### Pull request checklist

#### Description of changes

The check to see if the browser supports `ResizeObserver` was causing an error message when used with the `@testing-library/react` framework. This updates the check to avoid the error.

Also add a snapshot test for a Pivot that uses overflow behavior.


ecraig12345 (approved):

ecraig12345:
Looks like the snapshot is out of date--maybe an implicit conflict with Caleb's change earlier today?

PR #15576, opened 2020-10-18
Button: Correctly exporting old Button as part of the compat layer and new Button as the default in @fluentui/react.

### LINKED ISSUE #15562: Move compat Button exports to export correctly
Move the `compat` `Button` exports to the correct place and replace the regular exports with new `Button`.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15562 (also related to #15152)

#### Description of changes

This PR does a couple of things:
- Moves `Button` in `@fluentui/react-internal` to be exported as part of the `compat` folder.
- Changes the default `Button` export in `@fluentui/react` to export from `@fluentui/react-button` instead of from `@fluentui/react-internal`.
- Adds `docCategory Button` to all the type files in `@fluentui/react-button` so that they can be picked up by `api-docs`.
- Updates all references to the old `Button` so that they now import from the `compat` folder.

#### Focus areas to test

(optional)


ecraig12345:
One overall comment, I think to avoid confusing the two buttons we should switch old Button's docCategory to CompatButton or something like that (and add code on the website to change the display name to something nicer). Could either do that now or later.

ecraig12345:
Also we should add a compat API Extractor config to react-internal.

ecraig12345 (approved):
Mostly looks good, just see the two comments above. 

Also this is pretty cool! :tada:
image

Ridvanovskyy (approved):

packages/api-docs/config/api-docs.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
     path.resolve(__dirname, '../../date-time-utilities/dist/date-time-utilities.api.json'),
     path.resolve(__dirname, '../../merge-styles/dist/merge-styles.api.json'),
     path.resolve(__dirname, '../../react/dist/react.api.json'),
+    path.resolve(__dirname, '../../react-button/dist/react-button.api.json'),
```
ecraig12345, line 16: You'll probably also need to add an entry for `react-internal-compat.api.json` to get the compat button stuff to show up

khmakoto: Talked with @ecraig12345 offline. We'll address in a separate PR. Made an issue tracking this in #15600.

PR #15584, opened 2020-10-19
ContextualMenu: Remove unneeded resize dismiss check as callout handles it

### LINKED ISSUE #15369: ContextualMenu dismisses on window resize even with "calloutProps={{ preventDismissOnResize: true }}" set
ContextualMenu dismisses on window resize even with `calloutProps={{ preventDismissOnResize: true }}` set

### Environment Information

- **Package version(s)**: ^7.135.0

https://codepen.io/kestill/pen/wvWwdEE

#### Actual behavior:

ContextualMenu will be dismissed when the window is resized

#### Expected behavior:

ContextualMenu should stay open when the window is resized

Had a chat with @joschect, seems like the props are being passed to the underlying Callout as expected, but the ContextualMenu has a resize event of its own that is still dismissing the menu on window resize.

joschect:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #15369

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)


khmakoto (approved):
From the looks of it seems like this is a change by design, right? Anyways, it would probably be good to have some text explaining why this is done in the PR description instead of having to go to the issue.

PR #15587, opened 2020-10-19
Remove mdl2-theme package from master

xugao:
#### Pull request checklist

#### Description of changes
Remove `@uifabric/mdl2-theme` from master (it will be maintained and published from `7.0` branch). 

Related issue: #13384

#### Focus areas to test

(optional)


yarn.lock
```diff
@@ -1384,6 +1384,16 @@
     unique-filename "^1.1.1"
     which "^1.3.1"
 
+"@fluentui/theme@^1.5.1":
```
ecraig12345, line 1387: We need to ensure we're specifying dependencies in a way that doesn't accidentally pull in extra versions of packages that exist in master... May need to mess with `resolutions` in `package.json`? Or ideally figure out how to fix it "properly."

yarn.lock
```diff
@@ -4739,6 +4749,49 @@
...
+  integrity sha512-jj5FH7XuQXz8HgRhCXNApeUAynDBv2nfmTpzHawE55x7YR+oliQfsJSY8Ow4YbLAazIR5BuW1n9eKuGxxmIw1w==
+  dependencies:
+    "@fluentui/theme" "^1.5.1"
+    "@microsoft/load-themed-styles" "^1.10.26"
+    "@uifabric/merge-styles" "^7.19.1"
+    "@uifabric/set-version" "^7.0.23"
+    "@uifabric/utilities" "^7.32.4"
+    tslib "^1.10.0"
+
+"@uifabric/utilities@^7.32.4":
```
ecraig12345, line 4775: This one is bad too

yarn.lock
ecraig12345: Same issue for styling and utilities. I think we should try to find a "proper" fix for this if possible since it's an issue our consumers will hit too. (Or at least figure out some guidance which we use internally and share out.)

ecraig12345 (approved):

andrewchoi5 (approved):

PR #15589, opened 2020-10-19
Disable marquee selection undesirable in UPP

sjadelson:
#### Pull request checklist

#### Description of changes

MarqueeSelection is not the desired mouse selection paradigm for UPP. It also "hijacks" mouse selection in the message body.  This CL will remove MS in UPP, giving us the same experience as shipped OWA - no drag mouse selection.  We will give a better experience in a future checkin.

#### Focus areas to test

(optional)


packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -351,45 +343,43 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
-                    onPaste={_onPaste}
-                    onKeyDown={_onInputKeyDown}
-                    /* eslint-enable react/jsx-no-bind */
-                  />
-                </div>
-              )}
-            </div>
-          </SelectionZone>
-        </MarqueeSelection>
+        <SelectionZone selection={selection} selectionMode={SelectionMode.multiple}>
```
nebhatna, line 346: Could you please a TODO here so we dont lose track of things which we need to do like mouse selection etc.

sjadelson: Done

elisabethcvs (approved):

PR #15597, opened 2020-10-19
Removing unused disabled eslint rules from react and react-internal packages

czearing:
#### Pull request checklist

#### Description of changes

Removing unused disabled eslint rules from `react` and `react-internal` packages.


ecraig12345 (approved):
Thanks!

PR #15603, opened 2020-10-20
Rename @uifabric/icons to @fluentui/font-icons-mdl2

tomi-msft:
#### Pull request checklist

#### Description of changes

Rename @uifabric/icons to @fluentui/font-icons-mdl2

#### Focus areas to test

(optional)


apps/vr-tests/src/stories/Icon.stories.tsx
```diff
@@ -3,7 +3,7 @@ import Screener from 'screener-storybook/src/screener';
 import { storiesOf } from '@storybook/react';
 import { FabricDecorator } from '../utilities/index';
 import { Icon, IconType, getIconClassName, Fabric } from '@fluentui/react';
-import * as IconNames from '../../../../packages/icons/src/IconNames';
+import * as IconNames from '../../../../packages/font-icons-mdl2/src/IconNames';
```
ecraig12345, line 6: Can you change this to a package name import instead? (and add this package as a dep of vr-tests if needed)

change/@fluentui-react-checkbox-2020-10-19-17-41-22-icon-rename.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "minor",
```
ecraig12345, line 2: Most of these should probably be patch, not minor (though I'm not sure it matters in this case since we may not end up using all these changelogs as-is)

packages/react-examples/src/react/Icon/docs/IconOverview.md
```diff
@@ -2,4 +2,4 @@ In a user interface, an icon is an image that represents an application, a capab
 
 For a list of icons, visit our [icon documentation](#/styles/web/icons).
 
-Note that icons are not bundled by default and typically must be loaded by calling `initializeIcons` from the `@uifabric/icons` package at the root of your application. See the [icon documentation](#/styles/web/icons#fabric-react) for more details.
+Note that icons are not bundled by default and typically must be loaded by calling `initializeIcons` from the `@fluentui/font-icons-mdl2` package at the root of your application. See the [icon documentation](#/styles/web/icons#fabric-react) for more details.
```
ecraig12345, line 5: Not a new issue but we should probably discourage people from referencing the package directly
```suggestion
Note that icons are not bundled by default and typically must be loaded by calling `initializeIcons` from `@fluentui/react/lib/Icons` at the root of your application. See the [icon documentation](#/styles/web/icons#fabric-react) for more details.
```

ecraig12345:
Looks like you need an API update in react-internal.

ecraig12345 (approved):

PR #15605, opened 2020-10-20
Fix propagation of items and groups props from DetailsList through GroupedList

khmakoto:
#### Pull request checklist

#### Description of changes

_Cherry-pick of #15321._

_Original PR description:_

Fixed an issue in the recent rework to use `getDerivedStateFromProps` in the `GroupedList` component: the new function was inadvertently checking new input props values *against themselves* rather than the previous state. This issue is now fixed, and the `DetailsList` input state now propagates correctly through `GroupedList`.

Added some unit tests for `DetailsList` explicitly in order to validate that re-rendering a `DetailsList` with new inputs for `items` or `groups` properly updates for each type of transition.

#### Focus areas to test

Validate `DetailsList` when changing the `items` or `groups` props. Added a unit test explicitly for this.


dzearing (approved):

khmakoto:

PR #15612, opened 2020-10-20
Update postinstall.js

LavaAdventures:


LavaAdventures:
It is so if someone does not know what it does some how.

ecraig12345:
Sorry, but `console.log` is very a very standard, commonly-used JS method, so we're not going to add comments explaining what it does. The `chalk` part is a little less common, but easy to figure out what it does (adding colors) from a quick search online.

PR #15615, opened 2020-10-20
Rename build-related packages

ecraig12345:
Part of #13384, moving things to fluentui scope:
- Rename `@uifabric/build` to `@fluentui/scripts` (also matches the folder name)
- Rename `@uifabric/webpack-utils` to `@fluentui/webpack-utilities`
- Rename `@uifabric/test-utilities` to `@fluentui/test-utilities`
- Rename `@uifabric/jest-serializer-merge-styles` to `@fluentui/jest-serializer-merge-styles`

Also since I was already touching the entire world of just configs, I converted them to use actual TS import syntax and converted the task files to be TS not JS.


tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
```
xugao, line 1: what's this root tsconfig for?

xugao (approved):

tsconfig.json
ecraig12345: Better type resolution while editing just.config.ts throughout the repo

PR #15619, opened 2020-10-20
UnifiedPicker: Fixing the selection for the input text

nebhatna:
#### Pull request checklist

#### Description of changes
Selection was not working earlier for the input entered. This PR fixes that. Just updated one example so that we have other examples depicting paste usage too.

#### Focus areas to test

(optional)


packages/react-examples/src/react-experiments/UnifiedPeoplePicker/DoubleUnifiedPeoplePicker.Example.tsx
```diff
@@ -179,14 +159,54 @@ const UnifiedPeoplePickerExample = (): JSX.Element => {
...
+      .split(SEPARATOR_REGEX)
+      .map((token: string) => token.trim())
+      .filter((token: string) => !!token);
+
+  const _onInputChange = (filterText: string, composing?: boolean, resultItemsList?: IPersonaProps[]): void => {
     const allPeople = people;
+
+    const lastCharIndex = filterText.length - 1;
+    const lastChar = filterText[lastCharIndex];
     const suggestions = allPeople.filter((item: IPersonaProps) => _startsWith(item.text || '', filterText));
```
elisabethcvs, line 174: I think these can be in the else condition? They're not used in the if right?

packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -270,12 +270,25 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
     }
   };
-  const _onInputChange = (value: string, composing?: boolean) => {
+  const _onInputChange = (value: string, composing?: boolean, resultItemsList?: T[]) => {
     if (!composing) {
       // update query string
       setQueryString(value);
       !isSuggestionsShown ? showPicker(true) : null;
-      onInputChange ? onInputChange(value) : null;
+      if (!resultItemsList) {
```
elisabethcvs, line 278: I think this will always run? I think you can just declare it here so the prop can fill it in.

elisabethcvs (changes requested):
Made a few comments

packages/react-examples/src/react-experiments/UnifiedPeoplePicker/DoubleUnifiedPeoplePicker.Example.tsx
nebhatna: it is used for the if check :) i will move it closer to its usage

packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
nebhatna: i am not sure i understand.. what line are talking about :) ? we want to initialize the resiltItmsList to be an array in case it is not defined.

elisabethcvs: Correct. What I was trying to say is that I think it will always be undefined, so this clause will always run. But TBH it's fine either way

elisabethcvs (approved):

nebhatna:
Closing this as this got messed up with a merge.

PR #15620, opened 2020-10-20
Update package versions for @fluentui/react version 8 beta release

xugao:
#### Pull request checklist

#### Description of changes
- Major bump version for public packages related to `@fluentui/react` version 8
  - Update package version and all package dependencies (expect for northstar package dependencies).
  - Update existing change logs for v8 beta packages to have `"type": "prerelease"`.
  - sync non-beta packages with latest versions
- Reset version to `1.0.0` for private packages which has current version greater than `1.0.0`
- Reset couple packages from public to private: `react-conformance` and `storybook`
- Update beachball configs
  - by default, repo packages are only allowed to have change types `prerelease` or `none` and `tag` is set to be `beta`
  - for non beta package, I added `beachball` config in `package.json` to overwrite default beachball config (no behavior change compared to current) and set correct npm `tag`
- Added `bump:beachball` script in root `package.json` (useful for testing beachball related changes)
- Bumped beachball to latest, which includes [this fix](https://github.com/microsoft/beachball/pull/428) and [this fix](https://github.com/microsoft/beachball/pull/430)

#### Focus areas to test
run `yarn bump:beachball`


apps/codesandbox-react-northstar-template/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "codesandbox-react-northstar-template",
   "description": "Template for creating codesandbox environment for @fluentui/react-northstar packages.",
-  "version": "7.0.0",
+  "version": "1.0.0",
```
xugao, line 4: "reset" the package version to be `1.0.0` because
1. it's not related to fluentui/react
2. it's private / never published

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fluent-ui-react-repo",
-  "version": "7.0.0",
+  "version": "8.0.0-beta",
```
xugao, line 3: should we "reset" this to `1.0.0`? i don't think this version matters

packages/react-next/package.json
```diff
@@ -1,14 +1,14 @@
...
+  "version": "9.0.0-alpha",
   "beachball": {
     "disallowedChangeTypes": [
       "patch",
       "minor",
       "major"
     ]
   },
-  "description": "Next major release staging package: Reusable React components for building web experiences.",
+  "description": "Next majostr release staging package: Reusable React components for building web experiences.",
```
xugao, line 11:
```suggestion
  "description": "Next major release staging package: Reusable React components for building web experiences.",
```

packages/react-window-provider/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui/react-window-provider",
-  "version": "0.3.3",
+  "version": "2.0.0-beta",
```
xugao, line 3: bumping this one by 2 because we should have this versioned to be `1.0.0` to begin with since it's used in production. I will update this to `1.0.0` in `7.0.0` branch separately

apps/a11y-tests/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "a11y-tests",
-  "version": "7.0.0",
+  "version": "8.0.0-beta",
```
ecraig12345, line 3: For the unpublished packages do we even need to bother with the -beta, or can we just put them straight to 8? Or possibly take back to 1 if that makes more sense.

xugao: i intended to have `8.0.0-beta` everywhere without any `8.0.0` to avoid confusion. i think i will go with reseting `private` package version to `1.0.0` as you suggested :)

change/@fluentui-azure-themes-2020-10-09-16-58-10-rename-themes.json
```diff
@@ -1,5 +1,5 @@
 {
-  "type": "patch",
+  "type": "prerelease",
```
ecraig12345, line 2: Do we need to change dependentChangeType here? 

Also if we only allow `prerelease` in master we should update the cherry-picking instructions to say to delete and re-generate the change files.

xugao: yes we do need this. for example, `1.0.0-beta` with a new `patch` will be bumped to `1.0.0`. 
good call, for cherry-picking, the change files would require a bit more work.

ecraig12345: Makes sense. We may need to do another pass of fixing change types and package names in change files before publishing (and maybe again each day for a couple days) since there are likely open PRs with outdated change files.

package.json
```diff
@@ -129,6 +130,24 @@
...
+          "@fluentui/accessibility",
+          "@fluentui/docs",
+          "@fluentui/react-bindings",
+          "@fluentui/react-northstar",
+          "@fluentui/perf-test",
+          "@fluentui/e2e"
+        ],
+        "dependencies": [
+          "@uifabric/utilities",
+          "@fluentui/keyboard-key",
```
ecraig12345, line 145: I'm a bit nervous about this, especially the packages that have the same names.
1. Possibility of collisions such that other packages accidentally pull the old version not the local version. We may need some more `nohoist` settings to deal with this, though unfortunately that feature is super undocumented and may take some trial and error.
2. Even the packages with old names, it makes it easier for PRs using the old name to sneak in (hopefully it would be a lint error though)

xugao: >1. Possibility of collisions such that other packages accidentally pull the old version not the local version. We may need some more nohoist settings to deal with this, though unfortunately that feature is super undocumented and may take some trial and error.

This will not happen because non-northstar packages will be using latest versions of internal packages as package dependencies. for northstar packages, they will be using the packages from local `node_modules`. I also double checked locally. 

>2. Even the packages with old names, it makes it easier for PRs using the old name to sneak in (hopefully it would be a lint error though)

I just double checked locally using `@uifabric/merge-styles` / `@fluentui/merge-styles`.

VSCode is suggesting the correct imports:
image

Lint rule will also catch:
image

**please help double check the dev experience though, in case I missed anything**

packages/react-conformance/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui/react-conformance",
-  "version": "0.1.5",
+  "version": "1.0.0-beta",
```
ecraig12345, line 3: Not sure this one is quite at version 1 yet either...but also I don't know how likely it is that anyone would be using it

packages/react-monaco-editor/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui/react-monaco-editor",
-  "version": "0.13.18",
+  "version": "1.0.0-beta",
```
ecraig12345, line 3: This one I'm also less enthusiastic about bumping to 1.0, but it's probably unavoidable

packages/storybook/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui/storybook",
-  "version": "0.5.1",
+  "version": "1.0.0-beta",
```
ecraig12345, line 3: Does this one actually need to be published?

packages/ie11-polyfills/package.json
```diff
@@ -11,5 +11,11 @@
   "devDependencies": {
     "ie11-custom-properties": "4.1.0"
   },
-  "license": "MIT"
+  "license": "MIT",
+  "beachball": {
+    "tag": "latest",
+    "disallowedChangeTypes": [
```
ecraig12345, line 17: Bit confused, why does this one get a special config?

packages/react-avatar/package.json
```diff
@@ -47,18 +47,24 @@
...
   },
   "peerDependencies": {
     "@types/react": ">=16.8.0 <17.0.0",
     "@types/react-dom": ">=16.8.0 <17.0.0",
     "react": ">=16.8.0 <17.0.0",
     "react-dom": ">=16.8.0 <17.0.0"
+  },
+  "beachball": {
+    "tag": "latest",
+    "disallowedChangeTypes": [
```
ecraig12345, line 66: Same question here?

packages/react-next/package.json
```diff
@@ -1,14 +1,15 @@
...
   "beachball": {
+    "tag": "alpha",
     "disallowedChangeTypes": [
       "patch",
       "minor",
       "major"
     ]
   },
-  "description": "Next major release staging package: Reusable React components for building web experiences.",
+  "description": "Next majostr release staging package: Reusable React components for building web experiences.",
```
ecraig12345, line 12:
```suggestion
  "description": "Next major release staging package: Reusable React components for building web experiences.",
```

change/@fluentui-azure-themes-2020-10-09-16-58-10-rename-themes.json
xugao: `yarn checkchange` will break after my change for change files which are not set to `prerelease` type.

packages/react-compose/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fluentui/react-compose",
-  "version": "0.19.4",
+  "version": "1.0.0-beta",
```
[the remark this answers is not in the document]
xugao: we should bump since we are following the rule of in-production packages should be versioned greater than `1.0.0`

packages/ie11-polyfills/package.json
xugao: I updated repo-level beachball config. 
see "updated beachball config" section in my PR description :)

packages/react-compose/package.json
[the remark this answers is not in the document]
xugao: similar reason as why this PR was made: https://github.com/microsoft/fluentui/pull/15174

packages/storybook/package.json
xugao: i will change the package to `private`. we should only publish package which has clear reason

packages/react-compose/package.json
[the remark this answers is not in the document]
xugao: reverted for now. this bump doesn't block `beta`. i will confirm northstar folks on if we can safely bump `react-compose` and `keyboard-key` to `v1`

ecraig12345 (approved):
:tada: :tada: :tada:

package.json
```diff
@@ -129,6 +130,18 @@
...
           "webpack"
         ]
+      },
+      {
+        "packages": [
+          "@fluentui/accessibility",
+          "@fluentui/react-bindings",
+          "@fluentui/react-northstar"
+        ],
+        "dependencies": [
```
xugao, line 140: there are bunch missing pieces related to this change around webpack/jest/rollup configs. Elizabeth is addressing in PR #15644

PR #15623, opened 2020-10-20
UPP: Add onKeyDown prop

elisabethcvs:
#### Pull request checklist

#### Description of changes

Add a prop for onKeyDown so the consumer has the ability to modify that

I did test that this works by adding the overload to an example, but didn't feel like it added to the example to check it in

### DIFF
--- a/packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
+++ b/packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
@@ -176,7 +176,13 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
     onDragEnd: _onDragEnd,
   };
 
-  const _onBackspace = (ev: React.KeyboardEvent<HTMLDivElement>) => {
+  const _onKeyDown = (ev: React.KeyboardEvent<HTMLDivElement>) => {
+    // Allow the caller to handle the key down
+    if (props.onKeyDown) {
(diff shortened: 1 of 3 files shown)


packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -176,7 +176,13 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
...
-  const _onBackspace = (ev: React.KeyboardEvent<HTMLDivElement>) => {
+  const _onKeyDown = (ev: React.KeyboardEvent<HTMLDivElement>) => {
+    // Allow the caller to handle the key down
+    if (props.onKeyDown) {
+      props.onKeyDown(ev);
+    }
+
+    // Handle delete of items via backspace
     if (ev.which !== KeyCodes.backspace) {
       return;
```
nebhatna, line 187: i think rather than doing this, since now we are working on keydown instead of just backspace, we should check if the keyboard event is backspace then we remove the elements. rather than returning if the keyboard event is not backspace.

nebhatna (changes requested):
comment :)

packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
elisabethcvs: Good point

nebhatna (approved):

PR #15631, opened 2020-10-21
feat(UI builder): Add ability to compute narration for entering of landmarks and groups

adamsamec:
#### Description of changes
Modifies the screen reader narration computation feature  of the UI builder by adding the ability to compute and display narration when entering landmark or group elements or roles for components and their elements  that are added to the canvas and selected (in the Build and Design mode), or that are focused (in the Use mode). Specifically, these landmarks and group elements and roles are supported:

* Landmark elements: aside, footer, header, main, nav.
* Landmark roles: banner, complementary, contentinfo, form, main, navigation, region, search.
* Section elements: article, ol, ul.
* Section roles: article, list.
* Group and composite roles: grid, listbox, menu, radiogroup, tablist, toolbar, tree.


packages/fluentui/react-builder/src/components/ReaderNarration.tsx
```diff
@@ -1,12 +1,13 @@
...
-let narrationPath = null;
-let narrationPaths = [];
-let narrationTexts = {};
+const computer = new NarrationComputer();
+let prevSelector = null;
+let focusableElements = {};
+let elementsPaths = [];
+let selectedElementPath = null;
+let prevNarrationElement = null;
 const aomMissing = !(window as any).getComputedAccessibleNode;
```
assuncaocharles, line 11: _nit_
```suggestion
const aomMissing = window.hasOwnProperty('getComputedAccessibleNode');
```

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
```diff
@@ -16,98 +17,119 @@ export type ReaderNarrationProps = {
...
-  }, []); // End computeAndSave
-
   // Sets the complete screen reader narration text to be displayed.
   const setCompleteText = text => {
     setNarrationText(text !== null ? `Narration: ${text}` : null);
   }; // End setCompleteText
 
-  // Handles the "focusin" event by computing and saving the narration paths and texts.
+  // Handles the element path dropdown change event by updating the current and previous narration elements.
+  const handleElementPathChange = (event: any, props: DropdownProps) => {
```
assuncaocharles, line 29:
```suggestion
  const handleElementPathChange = (event: Event, props: DropdownProps) => {
```

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
```diff
@@ -23,14 +35,186 @@ export class NarrationComputer {
...
+    return focusableElements;
+  } // End getFocusableElements
+
+  // Recursively traverses the DOM tree rooted at the given element to find all the parents which have the ariaActiveDescendantElement property set, and saves the found elements to the given parents array.
+  findActiveDescendantsParents(element: IAriaElement, parents: IAriaElement[]) {
+    if (element.ariaActiveDescendantElement != null) {
+      // Begin if 1
+      parents.push(element);
+    } // End if 1
+    for (let i = 0; i < element.children.length; i++) {
```
assuncaocharles, line 58: `.forEach` usage?

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
```diff
@@ -23,14 +35,186 @@ export class NarrationComputer {
...
+      const focusableElement: FocusableElement = {
+        path: newPath,
+        element,
+      };
+      focusableElements.push(focusableElement);
+      // return;
+    } // End if 1
+
+    // Traverse down to all the children
+    for (let i = 0; i < element.children.length; i++) {
```
assuncaocharles, line 100: `.forEach`?

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
adamsamec: If I apply your suggestion, I get: `TS2322: Type '(event: Event, props: DropdownProps) => void' is not assignable to type '(event: MouseEvent<Element, MouseEvent> | KeyboardEvent<Element>, data: DropdownProps) => void'.`

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
adamsamec: It's an HTMLCollection array-like object which cannot be iterated using forEach().

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
adamsamec: Again, HTMLCollection has no forEach() method.

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
ling1726: shouldn't this be one of the events inherited from React.SyntheticEvent ? so MouseEvent if it's a click

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
```diff
@@ -16,98 +17,119 @@ export type ReaderNarrationProps = {
 
 export const ReaderNarration: React.FunctionComponent<ReaderNarrationProps> = ({ selector, inUseMode }) => {
   const ref = React.useRef<HTMLElement>();
-  const [selectedNarrationPath, setSelectedNarrationPath] = React.useState(null);
+  const [narrationElement, setNarrationElement] = React.useState(null);
```
ling1726, line 20: Is there a type you can assign here ? otherwise it's a very confusing implicit any

```suggestion
  const [narrationElement, setNarrationElement] = React.useState<type>(null);
```

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
adamsamec: Neither MouseEvent works.

adamsamec: However, `const handleElementPathChange = (event: MouseEvent | KeybordEvent, props: DropdownProps)` wowrks. Should I leave it like that?

adamsamec: I am sorry, MouseEvent | KeyboardEvent does not work either.

ling1726: In that case as we discussed offline React.SyntheticEvent will work since it's the parent of all react events... not very clean but I guess better than any :)

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
```diff
@@ -16,98 +17,119 @@ export type ReaderNarrationProps = {
...
+          let preselectedElementItem = focusableElementsItems[0];
+          focusableElementsItems.some(focusableElementItem => {
+            // Begin some 1
+            if (focusableElementItem.element.tabIndex >= 0) {
+              // Begin if 3
+              preselectedElementItem = focusableElementItem;
+              return true;
+            } // End if 3
+            return false;
+          }); // End some 1
```
ling1726, line 76: You generally shouldn't be using `some` and `every` for side effects. Especially when the boolean they return has no meaning in the code

Do you need to support IE11 ? this might be simpler

```suggestion
        preselectedItem = focusableElementItem.find(item => item.element.tabIndex >=0) || preselectedItem
```

ling1726, line 76: You generally shouldn't be using some and every for side effects. Especially when the boolean they return has no meaning in the code

Do you need to support IE11 ? this might be simpler

```suggestion
         
          let preselectedElementItem = focusableElementItems.find(item => item.element.tabIndex >=0) ||  focusableElementsItems[0];
```

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
adamsamec: Fixed to `React.useState<IAriaElement>`.

packages/fluentui/react-builder/src/components/ReaderNarration.tsx
adamsamec: Thanks, that's a good one.

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
```diff
@@ -23,14 +35,186 @@ export class NarrationComputer {
...
+      } // End while 2
+      // If the previous element is an ancestor of the element, return
+      if (elementParent === prevElement) {
+        // Begin if 1
+        return landmarksAndGroups;
+      } // End if 1
+      prevElementParent = prevElement.parentElement;
+      elementParent = elementParent.parentElement;
+    } // End while 1
+    commonAncestor = elementParent;
```
ling1726, line 134: Maybe this could be abstracted to be a function, at I think it's doing here is traversing up the tree to return the common ancestor... maybe a function called `getCommonAncestor(elementParent, prevElementParent, element)` ?

It will also make this code easier to read

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
```diff
@@ -23,14 +35,186 @@ export class NarrationComputer {
...
     let definitionName = this.getDefinitionName(element, platform, 'stateRules');
 
     // Retrieve the computed accessible node
     const node = await (window as any).getComputedAccessibleNode(element);
 
     // Compute and store all the narration parts
+    this.computeLandmarksAndGroups(element, prevElement, platform);
     this.computeUsage(definitionName, element, platform);
     this.computeDescription(definitionName, element, platform);
     this.computeNameAndTitle(node, element);
```
ling1726, line 220: You don't have to do this, just wondering what you think of moving the logic for each of these `this.compte<>` into seperate classes that encapsulate their behaviour, so the narration computer just has to instantiate each of the sub classes to use their behaviour. 

Doesn't need to be sub classes, you can probably just extract them to utility functions, so this class stays smaller.

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
adamsamec: Agree, I have extracted the function.

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
adamsamec: I have replaced it with Array.from().

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
adamsamec: Tried to replace it with Array.from(element.children).forEach(async (child: IAriaElement) => {...}); Although it does not complain at compile time, it does not work in runtime. Giving me some errors wihch seems unrelated to this problem.

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
adamsamec: After the Teams conversation, resolved to keep it as it is.

packages/fluentui/react-builder/src/narration/NarrationComputer.tsx
adamsamec: After a Teams meeting, we decided to keep it using the for statement.

jurokapsiar (approved):

PR #15632, opened 2020-10-21
7.0 - Remove MarqueeSelection from UnifiedPicker

sjadelson:
#### Pull request checklist

#### Description of changes

This change is a backport from V8, to support fixing an important bug in an app still using v7.  This change removes the MarqueeSelection (rubberband) selection from UnifiedPicker, which is not the right paradigm for mouse drag-selection in this control.  A proper drag-selection will be implemented in v8.  This change is minimal for v7, as having no drag-selection is better than the wrong one.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
+++ b/packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
@@ -8,7 +8,7 @@ import {
   FocusZone,
   SelectionZone,
   Autofill,
-in  IInputProps,
+  IInputProps,
   IDragDropEvents,
 } from 'office-ui-fabric-react';
 import { IUnifiedPickerProps } from './UnifiedPicker.types';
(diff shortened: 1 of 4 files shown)


packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -8,8 +8,7 @@ import {
   FocusZone,
   SelectionZone,
   Autofill,
-  IInputProps,
-  MarqueeSelection,
+in  IInputProps,
```
elisabethcvs, line 11: extra in, looks like a typo

elisabethcvs (changes requested):
looks like you've got some extra characters in there

elisabethcvs (approved):

packages/experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
sjadelson: Yes, saw that an fixed now.  Focus issues.

PR #15634, opened 2020-10-21
Remove unneeded peer deps on react-dom

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Part of #7160

#### Description of changes

Now that we've removed findDOMNode there's no reason for ~~most~~ some of our packages to have peer dependencies on react-dom.

EDIT: Turns out `Layer` uses `ReactDOM.createPortal`, so unfortunately we still need the peer dep for `react-internal` and everything above it. 😢


dzearing (approved):

behowell (approved):

ecraig12345:
Apparently we do still need react-dom for `@fluentui/react` (and anything that uses `@fluentui/react-internal`) because Layer uses `ReactDOM.createPortal`. :(

ecraig12345:
Updated with a lesser version of the change...only remove react-dom from packages *below* react-internal. (And move it to a normal dependency in packages which aren't published.)

PR #15644, opened 2020-10-22
Re-export date-time Calendar and DatePicker from @fluentui/react

### LINKED ISSUE #14679: Re-export date-time components from react

ecraig12345:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #14679

#### Description of changes

Re-export the `@uifabric/date-time` Calendar and DatePicker from `@fluentui/react`, and delete the old implementations. Plus some cleanup of interfaces and defaults (and actually sharing stuff) now that we're making this official.

#### In `date-time`

- Reuse shared types and defaults (especially for strings) from `date-time-utilities` instead of redefining everything
  - Leave old names as deprecated aliases
- Export a bunch of things from the root which were referenced by other parts of the public API
- Add missing function component type annotations and `React.RefAttributes` extension
- Add more `@docCategory` tags for the website, and minor doc wording cleanup
- Rename `ICalendarIconStrings` to `ICalendarNavigationIcons` ("strings" implies they're loc strings; old name available as alias)

#### In `date-time-utilities`

- Move some things around in the interfaces in `date-time-utilities` to isolate the `react-northstar`-specific stuff:
  - `ICalendarStrings` now contains the parts used by date-time and/or shared with northstar (and does not extend the callbacks interface `IDateFormatting`)
  - New `IDatepickerStrings` extends `ICalendarStrings` and contains the northstar-specific stuff
  - Same pattern for defaults
  - In `IDateFormatting`, remove extension of `IDateGridStrings` and move `parseDate` (northstar-specific) to `IDatepickerStrings`
- Add `docCategory` annotations to things that should show up on the website (and some minor doc wording cleanup)
- Properly export `TimeConstants` by name

#### In `@fluentui/react-internal`

- Delete old `Calendar` and `DatePicker`
- Delete re-export files of `dateMath`, `dateValues`, `timeValues` at old locations

#### In `@fluentui/react`

- Re-export `date-time`'s `Calendar` and `DatePicker`
- Add re-export files `DateTimeUtilities` and `WeeklyDayPicker` (mirroring `date-time`)
- Change the main index file to re-export directly from the other packages (not the individual component re-export files; this one is up for discussion)

#### In examples

- Remove unnecessary string overrides and other unnecessary explicit setting of props to the same value as the default

#### In northstar and scripts

Any northstar packages using `lerna-alias` to generate webpack or jest aliases was still pulling in local versions of `@fluentui/date-time-utilities` (instead of using the version from npm in node_modules), which wasn't obvious before, but some minor API changes made it show up in this PR. As a temporary workaround, add a custom `lernaAliasNorthstar` wrapper which has a hardcoded list of packages to NOT use the local versions.

#### Follow-ups

See #15646 for styling-related issues (only discovered once this change got to PR) which will need to be addressed ASAP but not right now.


scripts/lernaAliasNorthstar.js
```diff
@@ -0,0 +1,28 @@
+const lernaAlias = require('lerna-alias');
+
+// northstar packages should pull these from npm, not the repo
+const excludedPackages = ['@fluentui/date-time-utilities', '@fluentui/dom-utilities'];
```
xugao, line 4: also should exclude `'@uifabric/utilities'`

scripts/jest/index.js
```diff
@@ -1,4 +1,8 @@
-module.exports = {
+// @ts-check
+const lernaAlias = require('../lernaAliasNorthstar');
```
xugao, line 2: should we rename this to `northStar.js` or something to indicate more clearly this is used for northstar only?

packages/fluentui/react-builder/jest.config.js
```diff
@@ -1,5 +1,5 @@
-module.exports = {
-  ...require('@fluentui/scripts/jest'),
+const commonConfig = require('@fluentui/scripts/jest');
+
+module.exports = commonConfig({
   name: 'react-builder',
```
xugao, line 4: looks like this one has wrong `directory`

scripts/lernaAliasNorthstar.js
ecraig12345: That should happen automatically since the name has changed

scripts/jest/index.js
ecraig12345: Maybe, but we have this problem in various config files, so I don't consider it as important to fix right now. (though since I'm already changing all the configs using it, this might be a decent time)

xugao: ok; don't worry about it now

packages/fluentui/docs/tsconfig.json
```diff
@@ -8,10 +8,6 @@
       "@fluentui/dom-utilities": ["packages/dom-utilities/src"],
```
xugao, line 8: should remove this one too

packages/fluentui/docs/tsconfig.json
```diff
@@ -5,13 +5,8 @@
     "paths": {
       "@fluentui/*": ["packages/fluentui/*/src/index"],
       "@fluentui/a11y-testing": ["packages/a11y-testing/src/index"],
-      "@fluentui/dom-utilities": ["packages/dom-utilities/src"],
       "@fluentui/keyboard-key": ["packages/keyboard-key/src/index"],
       "@fluentui/react-compose": ["packages/react-compose/src/index"],
-      "@fluentui/merge-styles": ["packages/merge-styles/src"],
-      "@fluentui/set-version": ["packages/set-version/src"],
```
xugao, line 12: oh we need to keep `"@fluentui/set-version": ["packages/set-version/src"],`

ecraig12345:
The hovered day states still aren't working right in screener (will investigate).

Other non-blocking screener changes: 
- Padding around "Go to today" has changed and may be wrong--filed #15651 to address later
- Some very minor padding/margin, width, or text position changes (probably okay)

xugao (approved):

ecraig12345:
Hovering the day cells doesn't work in screener in RTL for some reason. I'm guessing this has to do with the fact that CalendarDayGrid relies on mouseIn events to add a class and apply hover styling (rather than applying it normally), and maybe the mouseIn doesn't get located/consumed/??? properly in RTL? Filed an issue #15657 to investigate but accepting the incorrect states for now to get this unblocked. ([This demo](https://codepen.io/ecraig12345/pen/VwjmOwv?editors=0010) confirms that it's a test-only issue.)

lorejoh12 (commented):
The only functionality bug with the examples is with the two externally controlled examples, "Calendar with week selection" and "Calendar with month selection" not updating the highlighted selected days as you click the previous and next buttons. However, this appears to be an existing issue, not related to this PR.

packages/date-time/src/components/Calendar/Calendar.base.tsx
```diff
@@ -322,7 +298,8 @@ export const CalendarBase = React.forwardRef(
       isDayPickerVisible: isDayPickerVisible,
       monthPickerOnly: monthPickerOnly,
       showMonthPickerAsOverlay: showMonthPickerAsOverlay,
-      overlayedWithButton: overlayedWithButton,
+      overlaidWithButton: overlaidWithButton,
+      overlayedWithButton: overlaidWithButton,
```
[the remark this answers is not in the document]
ecraig12345: Just fixing spelling.

ecraig12345:
@lorejoh12 Thanks for reviewing, and sorry about jumping the gun a bit with merging! I filed #15658 to follow up on the examples.

PR #15653, opened 2020-10-22
Updating Breadcrumb to a function component within react package

czearing:
#### Pull request checklist

#### Description of changes
1. Updating `Breadcrumb` to a function component with **react** package.
2. Removing `component-handles-ref` and `component-has-root-ref` from disabledTests in Breadcrumb.tests.tsx
3. Extending IBreadcrumbProps to ```React.RefAttributes<HTMLDivElement>```

#### Focus areas to test
`BreadCrumb`


packages/react/src/components/Breadcrumb/Breadcrumb.base.tsx
```diff
@@ -47,131 +41,139 @@ const nonActionableItemProps: Partial<IContextualMenuItemProps> = {
...
+  const reduceData = (data: IBreadcrumbData): IBreadcrumbData | undefined => {
+    const movedItem = renderedItems[data.props.overflowIndex!];
 
     if (!movedItem) {
       return undefined;
     }
 
-    renderedItems = [...renderedItems];
-    renderedItems.splice(overflowIndex!, 1);
+    data.renderedItems = [...data.renderedItems];
```
ecraig12345, line 74: You probably shouldn't be mutating the original `data` object. That's why the previous code was putting these back in at the end.

packages/react/src/components/Breadcrumb/Breadcrumb.base.tsx
```diff
@@ -47,131 +41,139 @@ const nonActionableItemProps: Partial<IContextualMenuItemProps> = {
...
     const movedItem = renderedOverflowItems.pop();
 
-    if (!movedItem || renderedItems.length >= maxDisplayedItems!) {
+    if (!movedItem || data.renderedItems.length >= data.props.maxDisplayedItems!) {
       return undefined;
     }
 
-    renderedItems = [...renderedItems];
-    renderedItems.splice(overflowIndex!, 0, movedItem);
+    data.renderedItems = [...data.renderedItems];
```
ecraig12345, line 94: Same here

packages/react/src/components/Breadcrumb/Breadcrumb.base.tsx
```diff
@@ -47,131 +41,103 @@ const nonActionableItemProps: Partial<IContextualMenuItemProps> = {
...
 
-    const contextualItems = renderedOverflowItems.map(
+    const contextualItems = data.renderedOverflowItems.map(
       (item): IContextualMenuItem => {
         const isActionable = !!(item.onClick || item.href);
         return {
           name: item.text,
           key: item.key,
-          onClick: item.onClick ? this._onBreadcrumbClicked.bind(this, item) : null,
+          onClick: item.onClick ? onBreadcrumbClicked.bind(ref, item) : null,
```
ecraig12345, line 140: Probably better to just do it this way since `this` isn't important.
```suggestion
          onClick: item.onClick ? ev => onBreadcrumbClicked(ev, item) : null,
```

ecraig12345 (approved):
Generally looks good, just a couple minor comments.

packages/react/src/components/Breadcrumb/Breadcrumb.base.tsx
```diff
@@ -24,14 +19,13 @@ import {
...
-/** @deprecated Use IBreadcrumbData */
-export type IBreadCrumbData = IBreadcrumbData;
-
 const getClassNames = classNamesFunction<IBreadcrumbStyleProps, IBreadcrumbStyles>();
 
-const OVERFLOW_KEY = 'overflow';
 const nullFunction = (): null => null;
 
+const OVERFLOW_KEY = 'overflow';
+const COMPONENT_NAME = 'BreadcrumbBase';
```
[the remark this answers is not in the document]
czearing: I talked with @dzearing and it may be worth removing Base from component's displayNames. Some of the regex from the conformance test should also be removed to enforce that the displayName matches the components name and does not contain Base.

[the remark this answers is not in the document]
ecraig12345: Sounds good to me!

ecraig12345:
Looks like there's a build error with the callback change.

ecraig12345:
Closing this since it's not going in version 8 at this point, sorry :(

PR #15655, opened 2020-10-22
Fix copy when keyboard focus is on the SelectedItemsList in the UnifiedPicker

elisabethcvs:
#### Pull request checklist

#### Description of changes

The _onCopy function was not getting called  on the wrapper div in the UnifiedPicker unless focus was in the autofill input. So if you were keyboard navigating the SelectedItemsList, keyboard copy never worked. But now that we removed mouse selection, there was no way to copy multiple items.

This adds the ability to copy when keyboard navigating. onCopy only gets called if the focus is in an input. So the SelectedItemsList now has a zero size input that's not keyboard selectable, that gets focus when the list gets focus. Copy is working.

#### Focus areas to test

- Tested that we can't see the input or access it via keyboard
- Tested that backspace still works
- Tested that multiselect still works

### DIFF
--- a/packages/react-experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
+++ b/packages/react-experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
@@ -74,6 +74,7 @@ const _SelectedItemsList = <TItem extends BaseSelectedItem>(
             // Add a zero zero size input. This is to pick up the copy command when we have
             // keyboard focus in the list
             // Focus is set to the input when the list gets focus via the _onFocus function
+            className={'ms-SelectedItemsList-copyInput'}
             ref={hiddenInput}
             style={{ height: '0px', width: '0px', border: 'none', outline: 'none' }}
             data-is-focusable={false}
--- a/packages/react-experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
+++ b/packages/react-experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
@@ -1,6 +1,16 @@
+import { css, classNamesFunction } from '@fluentui/react';
 import * as React from 'react';
+import {
+  ISelectedItemsList,
+  ISelectedItemsListProps,
+  BaseSelectedItem,
+  ISelectedItemsListStyleProps,
+  ISelectedItemsListStyles,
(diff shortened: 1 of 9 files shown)


packages/react-experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
```diff
@@ -59,11 +60,24 @@ const _SelectedItemsList = <TItem extends BaseSelectedItem>(
...
     <>
       {items.length > 0 && (
-        <div role={'list'}>
+        <div role={'list'} onFocus={_onFocus}>
+          <input
+            // Add a zero zero size input. This is to pick up the copy command when we have
+            // keyboard focus in the list
+            // Focus is set to the input when the list gets focus via the _onFocus function
+            ref={hiddenInput}
+            style={{ height: '0px', width: '0px', border: 'none', outline: 'none' }}
```
elisabethcvs, line 78: I probably need to put a class name on this so it can be styled yes?

packages/react-experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
```diff
@@ -59,11 +60,28 @@ const _SelectedItemsList = <TItem extends BaseSelectedItem>(
...
-        <div role={'list'}>
+        <div role={'list'} onFocus={_onFocus}>
+          <input
+            // Add a zero zero size input. This is to pick up the copy command when we have
+            // keyboard focus in the list
+            // Focus is set to the input when the list gets focus via the _onFocus function
+            className={'ms-SelectedItemsList-copyInput'}
+            ref={hiddenInput}
+            hidden={!props.getItemCopyText}
+            style={{ height: '0px', width: '0px', border: 'none', outline: 'none' }}
```
nebhatna, line 80: i know this is zero size input but is there a way we can put this in the styles file for cleanliness?

nebhatna (approved):
one minor comment

packages/react-experiments/src/components/SelectedItemsList/SelectedItemsList.tsx
elisabethcvs: yeah good call will update

PR #15666, opened 2020-10-23
Rename Pivot to Tabs, Part 1: Rename files and update imports

behowell:
#### Pull request checklist

#### Description of changes

This is the first of two PRs for renaming the Pivot component to Tabs. This PR only renames files and updates the import/export statements to refer to the new file. There will be a follow-up PR that actually renames the components and properties.

Splitting it out into two PRs allows git to track the renames so that source history is preserved.

Details about the API changes from this rename are here: https://hackmd.io/jDytiF_kQuW2hNELbxgNqA


packages/react/src/Pivot.ts
```diff
@@ -1 +0,0 @@
-export * from '@fluentui/react-tabs/lib/Pivot';
```
xugao, line 1: we should probably add `Tab` but keeping `Pivot` as is and deprecate. let's follow up tmr

behowell: I'm planning to export the `Pivot` component from Tabs.ts as a deprecated alias of `Tabs`, in Part 2. As of this PR, the component itself is still named `Pivot`.

If there's a reason to keep the _file_ Pivot.ts around, I can add it back. I suppose it would help avoid problems with places that import from the Pivot.ts file directly.

ecraig12345: We should keep `packages/react/src/Pivot.ts` since we consider that part of the public API that people are allowed to import directly from (maybe also `packages/react-tabs/src/Pivot.ts`).

apps/public-docsite-resources/src/components/pages/PivotPage.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
```
khmakoto, line 1: Since we're updating file names here should we also rename this to `TabsPage`? Similar comment for other files below.

packages/react-tabs/src/components/Pivot/index.ts
```diff
@@ -1,5 +1 @@
-export * from './Pivot';
-export * from './Pivot.base';
-export { PivotItem } from './PivotItem';
-export * from './Pivot.types';
-export * from './PivotItem.types';
+export * from '../Tabs/index';
```
khmakoto, line 1: Is this file still needed? Couldn't you just have the `Pivot.ts` file under `src` refer to `Tabs/index` instead of keeping this file around? That way we could remove the `Pivot` folder completely.

behowell: I was planning to use this file in Part 2 to re-export the Pivot, etc. names (as aliases of Tabs, etc.). Although I suppose I can just use react-tabs/src/Pivot.ts to do that...

khmakoto (commented):
You probably also need to update references to PivotPage in `AppDefinition.tsx`.

packages/react/src/Tabs.ts
```diff
@@ -0,0 +1 @@
+export * from '@fluentui/react-tabs/lib/Tabs';
```
xugao, line 1: let's not export `Tabs` yet until the all the renaming are done and we are ready to ship with it without futher breaking changes. Also, how are you planning to rename the props? `Pivot` needs to stay as is.

packages/react-tabs/src/components/Tabs/Tabs.test.tsx
```diff
@@ -21,10 +21,11 @@ describe('Pivot', () => {
...
 
-  isConformant({
-    Component: Pivot,
-    displayName: 'Pivot',
-  });
+  // TODO re-enable this test after renaming Pivot to Tabs
+  // isConformant({
+  //   Component: Pivot,
+  //   displayName: 'Pivot',
+  // });
```
xugao, line 28: we should not disable all the conformance tests. can cherry-pick the ones which make sense to disable

ecraig12345:
Somewhere (can't find it now) I saw a question about naming of PivotItem/TabPanel, and the new name choice being mainly based on aria. Have you checked what other frameworks call it?

behowell:
Abandoning this PR in favor of #15782, which creates Tabs as a copy of Pivot, rather than renaming it. This'll make it easier to make future API updates to Tabs without breaking existing users of Pivot.

PR #15673, opened 2020-10-23
Use type from state iff canRevealPassword is true

### LINKED ISSUE #15670: TextField: Can't change field type if initially set to password.
### Environment Information

- **Package version(s)**: v7.145.0 and above

### Details
https://github.com/microsoft/fluentui/blob/62b9b2e5daaf345097f6e2382896d64ed6c18e05/packages/office-ui-fabric-react/src/components/TextField/TextField.base.tsx#L488-L493

The line 492 line breaks the expected behavior of the component (whose initial `type` was `password`) when the `type` of component is updated by it's parent through props. This should only be considered when `this.props.canRevealPassword` is set to `true`.

https://codepen.io/faizal-hasanwala/pen/XWKpmNK?editors=0010

#### Actual behavior:

The input field type remains set to password.

#### Expected behavior:

The input field type should change to text.

Requested priority: Normal

faizal3199:
#### Pull request checklist

- [X] Addresses an existing issue: Fixes #15670

#### Description of changes

Set the type of TextField based on below priority list:
1. `this.state.type` iff `this.props.canRevealPassword` is set to `true`.
2. Use `type` from props.
3. Default to `text`.

#### Note:
`$ yarn change`: Command created no file.

### DIFF
--- a/change/office-ui-fabric-react-2020-10-24-10-42-39-7.0.json
+++ b/change/office-ui-fabric-react-2020-10-24-10-42-39-7.0.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "TextField: Use type from state iff canRevealPassword is true",
+  "packageName": "office-ui-fabric-react",
+  "email": "fahasanw@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-10-24T05:12:39.137Z"
+}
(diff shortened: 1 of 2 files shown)


ecraig12345 (approved):
Thanks for doing this! You'll also need to check out the repo locally and run `yarn change` to generate a change file.

khmakoto:

faizal3199:
Added the change file.

ecraig12345:
@faizal3199 Thanks again for doing this? Would you mind also cherry-picking the change to `master` to ensure it's included in Fluent UI React 8?

ecraig12345:
First sentence of my previous comment was meant to end in `!` (not a vaguely passive-aggressive `?`), sorry! 😆

faizal3199:
@ecraig12345 This bug doesn't exists in master branch. Reveal password button functionality (which introduced this bug) isn't present in master branch.

ecraig12345:
@faizal3199 Good catch, it's definitely supposed to be in master. 🤦‍♀️ In that case I'll add the feature to master including your fix.

faizal3199:
Thank you! 😄

PR #15692, opened 2020-10-26
Multi version docsite

ling1726:
#### Pull request checklist

#### Description of changes

Adds multi version support to the current N* docsite

#### Focus areas to test

(optional)


scripts/config.ts
```diff
@@ -9,8 +9,14 @@ const env = process.env.NODE_ENV || 'development';
...
 const __PERF__ = !!process.env.PERF;
 const __PROD__ = env === 'production';
-// This needs a trailing slash or images won't work
-const __BASENAME__ = process.env.DEPLOYBASEPATH ? `/${process.env.DEPLOYBASEPATH}/react-northstar/` : '/';
+let __BASENAME__ = process.env.PR_DEPLOY
+  ? // This needs a trailing slash or images won't work
+    `/pr-deploy-site/${process.env.BUILD_SOURCEBRANCH}/react-northstar/`
+  : '/';
+
+if (process.env.OFFICIALRELEASE) {
```
ling1726, line 17: I might not need `process.env.OFFICIALRELEASE` but I need to find out how/where `process.env.PR_DEPLOY` is set

ling1726: using this env variable would still give us an easy way to build docsites for single versions

packages/fluentui/docs/src/index.html
```diff
@@ -0,0 +1,116 @@
+<!DOCTYPE html>
```
ling1726, line 1: Any cleaner naming for this file ?

azure-pipelines.release-fluentui.yml
```diff
@@ -16,23 +22,31 @@ steps:
...
+      echo "Docsite base path published to for version $ver"
+    displayName: Set base path and version variables for job
   - script: |
       git config user.name "Fluent UI Build"
       git config user.email "fluentui-internal@service.microsoft.com"
       git remote set-url origin https://$(githubUser):$(githubPAT)@github.com/microsoft/fluentui.git
     displayName: Authenticate git for pushes
-
   - script: |
-      npx midgard-yarn install
```
layershifter, line 26: Why this is changed?

packages/fluentui/docs/src/components/Sidebar/VersionDropdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * A dropdown component that fetches available versions of the docsite from a hosted manifest file and renders available versions
+ * Makes assumptions about docsite routing and hosting
+ *
+ * @param props Component props
+ */
+export function VersionDropdown(props: { width: number }) {
+  const [versions, setVersions] = React.useState<string[]>([]);
+  React.useEffect(() => {
+    fetch('http://localhost:3000/manifest.json') // Assume that the manifest is hosted in the same blob storage
```
layershifter, line 16: Oops

azure-pipelines.release-fluentui.yml
ling1726: It wasn't working in the build for 47 when I first made these changes, I will revert and see if that works

scripts/config.ts
```diff
@@ -9,8 +9,14 @@ const env = process.env.NODE_ENV || 'development';
 const __DEV__ = env === 'development';
 const __PERF__ = !!process.env.PERF;
 const __PROD__ = env === 'production';
-// This needs a trailing slash or images won't work
-const __BASENAME__ = process.env.DEPLOYBASEPATH ? `/${process.env.DEPLOYBASEPATH}/react-northstar/` : '/';
+let __BASENAME__ = process.env.PR_DEPLOY
+  ? // This needs a trailing slash or images won't work
+    `/pr-deploy-site/${process.env.BUILD_SOURCEBRANCH}/react-northstar/`
+  : '/';
```
layershifter, line 15: Oops, we are using `DEPLOYBASEPATH` in `master`

azure-pipelines.release-fluentui.yml
```diff
@@ -16,23 +22,31 @@ steps:
...
     displayName: yarn build
-
   - script: |
       NODE_ENV=production yarn build:fluentui:docs
     displayName: yarn build:fluentui:docs
+  - task: PublishBuildArtifacts@1
+    displayName: 'Publish docsite as pipeline artifact'
+    inputs:
+      PathtoPublish: 'packages/fluentui/docs/dist'
+      ArtifactName: 'docsite_v$(docsiteVersion)'
```
layershifter, line 49: Is it required?

ling1726: no, I think but it's handy to have the artifacts available... storage is cheap and we can backfill easily if we ever move storage or have to clean it for some reason

packages/fluentui/docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -486,6 +487,7 @@ const Sidebar: React.FC<RouteComponentProps & SidebarProps> = props => {
         >
           Fluent <span style={gradientTextStyles}>UI</span>
         </Text>
+        <VersionDropdown width={props.width} />
```
layershifter, line 490:
```suggestion
        {process.env.NODE_ENV === 'production' && <VersionDropdown width={props.width} />}
```

So we hide this during local builds?

ling1726: This is should already be covered in the component itself, although you reminded me that I have no `catch` for the fetch.... is that enough for you or should we be more explicit to not render this locally at all ? It might be necessary to locally host a manifest to display this dropdown if you're working on it (like I have)

miroslavstastny (approved):

miroslavstastny (approved):

ecraig12345 (approved):

PR #15701, opened 2020-10-26
ResizeGroup: Improve perf by making sure resize data state change is batched.

xugao:
#### Pull request checklist

#### Description of changes
`React` is not batching the state changes when states are set in callbacks like `setTimeout` or `requestAnimationFrame`. See issue: https://github.com/facebook/react/issues/14259

In `ResizeGroup`, we call `updateResizeState` from `requestAnimationFrame` callback. `updateResizeState` call different `useState` setters which causes un-batched re-renders. The fix here is to group `resize` data into a single state in `useResizeData` hook. 

Also noticed couple places where we should memoize the callbacks returned from `updateResizeState`.

#### Focus areas to test
`perf-test` results for components which uses `ResizeGroup`: `Breadcrumb`, `CommandBar`

### DIFF
--- a/packages/react-internal/src/components/ResizeGroup/ResizeGroup.base.tsx
+++ b/packages/react-internal/src/components/ResizeGroup/ResizeGroup.base.tsx
@@ -314,13 +314,17 @@ type ResizeDataAction = {
   value: IResizeGroupState[keyof IResizeGroupState] | IResizeGroupState;
 };
 
+/**
+ * Use useReducer instead of userState because React is not batching the state updates
+ * when state is set in callbacks of setTimeout or requestAnimationFrame.
+ * See issue: https://github.com/facebook/react/issues/14259
+ */
 function resizeDataReducer(state: IResizeGroupState, action: ResizeDataAction): IResizeGroupState {
   switch (action.type) {
     case 'resizeData':
       return { ...action.value };
-    case 'dataToMeasure': {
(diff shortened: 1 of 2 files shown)


packages/react-internal/src/components/ResizeGroup/ResizeGroup.base.tsx
```diff
@@ -309,63 +309,61 @@ const hiddenDivStyles: React.CSSProperties = { position: 'fixed', visibility: 'h
 const hiddenParentStyles: React.CSSProperties = { position: 'relative' };
 const COMPONENT_NAME = 'ResizeGroup';
 
+type ResizeDataAction = {
+  type: 'resizeData' | keyof IResizeGroupState;
+  value: IResizeGroupState[keyof IResizeGroupState] | IResizeGroupState;
+};
+
+function resizeDataReducer(state: IResizeGroupState, action: ResizeDataAction): IResizeGroupState {
```
ecraig12345, line 317: Can you add comments inline about the reason for this optimization (basically what you wrote in the PR description)?

xugao: Done!

ecraig12345 (approved):
Nice fix!

micahgodbolt (approved):

PR #15707, opened 2020-10-27
[isConformant] Implementing component-handles-classname test

czearing:
### Pull request checklist
- [X] Addresses an existing issue: #14888

#### Summary
1. Adding a `component-handles-classname` test to isConformant's  defaultTests.
2. Removing redundant tests ComponentConformance.test.tsx from react-internal.
3. Removing redundant ComponentConformance.test.tsx from react.
4. Removing redundant tests in react-northstar isConformant tests.
5. Disabling tests that don't handle classNames correctly. (Either they aren't applied to the root element or not passing classNames correctly. )

#### What this test does:
1. Checks if the component handles the className prop.
2. Checks if the component handles any existing default provided classNames correctly.
3. Checks if the component puts the className on the root element.

```jsx
  /** Component file handles classname prop */
  'component-handles-classname': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
    const { Component, wrapperComponent, helperComponents = [], requiredProps, customMount = mount } = testInfo;

    let el = customMount(<Component {...requiredProps} />);
    let component = getComponent(el, helperComponents, wrapperComponent);

    const defaultClassNames =
      component
        ?.getDOMNode()
        ?.getAttribute('class')
        ?.split(' ') || [];

    const testClassName = 'testComponentClassName';
    const mergedProps: Partial<{}> = {
      ...requiredProps,
      className: 'testComponentClassName',
    };

    el = customMount(<Component {...mergedProps} />);
    component = getComponent(el, helperComponents, wrapperComponent);

    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    const classNames: any = component
      .getDOMNode()
      .getAttribute('class')
      ?.split(' ');

    it(`handles className prop`, () => {
      expect(classNames.indexOf(testClassName) >= 0).toEqual(true);
    });

    it(`handles component's default classNames`, () => {
      if (defaultClassNames.length && defaultClassNames[0] !== '') {
        for (const defaultClassName of defaultClassNames) {
          expect(classNames.indexOf(defaultClassName) >= 0).toEqual(true);
        }
      }
    });
  },
```

#### defaultErrorMessages

If the component doesn't correctly handle the classname prop.

If the default classname isn't correctly handled.

#### List of removed `react-northstar` tests:
- #### ( Basic Tests ) replaced with `name-matches-filename`, `exports-component`:    	* Component exports    	* Component has a constructor name
- #### ( ClassName Tests ) replaced with `component-handles-classname`:
     * ClassName applies to root element
     * ClassName applies to root component
     * User ClassName does not override the default classes

#### List of removed `ComponentConformance` tests:
- #### ( Component File Conformance ) replaced with `component-handles-classname`:
     * Component injects a className prop
- #### ( Top Level Component File Conformance ) replaced with `exported-top-level`, `has-top-level-file`:
     * Component has a top level file for each file in the directory

#### Disabled V8 Components
* `Breadcrumb`: Doesn't apply the test className to the root.
* `Callout`: Doesn't apply the test className to the root.
* `TeachingBubble`: Doesn't apply the test className to the root.
* `ContextualMenu`: Requires createPortal.
* `Rating`: Doesn't apply the defaultClassName to the root.
* `Callout`: Doesn't apply the test className to the root.
* `BaseSelectedItemsList`: Doesn't apply the test className to the root.
* `SelectedPeopleList`: Doesn't apply the test className to the root.
* `ButtonGrid`: Doesn't apply the test className to the root.
* `Modal`: Requires createPortal.
* `PersonaPresence`: Doesn't apply the test className to the root.
* `Coachmark`: Doesn't apply the test className to the root.
* `Dialog`: Refers to modalProps.className.
* `Nav`: Doesn't apply the test className to the root.
* `Panel`: Doesn't apply the test className to the root.
* `Sticky`: Doesn't apply the test className to the root.
* `SelectionZone`: Doesn't apply the test className to the root.
* `CalendarDayGrid`: Doesn't apply the test className to the root.


packages/react-internal/src/components/ColorPicker/__snapshots__/ColorPicker.test.tsx.snap
```diff
@@ -27,7 +27,7 @@ exports[`ColorPicker renders correctly 1`] = `
         }
   >
     <div
-      aria-describedby="ColorRectangle-description0"
+      aria-describedby="ColorRectangle-description32"
```
dzearing, line 30: Hmm. can you add `resetId` to ColorPicker.test beforeEach?

change/@fluentui-react-2020-10-26-17-19-22-feat-isConformant-classname-tests.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
```
ecraig12345, line 2: This can be none since it's just touching tests.
```suggestion
  "type": "none",
```

change/@fluentui-react-internal-2020-10-26-17-19-22-feat-isConformant-classname-tests.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "prerelease",
```
ecraig12345, line 2: Same here
```suggestion
  "type": "none",
```

packages/react-conformance/src/defaultErrorMessages.tsx
```diff
@@ -311,6 +311,106 @@ export const defaultErrorMessages = {
...
+        receivedErrorMessage(error),
+    );
+  },
+
+  'component-handles-classname': (
+    componentInfo: ComponentDoc,
+    testInfo: IsConformantOptions,
+    error: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    classNames: any,
```
ecraig12345, line 323: Can we give this a type?

packages/react-conformance/src/defaultErrorMessages.tsx
```diff
@@ -311,6 +311,106 @@ export const defaultErrorMessages = {
...
+    );
+  },
+
+  'component-handles-default-classname': (
+    componentInfo: ComponentDoc,
+    testInfo: IsConformantOptions,
+    error: string,
+    defaultClassName: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    classNames: any,
```
ecraig12345, line 372: Same here

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -146,6 +146,64 @@ export const defaultTests: TestObject = {
...
+    let component = getComponent(el, helperComponents, wrapperComponent);
+
+    const defaultClassNames =
+      component
+        ?.getDOMNode()
+        ?.getAttribute('class')
+        ?.split(' ') || [];
+
+    const testClassName = 'testComponentClassName';
+    const mergedProps: Partial<{}> = {
```
ecraig12345, line 163: What's with the type here?

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -146,6 +146,64 @@ export const defaultTests: TestObject = {
...
+      ...requiredProps,
+      className: 'testComponentClassName',
+    };
+
+    el = customMount(<Component {...mergedProps} />);
+    component = getComponent(el, helperComponents, wrapperComponent);
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const classNames: any = component
```
ecraig12345, line 172: This should be `string[] | undefined` it looks like?

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -146,6 +146,64 @@ export const defaultTests: TestObject = {
...
+      .getDOMNode()
+      .getAttribute('class')
+      ?.split(' ');
+
+    it(`handles className prop`, () => {
+      try {
+        expect(classNames.indexOf(testClassName) >= 0).toEqual(true);
```
ecraig12345, line 179: There's a better matcher for this: https://jestjs.io/docs/en/24.x/expect#tocontainitem
```suggestion
        expect(classNames).toContain(testClassName);
```

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -146,6 +146,64 @@ export const defaultTests: TestObject = {
...
+        throw new Error('component-handles-classname (handles className prop)');
+      }
+    });
+
+    it(`handles component's default classNames`, () => {
+      let defaultClassName: string = '';
+      try {
+        if (defaultClassNames.length && defaultClassNames[0] !== '') {
+          for (defaultClassName of defaultClassNames) {
+            expect(classNames.indexOf(defaultClassName) >= 0).toEqual(true);
```
ecraig12345, line 191:
```suggestion
            expect(classNames).toContain(defaultClassName);
```

packages/react-internal/src/components/ComponentConformance.test.tsx
```diff
@@ -269,17 +38,6 @@ describe('Top Level Component File Conformance', () => {
...
-  components.forEach(componentName => {
-    it(`${componentName} has a corresponding top level component file`, () => {
-      expect(
-        fs.existsSync(path.resolve(__dirname, `../${componentName}.ts`)) ||
-          fs.existsSync(path.resolve(__dirname, `../${componentName}.tsx`)),
-      ).toBeTruthy();
-    });
-  });
-
   // make sure that there is a version import in each corresponding top level component file
```
ecraig12345, line 41: Now that the previous test checking that the file exists is removed, what does a failure in the version import test look like if the top-level file doesn't exist? Want to make sure it's not too confusing/misleading.

packages/react/src/components/ComponentConformance.test.tsx
```diff
@@ -1,194 +0,0 @@
-import * as React from 'react';
```
ecraig12345, line 1: Not sure we're ready to delete this whole file? Instead it should be limited to only running the remaining tests for the components within this package I think.

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -146,6 +146,64 @@ export const defaultTests: TestObject = {
...
+    const { Component, wrapperComponent, helperComponents = [], requiredProps, customMount = mount } = testInfo;
+
+    let el = customMount(<Component {...requiredProps} />);
+    let component = getComponent(el, helperComponents, wrapperComponent);
+
+    const defaultClassNames =
+      component
+        ?.getDOMNode()
+        ?.getAttribute('class')
+        ?.split(' ') || [];
```
layershifter, line 160: I am a bit worried about this check...

Our components are using CSS-in-JS renderer and for N* we have a noop renderer when there is no React Context provided. Is it the same for V8/next? For example, are there any guarantees that we will not get `msComponent-root-XXX` in in `defaultClassNames`?

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -146,6 +146,64 @@ export const defaultTests: TestObject = {
...
+
+    const testClassName = 'testComponentClassName';
+    const mergedProps: Partial<{}> = {
+      ...requiredProps,
+      className: 'testComponentClassName',
+    };
+
+    el = customMount(<Component {...mergedProps} />);
+    component = getComponent(el, helperComponents, wrapperComponent);
```
layershifter, line 169: I suggest to avoid reassignment in this case and use separate variables

packages/react-internal/src/components/ComponentConformance.test.tsx
czearing: There aren't any visible changes in the top-level version test message in ComponentConformance but it is a bit confusing. I'm planning on re-releasing the component-has-version test after this pull request and it should improve the error message.

Before:


After:
image

packages/react/src/components/ComponentConformance.test.tsx
czearing: The existing ComponentConformance tests are redundant to the isConformant ones and both are currently running.

ComponentConformance tests

1.	Injects a className : **component-handles className**
2.	Contains a top-level file : **component-has-top-level-file**
3.	Top-level file contains a version import : currently commented out

Since the version test is currently commented out  (and will be re-released as **component-has-top-level-version** after this pull request ) it may be worth deleting the file for the time being.

packages/react-conformance/src/defaultTests.tsx
czearing: I believe that it should be fine. In scenarios where a defaultClassName isn't recieved to begin with. The test should still pass as it will only run if one is provided:
```jsx
if (defaultClassNames.length && defaultClassNames[0] !== '') {
...
}
``` 
Otherwise, it will test to see if the existing defaultClassName exists alongside the testClassName.

layershifter (approved):
Approved from N* side 👍

layershifter:
@dzearing is there something blocking on this PR from Fluent side?

layershifter:
@czearing can you please check this PR, it seems that after merge some change files are missing 💭

dzearing:
Will check with @czearing and @ecraig12345

dzearing:
Thanks, @ecraig12345 can you please merge once you've signed off?

packages/react-focus/src/components/FocusZone/__snapshots__/FocusZone.test.tsx.snap
```diff
@@ -9,7 +9,7 @@ exports[`FocusZone renders FocusZone correctly with aria-describedby and aria-la
       &:focus {
         outline: none;
       }
-  data-focuszone-id="FocusZone1"
+  data-focuszone-id="FocusZone3"
```
dzearing, line 12: This smells like resetIds is not being called in beforeEach.

packages/react-internal/src/components/CommandBar/__snapshots__/CommandBar.test.tsx.snap
```diff
@@ -41,7 +41,7 @@ exports[`CommandBar renders commands correctly 1`] = `
               padding-right: 14px;
               padding-top: 0;
             }
-        data-focuszone-id="FocusZone0"
+        data-focuszone-id="FocusZone2"
```
dzearing, line 44: resetIds ?

packages/react-internal/src/components/MessageBar/__snapshots__/MessageBar.test.tsx.snap
```diff
@@ -101,7 +101,7 @@ exports[`MessageBar renders MessageBar correctly 1`] = `
           @media screen and (-ms-high-contrast: active){& {
             -ms-high-contrast-adjust: none;
           }
-      id="MessageBar0"
+      id="MessageBar2"
```
dzearing, line 104: resetIds

packages/react-internal/src/components/TextField/MaskedTextField/__snapshots__/MaskedTextField.test.tsx.snap
```diff
@@ -91,7 +91,7 @@ exports[`MaskedTextField renders correctly 1`] = `
     >
       <input
         aria-invalid={false}
-        aria-labelledby="TextFieldLabel2"
+        aria-labelledby="TextFieldLabel8"
```
dzearing, line 94: resetIds

dzearing (approved):

packages/react-conformance/src/defaultTests.tsx
```diff
@@ -146,6 +146,60 @@ export const defaultTests: TestObject = {
     });
   },
 
+  /** Component file handles classname prop */
+  'component-handles-classname': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
+    const { Component, wrapperComponent, helperComponents = [], requiredProps, customMount = mount } = testInfo;
+
+    const defaultEl = customMount(<Component {...requiredProps} />);
```
ecraig12345, line 153: I'll fix this but each test (`it()`) should be standalone, and mounting components should happen inside the test itself.

ecraig12345 (approved):

PR #15710, opened 2020-10-27
Store website info in sessionStorage

ecraig12345:
#### Pull request checklist

#### Description of changes

None of the info on our website that we were putting in localStorage is actually that important to persist. Move to sessionStorage for easier compliance.

One of the things being read from localStorage was the language of the page (by DefaultFontStyles). So I updated the `language` utility to have an option for where to store the value and changed the default to sessionStorage.

### DIFF
--- a/apps/public-docsite/package.json
+++ b/apps/public-docsite/package.json
@@ -50,6 +50,7 @@
     "@fluentui/font-icons-mdl2": "^8.0.0-beta.1",
     "@fluentui/set-version": "^8.0.0-beta.0",
     "@fluentui/theme-samples": "^8.0.0-beta.2",
+    "@fluentui/utilities": "^8.0.0-beta.1",
     "@types/react": "16.8.25",
     "@types/react-dom": "16.8.4",
     "json-loader": "^0.5.7",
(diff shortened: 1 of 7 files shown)


khmakoto (approved):

scripts/where/index.ts
```diff
@@ -0,0 +1,162 @@
...
+const packageJsonGlob = '{apps,packages}/!(fluentui|web-components)/package.json';
+
+const components7 = lsTree('upstream/7.0', componentsGlob);
+const components8 = lsTree('upstream/master', componentsGlob);
+
+const componentFolders = glob.sync('packages/!(react-docsite-components)/src/components/!(*.*)', { cwd: gitRoot });
+
+const packages = lernaAlias({ sourceDirectory: false });
+
+function whereIsPackage(packageName: string) {
```
xugao, line 71: Nice :) but i don't think this is intended for this PR?

xugao:
I am good with this change. However, I am curious why `localStorage` is less compliant than `sessionStorage`?

xugao (approved):

scripts/where/index.ts
ecraig12345: Oops 😆 will remove

PR #15724, opened 2020-10-27
Spin button: ensure default fonts are set

### LINKED ISSUE #15362: SpinButton is not inheriting Fabric default fontFamily
### Environment Information

"office-ui-fabric-react": "^7.144.2",

https://codepen.io/jacklovett/pen/bGebNPa

#### Actual behavior:

Since office-ui-fabric-react-7.140.0 when font-family: inherit was added to a number of input components here: https://github.com/microsoft/fluentui/pull/15172, SpinButton now displays as "Times" instead of the Fabric default font value. 

#### Expected behavior:

Expected behaviour is that SpinButton components would inherit the default Fabric font styles

Requested priority: (Low)

LeslieW:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15362

#### Description of changes

Currently, the SpinButton font family is not inherited from Fabric and defaults to 'Times'.

This change adds `fonts.medium` to the SpinButton component styles so that font family results in the Fabric default. This change mimics the use of font styling across other button types. 

Before:
spin_button_before

After:
spin_button_after

#### Focus areas to test

(optional)


ecraig12345:
@xugao should probably review this since she made the change to add `font-family: inherit` in #15172.

apps/a11y-tests/src/tests/__snapshots__/ComponentExamples.test.tsx.snap
```diff
@@ -1,6836 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
xugao, line 1: pls revert this file

xugao: other than that, the change looks good to me. `SpinButton` has always been relying on `Fabric` wrapper to drop the default fonts, which I think we should change to let `SpinButton` to define fonts on it own (your change)

LeslieW: Thanks @xugao, done!

xugao: looks like there are some snapshots you missed to update. To do that, go to `packages/react-examples` and run `yarn update-snapshots`

LeslieW: Sounds good, thanks @xugao! Updated the snapshots.

xugao (approved):
LGTM! 

Could you please also make the same fix in `7.0` branch [here](https://github.com/microsoft/fluentui/blob/7.0/packages/office-ui-fabric-react/src/components/SpinButton/SpinButton.styles.ts#L148). That way the fix will be published for  `office-ui-fabric-react@v7` / `@fluentui/react@v7` and you can pick up in your app right away.

The change you have now in master will be published under `@fluentui/react@v8-beta`.

khmakoto (approved):

LeslieW:
Thanks @xugao and @khmakoto for your reviews! Sounds great, I created another [pull request](https://github.com/microsoft/fluentui/pull/15801) to add this change to the 7.0 branch.

Issue #15727, opened 2020-10-27
Export WindowProvider from react


ecraig12345:
Should `export * from '@fluentui/react-window-provider'` either in `@fluentui/react` or somewhere in the line of packages it exports.

PR #15736, opened 2020-10-27
Major bump react-window-provider and export from OUFR

xugao:
#### Pull request checklist

#### Description of changes
Related issue: #15727

- Major bump `react-window-provider` to `1.0.0` since it's used by other in-production packages. From [versioning perspective](https://semver.org/spec/v1.0.0.html#how-do-i-know-when-to-release-100), it's more correct to be versioned as `v1` instead of `v0`. Having versioned `v1` can also help avoid package duplications:
image

- export all from `@fluentui/react-window-provider` from OUFR (under `lib/WindowProvider`).
#### Focus areas to test

(optional)

### DIFF
--- a/packages/office-ui-fabric-react/src/Utilities.ts
+++ b/packages/office-ui-fabric-react/src/Utilities.ts
@@ -1,3 +1,2 @@
 import './version';
 export * from '@uifabric/utilities';
-export * from '@fluentui/react-window-provider';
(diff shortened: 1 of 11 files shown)


packages/office-ui-fabric-react/src/Utilities.ts
```diff
@@ -1,2 +1,3 @@
 import './version';
 export * from '@uifabric/utilities';
+export * from '@fluentui/react-window-provider';
```
ecraig12345, line 3: Wondering if we should make a separate file for this since it's not actually in utilities?

xugao: thought about it more - I will move to `lib/WindowProvider`

ecraig12345 (approved):

PR #15737, opened 2020-10-27
remove check for allowFreeForm to enable readOnly prop

### LINKED ISSUE #15406: Combobox - allowfreeform is false and disabled is false the input element becomes readonly although we can choose from dropdown
**Environment Information**
**Package version(s)**: 7.68.0
**Browser and OS versions:** Edge Version 85.0.564.68 (Official build) (64-bit) / Windows 10

**Describe the issue:**
Incorrect narration by screen reader i.e. read only is announced on combo box even if we are able to select options from combobox

[Codepen](https://codepen.io/ashik-microsoft/pen/ZEOYmoY?editable=true%3Dhttps%3A%2F%2Fdeveloper.microsoft.com%2F)

```
const { IComboBoxOption, IComboBoxStyles, VirtualizedComboBox, Fabric, initializeIcons } = window.Fabric;

// Initialize icons in case this example uses them
initializeIcons();

const comboBoxOption: IComboBoxOption[] = Array.from({ length: 1000 }).map((x, i) => ({
  key: `${i}`,
  text: `Option ${i}`,
}));

const comboBoxStyles: Partial<IComboBoxStyles> = { root: { maxWidth: '300px' } };

const ComboBoxVirtualizedExample: React.FC = () => (
  <Fabric className="ms-ComboBoxExample">
    <VirtualizedComboBox
      styles={comboBoxStyles}
      defaultSelectedKey="547"
      label="Scaled/virtualized example with 1000 items"
      allowFreeform={false}
      autoComplete="on"
      options={comboBoxOption}
      dropdownMaxWidth={200}
      useComboBoxAsMenuWidth
    />
  </Fabric>
);

const ComboBoxVirtualizedExampleWrapper = () => <Fabric><ComboBoxVirtualizedExample /></Fabric>;
ReactDOM.render(<ComboBoxVirtualizedExampleWrapper />, document.getElementById('content'))
```

Just check the combox input element by inspecting it and yo will see it as **readonly**.

fabric

**Actual behavior:**
The narrator narrates incorrect information i.e. 'Read Only' when the focus lands on combo box.

**Expected behavior:**
The narrator shouldn't narrates 'Read only' with the correct information. It misguide the user's to use the functionality of the control.

tomi-msft:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #15406 

#### Description of changes

When combobox with disabled and allowFreeForm set to false, it sets the readonly prop to true. This is confusing when screen reader reads this as readonly when there are selectable options in the combobox. This PR removes the check for allowFreeForm and will only set readonly prop to true when disabled is set to true.

#### Focus areas to test

(optional)


xugao:
Please make the fix in `7.0` branch as well. We should mark the issue as fixed in the `7.0` PR

joschect (approved):

PR #4518, opened 2018-04-11
Add ability to use Customizer to control the default layer host

ThomasMichon:
# Overview

This change provides a way to control all `Layer` components within a scope using `Customizer`, so they all use a default `LayerHost` with the provided `hostId`. This is accomplished by adding `hostId` to the set of customizable props of `Layer`.

Added an example for how to customize `Layer` in this manner so it works with a provided `LayerHost`.

PR #15756, opened 2020-10-29
Layer: fix not rendering into targeted host element defined by hostId prop

xugao:
#### Pull request checklist

#### Description of changes
- Fix `Layer` not rendering into host element (defined by `hostId` prop). See repro [here](https://fabricweb.z5.web.core.windows.net/oufr/8.0.0-beta.4/#/examples/layer)
- Removed  `Layer` example which uses `Customizer` to pass `layerProps`. I don't feel this is an example we need or a case which worths callout.
- Minor improvements made to `Layer` examples on hosted scenarios.

#### Focus areas to test

(optional)


packages/react-examples/src/react/Layer/Layer.Customized.Example.tsx
```diff
@@ -1,66 +0,0 @@
-import * as React from 'react';
```
ecraig12345, line 1: It looks like @ThomasMichon originally added this example in #4518, so I'd be curious about his thoughts on its continued relevance (or not) before we remove it.

packages/react-examples/src/react/Layer/Layer.Hosted.Example.tsx
```diff
@@ -92,8 +96,6 @@ const styles = {
           left: '50%',
           top: '50%',
           transform: 'translate(-50%, -50%)',
-          // double quotes required to make the string show up
```
ecraig12345, line 95: Not sure this should be removed? Seems potentially important to help demonstrate the example's point. Also if you're removing this you should remove the entire :before styles.

packages/react-internal/src/components/Layer/Layer.base.tsx
```diff
@@ -90,31 +90,21 @@ export const LayerBase: React.FunctionComponent<ILayerProps> = React.forwardRef<
     };
 
     React.useLayoutEffect(() => {
-      // During the initial render and any hostId updates:
-      //
+      createLayerElement();
```
ecraig12345, line 93: Is this a case where the logic didn't get translated quite correctly compared to the non-FC implementation?

xugao: yes. the bug here is mainly this logic
```
      if (hostId) {
        registerLayer(hostId, createLayerElement);
      }
      // The user didn't provide a hostId, create a new layer element.
      else {
        createLayerElement();
      }
```

We should always call `createLayerElement` otherwise it won't be called. `LayerHost` only `notifyHostChanged` (calls  `createLayerElement`) when it's mounted. I believe the "notification" logic between `Layer` and  `LayerHost` is mainly to make sure the order of mounting these 2 components do not matter

packages/react-examples/src/react/Layer/Layer.Hosted.Example.tsx
xugao: I now render the string in html (above) instead of using css  `after content`. the issue here was `I am a LayerHost with id=layerhost1"` is hard coded string and  here `layerhost1`(host id) needs to be dynamically set

ThomasMichon (approved):
Assuming that this is just a bug fix and not major behavior changes, I'm okay with it. We obviously need to make sure that rendering `Layer` without an override works fine and that rendering a `Layer` with the `Customizer` and an override `hostId` works fine, too.

packages/react-examples/src/react/Layer/Layer.Customized.Example.tsx
xugao: I talked to Thomas. He was originally added because TMP has implemented this case. However we looked at TMP code and the use case is no longer there.

Regardless, I decided keep the example for now instead of removing it. I also updated the `Customizer` deprecation message to be more explicit.

ecraig12345 (approved):

PR #15769, opened 2020-10-29
Point package.json to raw src code

ling1726:
#### Pull request checklist

#### Description of changes

Point package.json of N* packages to raw source code and create a prepublish step for N* modify package.json types and entrypoints to built code before publishing

#### Focus areas to test

(optional)

### DIFF
--- a/scripts/gulp/tasks/bundle.ts
+++ b/scripts/gulp/tasks/bundle.ts
@@ -14,16 +14,18 @@ const packageName = config.package;
 // Clean
 // ----------------------------------------
 
-task('bundle:package:clean', () =>
-  del(
-    [
-      `${paths.packageDist(packageName)}/es/*`,
-      `${paths.packageDist(packageName)}/commonjs/*`,
-      `${paths.packageDist(packageName)}/dts`,
-    ],
+task('bundle:package:clean', () => {
+  const pathsToDelete = [
+    `${paths.packageDist(packageName)}/es/*`,
+    `${paths.packageDist(packageName)}/commonjs/*`,
+    `${paths.packageDist(packageName)}/dts`,
+  ];
+
+  return del(
+    pathsToDelete.map(x => x.replace(/\\/g, '/')), // replace backslash if on windows
     { force: true },
-  ),
-);
+  );
+});
 
 // ----------------------------------------
 // Build
--- a/scripts/lernaAliasNorthstar.js
+++ b/scripts/lernaAliasNorthstar.js
@@ -1,7 +1,18 @@
 const lernaAlias = require('lerna-alias');
+const findGitRoot = require('./monorepo/findGitRoot');
+const path = require('path');
 
 // northstar packages should pull these from npm, not the repo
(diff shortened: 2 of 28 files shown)


scripts/gulp/tasks/bundle.ts
```diff
@@ -14,16 +14,18 @@ const packageName = config.package;
...
-    [
-      `${paths.packageDist(packageName)}/es/*`,
-      `${paths.packageDist(packageName)}/commonjs/*`,
-      `${paths.packageDist(packageName)}/dts`,
-    ],
+task('bundle:package:clean', () => {
+  const pathsToDelete = [
+    `${paths.packageDist(packageName)}/es/*`,
+    `${paths.packageDist(packageName)}/commonjs/*`,
+    `${paths.packageDist(packageName)}/dts`,
```
ecraig12345, line 21: You can probably get rid of the windows fix below by doing this:
```suggestion
    paths.packageDist(packageName, 'es', '*'),
    paths.packageDist(packageName, 'es', '*'),
    paths.packageDist(packageName, 'es', '*'),
```

scripts/lernaAliasNorthstar.js
```diff
@@ -1,28 +1,42 @@
 const lernaAlias = require('lerna-alias');
+const findGitRoot = require('./monorepo/findGitRoot');
+const path = require('path');
 
 // northstar packages should pull these from npm, not the repo
 const excludedPackages = ['@fluentui/date-time-utilities', '@fluentui/dom-utilities', '@fluentui/react-compose'];
 
+function deleteNorthstarAliases(aliases) {
```
ecraig12345, line 8: I'm confused, the whole point of this file was to be lerna-alias *specifically for* things under packages/fluentui.

ling1726: yes, I am trying to remove the aliases by pointing the entrypoints to raw src code, until publish time

ling1726: Since I haven't modified the entrypoints for packages outside of packages/fluentui I still need those aliases

ecraig12345: Ah okay, that makes sense. Can you add a comment about that?

ling1726: that's a good point ! I'll do that

scripts/tasks/ts.ts
```diff
@@ -45,4 +47,7 @@ export const ts = {
     // Use default tsbuildinfo for this variant (since it's the only variant)
     return tscTask({ ...extraOptions, outDir: 'lib', module: 'commonjs' });
   },
+  build: () => {
+    return tscTask({ build: true });
+  },
```
ling1726, line 52: @ecraig12345 any thoughts on this ? I converted `react-codesandbox-northstar` to use tsc -b to be able to take a dep on the N* packages

layershifter:
Closing for housekeeping 🏡 Hope that we will do it someday...

PR #15770, opened 2020-10-29
Allow ariaDescription as a prop on IContextualMenuItem

### LINKED ISSUE #15510: Allow ariaDescription as a prop on IContextualMenuItem
### Environment Information

- **Package version(s)**: all

### Describe the issue:

In office, we set the aria-description directly instead of using aria-describedBy for our tooltips. Can we get support for an ariaDescription optional prop on IContextualMenuItem  so that we can set the aria-label to be the menu item label and the aria-description to be the custom tooltip content? That way it would match the way we set things for all of our buttons throughout the product.

Here is a codepen that's not very helpful but you can see there isn't a prop to set ariaDescription on a menu item.
https://codepen.io/savannahbourgeois/pen/jOrWMMJ

#### Actual behavior:
No prop for setting ariaDescription.

#### Expected behavior:
There is an ariaDescription prop in the same way there is ariaLabel

Allow users a secondary way to set aria-description directly. Feel free to @ me if you need more info but since buttons already support an ariaDescription prop, it would be great if menu items did as well.

SavannahBourgeois:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15510

#### Description of changes
There are a few props that can be set on items to help give screen reader users more context. One of those is aria-DescribedBy which is a string of IDs that point to elements describing the item. The other is aria-Description which is often set as the description of the element or is placed into a hidden span and the ID of that span is added to the list of IDs in aria-DescribedBy. This second approach is better because it will merge the aria-DescribedBy IDs and concat them to be the aria-Description set on the element so you do not overwrite other aria-DescribedBy elements.

With this change, I add ariaDescribedBy and ariaDescription as options on IContextualMenuItem. If the user passes ariaDescription, we will generate a hidden span with the contents of that prop and set the ID of that span to the list of aria-DescribedBy.

Some more in-depth explanation on why I did it the way I did. Each of the contextual menu items is set up slightly different to account for the differences in content. Each of these has a keytip wrapper and if the user passes in keytips, the aria-DescribedBy is updated to point to the element housing the keytip. If we set aria-DescribedBy directly on the element, we would overwrite the possible aria-describedBy set by the keytip. We also can't rely on only passing in the merged aria-describedBy to the keytipProps since not all items have keytips. 

Instead there are two approaches done here to match the current syntax. In the splitButton, we continue to set aria-describedBy to the merged aria properties and include keytipAttributes['aria-describedby']. In the menu button and the anchor, we get a build error if we try to do the same thing saying that 'aria-describedby' is not a prop on keytipAttributes and instead we need to pass the merged aria-describedBy into the contents as a native prop. Since we also pass it as part of the keytip props with {...keytipAttributes}, if the aria-describedBy exists on the keytipAttributes it will override the other instance and properly set the merged aria-describedBy to be the ones we passed in and include the keytip spans. If it does not exist on the keytipAttributes, the passed in value will remain.

I have added an example of different context menu types with ariaDescriptions and verified they are read out by the screen reader when "Hear advanced detail" is selected.

Lastly, I'm not completely sure I made my change file correctly and don't know the proper tags to add. Essentially I would like this to go into Fabric 7 and have it pushed to Fabric 8 once the release is complete. This is an accessibility issue which we need to fix in Office so we don't want to wait until the release of V8

#### Focus areas to test
Contextual menus, keytips on contextual menus, screen reader


SavannahBourgeois:
@jspurlin as FYI

SavannahBourgeois:
@joschect can you take a look some time this week please? We would like to fix our accessibility issue in office. Thanks!

joschect (approved):

PR #15771, opened 2020-10-29
Fixing makeVariantClasses to have an alternative codepath for IE11.

dzearing:
When styling refers to css variables, IE11 won't handle it. This change will resolve the variables into static values at runtime, allowing the styling to refer to variables as long as they're resolvable from the variant tokens or from theme tokens.

I'm going to put the PR out first to see how bundle size is affected by adding the IE11 workaround directly to the bundle so that customers don't have to do any additional work. If the bundle size hit is too large, I will approach this by keeping it an injectable polyfill on demand.


packages/react-button/src/components/Button/useButtonClasses.ts
```diff
@@ -40,447 +40,450 @@ export const ButtonSizeVariants = {
...
 
-      // TODO: no references to palette.
-      ghost: {
-        background: 'transparent',
-        borderColor: 'transparent',
-        contentColor: palette?.neutralPrimary,
-        fontWeight: 'normal',
-        iconColor: palette?.themeDarkAlt,
-        forcedColorAdjust: 'none',
+      // focused: {
```
dzearing, line 404: I'm not sure how this is required, and also, it's not passing typings.

packages/react-button/src/components/SplitButton/useSplitButtonClasses.tsx
```diff
@@ -58,8 +63,8 @@ export const useSplitButtonClasses = makeVariantClasses<SplitButtonState, SplitB
       backgroundColor: 'var(--button-dividerColor)',
       position: 'absolute',
       right: menuButtonWidth,
-      top: 'calc(100% - var(--button-dividerLength, 100% + 8px))',
-      bottom: 'calc(100% - var(--button-dividerLength, 100% + 8px))',
+      top: 8,
```
dzearing, line 66: Technically, this should be supported now! Wasn't earlier.

packages/react-theme-provider/etc/react-theme-provider.api.md
```diff
@@ -36,23 +37,23 @@ export { IStyleFunctionOrObject }
...
 // @public
-export const makeVariantClasses: <TState = {}, TVariants = Record<string, any>>(options: MakeVariantClassesOptions<TVariants>) => (state: TState, theme?: Theme | undefined, renderer?: import(".").StyleRenderer | undefined) => void;
+export const makeVariantClasses: <TState = {}, TVariants = Record<string, any>>(options: MakeVariantClassesOptions<TVariants>) => (state: TState, theme?: Theme | undefined, renderer?: StyleRenderer | undefined) => void;
 
 // @public
 export type MakeVariantClassesOptions<TVariants = Variants> = {
     name?: string;
     prefix?: string;
-    styles?: Record<string, IStyle> | ((theme: Theme) => Record<string, IStyle>);
+    styles?: Record<string, IStyle_2> | ((theme: Theme) => Record<string, IStyle_2>);
```
dzearing, line 56: hmm. weird...

apps/vr-tests/src/stories/Button.stories.tsx
```diff
@@ -1,13 +1,13 @@
...
 import Screener, { Steps } from 'screener-storybook/src/screener';
 import { storiesOf } from '@storybook/react';
-import { Button } from '@fluentui/react-button';
+import { Button } from '@fluentui/react';
 import { AddIcon } from '@fluentui/react-icons-mdl2';
 import { TeamsTheme } from '@fluentui/storybook/lib/themes/v8/index';
 import { withThemeProvider } from '@fluentui/storybook';
 import { FabricDecorator, FabricDecoratorFullWidth } from '../utilities/index';
 
-storiesOf('Button Next', module)
```
khmakoto, line 10: Could you similarly remove the `Next` moniker in `CompoundButton`, `MenuButton`, `SplitButton` and `ToggleButton` stories as well since you're already doing it here?

khmakoto: We also probably want to check the rest of this file for usage of `Next` since I think there were multiple stories in each one.

packages/react-button/src/components/Button/Button.types.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
khmakoto, line 1: Why do we need this here?

packages/react-button/src/components/Button/Button.types.ts
```diff
@@ -145,22 +151,29 @@ export type ButtonTokens = ColorTokens &
...
+  size_larger?: TTokens;
+  size_largest?: TTokens;
+  size_regular?: TTokens;
+  size_small?: TTokens;
+  size_smaller?: TTokens;
+  size_smallest?: TTokens;
+
   transparent?: TTokens;
+
+  [key: string]: TTokens | undefined;
```
khmakoto, line 178: Should we really allow this? I think it undermines type safety.

packages/react-button/src/components/Button/useButtonClasses.ts
khmakoto: Should be passing typings now if you merge with the latest changes from master.

packages/react-theme-provider/src/callOrReturn.ts
```diff
@@ -0,0 +1,9 @@
+// type VariantClassesHook = <TState>(state: TState) =>
+/**
+ * Calls a function with the argument, or returns the given object.
+ * @param objOrFunc - Function or object.
+ * @param argument - Argument to pass if a function is provided.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const callOrReturn = (objOrFunc: any, argument: any) =>
```
khmakoto, line 8: Is this something that should be in `react-theme-provider`? Seems like a more general utility.

packages/react-theme-provider/src/makeVariantClasses.ts
```diff
@@ -3,6 +3,9 @@ import { tokensToStyleObject } from './tokensToStyleObject';
 import { Variants, Theme } from '@fluentui/theme';
 import { IStyle } from '@fluentui/merge-styles';
 import { makeClasses } from './makeClasses';
+import { isIE11 } from '@fluentui/utilities';
+import { StyleRenderer } from './styleRenderers/types';
+import { useVariantClassesIE11Override } from './useVariantClassesIE11Override';
```
khmakoto, line 9: You should remove `callOrReturn` in line 15 since you're already placing it in another file.

dzearing:
The bundle regression is bad, but it means that IE11 will work out of the box with 0 configuration for the customer. It's factored in a way we can remove later.

apps/vr-tests/src/stories/Button.stories.tsx
dzearing: done.

packages/react-theme-provider/src/callOrReturn.ts
dzearing: it's not exported by theme-provider and is only used there.

dzearing: But even then, moving it would not be a breaking change and we could do later if we find more use cases.

packages/react-theme-provider/src/makeVariantClasses.ts
dzearing: done.

packages/react-button/src/components/Button/Button.types.ts
dzearing: removed.

dzearing: it was because of size_small, etc. Moved to per line so it's more scoped and obvious.

packages/react-button/src/components/Button/Button.types.ts
dzearing: How would someone add their own variants?

dzearing: you're possibly right but we need a good story for how someone would extend.

packages/react-theme-provider/src/ie11.ts
```diff
@@ -0,0 +1,8 @@
+import { GlobalSettings, isIE11 } from '@fluentui/utilities';
+import { UseVariantClassesOverride, UseVariantClassesOverrideKey } from './makeVariantClasses';
+import { useVariantClassesIE11Override } from './useVariantClassesIE11Override';
+
+// Initialize the global settings override.
+if (isIE11()) {
+  GlobalSettings.setValue<UseVariantClassesOverride>(UseVariantClassesOverrideKey, useVariantClassesIE11Override);
+}
```
xugao, line 8: I don't think we need to drop this on global

packages/react-theme-provider/src/replaceCSSVariables.test.ts
```diff
@@ -0,0 +1,137 @@
+// '0 0 0 var(--button-focusInnerWidth, 1px) var(--button-focusInnerColor, white) inset'
```
xugao, line 1: remove

packages/react-theme-provider/src/makeStyles.ts
```diff
@@ -46,9 +15,10 @@ const graphSet = (graphNode: Map<any, any>, path: any[], value: any) => {
  */
 export function makeStyles<TStyleSet extends { [key: string]: IStyle }>(
   styleOrFunction: TStyleSet | ((theme: Theme) => TStyleSet),
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  graph: Map<any, any> = new Map(),
```
xugao, line 19:
```suggestion
  cache: Map<any, any> = new Map(),
```

packages/react-button/src/components/CompoundButton/useCompoundButtonClasses.ts
```diff
@@ -1,17 +1,24 @@
...
 
-export const useClasses = makeVariantClasses<CompoundButtonState, CompoundButtonVariants>({
+/**
+ * CompoundButton classes hook for mixing classes into a mutable state object.
+ */
+export const useCompoundButtonClasses = makeVariantClasses<CompoundButtonState, CompoundButtonVariants>({
   name: 'CompoundButton',
   prefix: '--button',
-  styles: {
+  styles: concatStyleSets(buttonStyles, {
```
xugao, line 21: we should change `styles` and `variants` to support arrays, then do concatStyleSets / merge inside `makeVariantClasses`

xugao (commented):
perf looks good to me. the potential regression occurred because new named function `useVariantClasses` is added.

Are we adding IE11 screener tests in separate PR?

packages/react-theme-provider/src/ie11.ts
dzearing: oh this should be removed.

layershifter:
We went with a different approach (#16631), but this will be useful for IE11 implementation. Closing for housekeeping 🏡

PR #15773, opened 2020-10-29
Fix React as external config

### LINKED ISSUE #10349: office-ui-fabric-react.umd.js requires "ReactDOM"
### Environment Information

- **Package version(s)**: office-ui-fabric-react@7.29.0

### Reproduce:

- `npm i office-ui-fabric-react`
- Open `node_modules\office-ui-fabric-react\dist\office-ui-fabric-react.umd.js` and look for `require("ReactDOM")` and `define(["React","ReactDOM"]`.

#### Actual behavior:

ReactDOM does not exist (anymore).

#### Expected behavior:

It probably should require [`react-dom`](https://www.npmjs.com/package/react-dom).
And about `define(["React"`: Shouldn't `React` be lowercased here?

Requested priority: I think this is a blocking issue since the UMD package just won't work. But since nobody noticed so far and I am no expert at all, I just let you decide.

filoxo:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #10349

#### Description of changes

As mentioned in the #10349, fluentui-react's build is configured to be able to use React as an external but *only* as a global. This change fixes that by using the proper config that will allow using external React as a global as well as with other methodologies (such as with SystemJS for in-browser module resolution).

#### Focus areas to test

- Inspect the output bundle and verify that:
  - `require("React"),require("ReactDOM")` is no longer present
  -  the correct casing `require("react"),require("react-dom")` is included
- Verify that it still works with React as a global (eg. react provided via a CDN)

### Notes

I wasn't able to find how to actually build the final bundle to be able to test these changes as I've described above.

### DIFF
--- a/scripts/webpack/webpack-resources.js
+++ b/scripts/webpack/webpack-resources.js
@@ -205,8 +205,16 @@ module.exports = {
               : output,
 
           externals: {
-            react: 'React',
-            'react-dom': 'ReactDOM',
+            react: {
+              commonjs: 'react',
+              amd: 'react',
+              root: 'React',
+            },
+            'react-dom': {
+              commonjs: 'react-dom',
+              amd: 'react-dom',
+              root: 'ReactDOM',
+            },
           },
 
           resolve: {


ecraig12345:
Thanks for doing this! We'll probably want to do it in the other packages as well--if you're interested I can explain how to do that now, or I can just handle it later.

Looks like the build is failing because you're missing a change file--you'll need to run `yarn change` from the root of the repo then push again. (The other failure is a known intermittent issue.)

filoxo:
I think I'd be open to fixing the other package's configs as well. Is there any info/docs on how I can build the final assets that are published? Or should I look in the build pipelines somewhere? That way I can follow the 'steps to verify' in the PR description.

I did include a [change file](https://github.com/microsoft/fluentui/pull/15773/files#diff-7aa7f5250db9ea3432511ce4300f592b3f0bf453e86232f4375a1a44e924c532) but it was generated before the major refactor that happened recently. Should I remove that and regenerate it instead?

layershifter:
@filoxo can you please merge master and recreate a change file? 🐱

filoxo:
@layershifter done.

ecraig12345:
Sorry for not following up on this sooner! To ensure this is fixed in all packages, I'd recommend making the change here instead:
https://github.com/microsoft/fluentui/blob/f55b5ce50951f599976ee6ef26b4a2df4dae735a/scripts/webpack/webpack-resources.js#L207-L210

(then revert the change in the packages/react webpack config)

filoxo:
I tried regenerating the change file but the command returns with `No change files are needed`. 🤷🏼

ecraig12345 (approved):

PR #15777, opened 2020-10-29
Use v8 react-button hooks in v0 button

levithomason:
# ⚠️ This PR is a WIP

Validating v8 hooks in v0.  Please wait for review.

### TODO 

- [ ] Run v8 examples and fix broken things (due to hook updates)


packages/react-button/src/components/Button/useButtonState.ts
```diff
@@ -32,5 +32,8 @@ export const useButtonState = (draftState: ButtonState) => {
     }
   }
 
-  draftState.disabled = draftState['aria-disabled'] = draftState.disabled || draftState.loading;
+  // TODO: v0 buttons are not disabled while loading, removed `disabled = loading` logic here
```
jurokapsiar, line 35: not disabling buttons while loading was intentional - while the UI is loading, it is useful to have the button still focused, because quite often, there is no other suitable place to move focus to.

packages/react-avatar/src/components/Image/useImage.tsx
```diff
@@ -1,15 +1,13 @@
...
-import { getSlots, makeMergeProps } from '@fluentui/react-compose/lib/next/index';
+import { getSlots, mergeProps } from '@fluentui/react-compose/lib/next/index';
 
 export const renderImage = (state: ImageState) => {
   const { slots, slotProps } = getSlots(state);
 
   return <slots.root {...slotProps.root} />;
 };
 
-const mergeProps = makeMergeProps<ImageState>();
```
levithomason, line 11: TODO: Add tests to show why we cannot rely on a static list of props to deeply merge.

ecraig12345:
@levithomason Guessing this can be closed now that we're taking a totally different approach?

ecraig12345:
Closing since we're no longer considering this approach

PR #15780, opened 2020-10-29
MenuButton: Extrapolating expanded functionality to be provided via function props and context

### LINKED ISSUE #15258: MenuButton: Make expanded functionality not rely on expectations of ContextualMenu being passed as a slot replacement
We need to find a way so that other things can be passed as slot replacements and the component should still work as expected.

*Note*: Currently exploring using `Context` to handle this.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15258

#### Description of changes

This PR extrapolates the `expanded` state in `MenuButton` to be provided via function props if a function is passed though the `menu` slot or via context if JSX is what's passed through the slot.

_Via function:_

```tsx
const FunctionMenu = (props: MinimalMenuProps) => {
  const { hidden, ...rest } = props;
  return !hidden ? <ContextualMenu {...(rest as IContextualMenuProps)} {...menuProps} /> : null;
};

{...}

<MenuButton menu={FunctionMenu} />
```

_Via JSX:_

```tsx
const JSXMenu = () => <ContextualMenu {...(useMenuContext() as IContextualMenuProps)} {...menuProps} />;

{...}

<MenuButton menu={<JSXMenu />} />
```

#### Focus areas to test

(optional)


khmakoto:
One question left to answer is if we should conditionally render the `menu` slot based on `expanded` or left that decision to the user now that they can import the context themselves.

packages/react-button/src/components/MenuButton/useMenuButtonClasses.ts
```diff
@@ -16,7 +17,7 @@ export const useMenuButtonClasses = makeVariantClasses<MenuButtonState, MenuButt
     root: [
       GlobalClassNames.root,
       {
-        // This seems like a bad selector.
+        // This seems likea  bad selector.
```
dzearing, line 20:
```suggestion
        // This seems like a  bad selector.
```

dzearing: Or remove the selector. It seems bad. :)

packages/react-examples/src/react-button/SplitButton/SplitButton.stories.tsx
```diff
@@ -16,30 +17,34 @@ const menuProps = {
   ],
 };
 
+const Menu = (defaultSlotType: string, props: MinimalMenuProps) => {
```
dzearing, line 20: Is this a render function `(Component, props) => JSX`? If so, is `string` the right type?

packages/react-button/src/components/MenuButton/useExpanded.tsx
```diff
@@ -68,17 +68,28 @@ export const useExpanded = <TDraftState extends ExpandedState>(draftState: TDraf
...
-      setExpandedValue(false);
+    setExpandedValue(false);
 
-      // TODO: should we re-focus the root?
-    }, [onMenuDismiss, setExpandedValue]),
+    // TODO: should we re-focus the root?
+  }, [onMenuDismiss, setExpandedValue]);
+
+  // Assign extra props to the menu slot.
+  draftState.menu = {
```
dzearing, line 80: Aren't you overwriting the menu object here?

packages/react-examples/src/react-button/SplitButton/SplitButton.stories.tsx
khmakoto: This is the expected function call but I think I can see a way of only passing `props` here. I'll investigate.

packages/react-button/src/components/MenuButton/useExpanded.tsx
khmakoto: I'm only using the already passed `Menu` object (which is passed as `draftState.menu.children`) and putting it inside the `ExpandedContext.Provider`. So in essence the only thing I'm doing here is making the user able to use the `Context` as they see fit.

packages/react-button/src/components/MenuButton/useMenuButtonClasses.ts
khmakoto: Thanks! I'll change this.

packages/react-button/src/components/MenuButton/useExpanded.tsx
khmakoto: Removing `Context` and passing `expand` as a prop.

packages/react-button/src/components/MenuButton/useMenuButtonClasses.ts
khmakoto: Changed!

packages/react-examples/src/react-button/SplitButton/SplitButton.stories.tsx
khmakoto: Made it so that only `props` need to be passed.

packages/react-button/etc/react-button.api.md
```diff
@@ -229,7 +233,7 @@ export interface SplitButtonProps extends ButtonProps, MenuButtonProps {
 export const splitButtonShorthandProps: string[];
 
 // @public (undocumented)
-export interface SplitButtonState extends Omit<SplitButtonProps, 'menu'>, MenuButtonState {
+export interface SplitButtonState extends Omit<SplitButtonProps, 'omit'>, MenuButtonState {
```
dzearing, line 236: Omit omit?

packages/react-button/src/components/SplitButton/SplitButton.types.ts
```diff
@@ -23,10 +23,7 @@ export interface SplitButtonProps extends ButtonProps, MenuButtonProps {
   menuButton?: ShorthandProps;
 }
 
-/**
- * {@docCategory Button}
- */
-export interface SplitButtonState extends Omit<SplitButtonProps, 'menu'>, MenuButtonState {
+export interface SplitButtonState extends Omit<SplitButtonProps, 'omit'>, MenuButtonState {
```
dzearing, line 26: Here's the `omit` Omit... not sure this was intentional.

packages/react-button/etc/react-button.api.md
khmakoto: Oh lol, that should be menu, but maybe it's not even needed if the build is not complaining.

dzearing (approved):

PR #15782, opened 2020-10-30
Create Tabs as a copy of Pivot

behowell:
#### Pull request checklist

#### Description of changes

This change clones the Pivot component to Tabs. For now, both components are exported side-by-side from the react-tabs package. Future PRs will improve the API and implementation of the Tabs component.

This is mostly a direct copy + renaming a few components. I've also removed some deprecated props and types from the copies.

The first commit in this branch copied all of the files, and then subsequent commits renamed and updated the components. Use this link to see just the renames/updates:
https://github.com/microsoft/fluentui/pull/15782/files/4b53efa3f91d4da7be9df3769219c74391b11ccb..HEAD


behowell:
The first commit in this branch copied all of the files, and then subsequent commits renamed and updated the components. Use this link to see just the renames/updates:
https://github.com/microsoft/fluentui/pull/15782/files/4b53efa3f91d4da7be9df3769219c74391b11ccb..HEAD

packages/react-tabs/src/next/components/Tabs/TabItem.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+import { getNativeProps, divProperties, initializeComponentRef } from '@fluentui/utilities';
+import { TabItemProps } from './TabItem.types';
+
+// const COMPONENT_NAME = 'TabItem';
+
+export class TabItem extends React.Component<TabItemProps, {}> {
```
ecraig12345, line 7: Sorry if I missed some discussion here but is there any way we can avoid entirely duplicating the code? The duplication almost guarantees it will get out of sync and bug fixes will be missed in one version or the other.

behowell: Pivot is going to be deprecated, and Tabs is going to be further modified after this PR for convergence work. In the interim, Pivot needs to continue to work without breaking changes. There are two basic approaches:
1. Rename Pivot to Tabs, and keep Pivot as a "thin" compat layer that accounts for all of the breaking changes.
   * The compat layer can get complicated, and is its own engineering problem. For example, Pivot uses old CommandButton for its tab buttons, and exposes that through its API via TabItem.headerButtonProps. If we replace that with the new Button, we need to keep headerButtonProps working. Possibly solvable, but potentially a lot of extra work.
2. Fork Pivot to /next/Tabs, and do the API breaking convergence work on the Tabs. In the future, Pivot will be deprecated and moved to /compat. Any new features and updates will be added to Tabs, not Pivot.

I was originally going to do 1, but after talking to Justin and David, decided to do 2, which is what this PR is.

ecraig12345: Got it, sorry. This would be a lot less confusing if we were in the office. :(

packages/react-examples/src/react-tabs/Tabs/Tabs.doc.tsx
```diff
@@ -0,0 +1,90 @@
+import * as React from 'react';
```
ecraig12345, line 1: Let's skip adding this file and the markdown files for now, since they're only used in the legacy demo app and the public site (neither of which really needs to include Tabs yet)

ecraig12345 (approved):
This looks okay once the extra doc files are removed, just please keep an eye out for Pivot PRs to ensure the changes are made in Tabs too.

packages/react-tabs/src/next/components/Tabs/Tabs.styles.ts
```diff
@@ -0,0 +1,280 @@
...
+  icon: 'ms-Pivot-icon',
+  tabIsSelected: 'is-selected',
+  tab: 'ms-Pivot-link',
+  tabContent: 'ms-Pivot-linkContent',
+  root: 'ms-Pivot',
+  rootIsLarge: 'ms-Pivot--large',
+  rootIsTabs: 'ms-Pivot--tabs',
+  text: 'ms-Pivot-text',
+  tabInMenu: 'ms-Pivot-linkInMenu',
+  overflowMenuButton: 'ms-Pivot-overflowMenuButton',
```
xugao, line 23: should we name these classNames too? maybe add a TODO and do  in future pr

packages/react-tabs/src/next/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from '../index';
 export { Pivot } from './Pivot';
+export * from './Tabs';
```
xugao, line 3: let's not export from `index` yet until we are done will all the breaking changes and ready to deprecate `Pivot` in favor of `Tabs`

xugao (approved):
Feel free to remove existing Pivot code in `/next` folder

xugao (approved):

PR #15783, opened 2020-10-30
[isConformant] Enabling consistent-callback-names test

czearing:
### Pull request checklist

### Description of changes
Fixing a bug that prohibited the `consistent-callback-names` conformance test from running and disabling it where necessary.

#### What this test does
1. Checks the component's props to find potential callbacks by searching for "on".
2. Given the set of callbacks it will check if any of them end with "ed" (onFocused, onChanged, ...)
3. If the test finds a callback ending with "ed" it will trigger the **defaultErrorMessage** and suggest for them to change.

NOTE: In the edge case that the test isn't accurate for a specific prop, it can individually be disabled by setting `ignoreProps`.

####  `consistent-callback-names`
```jsx
   'consistent-callback-names': (componentInfo: ComponentDoc, testInfo: IsConformantOptions) => {
    it(`has consistent custom callback names`, () => {
        const { testOptions = {} } = testInfo;
        const propNames = Object.keys(componentInfo.props);
        const ignoreProps = testOptions['consistent-callback-names']?.ignoreProps || [];

        for (const propName of propNames) {
          if (!ignoreProps.includes(propName) && /^on(?!Render[A-Z])[A-Z]/.test(propName)) {
            const words = propName.slice(2).match(/[A-Z][a-z]+/g);

            if (words) {
              // Make sure last word doesn't end with ed
              const lastWord = words[words.length - 1];
              expect(lastWord.endsWith('ed')).toBe(false);
            }
          }
        }
    });
  },
```
####  defaultErrorMessages
image


packages/react-conformance/src/defaultTests.tsx
```diff
@@ -244,9 +244,7 @@ export const defaultTests: TestObject = {
         const propNames = Object.keys(componentInfo.props);
         const ignoreProps = testOptions['consistent-callback-names']?.ignoreProps || [];
 
-        // Object.keys shouldn't be here and is causing this test not to run.
-        // TODO: Remove Object.keys after the package move and disable the test where necessary.
-        for (const propName of Object.keys(propNames)) {
+        for (const propName of propNames) {
```
ecraig12345, line 247: Oops, good catch!

packages/react-focus/src/components/FocusZone/FocusZone.test.tsx
```diff
@@ -78,10 +78,15 @@ describe('FocusZone', () => {
...
     Component: FocusZone,
     displayName: 'FocusZone',
-    disabledTests: [
-      // Their is no existing top level FocusZone.ts file.
-      'has-top-level-file',
-    ],
+    // Disabled Tests:
+    //
+    // has-top-level-file
+    // Their is no existing top level FocusZone.ts file.
```
ecraig12345, line 84:
```suggestion
    // There is no existing top level FocusZone.ts file.
```

ecraig12345: Also we should probably just create this file (it won't hurt anything and makes this package match the convention in case anyone expects that)

packages/react-internal/src/components/Callout/Callout.test.tsx
```diff
@@ -51,7 +51,16 @@ describe('Callout', () => {
...
+    // Disabled Tests:
+    //
+    // component-handles-ref
+    // Problem: Ref isn't receiving the correct element.
+    // Solution: Investigate the refs implementation.
+    //
+    // consistent-callback-names
+    // Problem: Contains onLayerMounted and onPositioned types.
+    // Solution: Remove the inconsistent callback names.
+    disabledTests: ['component-handles-ref', 'consistent-callback-names'],
```
ecraig12345, line 63: Instead of disabling this, it would probably be better to add `testOptions['consistent-callback-names'].ignoreProps` (same in other cases too)

PR #15784, opened 2020-10-30
Jspurlin/combo box focus fix 7.0

jspurlin:
#### Pull request checklist

#### Description of changes
Previous to this change there was a case when a controlled comboBox could move focus as part of either `onItemClick` or `onChange` prop callbacks, but then the comboBox would refocus the input  afterwards leading to unexpected focus behavior.

This PR allows the comboBox to handle its internal focus handling and potential menu closing prior to calling the `onItemClick` and `onChange` callbacks. This allows the comboBox to deal with its inner focus handling before communicating outwardly fixing the potential unexpected stealing of focus back to the input

#### Focus areas to test
Verified that the comboBox functionality didn't change and the comboBox no longer steals focus back if it was moved as part of `onItemClick` or `onChange`

### DIFF
--- a/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
+++ b/packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
@@ -759,11 +759,11 @@ export class ComboBox extends React.Component<IComboBoxProps, IComboBoxState> {
           return { ...item, index };
         })
         .filter(
-          (option) =>
+          option =>
             option.itemType !== SelectableOptionMenuItemType.Header &&
             option.itemType !== SelectableOptionMenuItemType.Divider,
         )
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/components/ComboBox/ComboBox.tsx
```diff
@@ -1601,6 +1603,11 @@ export class ComboBox extends React.Component<IComboBoxProps, IComboBoxState> {
           isOpen: false,
         });
       }
+
+      // Continue processing the click only after
+      // performing menu close / control focus(inner working)
+      onItemClick && onItemClick(ev, item, index);
```
jspurlin, line 1609: The only things changed as part of this PR is moving lines 1593 and 1594 to here (lines 1609 and 1610)

joschect (approved):

PR #15785, opened 2020-10-30
Add preventDefault logic for toggle button label area

### LINKED ISSUE #13721: The toggle should be rendered as inline-block instead of block
### Environment Information

- **Package version(s)**: https://developer.microsoft.com/en-us/fluentui#/controls/web/toggle

#### Actual behavior:

image

In the above screenshot, the red lines are clickable and will toggle the toggle, which is unexpected and inconsistent because the green lines are not clickable.

Accidentally clicking on those red lines causing unexpected toggling is very annoying.

#### Expected behavior:

image

only these green areas should be clickable

Are you willing to submit a PR to fix? YES

xw310:
#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #13721

#### Description of changes

Prevent clicking on label area from changing the state of toggle.

#### Focus areas to test

(optional)

### DIFF
--- a/packages/react-toggle/src/components/Toggle/Toggle.base.tsx
+++ b/packages/react-toggle/src/components/Toggle/Toggle.base.tsx
@@ -136,7 +136,7 @@ export const ToggleBase: React.FunctionComponent<IToggleProps> = React.forwardRe
 
     return (
       <RootType ref={forwardedRef as React.Ref<HTMLDivElement>} {...slotProps.root}>
-        {label && <Label {...slotProps.label} />}
+        {label && <Label {...slotProps.label} onClick={(ev: React.MouseEvent<HTMLElement>) => {ev.preventDefault();}}/>}
         <div {...slotProps.container}>
           <button {...slotProps.pill}>
             <span {...slotProps.thumb} />
--- a/packages/react-toggle/src/components/Toggle/Toggle.base.tsx
+++ b/packages/react-toggle/src/components/Toggle/Toggle.base.tsx
@@ -93,6 +93,10 @@ export const ToggleBase: React.FunctionComponent<IToggleProps> = React.forwardRe
       }
     };
 
+    const onClickLable = (ev: React.MouseEvent<HTMLElement>) => {
+      ev.preventDefault();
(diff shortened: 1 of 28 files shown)


xugao:
@xw310 - apologize for the delayed review and thank you for making the fix! 
The fix looks good. However, I believe you missed to update the tests - try run `yarn update-snapshots` under both `react-toggle` and `react-examples` packages

xw310:
Hi Xu,

Thanks for your reply! I tried to run yarn update-snapshots under both
packages you mentioned, but it seems that some snapshots are not correctly
updated under react-examples packages. Do you have some ideas why? Looking
forward to your reply!

Pipeline information here:
https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=152859&view=logs&j=258ec178-2d8b-5611-7b9b-60c5c95dae55&t=e99f753c-3f64-5289-3575-c1a650af1617&l=3280


Best regards,
Xinyang

xugao:
@xw310 - looks like you need to pull latest master then re-run the tests

packages/react-toggle/src/components/Toggle/Toggle.base.tsx
```diff
@@ -103,6 +107,7 @@ export const ToggleBase: React.FunctionComponent<IToggleProps> = React.forwardRe
         className: classNames.label,
         htmlFor: id,
         id: labelId,
+        onClick: onClickLable,
```
mbest, line 110: It's spelled `label` not `lable`

mbest:
Just from looking at the code, this would seem to make the whole label non-responsive.

xugao:
@mbest - I thought a bit more on what the right way to fix this issue should be after your comment. I agree we should **not** fix by making the label click no-op (and user has to click on the switch to toggle). Instead, we should address this by tweaking the css of the label & switch. Let me take a stab at fixing this. @xw310 - i will close your PR for now since it seems inactive.

xugao:
@xw310 @mbest - I made an alternative fix here: #16195. feel free to give a look, thanks!

PR #15797, opened 2020-10-30
SplitButton: Fix ariaDescription behavior so that it is correctly picked up by screen-readers

### LINKED ISSUE #15721: SplitButton - Narrator does not read ariaDescription
### Environment Information

- **Package version(s)**: office-ui-fabric-react 7.144.1
- **Browser and OS versions**: Microsoft Edge Version 88.0.673.0 (Official build) dev (64-bit), Windows 10 Version 1909 Build 18363.1139

Turn on Windows Narrator. Open the following codepen and focus the split button:
https://codepen.io/savannahbourgeois/pen/RwRpvvo

#### Actual behavior:

On focus, Windows Narrator reads: "Standard; split button; collapsed".
The `ariaDescription` value is not read.

#### Expected behavior:

Narrator should also read the value of `ariaDescription`:
"Standard; split button; collapsed; Sample Extra Description"

Requested priority: High. This is an accessibility bug affected by Microsoft internal SLA.

Products/sites affected: Power Query Online, Power Apps, Power BI, Power Automate (potentially all of Office Online as well)

Additional information from Office Online team:

> In FluentUI, when the split button has an ariaDescription passed, the ariaDescription is put into a span and they set aria-describedBy to point to that span. This works in the case of most buttons, however for the split button, the aria-describedBy is getting set on the inner primary button instead of the outer div that receives focus. 

>To fix this, it should hopefully be a one-line change in Fluent. They need to add ariaDescription to buttonProps https://github.com/microsoft/fluentui/blob/0b370e43432789ee9b170f60a8276b594493ca7a/packages/react-internal/src/compat/components/Button/BaseButton.tsx#L227 which is then passed to onRenderSplitButtonContent https://github.com/microsoft/fluentui/blob/0b370e43432789ee9b170f60a8276b594493ca7a/packages/react-internal/src/compat/components/Button/BaseButton.tsx#L253

>Then in onRenderSplitButtonContent, you will need to ensure that ariaDescribedBy https://github.com/microsoft/fluentui/blob/0b370e43432789ee9b170f60a8276b594493ca7a/packages/react-internal/src/compat/components/Button/BaseButton.tsx#L602 is set properly to the rendered span for the ariaDescription. I'm not sure how mergeAriaAttributeValues works but it is possible that there will need to be an additional step since it looks like ariaDescription is getting set directly to a string in the line linked above when aria-DescribedBy takes in an ID.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15721

#### Description of changes

This PR fixes an issue with split buttons where `ariaDescription` was being incorrectly passed into `aria-describedby` instead of providing the correct id there, which in turn was causing the description to not be read.

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-internal-2020-10-30-13-51-44-splitButtonAriaDescription.json
+++ b/change/@fluentui-react-internal-2020-10-30-13-51-44-splitButtonAriaDescription.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "SplitButton: Fix ariaDescription behavior so that it is correctly picked up by screen-readers.",
+  "packageName": "@fluentui/react-internal",
+  "email": "humbertomakotomorimoto@gmail.com",
+  "dependentChangeType": "patch",
+  "date": "2020-10-30T20:51:44.033Z"
+}
--- a/packages/react-internal/src/compat/components/Button/BaseButton.tsx
+++ b/packages/react-internal/src/compat/components/Button/BaseButton.tsx
@@ -599,8 +599,6 @@ export class BaseButton extends React.Component<IBaseButtonProps, IBaseButtonSta
       'data-is-focusable': false,
     });
 
-    const ariaDescribedBy = buttonProps.ariaDescription;
-
     if (keytipProps && menuProps) {
       keytipProps = this._getMemoizedMenuButtonKeytipProps(keytipProps);
     }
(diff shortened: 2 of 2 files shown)


khmakoto:
@msft-github-bot merge in 1 minute.

tomi-msft (approved):

PR #15800, opened 2020-10-30
Codemods: Fix an issue where files would not save synchronously.

joschect:
#### Pull request checklist

#### Description of changes

(give an overview)

#### Focus areas to test

(optional)

### DIFF
--- a/packages/codemods/src/modRunner/tests/upgrade.test.ts
+++ b/packages/codemods/src/modRunner/tests/upgrade.test.ts
@@ -1,4 +1,4 @@
-import { _upgradeTest } from '../../upgrade';
+import { _upgradeInternal } from '../../upgrade';
 import * as path from 'path';
 import * as fs from 'fs-extra';
 import * as os from 'os';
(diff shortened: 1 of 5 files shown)


jdhuntington (approved):

packages/codemods/src/modRunner/tests/upgrade.test.ts
```diff
@@ -0,0 +1,49 @@
...
+const tempDir = os.tmpdir();
+const tempPath = path.join(tempDir, 'codemods');
+const startingPath = process.cwd();
+
+describe('modRunner tests', () => {
+  it('saves syncronously correctly', () => {
+    const saveAsync = jest.fn();
+    const saveSync = jest.fn();
+    try {
+      const temp = fs.mkdtempSync(tempPath);
```
ecraig12345, line 16: It's probably a good idea to clean this up in `finally` (or move the temp dir creation/cleanup to a beforeEach/afterEach). Temp directories do get automatically cleaned up eventually but not always very quickly.

packages/codemods/src/upgrade.ts
```diff
@@ -4,11 +4,18 @@ import { Logger } from './modRunner/logger';
...
 const logger: Logger = console;
-export function upgrade(options: CommandParserResult) {
+
+interface UpgradeFunctions {
+  getTsConfigs: () => string[];
+  saveSync: (file: { saveSync: () => void }) => void;
+  saveAsync: (project: { save: () => void }) => void;
+}
+
+export function _upgradeTest(options: CommandParserResult, fns: UpgradeFunctions) {
```
ecraig12345, line 14: Might be better to call this something like `_upgradeInternal` since it's not just for tests.

packages/codemods/src/modRunner/tests/upgrade.test.ts
joschect: I added a `beforeEach` and `afterEach` to handle the folder cleanup and creation. Good idea!

ecraig12345 (approved):

PR #15802, opened 2020-10-31
Tabs: Add screener and perf tests

behowell:
#### Pull request checklist

#### Description of changes

Add screener and perf tests for Tabs. The screener tests are currently just copies of Pivot tests.


ecraig12345 (approved):

ecraig12345:
Looks like there's a version mismatch with some of the added/sorted deps. Try merging with master then running `yarn syncpack:fix`.

ecraig12345:
Not sure if syncpack:fix will do the right thing in this case but it's worth a try? Or manually fix the erorrs.

xugao (approved):

PR #15804, opened 2020-11-02
Correctly set focus in FocusTrapZone even if focus event does not bubble

### LINKED ISSUE #15530: FocusTrapZone does not tab properly when using preact
### Environment Information

- **Package version(s)**: 7.107.0
- **Browser and OS versions**: Edge Dev, Windows 10

https://924v8.csb.app/

Due to this requiring specific setup (using preact instead of react) I have reproed using CodeSandbox instead of CodePen

#### Actual behavior:

Pressing tab while focusing the last element in the modal keeps tab focus at the last item

#### Expected behavior:

Pressing tab while focusing the last element should focus the first tabbable element in the window

Requested priority: Normal

Products/sites affected: M365 in-app purchase (Dime)

taurheim:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15530

#### Description of changes

Currently FocusTrapZone only sets its internal focus state onFocus. In some cases if events do not bubble correctly, only onFocusCapture will be called, causing the FocusTrapZone to get "stuck" at the end and not loop back to the front. See more information in issue #15530

#### Focus areas to test

FocusTrapZone, Modal dialog usage of FocusTrapZone


layershifter:
@taurheim it seems that there failing unit tests due this change. Can you please check?

taurheim:
Hi @layershifter, the error in the build is "ERROR: there is an invalid change type detected @fluentui-react-internal-2020-11-02-08-54-48-taurheim-focustrap-onfocus.json: "patch" is not a valid change type"

Based on the documentation [here](https://github.com/microsoft/fluentui/wiki/Development-Workflow), patch should be a valid change type, right? Can you help me understand what I need to do to run a passing build?

layershifter:
@taurheim I think that now it's `"type": "prerelease",`. Can you please remove a change file and try to run `yarn change` again?

taurheim:
If someone more familiar with FocusTrapZone could take another look at this change that would be great -- to make sure I've covered everything I had to make another change, moving the code that was previously run onBlur to run onBlurCapture. This means that it will be run *before* blur is called on the target element. Similarly to the other change (setting internal state on bumper focus), it is necessary for cases where onBlur and onFocus do not bubble (see #15530 for more explanation). I don't see any issues with this change but there may be repercussions I haven't considered and want to make sure I'm not breaking a key scenario here :)

layershifter:
@dzearing can you please check this one?

khmakoto (approved):

packages/react-internal/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -156,6 +156,9 @@ export const FocusTrapZone: React.FunctionComponent<IFocusTrapZoneProps> & {
             focus();
           } else {
             nextFocusable.focus();
+
+            // Compatibility for preact where onFocus does not bubble
+            internalState.hasFocus = true;
```
xugao, line 161: Why is this set here instead of `onRootFocusCapture`?

xugao: also, for consistency, I think `internalState.hasFocus` should be set/unset in `onFocusCapture`/`onBlurCapture`

packages/react-internal/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -171,9 +174,8 @@ export const FocusTrapZone: React.FunctionComponent<IFocusTrapZoneProps> & {
     [onFocus, internalState],
   );
 
-  const onRootBlur = React.useCallback(
+  const onRootBlurCapture = React.useCallback(
     (ev: React.FocusEvent<HTMLDivElement>) => {
-      onBlur?.(ev);
```
xugao, line 176: we should probably call `onBlurCapture` (passed from user `props`) here

packages/react-internal/src/components/FocusTrapZone/FocusTrapZone.tsx
taurheim: I tried setting it in onRootFocusCapture, but my understanding and observed behaviour (based on [this](https://webcache.googleusercontent.com/search?q=cache:UO3E07JSSIYJ:https://harrymoreno.com/2015/12/27/event-delegation-in-javascript.html+&cd=3&hl=en&ct=clnk&gl=us)) is that the "Capture" phase goes from root to target, meaning that `onRootFocusCapture` gets called before `onBumperFocus`, which causes issues as `internalState.hasFocus` needs to be set *after* `onBumperFocus` has executed his logic. I agree that the logic as it stands is confusing, so I've made a change to remove the handler on the bumper and instead check for bumper focus directly in onRootFocusCapture. Let me know what you think!

packages/react-internal/src/components/FocusTrapZone/FocusTrapZone.tsx
```diff
@@ -166,14 +167,13 @@ export const FocusTrapZone: React.FunctionComponent<IFocusTrapZoneProps> & {
   const onRootFocus = React.useCallback(
```
xugao, line 167: we can get rid of this entirely and just pass `onFocus` below

xugao (commented):
The updates LGTM. The build failed due to some snapshots not update-to-date. I will approve once resolved!

taurheim:
I've fixed the tests - apologies for all the commits, I was having trouble getting a full test run to finish on my local machine :(

xugao (approved):

xugao:
@taurheim - I will go ahead help merge this PR. for the change to be published for v7, could you also make the same fix in `7,0` branch?

PR #15809, opened 2020-11-03
(fix) DetailsList group header aria-posinset and aria-setsize attributes

### LINKED ISSUE #15355: List components: Missing screen reader index count
There are a couple of issues regarding the index count of items in `List` components:
- `aria-posinset` and `aria-setsize` attributes are being flagged as not allowed in group headers given that they have the `row` role to account for correct `grid` semantics. We should remove them and use `aria-rowindex` and `aria-rowcount` instead.
- Index count is missing for items within a group so the total number of them are not read when navigating to the group (see #11817) nor are their positions in the set read when navigating through them (see #11819). This can be fixed by passing the correct `row` role to these items and setting both `aria-rowindex` and `aria-rowcount`.
- We've also been talking about having the items in the list match the visual display rather than including hidden/collapsed rows and having the same index across both headers and items. This is a tentative solution that is pending further discussion before committing to it. This can be evaluated by the following example where `Group Header 2` is collapsed:

```
Group Header 1      -   rowindex=1   rowcount=6
    List item 1.1   -   rowindex=2   rowcount=6
    List item 1.2   -   rowindex=3   rowcount=6
Group Header 2      -   rowindex=4   rowcount=6
Group Header 3      -   rowindex=5   rowcount=6
    List item 3.1   -   rowindex=6   rowcount=6
```

And when we expand it then the attributes change to reflect it:

```
Group Header 1      -   rowindex=1   rowcount=8
    List item 1.1   -   rowindex=2   rowcount=8
    List item 1.2   -   rowindex=3   rowcount=8
Group Header 2      -   rowindex=4   rowcount=8
    List item 2.1   -   rowindex=5   rowcount=8
    List item 2.2   -   rowindex=6   rowcount=8
Group Header 3      -   rowindex=7   rowcount=8
    List item 3.1   -   rowindex=8   rowcount=8
```

smhigley:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15355

#### Description of changes

Updates the Grouped Detailslist to use grid row semantics for index and count on the group header rows. Does not fix the actual index values on nested grid rows.

#### Focus areas to test

- Grouped Detailslist's group headers should have `aria-rowindex` and `aria-rowcount`, both of which should include the header row when calculating index/count
- GroupedList should use `aria-posinset` and `aria-setsize` for group headers

### DIFF
--- a/packages/react/src/components/GroupedList/GroupHeader.base.tsx
+++ b/packages/react/src/components/GroupedList/GroupHeader.base.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { IProcessedStyleSet, ITheme } from '../../Styling';
-import { composeRenderFunction, classNamesFunction, getRTL, getRTLSafeKeyCode, KeyCodes } from '../../Utilities';
+import { composeRenderFunction, classNamesFunction, getId, getRTL, getRTLSafeKeyCode, KeyCodes } from '../../Utilities';
 import { SelectionMode } from '../../Selection';
 import { Check } from '../../Check';
 import { Icon } from '../../Icon';
@@ -27,6 +27,7 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
   };
 
   private _classNames: IProcessedStyleSet<IGroupHeaderStyles>;
+  private _id: string;
 
   public static getDerivedStateFromProps(
     nextProps: IGroupHeaderProps,
@@ -49,6 +50,8 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
 
   constructor(props: IGroupHeaderProps) {
     super(props);
+
+    this._id = getId('GroupHeader');
 
     this.state = {
       isCollapsed: (this.props.group && this.props.group.isCollapsed) as boolean,
@@ -110,8 +113,6 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
       return null;
     }
 
-    const groupLabel = `${group.name} (${group.count}${group.hasMoreData ? '+' : ''})`;
-
     return (
       <div
         className={this._classNames.root}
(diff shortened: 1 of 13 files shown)


smhigley:
Screen reader support testing results for the changes in this PR: https://codepen.io/smhigley/full/WNoGBoZ

packages/react/src/components/GroupedList/GroupHeader.base.tsx
```diff
@@ -107,6 +109,9 @@ export class GroupHeaderBase extends React.Component<IGroupHeaderProps, IGroupHe
     if (!group) {
       return null;
     }
+
+    const groupLabel = `${group.name} (${group.count}${group.hasMoreData ? '+' : ''})`;
```
cschlechty, line 113: not sure how well this will localize

cschlechty (approved):

packages/react/src/components/GroupedList/GroupHeader.base.tsx
smhigley: Good point! I copied it more or less from `_onRenderTitle` (minus the wrapping nodes), so I suppose it'll at least be consistent? 😅. Would it be better to just have the `group.count` after the name, and leave off the "+" sign?

ThomasMichon: You should take in a function as props (passed to the `DetailsList` or whatever by the host) which accepts these fields and produces the final string:
```ts
const groupLabel = props.getGroupAriaLabel?({
  name: group.name,
  count: group.count,
  hasMoreData: group.hasMoreData
 });
If it's not present, I'm sorry but we should just use `ariaLabel` directly.

ThomasMichon (changes requested):

packages/react/src/components/GroupedList/GroupHeader.base.tsx
smhigley: @cschlechty / @ThomasMichon I updated the PR to use `aria-labelledby` pointing at the group title cell. It's still author-overridable through `group.ariaLabel`, and I think it sets a better default by avoiding separate visible vs. screen reader text.

Let me know if that addresses your concerns! :)

ThomasMichon (approved):

PR #15814, opened 2020-11-03
Callout ordering and line opacity

shivasai09:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15815
- [x] Include a change request file using `$ yarn change`

#### Description of changes

fixing the opacity of the line, when we select a line from the chart, the other lines(un-selected) should have the same opacity as the rectangle

also fixed the ordering of the line data in the callout. i have made it to descending oreder
#### Focus areas to test

vertical stacked bar chart

**after fix**
image
image

**before fix**
image
image

PR #15816, opened 2020-11-03
Charting updates and fixes

mbest:
#### Pull request checklist

#### Description of changes

- When `hideLegend` is true, no space is reserved for the legend
- turn on `strictBindCallApply` compiler option and fix resulting errors
- For VerticalStackedBarChart, support turning off hover action on legends
- VerticalStackedBarChart: add minimum bar height option
- VerticalStackedBarChart: add `onBarClick` callback option
- VerticalStackedBarChart: simplify how data is passed to callbacks
- VerticalStackedBarChart: don't output empty stack `<g>`
- VerticalStackedBarChart: disallow `yMinValue` since setting it breaks the scale of the chart


packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
```diff
@@ -213,25 +210,23 @@ export class VerticalStackedBarChartBase extends React.Component<
...
         // we inject index also , it will be helpful to draw lines when x asix is
         // of string type
-        linesData.forEach((line: LineData) => {
-          if (line.x === undefined) {
-            line.x = item.xAxisPoint;
-          }
-          if (line.index === undefined) {
-            line.index = index;
-          }
+        item.lineData.forEach(line => {
```
shivasai09, line 220: could you please revert this change. we need to set `x` value and `index` value only when it is undefined, this new code will override it. which we don't want

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
```diff
@@ -253,8 +248,12 @@ export class VerticalStackedBarChartBase extends React.Component<
     return lineLegends;
   };
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  private _createLines = (xScale: any, yScale: any, containerHeight: number, containerWidth: number): JSX.Element => {
+  private _createLines = (
+    xScale: NumericScale,
```
shivasai09, line 252: it can also be dateScale

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
```diff
@@ -299,27 +298,26 @@ export class VerticalStackedBarChartBase extends React.Component<
...
               isLegendSelected && selectedLegendTitle === item
-                ? this._linehover.bind(this, circlePoint)
-                : this._onStackHover.bind(
-                    this,
-                    circlePoint.x,
-                    find(this.props.data, (point: IVerticalStackedChartProps) => point.xAxisPoint === circlePoint.x)
-                      ?.lineData,
-                  )
+                ? this._lineHover.bind(this, circlePoint)
+                : this._onStackHover.bind(this, circlePoint.x)
```
shivasai09, line 310: `circlePoint.x`?? you need to send the StackData right? this will break the hover on circle i belive

shivasai09: image

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
mbest: For strings, it still uses a numeric scale using the indexes. It cannot be dateScale

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
mbest: Nope. It works fine
image

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
mbest: My new code makes a copy of the objects, so it will need to set these each time this function is run.

mbest: BTW, the previous code was modifying the objects (props) passed into this component, which is generally frowned upon.

Raghurk (approved):

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
Raghurk: @shivasai09 can you make sure this does not break our code?

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
shivasai09: no it isn't
image

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
```diff
@@ -213,25 +210,23 @@ export class VerticalStackedBarChartBase extends React.Component<
...
-            line.x = item.xAxisPoint;
-          }
-          if (line.index === undefined) {
-            line.index = index;
-          }
+        item.lineData.forEach(line => {
+          linesData.push({
+            ...line,
+            index,
+            x: item,
```
shivasai09, line 224: @mbest  why are you setting `item` to `x` instead of `x.xAxisPoint`

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
shivasai09: sure.

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
shivasai09, line 310: @mbest 

it's not working fine
image

mbest: Fixed it: 
image

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
mbest: Using `xAxisPoint` means having to do a `find` later. This change directly links to the item. I've renamed the property to make it a bit more informative.

shivasai09:
@mbest , `_onStackHover` and `_onStackFocus` looks almost similar, can you make it reusable. since this pr is doing clean-up, it would be appropriate to make this change..

what say?

mbest:
> @mbest , `_onStackHover` and `_onStackFocus` looks almost similar, can you make it reusable. since this pr is doing clean-up, it would be appropriate to make this change..

Sure. I also did the same for the "rect" versions.

mbest:
> @mbest , `_onStackHover` and `_onStackFocus` looks almost similar, can you make it reusable. since this pr is doing clean-up, it would be appropriate to make this change..

Sure. I also did the same for the "rect" versions.

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
```diff
@@ -559,82 +565,46 @@ export class VerticalStackedBarChartBase extends React.Component<
...
-      refSelected: mouseEvent,
+      refSelected,
       isCalloutVisible: true,
       YValueHover: isLinesPresent
-        ? [...lineData!, ...found!.chartData.slice().reverse()]
-        : found!.chartData.slice().reverse(),
-      hoverXValue: xAxisPoint,
-      stackCalloutProps: found!,
-      activeXAxisDataPoint: xAxisPoint,
+        ? [...lineData!, ...stack.chartData.slice().reverse()]
```
shivasai09, line 589: hey @mbest , i think while taking master to your branch , you have reverted the behavior which we merged in https://github.com/microsoft/fluentui/pull/15814/files.. we were sorting the `lineData` in descending order

could you please put it back.?

Thanks,
For making the code clean up on this.

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
```diff
@@ -559,82 +565,46 @@ export class VerticalStackedBarChartBase extends React.Component<
...
-      (sinlgePoint: { xAxisPoint: string | number; chartData: IVSChartDataPoint[] }) =>
-        sinlgePoint.xAxisPoint === xAxisPoint,
-    );
-    if (isLinesPresent) {
-      lineData!.forEach((item: ILineDataInVerticalStackedBarChart & { shouldDrawBorderBottom?: boolean }) => {
-        item.data = item.data || item.y;
-        item.shouldDrawBorderBottom = true;
-      });
+  private _onStackFocus(stack: IVerticalStackedChartProps, groupRef: IRefArrayData): void {
+    if (groupRef.refElement) {
```
shivasai09, line 604: is this if condition really necessary ?

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
mbest: Merged now.

shivasai09: hey, can you see line data sorted.?

i am not able to [see](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/pull/15816/merge/charting/dist/index.html#/examples/VerticalStackedBarChart), 
image


it should be in descending. like yellow should be on last and pink should be on first


if it's working for you in local, can you post a pic here. then i guess we are good merge this pr

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
shivasai09: i have noticed one issue. and mentioned in below comments

packages/charting/src/components/VerticalStackedBarChart/VerticalStackedBarChart.base.tsx
mbest: image

shivasai09:
@mbest  once i hover on any legend, and then if i mouse out  on that , it stays in hover state only
image

This happens only in legends. but for line chart and all it works fine.
why are you changing this behavior?

mbest:
> @mbest once i hover on any legend, and then if i mouse out on that , it stays in hover state only
> This happens only in legends. but for line chart and all it works fine.
> why are you changing this behavior?

The actual effect of legend interaction happens due to the callback functions provided. I changed `Legend` to update the UX only when there's also a callback for that interaction, because I needed to add support for not reacting to hover or focus.

To ensure the `Legend` examples work as they did, I've added `onMouseOutAction` callbacks to them.

shivasai09 (approved):

PR #15823, opened 2020-11-03
Remove wrapper around Nav which inhibits proper styling

ThomasMichon:
#### Description of changes

Updated `Nav` to reuse the `FocusZone` element as the 'root' element of the component, allowing `className` to apply properly to the entire box, not just the content of the `FocusZone`. This allows `className` to be used to apply flex-styling and positioning.

#### Focus areas to test

Validate `Nav` examples.

### DIFF
--- a/change/office-ui-fabric-react-2020-11-03-14-38-53-nav-style.json
+++ b/change/office-ui-fabric-react-2020-11-03-14-38-53-nav-style.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Remove wrapper element around Nav inhibiting proper styling",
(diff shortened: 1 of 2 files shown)


ecraig12345:
This seems like a pretty significant (potentially breaking) change to make within v7, though it's definitely the right thing to do and we could change it for v8. @dzearing any thoughts? @ThomasMichon were you able to test this out with any of your code internally, or at least look at current usage to get an idea of whether anything is likely to break?

ecraig12345:
I'm going to close this for housekeeping since as previously stated it's too risky for v7. If you want to resubmit to master for v8 that's fine, but it would need to be in the next couple weeks before we release.

PR #15825, opened 2020-11-03
Pass React event to SelectionZone onItemContextMenu handler

ThomasMichon:
#### Description of changes

Revised `onItemContextMenu` in `SelectionZone` so that the React synthetic event is passed to the callback instead of the `nativeEvent`. The issue is that `nativeEvent` is a DOM event *which has already propagated to `window`*, thus making it impossible to stop propagation to `oncontextmenu` handlers higher in the React tree.

Since most handlers are likely to use only a few fields of the `event` parameter, it should be safe to simply force-cast the React event to `Event`, since there is significant overlap.

#### Focus areas to test

Validate examples which use the context menu callback of `SelectionZone`.

### DIFF
--- a/packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
+++ b/packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
@@ -408,7 +408,7 @@ export class SelectionZone extends React.Component<ISelectionZoneProps, ISelecti
 
         this._onInvokeMouseDown(ev, index);
 
-        const skipPreventDefault = onItemContextMenu(selection.getItems()[index], index, ev.nativeEvent);
+        const skipPreventDefault = onItemContextMenu(selection.getItems()[index], index, (ev as unknown) as Event);
 
         // In order to keep back compat, if the value here is undefined, then we should still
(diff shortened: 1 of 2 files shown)


packages/office-ui-fabric-react/src/utilities/selection/SelectionZone.tsx
```diff
@@ -408,7 +408,7 @@ export class SelectionZone extends React.Component<ISelectionZoneProps, ISelecti
 
         this._onInvokeMouseDown(ev, index);
 
-        const skipPreventDefault = onItemContextMenu(selection.getItems()[index], index, ev.nativeEvent);
+        const skipPreventDefault = onItemContextMenu(selection.getItems()[index], index, (ev as unknown) as Event);
```
dzearing, line 411: this would need a big comment explaining whats up here.

dzearing: This really should have been the React event in the first place. Not sure why the native even was passed.

layershifter:
@ThomasMichon can you please address review comments?

PR #15826, opened 2020-11-03
ContextualMenu: Using MenuContext.Consumer to consume context from MenuButton

khmakoto:
#### Pull request checklist

#### Description of changes

This PR follows up #15780 by making `ContextualMenu` consume the `MenuContext` used in `MenuButton` so that you can pass `ContextualMenu` directly as JSX:

```tsx
<MenuButton menu={<ContextualMenu {...menuProps} />} />
```

The `MenuContext` is now moved into a newly created `@fluentui/react-shared-contexts` package and out of the `@fluentui/react-button` package.

#### Focus areas to test

(optional)


dzearing (approved):

packages/react-internal/package.json
```diff
@@ -55,6 +55,7 @@
     "sinon": "^4.1.3"
   },
   "dependencies": {
+    "@fluentui/react-button": "^1.0.0-beta.6",
```
ecraig12345, line 58: If you're doing this it might be good to remove the direct dep from the main react package and add the re-export files here.

khmakoto: I'm having a chat with Jon and David on the side that might lead us to not need this here, but if we do I'll update as per your request.

joschect (approved):

packages/react-shared-contexts/README.md
```diff
@@ -0,0 +1,8 @@
+# @fluentui/react-shared-contexts
+
+**Shared contexts for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
+([formerly Office UI Fabric React](https://developer.microsoft.com/en-us/office/blogs/ui-fabric-is-evolving-into-fluent-ui/))
```
ecraig12345, line 4: Let's not include this line for new packages (and we should probably remove it from the template)
```suggestion
```

packages/react-shared-contexts/README.md
```diff
@@ -0,0 +1,8 @@
+# @fluentui/react-shared-contexts
+
+**Shared contexts for [Fluent UI React](https://developer.microsoft.com/en-us/fluentui)**
```
ecraig12345, line 3: Are these for northstar too? If so we should probably update this to something like:
```
**Shared contexts for [`@fluentui/react`](v8 doc link) and [`@fluentui/react-northstar`](v0 doc link)**
```

packages/react-shared-contexts/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "react-app-polyfill": "~1.0.1",
+    "react-dom": "16.8.6",
+    "react-test-renderer": "^16.3.0",
+    "sinon": "^4.1.3"
+  },
+  "dependencies": {
+    "@fluentui/set-version": "^8.0.0-beta.0",
+    "prop-types": "^15.7.2",
+    "tslib": "^1.10.0"
+  },
```
ecraig12345, line 53: Did you generate this with `yarn create-package` or by copying something? Several of these deps shouldn't be here. (I'm fine with either removing the test-related deps, or keeping them if you think there might be tests later.)
```suggestion
  "devDependencies": {
    "@fluentui/eslint-plugin": "^1.0.0-beta.0",
    "@types/enzyme": "3.10.3",
    "@types/enzyme-adapter-react-16": "1.0.3",
    "@types/jest": "~24.9.0",
    "@types/react": "16.8.25",
    "@types/react-test-renderer": "^16.0.0",
    "@fluentui/scripts": "^1.0.0",
    "@fluentui/test-utilities": "^8.0.0-beta.1",
    "enzyme": "~3.10.0",
    "enzyme-adapter-react-16": "^1.15.0",
    "react": "16.8.6",
    "react-test-renderer": "^16.3.0"
  },
  "dependencies": {
    "@fluentui/set-version": "^8.0.0-beta.0",
    "tslib": "^1.10.0"
  },
```

ecraig12345: Alternate version if you want to remove the test stuff (in that case you should also remove the jest config and test-related scripts)
```suggestion
  "devDependencies": {
    "@fluentui/eslint-plugin": "^1.0.0-beta.0",
    "@types/react": "16.8.25",
    "@fluentui/scripts": "^1.0.0",
    "react": "16.8.6"
  },
  "dependencies": {
    "@fluentui/set-version": "^8.0.0-beta.0",
    "tslib": "^1.10.0"
  },
```

packages/react-shared-contexts/tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "importHelpers": true,
+    "noImplicitAny": true,
+    "noUnusedLocals": true,
+    "strictNullChecks": true,
+    "forceConsistentCasingInFileNames": true,
+    "moduleResolution": "node",
+    "preserveConstEnums": true,
+    "skipLibCheck": true,
+    "typeRoots": ["../../node_modules/@types", "../../typings"],
+    "types": ["jest", "custom-global", "node"]
```
ecraig12345, line 20: You can also remove the jest reference here if you decide to remove test stuff
```suggestion
    "typeRoots": ["../../node_modules/@types", "../../typings"],
    "types": ["jest", "custom-global"]
```

packages/react-shared-contexts/jest.config.js
```diff
@@ -0,0 +1,6 @@
+let path = require('path');
+let { createConfig, resolveMergeStylesSerializer } = require('@fluentui/scripts/jest/jest-resources');
+module.exports = createConfig({
+  setupFiles: [path.resolve(path.join(__dirname, 'config', 'tests.js'))],
+  snapshotSerializers: [resolveMergeStylesSerializer()],
```
ecraig12345, line 5:
```suggestion
let { createConfig } = require('@fluentui/scripts/jest/jest-resources');
module.exports = createConfig({
  setupFiles: [path.resolve(path.join(__dirname, 'config', 'tests.js'))],
```

ecraig12345 (changes requested):
A few changes needed to the new package

packages/react-shared-contexts/README.md
khmakoto: It's not going to be used for northstar right now. I'd prefer to not add that until needed.

packages/react-shared-contexts/package.json
khmakoto: There might be tests later, so I'd prefer to keep them there. This is by copying since the `create-package` script failed several times for me. I plan to look into it later.

packages/react-shared-contexts/README.md
khmakoto: Removed.

packages/react-shared-contexts/package.json
khmakoto: Done!

packages/react-shared-contexts/jest.config.js
khmakoto: Done!

packages/react-shared-contexts/tsconfig.json
khmakoto: Done!

packages/react-shared-contexts/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "outDir": "lib",
+    "module": "commonjs",
+    "lib": ["es5", "es2015.promise", "dom"],
```
ecraig12345, line 6: Forgot one thing, we should generally only include promise types if it's used
```suggestion
    "lib": ["es5", "dom"],
```

khmakoto: Done!

ecraig12345 (approved):

PR #15831, opened 2020-11-04
refactor(Tree): BREAKING CHANGE - rework tree as hook

YuanboXue-Amber:
#### Pull request checklist

## Description of the re-work ideas:

The key is to create a **flattened tree** (see `flattenTree.ts`).
The flattened tree is an object, with key being the id of each tree item, and value being the information of the corresponding tree item: its parent/child, was it selected, ..., etc.

Initially the flattened tree is created thru Tree's `props.items`.
Then it gets re-created every time when state changes. Meaning every time users select/unselect, expand/collapse a tree node, there will be O(n) time spent on re-creating flattened tree, where n being number of tree items.

Why flatten tree:
- provides a convenient way of navigating up/down tree. 
- being the single source of truth for the state (expanded/collapsed, selected) of the tree
- together with `getItemById`, provides a context for tree items to access tree information.

The flattened tree is then used in `useTree` hook. This hook is a building block for `Tree` component, and a `useVirtualTree` hook (not in this PR yet) for virtualized tree component.
 
## Breaking changes:

- The following props are removed from `TreeItem`:

  ```ts
    /** A state of selection indicator. */
    selected?: boolean;

    /** Whether or not tree item is part of the selectable parent. */
    selectableParent?: boolean;

    indeterminate?: boolean;
  ```

  Reason for removal:

  1. These props are not in 0.47 version.
  2. `selected` and `indeterminate` can be accessed from the flattened tree. (In fact, if possible, I'd like to remove all props that can be accessed from the flattened tree, for example `parent`, `index`, ..., .etc). `selectableParent` is useless in my opinion.

- Tree does not create and pass `contentRef` prop to each TreeItem anymore.
Originally Tree creates and keeps a map of refs to each tree items, and assigned it to tree item's `contentRef` prop. The new re-worked tree does not do this anymore. 
But `contentRef` is still kept as a prop for TreeItem.

#### Focus areas to test

(optional)


YuanboXue-Amber:
@layershifter @assuncaocharles

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -228,235 +160,41 @@ export const Tree: ComponentWithAs<'div', TreeProps> &
   });
 
   const treeRef = React.useRef<HTMLElement>();
```
layershifter, line 162: I don't see any usage of `treeRef`, can we remove it?

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -228,235 +160,41 @@ export const Tree: ComponentWithAs<'div', TreeProps> &
...
-          ...renderedItems,
-          renderedItem,
-          ...(isSubtreeExpanded ? renderItems(item.items as TreeItemProps[], level + 1, id) : ([] as any)),
-        ];
-      }, []);
-    };
-    const itemsRendered = renderItems(items as TreeItemProps[]);
-
-    return itemsRendered;
+    if (!tobeRenderedItemsProps) return [];
```
layershifter, line 189:
```suggestion
```

Can we always return an array?

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -26,22 +24,33 @@ import {
...
+  toggleActive: (ids: string[], e: React.SyntheticEvent) => void;
+  toggleSelect: (ids: string[], e: React.SyntheticEvent) => void;
+  focusFirstChild: (itemId: string) => void;
+  focusParent: (itemId: string) => void;
+  siblingsExpand: (e: React.SyntheticEvent, itemProps: TreeItemProps) => void;
+  registerItemRef: (id: string, node: HTMLElement) => void;
+  getItemById: (id: string) => BaseFlatTreeItem & { [key: string]: any };
+}
+
+export const TreeContext = React.createContext<TreeRenderContextValue>({
```
layershifter, line 45: Let's move out `TreeContext` to a separate file to avoid circular references

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -26,22 +24,33 @@ import {
...
+import { BaseFlatTreeItem } from './hooks/flattenTree';
+
+export interface TreeRenderContextValue {
+  toggleActive: (ids: string[], e: React.SyntheticEvent) => void;
+  toggleSelect: (ids: string[], e: React.SyntheticEvent) => void;
+  focusFirstChild: (itemId: string) => void;
+  focusParent: (itemId: string) => void;
+  siblingsExpand: (e: React.SyntheticEvent, itemProps: TreeItemProps) => void;
+  registerItemRef: (id: string, node: HTMLElement) => void;
+  getItemById: (id: string) => BaseFlatTreeItem & { [key: string]: any };
```
layershifter, line 42:
```suggestion
  getItemById: (id: string) => BaseFlatTreeItem;
```

I would like to avoid this part if possible

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -26,22 +24,33 @@ import {
...
+}
+
+export const TreeContext = React.createContext<TreeRenderContextValue>({
+  toggleActive: _.noop,
+  toggleSelect: _.noop,
+  focusFirstChild: _.noop,
+  focusParent: _.noop,
+  siblingsExpand: _.noop,
+  registerItemRef: _.noop,
+  getItemById: id => ({ id } as any),
```
layershifter, line 52:
```suggestion
  getItemById: id => ({ id } as Partial<BaseFlatTreeItem>),
```

Something like should be able to replace any...

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react';
+import { TreeProps } from '../Tree';
+import { flattenTree, getTobeRenderedItemsProps } from './flattenTree';
+import { useTreeActiveState } from './useTreeActiveState';
+import { useTreeBehavior } from './useTreeBehavior';
+
+export function useTree(props: TreeProps) {
```
layershifter, line 7: I don't think that it's a right idea to depend on `TreeProps`, I suggest to create `UseTreeOptions` type and `TreeProps` can extend it.

Motivation: hook can be used without `Tree` component, `TreeProps` contains non related properties

```suggestion
export function useTree(options: UseTreeOptions) {
```

packages/fluentui/react-northstar/src/components/Tree/hooks/useSelectableTree.tsx
```diff
@@ -0,0 +1,40 @@
...
+export function useSelectableTree(props: TreeProps) {
+  const baseTree = useTree(props);
+  const { flatTree: baseflatTree, tobeRenderedItemsProps } = baseTree;
+
+  const { selectedItemIds, flatTree, toggleSelect } = useTreeSelectState(props, baseflatTree);
+
+  React.useMemo(() => {
+    tobeRenderedItemsProps.forEach(itemProps => {
+      // tree item is selectable when it is in a selectable tree, and does not receive selectable=false prop
+      itemProps.selectable = props.selectable ? itemProps.selectable !== false : false;
```
ling1726, line 25: By transitivity

(false !== false) === false
(true !== false) === true

```suggestion
      itemProps.selectable = props.selectable ? itemProps.selectable : false;
```

ling1726: Should also work

```suggestion
      itemProps.selectable = props.selectable && itemProps.selectable
```

packages/fluentui/react-northstar/src/components/Tree/hooks/useStableProps.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+
+export function useStableProps<P>(props: P) {
+  const stableProps = React.useRef<P>(props);
+
+  React.useEffect(() => {
+    stableProps.current = props;
+  });
```
ling1726, line 8: This effect will be run on every single rerender, is that intended ?

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
...
+    initialValue: initialActiveItemIds,
+  });
+
+  // compute a new flattened tree based on activeItemIds state
+  const updatedFlatTree = React.useMemo(() => {
+    const updatedFlatTree = { ...flatTree };
+    Object.keys(updatedFlatTree).forEach(key => {
+      if (updatedFlatTree[key].expanded) {
+        updatedFlatTree[key].expanded = false;
+      }
```
ling1726, line 42: The check is unnecessary

```suggestion
        updatedFlatTree[key].expanded = false;
```

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
```diff
@@ -0,0 +1,144 @@
...
+  /** true if the tree item has subtree, indicating childrenIds are not undefined */
+  hasSubtree: boolean;
+
+  /** children ids of the tree item. For leaf tree item, childrenIds is undefined */
+  childrenIds?: string[];
+};
+
+export type FlatTree = {
+  [key: string]: BaseFlatTreeItem & { [key: string]: any };
+};
```
layershifter, line 45: nit

```suggestion
export type BaseFlatTree = Record<string, BaseFlatTreeItem>;
```

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
...
+    });
```
ling1726, line 43: You don't want to iterate the object twice
```suggestion
    const updatedFlatTree = {};
    Object.keys(flatTree).forEach(key => {
     updatedFlatTree[key] = {...flatTree[key], expanded: false};
    });
```

ling1726: `reduce` instead of `forEach` will mean you don't have `updatedFlatTree` existing as an empty object, but that's more of a nit

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      if (items?.length) {
+        flatten(items, level + 1, id);
+      }
+    });
+  };
+
+  flatten(items, 1);
+  return result;
+}
```
layershifter, line 81: I reworked this function to be more readable and avoid using `.length` as it will always count a length of an array

```suggestion
export function flattenTree(
  items: ObjectShorthandCollection<TreeItemProps>[],
  level: number = 1,
  parent: string | null = null,
  result: BaseFlatTree = {},
): BaseFlatTree {
  const itemsInLeaf = items.length;

  items.forEach((item, indexAmongSiblings) => {
    const { id, expanded, items: childrenItems } = item;
    const hasSubtree = childrenItems ? !!childrenItems.length : false;

    result[id] = {
      index: indexAmongSiblings + 1, // Used for aria-posinset and it's 1-based.
      level,
      expanded: hasSubtree ? !!expanded : undefined,
      parent: parent == null ? undefined : parent,
      treeSize: itemsInLeaf,
      hasSubtree,
      childrenIds: [],
    };

    if (hasSubtree) {
      flattenTree(childrenItems, level + 1, id, result);
    }

    if (parent) {
      result[parent].childrenIds.push(id);
    }
  });

  return result;
}
```

A test on CSB to ensure that implementation matches: https://codesandbox.io/s/throbbing-dew-ewbcp?file=/src/flatten.spec.js:852-1603

The single difference is that it always assigns `childrenItems`. Perf is the same:

image

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.tsx
```diff
@@ -0,0 +1,49 @@
...
+  // build flattened tree from props.items
```
layershifter, line 8: Let's avoid comments about _what it does_ and replace them with _why_ if it's needed

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
...
+  activeItemIds: string[];
+  flatTree: FlatTree;
+  toggleActive: (ids: string[], e: React.SyntheticEvent) => void;
+}
+
+export function useTreeActiveState(
+  props: Pick<TreeProps, 'defaultActiveItemIds' | 'activeItemIds' | 'exclusive'>,
+  flatTree: FlatTree,
+): TreeActiveState {
+  const initialActiveItemIds = React.useMemo(() => {
```
layershifter, line 19:
```suggestion
  // We need this because we want to handle `expanded` prop on `items`, should be deprecated and removed
  const initialActiveItemIds = React.useMemo(() => {
```

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
...
+    Object.keys(updatedFlatTree).forEach(key => {
+      if (updatedFlatTree[key].expanded) {
+        updatedFlatTree[key].expanded = false;
+      }
+    });
+    activeItemIds.forEach(activeId => {
+      updatedFlatTree[activeId].expanded = true;
+    });
+    return updatedFlatTree;
+  }, [activeItemIds, flatTree]);
```
layershifter, line 48:
```suggestion
  // We want to set `expanded` value on all items to simplify rendering later
  // There is no sense to recreate a flat tree so we simply mutating it
  const flatTreeItems = React.useMemo(() => {
    const updatedFlatTree = { ...baseFlatTreeItems };

    Object.keys(updatedFlatTree).forEach(key => {
      if (updatedFlatTree[key].expanded) {
        updatedFlatTree[key].expanded = false;
      }
    });

    activeItemIds.forEach(activeId => {
      updatedFlatTree[activeId].expanded = true;
    });

    return updatedFlatTree;
  }, [activeItemIds, baseFlatTreeItems]);
```

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
...
+import * as _ from 'lodash';
+
+export interface TreeActiveState {
+  activeItemIds: string[];
+  flatTree: FlatTree;
+  toggleActive: (ids: string[], e: React.SyntheticEvent) => void;
+}
+
+export function useTreeActiveState(
+  props: Pick<TreeProps, 'defaultActiveItemIds' | 'activeItemIds' | 'exclusive'>,
```
layershifter, line 16: Let's use `UseTreeOptions` instead

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
+import * as React from 'react';
+import { useAutoControlled } from '@fluentui/react-bindings';
+import { TreeProps } from '../Tree';
+import { useStableProps } from './useStableProps';
+import { FlatTree } from './flattenTree';
+import { findIndex, removeItemAtIndex } from './utils';
+import * as _ from 'lodash';
+
+export interface TreeActiveState {
```
layershifter, line 9:
```suggestion
export interface UseTreeActiveStateResult {
```

nit

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+      return result;
+    },
+    [props.exclusive, updatedFlatTree],
+  );
+
+  const stableProps = useStableProps(props);
+  const toggleActive = React.useCallback(
+    (ids: string[], e: React.SyntheticEvent) => {
+      const nextActiveItemIds = ids.reduce((prev, currId) => toggleActiveOnOneId(prev, currId), activeItemIds);
```
layershifter, line 78: I don't understand this part and `toggleActiveOnOneId` function at all. What it does?

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
...
+      }
+    });
+    activeItemIds.forEach(activeId => {
+      updatedFlatTree[activeId].expanded = true;
+    });
+    return updatedFlatTree;
+  }, [activeItemIds, flatTree]);
+
+  const toggleActiveOnOneId = React.useCallback(
+    (ids: string[], id: string): string[] => {
```
layershifter, line 51:
```suggestion
    (activeIds: string[], idToToggle: string): string[] => {
```

Can we do better naming naming there?

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
...
+          // need to collapse everything else, except id and its ancestors
+          const ancestors = getAncestorsIds(updatedFlatTree, id);
+          return [...ancestors, id];
+        }
+        return [...ids, id];
+      }
+
+      return result;
+    },
+    [props.exclusive, updatedFlatTree],
```
layershifter, line 72: **(!)** Having a dependency there on `updatedFlatTree` is a bad idea as `toggleActive` is a part of `contextValue`

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeBehavior.tsx
```diff
@@ -0,0 +1,75 @@
...
+      }
+
+      if (flatTree[firstChild]?.hasSubtree) {
+        firstChildRef.focus();
+      } else {
+        // when node is leaf, need to focus on the inner treeTitle
+        const firstElementChild = firstChildRef.firstElementChild;
+        if (firstElementChild) {
+          (firstElementChild as HTMLElement).focus();
+        }
```
ling1726, line 36:
```suggestion
        (firstElementChild as HTMLElement)?.focus();
```

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,105 @@
...
+    (ids: string[], id: string): string[] => {
+      if (!updatedFlatTree[id]?.hasSubtree) {
+        // leaf node does not have the concept of active/inactive
+        return ids;
+      }
+
+      let result: string[];
+      const index = findIndex(ids, id);
+      if (index >= 0) {
+        result = removeItemAtIndex(ids, index);
```
layershifter, line 60: Why we need to use `findIndex` there?

---

- to understand if `id` exists in an array we can use `.indexOf` or `.includes`
- to remove an item from an array we can use [`_.without()`](https://lodash.com/docs/4.17.15#without)
- we don't care about ordering in `ids`

Let's remove `findIndex` and `removeItemAtIndex` if possible

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export function getTobeRenderedItemsProps(
```
layershifter, line 83: _nit_ Is this function is a part of `flattenTree`?

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
```diff
@@ -0,0 +1,144 @@
...
+  items: ObjectShorthandCollection<TreeItemProps>, // tree's props.items
+  flatTree: FlatTree,
+  treeRenderItemTitle?: ShorthandRenderFunction<TreeTitleProps>,
+): TreeItemProps[] {
+  if (!items) {
+    return [];
+  }
+
+  return items.reduce((prevProps: TreeItemProps[], currItem) => {
```
layershifter, line 92: Can you please explain why we are iterating there through `items`? `items` is non-flat array, I don't see any guarantee that we will iterate through all items

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
```diff
@@ -0,0 +1,144 @@
...
+      renderItemTitle: renderItemTitle ?? treeRenderItemTitle,
+      selectable,
+      selectionIndicator,
+      // props from flatTree:
+      expanded,
+      parent,
+      level,
+      index,
+      treeSize,
+    };
```
layershifter, line 136: I probably lost the idea there. I see two issues:
- tree is no way to pass additional props in this case
- I don't understand the value of flattening in this case as there we are doing it second time (?)

---

I suggest to modify `flattenTree` to include all necessary props instead

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeSelectState.tsx
```diff
@@ -0,0 +1,136 @@
...
+    });
+    selectedItemIds.forEach(id => {
+      if (!updatedFlatTree[id].hasSubtree) {
+        updatedFlatTree[id].selected = true;
+      }
+    });
+
+    // traverse all tree nodes in updatedFlatTree for once,
+    // to calculate the selection state of the parent nodes based on leaf nodes
+    const traverseTree = nodes => {
```
ling1726, line 50: consider better naming, a function called `traverseTree` that returns a boolean or 'indeterminate' doesn't really tell me what it does

ling1726: do you even use the return values ?

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.tsx
```diff
@@ -0,0 +1,49 @@
...
+    () => getTobeRenderedItemsProps(props.items, flatTree, props.renderItemTitle),
+    [props.items, props.renderItemTitle, flatTree],
+  );
+
+  // === manage ref to all to-be-rendered tree items  ===
+  const nodes = React.useRef({});
+  const registerItemRef = React.useCallback((id, node) => {
+    nodes.current[id] = node;
+  }, []);
+  const getItemRef = React.useCallback(id => nodes.current[id], []);
```
layershifter, line 25: Will be great to have typings there

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  // === context provided by tree, consumed by tree items ===
+  const { focusParent, focusFirstChild, siblingsExpand } = useTreeBehavior(
```
layershifter, line 28: Is there any motivation to keep this separately from `useTreeState()`?

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.tsx
```diff
@@ -0,0 +1,49 @@
...
+    flatTree,
+    toggleActive,
+    props.exclusive,
+    getItemRef,
+  );
+
+  const getItemById = React.useCallback(id => flatTree[id], [flatTree]);
```
layershifter, line 35: **(!)** This function is a part of React Context value, but again may be unstable because dependency on `flatTree`

packages/fluentui/react-northstar/src/components/Tree/hooks/useSelectableTree.tsx
```diff
@@ -0,0 +1,40 @@
...
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [tobeRenderedItemsProps, flatTree]);
+
+  const getItemById = React.useCallback(id => flatTree[id], [flatTree]);
```
layershifter, line 30:
```suggestion
```

This comes from `baseTree`

packages/fluentui/react-northstar/src/components/Tree/hooks/utils.ts
```diff
@@ -0,0 +1,9 @@
+import * as _ from 'lodash';
+
+export const removeItemAtIndex = (items: string[], itemIndex: number): string[] => {
+  return _.concat(_.slice(items, 0, itemIndex), _.slice(items, itemIndex + 1));
+};
```
ling1726, line 5: This function is really problematic, you tend to use it in iterations, and every time you call it you will trigger more iterations.... better to follow a pattern of identifying items to exclude and recreating the original array without them... you sacrifice more space for time

packages/fluentui/react-northstar/src/components/Tree/hooks/useStableProps.tsx
layershifter: Yes. I would like to replace this with another hook on next iteration

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeSelectState.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+      setSelectedItemIdsState(nextSelectedItemIds);
+    },
+    [selectedItemIds, setSelectedItemIdsState, stableProps, toggleSelectOnOneId],
+  );
+
+  return { selectedItemIds, flatTree: updatedFlatTree, toggleSelect };
+}
+
+function getLeavesOfSubTree(flatTree, subTreeRootId) {
```
ling1726, line 123: nit: the fact that it is a subtree is kind of irrelevant
```suggestion
function getLeafNodes(flatTree, rootId) {
```

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
YuanboXue-Amber: I think we can.
Originally `treeRef` was only used for finding first child, and it is used to traverse DOM. 
We now uses flat tree to get the id of the first child instead of traversing DOM, so I will remove it.

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
YuanboXue-Amber: Oh in fact we are, let me remove this useless check

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
YuanboXue-Amber: Right. It was in a separate file and I put it to tree. Didn't think about circular reference. Let me put it out as 'context.ts'

packages/fluentui/react-northstar/src/components/Tree/hooks/useSelectableTree.tsx
YuanboXue-Amber: I don't think this produce the same result.

When `props.selectable=true` (this specifies if tree is selectable or not), by default `itemProps.selectable` (this specifies if tree item is selectable or not) should be true, unless the input explicitly specifies `itemProps.selectable=false`, so it goes:

```
props.selectable | itemProps.selectable | result
------------------------------------------------
true             | true                 | true
                 | undefined            | true
                 | false                | false 
```

ling1726: oh I see so `itemProps.selectable` can be undefined, then it does make sense. My bad

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
YuanboXue-Amber: I used `findIndex` and `removeItemAtIndex` because I wasn't sure what to use, array function or lodash. So I put them in one place for easy replace.

I could change it all to `.indexOf` and `_.without()`, sounds good?
btw I found out `_.without()` is faster than slice and concat :D 
Screenshot 2020-11-05 at 14 04 52

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
YuanboXue-Amber: I have to use items here, because flatTree does not keep order.

If I have tree item with id that can be parsed to int, for example '1' '2' ... They will always be in front of all other items in flatTree.

https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
YuanboXue-Amber: I have to use items here, because flatTree does not keep order.

If I have tree item with id that can be parsed to int, for example '1' '2' ... They will always be in front of all other items in flatTree.

https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
YuanboXue-Amber: I could move it to `useTree.tsx`?

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
layershifter: > I could change it all to .indexOf and _.without(), sounds good?

Yes, please

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
YuanboXue-Amber: > tree is no way to pass additional props in this case

This can be done by iterate thru the returned `tobeRenderedItemsProps` again

> I don't understand the value of flattening in this case as there we are doing it second time (?)

Same reason as my previous comment. Flat tree does not keep order.

Another way could be, I remember all the root nodes of the tree, and generate `tobeRenderedItemsProps` based on their expanded props and childrenIds, and the ids of their children, and so on...

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.tsx
YuanboXue-Amber: Yes. Because this is focusing on keyboard navigation. It is not related to either active state or select state. So I set it aside.

packages/fluentui/react-northstar/src/components/Tree/hooks/utils.ts
YuanboXue-Amber: Shift raise the similar concern. Going to change this to _.without :)

packages/fluentui/react-northstar/src/components/Tree/hooks/useSelectableTree.tsx
YuanboXue-Amber: Hmm... not exactly.
Because `useTreeSelectState` returns a new flatTree, and when we want to `getItemById` in a tree item, we would like the new flatTree, not the one from baseTree

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.tsx
YuanboXue-Amber:
```suggestion
  const nodes = React.useRef({});
  const registerItemRef = React.useCallback((id: string, node: HTMLElement) => {
    nodes.current[id] = node;
  }, []);
  const getItemRef = React.useCallback((id): HTMLElement => nodes.current[id], []);
```

like so? (not sure if I get the idea properly.)

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
YuanboXue-Amber: I see, your way has less loop. I did a test and it is twice faster for 100 items. Going to change to use the same idea :) 
```suggestion
  const updatedFlatTree = {};
  Object.keys(flatTree).forEach((key) => {
    updatedFlatTree[key] = { ... flatTree[key] };
    if (updatedFlatTree[key].expanded) {
      updatedFlatTree[key].expanded = false;
    }
  });
```
(I didn't commit your suggestions directly because I wanted to add a check of `if (updatedFlatTree[key].expanded)`, so I keep the expanded=undefined to continue be undefined)

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
YuanboXue-Amber: Shift I saw you added comment.

But the code here, it is not mutating, but recreating.

It is recreating here, because we need to know when to re-calculate the `tobeRenderedItemsProps`.
```ts
// this is from useTree.tsx

 const tobeRenderedItemsProps = React.useMemo(
    () => getTobeRenderedItemsProps(props.items, flatTree, props.renderItemTitle),
    [props.items, props.renderItemTitle, flatTree], // this will never be recalculated in the case of mutating
  )
```

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeSelectState.tsx
YuanboXue-Amber: Yes it is used in line 56.

This function may get changed later. Shift hasn't reviewed this part yet, so I'm not sure if my implementation here is good enough.

This `traverseTree` is used for calculating the select state of a tree. 
The parent tree node can have selected/not-selected/indeterminate state, based on if its children is fully-selected/not-selected/half-selected.
So `traverseTree`, is doing a depth first traverse down the tree, til it reaches leaf nodes, then gather the selection state of all leaf children, and then pass this information back to parent. And parent to its parent, and so on.

It's really hard for me to came up a name to make sense here.... 🙈

ling1726: alright, that's fine, let's resolve this comment for now if you'll revisit this

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
YuanboXue-Amber: I added a check to make sure `childrenIds` is undefined instead of `[]` for leaf nodes. I discover I was checking if an item is leaf by if childrenIds is null. So I'd like to keep this.

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
YuanboXue-Amber: `toggleActiveOnOneId` deal with the case when one item's active state was changed. It accepts an array of active Ids and the one Id to be toggle, and returns a new array of active Ids.

`toggleActive` accepts an array of Ids, and it calls `toggleActiveOnOne` on these Ids one by one, to get the final active Ids.

I understand I didn't give proper name here. Hope the new naming clears it up.

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
YuanboXue-Amber: I understand. However I'm not sure what would be a good way to avoid so, as we are changing the flatTree based on the active ids change. Let's talk about this tomorrow

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.tsx
layershifter: Yes 👍

layershifter: Please also type `useRef`, `useRef<Record<string, HTMLElement>>(

packages/fluentui/react-northstar/src/components/Tree/hooks/useStableProps.tsx
ling1726: Per or offline discussion, I see why this is necessary now

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
YuanboXue-Amber: Talked about it on the call, we are not deep cloning items, make a comment about that.

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
YuanboXue-Amber: So in `useTreeSelectState.tsx`:

```ts 
export interface SelectableFlatTreeItem extends BaseFlatTreeItem {
  selected?: boolean | 'indeterminate';
} 
```

And when I need to use `flatTree[id].selected`, I do `(flatTree[id] as SelectableFlatTreeItem).selected = true`

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
YuanboXue-Amber: Resolved on the call, now returning `toggleItemActive` and `expandSiblings`

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
YuanboXue-Amber: Resolved  by `useGetItemById`

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.tsx
YuanboXue-Amber: Resolved  by `useGetItemById`

packages/fluentui/react-northstar/src/components/Tree/hooks/useSelectableTree.tsx
YuanboXue-Amber: Resolved by returning an array of ids in order

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
YuanboXue-Amber: now it's in Tree.tsx

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
YuanboXue-Amber: Resolved by returning an array of ordered ids

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
```diff
@@ -0,0 +1,114 @@
+import { ObjectShorthandValue } from '../../../index';
```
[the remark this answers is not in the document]
YuanboXue-Amber: Question: why is importing from `types` better than `index`?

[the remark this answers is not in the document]
YuanboXue-Amber: Oh sorry I just saw the other comments pointing out the circular reference

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeActiveState.tsx
```diff
@@ -0,0 +1,100 @@
...
+  );
+
+  return {
+    activeItemIds,
+    toggleItemActive,
+    expandSiblings,
+  };
+}
+
+function getAncestorsIds(getItemById: (id: string) => BaseFlatTreeItem, item: BaseFlatTreeItem): string[] {
```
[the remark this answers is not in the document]
YuanboXue-Amber: Question: 
`GetItemById` would be in defined `useTree.tsx` (because `useGetItemById` was moved into `useTree`)
So there's circular reference:
`useTree` imports `useTreeActiveState`
`useTreeActiveState` imports `GetItemById` exported by `useTree`.
How do I deal with it? Should I create a separate type file?

packages/fluentui/react-northstar/src/components/Tree/hooks/useTreeBehavior.tsx
```diff
@@ -0,0 +1,54 @@
...
+        firstChildRef.focus();
+      } else {
+        // when node is leaf, need to focus on the inner treeTitle
+        (firstChildRef.firstElementChild as HTMLElement)?.focus();
+      }
+    },
+    [getItemById, getItemRef],
+  );
+
+  const siblingsExpand = React.useCallback(
```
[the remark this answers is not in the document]
YuanboXue-Amber: Reason for not removing: 
I would like `TreeItem` to call `siblingsExpand` passed down from context. 
Without this, `useTree` hook has no way to know when the '*' key is pressed. 
With this, `useVirtualTree` is able to change `siblingsExpand` to something like below, that sets focus when '*' key is pressed.

```ts
const siblingsExpand = React.useCallback(
    (e: React.KeyboardEvent, id: string) => {
      baseSiblingsExpand(e, id);
      stableSetFocusId(id);
    },
    [baseSiblingsExpand, stableSetFocusId],
  );

```

packages/fluentui/react-northstar/src/components/Tree/hooks/useSelectableTree.tsx
```diff
@@ -0,0 +1,67 @@
...
+    });
+    selectedItemIds.forEach(id => {
+      if (!flatTree[id].hasSubtree) {
+        (flatTree[id] as SelectableFlatTreeItem).selected = true;
+      }
+    });
+
+    // traverse all tree nodes in updatedFlatTree for once,
+    // to calculate the selection state of the parent nodes based on leaf nodes
+    const traverseTree = nodes => {
```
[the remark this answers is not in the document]
YuanboXue-Amber: Just a bit of context of why I do so:

For a tri-state selectable tree, when one tree node is selected/un-selected, the following needs to happen:
For example a node is changed to be selected:
- mark all descendents as selected
- check all ancestor
  - if an ancestor is not selected, put it to `indetermined` state.
  - if an ancestor is `indetermined`, check all its direct children, to know if this ancestor should be changed to selected, or retain its state.
 
I could implement such logic for each node being selected/un-selected. I just find my way of traversing the entire tree for once easier to understand, than the logic above. But I'm open to changes if you guys don't think so.

packages/fluentui/react-northstar/src/components/Tree/Tree.tsx
```diff
@@ -227,253 +137,63 @@ export const Tree: ComponentWithAs<'div', TreeProps> &
...
+      return TreeItem.create(item.item, {
+        defaultProps: () =>
+          getA11yProps('item', {
+            key: id,
+            expanded,
+            parent,
+            level,
+            index,
+            treeSize,
+            selectable: selectable ? item.item.selectable !== false : false,
```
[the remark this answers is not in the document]
YuanboXue-Amber: Like so?

```suggestion
        overrideProps: () =>
          getA11yProps('item', {
            expanded,
            parent,
            level,
            index,
            treeSize,
            selectable: selectable ? item.item.selectable !== false : false,
          }),
```

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.ts
```diff
@@ -0,0 +1,174 @@
...
+    nodes.current[id] = node;
+  }, []);
+  const getItemRef = React.useCallback((id): HTMLElement => nodes.current[id], []);
+
+  // can be used for keyboard navigation ===
+  const focusItemById = React.useCallback(
+    (id: string) => {
+      if (id == null) {
+        return;
+      }
```
layershifter, line 127:
```suggestion
```

According to typings in arguments we cannot have anything except string there, let's avoid unnecessary conditions

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.ts
```diff
@@ -0,0 +1,174 @@
...
+      const itemRef = getItemRef(id);
+      if (itemRef == null) {
```
layershifter, line 129: There should an ESLint rule that prevents non-strict checks, I  would rather go with as it's safer:

```jsx
      if (itemRef instanceof HTMLElement) {
        if (getItemById(id)?.hasSubtree) {
          itemRef.focus();
        } else {
          // when node is leaf, need to focus on the inner treeTitle
          (itemRef.firstElementChild as HTMLElement)?.focus();
        }
      }
```

packages/fluentui/react-northstar/src/components/Tree/hooks/useGetItemById.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function useGetItemById(flatTree: BaseFlatTree): GetItemById {
+  // An exception is thrown there to ensure that a proper callback will assigned to ref
+  const callbackRef = React.useRef<GetItemById>(() => {
+    throw new Error('Callback is not assigned yet');
+  });
+  // We are assigning a callback during render as it can be used during render and in event handlers
+  callbackRef.current = itemId => {
+    return process.env.NODE === 'production' ? flatTree[itemId] : Object.freeze(flatTree[itemId]);
+  };
```
layershifter, line 18:
```suggestion

  // We are assigning a callback during render as it can be used during render and in event handlers. In dev mode we
  // are freezing objects to prevent their mutations
  callbackRef.current = itemId =>
    process.env.NODE === 'production' ? flatTree[itemId] : Object.freeze(flatTree[itemId]);

```

nit, updated comment and made it more readable

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
```diff
@@ -0,0 +1,130 @@
...
+  // returns an extra array of orderedItemIds because flattened tree object does not keep the order of insertion
+  return flattenSubTree(items, 1, SECRET_ROOT_ID, flatTree, true, activeItemIds, []);
+}
+
+function flattenSubTree(
+  items: ObjectShorthandValue<TreeItemProps>[],
+  level: number = 1,
+  parent: string,
+  flatTree: BaseFlatTree,
+  isVisible: boolean = true,
```
layershifter, line 80: Should it be `isParentVisible`?

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (!items) {
+    return { flatTree, visibleItemIds };
+  }
+
+  const itemsInLeaf = items.length;
+
+  items.forEach((item, indexAmongSiblings) => {
+    const { id, items: childrenItems } = item;
+    const hasSubtree = childrenItems ? !!childrenItems.length : false;
+    const expanded = hasSubtree && activeItemIds.includes(id);
```
layershifter, line 93:
```suggestion
    const expanded = hasSubtree && activeItemIds.indexOf(id) !== -1;
```

nit: let's be consistent with other code

packages/fluentui/react-northstar/src/components/Tree/hooks/flattenTree.ts
YuanboXue-Amber: Yes it would make more sense

ling1726 (approved):

layershifter (approved):

packages/fluentui/react-northstar/src/components/Tree/hooks/useTree.ts
```diff
@@ -0,0 +1,286 @@
...
+import { ObjectShorthandCollection } from '../../../types';
+import { TreeItemProps } from '../TreeItem';
+import { FlatTreeItem, flattenTree, FlatTree } from './flattenTree';
+import * as _ from 'lodash';
+
+type GetItemById = (id: string) => FlatTreeItem;
+
+/**
+ * This hook returns a stable `getItemById()` function that will lookup in latest `flatTree`.
+ * This is used used to have stable callbacks that can be passed to React's Context.
```
miroslavstastny, line 12: nit nit nit

```suggestion
 * This is used to have stable callbacks that can be passed to React's Context.
```

miroslavstastny (approved):

PR #15846, opened 2020-11-05
Update content in Themes doc page

xugao:
#### Pull request checklist

#### Description of changes
Updated `/Themes` page from doc site 
- For details about how to apply theme (e.g. `loadTheme` usage), I added link to [this wiki](https://github.com/microsoft/fluentui/wiki/How-to-apply-theme-to-Fluent-UI-React-components)
- For details about `defaultFontStyle`, we have [this page](https://developer.microsoft.com/en-us/fluentui#/styles/web/typography) already which contains the same info.

#### Focus areas to test
Check out deployed site.


ecraig12345:
I'm not sure we should outright remove this--for one thing because people likely have existing links, and also because it makes the content less discoverable for people who are using the website as the main entry point for our documentation. If we want to replace most of the content with links to the wiki for now that might be okay.

I also think we need to preserve the more detailed example of customizing default fonts since this has been a major point of confusion for customers in the past.

xugao:
@ecraig12345 -sorry for the delay. I just resolved your concerns. let me know what you think

packages/react-examples/src/react/Theme/docs/ThemesOverview.md
```diff
@@ -1,69 +1 @@
...
-      fontWeight: 'semibold',
-    },
-    xLarge: {
-      fontSize: '22px',
-      fontWeight: 'semibold',
-    },
-  },
-});
-```
+The entire color palette of the controls is theme-able. We provide a set of sensible defaults, but you can override all colors individually. To do this, you can reference [this wiki](https://github.com/microsoft/fluentui/wiki/How-to-apply-theme-to-Fluent-UI-React-components) for more details.
```
ecraig12345, line 1: Would you mind also adding a link to the font customization docs?

xugao: Updated. I don't want to specifically callout "here is how to override default fonts" because it's too specific of a case. I added the links to the color/typography pages in a more generic way. let me know what you think

packages/react-examples/src/react/Theme/docs/ThemesOverview.md
```diff
@@ -1,69 +1 @@
...
-      fontWeight: 'semibold',
-    },
-    xLarge: {
-      fontSize: '22px',
-      fontWeight: 'semibold',
-    },
-  },
-});
-```
+Fluent UI React components are theme-able. We provide a set of sensible defaults, but you can override [colors](https://developer.microsoft.com/en-us/fluentui#/styles/web/colors/theme-slots) or [fonts](https://developer.microsoft.com/en-us/fluentui#/styles/web/typography) individually. To do this, you can reference [this wiki](https://github.com/microsoft/fluentui/wiki/How-to-apply-theme-to-Fluent-UI-React-components) for more details.
```
ecraig12345, line 1: I think we should use "themeable" not "theme-able"? But not totally sure. Also one other minor suggestion inline.
```suggestion
Fluent UI React components are themeable. We provide a set of sensible defaults, but you can override [colors](https://developer.microsoft.com/en-us/fluentui#/styles/web/colors/theme-slots) or [fonts](https://developer.microsoft.com/en-us/fluentui#/styles/web/typography) individually. See [this wiki](https://github.com/microsoft/fluentui/wiki/How-to-apply-theme-to-Fluent-UI-React-components) for more details.
```

ecraig12345 (approved):
Thanks for the updates! One more suggestion but it's not essential.

packages/react-examples/src/react/Theme/docs/ThemesOverview.md
ecraig12345: People wanting to override the default fonts (mainly to get around issues with the assets license, or various on-prem scenarios) is more common than you might think, which is why we added docs on it. But I can see how it might be better to stick with more generic information.

PR #15847, opened 2020-11-05
Calendar date picker spacing around "Go to today"

### LINKED ISSUE #15651: Calendar/DatePicker "Go to today" padding
The date-time version of Calendar/DatePicker has less bottom padding around the "Go to today" button compared to the old OUFR version. The left/right padding may also be a tiny bit different. Not a functionality issue but it doesn't look so great. (I found this while looking at screener results for #15644.) cc @lorejoh12 

Compare the first [DatePicker example from OUFR](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/7.0/fabric-website-resources/dist/index.html#/examples/datepicker):
image

With the first [DatePicker example from date-time](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/master/date-time/demo/index.html#/examples/datepicker)
image

LeslieW:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15651

#### Description of changes

Currently the spacing around the "Go to today" button varies between the [first DatePicker example from OUFR](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/7.0/fabric-website-resources/dist/index.html#/examples/datepicker) and the [first DatePicker example from date-time](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/master/date-time/demo/index.html#/examples/datepicker). In addition to reduced padding, first DatePicker example from date-time also behaves differently when presented with months of varying lengths. For instance, when August is selected (or any month with 6 weeks due to transitions), the height of the component changes.

Current DatePicker example from date-time:
before_nov
before_aug

This change adds space beneath the button and eliminates variability in overall component height by adding `min-height: 238`. While multiple approaches were considered, this one was selected for its simplicity.

Updated DatePicker example from date-time:
after_nov
after_aug

#### Focus areas to test

(optional)

### DIFF
--- a/change/@fluentui-react-date-time-2020-11-05-10-29-25-calendar_date_picker.json
+++ b/change/@fluentui-react-date-time-2020-11-05-10-29-25-calendar_date_picker.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Update Calendar spacing around \"Go to today\" button.",
+  "packageName": "@fluentui/react-date-time",
+  "email": "lesliewilliams234@gmail.com",
+  "dependentChangeType": "patch",
+  "date": "2020-11-05T18:29:25.460Z"
+}
--- a/packages/react/src/components/Calendar/Calendar.styles.ts
+++ b/packages/react/src/components/Calendar/Calendar.styles.ts
@@ -45,8 +45,7 @@ export const styles = (props: ICalendarStyleProps): ICalendarStyles => {
         boxSizing: 'content-box',
         padding: '0 4px',
         alignSelf: 'flex-end',
-        marginRight: 16,
-        marginTop: 3,
+        margin: '3px 16px 11px 0',
         fontSize: FontSizes.small,
         fontFamily: 'inherit',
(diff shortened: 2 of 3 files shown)


LeslieW:
Just wanted to follow up on this – any recommended next steps? Thanks!

lorejoh12:
So I don't think this works when we have either `isMonthPickerVisible={false}` or `showMonthPickerAsOverlay={true}`. The deploy link isn't working, but I tried it out locally and we end up with a large blank white space underneath the date grid (as the Go to today button floats underneath that when either of the two properties are set like this).

This approach also doesn't take into account the possibility of the control growing or shrinking. If you go to Feb 2026, there's only 4 weeks in the month (which results in an even larger space). May 2026, however, has 6 weeks, so just shrinking the size to accommodate probably won't work either.

It looks like the issue you're actually trying to solve is that there isn't sufficient space underneath the Go to today button. In that case, maybe adjusting the margin on the button itself and letting flex layout handle things would be the way to go here.

PR #15852, opened 2020-11-06
Renaming @fluentui/file-type-icons to @fluentui/react-file-type-icons

dzearing:
Before we release, we should correct the package name to indicate it is react-specific. This enables us to extract out a separate package for other platform support.


apps/public-docsite/src/pages/Styles/FileTypeIconsPage/docs/web/FileTypeIconsImplementation.md
```diff
@@ -19,7 +19,7 @@ When specifying `size`, stick to these default sizes so the images appear as int
...
+Here's a [simple demo](https://github.com/microsoft/fluentui/blob/master/packages/react-examples/src/react-experiments/FileTypeIcon/FileTypeIcon.Basic.Example.tsx) of how to use the [`@fluentui/react-file-type-icons` package](https://github.com/microsoft/fluentui/tree/master/packages/file-type-icons).  The code for the main method you'll use ([`getFileTypeIconProps`](https://github.com/microsoft/fluentui/blob/master/packages/file-type-icons/src/getFileTypeIconProps.ts)) is also well-commented. The icons are kept up-to-date on Fluent UI's CDN, so we recommend not copying or referencing these assets directly.
```
ecraig12345, line 22:
```suggestion
Here's a [simple demo](https://github.com/microsoft/fluentui/blob/master/packages/react-examples/src/react-experiments/FileTypeIcon/FileTypeIcon.Basic.Example.tsx) of how to use the [`@fluentui/react-file-type-icons` package](https://github.com/microsoft/fluentui/tree/master/packages/react-file-type-icons).  The code for the main method you'll use ([`getFileTypeIconProps`](https://github.com/microsoft/fluentui/blob/master/packages/react-file-type-icons/src/getFileTypeIconProps.ts)) is also well-commented. The icons are kept up-to-date on Fluent UI's CDN, so we recommend not copying or referencing these assets directly.
```

apps/public-docsite/src/pages/Styles/FileTypeIconsPage/docs/web/FileTypeIconsImplementation.md
```diff
@@ -1,6 +1,6 @@
-The most durable and future-proof way to get the right icon for a file or item is to use the [file-type-icons package](https://github.com/microsoft/fluentui/tree/master/packages/file-type-icons) from Fluent UI (`@fluentui/file-type-icons` on npm).
+The most durable and future-proof way to get the right icon for a file or item is to use the [file-type-icons package](https://github.com/microsoft/fluentui/tree/master/packages/file-type-icons) from Fluent UI (`@fluentui/react-file-type-icons` on npm).
```
ecraig12345, line 1:
```suggestion
The most durable and future-proof way to get the right icon for a file or item is to use the [react-file-type-icons package](https://github.com/microsoft/fluentui/tree/master/packages/react-file-type-icons) from Fluent UI (`@fluentui/react-file-type-icons` on npm).
```

apps/public-docsite/src/pages/Styles/FileTypeIconsPage/docs/web/FileTypeIconsImplementation.md
```diff
@@ -19,7 +19,7 @@ When specifying `size`, stick to these default sizes so the images appear as int
...
+Here's a [simple demo](https://github.com/microsoft/fluentui/blob/master/packages/react-examples/src/react-experiments/FileTypeIcon/FileTypeIcon.Basic.Example.tsx) of how to use the [`@fluentui/react-file-type-icons` package](https://github.com/microsoft/fluentui/tree/master/packages/file-type-icons).  The code for the main method you'll use ([`getFileTypeIconProps`](https://github.com/microsoft/fluentui/blob/master/packages/file-type-icons/src/getFileTypeIconProps.ts)) is also well-commented. The icons are kept up-to-date on Fluent UI's CDN, so we recommend not copying or referencing these assets directly.
 
 #### Map icons to extensions - [TypeScript mapping file](https://github.com/microsoft/fluentui/blob/master/packages/file-type-icons/src/FileTypeIconMap.ts)
```
ecraig12345, line 24: There are also some github links in this section that need to be updated

packages/react-file-type-icons/CHANGELOG.json
```diff
@@ -1,70 +1,70 @@
 {
-  "name": "@fluentui/file-type-icons",
+  "name": "@fluentui/react-file-type-icons",
   "entries": [
     {
       "date": "Mon, 26 Oct 2020 12:39:59 GMT",
-      "tag": "@fluentui/file-type-icons_v8.0.0-beta.1",
+      "tag": "@fluentui/react-file-type-icons_v8.0.0-beta.1",
```
ecraig12345, line 6: I think we should generally leave the existing entries as they are since we didn't change the existing tags and IIRC these entries won't be used again when generating release notes.

packages/react-file-type-icons/CHANGELOG.md
```diff
@@ -1,33 +1,33 @@
-# Change Log - @fluentui/file-type-icons
+# Change Log - @fluentui/react-file-type-icons
 
 This log was last generated on Mon, 26 Oct 2020 12:39:59 GMT and should not be manually modified.
 
 <!-- Start content -->
 
-## [8.0.0-beta.1](https://github.com/microsoft/fluentui/tree/@fluentui/file-type-icons_v8.0.0-beta.1)
+## [8.0.0-beta.1](https://github.com/microsoft/fluentui/tree/@fluentui/react-file-type-icons_v8.0.0-beta.1)
```
ecraig12345, line 7: Revert all the links in this file too

ecraig12345 (changes requested):

packages/react-file-type-icons/CHANGELOG.json
dzearing: done

packages/react-file-type-icons/CHANGELOG.md
dzearing: done

apps/public-docsite/src/pages/Styles/FileTypeIconsPage/docs/web/FileTypeIconsImplementation.md
dzearing: done.

ecraig12345 (approved):

bigbadcapers:
Hey @dzearing qq for you, just confirming, this will be the new package name for 8.0 and beyond only right? Can you confirm it will remain with the old name in 7?

dzearing:
yes 8 and beyond; we don't want to touch the v7 one to ensure it doesn't break partners.

Full context:

We were renaming the npm scope from @uifabric to @fluentui. Some ODSP partners in IDC have a PR open to tweak the API to be more non-react friendly. We looped in some friends in Graph Toolkit who want to use the file-type-icons in web components. The feedback was:

* Can you make an API that just returns the data we need to render the icon? (e.g. the url)
* Can you separate that API out into a package that doesn't pull in React? That would help with non-React use cases.

So, since we're already renaming the NPM scope, we should also make the package name react-specific. That opens up a couple options for splitting out any of the non-react specific stuff into its own package later on.

PR #15860, opened 2020-11-07
api-docs: Surfacing CompatButton API

### LINKED ISSUE #15600: Include compat api files in api-docs.js
The compat api files for `@fluentui/react` and `@fluentui/react-internal` are not being included in `api-docs.js` because of a problem regarding conflicting names with the non-compat versions. We need to fix this if we want compat types to be able to be pulled up in the website.

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15600

#### Description of changes

This PR makes changes in the `@fluentui/api-docs` package to allow for `compat` APIs to be surfaced and surfaces the `CompatButton` API.

#### Focus areas to test

(optional)


packages/api-docs/src/generatePageJsonFiles.ts
```diff
@@ -1,9 +1,12 @@
+import * as fs from 'fs';
 import * as path from 'path';
 import * as fse from 'fs-extra';
 import { ApiModel } from '@microsoft/api-extractor-model';
 import { IPageJsonOptions } from './types';
 import { generatePageJson } from './pageJson';
 
+const tmpFile = path.join(__dirname, 'tmp.json');
```
ecraig12345, line 8: nit, maybe give this a more descriptive name (also ideally it ought to go under `os.tmpdir()` to be very sure it doesn't accidentally get checked in)

packages/api-docs/src/generatePageJsonFiles.ts
```diff
@@ -23,7 +26,24 @@ export function generatePageJsonFiles(options: IPageJsonOptions): void {
...
-    apiModel.loadPackage(apiJsonPath);
+
+    // If the file belongs to the compat layer.
+    if (apiJsonPath.indexOf('compat') !== -1) {
+      // Get the json info and change the package name.
+      const jsonObject = require(apiJsonPath);
+      jsonObject.name = jsonObject.name + '-compat';
+
+      // Write the info into a temporary file.
+      fs.writeFileSync(tmpFile, JSON.stringify(jsonObject));
```
ecraig12345, line 37: Nit, it's better to use fs-extra (`fse`) here--it has all the same methods plus some improvements, and is already used in this file. Also it has a convenience method `writeJSONSync` which is nice here.

ecraig12345 (approved):
Nice workaround! A couple minor comments but generally looks good.

PR #15874, opened 2020-11-09
UPP+SelectedItemsList: Remove hidden input to reenable drag, add new ctrl-c fix

elisabethcvs:
#### Pull request checklist

#### Description of changes

Previously, the selected items list had a zero size input that had focus when the user was keyboard selected in the list, in order to pick up the ctrl-c copy command. Sadly, this broke drag, as you can't drag the list items when focus was in the input.

This fix removes all that code and adds and overload in the UPP onKeyDown function to pick up the ctrl-c command and run copy from there. Yes, I know this will have localization issues. I can't figure out a way to make this work with the current architecture. Hoping that this will serve as a patch fix while we look at redoing how the list part of this control is constructed.

Other attempts:
- We could remove the focus in the input on click, but then users wouldn't be able to click to select items which would be a confusing user experience
-  Tried adding the contenteditable prop to the persona. This works, but then the text is editable. Which does nothing but confuse the user.

If anyone has any ideas on how to fix this I am very open to it.


sjadelson (approved):

packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
```diff
@@ -244,8 +244,25 @@ export const UnifiedPicker = <T extends {}>(props: IUnifiedPickerProps<T>): JSX.
       props.onKeyDown(ev);
     }
 
+    // Handle copy if focus is in the selected items list
```
nebhatna, line 247: i would just add a comment here mentioning that this is a workaround we are doing till we figure out the right way of doing this, or maybe a WI number, basically something to know that we need to update this in future :)

nebhatna (approved):
Added a minor comment

packages/react-experiments/src/components/UnifiedPicker/UnifiedPicker.tsx
elisabethcvs: Added a comment in the newest iteration

nebhatna (approved):

PR #15880, opened 2020-11-10
U/elisabethcvs/fixdisplaynameregex

### LINKED ISSUE #15878: getInitials should strip [xyz] (angled braces) just like it strips (xyz)
### Environment Information

#### Actual behavior:

RegEx for the code today in
```
const UNWANTED_CHARS_REGEX: RegExp = /\([^)]*\)|[\0-\u001F\!-/:-@\[-`\{-\u00BF\u0250-\u036F\uD800-\uFFFF]/g;

function cleanupDisplayName(displayName: string): string {
  displayName = displayName.replace(UNWANTED_CHARS_REGEX, '');
  displayName = displayName.replace(MULTIPLE_WHITESPACES_REGEX, ' ');
  displayName = displayName.trim();

  return displayName;
}
```

Takes `Kevin Andersen (EEE)` and turns it into `Kevin Andersen` but takes `Kevin Andersen [EEE]` and turns it into `Kevin Andersen EEE` resulting in initials of `KE` in the latter case (undesired).

#### Expected behavior:

Strip the content of [xyz] just like we strip the content of (xyz)
Probably should do the same thing for the curly braces too

elisabethcvs:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15878 

#### Description of changes
 
Currently, getInitials takes Kevin Andersen (EEE) and turns it into Kevin Andersen but takes Kevin Andersen [EEE] and turns it into Kevin Andersen EEE resulting in initials of KE in the latter case (undesired).

I've modified it so that the regex strips [xyz] and {xyz} now as well.

I've also updated:
- the tests for initials, as they previously did not cover these cases
- the Persona initials example to contain the old case and new one
- the persona tests to cover this case as well

I think ideally we would like to backport this to 7, but as it has the potential for breaks, it might be safer just in 8. Open to suggestions here.

### DIFF
--- a/packages/utilities/src/initials.ts
+++ b/packages/utilities/src/initials.ts
@@ -3,7 +3,7 @@
  * The first part matches characters within parenthesis, including the parenthesis.
  * The second part matches special ASCII characters except space, plus some unicode special characters.
  */
-const UNWANTED_CHARS_REGEX: RegExp = /\([^)]*\)|[\0-\u001F\!-/:-@\[-`\{-\u00BF\u0250-\u036F\uD800-\uFFFF]/g;
+const UNWANTED_CHARS_REGEX: RegExp = /[\(\[\{][^\)\]\}]*[\)\]\}]|[\0-\u001F\!-/:-@\[-`\{-\u00BF\u0250-\u036F\uD800-\uFFFF]/g;
 
 /**
  * Regular expression matching phone numbers. Applied after chars matching UNWANTED_CHARS_REGEX have been removed
--- a/packages/utilities/src/initials.ts
+++ b/packages/utilities/src/initials.ts
@@ -1,10 +1,18 @@
 /**
- * Regular expression matching characters to ignore when calculating the initials.
- * The first part matches characters within parenthesis, including the parenthesis.
- * The second part matches special ASCII characters except space, plus some unicode special characters.
+ * Regular expressions matching characters to ignore when calculating the initials.
  */
-// eslint-disable-next-line @fluentui/max-len
-const UNWANTED_CHARS_REGEX: RegExp = /[\(\[\{][^\)\]\}]*[\)\]\}]|[\0-\u001F\!-/:-@\[-`\{-\u00BF\u0250-\u036F\uD800-\uFFFF]/g;
+
+/**
+ * Regular expression matching characters within various types of enclosures, including the enclosures themselves
+ *  so for example, (xyz) [xyz] {xyz} all would be ignored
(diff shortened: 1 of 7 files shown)


packages/utilities/src/initials.ts
```diff
@@ -3,7 +3,8 @@
  * The first part matches characters within parenthesis, including the parenthesis.
  * The second part matches special ASCII characters except space, plus some unicode special characters.
  */
-const UNWANTED_CHARS_REGEX: RegExp = /\([^)]*\)|[\0-\u001F\!-/:-@\[-`\{-\u00BF\u0250-\u036F\uD800-\uFFFF]/g;
+// eslint-disable-next-line @fluentui/max-len
+const UNWANTED_CHARS_REGEX: RegExp = /[\(\[\{][^\)\]\}]*[\)\]\}]|[\0-\u001F\!-/:-@\[-`\{-\u00BF\u0250-\u036F\uD800-\uFFFF]/g;
```
mtennoe, line 7: Hm, the regex becomes quite unreadable after a while. Can we try to split it into sub-regexes with descriptive names, and then use a combination of different regexes instead? That will make it a lot easier to understand and maintain this in the future

elisabethcvs: Would splitting it into two for the brackets and then special characters  work, or were you thinking even more broken down then that

philipzloh: Is the suggestion to move to something like this?

```
const UNWANTED_CHARS_REGEX: RegExp = /[\(\[\{][^\)\]\}]*[\)\]\}]|[\0-\u001F\!-/:-@\[-`\{-\u00BF\u0250-\u036F\uD800-\uFFFF]/g;

====

var openBrackets = "[\(\[\{][^\)\]\}]";
var notCloseBrackets = "[^\)\]\}]";
var closeBrackets = "[\)\]\}]";
var miscSpecialChars = "[\0-\u001F\!-/:-@\[-`\{-\u00BF\u0250-\u036F\uD800-\uFFFF]";

const UNWANTED_CHARS_REGEX = new RegExp(openBrackets + notCloseBrackets + "*" + closeBrackets + "|" + miscSpecialChars, "g");
```

philipzloh: oh Elisabeth's new change is good too

mtennoe: Yup, like the latest change! A lot easier to read :)

mtennoe (approved):

Issue #15853, opened 2020-11-06
Certain interfaces and components are not exported, yet essential


Dalimil:
1) The only way to import `CalloutContentBase` is like this:
`import { CalloutContentBase } from 'office-ui-fabric-react/lib/components/Callout/CalloutContent.base';`
2) The only way to import `PanelBase` is like this:
`import { PanelBase } from 'office-ui-fabric-react/lib/components/Panel/Panel.base';`
3) Same for `IButtonClassNames`
4) Same for `IContextualMenuClassNames`
5) Same for `IMenuItemClassNames`
6) Same for test utilities `expectOne` and `expectMissing`

All of these should properly export these, otherwise deep imports (officially unsupported) are unavoidable.
This blocks transition from office-online-fabric-react to fluentui/react

ecraig12345:
If you need the test utilities, those should probably be added to `@fluentui/test-utilities` and exported from there. (Not sure why that wasn't done to begin with.)

paulgildea:
@Dalimil Thanks for filing this issue.

@behowell / @JustSlone Can we coordinate with @Dalimil to understand their deep import needs and determine the best way forward. I noticed a some dependencies on base components, which leads me to believe that the components `Panel` and `Callout` might be missing something. If so let's see if we can contribute something in.

ecraig12345:
@Dalimil already submitted a PR fixing most of these (thanks!) and we're discussing there.

PR #15854, opened 2020-11-06
Add missing exports for CalloutContentBase, PanelBase and more

Dalimil:
#### Pull request checklist

- [x] Addresses an existing issue: Addresses #15853

#### Description of changes

These files need to be exported to avoid deep imports (incompatible in fluentui/react)

PR #15882, opened 2020-11-10
Add missing exports to API

Dalimil:
#### Pull request checklist

- [x] Addresses an existing issue: Addresses #15853

#### Description of changes

These files need to be exported to avoid deep imports (incompatible in fluentui/react)


Dalimil:
I already made the same change for v7 here: https://github.com/microsoft/fluentui/pull/15854

This is just a follow up to get the same in v8

xugao (approved):

PR #15891, opened 2020-11-10
fix(DropdownSelectedItem): text overflow

assuncaocharles:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15719
- [ ] Include a change request file using `$ yarn change`

#### Description of changes

Agreed with designers on the value for `max-width` for selected items 

Before:

97340922-9fed4500-1841-11eb-9d3a-db01c6bdce13

After:

Screen Shot 2020-11-18 at 17 47 04

#### Focus areas to test

(optional)

PR #15898, opened 2020-11-10
fix(Dropdown): adjust dropdown height

### LINKED ISSUE #15893: Multiple Selection dropdown has additional space after selection list
After the PR
https://github.com/microsoft/fluentui/pull/14864/files#diff-0dfcfdca092d43b516fbc6007112238d042697c91fdced73011fa22d664d65f2R773-R777

Dropdown has additional space after user selection list:
image

Before there was no additional space: 

image

As @assuncaocharles  correctly figured out it is cause by adding "listbox" which is followed by "dropdown" button itself. 
We need to add there "listbox" because of accessibility.

assuncaocharles:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15893

#### Description of changes

In the PR #14864 it was introduced a new `div` to serve as `listbox` for the selected items what excluded the `button` that previously was in the same level as the selected items. This generated regression for dropdown with multiple items selected, this PR aims to fix it by changing the css to looks similar to what we had previously  

#### Focus areas to test

(optional)


ling1726:
Do you mind leaving a screenshot of before/after ?

assuncaocharles:
> Do you mind leaving a screenshot of before/after ?

It's on the issue linked, if you think it's necessary I can duplicate it here

packages/fluentui/react-northstar/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -140,7 +140,8 @@ export const dropdownStyles: ComponentSlotStylesPrepared<DropdownStylesProps, Dr
 
   selectedItems: ({ props: p, variables: v }): ICSSInJSStyle => ({
     display: 'flex',
-    flexWrap: 'wrap',
+    minHeight: pxToRem(28),
+    marginBottom: pxToRem(4),
```
ling1726, line 144: Why do you need these values ? setting the width and height of the trigger button should already extend the button down the right hand side of the dropdown

ling1726: I understand the bottom margin since it will cut off the some of the chips, but why the min-height ?

packages/fluentui/react-northstar/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -149,6 +150,9 @@ export const dropdownStyles: ComponentSlotStylesPrepared<DropdownStylesProps, Dr
 
   triggerButton: ({ props: p, variables: v }): ICSSInJSStyle => {
     return {
+      marginBottom: pxToRem(-4),
```
ling1726, line 153: Do we want there to be a small unclickable area in that button since it is a part of the dropdown ? without the margin the button height will extend the height of the parent container

packages/fluentui/react-northstar/src/themes/teams/components/Dropdown/dropdownStyles.ts
assuncaocharles: To keep the current height otherwise it will be smaller than before

ling1726: Makes sense, thanks

ling1726 (dismissed):

packages/fluentui/react-northstar/src/themes/teams/components/Dropdown/dropdownStyles.ts
assuncaocharles: So that is tricky, because what I am trying to achieve is to have less visual regression as possible but if we gonna say that having small regression is fine since we are changing the DOM I would be ok with removing this

ling1726: Yeah this isn't a show stopper for me... it's quite a minor thing I noticed playing with it. I'm not entirely sure about the regression scenario so you could ping someone who's been around longer about that.

Other than that looks good to me, nice catch on the flex-wrap :)

kolaps33:
I tried the fix with the JAWS/NVDA and I cannot see any regression. Looks good to me.

kolaps33 (approved):

pompomon (approved):

layershifter:
Verified, it works as expected. The single thing that looks strange is the blinking button:

dblink

layershifter (approved):

assuncaocharles:
> Verified, it works as expected. The single thing that looks strange is the blinking button:
> 
> dblink

Yeah, that is strange but it's not related to this change. We should change some amount of things to match dropdown with design spec anyway

assuncaocharles:
P.S: We have some screen regressions related to text overflow that was introduced in #15891

ling1726 (approved):

packages/fluentui/react-northstar/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1548,17 +1550,17 @@ export const Dropdown: ComponentWithAs<'div', DropdownProps> &
                 onClick={search && !open ? handleContainerClick : undefined}
               >
                 <div ref={selectedItemsRef} className={cx(dropdownSlotClassNames.selectedItems, classes.selectedItems)}>
+                  {!search && renderTriggerButton(getToggleButtonProps)}
```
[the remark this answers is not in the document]
assuncaocharles: Done

specs/Breadcrumb.md
```diff
@@ -18,12 +18,11 @@ A Breadcrumb trail consists of a list of links to the parent pages of the curren
 
 ## Props
 
-
 ## &lt;Breadcrumb&gt; Properties <a href="#breadcrumb-divider-properties" id="breadcrumb-divider-properties"></a>
 
-| Property Name | Type             | Default Value | Description                                          |
-| ------------- | ---------------- | ------------- | ---------------------------------------------------- |
-| `size`        | `enum`           | `medium`      | The Breadcrumb size                                  |
+| Property Name | Type   | Default Value | Description         |
```
assuncaocharles, line 23:
```suggestion
| Property Name | Type               | Default Value | Description                                            |
```

specs/Breadcrumb.md
```diff
@@ -18,12 +18,11 @@ A Breadcrumb trail consists of a list of links to the parent pages of the curren
 
 ## Props
 
-
 ## &lt;Breadcrumb&gt; Properties <a href="#breadcrumb-divider-properties" id="breadcrumb-divider-properties"></a>
 
-| Property Name | Type             | Default Value | Description                                          |
-| ------------- | ---------------- | ------------- | ---------------------------------------------------- |
-| `size`        | `enum`           | `medium`      | The Breadcrumb size                                  |
+| Property Name | Type               | Default Value | Description                                            |
```
assuncaocharles, line 23:
```suggestion
| Property Name | Type            | Default Value | Description                                         |
```

specs/Breadcrumb.md
```diff
@@ -18,12 +18,11 @@ A Breadcrumb trail consists of a list of links to the parent pages of the curren
 
 ## Props
 
-
 ## &lt;Breadcrumb&gt; Properties <a href="#breadcrumb-divider-properties" id="breadcrumb-divider-properties"></a>
 
-| Property Name | Type             | Default Value | Description                                          |
-| ------------- | ---------------- | ------------- | ---------------------------------------------------- |
-| `size`        | `enum`           | `medium`      | The Breadcrumb size                                  |
+| Property Name | Type            | Default Value | Description                                         |
```
assuncaocharles, line 23:
```suggestion
| Property Name | Type             | Default Value | Description                                          |
```

specs/Breadcrumb.md
```diff
@@ -18,12 +18,11 @@ A Breadcrumb trail consists of a list of links to the parent pages of the curren
 
 ## Props
 
-
 ## &lt;Breadcrumb&gt; Properties <a href="#breadcrumb-divider-properties" id="breadcrumb-divider-properties"></a>
```
assuncaocharles, line 21:
```suggestion

## &lt;Breadcrumb&gt; Properties <a href="#breadcrumb-divider-properties" id="breadcrumb-divider-properties"></a>
```

specs/Breadcrumb.md
```diff
@@ -22,8 +22,8 @@ A Breadcrumb trail consists of a list of links to the parent pages of the curren
...
 
 | Property Name | Type             | Default Value | Description                                          |
-| ------------- | ---------------- | ------------- | ---------------------------------------------------- |
-| `size`        | `enum`           | `medium`      | The Breadcrumb size                                  |
+| ------------- | ------ | ------------- | ------------------- |
```
assuncaocharles, line 25:
```suggestion
| ------------- | ---------------- | ------------- | ---------------------------------------------------- |
```

specs/Breadcrumb.md
```diff
@@ -23,7 +23,7 @@ A Breadcrumb trail consists of a list of links to the parent pages of the curren
 
 | Property Name | Type             | Default Value | Description                                          |
 | ------------- | ---------------- | ------------- | ---------------------------------------------------- |
-| `size`        | `enum`           | `medium`      | The Breadcrumb size                                  |
+| `size`        | `enum` | `medium`      | The Breadcrumb size |
```
assuncaocharles, line 26:
```suggestion
| `size`        | `enum`           | `medium`      | The Breadcrumb size                                  |
```

PR #15902, opened 2020-11-10
SelectedPersona: Fix hover/active states in normal and high contrast modes

salaman:
#### Pull request checklist

#### Description of changes

Fixed and cleaned up SelectedPersona styles to have more consistent hover/active states in normal and HC themes.

##### Screenshots

Left: before, right: after

Normal:
image image

Hover:
image image

Selected:
image image

Selected + hover:
image image

Invalid:
image image

Invalid + hover:
image image

Invalid + selected:
image image

Normal + high contrast:
image image

Hover + high contrast:
image image

Selected + high contrast:
image image

Selected + hover + high contrast:
image image

Invalid + high contrast:
image image

Invalid + hover + high contrast:
image image

Invalid + selected + high contrast:
image image

### DIFF
--- a/change/@fluentui-react-experiments-2020-11-10-11-24-56-upp-hover-states.json
+++ b/change/@fluentui-react-experiments-2020-11-10-11-24-56-upp-hover-states.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "SelectedPersona: Fix hover/active states in normal and high contrast modes",
+  "packageName": "@fluentui/react-experiments",
+  "email": "chrp@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-11-10T19:24:56.067Z"
+}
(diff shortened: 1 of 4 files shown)


elisabethcvs:
Thanks for putting all the screenshots, definitely helps. I think for the invalid hover+selected we probably want to keep the red text. Not sure if it meets contrast ratios but I think the info is important. Maybe there's a lighter/darker version?

elisabethcvs (changes requested):
Comment

salaman:
@elisabethcvs the current OWA style doesn't preserve the red in selected state (presumably because it looks hideous against the blue) and I'm not sure it's worth the added complexity to have a different colour. There's also no `redLight`, only `red` and `redDark`. If you feel strongly for it I can add it back but it doesn't visually look good and I expect fails the contrast ratio check.

elisabethcvs:
Nope if that's the existing default behavior then that sounds good

elisabethcvs (approved):

PR #15914, opened 2020-11-11
use react-button in docs

xugao:
use `react-button` and `react-theme-provider` in northstar docs; get screener running


ecraig12345:
Closing since we're no longer using this approach

Issue #14691, opened 2020-08-21
Add version 8 to website (all versions)


JustSlone:
(from @ecraig12345)

#### Any time before release
- [ ] Look into making same website release pipeline tasks work across 5/6/7/master (task group with variable?)
- [ ] Low priority: Look into driving version switcher from manifest or something (to avoid needing manual updates across multiple files and branches) -- would be nice if there's an *easy* way to do it, but not essential

#### During the final pre-release freeze:
- [x] Update version switching everywhere to include 8 (and use by default)
  - [x] version switcher in React homepage (in 5, 6, 7, master)
    - This is at `apps/fabric-website/src/pages/HomePage/HomePage.base.tsx` in 6/7/master; not sure where in 5
  - [x] `apps/fabric-website/index.html` (master, forget if it's needed in 5/6/7)
  - [ ] `apps/fabric-website/homepage.htm` (master only; delete or ignore in other branches)
  - [ ] others?
- [ ] Enable website release of 8 from master

PR #15917, opened 2020-11-11
Website: Adding version switcher across the website

khmakoto:
#### Pull request checklist

- [x] Addresses an existing issue: Part of #14691

#### Description of changes

This PR adds a version switcher across the website instead of only having it on the front page:

image

This PR also adds version 8 were required.


packages/react-docsite-components/src/components/Page/Page.types.ts
```diff
@@ -112,6 +112,22 @@ export interface IPageProps<TPlatforms extends string = string> {
...
+   */
+  versionOptions?: IContextualMenuItem[];
+
+  /**
+   * Callback that determines what happens when a version is chosen from the version dropdown.
+   */
+  onVersionMenuClick?: (event: React.MouseEvent<HTMLElement, MouseEvent>, item: IContextualMenuItem) => void;
+
+  /**
+   * The data regarding the current version of the library.
```
ecraig12345, line 127: Might be good to make clear that this is a package.json object. Or even better, pull out and pass through only the needed info ahead of time.

packages/react-docsite-components/src/components/Page/Page.types.ts
```diff
@@ -112,6 +112,22 @@ export interface IPageProps<TPlatforms extends string = string> {
 
   /** Currently selected platform. */
   platform?: TPlatforms;
+
+  /**
+   * The available versions whose documentation is presented on the website.
+   */
+  versionOptions?: IContextualMenuItem[];
```
ecraig12345, line 119: Wondering if it would be better to bundle these under a single property. Also if it might be better to just pass the list of versions (or possibly key/text pairs) and let the dropdown rendering code handle converting to menu items, and have the callback just take the chosen version number as a parameter.

packages/react-docsite-components/src/components/PageHeader/PageHeader.module.scss
```diff
@@ -0,0 +1,5 @@
+@import '../../styles/common';
+
+.root {
+  width: calc(100% - #{$SideRail-width});
```
ecraig12345, line 4: Would be good to check what happens with the dropdown on mobile. (The site may totally be a mess on mobile right now but I do see some styling for things like hiding the side rail, which would also mess with the dropdown position.)

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
```diff
@@ -39,15 +53,51 @@ const getStyles: IStyleFunction<IPageHeaderStyleProps, IPageHeaderStyles> = prop
...
+    currentVersionData,
+    onVersionMenuClick,
+    pageTitle = 'Page title',
+    pageSubTitle,
+    theme,
+    versionOptions,
+  } = props;
   const styles = getClassNames(getStyles, { theme });
 
+  const monoFont =
```
ecraig12345, line 66: What was the thinking behind having this in a mono font? I'm not so sure about that, though if we keep it you should probably reuse the `codeFontFamily` variable from `CodeSnippet.styles.ts` to keep things consistent.

packages/react-docsite-components/src/components/PageHeader/PageHeader.module.scss
ecraig12345: Hmm this definitely comes out looking funny on the main controls page, at least on a smaller desktop screen. 
image

I'm also not seeing it even on a specific control page.
image

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
khmakoto: Well, I was just using the same styling that we have in the front page, happy to remove this if needed.

packages/react-docsite-components/src/components/PageHeader/PageHeader.module.scss
khmakoto: You should now be seeing it in every component page as well, although I still need to address the styling.

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
khmakoto: Removed.

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
```diff
@@ -39,15 +58,47 @@ const getStyles: IStyleFunction<IPageHeaderStyleProps, IPageHeaderStyles> = prop
...
+      : undefined;
 
   return (
-    <header className={css(styles.root, props.className)}>
+    <header className={styles.root}>
       <h1 className={styles.title}>
         {pageTitle}
         {pageSubTitle && <span className={styles.subTitle}>{pageSubTitle}</span>}
       </h1>
+      {versionSwitcherDefinition && versionOptions && (
```
ecraig12345, line 79: We should probably hide this on mobile, at least phones. Not exactly sure how to handle the breakpoints here, since within the UHF "mobile" range (up to 768px) the dropdown is bad on the smaller end and fine on the larger end (then bad again up to around 900px).

Medium-sized phone (375px):
image

Tablet vertical (767px--this is just below the UHF "mobile" breakpoint of 768px, since exactly 768px seems to have some bugs, unrelated to your PR)
image

800px
image

800px with abbreviated version
image

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
```diff
@@ -39,15 +58,47 @@ const getStyles: IStyleFunction<IPageHeaderStyleProps, IPageHeaderStyles> = prop
...
+        <ActionButton
+          allowDisabledFocus={true}
+          styles={{
+            root: { height: '1em', padding: '12px 0' },
```
ecraig12345, line 83: Not set on this but I think it would look better to have the button text be the same color as "Web"

Also the vertical alignment is off
image

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
```diff
@@ -39,15 +58,47 @@ const getStyles: IStyleFunction<IPageHeaderStyleProps, IPageHeaderStyles> = prop
...
-  const styles = getClassNames(getStyles, { theme });
+  const { className, pageTitle = 'Page title', pageSubTitle, theme, versionSwitcherDefinition } = props;
+  const styles = getClassNames(getStyles, { className, theme });
+
+  const versionOptions: IContextualMenuItem[] | undefined =
+    versionSwitcherDefinition && versionSwitcherDefinition.versions
+      ? versionSwitcherDefinition.versions.map(version => ({
+          key: version,
+          text: version,
+          checked: version === versionSwitcherDefinition.currentVersion,
```
ecraig12345, line 69: I don't think the `checked` logic is working

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
khmakoto: This should be solved now.

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
khmakoto: I think it's working, it just isn't doing anything with it, nor is it doing anything with it in the front page either. Should we just remove it since it's not doing anything either way?

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
```diff
@@ -1,23 +1,42 @@
...
   return {
-    root: {
-      position: 'relative',
-      marginBottom: appPaddingSm,
-      selectors: {
-        [`@media screen and (min-width: ${ScreenWidthMinUhfMobile}px)`]: {
-          marginBottom: 0,
-          padding: `${appPaddingLg}px 0`,
-          minHeight: pageHeaderFullHeight,
+    root: [
```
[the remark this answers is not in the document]
ecraig12345: Hmm I see what you mean...looking at the master/7 PR deploy site to avoid interference by UHF styling, in the [7.0 version](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/7.0/fabric-website/dist/index.html?devhost#/controls/web/button) it's fine but [in master](http://fabricweb.z5.web.core.windows.net/pr-deploy-site/refs/heads/master/public-docsite/index.html?devhost#/controls/web/button) there's the extra space. I guess I'll look into it sometime before release.

packages/react-docsite-components/src/components/PageHeader/PageHeader.tsx
ecraig12345: You mean the ActionButton isn't using it? If so let's just remove it.

ecraig12345 (approved):

PR #15949, opened 2020-11-16
Fixes #15729: add type button to textfield of type password

### LINKED ISSUE #15729: fluent ui input with canRevealPassword will trigger post if it's in a form
### Environment Information

- **Package version(s)**: (7.147.1)
- **Browser and OS versions**: (win 10)

#### Actual behavior:

when input control is used in a form, the 'eye ball' to reveal password, which is turned on by adding canRevealPassword, will listen to the submit event of the form and trigger post

#### Expected behavior:

'eye ball' will not listen to submit event.

### Priorities and help requested: medium, I can walk around using keypress event

Are you willing to submit a PR to fix? (Yes, No) Yes, if guidance is provided

Requested priority: (Blocking, High, Normal, Low) normal

zemirco:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15729

#### Description of changes

Currently pressing the "reveal password" button submits the form if the TextField is inside a form. This is because the default button type inside a form is `submit`. By setting it to `button` explicitly we prevent the form from being submitted.


ecraig12345:
Thanks for doing this! Looks like you need to update snapshots in a couple different places. Full steps for doing that (starting from the root of the repo):
```
yarn buildto react-examples
cd packages/react-internal
yarn update-snapshots
cd ../react-examples
yarn update-snapshots
```

Also, just a heads up that if you'd like this change in version 7, you'll need to make an additional PR with the same change against the `7.0` branch (PRs to `master` will only affect the version 8 beta releases).

ecraig12345 (approved):

PR #15952, opened 2020-11-16
makeStyles: hook produced should take optional options object, rather than specific args

dzearing:
Current code does this:

```tsx
const useStyles = makeStyles({ ... });

const Foo = (p) => {
  const styles = useStyles(p.theme, p.renderer);
  return <div className={styles.foo} />;
};
```
We should avoid having the hook take in specific arguments and instead take in an options bag. This lets us extend the input in the future as needed, which might be required in the case of passing along classnames which should override the map:

```tsx
const Foo = (p) => {
  const styles = useStyles({ theme: p.theme, renderer: p.renderer });
  return <div className={styles.foo} />;
};
```


khmakoto (approved):

xugao (approved):
LGTM! could you make the same change in 7.0 branch?

PR #15959, opened 2020-11-17
Creating an applyClasses helper and moving logic out of makeClasses.

dzearing:
Rather than burying the "auto class distribution" logic in the `makeClasses` hook, I've moved it into a very testable, standalone helper which now can be used in conjunction with `makeStyles`, thus reducing the need to have `makeClasses`. Before I proceed to remove `makeClasses`, need to use this directly in `makeVariantClasses`.

I'd like to get the step 1 here checked in however, and then proceed with additional work as a step 2 in a separate PR.

Usage of `applyClasses`:

```jsx
const useStyles = makeStyles({
  root: { ... },
  _primary: { ... } // styles when primary is applied
  _size_small: { ... } // styles when size equals value "small"
});

const Button = (p) => {
  const state = { ...p };
  const styles = useStyles();

  // Automatically mix in root/slot classes and conditionally apply modifiers and
  // enums to the root className based on state.
  applyClasses(state, styles);

 return <button { ...state } />
};
```

### DIFF
--- a/change/@fluentui-react-theme-provider-2020-11-17-09-18-25-feat-apply-classes.json
+++ b/change/@fluentui-react-theme-provider-2020-11-17-09-18-25-feat-apply-classes.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Moving the auto-distribution logic out of makeClasses and into an applyClasses helper.",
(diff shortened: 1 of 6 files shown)


xugao (approved):
LGTM. Please cherry-pick the change to 7.0 branch.

PR #15973, opened 2020-11-18
Add AriaDescription for ContextMenu items

### LINKED ISSUE #15510: Allow ariaDescription as a prop on IContextualMenuItem

SavannahBourgeois:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15510

#### Description of changes
There are a few props that can be set on items to help give screen reader users more context. One of those is aria-DescribedBy which is a string of IDs that point to elements describing the item. The other is aria-Description which is often set as the description of the element or is placed into a hidden span and the ID of that span is added to the list of IDs in aria-DescribedBy. This second approach is better because it will merge the aria-DescribedBy IDs and concat them to be the aria-Description set on the element so you do not overwrite other aria-DescribedBy elements.

With this change, I add ariaDescribedBy and ariaDescription as options on IContextualMenuItem. If the user passes ariaDescription, we will generate a hidden span with the contents of that prop and set the ID of that span to the list of aria-DescribedBy.

I have added an example of different context menu types with ariaDescriptions and verified they are read out by the screen reader when "Hear advanced detail" is selected.

#### Focus areas to test
Contextual menus, keytips on contextual menus, screen reader

I already checked in this change for Master, this is adding it to V7 (https://github.com/microsoft/fluentui/pull/15770)


packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.types.ts
```diff
@@ -539,6 +539,16 @@ export interface IContextualMenuItem {
    */
   [propertyName: string]: any;
 
+  /**
+   * Detailed description of the menu item for the benefit of screen readers.
+   */
+  ariaDescription?: string;
+
+  /**
```
ThomasMichon, line 547: `ariaDescribedBy` should not be *passed in* to `IContextualMenuItem`. This is because the 'element which describes the item' will the `ContextualMenuButton`'s own `Tooltip`. In fact, `title` already is used for a tooltip, so we really should avoid overloading `ariaDescription` here as well. Between `text`, `ariaLabel`, and `title`, we already have all fields covered.

ThomasMichon (changes requested):

packages/office-ui-fabric-react/src/components/ContextualMenu/ContextualMenu.types.ts
SavannahBourgeois: Thanks for the input, sing title for the tooltip is not really accessible and not an option for us. Here is a wiki explaining why you should avoid using title (https://sarahmhigley.com/writing/tooltips-in-wcag-21/) and accessibility issues with browser tooltips.  The current model with text, ariaLabel and title does not support our needs in Office. We have some controls that have a title displayed, and then have additional information in a tooltip and these items can be in context menus. We are currently passing the tooltip to aria-label but that means the actual text in the menu item is not read out which means screen reader users get a different experience than non-screen reader users. Happy to discuss more!

SavannahBourgeois: @ThomasMichon I may have misunderstood here. Are you suggesting I just remove the ariaDescribedBy passed in and leave the ariaDescription? Also in Office we don't use FluentUI's tooltips for a variety of reasons so we can't rely on the FluentUI Tooltip setting anything for us.

SavannahBourgeois:
@ThomasMichon or @joschect can one of you take a look please? I would love to be able to get this in soon.

joschect (approved):

ThomasMichon (approved):

joschect (approved):

PR #15976, opened 2020-11-18
Support themed react-button without ThemeProvider

xugao:
#### Pull request checklist

#### Description of changes
1. Ensure tokens inside theme are populated/mapped in default theme by moving `getTokens` logic out from `ThemeProvider` into `mergeTheme`/`createTheme`.
2. Updated variants definitions for Buttons so they are by default themed without using `ThemeProvider` or `loadTheme`. The downside is we cannot directly using the css variables for global tokens when defining variants for Buttons. Instead need to directly access values from theme.
3. Updated Button stories and screener tests to not wrap with `ThemeProvider` by default.

#### Focus areas to test

(optional)


packages/theme/src/utilities/getTokens.ts
```diff
@@ -29,32 +29,32 @@ export function getTokens(theme: Theme, userTokens?: RecursivePartial<Tokens>):
...
           },
 
           disabled: {
             background: semanticColors.primaryButtonBackgroundDisabled,
             contentColor: semanticColors.buttonTextDisabled,
             dividerColor: palette.neutralTertiaryAlt,
-            secondaryContentColor: 'var(--color-brand-disabled-contentColor)',
-            borderColor: 'var(--color-brand-disabled-background)',
-            iconColor: 'var(--color-brand-disabled-contentColor)',
+            secondaryContentColor: semanticColors.buttonTextDisabled,
```
dzearing, line 55: shouldn't this be `brand?.disabled?contentColor`

dzearing (approved):

khmakoto (approved):

PR #15977, opened 2020-11-18
Allow for custom overflow button in breadcrumb with correct props

### LINKED ISSUE #15881: Breadcrumb provides onRender for items but doesn't respect it for overflow items
### Environment Information

- **Package version(s)**: 7.123.9

The breadcrumb component provides a `onRenderItem` that you can use to render custom content on the item itself, but as soon as the item goes in the overflow menu the `onRenderItem` is not respected. I would expect that either the same method is used for rendering the overflow items or the API should provide an `onRenderOverflownItem` to be able to do this.

In our app we use React-router, and therefore we need to use their custom `Link` component for internal navigation, so in this case the overflow items are not aware of the router and reload the whole page. Apart from this, the styling needs to be written twice (for items and overflown items) and the focus-visible doesn't respect styles for non-href items (The same way that the mouse provides styles for)

Are you willing to submit a PR to fix? **Yes**

Requested priority: High

Products/sites affected: Azure IoT Central

PatoBeltran:
#### Pull request checklist

- [x] Addresses an existing issue: Fixes #15881

#### Description of changes

While working on breadcrumb I noticed two issues:

- The `IContextualMenuItem` type mentions that `name` is deprecated and we should use `text` instead, but when trying to render custom overflow items I noticed that the breadcrumb was transforming my `text` to `name` on each item and removing `text`, so I added it so we don't have to rely on deprecate props.
- I need a way to render custom overflow menu items (to render router aware links instead of normal <a> tags), since other components (CommandBar and OverflowSet) have a prop to render a custom overflow button, I created my own `OverflowMenu` component that uses a `IconButton` with it's own `onRenderOverflowItem` to allow this. While working with the breadcrumb I noticed there's no way to do this, since the current prop (`onRenderOverflowIcon`) only allows you to render a custom icon, but not the wrapper button, so if I use this to render my `OverflowMenu` component I end up with a button with a menu inside a button with a menu. Here I'm adding a prop that matches exactly the prop in the CommandBar to allow for this.

Sorry that I didn't open an issue first, since it's a simple change I thought it would be faster just to send a PR and discuss any chances needed here.

This is a bit related to https://github.com/microsoft/fluentui/issues/15881

### DIFF
--- a/change/@fluentui-react-2020-11-18-12-29-27-pb-breadcrumb-fixes.json
+++ b/change/@fluentui-react-2020-11-18-12-29-27-pb-breadcrumb-fixes.json
@@ -1,0 +1,8 @@
+{
+  "type": "prerelease",
+  "comment": "Add text to breacrumb overflow items",
+  "packageName": "@fluentui/react",
+  "email": "pbelsal@gmail.com",
+  "dependentChangeType": "patch",
+  "date": "2020-11-18T20:29:27.909Z"
+}
--- a/packages/react/etc/react.api.md
+++ b/packages/react/etc/react.api.md
@@ -383,6 +383,7 @@ export interface IBreadcrumbProps extends React.HTMLAttributes<HTMLElement> {
     onRenderItem?: IRenderFunction<IBreadcrumbItem>;
     onRenderOverflowIcon?: IRenderFunction<IButtonProps>;
     overflowAriaLabel?: string;
+    overflowButtonAs?: IComponentAs<IButtonProps>;
     overflowIndex?: number;
     // (undocumented)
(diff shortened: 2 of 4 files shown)


khmakoto (approved):
You need to run `yarn change` to generate a change file and you should also probably build `@fluentui/react` so that the API file is updated. Other than that though, the actual change looks good 👍

xugao (approved):
LGTM. In case you don't know, this change will be published under `@fluentui/react@v8-beta` version. if you need this change to be published under `@fluentui/react@v7`, pls cherry-pick the change to `7.0` branch as well.

khmakoto:

